PR #61, opened 2021-09-22
Integration tests using puppeteer

devinea:
WIP


packages/fiori-freestyle/test/integration/__tests__/app.tests.ts
```diff
@@ -0,0 +1,347 @@
...
+        });
+        throw e;
+    }
+    await expect(page).toMatchElement(config.selectorKey, {
+        text: config.listPageTitle
+    });
+    await expect(page.title()).resolves.toMatch(config.title);
+};
+
+jest.setTimeout(200000);
```
IainSAP, line 52: Is this a debugging timeout that can be removed?

packages/fiori-freestyle/test/integration/__tests__/app.tests.ts
```diff
@@ -0,0 +1,347 @@
...
+        rmdirSync(reportDir, { recursive: true });
+        mkdirSync(reportDir);
+    });
+    describe('allTemplate/listdetail', () => {
+        test('npm run start - allTemplate/listdetail - should display Fiori page', async () => {
+            await checkApp({
+                name: 'npm run start',
+                appId: 'testme',
+                listSelectorKey: '[id="__item2-__clone0-titleText-inner"]',
+                selectorKey: '[id="application-testme-app-component---list--listPageTitle-inner"]',
```
IainSAP, line 64: Do we need to use selectors that include the app id? Only one app can be loaded. If we dont tie to dynamic values there tests can be re-used?

IainSAP (commented):
Couple of queries

packages/fiori-freestyle/test/integration/__tests__/app.tests.ts
devinea: Yeah should be able to make this more general. Trying to find what ones are common first

packages/fiori-freestyle/test/integration/__tests__/app.tests.ts
devinea: Probably still needed, but will look to make it lower if possible.

vobu:
just a general remark on the subject of UI5 IDs and test runners: we (aka j&s, my company) faced a similar challenge that `UIveri5` didn't fulfill our requirements in terms of integration tests for a UI5 app.
So we went ahead and integrated Webdriver.IO with UI5's test API (`sap.ui.test.RecordReplay.*`) and calles it `wdi5`: https://github.com/js-soft/wdi5
It also comes in a lightweight, browser-scope only variant: https://github.com/js-soft/wdi5/blob/develop/wdio-ui5-service/README.md
With it you can do:
```js
const oList = {
  selector: {
    // sap.ui.test.RecordReplay.ControlSelector: same as OPA5 or UIveri5
    id: 'UI5control_ID',
    viewName: 'your.namespace.App'
  }
}
// UIveri5-compatible .asControl()
const list = browser.asControl(oList)
// native UI5 methods!
list.getItems()
...
```
Not trying to sell/advertise, just mentioning alternatives to pupeteer.

vobu:
oh, and i forgot to mention that you can run `wdio`/`wdi5` programmatically in test runners such as `jest`. 
example: https://github.com/js-soft/wdi5/blob/5d718448c0fd66316e78fcb0c240690cefb36c39/test/wdi5-multiversion.js#L38

devinea:
@vobu Thanks for the info. puppeteer is quick and easy for now. We might swap it out later when we have OPA added to the generated projects. Long term plan still to be decided.

nlunets (changes requested):
First of all, thanks for working on this, it works in certain situation which is quite nice.

Looking at the code i think there are some things that could be changed and improved.

I could run the test on Mac without problem but couldn't get them to work on windows (multiple issues with port already used and killing them didn't seem to work), nor on WSL on windows (problem during the globalSetup but this might be due to the WSL setup itself and its lack of X server).
On windows the biggest issue is that the test not running was not failing the command so this could lead to un-noticeable issues.

While getting the test running on mac i noticed that they are difficult to follow since the starting of the server happens asynchronously at the same time as the tests are waiting for them. We also start all server from the get go which seems a bit heavy (18 of them !) especially since they also might need to download some stuff, maybe it would make sense to either, load the server on demand, or find a way to server everything from one common location ? From a debugging perspective it might make it easier if the test were responsible for their server startup... 

I also got some telemetry issues in the log which didn't look very nice NOENT: no such file or directory, open 'ux-ui5-tooling@1.3.5/node_modules/@sap/ux-ui5-tooling/package.json

As for the content of the test itself, it's relatively basic for sure and only guarantees that the application component can be started which is not guaranteeing much about the app itself, but this is something that can be improved later on.

Maybe if the goal is that this is all, more or less automated, you should consider pushing that idea to the end and also make it so that the test definition gets generated and then used in a generic way, maybe not directly reusing the unit test (because that introduce an awkward dependency between the two) but with some dedicated script to prepare the integration tests.

Finally, process wise you may want to add some information in the readme on how to run this and also maybe how new ones should be defined if there are new templates ?

devinea:
@nlunets 
I am coming back to this again finally.

> but couldn't get them to work on windows (multiple issues with port already used and killing them didn't seem to work), nor on WSL on windows (problem during the globalSetup but this might be due to the WSL setup itself and its lack of X server).

WSL for sure won't work due to lack of X server as you say. Windows builds seem to be working fine and running locally on windows. Could it be a clash with software you have installed? 

> On windows the biggest issue is that the test not running was not failing the command so this could lead to un-noticeable issues.

I haven't been able to re-create. Any tips?

> While getting the test running on mac i noticed that they are difficult to follow since the starting of the server happens asynchronously at the same time as the tests are waiting for them. We also start all server from the get go which seems a bit heavy (18 of them !) especially since they also might need to download some stuff, maybe it would make sense to either, load the server on demand, or find a way to server everything from one common location ? From a debugging perspective it might make it easier if the test were responsible for their server startup...

Yeah I think this is something that may evolve over time, but is ok for now. The 18 servers don't use much resources so ok the time being, obviously it won't scale longer term.
In tool-suite we start the servers in the test, but that set-up is slow and difficult to run tests concurrently.
In this set-up the server logs are hard to tell which server it belongs too. Will look into improving that in a later PR.

> I also got some telemetry issues in the log which didn't look very nice NOENT: no such file or directory, open 'ux-ui5-tooling@1.3.5/node_modules/@sap/ux-ui5-tooling/package.json

I think that issue has being fixed in telemetry the up coming release. 

> As for the content of the test itself, it's relatively basic for sure and only guarantees that the application component can be started which is not guaranteeing much about the app itself, but this is something that can be improved later on.

💯 

Maybe if the goal is that this is all, more or less automated, you should consider pushing that idea to the end and also make it so that the test definition gets generated and then used in a generic way, maybe not directly reusing the unit test (because that introduce an awkward dependency between the two) but with some dedicated script to prepare the integration tests.

> In tool-suite we also generate the test definition dynamically. The boiler plate is less for sure. But running and debugging individual tests then becomes much more awkward as the jest tooling doesn't support that well. So want to opt for something easier to work with here as the cost of being more verbose. 


> Finally, process wise you may want to add some information in the readme on how to run this and also maybe how new ones should be defined if there are new templates ?

Will do.

unseen1980 (approved):

packages/odata-service-writer/src/index.ts
```diff
@@ -91,11 +91,11 @@ async function generate(basePath: string, data: OdataService, fs?: Editor): Prom
     );
 
     const mockserverMiddleware = getMockServerMiddlewareConfig(data);
```
IainSAP, line 93: Think this loc can be move to here also? https://github.com/SAP/open-ux-tools/pull/61/commits/308c4659c65d02ea5f7a346cfcd022b0e7a408f3#diff-11060fa3e2b31c89f7a31e60dfe2c764b0f9b32db732156920d506581236a4c4R98

PR #63, opened 2021-09-23
update(FF): Updates FF Worklist controllers and tests

IainSAP:
Adds odata version agnostic controllers for FF worklist + test


packages/fiori-freestyle/src/index.ts
```diff
@@ -9,13 +9,13 @@ import {
...
     Template,
-    Package,
-    OdataVersion
+    Package
 } from '@sap/open-ux-tools-types';
 import { TemplateType } from '@sap/open-ux-tools-types'; // This is an enum dont import as type, we lose runtime values
 import { UI5Config } from '@sap/ux-ui5-config';
 import { getPackageTasks } from '@sap/open-ux-tools-common';
 import { getUI5Libs } from './data/ui5Libs';
+import { cloneDeep } from 'lodash';
```
devinea, line 18: `import cloneDeep from 'lodash/cloneDeep';`

devinea (commented):
@IainSAP - Can you update the description https://github.com/SAP/open-ux-tools/pull/63#issue-741577309? Not v4 specific any more.

packages/fiori-freestyle/templates/worklist/add/webapp/controller/Worklist.controller.js
```diff
@@ -128,8 +126,13 @@ sap.ui.define([
...
 		 */
 		_showObject : function (oItem) {
-			this.getRouter().navTo("object", {
-				objectId: oItem.getBindingContext().getProperty("<%=template.settings.entity.key%>")
+			var that = this;
+
+			oItem.getBindingContext().requestCanonicalPath().then(function (sObjectPath) {
+				that.getRouter().navTo("object", {
+					objectId_Old: oItem.getBindingContext().getProperty("ID"),
+					objectId : sObjectPath.slice("/Products".length) // /Products(3)->(3)
```
tobiasqueck, line 134: This doesn't look right. I would expect a templating string instead of `Products`

packages/fiori-freestyle/templates/worklist/add/webapp/controller/Worklist.controller.js
```diff
@@ -100,7 +98,7 @@ sap.ui.define([
 				var sQuery = oEvent.getParameter("query");
 
 				if (sQuery && sQuery.length > 0) {
-					aTableSearchState = [new Filter("<%=template.settings.entity.idProperty%>", FilterOperator.Contains, sQuery)];
+					aTableSearchState = [new Filter("ID", FilterOperator.Contains, sQuery)];
```
tobiasqueck, line 101: Shouldn't the property match the ID property of the chosen entity? Or is `ID` always valid no matter the entity.
If not, this also needs to be looked at in in 133

packages/fiori-freestyle/templates/worklist/add/webapp/controller/Object.controller.js
```diff
@@ -113,17 +99,19 @@ sap.ui.define([
...
 
-			var oResourceBundle = this.getResourceBundle(),
-				oObject = oView.getBindingContext().getObject(),
-				sObjectId = oObject.<%=template.settings.entity.key%>,
-				sObjectName = oObject.<%=template.settings.entity.idProperty%>;
+			var oResourceBundle = this.getResourceBundle();
+
+			oView.getBindingContext().requestObject().then((function (oObject) {
+				var sObjectId = oObject.ID,
+					sObjectName = oObject.relatedProduct_ID;
```
tobiasqueck, line 106: Templating string instead of `relatedProduct_ID`?

packages/fiori-freestyle/templates/worklist/add/webapp/controller/Object.controller.js
```diff
@@ -64,12 +62,7 @@ sap.ui.define([
 		 */
 		_onObjectMatched : function (oEvent) {
 			var sObjectId =  oEvent.getParameter("arguments").objectId;
-			this.getModel().metadataLoaded().then( function() {
-				var sObjectPath = this.getModel().createKey("<%=template.settings.entity.name%>", {
-					<%=template.settings.entity.key%> :  sObjectId
-				});
-				this._bindView("/" + sObjectPath);
-			}.bind(this));
+			this._bindView("/Products" + sObjectId);
```
tobiasqueck, line 65: Templating string instead of `Products`?

tobiasqueck (changes requested):
Please see the inline comments. It looks like a handful of copy & paste issues happened in which templating strings have been replaced with hardcoded values.

tobiasqueck:
Not required but if you add `client: 100` to the test config at https://github.com/SAP/open-ux-tools/blob/feature/v4_controllers_worklist/packages/fiori-freestyle/test/worklist.test.ts#L55 then the generated app works when you keep it using the DEBUG mode.

tobiasqueck (dismissed):
* reviewed the fix of the clone method
* validated that the generated worklist templates work for both v2/v4
* didn't validate that the removal of the escaping in the ejs templates works as expected but discussed with @IainSAP and here explained that it is needed
* ran the tests

All in all, it looks very good and is a great step forward simplifying our templates.

packages/fiori-freestyle/package.json
```diff
@@ -19,19 +19,21 @@
...
-        "mem-fs": "^2.1.0",
-        "mem-fs-editor": "^9.0.0",
-        "@sap/ux-ui5-application-template": "workspace:*",
+        "@sap/open-ux-tools-common": "workspace:*",
         "@sap/ux-odata-service-template": "workspace:*",
+        "@sap/ux-ui5-application-template": "workspace:*",
         "@sap/ux-ui5-config": "workspace:*",
-        "@sap/open-ux-tools-common": "workspace:*"
+        "ejs": "^3.1.6",
+        "lodash.clonedeep": "4.5.0",
```
devinea, line 27:
```suggestion
        "lodash": "4.17.21",
```
I think "lodash.clonedeep" is an older approach.

packages/fiori-freestyle/src/index.ts
```diff
@@ -9,13 +9,13 @@ import {
...
     Template,
-    Package,
-    OdataVersion
+    Package
 } from '@sap/open-ux-tools-types';
 import { TemplateType } from '@sap/open-ux-tools-types'; // This is an enum dont import as type, we lose runtime values
 import { UI5Config } from '@sap/ux-ui5-config';
 import { getPackageTasks } from '@sap/open-ux-tools-common';
 import { getUI5Libs } from './data/ui5Libs';
+import cloneDeep from 'lodash.clonedeep';
```
devinea, line 18:
```suggestion
import cloneDeep from 'lodash/cloneDeep';
```

devinea (approved):
Changes look good. 
Backward compatibility with tool-suite checked elsewhere by @tobiasqueck

packages/fiori-freestyle/src/index.ts
ullasholla: @devinea, I could not find any indication that `lodash.deepClone` is the older approach. They still seem to have the link to the per method packages in their doco: https://www.npmjs.com/search?q=keywords:lodash-modularized

Looking at the way they build the packages (https://lodash.com/custom-builds) and also comparing the dates of the last changes (2 years ago), both `lodash.deepClone` and `lodash/deepClone` seem to be the same implemenation.

Package sizes are different though. `lodash.deepClone` is 45kb while `lodash` is 1.4 MiB. Treeshaking in our webpack config might pare it down to what's needed but do you think we should still use `lodash/deepClone`?

ullasholla: They also have this: https://www.npmjs.com/package/lodash-es

devinea (dismissed):
Thanks @ullasholla

devinea:
Some git conflicts

tobiasqueck (approved):
Approved (again) since the additional change only took care of merges to `package.json` and  `pnpm-lock.yaml`

packages/fiori-freestyle/templates/worklist/add/webapp/controller/Object.controller.js
```diff
@@ -19,37 +20,34 @@ sap.ui.define([
...
-
 		/* =========================================================== */
 		/* event handlers                                              */
 		/* =========================================================== */
 
 
 		/**
-		 * Event handler for navigating back.
+		 * Event handler  for navigating back.
+		 * It there is a history entry we go one step back in the browser history
```
unseen1980, line 39: Typo here? Probably meant ``` If there is...```

packages/fiori-freestyle/templates/worklist/add/webapp/controller/Object.controller.js
```diff
@@ -19,37 +20,34 @@ sap.ui.define([
...
 		 * @public
 		 */
 		onNavBack : function() {
-			this.getRouter().navTo("worklist", {}, true);
+			var sPreviousHash = History.getInstance().getPreviousHash();
+			if (sPreviousHash !== undefined) {
+				// eslint-disable-next-line sap-no-history-manipulation
+				history.go(-1);
+			} else {
+				this.getRouter().navTo("worklist", {}, true);
```
unseen1980, line 49: Is it correct that in one case we use the Router for navigation and in the other we directly manipulating the history? Couldn't we use the router for both?

unseen1980 (approved):

packages/fiori-freestyle/templates/worklist/add/webapp/controller/Object.controller.js
nlunets: I think this is correct as the router doesn't provide the abstraction necessary for this.
In case there is already a history stack this means that we come from the worklist so our history looks like
`[worklist, detailPage]`
At that point doing a navTo without replace context would have us end up with
`[worklist, detailPage, worklist]` which is not what we want
Doing a navTo with replace would end up with
`[worklist, workslist]` which is also not what we want.

So history.go(-1) is the only way to get to the right state.

PR #64, opened 2021-09-23
Add Node V16 build and compatibility

devinea:
- Adapt code due to API change in node 16.
https://github.com/nodejs/node/pull/37216/files#diff-11c13307cb334211e0d11f4e7e43bdc0a34e437779a63a6ff645e62448d973d2R1053-R1055

- Only write files to disk if UX_DEBUG environment variable is set when running unit tests.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        node-version: [12.x, 14.x]
+        node-version: [12.x, 14.x, 16.x]
     
     runs-on: ${{ matrix.os }}
     timeout-minutes: 15
--- a/packages/fiori-freestyle/test/basic.test.ts
+++ b/packages/fiori-freestyle/test/basic.test.ts
@@ -1,6 +1,6 @@
 import { FreestyleApp, generate, TemplateType } from '../src';
 import { join } from 'path';
-import { rmdirSync } from 'fs';
+import { removeSync } from 'fs-extra';
 import { testOutputDir, debug } from './common';
 import { OdataVersion } from '@sap/open-ux-tools-types';
 import { sample } from './sample/metadata';
@@ -55,7 +55,7 @@ describe(`Fiori freestyle template: ${TEST_NAME}`, () => {
     ];
 
     beforeAll(() => {
-        rmdirSync(curTestOutPath, { recursive: true });
+        removeSync(curTestOutPath);
     });
 
     test.each(configuration)('Generate files for template: $name', async ({ name, config }) => {
--- a/packages/fiori-freestyle/test/basic.test.ts
+++ b/packages/fiori-freestyle/test/basic.test.ts
@@ -55,16 +55,22 @@ describe(`Fiori freestyle template: ${TEST_NAME}`, () => {
     ];
 
     beforeAll(() => {
-        removeSync(curTestOutPath);
+        removeSync(curTestOutPath);  // even for in memory
     });
 
     test.each(configuration)('Generate files for template: $name', async ({ name, config }) => {
         const testPath = join(curTestOutPath, name);
         const fs = await generate(testPath, config);
         expect((fs as any).dump(testPath)).toMatchSnapshot();
-        // write out the files for debugging
+
         return new Promise((resolve) => {
-            fs.commit(resolve);
+					// write out the files for debugging
+					if (debug?.enabled) {
+						fs.commit(resolve);
+					} else
+					{
+						resolve(true)
+					}
         });
     });
 });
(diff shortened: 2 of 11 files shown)


tobiasqueck (changes requested):
@devinea instead of using a new library to delete the debug folder before all tests, could we instead change the debugging approach and only commit the folders to the file system if debug mode is enabled? 
It would remove the requirement to delete anything (btw I think that was also the initial implementation of template tests by @ullasholla )

devinea:
@tobiasqueck that is sort of a separate issue. In debug mode the dir stills needs to be deleted. Fs api's have a breaking change so it is easier to switch to fs-extra or rimraf for consistent API across versions.

tobiasqueck:
While I agree that this is kind of a separate issue, I saw it while reviewing this code and, I don't think it is a good idea to fix an issue that only exists because of another deeper issue: we have unit tests that rely on the file system. To be precise, not the tests itself rely on it but test preparation method that do things that are not related to the tests but for a very different problem: generating artifacts to manually review them.

> In debug mode the dir stills needs to be deleted

No, there shouldn't be the need to delete anything because the tests will not write anything to the filestystem and they also don't rely  on the file system. Only the debug mode would commit/write data to the fs, so that a developer can review. Under a normal (not debug) run nothing is ever written.

devinea:
@tobiasqueck - Based on our discussion I have added the UX_DEBUG env to control writing the test-output or not via fs-commit. Since the test folders should be deleted anyway when UX_DEBUG is either true or false I've kept `fs-extra` in place for that.

.github/workflows/pipeline.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        node-version: [12.x, 14.x]
+        node-version: [12.x, 14.x, 16.x]
```
devinea, line 14: @kranthie-sap @donal-tobin-sap - Do we want to limit the windows build to only 1 node version?

tobiasqueck (dismissed):
- checked that it works even if the `test-output` folder doesn't exist
- reviewed the tests and verified that no data is written to the file system if one is not running in debug mode
- looked at general coding style and all looks good to me
- I did not check that data is written to the filesystem if it runs in debug mode

.github/workflows/pipeline.yml
kranthie-sap: @devinea: Let's go with all 3 versions on all platforms for now. If the pipeline takes way too long to run, we can look at other options like running node v16 (non LTS version) on a scheduled run.

unseen1980 (dismissed):

packages/fiori-freestyle/test/basic.test.ts
```diff
@@ -55,16 +55,22 @@ describe(`Fiori freestyle template: ${TEST_NAME}`, () => {
...
         expect((fs as any).dump(testPath)).toMatchSnapshot();
-        // write out the files for debugging
+
         return new Promise((resolve) => {
-            fs.commit(resolve);
+					// write out the files for debugging
+					if (debug?.enabled) {
+						fs.commit(resolve);
+					} else
+					{
```
ullasholla, line 71:
```suggestion
					} else {
```
Minor change to maintain consistency with other formatting.

packages/ui5-application/test/options.test.ts
```diff
@@ -28,7 +29,14 @@ describe('UI5 templates', () => {
...
-        fs.commit(() => 0)
         expect((fs as any).dump(projectDir)).toMatchSnapshot();
+        return new Promise((resolve) => {
+            // write out the files for debugging
+            if (debug) {
+                fs.commit(resolve);
+            } else {
+                resolve(true);
+            }
+        });
```
ullasholla, line 40: This is good! Anyway we can abstract this as It seems to be required in most of the tests? Could always do it later.

ullasholla:
> @tobiasqueck that is sort of a separate issue. In debug mode the dir stills needs to be deleted. Fs api's have a breaking change so it is easier to switch to fs-extra or rimraf for consistent API across versions.

@devinea , what's the breaking change in `rmdirSync` again? Curious because this might affect tools-suite too.

ullasholla (changes requested):
Changes look good, thanks.

There's one minor formatting change required (there's a suggestion that can be committed directly, if you're ok with it).

There's a comment about abstracting common code in the tests. Up to you how you handle this.

I'm curious about the breaking change in the `fs` API though. This might have an impact on our other code. Mind posting a link or two about it please?

devinea:
@ullasholla 

> There's one minor formatting change required (there's a suggestion that can be committed directly, if you're ok with it).

Applied. Thanks. 

> There's a comment about abstracting common code in the tests. Up to you how you handle this.

Yeah makes sense. Can look at that in another PR.

> I'm curious about the breaking change in the fs API though. This might have an impact on our other code. Mind posting a link or two about it please?

See here https://github.com/nodejs/node/pull/37216/files#diff-11c13307cb334211e0d11f4e7e43bdc0a34e437779a63a6ff645e62448d973d2R1053-R1055

tobiasqueck (approved):
Approved again. Only formatting changes since my initial review.

ullasholla (approved):
Thanks @devinea, looks good.

PR #65, opened 2021-09-24
feat: enhancements as result of easy-ui5 integration

vobu:
this PR is based on using `fiori-freestyle` for scaffolding a basic UI5 app in `easy-ui5`. 
(more precise, `fiori-freestyle`'s `generate()`-API is used  in the `generator-template-ui5-project` sub-generator to scaffold the UI5 app)

as a result, the following enhancements were introduced to `open-ux-tools`/`fiori-freestyle`:
- dynamically set initial view- and controller-name at scaffolding time: https://github.com/vobu/open-ux-tools-1/commit/bb8a2b508b58321e7cea51ec5f7993b6e4720f61
- same for manifest.json: https://github.com/vobu/open-ux-tools-1/commit/1aef09425831277760531d4a1d4dbac1a016f572
- provide CDN- and local resources for bootstrapping UI5 in `index.html`: https://github.com/vobu/open-ux-tools-1/commit/c451411f908539c5a46e6a8c14d13c7ec965249f

Also, a small regex fix is included: https://github.com/vobu/open-ux-tools-1/commit/86a9cb4c4974d3607af53aea2da6caf63d0ed3cc

### DIFF
--- a/packages/fiori-freestyle/src/index.ts
+++ b/packages/fiori-freestyle/src/index.ts
@@ -106,7 +106,7 @@ function copyTemplates(tmplPath: string, basePath: string, ffApp: FreestyleApp<u
         return path
             .replace(new RegExp(replaceVer), '') // remove version template path
             .replace(/\$ViewName/, ffApp.ui5?.initialViewName || "View1") // dynamically set inital view file name
-            .replace(/\$ControllerName/, ffApp.ui5?.initialControllerName || "View1") // dynamically set inital controller file name
+            .replace(/\$ControllerName/, ffApp.ui5?.initialControllerName || "View1"); // dynamically set inital controller file name
     };
     // Ignore other odata version specific template folders
     const ignoreFolderPattern = Object.values(OdataVersion).filter((ver) => ver !== ffApp.service?.version);
--- a/packages/fiori-freestyle/src/index.ts
+++ b/packages/fiori-freestyle/src/index.ts
@@ -102,7 +102,12 @@ async function generate<T>(basePath: string, data: FreestyleApp<T>, fs?: Editor)
 function copyTemplates(tmplPath: string, basePath: string, ffApp: FreestyleApp<unknown>, fs: Editor) {
     // Remove odata versions specific path while copying template files
     const replaceVer = `\\${sep}v${Object.values(OdataVersion).join(`|\\${sep}v`)}`;
-    const removeVersionTmplPath = (path: string): string => path.replace(new RegExp(replaceVer), '');
+    const destPathManipulations = (path: string): string => {
+        return path
+            .replace(new RegExp(replaceVer), '') // remove version template path
+            .replace(/\$ViewName/, ffApp.ui5?.initialViewName || "View1") // dynamically set inital view file name
+            .replace(/\$ControllerName/, ffApp.ui5?.initialControllerName || "View1") // dynamically set inital view file name
+    };
     // Ignore other odata version specific template folders
     const ignoreFolderPattern = Object.values(OdataVersion).filter((ver) => ver !== ffApp.service?.version);
     // By template type
@@ -113,7 +118,7 @@ function copyTemplates(tmplPath: string, basePath: string, ffApp: FreestyleApp<u
         {},
         {
             globOptions: { ignore: ignoreFolderPattern.map((folder) => `**/v${folder}/**`) },
-            processDestinationPath: removeVersionTmplPath
+            processDestinationPath: destPathManipulations
         }
     );
 }
--- a/packages/fiori-freestyle/src/index.ts
+++ b/packages/fiori-freestyle/src/index.ts
@@ -106,7 +106,7 @@ function copyTemplates(tmplPath: string, basePath: string, ffApp: FreestyleApp<u
         return path
             .replace(new RegExp(replaceVer), '') // remove version template path
             .replace(/\$ViewName/, ffApp.ui5?.initialViewName || "View1") // dynamically set inital view file name
-            .replace(/\$ControllerName/, ffApp.ui5?.initialControllerName || "View1") // dynamically set inital view file name
+            .replace(/\$ControllerName/, ffApp.ui5?.initialControllerName || "View1") // dynamically set inital controller file name
     };
     // Ignore other odata version specific template folders
     const ignoreFolderPattern = Object.values(OdataVersion).filter((ver) => ver !== ffApp.service?.version);
--- a/packages/fiori-freestyle/templates/basic/add/webapp/Component.js
+++ b/packages/fiori-freestyle/templates/basic/add/webapp/Component.js
@@ -1,7 +1,7 @@
 sap.ui.define([
 		"<%=app.baseComponent%>",
 		"sap/ui/Device",
-		"<%=app.id.replace('.', '/')%>/model/models"
+		"<%=app.id.replace(/\./g, '/')%>/model/models"
 	],
     function (UIComponent, Device, models) {
         "use strict";
--- a/packages/types/src/ui5App.ts
+++ b/packages/types/src/ui5App.ts
@@ -22,6 +22,11 @@ export interface App {
 	localStartFile?: string;
 }
 
+export enum BootstrapSrc {
+	CdnOpenUI5 = "Content delivery network (OpenUI5)",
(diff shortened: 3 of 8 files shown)


vobu:
will add some tests backing up the introduced enhancements, so converting this to `draft` until

packages/ui5-application/templates/core/webapp/index.html
```diff
@@ -12,7 +12,10 @@
     </style>
     <script
         id="sap-ui-bootstrap"
-        src="resources/sap-ui-core.js"
+        src="<%= ui5.bootstrapSrc === 'Content delivery network (OpenUI5)' ?
+            'https://openui5.hana.ondemand.com/resources/sap-ui-core.js' :
+                ui5.bootstrapSrc === 'Content delivery network (SAPUI5)' ?
+                    'https://sapui5.hana.ondemand.com/resources/sap-ui-core.js' : 'resources/sap-ui-core.js'  %>"
```
devinea, line 18: This approach is not really compatible with Fiori tools. 
We put the host the ui5*.yaml and proxy with https://www.npmjs.com/package/@sap/ux-ui5-tooling e.g.  cdn host is https://ui5.sap.com. This allows switching ui5 versions with the ui5*.yaml or launch config for preview
```
  customMiddleware:
    - name: fiori-tools-proxy
      afterMiddleware: compression
      configuration:
        ignoreCertError: false
        backend:
          - path: /sap
            url: https://somehost:44300
            client: "500"
            destination: ECXCLNT500
        ui5:
          path:
            - /resources
            - /test-resources
          url: https://ui5.sap.com
          version: ""
```
CDN can be passed into export const `getFioriToolsProxyMiddlewareConfig` to use a value other on than the default.
Can this approach work for easy-ui5 also?

devinea (commented):
@vobu - changes look good for the templates. Question on cdn below.

packages/ui5-application/templates/core/webapp/index.html
vobu: > This approach is not really compatible with Fiori tools.

given the type def...
```
export enum BootstrapSrc {
	CdnOpenUI5 = "Content delivery network (OpenUI5)",
	CdnSAPUI5 = "Content delivery network (SAPUI5)",
	Local = "Local Resources"
}
```
..., doing `ui5.bootStrapSrc = BootstrapSrc.Local` or simply leaving it `undefined`, this would result in 
```js
<script
        id="sap-ui-bootstrap"
        src="resources/sap-ui-core.js
// ...
>
```
which should be fine with the fiori tools proxy, right?

btw: the fiori tools proxy to the ui5 cdn has a problem in its' default definition anyway: given the default config you mentioned above...
```
  customMiddleware:
    - name: fiori-tools-proxy
      afterMiddleware: compression
      configuration:
        ignoreCertError: false
        backend:
          - path: /sap
            url: https://somehost:44300
            client: "500"
            destination: ECXCLNT500
        ui5:
          path:
            - /resources <-- this is a problem
            - /test-resources
          url: https://ui5.sap.com
          version: ""
```
...proxying `/resources` to the cdn prevents any ui5 tooling module shims from working. 
the shims need `resources/path/to/shim` working for the ui5 tooling.
so a solution could be to only proxy `sap` resources to the cdn:
```
 ui5:
          path:
            - /resources/sap*
            - /test-resources
```

tobiasqueck: @vobu do you have an example project with a module shim? I am wondering if this is easier solvable by running the middleware after a different middleware i.e. after the module shims are resolved.

Otherwise, the change to `resources/sap*` looks reasonable as well but requires some testing on our side first.

packages/types/src/ui5App.ts
```diff
@@ -22,6 +22,11 @@ export interface App {
 	localStartFile?: string;
 }
 
+export enum BootstrapSrc {
+	CdnOpenUI5 = "Content delivery network (OpenUI5)",
+	CdnSAPUI5 = "Content delivery network (SAPUI5)",
+	Local = "Local Resources"
+}
```
tobiasqueck, line 29: I am not in favor of this approach. This looks like a question for the prompting and based on that the template should be called with the location/url of the UI5 sources. Such an approach would also allow to point to a specific SAP system for testing with its UI5 version.

Would this work for `easyUI5`?

packages/ui5-application/templates/core/webapp/index.html
vobu: there's a whole thread with examples here: https://answers.sap.com/questions/13461279/ui5-project-shims-not-working.html?childToView=13479438

packages/fiori-freestyle/templates/basic/add/webapp/Component.js
```diff
@@ -1,7 +1,7 @@
 sap.ui.define([
 		"<%=app.baseComponent%>",
 		"sap/ui/Device",
-		"<%=app.id.replace('.', '/')%>/model/models"
+		"<%=app.id.replace(/\./g, '/')%>/model/models"
```
tobiasqueck, line 4: Oha ... that is bad bug

packages/types/src/ui5App.ts
vobu: actually the enum is aligned with exactly such a prompt in easy-ui5:
https://github.com/ui5-community/generator-ui5-project/blob/b7ebdfcefd0dd022a9c1eacf796dd3646bbf9368/generators/app/index.js#L64

packages/fiori-freestyle/src/index.ts
```diff
@@ -102,7 +102,12 @@ async function generate<T>(basePath: string, data: FreestyleApp<T>, fs?: Editor)
 function copyTemplates(tmplPath: string, basePath: string, ffApp: FreestyleApp<unknown>, fs: Editor) {
     // Remove odata versions specific path while copying template files
     const replaceVer = `\\${sep}v${Object.values(OdataVersion).join(`|\\${sep}v`)}`;
-    const removeVersionTmplPath = (path: string): string => path.replace(new RegExp(replaceVer), '');
+    const destPathManipulations = (path: string): string => {
+        return path
+            .replace(new RegExp(replaceVer), '') // remove version template path
```
tobiasqueck, line 107: Allowing to provide a custom view/controller name sounds good. I would like to see the defaulting in one place i.e. https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-application/src/data/defaults.ts#L51 instead of in all `ejs` templates and here.
Btw, would you add, both view and controller name as parameter or wouldn't it be better style to just ask for the view name and make sure the controller name matches? I am aware that it is technically possible to have different names but I am wondering if we should enforce best practices.

tobiasqueck:
All in all, I see one major bug you fixed and a set of reasonable requests to enhance the API. I will create corresponding feature requests and bug reports which I will send you for review. For each, I would like to see a separate PR for better tracking.

Btw, https://github.com/SAP/open-ux-tools/pull/63 is supposed to go today (or this week) simplifying the `fiori-freestyle` template leading to incompatible changes with your PR that need to be revisited.

vobu:
right on, then i'll move to stand-by and wait what's coming my way after #63.
only then will i refactor the easy-ui5 integration.

packages/types/src/ui5App.ts
tobiasqueck: I meant, would it also work for `easyUI5` to keep their prompts and then convert it into the target url for UI5? The template could then use the provided url in the proxy middleware.

packages/fiori-freestyle/templates/basic/add/webapp/Component.js
tobiasqueck: @vobu here is a report of the issue you found: https://github.com/SAP/open-ux-tools/issues/70. Would you mind fixing it in a separate PR, so that we can keep it separate?

vobu: will do this morning

vobu: &rarr; https://github.com/SAP/open-ux-tools/pull/77

tobiasqueck:
@vobu can we close this PR? I think the PRs #77, #158 and #144 covered everything outlined in #78 which was defined based on your PR.

vobu:
yep, closing it.
thx for the cooperation on all thingies so far!!!

PR #77, opened 2021-09-29
fix: app-id creation at scaffolding time in Component.js (fixes #70)

### LINKED ISSUE #70: [Fiori Freestyle] The generated Component.js in the basic template has an invalid import if its app Id contains more than one dot
### **Bug Description**
If you use the `fiori-freestyle` templates - basic template and generate a project while providing an application ID with more than one dot e.g. my.demo.App then the model import in the generated `Component.js` is broken because only the first`.` is replaced with a `/`.

### **Expected Behavior**
The `Component.js` works and all the dots are properly replaced.

### **Steps to Reproduce**
1. Modify the configuration in `test/basic.test.ts` to use `my.demo.App` as `config>app>id` value
2. Run the test
3. Check the test `test-output/basicTemplate/**/Component.js`
4. See see issue in the imports at the beginning of the file

### **Context**
- OS/Platform: any
- Browser: independent
- Affected component: `fiori-freestyle`

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

### **Stakeholder Info** (if applicable)

Reported by @vobu

vobu:
replace all occurrences of "." with "/" in `Component.js` template

### DIFF
--- a/packages/fiori-freestyle/templates/basic/add/webapp/Component.js
+++ b/packages/fiori-freestyle/templates/basic/add/webapp/Component.js
@@ -1,7 +1,7 @@
 sap.ui.define([
 		"<%=app.baseComponent%>",
 		"sap/ui/Device",
-		"<%=app.id.replace('.', '/')%>/model/models"
+		"<%=app.id.replace(/\./g, '/')%>/model/models"
 	],
     function (UIComponent, Device, models) {
         "use strict";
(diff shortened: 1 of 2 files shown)


tobiasqueck (approved):
- [x] project installs, build and tests are running
- [x] reported issue is fixed
- [x] code is readable and linted

The same issue can also be found at https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-application/src/data/defaults.ts#L37, however, this part of the code contains a few other issues, therefore, it needs to be reported as a separate issue.

unseen1980 (approved):
Thanks 👍

PR #88, opened 2021-10-08
fix: break circular dependencies

ullasholla:
This PR breaks circular dependencies between `ui5-config` and `odata-service-template`, and `ui5-config` and `open-ux-tools-types`.

It replaces the types used in the `ui5-config` tests with the types from the `open-ux-tools-types` module.


tobiasqueck (approved):
I am generally not in favor of moving things into a central types project instead of defining them where they belong, however, it looks like the enum exists twice already and this is just changing the the reference from one to another, therefore, I am ok with this change.

PR #90, opened 2021-10-09
fix(free-style): remove change_loader.js and refs

ullasholla:
This PR removes `webapp/test/change_loader.js` and references to it in `webapp/test/flpSandbox.html`.


unseen1980 (dismissed):
Thanks @ullasholla

Issue #93, opened 2021-10-11
Service client is not properly formed in package.json task outputs


IainSAP:
### **Bug Description**
Setting the client of the service type omits the 'sap-client' query param key from the package.json tasks.

### **Expected Behavior**
The correctly formed tasks are output as so :
```
{
...
              "start": "fiori run --open 'test/flpSandbox.html?sap-client=100#testApp-tile'",
              "start-local": "fiori run --config ./ui5-local.yaml --open 'test/flpSandbox.html?sap-client=100#testApp-tile'",
              "start-mock": "fiori run --config ui5-mock.yaml --open 'test/flpSandbox.html?sap-client=100#testApp-tile'",
              "start-noflp": "fiori run --open 'index.html?sap-client=100'",
...
}
```

### **Steps to Reproduce**
Generate a Fiori Freestyle app setting the `service.client` value.

### **Context**
- OS/Platform: {*}
- Browser: {*}
- Affected component: fiori-freestyle, common

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

IainSAP:
Done

PR #94, opened 2021-10-11
fix(common): Write correct sap-client format

IainSAP:
https://github.com/SAP/open-ux-tools/issues/93


packages/fiori-freestyle/test/__snapshots__/worklist.test.ts.snap
```diff
@@ -6929,12 +6929,12 @@ archive.zip
...
   },
   \\"scripts\\": {
-    \\"start\\": \\"fiori run --open 'test/flpSandbox.html#wrk1-tile'\\",
+    \\"start\\": \\"fiori run --open 'test/flpSandbox.html?sap-client=123#wrk1-tile'\\",
     \\"build\\": \\"ui5 build --config=ui5.yaml --clean-dest --dest dist\\",
     \\"deploy-config\\": \\"fiori add deploy-config\\",
-    \\"start-local\\": \\"fiori run --config ./ui5-local.yaml --open 'test/flpSandbox.html#wrk1-tile'\\",
-    \\"start-noflp\\": \\"fiori run --open 'index.html'\\",
-    \\"start-mock\\": \\"fiori run --config ui5-mock.yaml --open 'test/flpSandbox.html#wrk1-tile'\\"
+    \\"start-local\\": \\"fiori run --config ./ui5-local.yaml --open 'test/flpSandbox.html?sap-client=123#wrk1-tile'\\",
```
devinea, line 6935: Assume this sap-client is ignored for this app in preview as it is not a real ABAP backend.

packages/common/package.json
```diff
@@ -11,7 +11,9 @@
         "lint:fix": "eslint . --ext .ts --fix",
         "watch": "tsc --watch",
         "link": "pnpm link --global",
-        "unlink": "pnpm unlink --global"
+        "unlink": "pnpm unlink --global",
+        "test": "jest --ci --forceExit --detectOpenHandles --colors",
+        "test-u": "jest --ci --forceExit --detectOpenHandles --colors -u"
```
ullasholla, line 16: Can we remove this script and use `pnpm run test -- -u` instead when we need to update the snapshots? (We have the script in all our modules, so it's a general observation). I'd rather constrain the number of scripts we have in our `package.json` files. This also ensures that when we update snapshots it's done intentionally and not accidentally (for example, I have autocomplete on my shell and this script makes it easy to inadvertently update snapshots).
```suggestion

```

packages/common/test/utils.test.ts
```diff
@@ -0,0 +1,34 @@
+import { getPackageJsonTasks } from '../src/utils';
+
+describe('Test common utils', () => {
```
ullasholla, line 3: There are 3 tests inside this `test()` method. Would it be possible to split them into 3 and add descriptive comments so that it's clear which usecase is being tested?

packages/common/src/utils/index.ts
```diff
@@ -17,3 +18,49 @@ export function getFilePaths(dir: string): string[] | [] {
     } catch (err) {}  // eslint-disable-line no-empty
```
ullasholla, line 18: I know this PR did not introduce this function, but it's probably a good time to look at it. Ignoring errors is usually a code-smell. In what cases would ignoring exceptions reading file paths be ok?

ullasholla (changes requested):
Looks ok, thanks for the PR. Had a few comments. Could you please have a look?

packages/common/package.json
IainSAP: We would need to update the build scripts also, so Ill leave that for @kranthie-sap

packages/common/test/utils.test.ts
IainSAP: Not sure, there are no test cases per se but just various combinations of parameters that should be tested at the very least.

devinea (approved):

devinea (approved):

packages/common/test/utils.test.ts
IainSAP: They are testing that the outputs of `getPackageJsonTasks` are as expected based on those inputs. Im not sure I understand what the ask is here?

packages/common/package.json
IainSAP: Lets leave it until we remove them all for consistency.

packages/common/test/utils.test.ts
```diff
@@ -0,0 +1,34 @@
...
+
+        tasks = getPackageJsonTasks(true, false, undefined, 'testApp-tile', undefined, 'testLocalStart.html');
+        expect(tasks).toMatchInlineSnapshot(`
+            Object {
+              "start": "echo \\\\\\"This application was generated with a local metadata file and does not reference a live server. Please add the required server configuration or start this application with mock data using the target: npm run start-mock\\\\\\"",
+              "start-local": "fiori run --config ./ui5-local.yaml --open 'testLocalStart.html#testApp-tile'",
+              "start-noflp": "echo \\\\\\"This application was generated with a local metadata file and does not reference a live server. Please add the required server configuration or start this application with mock data using the target: npm run start-mock\\\\\\"",
+            }
+        `);
+    });
```
ullasholla:
```suggestion
        describe('package.json task generation', () => {
        test('non empty sap-client is handled correctly', () => {
            let tasks = getPackageJsonTasks({
                localOnly: false,
                addMock: true,
                sapClient: '100',
                flpAppId: 'testApp-tile'
            });
            expect(tasks).toMatchInlineSnapshot(`
            Object {
              "start": "fiori run --open 'test/flpSandbox.html?sap-client=100#testApp-tile'",
              "start-local": "fiori run --config ./ui5-local.yaml --open 'test/flpSandbox.html?sap-client=100#testApp-tile'",
              "start-mock": "fiori run --config ./ui5-mock.yaml --open 'test/flpSandbox.html?sap-client=100#testApp-tile'",
              "start-noflp": "fiori run --open 'index.html?sap-client=100'",
            }
        `);
        });

        test('handles empty sap-client correctly', () => {
            let tasks = getPackageJsonTasks({ localOnly: false, addMock: true });
            expect(tasks).toMatchInlineSnapshot(`
            Object {
              "start": "fiori run --open 'test/flpSandbox.html'",
              "start-local": "fiori run --config ./ui5-local.yaml --open 'test/flpSandbox.html'",
              "start-mock": "fiori run --config ./ui5-mock.yaml --open 'test/flpSandbox.html'",
              "start-noflp": "fiori run --open 'index.html'",
            }
        `);
        });

        test('for local metadata file works correctly', () => {
            let tasks = getPackageJsonTasks({
                localOnly: true,
                addMock: false,
                sapClient: undefined,
                flpAppId: 'testApp-tile',
                localStartFile: 'testLocalStart.html'
            });
            expect(tasks).toMatchInlineSnapshot(`
            Object {
              "start": "echo \\\\\\"This application was generated with a local metadata file and does not reference a live server. Please add the required server configuration or start this application with mock data using the target: npm run start-mock\\\\\\"",
              "start-local": "fiori run --config ./ui5-local.yaml --open 'testLocalStart.html#testApp-tile'",
              "start-noflp": "echo \\\\\\"This application was generated with a local metadata file and does not reference a live server. Please add the required server configuration or start this application with mock data using the target: npm run start-mock\\\\\\"",
            }
        `);
        });
    });
```

Does this make more sense? 
The function `getPackageJsonTasks` accepts 6 parameters, 2 boolean and 4 optional strings. The contents of package.json is different based on the input. 
One test with 3 calls embedded with the description `'Test package json task generation'` could be changed to describe why we're choosing to test those 3 calls.

ullasholla: It doesn't need to help me, it keeps the code base clean. I wouldn't call having simple, focused tests 'boiler-plate'.

>I think grouping them with comments makes more sense

Did I miss the comments?

PR #100, opened 2021-10-11
chore: ensure pretty-quick respects .prettierignore file

ullasholla:
This PR adds `--ignore-path .prettierignore` to the `pretty-quick` config so that `.prettierignore` settings are respected.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,1 @@
-npx pretty-quick --staged
+npx pretty-quick --staged --ignore-path .prettierignore


.husky/pre-commit
```diff
@@ -1 +1 @@
-npx pretty-quick --staged
+npx pretty-quick --staged --ignore-path .prettierignore
```
kranthie-sap, line 1: @ullasholla: Does `pnpx` work here?

kranthie-sap (changes requested):

.husky/pre-commit
ullasholla: No, didn't work for me. It couldn't find `pnpx` or `pnpm` 🤷‍♂️ 
That's the first thing I tried. Does it work for you?

ullasholla: I must have had a typo before, just `pnpm` works. Thanks @kranthie-sap

kranthie-sap (approved):
Looks good to me. Thanks for the prepare script too.

PR #105, opened 2021-10-13
fix(odata-service): Add mock config to ui5-local.yaml

IainSAP:
https://github.com/SAP/open-ux-tools/issues/106

Adds mock server config to ui5-local.yaml.

### DIFF
--- a/packages/odata-service-writer/src/index.ts
+++ b/packages/odata-service-writer/src/index.ts
@@ -102,7 +102,7 @@ async function generate(basePath: string, data: OdataService, fs?: Editor): Prom
             Object.assign(data, { appid })
         );
         await addMiddlewareConfig(fs, basePath, 'ui5-mock.yaml', proxyMiddleware.config, proxyMiddleware.comments);
-        await addMiddlewareConfig(fs, basePath, 'ui5-mock.yaml', mwMock);
+        await addMiddlewareConfig(fs, basePath, 'ui5-mock.yaml', mockserverMiddleware);
         await addMiddlewareConfig(fs, basePath, 'ui5-mock.yaml', appReloadMiddleware);
         // create local copy of metadata and annotations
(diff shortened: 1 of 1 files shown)


packages/odata-service-writer/src/index.ts
```diff
@@ -89,6 +89,9 @@ async function generate(basePath: string, data: OdataService, fs?: Editor): Prom
         proxyLocalMiddleware.config,
         proxyLocalMiddleware.comments
     );
+
+    const mwMock = getMockServerMiddlewareConfig(data);
+    await addMiddlewareConfig(fs, basePath, 'ui5-local.yaml', mwMock);
```
ullasholla, line 94:
```suggestion
    const mockserverMiddleware = getMockServerMiddlewareConfig(data);
    await addMiddlewareConfig(fs, basePath, 'ui5-local.yaml', mockserverMiddleware);
```
Use a more descriptive name? I can't use the suggestion to capture all the usages of the variable, but you get the picture.

ullasholla: Also, don't we do this only for v4 services?

ullasholla (changes requested):
Thanks. 1 minor point about using a more descriptive variable name (`mwMock` -> `mockserverMiddleware`) and a clarification about generating this for v2 & v4 services. My understanding was that it was relevant for v4 services.

Cheers for updating the snapshots.

devinea (approved):

packages/odata-service-writer/src/index.ts
ullasholla: One more clarification: don't we do this conditionally in tools-suite? I see a check for a v4 service there.

ullasholla: Thanks @IainSAP

PR #108, opened 2021-10-14
fix(ui5-app): Write theme to scripts

IainSAP:
Write ui5 theme to index.html


unseen1980 (approved):

petermuessig:
Hi @ullasholla ,

the `themelib_sap_horizon` contains only the `sap_horizon` theme so far: 
https://github.com/SAP/openui5/tree/master/src/themelib_sap_horizon

The names for the theme variants of that theme family are not yet decided - most probably they will be called `sap_horizon_dark`, `sap_horizon_hcb` and `sap_horizon_hcw` - but this is not yet carved in stone...

If you want to query the available themes an option might be to lookup all `.theming` files in the resources of the dependencies. This file can be found at the following location `resources/sap/ui/core/themes/*/.theming`. This file is the theme metadata for the UI theme designer and provides the theme metadata such as the name as JSON content: 
https://github.com/SAP/openui5/blob/master/src/themelib_sap_horizon/src/sap/ui/core/themes/sap_horizon/.theming

Cc: @matz3 @RandomByte @andreas-roessler @Stanley-78 - maybe you have a better idea.

PR #112, opened 2021-10-15
feature(ui5-application-writer): Add support for 'sapuxLayer' option

IainSAP:
Adds `sapuxLayer` as an option

### DIFF
--- a/packages/fiori-freestyle-writer/templates/common/add/webapp/test/flpSandbox.html
+++ b/packages/fiori-freestyle-writer/templates/common/add/webapp/test/flpSandbox.html
@@ -54,10 +54,10 @@
 		};
 	</script>
 
-	<script src="/test-resources/sap/ushell/bootstrap/sandbox.js" id="sap-ushell-bootstrap"></script>
+	<script src="../test-resources/sap/ushell/bootstrap/sandbox.js" id="sap-ushell-bootstrap"></script>
 	<!-- Bootstrap the UI5 core library -->
 	<script id="sap-ui-bootstrap"
-		src="/resources/sap-ui-core.js"
+		src="../resources/sap-ui-core.js"
 		data-sap-ui-libs="<%- ui5.ui5Libs %>"
 		data-sap-ui-async="true"
 		data-sap-ui-preload="async"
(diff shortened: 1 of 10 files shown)


packages/ui5-application-writer/src/index.ts
```diff
@@ -28,7 +28,7 @@ async function generate(basePath: string, ui5App: Ui5App, fs?: Editor): Promise<
     // Add optional features
     if (ui5App.appOptions) {
         Object.entries(ui5App.appOptions).forEach(([key, value]) => {
-            if (value === true) {
+            if (value !== false) {
```
ullasholla, line 31:
```suggestion
            if (value) {
```
Will a check for 'truthy' values suffice here?

packages/ui5-application-writer/test/options.test.ts
```diff
@@ -11,9 +14,20 @@ describe('UI5 templates', () => {
         removeSync(outputDir); // even for in memory
     });
 
-    it('generates options', async () => {
-        const projectDir = join(outputDir, 'testapp_options');
-        const fs = await generate(projectDir, {
+    afterEach(() => {
```
ullasholla, line 17: Nice. This is much cleaner.

packages/ui5-application-writer/test/options.test.ts
```diff
@@ -11,9 +14,20 @@ describe('UI5 templates', () => {
...
+            // write out the files for debugging
+            if (debug) {
+                fs.commit(resolve);
+            } else {
+                resolve(true);
+            }
+        });
+    });
+
+    it('generates simple options', async () => {
```
ullasholla, line 28: Do we mean 'with default' options?

IainSAP: No there are no default options. There are simple types (boolean) which simply add some file entries. Then there are options that require template params.

packages/ui5-application-writer/src/index.ts
IainSAP: The condition is testing for cases where the value is of type boolean and is false explicitly.

ullasholla: Is this the explanatory comment?
```
// Add optional features
```

So do we treat an option set to `undefined` or `''` as true?

packages/ui5-application-writer/test/options.test.ts
IainSAP: You could rightly ask what about writing a value thats a boolean...well that would be: 
```
appOptions: {
    someNewProp: 'true'
}
...
<%- someNewProp %>
...
```

PR #123, opened 2021-10-18
refactor: rm common types module

ullasholla:
This PR removes the common types module `@sap-ux/open-ux-tools-types` as it makes more sense for each module to supply types relevant to its functionality.


packages/odata-service-writer/package.json
```diff
@@ -36,7 +36,6 @@
         "fs-extra": "10.0.0",
         "typescript": "4.2.4",
         "@sap-ux/ui5-application-writer": "workspace:*",
```
tobiasqueck, line 38: Why do we have this devDependency in the `odata-service-writer`?

tobiasqueck (changes requested):
Please have a look at my inline comment. I have focussed my review on the `package.json` files to make sure that the removal of the `types` module doesn't introduce a circular dependency or anything else that is odd.

packages/ui5-config/src/types.ts
```diff
@@ -35,5 +57,3 @@ export interface MiddlewareConfig {
         ignoreCertError?: boolean;
     };
 }
-// TypeDefs using TypeScript 4.1 feature will break consumer compilation < 4.1
-// export { NodeComment, Path } from '@sap-ux/yaml';
```
IainSAP, line 39: We still need to prevent export of types that use 4.1 features to avoid breaking the consumption by tools-suite.

Issue #134, opened 2021-10-21
BUG - Mock server middleware is written even if no metadata provided


IainSAP:
### **Bug Description**
Mock server middleware is written even if no metadata provided

This loc https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-writer/src/index.ts#L94
should be here: 
https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-writer/src/index.ts#L99

### **Expected Behavior**
No mockserver middleware written if no metadata provided

### **Steps to Reproduce**
2. ...
3. ...

### **Context**
- OS/Platform: {...}
- Browser: {...}
- Affected component: {...}

### **Log Output / Stack Trace / Screenshots**
```
{...}
```
### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: {...}
- Business impact: {...}

PR #135, opened 2021-10-21
fix(odata-service): Dont add mock config if no metadata

IainSAP:
Fix for https://github.com/SAP/open-ux-tools/issues/134


tobiasqueck (changes requested):
Generally the change makes sense, the mockserver requires the metadata, otherwise, it doesn't work, however, I am wondering in what use case do we add an odata service without knowing/having the metadata?

unseen1980 (approved):

Issue #130, opened 2021-10-20
Pass sap-client in to getFioriToolsProxyMiddlewareConfig so it is written to the ui5.yaml files


devinea:
### **Bug Description**
Pass sap-client in to getFioriToolsProxyMiddlewareConfig so it is written to the ui5.yaml files
Noted in https://github.com/SAP/open-ux-tools/pull/127 
### **Expected Behavior**
A clear and concise description of what you expect to happen.

### **Steps to Reproduce**
2. ...
3. ...

### **Context**
- OS/Platform: {...}
- Browser: {...}
- Affected component: {...}

### **Log Output / Stack Trace / Screenshots**
```
{...}
```
### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: {...}
- Business impact: {...}

PR #138, opened 2021-10-21
Fix missing client in fiori proxy config

devinea:
#130


tobiasqueck (approved):
Compared the snapshot with my expectation, especially check for the leading 0.

kranthie-sap:
@devinea: Yes, please run `pnpm cset`, select the updated packages, patch bump and add a message for this change. You will then need to push the newly added changeset config files.

PR #141, opened 2021-10-21
chore: update version pipeline to skip git hooks

kranthie-sap:
* Update `version` pipeline to skip the git commit hook.


kranthie-sap:
Merging using admin privileges to test the version pipeline.

Issue #78, opened 2021-09-29
FEATURE - Enhance Fiori freestyle API for integration into `easy-ui5`


tobiasqueck:
_As a **developer**,
I want to **generate Fiori freestyle project with more flexibility**,
so that I **can choose my own view names or select the source of UI5**_

### Description
When @vobu tried integrating the `fiori-freestyle` template into the `easy-ui5` following features were not possible to be set:

1. Configure a custom view/controller name instead of the currently hardcoded name
2. Configure the target url that is used to load UI5 in the preview
3. Generate a project using OpenUI5 instead of SAPUI5.

### Technical Design
Coming up with a technical design that works with both `@sap/generator-fiori` as well as with `easy-ui5` needs to be evaluated.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I am using `easy-ui5`
**when** I execute the newly integrated `fiori-freestyle` templates
**then** I will be able to define my custom view name.

**Given** I am using `easy-ui5`
**when** I execute the newly integrated `fiori-freestyle` templates
**then** I will be able to use OpenUI5 loaded from a custom location.

**Given** I am using `@sap/generator-fiori`
**when** I generate a Fiori freestyle project
**then** I want it to behave as today without requiring me to enter more information.

### Notes
_none_

### Tasks
- [x] Discuss an extension of the API with experts from both `@sap/generator-fiori` as well as with `easy-ui5`
- [x] Implement the agreed API extension 
- [x] Experts from both `@sap/generator-fiori` as well as with `easy-ui5` verify the implementation
- [x] Unit Tests added
- [x] Test Cases defined

tobiasqueck:
@vobu could you please check if I collected all requested changes from #65?

vobu:
> @vobu could you please check if I collected all requested changes from #65?

in principle all good. 
would still word things differently for "Description" > "2."
"Configure a custom bootstrap URL for consuming UI5 runtime resources"
this would also make "3." obsolete IMO.

Acceptance criteria are spot on.

tobiasqueck:
The features described are implemented and the acceptance criteria is fulfilled. 
@vobu did we miss something?

vobu:
just FTR: /me thinks, we're good 👍

PR #165, opened 2021-11-02
refactor to be more backward compatible with mem-fs-editor (and yeoman-environment ^2.10.0)

devinea:
https://github.com/SAP/open-ux-tools/pull/158
https://github.com/SAP/open-ux-tools/issues/78

refactor to be more backward compatible with mem-fs-editor (and yeoman-environment ^2.10.0)
`processDestinationPath` is not available in older versions


ullasholla:
@devinea, thanks looks good.

I added a test for the basic template, didn't realize we didn't have one.

Also moved the `viewName` variable inside the `if` trusting that `setDefault()` will do its job: https://github.com/SAP/open-ux-tools/pull/165/commits/e63a85a3fec137660718f6eec58ab66f5e87a212
The code seemed a bit clearer to me this way. Let me know if you're unhappy with this and I can revert it.

I wonder if there are more incompatibilities between the two mem-fs-editor versions. Does it make sense to use the same version as yeoman-environment@2.10.0+? @tobiasqueck  & @IainSAP, thoughts about this?

devinea:
@ullasholla  We do have some tests for basic here https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-freestyle-writer/test/basic.test.ts#L8 

I am ok with the change to the if statement. 👍 
I didn’t want to rush into a downgrade of `mem-fs-editor` as we have started to investigate an upgrade to `yeoman-environment@3+` in tool-suite. But being on different major versions has its risks.

IainSAP:
@devinea Can we doc in the generate API the version expectation of `mem-fs-editor` so at the consumers are aware and we are covered?

packages/fiori-freestyle-writer/src/index.ts
```diff
@@ -34,15 +34,30 @@ async function generate<T>(basePath: string, data: FreestyleApp<T>, fs?: Editor)
...
-                return path.replace('View1', (ffApp.template.settings as unknown as BasicAppSettings).viewName!);
-            } else {
-                return path;
-            }
-        }
-    });
+    fs.copyTpl(join(tmplPath, ffApp.template.type, 'add', `**/*.*`), basePath, ffApp, undefined, {});
+
+    if (ffApp.template.type === TemplateType.Basic) {
+        // Todo: need to find a way around the casts
```
tobiasqueck, line 40: I don't think you need `as unknown` here.
`(ffApp.template.settings as BasicAppSettings).viewName` should be sufficient.

packages/fiori-freestyle-writer/src/index.ts
```diff
@@ -34,15 +34,30 @@ async function generate<T>(basePath: string, data: FreestyleApp<T>, fs?: Editor)
...
+            ffApp,
+            undefined,
+            {}
+        );
+        const controllerTarget = join(basePath, 'webapp', 'controller', `${viewName}.controller.js`);
+        fs.copyTpl(
+            join(tmplPath, ffApp.template.type, 'addViewFiles', 'controller', controllerTmplName),
+            controllerTarget,
+            ffApp,
+            undefined,
```
tobiasqueck, line 57: We don't need `undefined` and `{}` here, correct?

tobiasqueck (approved):
Tests still run, code looks good. 
I have not validated that it works better with `yeoman-environment` 2.10 but I assume that has been tested by @IainSAP

unseen1980 (approved):

IainSAP (approved):
Tested consumption, files output as expected.

PR #167, opened 2021-11-02
Port tool-suite change - add delay to appreload middleware

devinea:
Adding `delay` property as it is now recommended configuration


docirl (approved):

tobiasqueck (approved):
Type properly enhanced, value for new property set, snapshots updated.
Everything looks good.

PR #179, opened 2021-11-09
fix: use consistent whitespace in templates

ullasholla:
Some templates in `fiori-freestyle-writer` and `ui5-application-writer` use a mix of tabs and spaces for indentation. This is generally not a good idea. This PR normalizes the whitespace and formats the templates using 4 spaces.

The templates have been updated using the shell command:
```shell
find <templates dirname>  ! -type d -exec bash -c 'expand -t 4 "$0" > /tmp/e && mv /tmp/e "$0"' {} \; 
```
Source: https://stackoverflow.com/questions/11094383/how-can-i-convert-tabs-to-spaces-in-every-file-of-a-directory

The associated test snapshots have been updated as well.

Tagging: @tobiasqueck, @devinea & @IainSAP


tobiasqueck (approved):
Looks good. I only see whitespace changes and they are finally consistent. Thank you.

devinea (approved):
Thanks @ullasholla lgtm

PR #182, opened 2021-11-12
feat: enhanced ui5-config to also add abap-deploy middleware

tobiasqueck:
Proposal to enhance the `ui5-config` module to allow adding the `abap-deploy-task` required for `ui5-deploy.yaml` in Fiori projects.

@ullasholla @IainSAP @devinea fyi

### DIFF
--- a/packages/ui5-config/test/index.test.ts
+++ b/packages/ui5-config/test/index.test.ts
@@ -119,6 +119,19 @@ describe('UI5Config', () => {
         expect(ui5Config.toString()).toMatchSnapshot();
     });
 
+    test('addCustomTask', () => {
+        ui5Config.addCustomTasks([
+            {
+                name: 'ui5-task-zipper',
+                afterTask: 'generateCachebusterInfo',
+                configuration: {
+                    archiveName: 'my-archive'
+                }
+            }
+        ]);
+        expect(ui5Config.toString()).toMatchSnapshot();
+    });
+
     describe('addAbapDeployTask', () => {
         const app: AbapApp = {
             name: '~name',
--- a/packages/ui5-config/test/index.test.ts
+++ b/packages/ui5-config/test/index.test.ts
@@ -1,4 +1,3 @@
-import { getFioriToolsProxyMiddlewareConfig } from '../src/middlewares';
 import { UI5Config, AbapApp } from '../src';
 
 describe('UI5Config', () => {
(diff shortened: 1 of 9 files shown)


tobiasqueck:
The current flow in the Fiori deployment generator copies the existing `ui5.yaml` then deletes the `server` and adds the builder. To do a deletion, we also need an additional `deleteNode` method that I have added as stub and also a test case showing what the expectation is BUT it is not yet implemented.

longieirl (dismissed):

packages/ui5-config/test/index.test.ts
```diff
@@ -117,4 +117,49 @@ describe('UI5Config', () => {
...
+                },
+                app
+            );
+            expect(ui5Config.toString()).toMatchSnapshot();
+        });
+    });
+
+    test('removeMiddleware', () => {
+        const { config, comments } = getFioriToolsProxyMiddlewareConfig([], {});
+        ui5Config.addFioriToolsAppReloadMiddleware();
```
ullasholla, line 161: Could we use an already existing middleware that we delete instead? This adds a dependency to `addFioriToolsAppReloadMiddleware`. The problem with that is that bugs in that method could make this testcase fail. Or worse, bugs there could make this testcase pass when there's a bug in `removeMiddleware`.

packages/yaml/src/yaml-document.ts
```diff
@@ -228,6 +228,36 @@ export class YamlDocument {
...
+     * @param {string} path.matcher.value - the value
+     * @returns {YamlDocument} the YamlDocument instance
+     * @memberof YamlDocument
+     */
+    deleteAt({ path, matcher }: { path: string; matcher: { key: string; value: string } }): YamlDocument {
+        const pathArray = this.toPathArray(path);
+        const seq = this.document.getIn(pathArray) as YAMLSeq<yaml.Node>;
+        if (!seq) {
+            throw new Error(t('error.seqDoesNotExist', { path }));
+        }
```
ullasholla, line 247: We would need to the case when the node exists but is not a `YAMLSeq`. Otherwise the code after will fail.

packages/yaml/test/yaml-document.test.ts
```diff
@@ -251,6 +251,26 @@ seq1:
         });
     });
 
+    describe('deleteAt', () => {
+        it('delete node in existing empty sequence', async () => {
```
ullasholla, line 255: Could you please add tests to cover the cases where the node does not exist and when the node exists but is not a sequence?

ullasholla (changes requested):
The proposed API for `ui5-config` looks good. A couple of small changes to tests and the `deleteAt()` method may be needed.

packages/ui5-config/test/index.test.ts
tobiasqueck: 4bf4bcd

packages/yaml/src/yaml-document.ts
tobiasqueck: 8c7e5b3

packages/yaml/test/yaml-document.test.ts
tobiasqueck: 8c7e5b3

longieirl (approved):

ullasholla (approved):
Thanks @tobiasqueck. Code looks good. All the previous review comments have been addressed.

ullasholla:
> I believe there is a `deleteIn` API for removing items from yaml collections. https://eemeli.org/yaml/#collections e.g. `yamlDoc.deleteIn(['server', 'customMiddleware', index]);`

@IainSAP, not sure I understand. Yes, there's a `deleteAt` method in the library linked. It expects a full path. The one in this PR takes matcher and finds the element and then deletes it. What are you suggesting again?

IainSAP:
> > I believe there is a `deleteIn` API for removing items from yaml collections. https://eemeli.org/yaml/#collections e.g. `yamlDoc.deleteIn(['server', 'customMiddleware', index]);`
> 
> @IainSAP, not sure I understand. Yes, there's a `deleteIn` method in the library linked. It expects a full path. The one in this PR takes a matcher and finds the element and then deletes it. What are you suggesting again?

Maybe I've missed something. The proposed API also takes a path. But its probably possible to achieve the same using the existing apis rather than splicing
e.g. Something like (pseudo code)
```
    const index = document
        .getIn(...path).items.findIndex((item) => item.get(key) === value);
    ui5LocalYaml.deleteIn([...path, index]);
```
Im ok with either.

IainSAP (approved):

PR #187, opened 2021-11-17
fix(infra): override module versions

donal-tobin-sap:
upgrade a couple of modules that generate warnings for `pnpm audit`


kranthie-sap (dismissed):

tobiasqueck (approved):

devinea (approved):

tobiasqueck (changes requested):
@donal-tobin-sap changed my mind about the approval :) . Could you please add a document where you list the libs and the reason why we put in the version limitations? Right now, with two it is not a big deal, but I imagine the list is growing so properly tracking them will be helpful in the future.

tobiasqueck:
I have added a document at https://github.com/SAP/open-ux-tools/blob/fix/upgrade-modules/docs/version-overrides.md that I would be ok with as documentation. @donal-tobin-sap please have a look and change it if a different way of documentation would be preferred.

Otherwise, I have changed also a few versions of modules in the root `package.json` to reduce the amount of required overrides. In general, we should first look at why a vulnerable version is used and if possible upgrade the modules causing it.

tobiasqueck:
I needed to add a changeset since one of the dependencies that need updating is not just a `devDependency`

tobiasqueck:
Also added `pnpm audit` as `pre-commit` command so that changed dependencies are not accidentally bring in new vulnerabilities.

devinea (approved):
Thanks. Approach looks good. 
docs/version-overrides.md is clear, but a bit more detailed on than tool-suite doc. Would that format suffice?

tobiasqueck:
> docs/version-overrides.md is clear, but a bit more detailed on than tool-suite doc. Would that format suffice?

This is the output you get when running `pnpm audit` so I went for the 'lazy` option allowing to copy paste :)

devinea:
Fair enough. Thanks @tobiasqueck

tobiasqueck (dismissed):

donal-tobin-sap (commented):
👍🏽 Looks good to me.

tobiasqueck (approved):

devinea (approved):

PR #188, opened 2021-11-17
chore: add support to generate sonar coverage reports

kranthie-sap:
* Add `jest-sonar` dependency.
* Update `jest` config in all modules to use new reporter to generate jest coverage reports that Sonar Cloud can use to publish code coverage metrics.
* Add sonar cloud configuration file


FredAtWork (dismissed):

tobiasqueck (approved):

tobiasqueck (dismissed):

packages/fe-fpm-writer/test/unit/section.test.ts
```diff
@@ -136,7 +136,7 @@ describe('CustomSection', () => {
             expect(fs.read(fragmentPath)).toMatchSnapshot();
         });
 
-        const testVersions = [1.9, 1.85, 1.84,  1.86, 1.90];
+        const testVersions = [1.9, 1.85, 1.84, 1.86, 1.9];
```
devinea, line 139: Why do we have `1.9` twice?

packages/fiori-freestyle-writer/jest.config.js
```diff
@@ -3,7 +3,16 @@ module.exports = {
         '^.+\\.ts$': 'ts-jest'
     },
     collectCoverage: true,
-    coverageReporters: ['text','html'],
     collectCoverageFrom: ['src/**/*.ts'],
+    reporters: [
+        'default',
```
devinea, line 8: I assume `default`  covers  `coverageReporters: ['text','html'],` ?

sonar-project.properties
```diff
@@ -0,0 +1,21 @@
+sonar.organization=sap-1
+sonar.projectKey=SAP_open-ux-tools
+sonar.sources=.
+sonar.inclusions=**/*.ts
+sonar.exclusions=**/*.test.ts
```
devinea, line 5:
```suggestion
sonar.exclusions=**/*.test.ts, **/templates/**
```
I think the templates should be excluded. Is this the place to configure it?

packages/fe-fpm-writer/test/unit/section.test.ts
tobiasqueck: Why is this change in this PR? This is a regression currently in #269 which @stefanschreck is going to fix before merging. It should not be here ...

kranthie-sap: @tobiasqueck: It looks like it was updated by prettier in the git hook. I'll undo this change.

sonar-project.properties
kranthie-sap: @devinea: Yes, I added this in the latest commit.

packages/fiori-freestyle-writer/jest.config.js
kranthie-sap: @devinea: I couldn't find docs on the default reporters of jest. But, I did verify that the coverage is shown in the console and the html files are in the `coverage/lcov-report` folder.

devinea (dismissed):
Thanks @kranthie-sap

tobiasqueck (dismissed):

tobiasqueck (approved):

stefanschreck (approved):

Issue #73, opened 2021-09-28
FEATURE - CLI to enhance Fiori elements for OData v4 projects with the flexible programming model


tobiasqueck:
_As a **developer**,
I want to **be able to efficiently add flexible programming model extensions** to my Fiori elements for OData v4 application,
so that I can **focus on the actually functionality** instead of integration syntax._

### Description
A developer using the flexible programming model is likely to be a technical person that is familiar with (or even fond of) the CLI to generate project artifacts. With this feature we want to empower those developers to efficiently add FPM extensions like custom pages or custom actions to their projects. 
In addition, the more UI based tools (Fiori tools page modeler and guide development) can also make use of the templates and prompts used by the CLI, so that FPM extensions generated with the CLI or other tools are consistent.

### Technical Design
* separate the functionality into three modules
  * templates: containing all templating logic and exposing functions for each available FPM extension e.g. page, section, action etc.
  * prompts: generic definition of prompts that can be used by the CLI but also by UI based tools
  * cli: create a module that can be executed with `npm init` without the requirement to install anything like `yo` etc. (https://docs.npmjs.com/cli/v7/commands/npm-init) 
* templates and prompts have are to be isolated
* cli will depend on templates and prompts and transform output from prompts to the required input for the template
* other tools should use templates and prompts as needed but should not require the cli module

### Acceptance Criteria
**Given** I am in the root folder of an existing Fiori elements v4 project
**when** I execute `npm init @sap-ux/fiori-fpm`
**then** I will be able to use a simple CLI based generator adding FPM boilerplate code to my project.

**Given** I am NOT in the root folder of an existing Fiori elements v4 project
**when** I execute `npm init @sap-ux/fiori-fpm`
**then** I will be told what I did wrong.

**Given** I am the developer of tools for Fiori elements
**when** I want to reuse either the templates or the prompts
**then** I will be able to do that without any other unnecessary modules.

### Tasks
- [ ] validate with Fiori tools page modeler and guided development experts that the exposed modules are reusable for them
- [ ] create template module
- [ ] create prompts module
- [ ] create cli module using templates and prompts
- [ ] Unit Tests added
- [ ] Test Cases defined

tobiasqueck:
Current state:

- custom page, action and column have been released as `@sap-ux/fe-fpm-writer`
- Fiori tools team is working on replacing the existing custom page implementation with the `@sap-ux/fe-fpm-writer`
[...]

PR #212, opened 2021-12-09
feat(fev4-fpm): extract custom section from Fiori tools in to this module

815are:
# Issue

3rd point from #73

Changes:
1. Support for Custom Section creation - 3rd point from #73;
2. Deleted - https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/templates/column/ext/CustomColumnFragment.xml and created new template in `common` - https://github.com/SAP/open-ux-tools/blob/feat/73/customSection/packages/fe-fpm-writer/templates/common/Fragment.xml
In result template used by `column` and `section`. Also removed `xmlns:l="sap.ui.layout"` from fragment, because did not find reference where `<l:` would be used.
3. Moved part form `column/index.ts` - https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/column/index.ts#L52-L59 into `common` and reusing for both `column` and `section`
https://github.com/SAP/open-ux-tools/blob/feat/73/customSection/packages/fe-fpm-writer/src/common/defaults.ts#L40
4. Correction regarding integration test - https://github.com/SAP/open-ux-tools/blob/feat/73/customSection/packages/fe-fpm-writer/test/integration/index.test.ts#L33
I noticed that when I added new step, then after snapshot update - previous step's result was not included into snapshot. It happened because code part `fs.copy(join(testInput, 'basic-lrop'), targetPath);` was called on each step and in result it lead that only last step was checked against snapshot.


packages/fe-fpm-writer/src/section/types.ts
```diff
@@ -0,0 +1,46 @@
...
+export interface CustomSection extends CustomElement {
+    /**
+     * Name of the routing target
+     */
+    target: string;
+
+    /**
+     * Extension file type. Currently only xml fragment is supported.
+     */
+    type: CustomSectionType;
```
tobiasqueck, line 20: Do we plan to support to anything else here in the future (in the Fiori tools)? If not, I would consider removing it or at least making it optional and then using `XMLFragment` if the nothing is provided by the caller.

tobiasqueck (changes requested):
- code is well structured and easy to read
- test coverage is 100%
- tests are well structured allowing to easily understand the exposed API
- integration test properly extended
- thanks for making the fragment coding reusable and extracting it to common
- just one minor comment

stefanschreck (dismissed):

packages/fe-fpm-writer/src/section/types.ts
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Optional control XML that will be generated into the fragment of section. If the property isn't provided then a sample control will be generated.
+     */
+    control?: string;
+}
+
+export interface InternalCustomSection extends CustomSection, InternalCustomElement {
+    content: string;
+    // Target section file format - currently only xml fragment is supported.
+    type: typeof CUSTOM_SECTION_XML_FRAGMENT;
```
tobiasqueck, line 36: @815are we don't need this one anymore, correct? (same for line 3)

packages/fe-fpm-writer/templates/section/manifest.section.json
```diff
@@ -0,0 +1,33 @@
...
+                                            "name": "<%- ns %>.<%- name %>",
+                                            "position": {
+                                                <% if (position.placement) { %> 
+                                                    "placement": "<%- position.placement %>", 
+                                                <% } %>
+                                                <% if (position.anchor) { %> 
+                                                    "anchor": "<%- position.anchor %>" 
+                                                <% } %>
+                                            },
+                                            "type": "<%- type %>",
```
tobiasqueck, line 21: This can also be removed since Fiori elements is not evaluating the property, correct?

packages/fe-fpm-writer/src/section/types.ts
815are: @tobiasqueck 
According to latest information from Fiori team(from email conversation). According to Fiori team statement - there was plan at beginning to use `type` and add additional type like `XMLView`, but according to latest statement team decided to stick with fragments. Also support for `JSFragments` not planned as well.
Proposal from Fiori team was that we ignore such property.

I deleted `type` property in commit 1a331770d57845ac7aa54cd89b172837f242a64d

packages/fe-fpm-writer/src/section/types.ts
815are: yes, deleted in commit 1a33177

packages/fe-fpm-writer/templates/section/manifest.section.json
815are: yes, deleted in commit 1a33177

tobiasqueck (approved):
Thank you for following up with the Fiori elements team and simplifying the API accordingly.

stefanschreck (approved):

tobiasqueck (approved):

stefanschreck (approved):

PR #238, opened 2021-12-27
fix(fev4-fpm): Custom Section. Two 'manifest.json' generations versions for different UI5 versions below and above 1.86

815are:
Part of 3rd point from #73

Main PR was - https://github.com/SAP/open-ux-tools/pull/212

According to latest documentation `name` property in `manifest.json` is deprecated since 1.86 and new approach is:
```
"NewCustomSection": {
        "position": {
          "anchor": "DummyFacet",
          "placement": "After",
        },
        "template": "sapux.fe.fpm.writer.test.extensions.custom.NewCustomSection",
        "title": "New Custom Section",
      },
```

Old approach is(<1.86):
```
"NewCustomSection": {
        "name": "sapux.fe.fpm.writer.test.extensions.custom.NewCustomSection",
        "position": {
          "anchor": "DummyFacet",
          "placement": "After",
        },
        "title": "New Custom Section",
        "type": "XMLFragment",
      },
```

In scope of PR - we support both variants by specifying target UI5 version.

API was not changed and still accepts property `name` - similar like it is for `columns`. In result property `name` is transferred to `template` or `name` depending on passed target UI5 version

### DIFF
--- a/.changeset/fresh-ears-doubt.md
+++ b/.changeset/fresh-ears-doubt.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fe-fpm-writer': patch
+---
+
+Custom Section. Fix to support old(<1.86) and new(>=1.86) manifest.json syntaxes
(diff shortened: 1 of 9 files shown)


stefanschreck (approved):

tobiasqueck (approved):
* build and test still work
* code looks good
* using a different subfolder for the different version keeps it nicely separated

PR #258, opened 2022-01-11
feat(all): switch ui5 CDNs

longieirl:
Switch CDN for ui5 libs from ui5.sap.com -> sapui5.hana.ondemand.com.


devinea (approved):
lgtm. patch release would probably suffice.

tobiasqueck (changes requested):
Let's wait with this one for a moment and clarify internally whether the preview urls should actually change.

Issue #263, opened 2022-01-14
FEATURE - Support use of existing fragments and code for extension creation


cfg74:
_As a **developer**,
I want to **use existing fragments** and code for extension creation,
so that I **can make use of previous efforts**_

### Description

_Creation of extensions shall allow the reuse of existing fragments and code to speed up development._

Reasons for reuse:
1. Copy code from another project
2. Use existing code already in project in case of delete/re-create, undo/redo, today only start from scratch is possible as existing code will be cleared and overwritten
3. Support cases where multiple extension need to share a single fragment or code (e.g. V2 custom columns)

### Technical Design
_What modules need to be created or enhanced?_ 
_How would the functionality be exposed and how can it be consumed?_

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a ready code snippet / fragment ...
**when** I create a new extension
**then** I can use these existing artifacts or let a fresh one being created

### Notes
@stefanschreck @tobiasqueck 

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

stefanschreck:
_What modules need to be created or enhanced?_
This is relevant for all extension types that include a fragment, i.e. custom page, custom section and custom column. But the same requirement is also valid for controller coding, thus the solution should be applied to custom actions as well. It must be a general pattern for all extension types.

_How would the functionality be exposed and how can it be consumed?_
What we discussed are the two options: either have an additional input parameter for controlling the overwrite, or not to introduce such a parameter and generally check the file existence.
Design proposal is to take the second option: for each view or controller, check the existence first, and do not overwrite if the file is there.

stefanschreck:
Fine, thanks. One additional remark, or special case: if the user passes the `control` content  in the interface and specifies an existing control by `name`, the expectation probably is that the content get overwritten. I will program it like this.

stefanschreck:
@cfg74 Issue can be closed now

tobiasqueck:
@cfg74 does https://github.com/SAP/open-ux-tools/pull/269 sufficiently cover your request?

cfg74:
Lgtm, thanks!

PR #269, opened 2022-01-19
FPM writer: do not overwrite files

stefanschreck:
Solves feature request https://github.com/SAP/open-ux-tools/issues/263.

All extension writers are enhanced so that they do not overwrite existing fragments, views, or controllers. 
Instead, the given file is taken but the program continues and refers the existing file in the manifest.
New tests have been added for this scenario.

For further information please check the feature request.


tobiasqueck (changes requested):
Looks good. I have just two minor requests for improvements.
I didn't run the unit tests locally but the snapshot updates look correct and the CI test runs succeed, so I assume they are good.

packages/fe-fpm-writer/src/section/index.ts
```diff
@@ -70,12 +70,11 @@ export function generateCustomSection(basePath: string, customSection: CustomSec
...
     // add event handler if requested
-    if (completeSection.eventHandler) {
-        fs.copyTpl(
-            join(root, 'common/EventHandler.js'),
-            join(completeSection.path, `${completeSection.name}.js`),
-            completeSection
-        );
+    const controllerPath = join(completeSection.path, `${completeSection.name}.js`);
+    if (!fs.exists(controllerPath)) {
+        if (completeSection.eventHandler) {
```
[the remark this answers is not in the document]
stefanschreck: right, I will change this.

packages/fe-fpm-writer/test/unit/column.test.ts
```diff
@@ -80,7 +82,27 @@ describe('CustomAction', () => {
             expect(fs.read(expectedFragmentPath)).toMatchSnapshot();
             expect(fs.read(expectedFragmentPath.replace('.fragment.xml', '.js'))).toMatchSnapshot();
         });
-        test('generateCustomColumn 1.85, no handler, all properties', () => {
+        test('version 1.86, with existing handler, all properties', () => {
+            const controllerPath = 'my.test.App.ext.ExistingHandler.onCustomAction';
+            fs.write(controllerPath, 'dummyContent');
+            const testCustomColumn: CustomTableColumn = {
+                ...customColumn,
+                eventHandler: controllerPath,
```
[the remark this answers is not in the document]
stefanschreck: @tobiasqueck done, please check again

packages/fe-fpm-writer/test/unit/section.test.ts
```diff
@@ -136,7 +152,7 @@ describe('CustomSection', () => {
             expect(fs.read(fragmentPath)).toMatchSnapshot();
         });
 
-        const testVersions = [1.9, 1.85, 1.84,  1.86, 1.90];
+        const testVersions = [1.9, 1.85, 1.84, 1.86, 1.9];
```
tobiasqueck, line 155: that looks wrong. Why `1.9` and not `1.90`, btw, now you have `1.9` twice in the array

packages/fe-fpm-writer/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-    "include": ["src"],
+    "include": ["src/**/*.ts"],
```
tobiasqueck, line 2: Why this change?
We have `src` in every other project e.g. https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-writer/tsconfig.json#L2

packages/fe-fpm-writer/test/unit/section.test.ts
stefanschreck: was a mistake. I'll undo.

packages/fe-fpm-writer/tsconfig.json
stefanschreck: I tried to get rid of my local eslint problems, and had tried several things. When reading the [documentation](https://www.typescriptlang.org/tsconfig#include), I thought it better to restrict to the ts files here. But as .ts, .tsx, and .d.ts are the default we can also neglect the suffix. Indeed, it is not the only tsconfig file at open-ux-tools which uses the suffix. Finally, it might better to work without suffix as I can also find a glob definition under https://github.com/SAP/open-ux-tools/blob/main/tsconfig.json.

stefanschreck: @tobiasqueck all done

packages/fe-fpm-writer/test/unit/section.test.ts
tobiasqueck: @stefanschreck  You re-requested a review but I still see the issue. Did you forget to push something?

stefanschreck: weird. the commit was pushed but without this correction. need a new commit.

stefanschreck: @tobiasqueck the change gets condensed during the commit, as 1.9 and 1.90 are the same **numbers**. I can only choose a different number, e.g. 1.98.

tobiasqueck (approved):

stefanschreck:
Thanks, Tobias.
@815are @Klaus-Keller additional reviewer needed...

815are (approved):

PR #283, opened 2022-01-25
Update pnpm version used and remove double recursive

devinea:
No change set needed as this does no impact released modules.
yml file is easier to read with diff https://github.com/SAP/open-ux-tools/pull/283/files?diff=split&w=1


tobiasqueck (approved):
Thanks, this solves the random build issues I saw in my PR.

unseen1980 (approved):

Issue #288, opened 2022-01-25
BUG - Double `?` in `start` task


IainSAP:
### Related Feature
Feature request: (https://github.com/SAP/open-ux-tools/issues/271)

### Description
This PR https://github.com/SAP/open-ux-tools/commit/030f490b6bbf0ed89a1a8ec6625af4879ff2ab78 has introduced an error in the package json start script when a client value is provided. It now adds 2 query param delimiters '?':

` "start": "fiori run --open 'test/flpSandbox.html??sap-client=012&sap-ui-xx-viewCache=false#felrop1-tile'",`

### Steps to Reproduce
Generate a Fiori Freestyle app providing a client value via the service interface.

### Expected results
The correct single '?' is used.

### Actual results
` "start": "fiori run --open 'test/flpSandbox.html??sap-client=012&sap-ui-xx-viewCache=false#felrop1-tile'",`

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

IainSAP:
fyi: @slin-sap

IainSAP:
Thx for the fix @slin-sap

PR #289, opened 2022-01-25
fix(fiori-freestyle-writer): Fixes multiple query delims (`??`)

IainSAP:
Fix for : https://github.com/SAP/open-ux-tools/issues/288

### DIFF
--- a/packages/fiori-freestyle-writer/src/packageConfig.ts
+++ b/packages/fiori-freestyle-writer/src/packageConfig.ts
@@ -28,21 +28,27 @@ export function getPackageJsonTasks({
     startFile?: string;
     localStartFile?: string;
 }): { start: string; 'start-local': string; 'start-noflp': string; 'start-mock'?: string } {
-    const sapClientParam = sapClient ? `?sap-client=${sapClient}` : '';
-    let urlParam = [sapClientParam, 'sap-ui-xx-viewCache=false'].filter((param) => !!param).join('&');
-    urlParam = urlParam ? `?${urlParam}` : '';
-    const params = `${urlParam}${flpAppId ? `#${flpAppId}` : ''}`;
+
+    const addParams = [['sap-ui-xx-viewCache', 'false']];
+    if (sapClient) {
+        addParams.push([`sap-client`, `${sapClient}`]);
+    }
+    const searchParams = new URLSearchParams(addParams);
+    const flpParams = `?${searchParams}${flpAppId ? `#${flpAppId}` : ''}`;
+
     const startCommand = localOnly
         ? `echo \\"${t('info.mockOnlyWarning')}\\"`
-        : `fiori run --open '${startFile || 'test/flpSandbox.html'}${params}'`;
+        : `fiori run --open '${startFile || 'test/flpSandbox.html'}${flpParams}'`;
+
     const startLocalCommand = `fiori run --config ./ui5-local.yaml --open '${
         localStartFile || 'test/flpSandbox.html'
-    }${params}'`;
+    }${flpParams}'`;
+
     const startNoFlpCommand = localOnly
         ? `echo \\"${t('info.mockOnlyWarning')}\\"`
-        : `fiori run --open '${'index.html'}${urlParam}'`;
+        : `fiori run --open '${'index.html?'}${searchParams}'`;
 
-    const mockTask = `fiori run --config ./ui5-mock.yaml --open 'test/flpSandbox.html${params}'`;
+    const mockTask = `fiori run --config ./ui5-mock.yaml --open 'test/flpSandbox.html${flpParams}'`;
     return Object.assign(
         {
             start: startCommand,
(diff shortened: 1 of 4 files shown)


devinea (approved):
Thanks @IainSAP lg

packages/fiori-freestyle-writer/src/packageConfig.ts
```diff
@@ -28,21 +28,27 @@ export function getPackageJsonTasks({
     startFile?: string;
     localStartFile?: string;
 }): { start: string; 'start-local': string; 'start-noflp': string; 'start-mock'?: string } {
-    const sapClientParam = sapClient ? `?sap-client=${sapClient}` : '';
-    let urlParam = [sapClientParam, 'sap-ui-xx-viewCache=false'].filter((param) => !!param).join('&');
-    urlParam = urlParam ? `?${urlParam}` : '';
-    const params = `${urlParam}${flpAppId ? `#${flpAppId}` : ''}`;
+
+    const addParams = [['sap-ui-xx-viewCache', 'false']];
```
ullasholla, line 32: Does `addParams` stand for additional parameters?

ullasholla:
Thanks for the fix Iain. It's very hard to spot bugs in that method, the cognitive complexity of those 20 lines is pretty high I must say. Your fix certainly cleans it up a bit.

packages/fiori-freestyle-writer/src/packageConfig.ts
IainSAP: Sure...or params to be added :)

packages/fiori-freestyle-writer/src/packageConfig.ts
```diff
@@ -28,21 +28,27 @@ export function getPackageJsonTasks({
...
+    if (sapClient) {
+        addParams.push([`sap-client`, `${sapClient}`]);
+    }
+    const searchParams = new URLSearchParams(addParams);
+    const flpParams = `?${searchParams}${flpAppId ? `#${flpAppId}` : ''}`;
```
slin-sap, line 37: Thanks Iain, I forgot to remove '?' in ?sap-client=${sapClient}

If this disable cache param is removed in the future for some reason, searchParams could be empty.  It is better to check searchParams is empty and decide if we add ? instead of always having it in the string template.

slin-sap (changes requested):

packages/fiori-freestyle-writer/src/packageConfig.ts
```diff
@@ -28,21 +28,27 @@ export function getPackageJsonTasks({
...
+
     const startLocalCommand = `fiori run --config ./ui5-local.yaml --open '${
         localStartFile || 'test/flpSandbox.html'
-    }${params}'`;
+    }${flpParams}'`;
+
     const startNoFlpCommand = localOnly
         ? `echo \\"${t('info.mockOnlyWarning')}\\"`
-        : `fiori run --open '${'index.html'}${urlParam}'`;
+        : `fiori run --open '${'index.html?'}${searchParams}'`;
```
slin-sap, line 49: Same here if searchParams is empty.

IainSAP: Not sure thats a concern since its initialised.

devinea (approved):

packages/fiori-freestyle-writer/src/packageConfig.ts
IainSAP: @slin-sap Should I make the cache param configurable via an additional function param, is it likely to be toggled?

slin-sap: @IainSAP We never know if we might turn off this cache param and left with a single optional param again. Despite this, I think it is better to separate the duties. 

E.g.

```
// Complete build of query params of url
searchParams = new URLSearchParams(addParams);
searchParams = searchParams ? searchParams : '';
// Complete build of hash fragment of url
hashFragment = flpAppId ? `#${flpAppId}` : '';


// Building the full params of the url can be simply concatenate 
// the query part and the hash part together, since previously code section already handled the empty string cases.
const flpParams = `${searchParams}${hashFragment}`;

// Finally, in the code section for building various start commands, 
// no need to think about if flpParams or searchParams have provided ? / # prefix. 
// Only need to focus on concatenating the url param parts to the html page path. 
const startLocalCommand = `fiori run --config ./ui5-local.yaml --open '${
    localStartFile || 'test/flpSandbox.html'
}${flpParams}'`;
    
const startNoFlpCommand = localOnly
  ? `echo \\"${t('info.mockOnlyWarning')}\\"`
  : `fiori run --open '${'index.html'}{searchParams}'`;
```

IainSAP: Sorry , I dont see an issue with the current implementation. Can you take over the PR please?

Issue #294, opened 2022-01-31
FEATURE - Optionally write local annotations file from `odata-service-writer` generate API


IainSAP:
As an app. developer, I want to optionally write a local annotations file 'annotation.xml' when adding services via the `odata-service-writer` module, so that this information is available to the applications.

### Description

Local annotations files are required for the addition of local annotations. This include schema references to annotation vocabularies and a reference to the local copy of backend annotations. When additional datasources are added we should also allow additional local annotation files and `manifest.json` entries to be generated with a provided name.

### Technical Design
An optional property `localAnnotationsName` will be added to the `OdataService` interface to allow the annotations references file to be written by the `odata-service-writer` module in the form:

```
interface OdataService {
...
    localAnnotationsName?: string;
...
}
```

Since it is only required for Fiori elements writer currently if the  local service name is omitted then neither the file nor the manifest.json datasource entry will be written.

```
"dataSources": {
            "mainService": {
                ...
                "settings": {
                    "annotations": [
                        "<localAnnotationsName>"
                    ],
                   ...
                }
            },
            "<localAnnotationsName>": {
                "type": "ODataAnnotation",
                "uri": "annotations/<localAnnotationsName>.xml",
                "settings": {
                    "localUri": "annotations/<localAnnotationsName>.xml"
                }
            }
        },
```

### Acceptance Criteria

When adding a service, passing the optional property `localAnnotationsName` to any value, a file `<localAnnotationsName>.xml` will be written to the folder `<app-root>/webapp/annotations/`. A corresponding entry in the application `manifest.json` will be added also.

### Notes

Addition of multiple annotation datasource references to a single datasource is not covered by this TBI.

### Tasks
- [ ] Update `odata-service-writer` `generate` handle new `OdaraService` property 'localAnnotationsName'
- [ ] Unit Tests added

PR #298, opened 2022-01-31
feat(odata-service-writer): Adds local annotations support

IainSAP:
https://github.com/SAP/open-ux-tools/issues/294 

Needed for yet to be created `fiori-elements-writer`.


IainSAP:
Additional test required to cover uncovered loc

packages/odata-service-writer/src/data/annotations.ts
```diff
@@ -0,0 +1,97 @@
...
+        throw new Error(t('error.invalidXML'));
+    }
+    const options = {
+        attributeNamePrefix: '',
+        ignoreAttributes: false,
+        ignoreNameSpace: true,
+        parseAttributeValue: true
+    };
+    const parser = new XMLParser(options);
+    return parser.parse(xml);
```
devinea, line 47: Need a try catch here on the call to 'parse'?

packages/odata-service-writer/templates/extend/manifest.json
```diff
@@ -6,7 +6,8 @@
                 "type": "OData",
                 "settings": {
                     "annotations": [<%if (locals.annotations && annotations.technicalName) {%>
-                       "<%- annotations.technicalName %>"<% } %>
+                        "<%- annotations.technicalName %>"<% if (locals.localAnnotationsName) {%>,<%}%><% } %><% if (locals.localAnnotationsName) { %>
+                        "<%- localAnnotationsName %>"<% } %>
                     ],
                     "localUri": "localService/metadata.xml"<%if (version === '4') {%>,
                     "odataVersion": "4.0"<% } %><%if (version === '2') {%>,
```
devinea, line 13: For completeness shouldn't we add odataVersion for V2 also. Makes it easier for other tools to check when working on the project.

packages/odata-service-writer/test/index.test.ts
```diff
@@ -73,4 +73,41 @@ describe('ODataService templates', () => {
         const fsEditor = await generate(testDir, validServiceConfig as OdataService, fs);
         expect((fsEditor as any).dump(testDir)).toMatchSnapshot();
     });
+
+    it('generate: project with local annotations', async () => {
+        const serviceConfigWithAnnotations: OdataService = {
+            url: 'http://localhost',
+            path: '/sap/odata/testme',
+            version: OdataVersion.v2,
+            metadata: readFileSync(join(__dirname, 'test-data', 'sepmra_prod_man_v2', `metadata.xml`), 'utf-8'),
```
devinea, line 82: test is async. Should be able to use the async read file too?

devinea (commented):
@IainSAP - Overall looks good. few minor comments

packages/odata-service-writer/test/index.test.ts
IainSAP: I would still have to await - which effectively makes it sync again. But I dont think the test should be async at all so I can remove that thx

packages/odata-service-writer/src/data/annotations.ts
IainSAP: I dont think so since we validate the xml here https://github.com/SAP/open-ux-tools/pull/298/files/3e6e0ce7283018ce3d19fd73d0bc3f7169f28155#diff-f507de448682b891118eb30af7f854e31a976fa9ccdcfc1818dea12ef9a296eaR37

packages/odata-service-writer/templates/extend/manifest.json
IainSAP: We do here https://github.com/SAP/open-ux-tools/pull/298/files/3e6e0ce7283018ce3d19fd73d0bc3f7169f28155#diff-36edc1a05a6eaa8b903f997faf779f4e228cf7ccd8aab52c8c124e6c520bf648L13 ;)

devinea (dismissed):
Thanks @IainSAP for the updates

unseen1980 (dismissed):

packages/odata-service-writer/src/data/annotations.ts
```diff
@@ -0,0 +1,101 @@
...
+export function getAnnotationNamespaces({ metadata, annotations }: Partial<OdataService>): NamespaceAlias[] {
+    // enhance service with annotations namespaces
+    const schemaNamespaces = metadata ? getNamespaces(metadata) : [];
+
+    return schemaNamespaces.map((schema: NamespaceAlias) => {
+        // Check if alias exists in backend annotation file, if so use it
+        if (annotations) {
+            const annotationAlias =
+                annotations.xml && schema.namespace
+                    ? getAliasFromAnnotation(annotations.xml, schema.namespace)
```
ullasholla, line 22: Does it make sense to memoize this function call or store an intermediate result? I'm not sure how performant XML parsing is.

packages/odata-service-writer/src/index.ts
```diff
@@ -88,26 +88,40 @@ async function generate(basePath: string, data: OdataService, fs?: Editor): Prom
...
+            prettifyXml(service.metadata, { indent: 4 })
+        );
+
+        // Adds local annotations to datasources section of manifest.json and writes the annotations file
+        if (service.localAnnotationsName) {
+            const namespaces = getAnnotationNamespaces(service);
+            fs.copyTpl(
+                join(templateRoot, 'add', 'annotation.xml'),
+                join(basePath, 'webapp', 'annotations', `${service.localAnnotationsName}.xml`),
+                { ...service, namespaces }
```
ullasholla, line 109: Do we use the `namespaces` in the `manifest.json` template?

ullasholla (changes requested):
Hi @IainSAP, thank you. The PR looks good. There is good test coverage.
Had a couple of minor points, could you look at them when you can please?

packages/odata-service-writer/src/index.ts
IainSAP: No, it's only used in the local annotation file.

packages/odata-service-writer/src/data/annotations.ts
IainSAP: Memoize pattern is generally useful when the same combination arguments are passed multiple times, which in the case of `getAliasFromAnnotation` wont happen (unless there are duplicated `namespace` references). However, it would might be beneficial to only parse the same xmls once for sure. Im guessing thats what you are referring to. 

In the case of multiple namespaces the same `annotation` xml will be parsed repeatedly. We would need to generate an id for each passed xml as an equality check and store this along with the parsed result. Im not sure its worth the effort currently as `fast-xml-parser` is quite performant. @devinea @tobiasqueck any thoughts?

IainSAP: Alternatively we could parse the `annotation` xml before the call to `getAliasFromAnnotation` and pass the parsed result. This would save some processing and is a minor update, I think this makes sense.

devinea (approved):

devinea (approved):

ullasholla (approved):
@IainSAP, thanks for addressing the review comments.

PR #302, opened 2022-02-02
chore: update ui5-app-writer jest config

kranthie-sap:
* Update `jest.config` in `ui5-application-writer` module by removing the `roots` setting to check if it resolves the sonar coverage issue.


donal-tobin-sap:
The templates folder needs to be excluded from jest as the template files are not valid js and cannot be parsed.

Issue #303, opened 2022-02-02
BUG - dot files missing in writers with pnpm >=6.11.0-0


devinea:
### Related Feature
Feature request: _issue number_

### Description
After upgrading pnpm in https://github.com/SAP/open-ux-tools/pull/283 the published npm module is missing `/templates/core/.gitignore` in `/packages/ui5-application-writer` for example.
Bug logged with pnpm https://github.com/pnpm/pnpm/issues/4293

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

devinea:
Workaround applied to `pnpm pack` bug.

PR #305, opened 2022-02-02
bug(ui5-app) - missing dot files. Rename dot files to .tmpl file in the templates due to bug in pnpm

devinea:
#303


ullasholla (approved):
Thanks @devinea

PR #306, opened 2022-02-02
fix(ui5-application-writer): change eslint template file

docirl:
#303 

Revert changes to eslint template, the file is not getting renamed at generation. Previous implementation was working correctly.


devinea (approved):
Thanks @docirl

ullasholla (approved):

Issue #308, opened 2022-02-04
BUG - ui5-application-writer is not writing dependencies to manifest.json


IainSAP:
### Description
Only one ui5 lib dependency is written to the manifest.json file. All dependencies provided through `ui5App.ui5Libs` should appear in the ui5-local.yaml and the manifest.json file.

In addition there is no entry in `manifest.json` for `resources` which should be `"resources.json"` by default convention.

### Steps to Reproduce
Generate a ui5 app using ui5-application-writer providing additional ui5 libs. Only one appears in the manifest file. Also `sap.m` appears by default in the ui5-local.yaml but not in the manifest.json.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

IainSAP:
Fixed in #309

PR #309, opened 2022-02-04
fix(ui5-application-writer): Fixes incorrect deps written to manifest

IainSAP:
https://github.com/SAP/open-ux-tools/issues/308

Add ui5 dependencies to manifest as required. Fixes inconsistent default libs `sap.m` in ui5-local.yaml but `sap.ui.core` in manifest.


zdravko-georgiev (approved):
LGTM

packages/ui5-application-writer/src/data/defaultSettings.ts
```diff
@@ -1,5 +1,6 @@
 import { App, Package, UI5, UI5Framework } from '../types';
```
IainSAP, line 1: File rename is a temp workaround for sonar/lcov issues reporting the whole file as uncovered.

packages/ui5-application-writer/src/data/ui5Libs.ts
```diff
@@ -1,6 +1,14 @@
...
+/**
+ * Merges the specified ui5 libs with the defaults.
+ * 
+ * @param ui5Libs - The ui5 libraries to be merged with the defaults
+ * @returns UI5 libs with defaults
+ */
 export const getUI5Libs = (ui5Libs?: string | string[]): string[] => {
     const libs = Array.isArray(ui5Libs) ? ui5Libs : ui5Libs?.split(',') || [];
-    return ['sap.m'].concat(libs).filter((value, index, self) => {
+    return defaultUI5Libs.concat(libs).filter((value, index, self) => {
```
ullasholla, line 11: Do we need to handle the case where the default libs are also passed in? We'll have duplicated entries then?

ullasholla (dismissed):
Thanks @IainSAP, the fix looks good.
One question though: https://github.com/SAP/open-ux-tools/pull/309/files#r799688307

packages/ui5-application-writer/src/data/ui5Libs.ts
IainSAP: That occurred to me also, but this filter condition removes duplicates.

longieirl (approved):

devinea (approved):
@IainSAP looks great. Thanks.

Issue #318, opened 2022-02-08
BUG - ui5-application-writer is writing wrong UI5 versions


zdravko-georgiev:
### Description
Based on the [tests](https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-application-writer/test/data.test.ts) it is visible that the `ui5-application-writer` will write UI5 version in some cases without the patch number, e.g. `1.80`. Having an UI5 version without the patch number will result in the application not loading, because the proxy will forward the request, e.g. `/resources/sap-ui-core.js` --> `https://ui5.sap.com/1.80/resources/sap-ui-core.js`. The request will result in 404. Hence when sending a request to `https://ui5.sap.com` for a specific UI5 version, the version needs to contain the patch number.

### Expected results
All UI5 versions written by the `ui5-application-writer` have the patch number included.

### Actual results
In some cases the UI5 version written by the `ui5-application-writer` does not contain the patch number

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #319, opened 2022-02-08
Fix/318/fix invalid semantic versions

IainSAP:
Fix for: https://github.com/SAP/open-ux-tools/issues/318

Adds proper version -> semantic version conversion so that we no longer write invalid versions to yamls and manifest.json. Doing so was causing failures to load UI5 libraries.

Also, adds localisation strings for `ui5-application-writer`.

Updates ui5 version -> manifest schema version from https://github.com/SAP/ui5-manifest/blob/master/mapping.json

Updates to properly handle snapshot versions when comparing versions and looking up manifest descriptor version.


zdravko-georgiev (dismissed):
Thanks @IainSAP

packages/ui5-application-writer/src/data/validators.ts
```diff
@@ -9,11 +11,27 @@ import { Ui5App } from '../types';
  */
 export function validateAppId(appId: string): boolean {
     if (!appId) {
-        throw new Error('The property: app.id must have a value'); // todo: add localised parameterised message
+        throw new Error(t('error.missingRequiredProperty', { propertyName: 'app.id' }));
```
IainSAP, line 14: Test needed to cover : https://sonarcloud.io/code?id=SAP_open-ux-tools&pullRequest=319&selected=SAP_open-ux-tools%3Apackages%2Fui5-application-writer%2Fsrc%2Fdata%2Fvalidators.ts&line=33

IainSAP:
> @IainSAP changes look good. But can you add tests for UI5 snapshot versions. E.g. `snapshot-1.98` `semVer.coerce` might remove `snapshot-` in the minUI5Version incorrectly.

Yes it will remove that, currently adding tests.

packages/ui5-application-writer/test/data.test.ts
```diff
@@ -113,6 +113,42 @@ describe('Setting defaults', () => {
...
+        {
+            input: {
+                version: 'snapshot-1.80',
+                frameworkUrl: 'https://ui5.unreleased.sap.com'
+            },
+            expected: {
+                framework: 'SAPUI5',
+                frameworkUrl: 'https://ui5.unreleased.sap.com',
+                version: 'snapshot-1.80',
+                localVersion: 'snapshot-1.80',
```
devinea, line 126: Local version should be `1.80.0`

packages/ui5-application-writer/test/data.test.ts
```diff
@@ -113,6 +113,42 @@ describe('Setting defaults', () => {
...
+                minUI5Version: '1.60.0',
```
devinea, line 127: minUI5Version should be ‘snapshot-1.80’
@zdravko-georgiev correct me if i am wrong.

packages/ui5-application-writer/src/data/defaults.ts
```diff
@@ -75,19 +80,41 @@ export function mergeUi5(ui5: Partial<UI5>): UI5 {
         framework,
         frameworkUrl: ui5.frameworkUrl ?? defaultFrameworkUrl
     };
-    const typesVersion = parseFloat(merged.localVersion) >= 1.76 ? merged.localVersion : '1.71.18';
+    // typesVersion must be a valid npm semantic version, we know they cannot be null as already validated
+    const localSemVer = semVer.valid((semVer.coerce(merged.localVersion)))!;
+    const typesVersion = semVer.gte(localSemVer, '1.76.0') ? localSemVer : '1.71.18';
```
ullasholla, line 85: Use constants for `'1.76.0'` and `'1.71.18'` too? I know this PR does not add this code but touches it.

packages/ui5-application-writer/src/data/defaults.ts
```diff
@@ -113,8 +140,13 @@ function getLocalVersion({
...
         }
-        if (parseFloat(version) > parseFloat(result)) {
-            result = version;
+
+        if (version !== UI5_DEFAULT.DEFAULT_UI5_VERSION) {
+
+            // Update to a valid coerced version string e.g. 1.80 -> 1.80.0. Cannot be null as previously validated.
+            if (semVer.gt(semVer.coerce(version)!, semVer.coerce(result)!)) {
+                result = version;
+            }
```
ullasholla, line 149: Is the `else` before the addition (line 140) the same as the this condition: `if (version !== UI5_DEFAULT.DEFAULT_UI5_VERSION)`? If so, there seems to be a chance an open UI5 version could get overwritten by the addition.

packages/ui5-application-writer/src/i18n.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Helper function facading the call call to i18next.
+ *
+ * @param key i18n key
+ * @param options additional options
+ * @returns {string} localized string stored for the given key
+ */
+export function t(key: string, options?: TOptions): string {
+    return i18next.t(key, options);
+}
```
ullasholla, line 32: Can we directly use the `i18next.t()` here instead of wrapping it?

ullasholla (changes requested):
Thanks for the fix. Code looks good. Test coverage is good. Using `semver` is a good idea.

Have a couple of questions? Could you take a look?

packages/ui5-application-writer/test/data.test.ts
IainSAP: @devinea This is how it was implemented...if `minUI5Version` is not provided it defaults to `1.60` https://github.com/SAP/open-ux-tools/pull/319/files#diff-0f207248944a1732a5b7343b9d1bb29e22bbae5fa2acce7a573a344c04c856d2L72

But I think you are correct, I will update to use the current version provided if not present.

packages/ui5-application-writer/src/data/defaults.ts
IainSAP: I believe `result` would be overwritten by the addition, at least that's how it was previously implemented. But I see we can refactor here to make the intention clearer.

packages/ui5-application-writer/test/data.test.ts
zdravko-georgiev: @devinea @IainSAP What was the `localVersion` again? The version for the `ui5-local.yaml`? If yes, then `localVersion` can't be a snapshot. Only released UI5 versions 1.76.0 and above.

packages/ui5-application-writer/test/data.test.ts
zdravko-georgiev: @devinea @IainSAP We need to be careful here, If I am not mistaken snapshot versions are not valid for the `manifest.json`.

packages/ui5-application-writer/src/i18n.ts
IainSAP: I believe it's done like this due ordering of `i18next.init` on load of the file and we don't want consumers to have to manage that process. Also this is replicated code across many modules and should be consolidated to a single i18n utility in another TBI and PR so I will leave it as is for now.

packages/ui5-application-writer/test/data.test.ts
devinea: @zdravko-georgiev  Yeah `localVersion` is the one for `ui5-local.yaml`
@IainSAP  - Okay yeah I guess if `localVersion` is passed in as `snapshot-1.80` it should be output the same. Even if it is not really valid for the ui5-local.yaml.

packages/ui5-application-writer/src/data/version-to-descriptor-mapping.json
```diff
@@ -1,5 +1,10 @@
 {
-    "latest": "1.33.0",
+    "latest": "1.38.0",
+    "1.98": "1.38.0",
+    "1.96": "1.37.0",
+    "1.95": "1.36.0",
+    "1.93": "1.35.0",
+    "1.92": "1.34.0",
```
devinea, line 7: Thanks

devinea (approved):
Thanks @IainSAP

IainSAP:
Thx all

PR #332, opened 2022-02-17
fix(ui5-application-writer): Adds reuse lib loading to index.html

### LINKED ISSUE #331: BUG -  Reuse libs are not loaded for Freestyle apps index.html
### Description
When starting generated Fiori Freestyle apps the reuse libs specified in the manifest.json are not being loaded. This is becuase the `locate-reuse-libs.js` script is not loaded by the `index.html` file. 

### Steps to Reproduce
Generate a freestyle app. Add a re-use lib to the `manifest.json` file. Run `npm start-noflp`. The specified libs fail to load with 404s.

### Expected results
The reuse libs are loaded.

### Screenshots

OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

IainSAP:
Fixes #331

Adds optional support for reuse lib loading to ui5-application-writer


zdravko-georgiev:
Shouldn't we remove the mock server part from the [locate-reuse-libs.js](https://github.com/SAP/open-ux-tools/blob/3e66342c56b2a64e58fd618df8d7de0f4c93542c/packages/ui5-application-writer/templates/optional/loadReuseLibs/webapp/utils/locate-reuse-libs.js#L141-L149). For newly generated apps we are using now only the `"@sap/ux-ui5-fe-mockserver-middleware"`, right?

IainSAP:
> Shouldn't we remove the mock server part from the [locate-reuse-libs.js](https://github.com/SAP/open-ux-tools/blob/3e66342c56b2a64e58fd618df8d7de0f4c93542c/packages/ui5-application-writer/templates/optional/loadReuseLibs/webapp/utils/locate-reuse-libs.js#L141-L149). For newly generated apps we are using now only the `"@sap/ux-ui5-fe-mockserver-middleware"`, right?

@devinea Would like to keep it here, but Im open to removing it.

docirl (approved):

packages/fiori-freestyle-writer/src/defaults.ts
```diff
@@ -28,4 +28,6 @@ export function setDefaults(ffApp: FreestyleApp<unknown>): void {
     if (ffApp.template.type === TemplateType.Basic) {
         setBasicTemplateDefaults(ffApp.template.settings as BasicAppSettings);
     }
+    // All fiori-freestyle apps should use load reuse libs, unless explicitly overridden
+    ffApp.appOptions = Object.assign({ loadReuseLibs: true }, ffApp.appOptions);
```
devinea, line 32: The `loadReuseLibs`  param only applies to the index.html and not to the writing of `flpsandbox.html` and `locate-reuse-libs.js`.
Should be be considered in those cases too or just removed as a param?

IainSAP: The `locate-reuse-libs.js` file is taken care of by the `loadRuseLibs` param but you are correct that the `flpSandbox.html` needs a condition around adding the script reference, thx. Well spotted.

devinea:
> > Shouldn't we remove the mock server part from the [locate-reuse-libs.js](https://github.com/SAP/open-ux-tools/blob/3e66342c56b2a64e58fd618df8d7de0f4c93542c/packages/ui5-application-writer/templates/optional/loadReuseLibs/webapp/utils/locate-reuse-libs.js#L141-L149). For newly generated apps we are using now only the `"@sap/ux-ui5-fe-mockserver-middleware"`, right?
> 
> @devinea Would like to keep it here, but Im open to removing it.

Yeah I am in favour of keeping it as migration will still use ui5 mockserver if the project did before. So we can avoid having different versions on this file to maintain.

devinea (approved):
Thanks @IainSAP 
I haven't tried it locally. But changes look as expected.

Issue #333, opened 2022-02-17
TBI - Remove ui5 version from ui5.yaml and ui5-mock.yaml files


unseen1980:
### Description 
It has been agreed that the middleware should determine the version of ui5 in the generated or migrated Fiori app and it no longer needs to be specified in the `ui5.yaml` and `ui5-mock.yaml` files.

### Acceptance criteria

**Before:**

```
        ui5:
          path:
            - /resources
            - /test-resources
          url: https://ui5.sap.com
          version: 1.96.1
```

**After:**

```
        ui5:
          path:
            - /resources
            - /test-resources
          url: https://ui5.sap.com
```

### Tasks
- [ ] Remove version from ui5-config
- [ ] Update TS types
- [ ] Update tests/snapshots

PR #334, opened 2022-02-17
Tbi/ui5version in yaml/17765

unseen1980:
For TBI https://github.com/SAP/open-ux-tools/issues/333

- Remove version from ui5-config
- Update TS types
- Update tests/snapshots

### DIFF
--- a/packages/ui5-config/src/types.ts
+++ b/packages/ui5-config/src/types.ts
@@ -15,7 +15,6 @@ export interface ProxyUIConfig {
     [key: string]: unknown | undefined;
     path?: string[];
     url?: string;
-    version?: string;
     directLoad?: boolean;
 }
 
--- a/.changeset/tasty-lies-promise.md
+++ b/.changeset/tasty-lies-promise.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui5-config': minor
+---
+
+UI5 version removal from ui5.yaml and ui5-mock.yaml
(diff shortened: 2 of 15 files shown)


IainSAP (commented):
@unseen1980 I think we can remove this now https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-application-writer/src/index.ts#L38 ?

IainSAP (approved):
OK for me, not sure we need 2 change sets? Removal of version from ui5-config api doesn't result in an API change for ui5-application-writer. Thx @unseen1980

packages/ui5-config/src/types.ts
```diff
@@ -15,7 +15,6 @@ export interface ProxyUIConfig {
     [key: string]: unknown | undefined;
     path?: string[];
     url?: string;
-    version?: string;
```
zdravko-georgiev, line 18: The `version` is still a valid property. It just doesn't need to be set by default. I would keep it in the type's definition.

zdravko-georgiev (changes requested):

packages/ui5-config/src/types.ts
IainSAP: Isnt it redundant as its not referenced anymore?

zdravko-georgiev (approved):
LGTM

Klaus-Keller (approved):
Did a review of code, no build and test. However, out of curiosity I've tried to remove the `version` property from `ui5.yaml` und started the preview using an older `@sap/ux-ui5-middleware` (1.4.1) and it still worked. This situation should not happen as only new generated projects have the `version` removed, but you never know.

PR #344, opened 2022-02-23
chore: rm redundant lerna config file

ullasholla:
This PR removes the redundant `lerna.json` file. We use `pnpm` to manage our workspaces and this is a holdover from the past when `lerna` was being used.


kranthie-sap (approved):
Thanks, Ullas.

kranthie-sap (changes requested):
@ullasholla: Can you also remove the references in `.gitignore` and `CODEOWNERS`, please?

ullasholla:
@kranthie-sap:
> @ullasholla: Can you also remove the references in `.gitignore` and `CODEOWNERS`, please?

This entry in `.gitignore`?
```
lerna-debug.log*
```
I left than in because we also have a bunch of references to temp files from other tools that we don't use (grunt, istanbul, bower, etc).

Did not find anything in `CODEOWNERS`. Maybe I'm looking in the wrong place?

kranthie-sap:
> Did not find anything in `CODEOWNERS`. Maybe I'm looking in the wrong place?

@ullasholla: It looks like the latest version of the `CODEOWNERS` has an entry for `lerna.json`.

kranthie-sap (approved):

donal-tobin-sap (approved):

Issue #353, opened 2022-02-25
BUG - odata-service-writer: Annotations Technical Name is not uri encoded


IainSAP:
### Description
Generating an app with a service, passing annotations in the format:
```
annotations: {
            technicalName: '/NS/SERVICE',
            xml: '<HELLO><ANNOTATION></ANNOTATION></WORLD></HELLO>'
        }
```
results in an non-loading annotation file in the resulting application due to the entry generated in the `manifest.json`:
```
"/NS/SERVICE/": {
        "uri": "/sap/opu/odata/IWFND/CATALOGSERVICE;v=2/Annotations(TechnicalName='/NS/SERVICE',Version='0001')/$value/",
        "type": "ODataAnnotation",
        "settings": {
          "localUri": "localService/NS/SERVICE.xml"
        }
      },
```

### Steps to Reproduce
Generate a Fiori Freestyle application passing an annotation technical name containing path segments.

### Expected results
`manifest.json` contains correctly encoded path segments in the annotation uri :
e.g.
```
"/NS/SERVICE/": {
        "uri": "/sap/opu/odata/IWFND/CATALOGSERVICE;v=2/Annotations(TechnicalName='%2FNS%2FSERVICE',Version='0001')/$value/",
        "type": "ODataAnnotation",
        "settings": {
          "localUri": "localService/NS/SERVICE.xml"
        }
      },
```

### Actual results
```
"/NS/SERVICE/": {
        "uri": "/sap/opu/odata/IWFND/CATALOGSERVICE;v=2/Annotations(TechnicalName='/NS/SERVICE',Version='0001')/$value/",
        "type": "ODataAnnotation",
        "settings": {
          "localUri": "localService/NS/SERVICE.xml"
        }
      },
```

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

IainSAP:
Fixed by: https://github.com/SAP/open-ux-tools/pull/354

PR #354, opened 2022-02-25
fix(odata-service-writer): Fix for annotations technical uri encoding

IainSAP:
https://github.com/SAP/open-ux-tools/issues/353

Also fixes variable version in `fiori-elements-writer` test which would otherwise block subsequent PRs


Klaus-Keller (approved):
I've checked the test adjustments and the code added to encode the technical name. Looks good to me.

unseen1980 (approved):

theStiggySpd:
With 2 approvals (considering the absence of module experts), I am merging this pr

PR #359, opened 2022-03-01
Bug/312/invalid json

### LINKED ISSUE #312: BUG - FPM Invalid entry generated for custom column
### Related Feature
FPM Writer

### Description
Following the Fiori Elements V4 information it shall be possible to generate a custom column without `anchor`. 
When trying this with help of the FPM writer I get an exception afterwards, as the generated entry for the manifest is syntactically incorrect ( one `,`too much):

image

### Steps to Reproduce
Use the API `generateCustomColumn` of https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/column/index.ts with supplying `position.anchor.`

### Expected results
one `,`less -  syntactically correct JSON gets generated.

### Actual results
Syntactically wrong JSON

 "@sap-ux/fe-fpm-writer": "0.4.0"

OS:                      
- [X] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

tobiasqueck:
Fixes #312 and enforces use of `vm2` with version `>=3.9.6` due to https://github.com/advisories/GHSA-6pw2-5hjv-9pf7


package.json
```diff
@@ -49,7 +49,8 @@
     "pnpm": {
         "overrides": {
             "ansi-regex": "^5.0.1",
-            "tmpl": "^1.0.5"
+            "tmpl": "^1.0.5",
+            "vm2": ">=3.9.6"
```
donal-tobin-sap, line 53: I believe that ^ is the more appropriate form for the overrides https://pnpm.io/package_json#pnpmoverrides
Note also that there are a number of breaking changes, issues and performance issues fixed in these versions 3.9.6 to 3.9.9.

tobiasqueck: Thanks for the link! Based on it, I would go even further and use `"vm2@<3.9.6": "^3.9.6"`.

stefanschreck (approved):

donal-tobin-sap (approved):
👍🏽 Looks good to me.

Issue #358, opened 2022-03-01
TBI - Add telemetry setting service to be managed in @sap-ux/store 


slin-sap:
### Description (include screenshots)
What will this TBI achieve, and how will it be measured as done. Is it documentation, increase of test coverage, refactoring

Add a telemetry setting config in store module at <homedir>/.fioritools. 

### Value
It provides a central telemetry setting config for all fiori tool extensions and npm modules.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [x] Add telemetry setting data provider and service
- [x] Unit test

PR #362, opened 2022-03-01
tbi: telemetry settings store

slin-sap:
For TBI https://github.com/SAP/open-ux-tools/issues/358

Temporary implement of telemetry setting config using existing store framework so it doesn't block the main central telemetry config user story. Will move the telemetry specific implementation out once the store module is refactored.


packages/store/src/entities/telemetry-setting.ts
```diff
@@ -0,0 +1,19 @@
+import { EntityKey } from '.';
+import { serializable } from '../decorators';
+
+/** This represents the migration status of the systems from pure secure store
```
tobiasqueck, line 4: I don't understand this comment

tobiasqueck (changes requested):
Link to actual issue is missing (I am assuming it is #358) and neither issue nor PR contain sufficient information to understand why this is done.

ullasholla (approved):
Thanks @slin-sap, looks good. Test coverage is good too.

tobiasqueck (approved):
The code is simple and well structured. I am missing comments for each function, however, since the functions are self-explanatory I am not seeing it as a reason to block the merge.

Issue #363, opened 2022-03-02
BUG - fiori-elements-writer: ALP issues


IainSAP:
### Description
Reviewing the tests for `@sap-ux/fiori-elements-writer` it can be seen that the wrong template type is used for the Analytical List Page v2 tests here https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-elements-writer/test/alp.test.ts#L61.

Some alp settings booleans are added to manifest as strings e.g.:

```
"autoHide": "true",
"smartVariantManagement": "true",
```

The app will fail to load.

### Steps to Reproduce
Run the tests from `@sap-ux/fiori-elements-writer`. Notice that the test outputs for `alpTemplates/alpV2` is using List Report instead of AnalyticalListPage for the `pages` definition.

### Expected results
ALP v2 test generates an ALP v2 floorplan app.

### Actual results
ALP v2 test generates an LROP v2 floorplan.
When generating using the correct floorplan the app fails to load due booleans set as string values.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #366, opened 2022-03-02
fix(fiori-elements-writer): Fix for #363 - incorrectly configured test alp v2

IainSAP:
Fixes: #363 

Correct test inputs and fixes ALP boolean settings being written as strings instead of booleans.


packages/fiori-elements-writer/src/data/defaults.ts
```diff
@@ -11,8 +11,10 @@ import { getBaseComponent, getUi5Libs } from './templateAttributes';
 export function setDefaultTemplateSettings<T>(template: Template<T>): T {
     const templateSettings = template.settings;
     if (template.type === TemplateType.AnalyticalListPage) {
-        const alpSettings = template.settings as Partial<ALPSettingsV2>;
+        const alpSettings = template.settings as Partial<ALPSettingsV2 & ALPSettingsV4>;
 
+        // Assign defaults for properties of all ALP settings interfaces.
+        // Since only referenced properties are written by templates this has no side-effects.
```
tobiasqueck, line 17: I am not in favor of this statement and the following implementation. Yes, right now, our templating explicitly takes the properties required, however, in other templates e.g. in `fe-fpm-writer` we also use a `JSON.stringify` in one place to set properties in the `manifest.json`. Maybe someone else will use the very same in the future to serialize the settings in the ALP templates resulting in invalid generated manifests. Of course, I would expect our test cases to catch it but it wouldn't change the fact that we add wrong settings here that just by knowledge of other parts of code don't create a problem.
Please look at the function as an encapsulated unit that should behave as it would be expected.

packages/fiori-elements-writer/test/__snapshots__/alp.test.ts.snap
```diff
@@ -5340,20 +5340,22 @@ xmlns:edmx=\\"http://docs.oasis-open.org/odata/ns/edmx\\">
   },
   \\"sap.ui.generic.app\\": {
     \\"pages\\": {
-      \\"ListReport|SEPMRA_C_ALP_SlsOrdItemCubeALPResults\\": {
+      \\"AnalyticalListPage|SEPMRA_C_ALP_SlsOrdItemCubeALPResults\\": {
```
tobiasqueck, line 5343: Where is this coming from? Is it just a change in the test resulting into this new page id? I guess it doesn't matter what the exact string is, as long as it is unique but still wondering if this is expected.

tobiasqueck (changes requested):

packages/fiori-elements-writer/test/__snapshots__/alp.test.ts.snap
tobiasqueck: Just read the issue, and understand it now. Thanks!

IainSAP: The test was setting the wrong template type - thats part of the fix here.

packages/fiori-elements-writer/src/data/defaults.ts
IainSAP: OK, the alternative is rather ugly casts and type guards (which is how I originally implemented), but Ill update for your comment.

tobiasqueck: Maybe I am missing something but I'd assumed it is 'just' another nested `if/else` with an `Object.assign` in each block?

IainSAP: Almost yes - https://github.com/SAP/open-ux-tools/pull/366/commits/8f50dcb50162348bd2ea5b30b519f6fde76eba8a#diff-ff4edbeac8e42e9c7ce005e9ba864236b8e5909f154fde419beb25fa1098fa06R23-R50

IainSAP: Updated to use odata version as discussed.

tobiasqueck: Thank you!

tobiasqueck (approved):
reviewed and no objections to 
* updated settings handling v2 vs. v4 
* updated test case
* formatting changes

devinea (approved):

PR #371, opened 2022-03-03
fix(fe-fpm-writer): namespace resolution handle backslashes

815are:
While working on integration of custom sections(custom sections was 3rd point from https://github.com/SAP/open-ux-tools/issues/73)

Issue relates to custom extension like Custom Section, Custom Column.
Issue happens if we pass `folder` property with backslashes `ext\\fragment`, then it would resolved as:
```
"fragmentName": "project4.ext\fragment.Sefse",
```

Issue does not happen if we pass folder as `ext/fragment`.

In result extending regex for replace from `/\//g` to `/\/|\\/g`

### DIFF
--- a/.changeset/metal-ads-tell.md
+++ b/.changeset/metal-ads-tell.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fe-fpm-writer': minor
+---
+
+custom extension - consider backslashes when "folder" converted into namespace for "template"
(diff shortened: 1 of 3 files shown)


tobiasqueck (approved):
* clean and simple fix
* properly tested

zdravko-georgiev (approved):

PR #383, opened 2022-03-09
chore(open-ux-tools) - update dependencies

devinea:
chore(open-ux-tools) - update dependencies 

add exclusions for `typescript` `@types/node` from dependabot.

**downgrade `typescript` to 4.0.8 for wider compatibility**


.github/dependabot.yml
```diff
@@ -18,3 +18,8 @@ updates:
       interval: "weekly"
     allow:
       - dependency-type: "direct"
+    ignore:
```
ullasholla, line 21: Do we need to add `ts-node` here as well?

ullasholla:
Thanks @devinea, looks good. Had a question about excluding `ts-node` as well. Could you have a look please?

.github/dependabot.yml
devinea: @ullasholla looks like we don't really use `ts-node`. Scripts point to `test/index.ts` that do not exist https://github.com/SAP/open-ux-tools/search?q=ts-node

Is it ok to remove `test:local` scripts and `ts-node` dependency?

packages/fiori-elements-writer/package.json
```diff
@@ -11,7 +11,6 @@
         "lint:fix": "eslint . --ext .ts --fix",
         "test": "jest --ci --forceExit --detectOpenHandles --colors",
         "test-u": "jest --ci --forceExit --detectOpenHandles --colors -u",
-        "test:local": "node -r ts-node/register test/index.ts",
```
ullasholla, line 14: Thanks for cleaning this up. What was this script meant to be test? Can't recall now.

.github/dependabot.yml
ullasholla: The `examples` in the `axios-odata` branch use `ts-node`: https://github.com/SAP/open-ux-tools/blob/b34532e9c155a27e48639d3edf32152e37654324/examples/axios-odata/package.json#L9

Maybe keep `ts-node`? @tobiasqueck?

tobiasqueck: I am ok to remove it and require to run `pnpm build` before running `pnpm start`. I mainly use it because I am lazy and  I don't want to run two scripts which I guess is not a valid reason to keep it ...

packages/fiori-elements-writer/package.json
tobiasqueck: That was one of my old scripts that we removed when publishing the elements templates.

.changeset/thick-geckos-reply.md
```diff
@@ -0,0 +1,15 @@
...
+'@sap-ux/fiori-elements-writer': patch
+'@sap-ux/fiori-freestyle-writer': patch
+'@sap-ux/logger': patch
+'@sap-ux/odata-service-writer': patch
+'@sap-ux/store': patch
+'@sap-ux/ui5-application-writer': patch
+'@sap-ux/ui5-config': patch
+'@sap-ux/ui5-proxy-middleware': patch
+'test-project': patch
+'@sap-ux/yaml': patch
```
tobiasqueck, line 12: That might be a stupid question but why do we need to patch the modules? You only updated `devDependencies` so the resulting built module doesn't change, or do we expect that the `dist` content changes due to the change of the `typescript` version?

tobiasqueck (commented):
Changes look good, just one question about the changeset.

tobiasqueck (approved):

Klaus-Keller (approved):
- version changes look good. 
- clean up of test scripts are good.

approved

.github/dependabot.yml
ullasholla: I'm not fussed either way. Generally I'm a fan of lowering the barrier to explore and interact.

ullasholla (approved):
Thanks @devinea, lgtm

PR #398, opened 2022-03-22
minimist upgrade and add missing properties to package.json

tobiasqueck:
1. added override for `minimist` to remove usage of vulnerable version
2. added `bugs` and `repository` to each `package.json` to make it easier for consumers to navigate our monorepo
3. added missing `descriptions` to `package.json` if they were missing

Btw, I checked the commits since `minimist@1.2.5` (https://github.com/substack/minimist/commits/master) to make sure that it only contains the fix and not anything else.


zdravko-georgiev (approved):
The labels `logger`, `store`, `ui5-proxy-middleware` and `yaml` are not defined. Otherwise looks good.

tobiasqueck:
@zdravko-georgiev labels added, please check again.

zdravko-georgiev (approved):

IainSAP (approved):

longieirl (approved):

donal-tobin-sap (approved):
👍🏽 Looks good to me.

theStiggySpd:
@tobiasqueck Can we consider adding a prefix (e.g. `module:` or something similar) to respective module labels so that they show up grouped when someone is creating a bug? Nothing major, but just a convenience...
Right now it shows like this
image

theStiggySpd:
> > Can we consider adding a prefix (e.g. module: or something similar) to respective module labels so that they show up grouped when someone is creating a bug
> 
> If you think it helps, we can do that. Maybe we use the full module name e.g. `@sap-ux/fe-fpm-writer` and the description from the `package.json` as description so that they are grouped and human readable?
> 
> I personally don't like it, and I don't scroll to find labels but start typing and then select. I gave them the same color for grouping :)

its fine I think:-) . No strong opinion. We can change later

Issue #409, opened 2022-03-24
BUG - Optional ui5 version is mandatory


tobiasqueck:
### Related Feature
Feature request: #72

### Description
I have created my own generator using the @sap-ux/fiori-elements-writer module. When calling the `generate` function, I left the `ui5` property `undefined` because it is optional, however, if I do that, then the generator crashes. I can prevent the crash by setting a version like
```
appConfig.ui5 = {
    version: '1.94'
}
```

My example generator is at https://github.com/SAP/open-ux-tools/blob/example-for-writer/examples/simple-generator/src/app/index.ts.

### Steps to Reproduce
Call the generate function from @sap-ux/fiori-elements-writer with all mandatory values BUT none of the optional values.

### Expected results
A valid project is generated without errors.

### Actual results
```
TypeError: Invalid Version: undefined
    at new SemVer (/Users/I821846/SAPDevelop/git/sap/open-ux-tools/node_modules/.pnpm/semver@7.3.5/node_modules/semver/classes/semver.js:19:13)
    at compare (/Users/I821846/SAPDevelop/git/sap/open-ux-tools/node_modules/.pnpm/semver@7.3.5/node_modules/semver/functions/compare.js:3:3)
    at Object.lt (/Users/I821846/SAPDevelop/git/sap/open-ux-tools/node_modules/.pnpm/semver@7.3.5/node_modules/semver/functions/lt.js:2:29)
    at validateApp (/Users/I821846/SAPDevelop/git/sap/open-ux-tools/packages/fiori-elements-writer/dist/validate.js:42:26)
    at /Users/I821846/SAPDevelop/git/sap/open-ux-tools/packages/fiori-elements-writer/dist/index.js:56:36
    at Generator.next (<anonymous>)
```

### Screenshots
n/a

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #410, opened 2022-03-24
Making ui5 configuration optional as defined in the API

tobiasqueck:
Corresponding issue: #409
- corrected the minimum versions for the different floorplans
- added test case that is not providing a ui5 configuration
- fixed the code to support missing ui5 configuration

### DIFF
--- a/.changeset/swift-shoes-smile.md
+++ b/.changeset/swift-shoes-smile.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fiori-elements-writer': patch
+---
+
+Making ui5 configuration optional as defined in the API


packages/fiori-elements-writer/src/data/templateAttributes.ts
```diff
@@ -1,5 +1,8 @@
 import { OdataVersion, TemplateType } from '../types';
 
+// https://ui5.sap.com/versionoverview.html
+export const oldestSupportedUI5Version = '1.38.0';
```
devinea, line 4: @tobiasqueck This is lower than Fiori tools supports for FE V2 today. It might impact start-local, themes, ui5 libs used etc... in the templates.

tobiasqueck: In `main` the value is still `1.0.0`, so I think the `1.38.0` is already an improvement. What higher version would you suggest and why?

IainSAP (commented):
Looks good: 

-  Min ui5 version is set based on min supported version
- Additional test added

IainSAP (dismissed):

packages/fiori-elements-writer/src/data/templateAttributes.ts
tobiasqueck: Should I use `1.71` as it is the lowest supported by the spec (https://www.npmjs.com/package/@sap/ux-specification)?

devinea: @tobiasqueck We use `1.65.0` currently in tool-suite `ui5-info` as min ui5 FE V2. 
Since 1.71.0 is the supported version per https://ui5.sap.com/versionoverview.html and spec at this stage it would probably make sense to update tool-suite to 1.71.0. @IainSAP @theStiggySpd @ianquigley-sap - Do you agree?

tobiasqueck: Changed it to `1.65.0` for now to stay consistent with the Fiori tools, however, I will provide another PR next week with a proposal to cleanup the versioning.

IainSAP (approved):
Thx @tobiasqueck 

-  Starting ui5 version 1.65, as agreed

devinea (approved):
Thanks @tobiasqueck

packages/fiori-elements-writer/test/lrop.test.ts
```diff
@@ -29,6 +29,18 @@ describe(`Fiori Elements template: ${TEST_NAME}`, () => {
                 service: v4Service
             } as FioriElementsApp<LROPSettings>
         },
+        {
+            name: 'lropV4noUi5Version',
+            config: {
+                ...Object.assign(feBaseConfig('felropui5', false), {
```
nlunets, line 35: Isn't that redundant and also potentially wrong due to the assignment side effect ?

An object spread will already create a new object so you don't need to also call object.assign, except if you're really looking for the side effect of updating the feBaseConfig with your new value (like you're doing there).

If this was really intended i would still encourage to rewrite this part as it is very difficult to understand all that is happening here :)

packages/fiori-elements-writer/src/data/defaults.ts
```diff
@@ -80,5 +80,12 @@ export function setAppDefaults<T>(feApp: FioriElementsApp<T>): FioriElementsApp<
         (feApp.service as OdataService).model = defaultModelName;
     }
 
+    // minimum UI5 version depending on the template required
+    feApp.ui5 = feApp.ui5 ?? {};
```
nlunets, line 84: Why ?? and not || ?

tobiasqueck: It doesn't make a difference here but I prefer it with strings to accept an empty string as input therefore I use it everywhere

packages/fiori-elements-writer/test/lrop.test.ts
tobiasqueck: Argh ... I haven't looked at this one, just copied an existing test case and changed the input for the new case to test. Do you mind if I leave it in, and touch it separately?

packages/fiori-elements-writer/src/data/defaults.ts
nlunets: interesting !

packages/fiori-elements-writer/test/lrop.test.ts
nlunets: sure, it's in the test code so it's not too bad but i do hope it doesn't hide a real issue :)

nlunets (approved):

packages/fiori-elements-writer/test/lrop.test.ts
tobiasqueck: I don't think so, the code with or without the `Object.assign` would in this case result in the same.

Issue #408, opened 2022-03-24
BUG - utils/locate-reuse-libs.js added without requesting it


tobiasqueck:
### Related Feature
Feature request: #72

### Description
I have created my own generator using the `@sap-ux/fiori-elements-writer` module. When calling the generate method I set the `appOptions` property to `undefined` (and also tried `{}`) but in both case the `webapp/utils/locate-reuse-libs.js` script was generated. The only way to not generate it was explicitly setting the `loadReuseLibs: false`.
My example generator is at https://github.com/SAP/open-ux-tools/blob/example-for-writer/examples/simple-generator/src/app/index.ts.

### Steps to Reproduce
Call the `generate`  function from `@sap-ux/fiori-elements-writer`  with all mandatory values BUT none of the optional values.

### Expected results
The file `webapp/utils/locate-reuse-libs.js` is **not** generated.

### Actual results
The file `webapp/utils/locate-reuse-libs.js` is generated.

### Screenshots
n/a

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

IainSAP:
Fixed #411

PR #411, opened 2022-03-25
fix(fiori-elements-writer):fixes #407(types issues) #408(reuse libs off)

### LINKED ISSUE #407: BUG - the types in the fiori-elements-writer are misleading
### Related Feature
Feature request: #72

### Description
I have tried to create a simple Yeoman generator utilizing the`@sap-ux/fiori-elements-writer` module. While using it some of the types did not completely make sense to me. My example is WIP at https://github.com/SAP/open-ux-tools/blob/example-for-writer/examples/simple-generator/src/app/index.ts

#### EntityConfig
Could you please document the interface `EntityConfig` at https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-elements-writer/src/types.ts#L27? It is not clear what `mainEntity.type` does or `navigationEntity.Role`. For `navigationEntity` I also don't understand what `Name` and `EntitySet` exactly do.

#### Template Settings
I wanted to create an instance of the `LROPSettings` and I noticed the option property `template.settings.entityConfig.filterEntityType` which does not have an effect in an LROP template. This is an OVP specific property. Looking closer, I noticed the vice versa `OVPSettings` offer a `mainEntity` that does not belong to OVP. Could you please revisit https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-elements-writer/src/types.ts#L53-L67 and define types so that they match the real options.

### flpAppId
It is defined as property at https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-elements-writer/src/types.ts#L89 and based on my understanding it is only used for local flp sandbox script. Is there any value for the consumer of the `fiori-elements-writer` to configure this one? I would suggest to remove the property from the public api and set it within the writer to simplify the interface.

### Steps to Reproduce
n/a

### Expected results
I have no questions and know exactly what to do :)

### Actual results
I am confused :(

### Screenshots

OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

IainSAP:
Fixes for:

https://github.com/SAP/open-ux-tools/issues/407 - consolidate types to remove unused and doc
https://github.com/SAP/open-ux-tools/issues/408 - default reuse libs to false

### DIFF
--- a/packages/fiori-elements-writer/test/unit/defaults.test.ts
+++ b/packages/fiori-elements-writer/test/unit/defaults.test.ts
@@ -10,20 +10,26 @@ describe('Defaults', () => {
         const template: Template<ALPSettings> = {
             type: TemplateType.AnalyticalListPage,
             settings: {
-                entityConfig: {}
+                entityConfig: {
+                    mainEntityName: ''
+                }
             }
         };
 
         expect(setDefaultTemplateSettings(cloneDeep(template))).toMatchInlineSnapshot(`
             Object {
-              "entityConfig": Object {},
+              "entityConfig": Object {
+                "mainEntityName": "",
+              },
               "tableType": "AnalyticalTable",
             }
         `);
 
         expect(setDefaultTemplateSettings(cloneDeep(template), OdataVersion.v4)).toMatchInlineSnapshot(`
             Object {
-              "entityConfig": Object {},
+              "entityConfig": Object {
+                "mainEntityName": "",
+              },
               "selectionMode": "None",
               "tableType": "AnalyticalTable",
             }
@@ -32,7 +38,9 @@ describe('Defaults', () => {
         expect(setDefaultTemplateSettings(cloneDeep(template), OdataVersion.v2)).toMatchInlineSnapshot(`
             Object {
               "autoHide": undefined,
-              "entityConfig": Object {},
+              "entityConfig": Object {
+                "mainEntityName": "",
+              },
               "multiSelect": undefined,
               "qualifier": undefined,
               "smartVariantManagement": undefined,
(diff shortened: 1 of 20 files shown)


packages/fiori-elements-writer/test/unit/defaults.test.ts
```diff
@@ -10,20 +10,26 @@ describe('Defaults', () => {
         const template: Template<ALPSettings> = {
             type: TemplateType.AnalyticalListPage,
             settings: {
-                entityConfig: {}
+                entityConfig: {
+                    mainEntityName: ''
```
tobiasqueck, line 14: This test case is interesting because it shows that we accept an empty string as a valid entity name. Should we validate this and throw an error, is this something we can ignore and say it is the responsibility of the caller?

tobiasqueck (approved):
* thanks for removing the default `loadReuseLibs: true`
* separating OVP settings from the rest also looks good
* simplified entity is great
* thanks for the other minor changes improving the code quality (var names and unnecessary function calls)

I have added a minor question inline, otherwise, it looks good.

devinea (approved):
Changes look fine. Haven’t tested locally

packages/fiori-elements-writer/test/unit/defaults.test.ts
IainSAP: Im not sure since a consumer of the API may want to generate an app and fill in that detail later maybe (although a placeholder could be provided in that case). Otherwise we could validate for valid entity names (perhaps there are other disallowed values or characters) or document that we do not explicitly validate this and validation is a caller responsibility, as you say.

PR #416, opened 2022-03-28
Fix version incompatibility in yarn due to logform import

### LINKED ISSUE #405: BUG - Programmatic usage of `ui5-proxy-middleware` does not work
### Related Feature
Feature request: #228

### Description
I have tried using the `@sap-ux/ui5-proxy-middleware` in a CAP project programmatically by extending the `server.js` but it crashes (I assume) when trying to log because the logger is undefined when not running as part of the ui5 tooling. This issue should be independent of CAP and should be reproducible with any standalone `express` instance.

My sample code:
```javascript
const cds = require('@sap/cds');
const ui5Proxy = require('@sap-ux/ui5-proxy-middleware').ui5Proxy;

cds.on('bootstrap', (app) => {
    app.get('/resources/*', ui5Proxy({
        path: [
            '/resources', '/test-resources'
        ],
        url: 'https://ui5.sap.com'
    }));
});

// Delegate bootstrapping to built-in server.js of CDS
module.exports = cds.server
```
### Steps to Reproduce
1. Use any CAP project and add the snippet above to your `server.js`
2. Execute `cds run`
3. Open your browser with http://localhost:4004/resources/sap-ui-core.js
4. See error in the console

### Expected results
The browser returns the content of https://ui5.sap.com/resources/sap-ui-core.js

### Actual results
Console error:
```
[INTERNAL ERROR] TypeError: Cannot read property 'info' of undefined
    at Colorizer.colorize (/xxx/ui5-cap-event-app/node_modules/@sap-ux/logger/node_modules/logform/colorize.js:72:43)
```

### Screenshots
Not required

OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

tobiasqueck:
* fixes #405 
* removed explicit dependency to `logform` (moved it to `devDependencies` for the type import)
* use `logform.format` functionality exposed by `winston`

### DIFF
--- a/packages/logger/package.json
+++ b/packages/logger/package.json
@@ -27,7 +27,6 @@
     "dependencies": {
         "chalk": "4.1.2",
         "lodash": "4.17.21",
-        "logform": "2.3.2",
         "winston": "3.3.3",
         "winston-transport": "4.4.1"
     },
@@ -35,7 +34,8 @@
         "@types/debug": "4.1.7",
         "@types/lodash": "4.14.176",
         "@types/vscode": "1.63.1",
-        "jest-extended": "1.2.0"
+        "jest-extended": "1.2.0",
+        "logform": "2.3.2"
     },
     "files": [
         "dist",
--- a/.changeset/honest-ducks-drive.md
+++ b/.changeset/honest-ducks-drive.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/logger': patch
+---
+
+Fix version incompatibility in yarn due to logform import
--- a/packages/logger/src/winston-logger/adapter.ts
+++ b/packages/logger/src/winston-logger/adapter.ts
@@ -1,12 +1,11 @@
 import type { Transport, TransportOptions } from '../types';
 import { LogLevel } from '../types';
 import type WinstonTransport from 'winston-transport';
-import winston from 'winston';
+import winston, { format } from 'winston';
 import { ConsoleTransport, FileTransport, NullTransport, UI5ToolingTransport, VSCodeTransport } from '../transports';
 import { NullTransport as WinstonNullTransport } from './null-transport';
 import { VSCodeTransport as WinstonVSCodeTransport } from './vscode-output-channel-transport';
 import type { Format } from 'logform';
-import { format } from 'logform';
 import { inspect } from 'util';
 import chalk from 'chalk';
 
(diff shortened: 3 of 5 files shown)


tobiasqueck:
@zdravko-georgiev thanks for your identifying the problem. This fix solves the problem. I have verified this by creating a package locally from the sources and then used your tests snippet that failed with `0.1.4` but works with my local package.
```
const logger = require("@sap-ux/logger");
const myLogger = new logger.ToolsLogger({
  transports: [
    new logger.UI5ToolingTransport({ moduleName: "ui5-proxy-middleware" }),
  ],
});
myLogger.info("Hello");
```

zdravko-georgiev:
> @zdravko-georgiev thanks for identifying the problem. This fix solves the problem. I have verified it by creating a package locally from the sources and then used your tests snippet that failed with `0.1.4` but works with my local package.
> 
> ```
> const logger = require("@sap-ux/logger");
> const myLogger = new logger.ToolsLogger({
>   transports: [
>     new logger.UI5ToolingTransport({ moduleName: "ui5-proxy-middleware" }),
>   ],
> });
> myLogger.info("Hello");
> ```

Thanks for the fix @tobiasqueck!

zdravko-georgiev (approved):
Code ✅ 
Manual test ✅

packages/logger/package.json
```diff
@@ -27,15 +27,15 @@
...
         "winston": "3.3.3",
         "winston-transport": "4.4.1"
     },
     "devDependencies": {
         "@types/debug": "4.1.7",
         "@types/lodash": "4.14.176",
         "@types/vscode": "1.63.1",
-        "jest-extended": "1.2.0"
+        "jest-extended": "1.2.0",
+        "logform": "2.3.2"
```
ullasholla, line 38: Adding a note that this is only required because the type `Format` is not exposed by `winston` - it's used here: 
https://github.com/SAP/open-ux-tools/blob/34beb9f8899fc997a12351dfd0c34d78c31d8776/packages/logger/src/winston-logger/adapter.ts#L8
https://github.com/SAP/open-ux-tools/blob/34beb9f8899fc997a12351dfd0c34d78c31d8776/packages/logger/src/winston-logger/adapter.ts#L46

ullasholla (approved):
Thanks for fixing this @tobiasqueck :)

PR #422, opened 2022-03-29
Fix: adding a custom page to an empty FE FPM application with FCL enabled

tobiasqueck:
- replaced locally defined type for manifest and imported a generated one instead
- added test case for the not yet covered scenario
- fixed the scenario


stefanschreck (approved):

MarcelWaechter (approved):
the generated routes looks good to me... thanks

PR #424, opened 2022-03-30
moved manifest type to reusable module

tobiasqueck:


packages/ui5-proxy-middleware/package.json
```diff
@@ -35,12 +35,11 @@
         "yaml": "2.0.0-10"
     },
     "devDependencies": {
+        "@sap-ux/ui5-config": "workspace:*",
         "@types/express": "4.17.13",
-        "@ui5/server": "2.4.0",
-        "@ui5/project": "2.6.0",
```
zdravko-georgiev, line 40: Since we are removing the dependencies to `@ui5`, we can also remove this one here https://github.com/SAP/open-ux-tools/blob/main/package.json#L51 😃

ullasholla:
@tobiasqueck, don't know if you want reviews yet. The PR is marked as draft

packages/ui5-proxy-middleware/package.json
tobiasqueck: Done

tobiasqueck:
> Guess we can close this as Resolve as won't fix as this file is generated?

Done

tobiasqueck:
> Guess we can close this as Resolve as won't fix as this file is generated?

Changed it to ready and yes, would like to get reviews. There is still one piece missing but the type changes are done.

packages/ui5-proxy-middleware/src/base/utils.ts
```diff
@@ -150,17 +145,13 @@ export const getYamlFile = (args: string[]): string => {
...
 export const getWebAppFolderFromYaml = async (ui5YamlPath: string): Promise<string> => {
-    let webAppFolder = 'webapp';
-
     if (existsSync(ui5YamlPath)) {
-        const ui5Yaml = parseDocument(await promises.readFile(ui5YamlPath, { encoding: 'utf8' })).toJSON();
-
-        if (ui5Yaml.resources?.configuration?.paths?.webapp) {
-            webAppFolder = ui5Yaml.resources.configuration.paths.webapp;
-        }
+        const test = readFileSync(ui5YamlPath, { encoding: 'utf8' });
```
zdravko-georgiev, line 149: Can we have another name for this variable, e.g. `yamlFile`?

packages/ui5-proxy-middleware/src/base/utils.ts
```diff
@@ -150,17 +145,13 @@ export const getYamlFile = (args: string[]): string => {
...
+        const yaml = await UI5Config.newInstance(test);
```
zdravko-georgiev, line 150: `yamlContent` instead of `yaml`?

packages/ui5-proxy-middleware/src/base/utils.ts
zdravko-georgiev, line 149: Why did you switched to the `sync` function?

packages/ui5-proxy-middleware/src/base/utils.ts
```diff
@@ -242,7 +238,7 @@ export async function resolveUI5Version(version?: string, log?: ToolsLogger): Pr
  */
 export const injectUI5Url = async (htmlFilePath: string, ui5Configs: ProxyConfig[]): Promise<string | undefined> => {
     if (existsSync(htmlFilePath)) {
-        let html = await promises.readFile(htmlFilePath, { encoding: 'utf8' });
+        let html = readFileSync(htmlFilePath, { encoding: 'utf8' });
```
zdravko-georgiev, line 241: If we are going to use `readFileSync`, then the function `injectUI5Url` doesn't need to be defined as `async`.

packages/ui5-proxy-middleware/src/base/utils.ts
tobiasqueck: Thank you for the question! 
I tried to find out if there are any known advantages or disadvantages for one or the other approach by searching the internet and by discussing with @nlunets . Conclusion: if you have a use case where you want to read two or more files in parallel, using the `async` version, obviously is the better more flexible option. However, if you read a single file and force synchronization by using `await`, there is no advantage of using a promise. And if two things have no advantage or disadvantage compared to each other, then I always prefer going for the simpler one, which is the `readFileSync`.
Not important but also for consistency, it felt right to use `readFileSync` because we already use `existsSync`.

packages/ui5-proxy-middleware/src/base/utils.ts
tobiasqueck: https://github.com/SAP/open-ux-tools/pull/424/commits/6320c5d2d4d9064ba145e19c12fbbeaf76e7290f

packages/ui5-proxy-middleware/src/base/utils.ts
tobiasqueck: https://github.com/SAP/open-ux-tools/pull/424/commits/229211b2df005d465452fb9dd06715c21a8d32a0

packages/ui5-proxy-middleware/src/base/utils.ts
tobiasqueck: https://github.com/SAP/open-ux-tools/pull/424/commits/6320c5d2d4d9064ba145e19c12fbbeaf76e7290f

zdravko-georgiev: Thank you for the thorough explanation!

donal-tobin-sap:
> @donal-tobin-sap we removed an override that is not required anymore, could you please review

Yes, it is good, I believe it was a build issue and not a security issue in this case.

FYI only now, note that the override information did not match the actual override, as they referred to two different versions.

donal-tobin-sap (approved):
👍🏽 Looks good to me.

tobiasqueck:
> Getting an error when I try to use the ui5-proxy-middleware in an application.
@zdravko-georgiev how did you test? If you created a `tgz` and created a dummy project then it will contain a reference to the old version of the `ui5-config` which didn't have the function yet. It is cumbersome, but to test, you would have to build `ui5-config` and the replace the `node_modules/@sap-ux/config/dist` folder in your dummy project with the content built in `packages/ui5-config/dist`

zdravko-georgiev:
> > Getting an error when I try to use the ui5-proxy-middleware in an application.
> 
> @zdravko-georgiev how did you test? If you created a `tgz` and created a dummy project then it will contain a reference to the old version of the `ui5-config` which didn't have the function yet. It is cumbersome, but to test, you would have to build `ui5-config` and the replace the `node_modules/@sap-ux/config/dist` folder in your dummy project with the content built in `packages/ui5-config/dist`

Yes, this is what I did. The proposed workaround did the trick.

zdravko-georgiev (approved):
looks good 
manual testing `ui5-proxy-middleware` ✅

IainSAP (approved):
Checked that there are no unexpected changes to generated outputs

Issue #399, opened 2022-03-22
FEATURE - Worklist OData V4 support


quiggs1001:
**As a Fiori developer, I'd like to be able to use an OData V4 data source to generate a Worklist Fiori elements application.**

### Description

The worklist floorplan in the generator supports only V2 data sources currently. In order to generate a Worklist V4 application, it will be an LROP V4 application, but with the setting hideFilterBar=true in the manifest.

### Technical Design
_What modules need to be created or enhanced?_ 
_How would the functionality be exposed and how can it be consumed?_

### Acceptance Criteria
- Given I choose the Fiori elements Worklist template, I should be able to provide an OData V4 data source and successfully generate the FE Worklist application.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

IainSAP:
Closing as merged and tests added.

PR #448, opened 2022-04-12
Feat: worklist odata v4 support - 399

unseen1980:
https://github.com/SAP/open-ux-tools/issues/399

The only change needed to support Worklist V4 is the additional property hideFilterBar in manifest.json

1. hideFilterBar added when template type is worklist
2. Tests added
3. Snapshots updated
4. Min ui5 version set to 1.99.0

### DIFF
--- a/packages/fiori-elements-writer/test/unit/validate.test.ts
+++ b/packages/fiori-elements-writer/test/unit/validate.test.ts
@@ -26,11 +26,11 @@ describe('Validate', () => {
         const feApp: FioriElementsApp<LROPSettings> = {
             ...Object.assign(feBaseConfig('felrop1'), {
                 template: {
-                    type: TemplateType.Worklist,
+                    type: TemplateType.FormEntryObjectPage,
                     settings: {}
                 },
                 service: {
-                    version: OdataVersion.v4 // Worklist does not support v4
+                    version: OdataVersion.v2 // Form Entry Object Page does not support v2
                 }
             })
         } as FioriElementsApp<LROPSettings>;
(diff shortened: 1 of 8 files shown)


IainSAP:
@unseen1980 Compent.js base comopnent lib name is missing in generated app. You need to add an entry for Workist v4 here also: 
https://github.com/SAP/open-ux-tools/blob/dd6850a48d8dc15f7d21cbd35503cb5c9a24a47c/packages/fiori-elements-writer/src/data/templateAttributes.ts#L67

Once I make this change locally the app loads.

packages/fiori-elements-writer/test/unit/validate.test.ts
```diff
@@ -26,11 +26,11 @@ describe('Validate', () => {
         const feApp: FioriElementsApp<LROPSettings> = {
             ...Object.assign(feBaseConfig('felrop1'), {
                 template: {
-                    type: TemplateType.Worklist,
+                    type: TemplateType.FormEntryObjectPage,
                     settings: {}
                 },
                 service: {
-                    version: OdataVersion.v4 // Worklist does not support v4
+                    version: OdataVersion.v2 // Form Entry Object Page does not support v2
```
zdravko-georgiev, line 33: This is confusing. You are setting it to `v2` and you are saying it is not supported. Which one is it? 😉

unseen1980: Because worklist now supports v4, I didn't want to remove the test so I changed it into a different negative scenario which is FEOP-v2. It is a negative test. ```'Invalid ODataVersion for template type'```

zdravko-georgiev (approved):

devinea (approved):
Changes look good and inline with expected changes. 
Thanks @unseen1980

unseen1980:
Thanks!

Issue #453, opened 2022-04-21
BUG - wrong default parameter for the ui5-proxy-middleware


zdravko-georgiev:
### Description
The default value of the `secure` option should be `true`, when no option is provided. But in fact it is set to `false` when the `secure` options is missing.

### Expected results
Default value for `secure` is `true`

### Actual results
Default value for `secure` is `false`

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

zdravko-georgiev:
Fixed

PR #454, opened 2022-04-21
fix(ui5-proxy-middleware): set secure:true as default

zdravko-georgiev:
Issue: #453


donal-tobin-sap (dismissed):
👍🏽 Looks good to me.

zdravko-georgiev:
> 👍🏽 Looks good to me.

Donal please re-review. I just saw that my `changeset` was wrong.

donal-tobin-sap (approved):

Klaus-Keller (approved):
Code looks good, setting secure to `true` in case it is not provided is a good idea.
Unit tests exists.

Approved

packages/ui5-proxy-middleware/src/ui5/middleware.ts
```diff
@@ -21,7 +21,7 @@ module.exports = async ({ options }: MiddlewareParameters<Ui5MiddlewareConfig>):
     const router = express.Router();
     const config = options.configuration;
     const ui5Version = await resolveUI5Version(config.version, logger);
-    const secure = !!config.secure;
+    const secure = config.secure !== undefined ? !!config.secure : true;
```
donal-tobin-sap, line 24: If it is null as opposed to undefined then it will not get a default value, is that what you want to do?

zdravko-georgiev: It can be only `null` if the user decides to set it to `null`. So yes, it is intentional.

donal-tobin-sap, line 24:
```suggestion
    const secure = config.secure !== undefined && config.secure !== null ? !!config.secure : true;
```

donal-tobin-sap: 🆗

donal-tobin-sap (approved):

zdravko-georgiev:
Thanks @donal-tobin-sap

Issue #437, opened 2022-04-05
BUG - 


IainSAP:
### Description

The default model setting `groupId` for generated v4 based apps should be:

```
"groupId": "$auto"
```
Currently it disables batch (grouped) requests:
```
"groupId": "$direct"
```

This change has been validated with @SAP/ux-tools-team-elements and details of setting can be found here: https://help.sap.com/saphelp_uiaddon20/helpdata/en/5d/e13cf4dd1f4a3480f7e2eaaee3f5b8/frameset.htm

### Steps to Reproduce
Generate any v4 app.

### Expected results
Check the manifest.json for `mainService`...we should allow batch mode by default.

### Actual results

```
"": {
        "dataSource": "mainService",
        "preload": true,
        "settings": {
          "synchronizationMode": "None",
          "operationMode": "Server",
          "autoExpandSelect": true,
          "earlyRequests": true,
          "groupId": "$direct"
        }
      },
```

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

Issue #392, opened 2022-03-21
TBI - fiori-elements-writer: Add support for v4 services with Overview Page templates


IainSAP:
### Description (include screenshots)
OVPs also work with a v4 service as a base service, and there is no difference in how to configure it for either v2 or v4. This could be supported by allowing v4 services also when generating OVP.

### Value
Adds v4 service support for Overview Page floorplan generation.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

IainSAP:
Will also address https://github.com/SAP/open-ux-tools/issues/437 in same PR.

PR #459, opened 2022-04-25
Feature/fiori elements writer/392/ovp v4

IainSAP:
Adds feature: https://github.com/SAP/open-ux-tools/issues/392
Fixes issue: #437 

- Adds support for Overview Page floorplan generation when an odata v4 service is specified
- Sets the default v4 model property `groupId` to `$auto` : 
> Controls the model's use of batch requests: '$auto' bundles requests from the model in a batch request which is sent automatically before rendering


longieirl (dismissed):

unseen1980 (approved):

tobiasqueck (commented):
I checked https://sapui5.hana.ondemand.com/sdk/#/api/sap.ui.model.odata.v4.ODataModel%23overview and it looks like `groupId: $auto` is the default setting. Wouldn't it be better if the templates don't set it so that the default is used? 

Otherwise, the code changes look good. They adhere to our coding standards and the changes are easy to understand. Especially thanks for adding the additional information about the template structure in the `readme.md`

IainSAP:
> I checked https://sapui5.hana.ondemand.com/sdk/#/api/sap.ui.model.odata.v4.ODataModel%23overview and it looks like `groupId: $auto` is the default setting. Wouldn't it be better if the templates don't set it so that the default is used?
> 
> Otherwise, the code changes look good. They adhere to our coding standards and the changes are easy to understand. Especially thanks for adding the additional information about the template structure in the `readme.md`

Yes I had spotted the default was `$auto` - Im OK to remove the explicit setting in favour of the default fallback if you think that's better. I don't have an opinion either way.

devinea (approved):

PR #469, opened 2022-04-29
Feat/468/lrep service

tobiasqueck:
The PR enhances the `axios-extension` with a service interface for the lrep service allowing to deploy adaptation projects.


tobiasqueck:
@kranthie-sap `changeset` noticed that the example project `examples/odata-cli` is affected by the change in `axios-extension`, however, it is a private module and shouldn't be published. Is the pipeline aware of that?

packages/axios-extension/src/abap/lrep-service.ts
```diff
@@ -0,0 +1,163 @@
...
+type Layer = 'VENDOR' | 'CUSTOMER_BASE';
+
+/**
+ * Returns the namespace as string.
+ *
+ * @param namespace either as string or as object
+ * @returns serialized namespace
+ */
+function getNamespaceAsString(namespace: Namespace): string {
+    return namespace['id'] ? `apps/${namespace['reference']}/appVariants/${namespace['id']}/` : (namespace as string);
```
ullasholla, line 68: Does this work better with an explicit type check? If `string` ever has an `id` property this will silently break

packages/axios-extension/src/abap/lrep-service.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    public log: Logger;
+
+    /**
+     * Check whether a variant with the given namespace already exists.
+     *
+     * @param namespace either as string or as object
+     * @returns the Axios response object for futher processing
+     */
+    public async check(namespace: Namespace): Promise<AxiosResponse> {
```
ullasholla, line 85: Does this method need to be public? If so, does it make sense to wrap the raw axios response?

packages/axios-extension/src/abap/lrep-service.ts
```diff
@@ -0,0 +1,163 @@
...
+            }
+        });
+    }
+
+    /**
+     * Fetch a csrf token required for deployment. The token will be returned as header 'x-csrf-token'.
+     *
+     * @returns the Axios response object for futher processing
+     */
+    public async getCsrfToken(): Promise<AxiosResponse> {
```
ullasholla, line 99: Does this method need to be public?

packages/axios-extension/test/abap/lrep-service.test.ts
```diff
@@ -0,0 +1,95 @@
+import nock from 'nock';
+import fs from 'fs';
+import type { Message } from '../../src/abap/lrep-service';
+import { LayeredRepositoryService, createForAbap, AdaptationConfig } from '../../src';
+
+nock.disableNetConnect();
+
+describe('DesigntimeAdaptationService', () => {
```
ullasholla, line 8: `<nit>`Why call it `DesigntimeAdaptationService` here?`</nit>`

kranthie-sap:
> @kranthie-sap `changeset` noticed that the example project `examples/odata-cli` is affected by the change in `axios-extension`, however, it is a private module and shouldn't be published. Is the pipeline aware of that?

@tobiasqueck: The changeset version script will patch bump `examples/odata-cli` as a dependency was updated. However, the subsequent release workflow will only publish modules that are marked public.

packages/axios-extension/src/abap/lrep-service.ts
```diff
@@ -0,0 +1,163 @@
...
+        };
+        if (config.package) {
+            params['package'] = config.package;
+            if (config.package.toLowerCase() !== '$tmp') {
+                params['changeList'] = config.transport;
+            }
+        }
+        const response = await this.request({
+            method: checkResponse.status === 200 ? 'PUT' : 'POST',
+            url: '/dta_folder/',
```
ullasholla, line 131: Use a constant?

packages/axios-extension/src/abap/lrep-service.ts
```diff
@@ -0,0 +1,163 @@
...
+            data: archive,
+            params,
+            headers: {
+                'Content-Type': 'application/octet-stream',
+                'X-Csrf-Token': tokenResponse.headers['x-csrf-token']
+            }
+        });
+
+        const info = response.data ? JSON.parse(response.data) : {};
```
ullasholla, line 140: `JSON.parse` can throw an error, may need to handle it

ullasholla (changes requested):
Thanks @tobiasqueck, had a few comments. Could you have a look please?

packages/axios-extension/src/abap/lrep-service.ts
tobiasqueck: Does it need to be, no, but the guiding principle for the axios-extension is to offer convenience but still keep it flexible, therefore, I would like to keep it public

packages/axios-extension/src/abap/lrep-service.ts
tobiasqueck: Does it need to be, no, but the guiding principle for the axios-extension is to offer convenience but still keep it flexible, therefore, I would like to keep it public and also return the axios response.

The naming I will change.

packages/axios-extension/test/abap/lrep-service.test.ts
tobiasqueck: Because I changed the name after creating the test cases and forgot to change it here :/ . Will fix it.

packages/axios-extension/src/abap/lrep-service.ts
ullasholla: This service does the CSRF token handling differently to the other services/providers:https://github.com/SAP/open-ux-tools/blob/d9d13e7127d5cea330574622f5d9e1e0dd7bb7fe/packages/axios-extension/src/auth/connection.ts#L137-L140

Is it possible to extend the current implementation?

The intention of asking if we want to expose it is because it makes sense to keep the API surface area as small as possible (specifically I had [Hyrum's law](https://www.hyrumslaw.com/) in mind)

testojs (approved):
Thanks @tobiasqueck! Looks good to me.

packages/axios-extension/src/abap/lrep-service.ts
tobiasqueck: Could I say no? It is only used twice and I doubt it will ever be changed, so I find the code easier readable with the path as plain text.

packages/axios-extension/src/abap/lrep-service.ts
tobiasqueck: Done with https://github.com/SAP/open-ux-tools/pull/469/commits/b8ee2925a0889471a628b6eb9a0444917d1e9e83

packages/axios-extension/test/abap/lrep-service.test.ts
tobiasqueck: Fixed with https://github.com/SAP/open-ux-tools/pull/469/commits/b8ee2925a0889471a628b6eb9a0444917d1e9e83

packages/axios-extension/src/abap/lrep-service.ts
tobiasqueck: You are absolutely correct, the service specific CSRF handling was unnecessary. I have completely removed it. The CSRF token will be fetched with the first `GET` request which we anyhow always send to check if this is a new or existing adaptation project.
When removing it, I also noticed a slightly incorrect implementation in `Connection.ts`. Our code threw an exception for every http code `>=400`, however, that is wrong because the connection can be successful even if a `404` or a `500` is then returned, therefore, I have limited the check to `=401`.

Change is in https://github.com/SAP/open-ux-tools/pull/469/commits/b8ee2925a0889471a628b6eb9a0444917d1e9e83

packages/axios-extension/src/abap/lrep-service.ts
tobiasqueck: Fixed with https://github.com/SAP/open-ux-tools/pull/469/commits/632872d952a710b29d2a0114a841ddd0d88a4f58

packages/axios-extension/src/abap/abap-service-provider.ts
```diff
@@ -137,4 +138,19 @@ export class AbapServiceProvider extends ServiceProvider implements AbapServiceP
         }
         return this.services[AppIndexService.PATH] as AppIndexService;
     }
+
+    /**
+     * Create or get an existing instance of design time adaptation service.
+     *
+     * @returns an instance of the design time adaptation service.
+     */
+    public layeredRepository(): LayeredRepositoryService {
```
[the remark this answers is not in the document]
tobiasqueck: We use `catalog` or `ui5AbapRepository` already, so using `get...` here would be inconsistent. If we want to change it, let's do it in a separate request. Either we convert these methods into property getter and keep the name or add a `get` as prefix to all of them but let's discuss it separately.

packages/axios-extension/src/abap/lrep-service.ts
tobiasqueck: Done with https://github.com/SAP/open-ux-tools/pull/469/commits/3de6c2502ddeaa17de391f8a733d3f542bd116be.

packages/axios-extension/src/abap/lrep-service.ts
ullasholla: Of course, code readability trumps the rule of using constants over literals.

tobiasqueck: Well ... after I said it, I added the `undeploy` function and needed it again, and I couldn't get myself to have the same string three times in the file, so now it is a constant.

ullasholla (approved):
Thanks @tobiasqueck, the changes look good. The code is very readable, the API is clean and the test coverage is good

IainSAP (approved):
Thx @tobiasqueck 

- Clear easy to follow functions and doc
- Tests coverage is good

PR #485, opened 2022-05-13
feat(axios-extension): add support for login with reentrance ticket (Cloud ABAP)

ullasholla:
This PR adds support to connect to Embedded Steampunk systems using [reentrance tickets](https://launchpad.support.sap.com/#/notes/2822156). 

The authentication flow can be summarized as:
1. The user's default browser is opened so that the user can login to the system using SAML 2.0. 
2. After the user is authenticated, the backend system issues a reentrance ticket that is passed back to the code. Reentrance tickets are short-lived (expiry time of 2 minutes) one-time tokens that the backend systems issue. 
3. These tickets are passed back in `MYSAPSSO2` header to fetch cookies that are used to continue the session. 

The user will need to authenticate using the browser _every_ time with this method.

The example CLI code demonstrates how the API is used:
https://github.com/SAP/open-ux-tools/blob/0db5b7f8555a99473a75a2427b6d66dd4be183ab/examples/odata-cli/src/index.ts#L131-L135

The starting point for the code to handle fetch and use the reentrance ticket is here:https://github.com/SAP/open-ux-tools/blob/0db5b7f8555a99473a75a2427b6d66dd4be183ab/packages/axios-extension/src/auth/index.ts#L61-L69

## To test
There is an example module demonstrating how the API can be used: `examples/odata-cli`. The README.md file explains the pre-requisites. Once that is done, at a terminal, change your working directory to `examples/odata-cli` and:
```shell
pnpm build
pnpm start -- cloud
```
You should see a file `v2-catalog.json` created under the `.tmp` folder (re/move it if already exists first).


packages/axios-extension/jest.config.js
```diff
@@ -4,6 +4,14 @@ module.exports = {
...
     collectCoverage: true,
     collectCoverageFrom: ['src/**/*.ts'],
+    globals: {
+        'ts-jest': {
+            tsconfig: './test/tsconfig.json'
+        }
+    },
+    moduleNameMapper: {
+        '^@src/(.*)$': '<rootDir>/src/$1'
+    },
```
tobiasqueck, line 14: Why do we need the changes in the `jest.config.js`?

ullasholla: The change lets me use paths like this:
image

I prefer this over `import Foo from '../../src/foo'` -it's neater and lets me move the test around without breaking the import paths

packages/axios-extension/src/factory.ts
```diff
@@ -85,6 +85,25 @@ export function createForAbapOnBtp(
...
 }
 
+/**
+ * Create an instance of an ABAP service provider for a Cloud ABAP system
+ *
+ * @params options
+ * @param options.url ABAP Service Instance URL
+ * @returns instance of an ABAP service provider
+ */
+export function createForAbapOnCloud({
```
tobiasqueck, line 95: This function is misleading. We already have `createForAbapOnBtp` which is the same as cloud (well, not literally but you know what I mean). So, we should extend the existing function and allow both options, reentrance tickets and oauth

tobiasqueck: Or have to functions but then rename it `createForAbapOnBtpWithReentranceTicket`

packages/axios-extension/test/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+    "extends": "../tsconfig.json",
+    "include": ["../../src/**/*.ts", "**/*.ts"],
+    "exclude": ["dist", "node_modules"],
+    "compilerOptions": {
+        "types": ["jest", "node"],
+        "baseUrl": ".",
+        "paths": { "@src/*": ["../src/*"] }
+    }
+}
```
tobiasqueck, line 10: As one of the earlier comments, how does this file help and would it make sense to do it for all projects?

ullasholla: In this case, it's exclusively used for path mapping. 
The imports look this now:
```typescript
import Foo from @src/foo;
```
instead of:
```typescript
import Foo from '../../src/foo';
```

It would make sense in other projects too, but we should probably do it organically

packages/axios-extension/jest.config.js
tobiasqueck: I agree, it is better than `../..` . Let me also introduce it in my open PR for the new middleware and then we can slowly use it everywhere. Or should we introduce a `jest.config.js` in the repository root and start extending it in each package?

packages/axios-extension/test/abap/abap-service-provider.test.ts
```diff
@@ -11,9 +11,16 @@ import {
 } from '../../src';
 import { ATO_CATALOG_URL_PATH } from '../../src/abap/ato';
 
-nock.disableNetConnect();
-
 describe('AbapServiceProvider', () => {
+    beforeAll(() => {
+        nock.disableNetConnect();
+    });
```
tobiasqueck, line 17: I don't have any objections against this change, just curious, what is the advantage in running it in `beforeAll` compared to outside of the top-most describe? Similar question, do we need `enableNetConnect` in the `afterAll` call again? So far, I assumed that each test file is running separate so it won't affect the next one.

packages/axios-extension/src/auth/index.ts
```diff
@@ -49,3 +51,55 @@ export function attachUaaAuthInterceptor(
...
+            backendUrl: provider.defaults.baseURL,
+            logger: provider.log
+        });
+        if (apiUrl && apiUrl != provider.defaults.baseURL) {
+            // Reentrance tickets work with API hostnames. If the original URL was not one, this will replace it
+            // with the API hostname returned
+            provider.log.warn(
+                `Replacing provider's default base URL (${provider.defaults.baseURL}) with API URL: ${apiUrl}`
+            );
+            provider.defaults.baseURL = apiUrl;
```
tobiasqueck, line 97: This might be a problem if the first call is not executed on the `provider` but on a service created based on the provider because when we create an instance of a service (in `ServiceProvider.createService`) we copy the interceptors over to the service, however, the function here would modify the `baseUrl` of the provider but not affecting the service `baseUrl`.
Thinking about it, this is not a special problem here but more of a conceptual problem in the provider/service concept.

packages/axios-extension/src/factory.ts
```diff
@@ -80,23 +86,53 @@ export function createForAbap(config: AxiosRequestConfig & Partial<ProviderConfi
     return createInstance(AbapServiceProvider, config);
 }
 
+export enum Authentication {
+    OAuth = 'OAuth',
+    ReentranceTicket = 'ReentranceTicket'
+}
+
+export interface AbapEnvironmentBtp {
```
tobiasqueck, line 94: I don't like the interface names and both need a jsdoc explaining their different usages. I would prefer to rename them to describe the technical difference e.g. one using reentrance tickets the other oauth or to rename them based on their target e.g. standalone vs. embedded steampunk

packages/axios-extension/src/factory.ts
```diff
@@ -80,23 +86,53 @@ export function createForAbap(config: AxiosRequestConfig & Partial<ProviderConfi
...
+export type AbapEnvironment = AbapEnvironmentBtp | AbapEnvironmentCloud;
+
 /**
- * Create an instance of an ABAP service provider for an ABAP environment on SAP BTP.
+ * Create an instance of an ABAP service provider for a Cloud ABAP system.
  *
- * @param service ABAP environment service
- * @param refreshToken optional refresh token
- * @param refreshTokenChangedCb option callback for refresh token updates
+ * @param options
```
tobiasqueck, line 111: comment please

packages/axios-extension/src/factory.ts
```diff
@@ -80,23 +86,53 @@ export function createForAbap(config: AxiosRequestConfig & Partial<ProviderConfi
...
+    switch (options.authentication) {
+        case Authentication.OAuth:
+            const { service, refreshToken, refreshTokenChangedCb } = options;
+            provider = createInstance<AbapServiceProvider>(AbapServiceProvider, {
+                baseURL: service.url
+            });
+            attachUaaAuthInterceptor(provider, service, refreshToken, refreshTokenChangedCb);
+            break;
+        case Authentication.ReentranceTicket:
+            const { url, ...config } = options;
```
tobiasqueck, line 125: Where is `config ` coming from? I am lost here.

tobiasqueck: Is this a generic "give me all properties" and then in line 128 you say "spread all properties"? If yes, why wouldn't it be the same for both cases?

ullasholla: `config` is the type `ProviderConfiguration - (AbapEnvironmentCloud | AbapEnvironmentBtp)`.
I didn't use it in both cases as I moved the BTP code here - didn't want to break something that was already existing :)

packages/axios-extension/src/factory.ts
ullasholla: Will do. I prefer to use the target names (standalone, embedded Steampunk, etc), as it'll get confusing when Embedded Steampunk starts supporting OAuth (I expect it to be slightly different to the CF OAuth flow that we have right now).

packages/axios-extension/jest.config.js
devinea: FWIW - IMO `../..` is clearer.

ullasholla: Extending from a root `jest.config.js` sounds good, does it work well? I remember we tried something like this in tools-suite and gave up.

This works in concert with the path mapping in `tsconfig.json`.

packages/axios-extension/test/tsconfig.json
devinea: It can be better for performance to transpile `src` and `test` separately too.
Maybe worth considering as part of https://github.com/SAP/open-ux-tools/issues/495

devinea (approved):
Looks well structured. 👍

packages/axios-extension/src/factory.ts
tobiasqueck: yes

packages/axios-extension/src/factory.ts
tobiasqueck: Looks good

tobiasqueck (commented):
* code adheres to our quality standards (it is very well structure and very easy to understand even with only basic domain knowledge on the topic)
* very good test coverage
* two open points:
   * the issue with the interceptor BUT this is a bigger flaw in the provider/service concept that we need to address separately
   * usage of `@src` as import reference should not be in this PR because I would prefer to be consistent across all modules i.e. if we agree that it is a good idea, then it should be done in a separate PR for all projects

devinea (approved):

tobiasqueck (changes requested):
Could you please enable reentrance tickets in the backend proxy at https://github.com/SAP/open-ux-tools/blob/axios-reentrance-ticket/packages/backend-proxy-middleware/src/base/proxy.ts#L253

packages/axios-extension/src/auth/reentrance-ticket/index.ts
```diff
@@ -0,0 +1,42 @@
+import type { Logger } from '@sap-ux/logger';
+import type { AddressInfo } from 'net';
+import open = require('open');
+import { defaultTimeout } from '../connection';
+import { ABAPSystem } from './abap-system';
+import { setupRedirectHandling } from './redirect';
+
+/** DO NOT USE THIS SERVICE ENDPOINT DIRECTLY.
```
tobiasqueck, line 8: Thanks!

ullasholla:
> Could you please enable reentrance tickets in the backend proxy at https://github.com/SAP/open-ux-tools/blob/axios-reentrance-ticket/packages/backend-proxy-middleware/src/base/proxy.ts#L253

@tobiasqueck, I have made the changes now, could you review again please?

tobiasqueck (approved):

devinea (approved):

Issue #470, opened 2022-05-02
BUG - ui5-proxy-middleware doesn't work with a run configuration created by the SAP Fiori tools


zdravko-georgiev:
### Description
The `ui5-proxy-middleware` doesn't read the env variables `FIORI_TOOLS_UI5_URI` and `FIORI_TOOLS_UI5_VERSION` of a run configuration.

### Expected results
I can change the UI5 version of my application by using a run configuration

### Actual results
The UI5 version defined in a run configuration is not considered when running the application.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #490, opened 2022-05-17
fix(ui5-proxy-middleware): read FIORI_TOOLS_UI5_* env variables

zdravko-georgiev:
Fixes https://github.com/SAP/open-ux-tools/issues/470


tobiasqueck (commented):
Would it be possible to enhance that PR to also enable setting environment variables using `.env` file using `dotenv`?

zdravko-georgiev:
> Would it be possible to enhance that PR to also enable setting environment variables using `.env` file using `dotenv`?

Yes, will do

packages/ui5-proxy-middleware/src/base/utils.ts
```diff
@@ -204,20 +204,21 @@ export async function getUI5VersionFromManifest(args: string[]): Promise<string
  * @returns The UI5 version with which the application will be started
  */
 export async function resolveUI5Version(version?: string, log?: ToolsLogger): Promise<string> {
```
tobiasqueck, line 206: It took me a moment to fully understand what exactly the function assigns when and it seems that some conditions are checked multiple times. If I understood it correctly then we use the version in the following order:
1. env variable
2. input to the function
3. minVersion from manifest

If that is the case, I would suggest to restructure the code like
```
    let ui5Version: string;
    let ui5VersionInfo: string;
    let ui5VersionLocation: string;
	
	if (process.env.FIORI_TOOLS_UI5_VERSION || process.env.FIORI_TOOLS_UI5_VERSION === '') {
		ui5Version = process.env.FIORI_TOOLS_UI5_VERSION;
		ui5VersionLocation = 'CLI arguments / Run configuration';
	} else if (version) {
		ui5Version = version;
		ui5VersionLocation = getYamlFile(process.argv);
	} else {
		const minUI5Version = await getUI5VersionFromManifest(process.argv);
		ui5Version = minUI5Version && isNaN(parseFloat(minUI5Version)) ? minUI5Version : '';
		ui5VersionLocation =  'manifest.json';
	}
    
    if (log) {
        ui5VersionInfo = ui5Version ? ui5Version : 'latest';
        log.info(t('info.ui5VersionSource', { version: ui5VersionInfo, source: ui5VersionLocation }));
    }
    return ui5Version;
```

Less condition checks and the semantic is easier to grasp from the code structure (at least for me).

tobiasqueck (commented):
Please have a look at my suggestion of a code change, otherwise, all looks good. Environment variables can be used now, and they could also be provided using a `.env` file

packages/ui5-proxy-middleware/src/base/utils.ts
zdravko-georgiev: Thanks @tobiasqueck. Done with https://github.com/SAP/open-ux-tools/pull/490/commits/5d35da0b4a9269c4ce60cdb09ae0cb34ae547a56. I did only 2 small adjustments.

tobiasqueck: Thanks (and thanks for fixing my proposal that was ignoring `version = null`)

tobiasqueck (approved):

tobiasqueck (dismissed):

devinea (dismissed):

tobiasqueck (dismissed):

devinea (approved):

tobiasqueck (approved):

PR #492, opened 2022-05-17
fix: update snapshots

zdravko-georgiev:
Fixes failing tests


packages/fiori-elements-writer/test/__snapshots__/lrop.test.ts.snap
```diff
@@ -16955,7 +16955,7 @@ appDescription=A Fiori application.",
     \\"resources\\": \\"resources.json\\",
     \\"sourceTemplate\\": {
       \\"id\\": \\"@sap-ux/fiori-elements-writer:lrop\\",
-      \\"version\\": \\"0.3.2\\",
+      \\"version\\": \\"0.3.3\\",
```
IainSAP, line 16958: This wont work as the next publish will break again.

IainSAP (changes requested):

PR #493, opened 2022-05-17
fix(f-e-w/f-f-w): Fix unstable tests relying on dynamic version

IainSAP:
Recently added snapshots reference a dynamic value (the version of the `@sap-ux/fiori-elements-writer` and `@sap-ux/fiori-freestyle-writer`). This is blocking publish and would continue to break with each version. 

This PR mocks the dynamic values while maintaining the test coverage.


zdravko-georgiev (approved):
LGTM

devinea (approved):

PR #499, opened 2022-05-20
Fix issue in generated route in freestyle template

tobiasqueck:
Fix for #500
Noticed this while trying to use the latest version in easy-ui5

### DIFF
--- a/packages/fiori-freestyle-writer/templates/basic/extend/webapp/manifest.json
+++ b/packages/fiori-freestyle-writer/templates/basic/extend/webapp/manifest.json
@@ -26,7 +26,7 @@
             "routes": [
                 {
                     "name": "Route<%- template.settings.viewName %>",
-                    "pattern": "Route<%- template.settings.viewName %>",
+                    "pattern": ":?query:",
                     "target": [
                         "Target<%- template.settings.viewName %>"
                     ]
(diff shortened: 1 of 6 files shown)


IainSAP (approved):
Thx @tobiasqueck 

-  Test added to cover

packages/fiori-freestyle-writer/templates/basic/extend/webapp/manifest.json
```diff
@@ -26,7 +26,7 @@
             "routes": [
                 {
                     "name": "Route<%- template.settings.viewName %>",
-                    "pattern": "Route<%- template.settings.viewName %>",
+                    "pattern": ":?query:",
```
devinea, line 29:
```suggestion
                    "pattern": "/:?query:",
```
I was a bit confused when I read this first. Since I expect the idea is to set the default route to the app root would this adding the '/' be more obvious?

devinea (approved):
works well. One small query

packages/fiori-freestyle-writer/templates/basic/extend/webapp/manifest.json
tobiasqueck: Hmm, not sure. I wanted to be consistent with what we create in Fiori elements, so I copied it from there.

devinea: OK. Not a big deal.

PR #501, opened 2022-05-20
Making the `odata-service-writer` more flexible

tobiasqueck:
The currently exported `generate` function updates the `manifest.json` as well as the `ui5.yaml` and the`ui5-local.yaml`, lastly it creates a `ui5-mock.yaml`. These yaml files are a given in an (non-CAP) applications generated with the Fiori tools, however, other applications e.g. an application added to the `apps` folder of a CAP project, or an application generated with the `easy-ui5` generator do not have these files, as result, the writer throws an error when used with them.

This PR is making the `odata-service-writer` more flexible by

- find `package.json` and `ui5.yaml` files even if in parent folders
- only update the `ui5.yaml` if it is found
- only create a `ui5-mock.yaml` and update the `package.json` if both `ui5.yaml` and `package.json` exist
- only update `ui5-local.yaml` if it exists 

@devinea @IainSAP @ullasholla what do you think?

Note: I also made the tests a little cleaner BUT neither a semantic change in the tests or an update of the snapshots i.e. the change is backward compatible.
Another note: I later on added code that adds the `fiori-tools-proxy` to the `ui5.yaml` if it is missing, and since the tests used a yaml without it, the snapshots changed a little


packages/odata-service-writer/test/index.test.ts
```diff
@@ -38,47 +37,37 @@ describe('ODataService templates', () => {
...
-        await expect(generate(testDir, validServiceConfig, fs)).rejects.toEqual(
-            Error(t('error.requiredProjectPropertyNotFound', { property: `'sap.app'.id`, path: manifestPath }))
-        );
+    beforeEach(() => {
+        fs = create(createStorage());
     });
 
-    it('generate: valid project with standard valid input', async () => {
-        const testDir = join(outputDir, 'odata-service-v2');
+    async function createTestDir(name: string): Promise<string> {
```
ullasholla, line 44: Does this directory need to be (optionally) cleaned up after a successful test run?

packages/odata-service-writer/src/index.ts
```diff
@@ -11,24 +11,51 @@ import { t } from './i18n';
...
-                throw new Error(t('error.requiredProjectFileNotFound', { path }));
-            }
+export async function findProjectFiles(
+    basePath: string,
+    fs: Editor
+): Promise<{ packageJson?: string; ui5Yaml?: string }> {
+    const files: { packageJson?: string; ui5Yaml?: string } = {};
+    const parts = join(basePath).split('/');
+
+    while (parts.length > 0 && (!files.packageJson || !files.ui5Yaml)) {
```
ullasholla, line 42: Does it make sense to use a library like [`find-up`](https://github.com/sindresorhus/find-up) to do this? The library will take care things like path normalization and handle the various platforms (Windows path idiosyncrasies come to mind!). We make extensive use of `find-up` in tools-suite.

packages/odata-service-writer/src/index.ts
```diff
@@ -11,24 +11,51 @@ import { t } from './i18n';
...
 
 /**
- * Validates the provided base path.
+ * Validates the existence of the given files in the provided base path.
+ *
+ * @param basePath - the root path of an existing UI5 application
+ * @param files - list of files that need to exist
+ * @param fs - the memfs editor instance
+ */
+function validateBasePath(basePath: string, files: string[], fs: Editor) {
```
ullasholla, line 20: How about a more intension-revealing name like `ensureExists` or `validateExistence`? I don't get the `BasePath` in the function name. I know this is a pre-existing function

packages/odata-service-writer/src/index.ts
```diff
@@ -39,61 +66,55 @@ async function generate(basePath: string, service: OdataService, fs?: Editor): P
...
+    let ui5Config: UI5Config | undefined;
+    let ui5LocalConfig: UI5Config | undefined;
+    let ui5LocalConfigPath: string | undefined;
+    if (paths.ui5Yaml) {
+        ui5Config = await UI5Config.newInstance(fs.read(paths.ui5Yaml));
+        try {
+            ui5Config.addBackendToFioriToolsProxydMiddleware(service.previewSettings as ProxyBackend);
+        } catch (error: any) {
+            if (error.message === 'error.nodeNotFound') {
+                ui5Config.addFioriToolsProxydMiddleware({ backend: [service.previewSettings as ProxyBackend] });
```
ullasholla, line 89: Just confirming that this adds the backend config too

ullasholla (changes requested):
Thanks @tobiasqueck, the API is much better as a result of this PR.
Had a few questions inline with the code. Cheers.

packages/odata-service-writer/tsconfig.json
```diff
@@ -7,7 +7,8 @@
         "sourceMap": true,
         "baseUrl": "./src",
         "resolveJsonModule": true,
-        "noImplicitAny": false
+        "noImplicitAny": false,
+        "esModuleInterop": true
```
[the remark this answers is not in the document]
tobiasqueck: I experimented with the latest version of `find-up` when I added it but then remove the lib later, so no, it is not needed. Removed with https://github.com/SAP/open-ux-tools/pull/501/commits/8368cea5f0564209ab3acf75f585f205b66d61e3

packages/odata-service-writer/test/index.test.ts
tobiasqueck: `fs` is a reference to the `mem-fs-editor` here, therefore, nothing is written to the real filesystem. Additionally, `beforeEach` run, we create a new reference to `fs`, so each test start with a clean directory.

packages/odata-service-writer/src/index.ts
tobiasqueck: I tried using it but `find-up` works with the real filesystem while we work with `mem-fs`. I tried using directory callbacks in `find-up` to handle `mem-fs` but it only partially worked for me and got so complicated that doing it on my own seemed simpler (and worked).
If you want to give it a short, please try, I am happy to use `find-up` if we get it running with `mem-fs`.

packages/odata-service-writer/src/index.ts
tobiasqueck: Good point. Changed as requested with https://github.com/SAP/open-ux-tools/pull/501/commits/2e51b2c4884b03171a712be7c011dec8897c87e2

packages/odata-service-writer/src/index.ts
tobiasqueck: Confirmed. `error.nodeNotFound` is thrown if the `fiori-tools-proxy` middleware is not found, so we are adding it including the proxy configuration.

packages/odata-service-writer/src/index.ts
tobiasqueck: PS: now it works on windows as well ;) (https://github.com/SAP/open-ux-tools/pull/501/commits/783997f0bb83e64ca955f4c0159f14463fea4dcc)

devinea (approved):
Approach makes sense. 
Changes are expected. 
Haven't tested locally. 
Thanks

ullasholla (approved):
Thanks @tobiasqueck, the changes look good. All my previous questions have been addressed

Issue #515, opened 2022-06-07
BUG - axios-extension getAtoInfo call returns empty ATO configuration


slin-sap:
### Related Feature
Feature request: _issue number_

### Description

axios-extension getAtoInfo call returns empty ATO configuration

### Steps to Reproduce
1. Add following code in callAFewAbapServices() test function in examples/odata-cli/src/index.ts
```
        const atoSettings = await provider.getAtoInfo();
        console.log(atoSettings);
```
2. Build and run the CLI example project at examples/odata-cli
3. Run command `pnpm start -- abap`

### Expected results

A JS object ATO setting is returned that contains properties such as developmentPrefix, isTransportRequestRequired, transportationMode.  

### Actual results

Empty ATO setting {} is returned.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
Response has status 200. But the response content is actually application level error message in XML format:

```xml
<?xml version="1.0" encoding="utf-8"?>
<exc:exception xmlns:exc="http://www.sap.com/abapxml/types/communicationframework">
<namespace id="com.sap.adt"/>
<type id="ExceptionResourceNotAcceptable"/>
<message lang="EN">The message content is not acceptable. Accepted content types: application/vnd.sap.adt.ato.settings.v1+xml, application/vnd.sap.adt.ato.settings.v1+json</message>
<localizedMessage lang="EN">The message content is not acceptable. Accepted content types: application/vnd.sap.adt.ato.settings.v1+xml, application/vnd.sap.adt.ato.settings.v1+json</localizedMessage>
<properties/>
</exc:exception>
```

### Fix

The http request implemented in getAtoInfo() in packages/axios-extension/src/abap/abap-service-provider.ts is not accepted by backend. It seems to work fine by adding header:

```javascript
const response = await this.get(ATO_CATALOG_URL_PATH, {
    headers: { Accept: 'application/vnd.sap.adt.ato.settings.v1+xml' }
});
```

### Why was it missed
Specific parameter required by ADT service protocol. Test with actual backend is required.

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

tobiasqueck:
@slin-sap is this version specific e.g. will it work in all versions or is this something that cannot be used in e.g. older versions?

slin-sap:
[...]

PR #516, opened 2022-06-07
fix: empty ato settings response

slin-sap:
#515


slin-sap:
Proper fix that formats version specific ATO setting query based on ADT schema will be provided in #518.

Issue #462, opened 2022-04-25
FEATURE - Support custom views in list reports


stefanschreck:
_As a **developer**,
I want to **use the FPM writer to create a custom view in list report**,
so that I **define custom views for the multiple views' scenario**_

### Description

When I compare the capabilities of the FPM writer with the documentation of the FPM explorer I can find a gap: the custom views in list report are not supported, as described [here](https://sapui5untested.int.sap.eu2.hana.ondemand.com/test-resources/sap/fe/core/fpmExplorer/index.html#/customElements/customElementsOverview/customViewInLRContent).

We invested a lot over the last weeks in supporting ALP views and LR multiple views in the Application Modeler of the Fiori Tools suite, and detected the custom views as the last missing feature here. Thus is would be great to add the functionality to create new custom views at the FPM writer, similar to other existing API functions.

### Technical Design
Following the examples at the [FPM explorer ](https://sapui5untested.int.sap.eu2.hana.ondemand.com/test-resources/sap/fe/core/fpmExplorer/index.html#/customElements/customElementsOverview/customViewInLRContent) the new API function `createCustomView` shall 
- create a new fragment, following a template (note: the FPM explorer shows two examples, a table and a panel with a button, but I guess the table is - besides a chart - the most common use case; thus I'd like to propose to present a simple view template, plus an option to have a table as content).
- the new fragment shall be registered in manifest as a new view.
- create an event handler if required (optional parameter); if specified, the API shall also create the extends.extensions entry in manifest for this controller.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I want to add a new custom view to a list of multiple views at the FE list report
**when** I call the new API
**then** a view and (optionally) a controller gets created and the new view is added to the views section at the manifest; in a preview can see the additional view.

### Tasks
- [x] Create view
- [x] Create controller
- [x] Update views in manifest
- [x] Update extension in manifest
- [x] Unit Tests added
- [x] Test Cases defined

tobiasqueck:
@stefanschreck could you please rework the acceptance criteria and be more specific and what use case we need to cover?

tobiasqueck:
Are there other places e.g. in an object page where we could add a custom view? Would it make sense to design the API more flexible?

stefanschreck:
> Are there other places e.g. in an object page where we could add a custom view? Would it make sense to design the API more flexible?
@tobiasqueck 
[...]

PR #517, opened 2022-06-08
feat(fe-fpm-writer): support custom views

johannes-kolbe:
#462: Add support for custom views

- optional parameter: `control`, indicating whether a sample table xml should be generated into the fragment.
- extends manifest `extends/extension` in case of eventHandler
- extends views > paths

#### Tasks
- [x] Create view
- [x] Create controller
- [x] Update views in manifest
- [x] Update extension in manifest
- [x] Unit Tests added
- [x] Test Cases defined


packages/fe-fpm-writer/src/view/index.ts
```diff
@@ -0,0 +1,78 @@
...
+import { create } from 'mem-fs-editor';
+import type { CustomView, InternalCustomView } from './types';
+import { join } from 'path';
+import { render } from 'ejs';
+import { validateVersion, validateBasePath } from '../common/validate';
+import type { Manifest } from '../common/types';
+import { setCommonDefaults, getDefaultFragmentContent } from '../common/defaults';
+
+/**
+ * Enhances the provided custom view configuration with additional data.
```
stefanschreck, line 12: ...with default data
(it's called like this with the other extensions and I think this fits better)

packages/fe-fpm-writer/src/view/types.ts
```diff
@@ -0,0 +1,33 @@
...
+     */
+    label: string;
+
+    /**
+     * Unique tab identifier.
+     */
+    key: string;
+
+    /**
+     * If not set (i.e. undefined) then no event handler is linked. If it is set true, a new one is created and linked to the action.
```
stefanschreck, line 20: ...set to true, a new controller is created ...

packages/fe-fpm-writer/templates/view/ext/CustomViewWithTable.xml
```diff
@@ -0,0 +1,29 @@
...
+        </infoToolbar>
+        <columns>
+            <Column>
+                <Text text="Column" />
+            </Column>
+        </columns>
+        <items>
+            <ColumnListItem>
+                <cells>
+                    <%- content %>
```
stefanschreck, line 24: That's too complex, I would say: no-one would know how to correctly supply `content`. 
It's easier to maintain it afterwards in the XML, thus I would neglect this parameter.

packages/fe-fpm-writer/templates/view/manifest.json
```diff
@@ -0,0 +1,34 @@
...
+            }
+        },
+        <% } %>
+        "routing": {
+            "targets": {
+                "<%- target %>": {
+                    "options": {
+                        "settings": {
+                            "views": {
+                                "paths": [
```
stefanschreck, line 20: You cannot assume that there is no view defined yet. If there would be any view definition already, you would destroy it.

packages/fe-fpm-writer/test/unit/view.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    });
+
+    test('with new handler', () => {
+        //sut
+        generateCustomView(testDir, { ...customView, key: 'customViewKey', eventHandler: true }, fs);
+        const updatedManifest: any = fs.readJSON(join(testDir, 'webapp/manifest.json'));
+        const extension = updatedManifest['sap.ui5']['extends']['extensions'];
+        const views = updatedManifest['sap.ui5']['routing']['targets']['sample']['options']['settings']['views'];
+        expect(extension).toMatchSnapshot();
+        expect(views).toMatchSnapshot();
```
stefanschreck, line 44: That's the problem of snapshot testing, that you never look at the content...

packages/fe-fpm-writer/test/integration/__snapshots__/index.test.ts.snap
```diff
@@ -420,6 +441,15 @@ Object {
                     }
                   }
                 }
+              },
+              \\"views\\": {
```
stefanschreck, line 445: as expected, the snapshot only shows an example with exactly one view entry. Adding a custom view to a list of existing views is not supported.

stefanschreck (changes requested):

packages/fe-fpm-writer/src/view/types.ts
```diff
@@ -0,0 +1,33 @@
...
+     * If an existing event handler is to be used then its id needs to be provided as string.
+     */
+    eventHandler?: string | true;
+
+    /**
+     * If set, a sample table control XML will be generated into the fragment of the view.
+     */
+    tableControl?: true;
```
tobiasqueck, line 28: That one feels wrong. Shouldn't we allow providing an optional XML string as content. If the string is not provided, then we generate example XML.

stefanschreck:
@johannes-kolbe Would you please also update the task list of the requirement?

packages/fe-fpm-writer/templates/view/ext/CustomViewWithTable.xml
johannes-kolbe: Ok, thanks, will remove `content`.
If a user wants to generate a new eventHandler and a sample table, would it make sense to add a `Button` instead of `content` here to indicate/link that?

tobiasqueck: Would it be possible to reuse this common functionality https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/common/defaults.ts#L41

packages/fe-fpm-writer/src/view/index.ts
johannes-kolbe: renamed to `default data`

packages/fe-fpm-writer/src/view/types.ts
johannes-kolbe: Thanks, applied suggestion.

packages/fe-fpm-writer/templates/view/ext/CustomViewWithTable.xml
johannes-kolbe: @tobiasqueck - thanks, yes would like to reuse it! 
Current approach: 
If `control` is custom pre-filled, a new fragment is generated. If set to `true`, then table fragment is created with `<%- content %>` filled by either eventHandler, or Text. 
Would that work, or is another solution needed?

packages/fe-fpm-writer/test/unit/view.test.ts
johannes-kolbe: Still snapshots, but sympotom should be addressed now. If approach is used and preferred, I could add direct tests.

packages/fe-fpm-writer/test/integration/__snapshots__/index.test.ts.snap
johannes-kolbe: Done

packages/fe-fpm-writer/src/view/types.ts
johannes-kolbe: Renamed to `control?: string | true` - with updated logic.

tobiasqueck:
> Updated task list, testcases are unit tests?

No, test cases could be additional (manual) tests. For the FPM writer we have this one additional integration test (also with jest) that calls multiple generate functions on the same base project and there we check that it also works together. You did this, so you can check the box.

packages/fe-fpm-writer/test/unit/view.test.ts
tobiasqueck: What is the problem of snapshot testing? I don't understand the problem here.

packages/fe-fpm-writer/templates/view/ext/CustomViewWithTable.xml
tobiasqueck: I always prefer keeping functionality consistent. For columns we either use the provided control, or generate sample code if `control === undefined`. We don't have the additional `true`. https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/column/index.ts#L46-L50

What would happen in your proposal if control is `undefined`? The fragment would be empty, correct? This is, of course, not wrong, however, I wonder if simple generated content is the better option because as the developer I can see in the preview that the generation worked.

stefanschreck:
> > @johannes-kolbe Would you please also update the task list of the requirement?
> 
> Updated task list, testcases are unit tests?

I meant the task lsist at https://github.com/SAP/open-ux-tools/issues/462, now we have it twice.

packages/fe-fpm-writer/test/unit/view.test.ts
stefanschreck: @tobiasqueck Not especially here, but it is my general concern - but we hade this discussion already some times back.

packages/fe-fpm-writer/test/unit/view.test.ts
```diff
@@ -26,54 +41,113 @@ describe('CustomView', () => {
...
+        const { extension, views } = getManifestSegments();
+
         expect(extension).toMatchSnapshot();
         expect(views).toMatchSnapshot();
-
         expect(fs.read(expectedFragmentPath)).toMatchSnapshot();
         expect(fs.read(expectedFragmentPath.replace('.fragment.xml', '.js'))).toMatchSnapshot();
     });
+
+    test.only('with existing views', () => {
```
stefanschreck, line 124: .only must be removed

stefanschreck (changes requested):

packages/fe-fpm-writer/test/integration/__snapshots__/index.test.ts.snap
stefanschreck: ok

packages/fe-fpm-writer/templates/view/manifest.json
stefanschreck: ok

johannes-kolbe:
> > > @johannes-kolbe Would you please also update the task list of the requirement?
> > 
> > 
> > Updated task list, testcases are unit tests?
> 
> I meant the task lsist at #462, now we have it twice.

Ok, I couldn't edit it, that's why it was duplicated here.

packages/fe-fpm-writer/src/view/index.ts
```diff
@@ -0,0 +1,89 @@
...
+    // existing views
+    const existingViews = (manifest['sap.ui5']?.routing?.targets?.[data.target] as Ui5RoutingTarget<Ui5TargetSettings>)
+        .options?.settings?.views;
+    if (existingViews) {
+        config.views = existingViews;
+    }
+
+    // generate view content
+    if (typeof config.control === 'string') {
+        config.content = config.control;
```
tobiasqueck, line 37: I don't think you need that line ;)

packages/fe-fpm-writer/templates/view/ext/CustomViewWithTable.xml
tobiasqueck: Is there an advantage in creating an example table as content compared to using the default fragment content i.e. the button?
I tend to say no because the developer creating a custom view and wants to create a table, has already something in mind, and won't need this example ( in addition, if we want to add a sample table, we should create an example with the table macro). Thus, I would recommend doing it the same as we do for columns, if a control is provided, use it as content, otherwise use the default fragment content. 
It would be consistent with our other generate functions and provides enough of an example to see in the preview that the generation was successful.

stefanschreck (dismissed):

packages/fe-fpm-writer/templates/view/ext/CustomViewWithTable.xml
stefanschreck: @tobiasqueck 
I agree to the table as default content. The button is just an example in the FPM explorer, but not a real use case. On different view tabs of list report you usually show different lists. Charts in some cases.

johannes-kolbe: Updated generated xml fragment: https://github.com/SAP/open-ux-tools/pull/517/commits/f2e41eb0507915b08105ad35c044e3e4033d78b4#diff-372dd6c273bb1b939f13bfec4a5bc344527da25410d5fedab2f1f00e3fb8ee15R2

tobiasqueck (changes requested):
The manual test fails for me. Please execute `UX_DEBUG=true pnpm test` and then go to the folder `./test/test-output/lrop` and run `npm i && npm start` there. 
The generated app is not loading. It looks like there are 404s for your controller/view files.

tobiasqueck:
@johannes-kolbe it still doesn't work for me. The app now renders the filter bar but I don't see the table or any other view. 

Btw, when you look at it, could you do me a favor and also change the `test-input/package.json` and replace the existing `devDependencies` section with

```
"devDependencies": {
        "@ui5/cli": "2",
        "@sap/ux-specification": "latest",
        "@sap/ux-ui5-tooling": "1",
        "@sap/ux-ui5-fe-mockserver-middleware": "1",
        "rimraf": "3.0.2"
    }
```

Please also run `pnpm test -- -u` to update the test snapshots. It looks like there is an obsolete one now.

johannes-kolbe:
Closing this PR in favour of [#525](https://github.com/SAP/open-ux-tools/pull/525), resolving Issue with fork and sonarcloud.

Issue #518, opened 2022-06-08
TBI - Use ADT service to fetch transport numbers for ABAP deployment configuration


slin-sap:
### Description (include screenshots)
Link to tools-suite internal issue: #19643

During deploy configuration, user needs to provide transport number for the project to be deployed. It is more convenient to allow user to select an existing transport number from a dropdown list instead of manually typing the transport number. This TBI aims to provide the api function that can be called to fetch transport number list from ADT service provided by backend ABAP system.

#Architecture elaboration

The new API function will be implemented in axios-extension module. A service provider should provide API to call backend to fetch the transport number list. ADT service request can be specific to ABAP versions. Need to call discovery service first and parse the service schema.

tobiasqueck:
@slin-sap what is the state of this one? Anything missing or has the feature been implemented as requested?

PR #519, opened 2022-06-09
fix: adt services for abap service provider

slin-sap:
#518

- Provide adt-catalog-service for methods that implements ADT services requests.
- Implements ADT discovery schema to use request url matches the backend system ADT service version
- Implemented 2 ADT services: ATO settings and transport checks.
- Unit tests


packages/axios-extension/src/abap/types/adtTypes.ts
```diff
@@ -0,0 +1,30 @@
+export interface AdtSchemaData {
```
tobiasqueck, line 1: Minor thing: we used `types.ts` as name in other places when defining types and interfaces

tobiasqueck (commented):
It took me a quite a bit to understand exactly what you are doing (haven't used decorators in a while) but I like the concept. Could you please add jsdoc comments to each function outlining what it does, so that next time when looking at the code, it is easier to understand.

packages/axios-extension/src/abap/interface.ts
```diff
@@ -0,0 +1,12 @@
+import type { ODataVersion } from '../base/odata-service';
```
ullasholla, line 1: Can we think of another name for the file? `interface` seems too generic

packages/axios-extension/src/abap/types/adt-types.ts
```diff
@@ -0,0 +1,30 @@
+export interface AdtSchemaData {
```
ullasholla, line 1: Can we have some comments (JS Doc or otherwise) about these types please? I'm mainly focussed on what they're useful for

packages/axios-extension/src/abap/adt/handlers/transport.ts
```diff
@@ -0,0 +1,97 @@
+import XmlParser from 'fast-xml-parser';
+import * as xpath from 'xpath';
+import { DOMParser } from '@xmldom/xmldom';
+
+export function parseTransportRequests(xml: string): string[] {
+    if (XmlParser.validate(xml) !== true) {
+        return [];
```
ullasholla, line 7: Log something here? That probably means that the logger needs to be passed in somehow

packages/axios-extension/src/abap/adt/handlers/transport.ts
```diff
@@ -0,0 +1,97 @@
...
+ * @param doc
+ * @returns available transport numbers
+ */
+function getTransportChecksResponse(doc: Document): string[] {
+    const status = xpath.select1('//RESULT/text()', doc)?.toString();
+
+    switch (status) {
+        case AdtTransportStatus.Success:
+            return getTransportList(doc);
+        case AdtTransportStatus.Error:
```
ullasholla, line 33: Even if we don't handle errors now, we should probably log them

packages/axios-extension/src/abap/adt/handlers/transport.ts
```diff
@@ -0,0 +1,97 @@
+import XmlParser from 'fast-xml-parser';
+import * as xpath from 'xpath';
+import { DOMParser } from '@xmldom/xmldom';
+
+export function parseTransportRequests(xml: string): string[] {
+    if (XmlParser.validate(xml) !== true) {
+        return [];
+    }
+    const doc = new DOMParser().parseFromString(xml);
```
ullasholla, line 9: Can this call still fail? Does it need any error handling?

packages/axios-extension/src/abap/adt/handlers/transport.ts
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * @param doc
+ * @returns
+ * - For local package, an array list that contain a single empty string is returned.
+ * - For errors or other unkonwn reasons no transport number found, an empty array list is returned.
+ */
+function getTransportList(doc: Document): string[] {
+    const recording = xpath.select1('//RECORDING/text()', doc)?.toString();
+    const locked = (xpath.select1('//LOCKS', doc) as Element)?.textContent;
+    const localPackage = xpath.select1('//DLVUNIT/text()', doc)?.toString();
```
ullasholla, line 51: Are these paths part of some schema? Do we need to consider versions if such a schema is available? Can we log anything as an early warning if the schema changes underfoot?

packages/axios-extension/src/abap/adt/discovery.ts
```diff
@@ -0,0 +1,106 @@
+import type { AbapServiceProviderExtension } from '../abap-service-provider-extension';
+import XmlParser from 'fast-xml-parser';
+import type { AdtCategory, AdtSchemaData } from '../types';
+import type { ServiceProvider } from 'index';
+import 'reflect-metadata';
+
+const ADT_DISCOVERY_SERVICE = '/sap/bc/adt/discovery';
```
ullasholla, line 7: Does this service have a stable public API?

slin-sap: Yes

packages/axios-extension/src/abap/adt/handlers/transport.ts
slin-sap: Now you see the axios extension doesn't have looger passed in :P

packages/axios-extension/src/abap/adt/handlers/transport.ts
```diff
@@ -0,0 +1,97 @@
...
+    return list;
+}
+
+/**
+ * This function processes ADT response for existing project name that has been locked.
+ * A single, previously provided transport number is returned in the list.
+ * @param doc
+ * @returns
+ */
+function getLockedTransport(doc: Document): string[] {
```
ullasholla, line 90: Why does this return an array and not a scalar value?

packages/axios-extension/src/abap/adt/handlers/transport.ts
```diff
@@ -0,0 +1,97 @@
+import XmlParser from 'fast-xml-parser';
+import * as xpath from 'xpath';
+import { DOMParser } from '@xmldom/xmldom';
+
+export function parseTransportRequests(xml: string): string[] {
```
ullasholla, line 5: Find a better name than `parseTransportRequests`? Parsing technically means to break something down into components. What are we doing exactly? Extracting transport request numbers from the response? Some JS Doc please?

packages/axios-extension/src/abap/adt/handlers/transport.ts
slin-sap: Based on the WebIDE implementation, there is no need to handle schema changes at the response data level.

packages/axios-extension/src/abap/abap-service-provider.ts
```diff
@@ -153,4 +170,52 @@ export class AbapServiceProvider extends ServiceProvider implements AbapServiceP
...
     }
+
+    /**
+     *
+     * @param packageName Package name for deployment
+     * @param appName Fiori project name for deployment. A new project that has not been deployed before is also allowed
+     * @param schema Decorated para that will be autofilled
+     * @returns
+     */
+    @adt(AdtServiceConfigs[AdtServiceName.TransportChecks])
```
ullasholla, line 181: I don't understand the `@adt` decorator abstraction. Can't we use an `ADTServiceCatalog` or something that takes care of calling discovery and loading the schema as required? This feels too implicit/magic to me. Feels like the name `@adt` could be more descriptive too. Something along the lines of `@usesAdtService()`. Even then I don't see the point of exposing the concept of the ADT Schema to endusers. From what I can see users will try and access ADT services, discovery and schema loading are incidental to this purpose. It's also easy to miss using the decorator. If a catalog object were used instead, users would not miss dealing with an instance of it

ullasholla (changes requested):
Thanks @slin-sap. I had a few questions. Also, do you mind adding in a short description of what this PR exactly does?

packages/axios-extension/src/abap/adt/discovery.ts
ullasholla: Can you also add a link to the public documentation then please?

packages/axios-extension/src/abap/adt/handlers/transport.ts
ullasholla: `ServiceProvider` has an instance of logger: https://github.com/SAP/open-ux-tools/blob/9d4e3bcfd4a1d45e93ba603ecf345e34896cedf8/packages/axios-extension/src/base/service-provider.ts#L25

slin-sap: I added this one. It doesn't look like it instantiate from YUI logger tho.

tobiasqueck: > I added this one. It doesn't look like it instantiate from YUI logger 

Not sure I understand but do we need to make the `logger` module more flexible or what is missing?

slin-sap: For generators used in YUI, the vscode logger is passed into the generator as option. The log messages can be found on the output tab in vscode. Regular logs are not accessible on the console.

packages/axios-extension/src/abap/adt/handlers/transport.ts
slin-sap: The function name has been updated. Not sure why it is not updated here. Added JS doc

packages/axios-extension/src/abap/adt/README.md
```diff
@@ -0,0 +1,104 @@
...
+Before requesting an ADT service, ADT service schema should be checked to find
+out the ADT service availability and service contract
+(e.g. available response content types supported by ABAP backend).
+
+We use two decorators to minimize this repeated effort to check ADT schema in the code. Please
+see the Usage section below for details.
+
+### Usage
+
+ADT decorator is simple to use. Currently it can be used for methods that
```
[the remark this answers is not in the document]
slin-sap: The latest dev branch is updated: https://github.com/SAP/open-ux-tools/blob/fix/518/adt-services/packages/axios-extension/src/abap/adt/README.md But not reflected here.

packages/axios-extension/src/abap/abap-service-provider-extension.ts
```diff
@@ -0,0 +1,16 @@
...
+import type { AdtCollection } from './types';
+import type { Ui5AbapRepositoryService } from './ui5-abap-repository-service';
+
+export interface AbapServiceProviderExtension {
+    s4Cloud: boolean | undefined;
+    user(): Promise<string>;
+    catalog(oDataVersion: ODataVersion): CatalogService;
+    ui5AbapRepository(): Ui5AbapRepositoryService;
+    getSchemaStore(): AdtSchemaStore;
+    getAtoInfo(schema?: AdtCollection): Promise<AtoSettings>;
```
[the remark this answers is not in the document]
kranthie-sap: These may have been inherited from before, feel free to update or leave them for now.

packages/axios-extension/test/abap/abap-service-provider.test.ts
```diff
@@ -76,10 +76,10 @@ describe('AbapServiceProvider', () => {
     });
 
     // Discovery schema is cached, so separate this test suite from other ADT service tests
-    describe('Service unavailable in discovery', () => {
+    describe.only('ADT Services unavailable in discovery', () => {
```
ullasholla:
```suggestion
    describe('ADT Services unavailable in discovery', () => {
```

slin-sap: @ullasholla already fixed. Noticed it when the sonar qgate dropped.

packages/axios-extension/src/abap/abap-service-provider.ts
```diff
@@ -49,9 +44,24 @@ export class AbapServiceProvider extends ServiceProvider implements AbapServiceP
...
+
         if (!this.atoSettings) {
             try {
-                const response = await this.get(ATO_CATALOG_URL_PATH);
+                const acceptHeaders = {
+                    headers: {
+                        Accept: 'application/*'
+                    }
+                };
+                const response = await this.get(serviceSchema.href, acceptHeaders);
```
ullasholla, line 64: What's supposed to happen if `serviceSchema.href` is empty or `undefined`?

slin-sap: @ullasholla  If serviceSchema is not empty, href should always be available.

packages/axios-extension/src/abap/abap-service-provider.ts
```diff
@@ -153,4 +175,48 @@ export class AbapServiceProvider extends ServiceProvider implements AbapServiceP
...
+     */
+    public async getTransportRequests(packageName: string, appName: string): Promise<string[]> {
+        const serviceSchema = await this.getAdtCatalogService().getServiceDefinition(
+            AdtServiceConfigs[AdtServiceName.TransportChecks]
+        );
+        // Service not available on target ABAP backend version, return empty setting config
+        if (!serviceSchema) {
+            return [];
+        }
+        const urlPath = serviceSchema.href;
```
ullasholla, line 193: How about the case when `serviceSchema.href` is empty or `undefined`?

slin-sap: We can assume href always presents if serviceSchema is not empty.

packages/axios-extension/src/abap/adt-catalog/README.md
```diff
@@ -0,0 +1,60 @@
...
+ADT provides a discovery service that responds with ADT service schema, 
+which contains availability for ADT services.
+
+
+### ADT Service Implementation
+
+A lazy loading strategy is used for fetching Adt discovery schema. I.e. The discovery request is sent
+when the first time an ADT service request is triggered. 
+To implement ADT services with this lazy loading, developer needs to make sure to call the getServiceDefinition() function in 
+AdtCatalogService to fetch the service schema for the ADT service to be implemented. E.g.
```
ullasholla, line 13:
```suggestion
The ADT service schema is lazily loaded. The first time an ADT service request is triggered, a schema discovery request is made first.
To use an ADT service, make sure to call `getServiceDefinition()` in 
`AdtCatalogService` first to make sure the service schema is initialized correctly. E.g.
```

ullasholla:
An ADT service may need its own abstraction like we have one for OData services: https://github.com/SAP/open-ux-tools/blob/main/packages/axios-extension/src/base/odata-service.ts

We can refactor the current implementation into something like that, doesn't have to done in the current PR.

@tobiasqueck , @kranthie-sap, @slin-sap, thoughts?

packages/axios-extension/src/abap/abap-service-provider.ts
ullasholla: We're calling an external service and we have no control over its behavior (or bugs or misconfiguration). Calling `get()` later with an undefined/empty path may result in unpredictable and hard to debug behavior.

packages/axios-extension/src/abap/abap-service-provider.ts
ullasholla: Can we please code defensively and handle this?

.changeset/tender-kings-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/axios-extension': minor
+---
+
+ADT service support in abap service prodiver
```
ullasholla, line 5:
```suggestion
ADT service support in abap service provider
```

packages/axios-extension/src/abap/abap-service-provider.ts
ullasholla: Thanks @slin-sap. Looks good. One of the last things to do: can we please log an error/warning when we're returning nothing? This would be a case when there's a bug/misconfiguration in the backend. Logging would expedite trouble-shooting

packages/axios-extension/src/abap/abap-service-provider.ts
ullasholla: Thanks @slin-sap. Looks good. One of the last things to do: can we please log an error/warning when we're returning nothing? This would be a case when there's a bug/misconfiguration in the backend. Logging would expedite trouble-shooting

ullasholla (dismissed):
Thanks @slin-sap, looks good. All my earlier questions have been addressed

packages/axios-extension/src/abap/abap-service-provider-extension.ts
```diff
@@ -0,0 +1,16 @@
...
+import type { AdtSchemaStore } from './adt/adt-schema-store';
+import type { CatalogService } from './catalog';
+import type { AdtCollection } from './types';
+import type { Ui5AbapRepositoryService } from './ui5-abap-repository-service';
+
+export interface AbapServiceProviderExtension {
+    s4Cloud: boolean | undefined;
+    user(): Promise<string>;
+    catalog(oDataVersion: ODataVersion): CatalogService;
+    ui5AbapRepository(): Ui5AbapRepositoryService;
```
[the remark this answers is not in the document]
kranthie-sap: Since these are not changed in this PR, it can be cleaned up in another PR.

kranthie-sap (approved):

ullasholla (approved):

ullasholla (approved):

PR #531, opened 2022-06-16
feat(fe-fpm-writer): support 'requireSelection' property for custom actions

815are:
While working on integration of custom actions.
Noticed that 'requireSelection' property is not supported for custom actions.
In scope of this PR adding support for 'requireSelection'.

I am setting 'requireSelection' as optional 'boolean' only - because as tested it does not support binding definition.


tobiasqueck (approved):
* changes are easy to understand
* tests are properly enhanced

Issue #534, opened 2022-06-16
FEATURE - Support creation of ListReport and ObjectPage for FPM applications


tobiasqueck:
_As a **developer**,
I want to **have tools that support adding ListReportd and ObjectPages to FPM apps**,
so that I **can build fully flexible annotation driven applications**_

### Description
TBD
_What is the high level purpose and value of the feature?  The description should ensure that anyone reading understands what it is doing (at a high level) and most importantly why._

### Technical Design
_What modules need to be created or enhanced?_ 
_How would the functionality be exposed and how can it be consumed?_

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a ...
**when** I execute ...
**then** ... will happen.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #535, opened 2022-06-16
Add `generateObjectPage|ListReport` to `fe-fpm-writer`

tobiasqueck:
Implementation for the first part of #534

- [x] refactor generateCustomPage to be able to reuse parts for the new feature
- [x] generateObjectPage implemented
- [x] generateListReport implemented
- [x] unit tests created for generateObjectPage
- [x] unit tests created for generateListReport
- [x] integration test enhanced
- [x] changeset generated


packages/fe-fpm-writer/src/page/common.ts
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * @param routes existing application routes (from the manifest)
+ * @param config configuration object
+ */
+export function updateRoutes(
+    routes: ManifestNamespace.Route[],
+    config: InternalCustomPage | InternalObjectPage | InternalListReport
+) {
+    const newRoute: ManifestNamespace.Route = {
+        name: `${config.entity}${config.name}`
```
stefanschreck, line 28: what if the new route already exists?

packages/fe-fpm-writer/src/page/index.ts
```diff
@@ -1,146 +1,16 @@
...
-
-/**
- * Add a custom page to an existing UI5 application.
- *
- * @param {string} basePath - the base path
- * @param {CustomPage} data - the custom page configuration
- * @param {Editor} [fs] - the memfs editor instance
- * @returns {Promise<Editor>} the updated memfs editor instance
- */
-export function generateCustomPage(basePath: string, data: CustomPage, fs?: Editor): Editor {
```
stefanschreck, line 111: as this function gets deleted, it is an incompatibility that needs to be considered when consuming the new open-ux-tools version at tools-suite - coding needs to be adapted

packages/fe-fpm-writer/test/integration/index.test.ts
```diff
@@ -69,10 +91,33 @@ describe('use FPM with existing apps', () => {
         });
 
         test('generateCustomView in ListReport', () => {
+            //pre-requisite is at least one view based on annotations
+            fs.extendJSON(join(targetPath, 'webapp/manifest.json'), {
```
stefanschreck, line 95: why is this extension needed? the test should also run fine when no view exists yet (but we need to test both scenarios)

packages/fe-fpm-writer/src/page/types.ts
```diff
@@ -45,19 +106,24 @@ export interface CustomPage extends CustomElement {
...
+    /**
+     * The FCL flag is calculated to true if the existing app is configured to use the flexible column layout
+     */
+    fcl?: boolean;
+
+    /**
+     * Represents the name of the control aggregation to be used for the custom page.
+     * This is only used if FCL is enabled and is calculated based on the optional sourcepage.
+     */
+    controlAggregation?: 'beginColumnPages' | 'midColumnPages' | 'endColumnPages';
```
stefanschreck, line 119: Looks like FCL is not. covered by any (snapshot) test

packages/fe-fpm-writer/src/page/index.ts
tobiasqueck: The function didn't get deleted. It moved to the `custom.ts` file and is imported and exported here again. So, everything is backward compatible.

packages/fe-fpm-writer/test/integration/index.test.ts
tobiasqueck: No, it does not. You cannot add a custom view to a ListReport with at least one annotation based view.

tobiasqueck: We should consider adding this default view in such case but that would be a different topic and should be discussed with @johannes-kolbe as well

packages/fe-fpm-writer/src/page/index.ts
stefanschreck: Yes, this was clear, but the caller needs to adapt, as the function was moved and renamed. Just a reminder for us at tools-suite.

packages/fe-fpm-writer/test/integration/index.test.ts
stefanschreck: OK, got it

packages/fe-fpm-writer/src/page/common.ts
tobiasqueck: Then a 2nd route with the same name is added to the `routes` array. The resulting app still works but ignores the previously defined route.
For me that is fine for the writer module because it still generates valid content. The responsibility to check first is with the caller.

packages/fe-fpm-writer/src/page/types.ts
stefanschreck: I was looking for 'beginColumnPages' | 'midColumnPages' | 'endColumnPages' in the snapshots or tests and could not find any.

packages/fe-fpm-writer/src/page/index.ts
tobiasqueck: No, the caller doesn't need to adapt. Both previously possible calls are still valid
```
import { generateCustomPage } from '@sap-ux/fe-fpm-writer';
import { generateCustomPage } from '@sap-ux/fe-fpm-writer/dist/page';
```

packages/fe-fpm-writer/src/page/types.ts
tobiasqueck: That is indeed odd. Let me check.

tobiasqueck: Check https://github.com/SAP/open-ux-tools/blob/fpm-op/packages/fe-fpm-writer/test/unit/page/__snapshots__/custom.test.ts.snap#L81

johannes-kolbe (approved):

packages/fe-fpm-writer/src/page/types.ts
stefanschreck: OK

stefanschreck (approved):

packages/fe-fpm-writer/src/page/index.ts
stefanschreck: OK then

packages/fe-fpm-writer/src/page/common.ts
stefanschreck: OK

PR #536, opened 2022-06-17
feat(fe-fpm-writer): Custom Actions. JS handler file. Support update of existing js handler with new custom action and allow to pass method name

815are:
While working on integration of custom actions.
Noticed some limitations regarding event handler function in `js` handler:
1. We can not pass custom function name for new handler js;
2. We can not update existing handler js file with additional method.

In result I am extending `eventHandler` property with additiona type:
```
export interface EventHandlerConfiguration {
    /**
     * JS file name - existing or new.
     */
    fileName?: string;
    /**
     * Function name. If undefined then `onPress` is used as default value.
     */
    fnName?: string;
    /**
     * If file exists, then existing file should be appended with passed script fragment.
     */
    insertScript?: TextFragmentInsertion;
}
```
For first point's issue we can can pass `fnName` like:
```
eventHandler: {
  fnName: 'onClick'
}
```

For second point we need pass object like:
```
{
    fnName: 'onClick',
    fileName: 'Handler',
    insertScript: {
        fragment: ',\n         function(){}',
        position: {
             line: 8,
             character: 9
        }
    }
}
```

In result existing js handler file will be updated in was that passed fragment would be inserted into passed position.


stefanschreck (changes requested):
Hello Andis,
I am not sure about your enhancement. The first point is necessary, i.e. passing the name of an existing event handler.
But I think that in this case no complicated insert logic should take place, but you just skip the event handler creation. Modifying an existing controller code is too dangerous. And how would you calculate the exact position from a UI?

packages/fe-fpm-writer/src/action/types.ts
```diff
@@ -25,8 +46,9 @@ export interface CustomAction extends CustomElement {
         /**
          * If not set (i.e. undefined) then no event handler is linked. If it is set true, a new one is created and linked to the action.
          * If an existing event handler is to be used then its id needs to be provided as string.
+         * Object with 'CustomActionNewEventHandler' provides option to append existing js file with new action.
          */
-        eventHandler?: string | true;
```
tobiasqueck, line 29: Should we just change the type or allow `string | true | CustomActionEventHandler` and be backward compatible? I am ok with replacing it, since we only have one known consumer, so changing it now is easy. Just wanted to have it mentioned and discussed.

815are: Hi @tobiasqueck 

I rechecked type and I think `string | CustomActionEventHandler` fits well.
By providing string we can simply set full path to existing method.
And if we need to create new js file or append existing js file with additional code fragment, then we can use object with `CustomActionNewEventHandler` interface.

Interface could look following:
```
export interface CustomActionNewEventHandler {
    /**
     * JS file name - existing or new.
     */
    fileName?: string;
    /**
     * Function name. If undefined then `onPress` is used as default value.
     */
    fnName?: string;
    /**
     * If file exists, then existing file should be appended with passed script fragment.
     */
    insertScript?: {
        /**
         * Insert position for passed script fragment.
         */
        position: FileContentPosition;
        /**
         * Full fragment of script to insert.
         */
        content: string;
    };
}

export type CustomActionEventHandler = string | true | CustomActionNewEventHandler;
```

Some examples would be:
1. New event handler(js file and method) with default file and function name:
```
{
    name: 'dummy',
    eventHandler: {},
    folder: join('ext', 'fragments')
}
```
New file at `ext/fragments/Dummy.js` with `onPress` will be created.

2. New event handler with custom function and file name:
```
{
    name: 'dummy',
    eventHandler: {
        fnName: 'onTest',
        fileName: 'newAction.js',
    },
    folder: join('ext', 'fragments')
}
```
New file at `ext/fragments/newAction.js` will be created. And in manifest we will have:
```
Dummy": {
  "press": "project4.ext.fragment.newAction.onPress",
  ...
}
```

3. Append script fragment to existing file(ext/fragments/MyExistingAction.js):
```
{
    eventHandler: {
        fnName: 'onHandleSecondAction',
        fileName: 'MyExistingAction.js',
        insertScript: {
            content: `,\n    onHandleSecondAction: function(){\n      console.log();\n    }`,
            position: {
                line: 8,
                character: 9
            }
        }
    },
    folder: join('ext', 'fragments')
}
```
`MyExistingAction.js` would be appended with additional js code fragment.

4. Use existing event handler:
```
{
    name: 'dummy',
    eventHandler: 'project4.ext.fragment.newAction.onPress',
    folder: join('ext', 'fragments')
}
```
No new or existing js files modified and entry in manifest would be:
```
Dummy": {
  "press": "project4.ext.fragment.newAction.onPress",
  ...
}
```

Is such interface ok for you?

zdravko-georgiev:
Something seems off with the `fe-fpm-writer` configuration. Only 2 checks were run and I can submit the change although there is -1 from Stefan 🤔

tobiasqueck:
> Something seems off with the `fe-fpm-writer` pipeline configuration. Only 2 checks were run and I can submit the change although there is -1 from Stefan 🤔

@kranthie-sap could you please have a look?

packages/fe-fpm-writer/src/action/types.ts
815are: @tobiasqueck 

I rechecked interface again and I think `true | string | EventHandlerConfiguration` fits well.
I think we need `string` for sure - in case to provide existing without any change in target js files.

We can get rid of `true` and use `eventHandler: {}` to setup handler with default values like:
1. file name by default equals to `config.name`;
2. function name in handler by default equals to `onPress`;

We can provide custom values by using properties `fileName` or `fnName` like:
```
eventHandler: {
    fnName: 'onPress',
    fileName: config.name
}
```
But I think it is better to have `fileName` and `fnName` as optional properties, which leads that shortest way to create handler is `eventHandler: {}`, which would be equal to `eventHandler: true`. So I think in result there no much benefit to remove `true` from `eventHandler?: true | string | EventHandlerConfiguration`.

Please correct if you have different interface in mind.

815are:
> @815are thanks a lot for the proposal. I am ok with adding the new logic allowing to reuse an existing handler or even inserting a the handler snippet. As @stefanschreck pointed out, inserting the code, and doing it properly formatted is not an easy thing, therefore, I would like to keep the writer simple, and only allow inserting a string in a specific line. I would prefer not thinking about prepending commas but really leave the string generation completely to the caller. In addition, once we agree on a good API/parameters, could we please do the very same for the other handlers as well, so that we consistently allow this new functionality for all custom elements with an event handler.

@tobiasqueck 
I updated PR with requirement `do the very same for the other handlers as well, so that we consistently allow this new functionality for all custom elements with an event handler`.
In result reusable method [applyEventHandlerConfiguration](https://github.com/SAP/open-ux-tools/blob/feat/fpm-support-passing-function-name-for-new-custom-action/packages/fe-fpm-writer/src/common/event-handler.ts#L16) is used by other custom extensions

zdravko-georgiev (dismissed):
Code looks good, just one small typo
Tests looks good

zdravko-georgiev (approved):

packages/fe-fpm-writer/src/action/types.ts
```diff
@@ -22,12 +22,7 @@ export interface CustomAction extends CustomElement {
         enabled?: string | true;
         requiresSelection?: boolean;
         position?: Position;
-        /**
-         * If not set (i.e. undefined) then no event handler is linked. If it is set true, a new one is created and linked to the action.
-         * If an existing event handler is to be used then its id needs to be provided as string.
-         */
-        eventHandler?: string | true;
-    };
+    } & EventHandler;
```
tobiasqueck, line 25: Not a real problem but could you please add the `EventHandler` to the extends section like you did it for the column types. I am aware that in this specific case it results in the same type but I'd like it to be used consistently.

tobiasqueck (dismissed):
* Code is clean and easy to understand
* extensive set of unit tests added
* i didn't run the integration tests locally but I assume you did it and it works

packages/fe-fpm-writer/src/action/types.ts
815are: correction submitted - https://github.com/SAP/open-ux-tools/pull/536/commits/93a705e1446ff30d218d7d563ea717d7b1c1b4ce

In result I have:
```
export interface CustomActionSettings extends EventHandler {
   ...
}

export interface CustomAction extends CustomElement {
    target: CustomActionTarget;
    settings: CustomActionSettings;
}
```

If we compare with other extension, then for other extensions interface is:
```
export interface CustomTableColumn extends CustomElement, EventHandler
```

`eventHandler` property was placed in `settings` for custom action also before my changes - [action types](https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/action/types.ts#L29)
In result I did not changed placement of `eventHandler` and left it in `settings`.

I also can move it to root level to have it same as for other extensions, like:
```
export interface CustomAction extends CustomElement, EventHandler {
```

zdravko-georgiev (dismissed):

tobiasqueck (approved):

zdravko-georgiev (approved):

PR #540, opened 2022-06-21
Remove express when not required 

tobiasqueck:
I noticed that `axios-extension` is using `express` for a very simple server use case. I replaced it with usage of the `http` module and can reduce the installation size of the package and all its dependencies from 13m to 11m.
I also noticed that our middlewares are using a specific version of express and when they are used somewhere else by e.g. the ui5-tooling using a different express version, then express gets installed twice, so I replaced it as well with the simple `http` module

- [x] remove express from `axios-extension` (11mb -> 9mb installation size) 
- [x] remove express from `backend-proxy-middleware` (22mb -> 19mb installation size)
- [x] remove express from `ui5-proxy-middleware` (17mb -> 15mb installation size)


tobiasqueck:
@ullasholla please closely look at the changes to the `axios-extension` to make sure I didn't change any logic. @zdravko-georgiev please do the same for the middlewares.
All unit tests run without issues and my manual tests also didn't show a problem but still, please look very closely.

zdravko-georgiev:
> @ullasholla please closely look at the changes to the `axios-extension` to make sure I didn't change any logic (please also checkout the new unit test)
> 
> @zdravko-georgiev please do the same for the middlewares - especially the `ui5-proxy-middleware` because I also reduced the cognitive complexity of some of the code. All unit tests run without issues and my manual tests also didn't show a problem but still, please look very closely.

Thanks @tobiasqueck, I will check

zdravko-georgiev (dismissed):
Code looks good.
Didn't found any issues when testing manually

zdravko-georgiev (dismissed):

packages/axios-extension/src/auth/reentrance-ticket/redirect.ts
```diff
@@ -30,35 +29,38 @@ export interface SetupRedirectOptions {
...
-        } else {
-            logger.error('Error getting reentrance ticket');
-            logger.debug(req);
-            res.set('Content-Type', 'text/html');
-            res.status(500);
-            res.send(Buffer.from(redirectErrorHtml()));
-            server.close();
-            reject(new ConnectionError('Error getting reentrance ticket'));
+    server = http.createServer((req, res) => {
+        const reqUrl = new URL(req.url, `http:${req.headers.host}`);
```
ullasholla, line 44: The second parameter to the `URL()` constructor gets interpolated to a value like `http:localhost:30034/redirect?...`

The code works but the URL standard specifies that the [absolute URL string](https://url.spec.whatwg.org/#absolute-url-string) must have `//` after the scheme. Consider changing this?

packages/axios-extension/src/auth/uaa.ts
```diff
@@ -179,24 +178,27 @@ export class Uaa {
...
             const timer = setTimeout(handleTimeout, timeout);
-            app.get(Redirect.path, (req, res) => {
-                res.set('Content-Type', 'text/html');
-                res.send(Buffer.from(redirectSuccessHtml(this.logoutUrl, this.systemId)));
-                this.log.info('Got authCode');
-                resolve({ authCode: req.query.code + '', redirect });
-                if (timer) {
-                    clearTimeout(timer);
+            server = http.createServer((req, res) => {
+                const reqUrl = new URL(req.url, `http:${req.headers.host}`);
```
ullasholla, line 191: Question about the interpolated base URL. Same as: https://github.com/SAP/open-ux-tools/pull/540/files#r974651046

ullasholla (dismissed):
Thanks @tobiasqueck, looks good.
Have two minor queries. Could you take a look please?

packages/axios-extension/src/auth/uaa.ts
tobiasqueck: https://github.com/SAP/open-ux-tools/pull/540/commits/6eca7ee931b046459cc6f27a29d3d4ea4d99d050

packages/axios-extension/src/auth/reentrance-ticket/redirect.ts
tobiasqueck: https://github.com/SAP/open-ux-tools/pull/540/commits/6eca7ee931b046459cc6f27a29d3d4ea4d99d050

ullasholla (approved):
@tobiasqueck, thanks for addressing the review comments. The changes and tests look good to me

tobiasqueck:
I admin-merged it because @zdravko-georgiev is on vacation, however, he approved the relevant parts for him a few times already.

PR #543, opened 2022-06-21
feat: new writer for OPA tests with Fiori elements for OData v4

vescalier:
Generates files for OPA tests for Fiori elements V4 app
- `test/testsuite.qunit.[html | js]`: test suite files
- `test/integration/opaTests.qunit.[html | js]`: OPA test files
- `test/integration/FirstJourney.js`: basic OPA journey that launches the app, checks ListReport visibility, triggers data load and clicks on the first row when the initial page of the app is a ListReport. If not, just checks the first page visiblity. If no initial page could be found in the manifest, the journey just launches and terminates the app.
- `test/integration/pages/*`: page files, one for each page we could find in the app manifest

All these files are overwritten in case they already exist in the `webapp` folder


vescalier:
@devinea @tobiasqueck @IainSAP

packages/fiori-elements-opa-writer/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "devDependencies": {
+        "@types/ejs": "3.1.0",
+        "@types/mem-fs": "1.1.2",
+        "@types/fs-extra": "9.0.13",
+        "@types/mem-fs-editor": "7.0.0",
+        "fs-extra": "10.0.0"
+    },
+    "engines": {
+        "pnpm": ">=6.26.1",
+        "node": ">=12.22.5 < 13.0.0 || >= 14.16.0 < 15.0.0 || >=16.1.0 < 17.0.0"
```
tobiasqueck, line 44: v12 is not supported by our modules anymore, please change it to `"node": ">= 14.16.0 < 15.0.0 || >=16.1.0 < 17.0.0 || >=18.0.0 < 19.0.0"`

packages/fiori-elements-opa-writer/src/index.ts
```diff
@@ -0,0 +1,138 @@
...
+        config,
+        undefined,
+        {
+            globOptions: { dot: true }
+        }
+    );
+
+    // Pages files (one for each page in the app)
+    config.pages.forEach((page) => {
+        fs?.copyTpl(
```
tobiasqueck, line 107: Would it be possible to extract and expose the generation of a page object also an additional function e.g. `generatePageObject`. If we have this, then the Fiori tools page modeler could also call it whenever a new page is being added with the `generate[TYPE]Page` functions in the `fe-fpm-writer`?.

tobiasqueck:
Couple of questions:

- Would it also be possible to create a test skeleton and a minimal journey for an FE v4 FPM app that just has a custom page (no LR or OP)?
- Could we also do something for OVP?
- Could we do something for FE v2? Maybe similar to the custom page scenario, just create the skeleton and a dummy start page.

packages/fiori-elements-opa-writer/src/index.ts
```diff
@@ -3,12 +3,45 @@ import { create as createStorage } from 'mem-fs';
...
+ * @returns Page configuration object, or undefined if the target type is not supported
+ */
+function createPageConfig(manifest: Manifest, targetKey: string): FEV4OPAPageConfig | undefined {
+    const appTargets = manifest['sap.ui5']?.routing?.targets;
+    const target = appTargets && appTargets[targetKey];
+    const appID = manifest['sap.app'].id;
+    const appPath = appID.split('.').join('/');
+
+    if (
+        target &&
```
tobiasqueck, line 24: small thing: you could start with `target?.id` and then remove line 27

packages/fiori-elements-opa-writer/src/index.ts
```diff
@@ -3,12 +3,45 @@ import { create as createStorage } from 'mem-fs';
...
 /**
- * Create the configuration object from the app descriptor and the FLP app name.
+ * Create the page configuration object from the app descriptor and the target key.
+ * @param manifest - the app descriptor of the app
+ * @param targetKey - the key of the target in the manifest
+ * @returns Page configuration object, or undefined if the target type is not supported
+ */
+function createPageConfig(manifest: Manifest, targetKey: string): FEV4OPAPageConfig | undefined {
+    const appTargets = manifest['sap.ui5']?.routing?.targets;
+    const target = appTargets && appTargets[targetKey];
```
tobiasqueck, line 19: Would it be possible to define the type for target somewhere central or at least here, so that you don't have to do it again and again below.

tobiasqueck:
Could we please rename the module to `@sap-ux/ui5-test-writer` to keep it as the module for further test generation functionality.

packages/fiori-elements-opa-writer/src/index.ts
vescalier: Not sure I get what you mean...

packages/fiori-elements-opa-writer/src/index.ts
vescalier: done

packages/fiori-elements-opa-writer/src/index.ts
tobiasqueck: In line 19, you are getting a reference to `target`. The type that reference seems not to be clear because you have to explicitly define the type of target in line 26, 28 and 35. In each of the lines, you don't have any conditions leading to the type you used, therefore, I'd say that you already know in line 19 exactly the type of target (or at least you exactly know a subset).
So, line 19 should look like `const target: { name?: string; options?: OptionsTypeXYZ } = ...;` BUT if that is not allowed, then I am wondering if you are making an incorrect assumption of what `appTargets && appTargets[targetKey]` will be.

devinea:
@vescalier - can this be closed?

tobiasqueck:
That PR is obsolete, it is superseded by #573

PR #545, opened 2022-06-22
FIX: `store` module does not read the `api-hub` key from the secure storage

tobiasqueck:
- [x] fix a wrong entity name mapping when trying to read the api-hub key from the secure storage
- [x] update for the manual test script to also play with api-hub key


tobiasqueck (commented):
Since I created the draft PR, I cannot officially approve it but I still reviewed

- code changes are easy to understand
- overall code is clean
- tests add and are also easy to understand
- successfully tested the migration from an old entry to the new format

zdravko-georgiev:
@tobiasqueck Do we need to add a changeset?

zdravko-georgiev (approved):

ullasholla (approved):
@tobiasqueck and @zdravko-georgiev, approving this as we're in agreement about the PR

Issue #546, opened 2022-06-22
BUG - getCredentialsForDestinationService returns wrong credentials


zdravko-georgiev:
### Description
The function [getCredentialsForDestinationService](https://github.com/SAP/open-ux-tools/blob/main/packages/btp-utils/src/app-studio.ts#L35) returns the wrong credentials because it fails to read correctly the `clientId` and `clientSecret` from the service info. 

#### Current State
```Typescript 
  const serviceInfo = await cfGetInstanceKeyParameters(instance);
  const clientId = serviceInfo.uaa?.clientid || serviceInfo.clientid;
  const clientSecret = serviceInfo.uaa?.clientsecret || serviceInfo.clientsecret;

```
#### Correct State
```Typescript 
  const serviceInfo = await cfGetInstanceKeyParameters(instance);
  const clientId = serviceInfo.credentials.uaa?.clientid || serviceInfo.credentials.clientid;
  const clientSecret = serviceInfo.credentials.uaa?.clientsecret || serviceInfo.credentials.clientsecret;

```

### Expected results
Correct credentials are returned

### Actual results
Wrong credentials are returned

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

zdravko-georgiev:
Fixed

PR #547, opened 2022-06-22
fix(btp-utils): fix getCredentialsForDestinationService

zdravko-georgiev:
Issue: https://github.com/SAP/open-ux-tools/issues/546

### DIFF
--- a/packages/btp-utils/src/app-studio.ts
+++ b/packages/btp-utils/src/app-studio.ts
@@ -34,9 +34,9 @@ export function getAppStudioProxyURL(): string | undefined {
  */
 export async function getCredentialsForDestinationService(instance: string): Promise<string> {
     try {
-        const serviceInfo = await cfGetInstanceKeyParameters(instance);
-        const clientId = serviceInfo.uaa?.clientid || serviceInfo.clientid;
-        const clientSecret = serviceInfo.uaa?.clientsecret || serviceInfo.clientsecret;
+        const serviceCredentials = (await cfGetInstanceKeyParameters(instance)).credentials;
+        const clientId = serviceCredentials.uaa?.clientid || serviceCredentials.clientid;
+        const clientSecret = serviceCredentials.uaa?.clientsecret || serviceCredentials.clientsecret;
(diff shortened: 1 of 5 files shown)


packages/btp-utils/src/app-studio.ts
```diff
@@ -34,9 +34,9 @@ export function getAppStudioProxyURL(): string | undefined {
  */
 export async function getCredentialsForDestinationService(instance: string): Promise<string> {
     try {
-        const serviceInfo = await cfGetInstanceKeyParameters(instance);
-        const clientId = serviceInfo.uaa?.clientid || serviceInfo.clientid;
-        const clientSecret = serviceInfo.uaa?.clientsecret || serviceInfo.clientsecret;
+        const serviceCredentials = (await cfGetInstanceKeyParameters(instance)).credentials;
```
tobiasqueck, line 37: Is it possible that `await cfGetInstanceKeyParameters(instance)` returns `undefined` or `null`?

tobiasqueck (dismissed):
Small comment, if a null check is not required, then all is good.

packages/btp-utils/src/app-studio.ts
zdravko-georgiev: Yes, it is possible https://github.com/SAP/cloud-foundry-tools-api/blob/master/src/cf-local.ts#L592. I will update the PR.

zdravko-georgiev: @tobiasqueck On second thought, if it returns `undefined`, then an error will be thrown that we can get credentials. Which is OK, right? I mean the whole flow can't work without credentials AFAIK.

zdravko-georgiev: Done with https://github.com/SAP/open-ux-tools/pull/547/commits/e1d0434d06b5a6cb816f3de2313a8892f58ab088

tobiasqueck (approved):

tobiasqueck (approved):

Klaus-Keller (approved):
Code looks good, test exists, approved.

PR #551, opened 2022-06-23
made route generation ready for FCL

tobiasqueck:
Follow up to #534 after testing the resulting applications with FCL

- [x] removed unused input property
- [x] enhanced the route generation logic to hand nested routes and first level routes --> this results in no route starting with a `/` which caused problems when using FCL
- [x] enhanced unit tests
- [x] refactored the route generation coding to decrease the cognitive complexity

### DIFF
--- a/packages/backend-proxy-middleware/CHANGELOG.md
+++ b/packages/backend-proxy-middleware/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/backend-proxy-middleware
 
+## 0.6.3
+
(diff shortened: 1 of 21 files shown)


johannes-kolbe (approved):
LGTM, manual test showed double '//' resolved.

Klaus-Keller (approved):
- Code looks good 
- Tests and test coverage are excellent
- Built and ran test locally from branch, but didn't integrate it back in specification

Approved

PR #555, opened 2022-06-27
feat(fe-fpm-writer): add support to generate building blocks

kranthie-sap:
The following building blocks can be generated using the fe-fpm-writer module:
* [Chart](https://ui5.sap.com/#/api/sap.fe.macros.Chart)
* [FilterBar](https://ui5.sap.com/#/api/sap.fe.macros.FilterBar)
* [Field](https://ui5.sap.com/#/api/sap.fe.macros.Field)


packages/fe-fpm-writer/src/building-block/index.ts
```diff
@@ -0,0 +1,155 @@
...
+    fs: Editor
+): Editor {
+    // Validate the base and view paths
+    validateBasePath(basePath, fs);
+    if (!fs.exists(join(basePath, viewPath))) {
+        throw new Error(`Invalid view path ${viewPath}.`);
+    }
+
+    // Read the view xml and template files and update contents of the view xml file
+    const viewDocument = getViewDocument(basePath, viewPath, fs);
```
tobiasqueck, line 34: Can we also handle fragments?

packages/fe-fpm-writer/src/building-block/index.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    return viewDocument;
+}
+/**
+ * Returns the macros namespace from the xml view document.
+ *
+ * @param {Document} viewDocument - the view xml file document
+ * @returns {string} the macros namespace
+ */
+function getMacrosNamespace(viewDocument: Document): string {
```
tobiasqueck, line 77: This is more than just getting it, correct? if you don't find it, you add the namespace. If I understand it correctly, then, I would suggest to change the function name and description to indicate that.

packages/fe-fpm-writer/src/building-block/types.ts
```diff
@@ -0,0 +1,342 @@
...
+
+/**
+ * Represents a Chart building block control.
+ * Usually, a contextPath and metaPath is expected.
+ *
+ * @example
+ * <macro:Chart id="Mychart" contextPath="/RootEntity" metaPath="@com.sap.vocabularies.UI.v1.Chart" />
+ * @extends {BuildingBlock}
+ */
+export interface Chart extends BuildingBlock {
```
tobiasqueck, line 45: Could you please also add the optional event handlers.

tobiasqueck (commented):
Code is very clean and easy to understand. I just added a few minor comments, would be great if you can address these. I have not locally tested it yet.
Do you think you can also add one or two building blocks to the integration test? If not, don't worry, we can do it later, however, would be nice to have.

packages/fe-fpm-writer/templates/building-block/field/View.xml
```diff
@@ -0,0 +1,8 @@
+<<%- macrosNamespace %>:Chart
+    id="<%- data.id %>"<% if (data.metaPath) { %>
```
nlunets, line 2: Probably Field rather than Chart :)

nlunets (changes requested):

nlunets:
Overall looks good but one tiny problem with the Field :)

packages/fe-fpm-writer/templates/building-block/field/View.xml
kranthie-sap: Thanks for catching it. :)

packages/fe-fpm-writer/src/building-block/types.ts
kranthie-sap: @tobiasqueck: I see that I missed the `selectionChange` event. I will add this. I haven't added any inherited events, properties or methods. Let me know if you need any other events or properties and I will add them for now. Hopefully, when the types from Fiori elements are used, all properties can be used.

packages/fe-fpm-writer/src/building-block/index.ts
kranthie-sap: The code should work for both xml views and xml fragment files. I updated the public function's variable name and docs to indicate this. I haven't updated this in other places in the file for better readability.

packages/fe-fpm-writer/src/building-block/types.ts
kranthie-sap: I added the `selectionChange` event to the interface and the template.

packages/fe-fpm-writer/templates/building-block/field/View.xml
kranthie-sap: Updated the template and test snapshots.

packages/fe-fpm-writer/src/building-block/index.ts
kranthie-sap: Good catch. Updated the function name.

kranthie-sap:
@tobiasqueck:
> Do you think you can also add one or two building blocks to the integration test? If not, don't worry, we can do it later, however, would be nice to have.

Yes, I plan to update the integration tests in the next PR when I add support for other building blocks.

nlunets (approved):
All good now :)

packages/fe-fpm-writer/src/building-block/types.ts
tobiasqueck: Thanks, I am ok with it for now, and we can enhance it later.

packages/fe-fpm-writer/src/building-block/index.ts
```diff
@@ -0,0 +1,155 @@
...
+ * Generates a building block into the provided xml view file.
+ *
+ * @param {string} basePath - the base path
+ * @param {string} viewOrFragmentPath - the path of the view or fragment xml file relative to the base path
+ * @param {string} aggregationPath - the aggregation xpath
+ * @param {BuildingBlock} buildingBlockData - the building block data
+ * @param {Editor} [fs] - the memfs editor instance
+ * @returns {Editor} the updated memfs editor instance
+ */
+export function generateBuildingBlock<T extends BuildingBlock>(
```
tobiasqueck, line 20: Sorry, I missed this in my first review, however, i would like to have a the `generate` functions consistent in the module. We always have three parameters: `basePath`, an object will all specific configs, and optional `fs`.

packages/fe-fpm-writer/src/building-block/index.ts
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * Returns the view xml file document.
+ *
+ * @param {string} basePath - the base path
+ * @param {string} viewPath - the path of the xml view relative to the base path
+ * @param {Editor} fs - the memfs editor instance
+ * @returns {Document} the view xml file document
+ */
+function getViewDocument(basePath: string, viewPath: string, fs: Editor): Document {
```
tobiasqueck, line 49: The function also works with fragments, also indicated in your public function, so I'd suggest to rename it as well.

tobiasqueck (changes requested):

nlunets (approved):

tobiasqueck (approved):
Provided fixes for my complaints :)

Otherwise, the code looks great, both src and test are well written and easy to understand.
Local tests also worked without issues.

packages/fe-fpm-writer/src/building-block/index.ts
tobiasqueck: https://github.com/SAP/open-ux-tools/pull/555/commits/9229ec75d8eff92ceeca032f392eda1c602d6b5a

PR #560, opened 2022-06-30
chore(open-ux-tools) ignore source map files when publishing to npm

devinea:
Source map files are not needed in the tgz published to npm. Removing reduces the install size.

### DIFF
--- a/.changeset/tricky-mice-whisper.md
+++ b/.changeset/tricky-mice-whisper.md
@@ -1,0 +1,19 @@
+---
+'@sap-ux/odata-cli': patch
+'@sap-ux/generator-simple-fe': patch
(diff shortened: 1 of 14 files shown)


zdravko-georgiev (approved):
Thanks Austin

IainSAP:
@devinea This change would mean we can never debug - this does not only remove during publish. We need to still be able to generate dist/tar including source maps for debugging on BAS for example.

IainSAP (changes requested):
Please check my comment and see if this is correct?

tobiasqueck:
>  Im not suggesting we publish with source maps, but we still need a way to generate a tgz that includes them.

That sounds like a good compromise to me.

Issue #575, opened 2022-07-13
BUG - backend-proxy doesn't support using direct url in BAS


zdravko-georgiev:
### Description
Using a direct OData service URL, e.g. https://services.odata.org/V2/OData/OData.svc/, in BAS doesn't work with the `backend-proxy-middleware`. The local server fails to start because in the case of BAS the `backend-proxy-middleware` expects a destination and thus ignores the url from the `ui5.yaml`. As a result an exception is thrown and the server fails to start

### Expected results
I can start an application using a direct OData service URL in BAS

### Actual results
I can't start an application using a direct OData service URL in BAS

### Screenshots
![image](https://user-images.githubusercontent.com/91968887/178690569-c1bcacae-6cc7-493e-a106-fb67c1296362.png)

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [x] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #576, opened 2022-07-13
fix(backend-proxy-middleware): support direct OData service url in BAS

zdravko-georgiev:
Fixes https://github.com/SAP/open-ux-tools/issues/575

### DIFF
--- a/.changeset/chilly-dogs-call.md
+++ b/.changeset/chilly-dogs-call.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/backend-proxy-middleware': patch
+---
+
+Support using direct OData service URLs in BAS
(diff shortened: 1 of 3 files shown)


Klaus-Keller (approved):
- Code looks good
- Unit test coverage is added
Approved

Klaus-Keller (approved):
Thanks for fixing the code smell

devinea (approved):
Thanks @zdravko-georgiev 
Changes and tests look good.

tobiasqueck:
@zdravko-georgiev while your PR fixes the problem, it is from a typing perspective incorrect what you did. Could you do me a favor and add another correction?

The problem that you uncovered is that the type `DestinationBackendConfig` was incorrectly defined, it should have contained the optional `url` parameter. Once you correct that, you can remove the incorrect type assignment in https://github.com/SAP/open-ux-tools/blob/main/packages/backend-proxy-middleware/src/base/proxy.ts#L290-L291 and just do `proxyOptions.target = backend.url;`.

Technically it doesn't make a difference for fixing this issue but it corrects the type, hopefully helping to prevent and uncover similar issues in the future.

zdravko-georgiev:
Thanks for reviewing @tobiasqueck, will prepare a PR with the correction.

PR #589, opened 2022-07-19
feat(ui5 application writer): trim themelib configuration property

korotkovao:
Trim `themelib` property value.
https://github.com/SAP/open-ux-tools/issues/590

### DIFF
--- a/packages/ui5-config/src/ui5config.ts
+++ b/packages/ui5-config/src/ui5config.ts
@@ -84,14 +84,14 @@ export class UI5Config {
         ui5Framework: string,
         ui5Version: string,
         ui5Libraries: string[],
-        ui5Theme = 'sap_fiori_3'
+        ui5Theme = 'sap_horizon'
     ): UI5Config {
         const libraryObjs = [];
         for (const library of ui5Libraries) {
             libraryObjs.push({ name: library });
         }
         // Add theme lib (dark theme versions are provided by base theme lib)
-        libraryObjs.push({ name: `themelib_${ui5Theme.replace(/_dark$/, '')}` });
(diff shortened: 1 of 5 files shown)


devinea (commented):
Changes and test look good. Thanks @korotkovao 
Can you add a match changeset?
https://github.com/SAP/open-ux-tools/blob/9d4e3bcfd4a1d45e93ba603ecf345e34896cedf8/README.md#create-changesets-for-feature-or-bug-fix-branches

packages/ui5-config/src/ui5config.ts
```diff
@@ -91,7 +91,7 @@ export class UI5Config {
             libraryObjs.push({ name: library });
         }
         // Add theme lib (dark theme versions are provided by base theme lib)
-        libraryObjs.push({ name: `themelib_${ui5Theme.replace(/_dark$/, '')}` });
+        libraryObjs.push({ name: `themelib_${ui5Theme.replace(/_dark|_hcw|_hcb/, '')}` });
```
IainSAP, line 94: Are you sure about dropping the `$`. This changes the regex match to match anywhere in the string not just the end. I think we should retain the precision.

IainSAP (changes requested):

packages/ui5-config/src/ui5config.ts
korotkovao: Thanks @IainSAP, i have added it back

IainSAP (dismissed):
Thx @korotkovao

devinea (dismissed):
Thanks @korotkovao

korotkovao:
Thank you @devinea and @IainSAP

IainSAP (approved):

PR #597, opened 2022-07-20
feat(fe-fpm-writer): Custom action. Empty string 'anchor' property created if position provided without 'anchor'.

815are:
Bug fix for issue, when we pass `position` without `anchor` in `setting` object to `generateCustomAction` as:
```
 settings: {
       ...settings,
       position: {
            placement: 'Before'
       }
}
```

problem - in manifest it is created as:
```
"position": {
      "placement": "Before",
      "anchor": ""
},
```

correct way should be:
```
"position": {
      "placement": "Before"
},
```

### DIFF
--- a/.changeset/witty-boats-divide.md
+++ b/.changeset/witty-boats-divide.md
@@ -1,5 +1,5 @@
 ---
-'@sap-ux/fe-fpm-writer': minor
+'@sap-ux/fe-fpm-writer': patch
 ---
 
 Custom Action. Optional "anchor" position property. Do not include "anchor" property if undefined is passed
--- a/.changeset/witty-boats-divide.md
+++ b/.changeset/witty-boats-divide.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fe-fpm-writer': minor
+---
+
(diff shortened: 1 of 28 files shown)


.changeset/witty-boats-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/fe-fpm-writer': minor
```
zdravko-georgiev, line 2: Do we really need minor version bump? Isn't `patch` sufficient?

zdravko-georgiev (approved):
Code looks good, just one comment regarding the version bump

stefanschreck (approved):

.changeset/witty-boats-divide.md
stefanschreck: that's the question indeed

815are: thanks, makes sense, I was not aware of `patch`.
Corrected to `patch` with commit 26c545bef8dbbf0a7449d80f247b4f092aa1fec7

stefanschreck (approved):

zdravko-georgiev (approved):

PR #598, opened 2022-07-21
fix: support existing connection sessions

slin-sap:
#518 

Add optional input parameter `cookies` when creating abap service provider. `cookies` contains serialized cookie sessions for an existing connection. If `cookies` is passed to the abap service provider, the provider doesn't require authentication and can directly request the backend abap services.


kranthie-sap (approved):
Looks good to me.

ullasholla (approved):
Thanks @slin-sap, looks good. We may need to account for the CSRF token later

PR #601, opened 2022-07-27
chore(deps): override glob-parent to 6.0.1

zdravko-georgiev:


longieirl (approved):

devinea (approved):

donal-tobin-sap (approved):
👍🏽 Looks good to me.

**N.B.** strictly not required as 5.1.2 is perfectly valid as well (despite Mend reports otherwise) as stated by GitHub Security.

tobiasqueck:
@zdravko-georgiev @donal-tobin-sap why did we need this? All nested dependencies to `glob-parent` got to version `6.0.2`.

tobiasqueck:
Thinking about it a bit more, I think it is a bad idea because now you are enforcing a `6.x` version even if it is used nested with a `5.x` version. How do we ensure that the nested usage is not broken?

zdravko-georgiev:
> @zdravko-georgiev @donal-tobin-sap why did we need this? All nested dependencies to `glob-parent` got to version `6.0.2` or `5.1.2`.

@tobiasqueck The security scan for the `@sap/ux-specification` package reported the version `5.1.2` as high risk. The `glob-parent` dependency comes from the dependency to the `mem-fs-editor`. Since `mem-fs-editor` still uses `glob-parent@5.1.2` the idea was to enforce version `6.0.2`. Interesting enough if I install now e.g. `@sap-ux/fe-fpm-writer@0.13.4` and I check the `glob-parent` then the version is still `5.1.2`... so I am wondering now if this change had any effect at all?

tobiasqueck:
I am confused. https://github.com/advisories/GHSA-cj88-88mr-972w states that versions before `5.1.2` and before `6.0.1` are problematic. `5.1.2` itself should be fine.

If I remove the entry from the `overrides` and then run `pnpm install && pnpm why glob-parent` then I don't see any problematic version (`pnpm audit` confirms it as well), therefore, could we please revert this change.

Issue #608, opened 2022-08-02
BUG - backend-proxy-middleware fails with a full_url destination


zdravko-georgiev:
### Description
The `backend-proxy-middleware` fails to proxy requests when full_url destination is used. The root causes for the issues are:

1. The check [isFullUrlDestination](https://github.com/SAP/open-ux-tools/blob/main/packages/btp-utils/src/destination.ts#L93) in `btp-utils` is wrong. It is checking for the existence of the `full_url` in the `WebIDEUsage`, but it should be checking in `WebIDEAdditionalData`.
2. The replace logic in [backend-proxy-middleware](https://github.com/SAP/open-ux-tools/blob/main/packages/backend-proxy-middleware/src/base/proxy.ts#L224) needs to be adapted for this special use case.

### Expected results
Requests when using full_url destination are proxied correctly

### Actual results
Requests when using full_url destination are not proxied correctly

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #609, opened 2022-08-02
fix(backend-proxy-middleware): fix handling of full url destinations

zdravko-georgiev:
Fixes: https://github.com/SAP/open-ux-tools/issues/608

### DIFF
--- a/packages/backend-proxy-middleware/src/base/proxy.ts
+++ b/packages/backend-proxy-middleware/src/base/proxy.ts
@@ -220,10 +220,8 @@ export async function enhanceConfigsForDestination(
         if (destination) {
             // in case of a full url destination remove the path defined in the destination from the forwarded call
             if (isFullUrlDestination(destination)) {
-                const destPath = new URL(destination.Host).pathname.replace(/\/$/, '');
-                if (backend.path.startsWith(destPath) && !backend.pathReplace) {
-                    backend.pathReplace = backend.path.replace(destPath, '');
-                }
+                backend.path = new URL(destination.Host).pathname.replace(/\/$/, '');
+                backend.pathReplace = backend.pathReplace ?? '/';
             }
         } else {
             throw new Error();
--- a/packages/backend-proxy-middleware/test/base/proxy.test.ts
+++ b/packages/backend-proxy-middleware/test/base/proxy.test.ts
@@ -219,7 +219,8 @@ describe('proxy', () => {
             mockListDestinations.mockResolvedValueOnce({
                 [backend.destination]: {
                     Host: 'http://backend.example/sap',
-                    WebIDEUsage: `${WebIDEUsage.ODATA_GENERIC},${WebIDEUsage.FULL_URL}`
+                    WebIDEUsage: `${WebIDEUsage.ODATA_GENERIC}`,
+                    WebIDEAdditionalData: 'full_url'
                 }
             });
             const proxyOptions: OptionsWithHeaders = { headers: {} };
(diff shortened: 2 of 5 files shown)


IainSAP (approved):
Looks good, thx @zdravko-georgiev

devinea (dismissed):
@zdravko-georgiev Changes look good per issue.

packages/backend-proxy-middleware/test/base/proxy.test.ts
```diff
@@ -219,14 +219,33 @@ describe('proxy', () => {
             mockListDestinations.mockResolvedValueOnce({
                 [backend.destination]: {
                     Host: 'http://backend.example/sap',
-                    WebIDEUsage: `${WebIDEUsage.ODATA_GENERIC},${WebIDEUsage.FULL_URL}`
+                    WebIDEUsage: `${WebIDEUsage.ODATA_GENERIC}`,
+                    WebIDEAdditionalData: 'full_url'
```
tobiasqueck, line 223: Why did you replace the constant here?

tobiasqueck (changes requested):

packages/backend-proxy-middleware/test/base/proxy.test.ts
zdravko-georgiev: Because there is no `full_url` value for the property `WebIDEUsage`. The idea was to not cause any confusion.

tobiasqueck: Let me rephrase, shouldn't it be
```
[backend.destination]: {
                    Host: 'http://backend.example/sap',
                    WebIDEUsage: WebIDEUsage.ODATA_GENERIC,
                    WebIDEAdditionalData: WebIDEUsage.FULL_URL
                }
```

tobiasqueck: Even cleaner, should we update the constants?

zdravko-georgiev: Yes, we should update the constants. I already did some update here https://github.com/SAP/open-ux-tools/pull/609/files#diff-32ab4a9ed3098e009c7d23e32540229f4c16ead04750a87224d55d42acd4664fL8. I can introduce one new enum called `WebIDEAdditionalData`?

packages/backend-proxy-middleware/src/base/proxy.ts
```diff
@@ -220,10 +220,8 @@ export async function enhanceConfigsForDestination(
         if (destination) {
             // in case of a full url destination remove the path defined in the destination from the forwarded call
             if (isFullUrlDestination(destination)) {
-                const destPath = new URL(destination.Host).pathname.replace(/\/$/, '');
-                if (backend.path.startsWith(destPath) && !backend.pathReplace) {
-                    backend.pathReplace = backend.path.replace(destPath, '');
-                }
+                backend.path = new URL(destination.Host).pathname.replace(/\/$/, '');
+                backend.pathReplace = backend.pathReplace ?? '/';
```
tobiasqueck, line 224: Could you explain the change? If I look at the two implementations the behavior changes for two cases
1. if `pathReplace` is undefined/null you now set it to `/`.
2. if the `pathReplace` is undefined and the `path` contains the path of of the full url destinations, then `pathReplace` is used to make it work

(1) surprises me because, I'd expect that we don't do anything if `pathReplace` is undefined and (2) seemed logical to me, and I am confused that it has been removed.

packages/backend-proxy-middleware/test/base/proxy.test.ts
tobiasqueck: A new enum would be my preference. Do we know more possible values? I couldn't find any doc.

packages/backend-proxy-middleware/src/base/proxy.ts
zdravko-georgiev: Sure, I will try to explain the best I could. Let's say that we have a full url destination with the following URL `https://services.odata.org/V4/Northwind/Northwind.svc/`. When an application is generated then in the `manifest.json` we will have as OData URI `/V4/Northwind/Northwind.svc/` and in the `ui5.yaml` we will have as `/V4` as `path` for the `backend-proxy-middleware`.
Now when the application is started all backend request are starting with `/V4/Northwind/Northwind.svc/`, but since we have a full url destination this part of the request needs to be deleted. So I am setting the `backend.path = /V4/Northwind/Northwind.svc/` and `backend.pathReplace = '/'` and now e.g. when a request like this comes `/V4/Northwind/Northwind.svc/$metadata` it will be rewritten just to `/$metadata`.


> (2) seemed logical to me, and I am confused that it has been removed.
I removed it because the check `backend.path.startsWith(destPath)` always returns false, because `backend.path=/V4` and `destPath=/V4/Northwind/Northwind.svc/`, so `pathReplace` is not set the and the `pathReplace` logic is not triggering at all.

packages/backend-proxy-middleware/test/base/proxy.test.ts
zdravko-georgiev: This is what I found: https://help.sap.com/docs/help/825270ffffe74d9f988a0f0066ad59f0/5c3debce758a470e8342161457fd6f70.html?locale=en-US

image

zdravko-georgiev: Done with https://github.com/SAP/open-ux-tools/pull/609/commits/22a938b34e4d0153464b38bd2239261478153a98

devinea (approved):

packages/backend-proxy-middleware/src/base/proxy.ts
tobiasqueck: Thanks a lot!

tobiasqueck (approved):
Code looks good, corrected tests make sense (you could have used the enum in all tests, but it is ok without it)
Thanks for the explanation.

zdravko-georgiev:
> Code looks good, corrected tests make sense (you could have used the enum in all tests, but it is ok without it) Thanks for the explanation.

Thanks Tobias!

PR #612, opened 2022-08-05
Axios thoughts

tobiasqueck:
@kranthie-sap @ullasholla as a follow up to our discussion, here is how I'd prefer the ADT code to be abstracted.


packages/axios-extension/src/abap/abap-service-provider.ts
```diff
@@ -47,9 +46,7 @@ export class AbapServiceProvider extends ServiceProvider implements AbapServiceP
     public async getAtoInfo(): Promise<AtoSettings> {
```
kranthie-sap, line 46: Should this also belong to the ADT service?

kranthie-sap (changes requested):
Looks good to me. There's one comment about `getAtoInfo`.
Also, it looks like the pipeline failed due to linting errors.

packages/axios-extension/src/abap/abap-service-provider.ts
slin-sap: Yes. This is also ADT service.

tobiasqueck:
This PR/branch is obsolete. Some of the proposed changes with further improvements have been applied by @slin-sap in other PRs

Issue #621, opened 2022-08-17
FEATURE - Typescript support for Fiori projects


tobiasqueck:
_As a **persona**,
I want to **develop UI5/Fiori applications using Typescript**,
so that I **make less mistakes thanks to the awesomeness of types**_

### Description
All writers should offer the option to generate source files either as JS or TS files. 

### Technical Design
1. `ui5-application-writer`: this is the most important one since it will contain all the required configurations
2. `fiori-freestyle-writer`: requires changes to all templates but should be relatively easy
3. `fiori-elements-writer`: needs the additional option to add TS support BUT should otherwise stay unchanged
4. `fe-fpm-writer`: requires changes to all templates but should be relatively easy 

On first glance, no other module should require a change.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have the interest to use Typescript
**when** I execute any of our writers with the typescript option set to true
**then** I will have a working TS project following best practice guidelines from https://github.com/ui5-community/ui5-ecosystem-showcase/pull/568

**Given** I have generated a TS project with the writers
**when** I execute any functionality with the SAP Fiori tools
**then** it will work without issues like in JS project.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

tobiasqueck:
@petermuessig thank you for https://github.com/ui5-community/ui5-ecosystem-showcase/pull/568. We will be evaluating it and if it doesn't cause any problems with the SAP Fiori tools then we can move forward with adding TS support.

petermuessig:
The `ui5-tooling-transpile` extensions have been released to npm:
https://www.npmjs.com/package/ui5-tooling-transpile

tobiasqueck:
I have enhanced our `ui5-application-writer` and the `fiori-freestyle-writer` to support typescript at for the basic template in #626. Afterwards, I tested it with the Application Modeler extension in the SAP Fiori tools:
- App Info page worked without problems and I used it flawlessly to install the libs, add the spec (that is currently missing in the template) and open the preview 
- I also opened the page map and saw the correct application structure (after removing an incorrect entry in the manifest that was generated but that is an issue in my test case)
- when trying to enable FCL and add pages then things started breaking and I will investigate further

All in all, I am pretty happy with the first test results.
fyi: @ashleytung @petermuessig @Klaus-Keller

tobiasqueck:
[...]

PR #629, opened 2022-08-17
Fix/621/ui5 app

tobiasqueck:
The PR corrects two issues discovered while implementing #621 

- when the `sapux` flag was added, the spec was not added resulting in a project that wouldn't work in Application Modeler
- enhanced `examples/simple-generator` to illustrate how to fetch the latest UI5 version 
- updated the local mapping file


packages/fiori-freestyle-writer/test/__snapshots__/index.test.ts.snap
```diff
@@ -898,7 +898,7 @@ title=My Test App",
   \\"sap.ui5\\": {
     \\"flexEnabled\\": true,
     \\"dependencies\\": {
-      \\"minUI5Version\\": \\"1.60.0\\",
+      \\"minUI5Version\\": \\"\\",
```
devinea, line 901: `minUI5Version` with no value looks incorrect

devinea (commented):
Generally I am not sure we should be calling out to npm registry in the writer. 
It seem like it belongs in the prompting phase. 
Better to have a default spec version or throw error if none provided to the generate inputs?

packages/fiori-freestyle-writer/test/__snapshots__/index.test.ts.snap
tobiasqueck: Why? If no minUI5Version is provided then the Fiori tools will use the latest version which would be the expected behavior if no min version or version is provided.

tobiasqueck:
> Better to have a default spec version or throw error if none provided to the generate inputs?

Hmm ... I'd prefer to generate a valid project over making it fail if spec is not provided as devDependency BUT if that is the preference, I can live with it. @Klaus-Keller @IainSAP opinions?

packages/ui5-application-writer/src/data/defaults.ts
```diff
@@ -225,3 +226,56 @@ function getLocalVersion({
...
+    });
+}
+
+/**
+ * Retrieve the tag version of the @sap/ux-specification based on the given version.
+ *
+ * @param ui5Version UI5 version used in the project
+ * @returns version tag
+ */
+export async function getSpecTagVersion(ui5Version: string | undefined): Promise<string> {
```
Klaus-Keller, line 260: @tobiasqueck, for each UI5 version there is a `dist-tag` for module `@sap/ux-specification`, see: https://www.npmjs.com/package/@sap/ux-specification -> Versions tab

Name pattern is `UI5-{major}.{minor}`, e.g. `UI5-1.103`. Each `dist-tag` points to the respective `@sap/ux-specification` version.

This means, there is no need to call `npm view`. As long as you know the UI5 version, you can build the `dist-tag` and write it to `package.json`, e.g.:
```
"devDependencies": {
    "@sap/ux-specification": "UI5-1.103"
    ...
```

if no UI5 version is set, you can use `latest`, that is then aligned with `"minUI5Version" : ""`.

tobiasqueck: @Klaus-Keller thanks a lot. I ported that code from our Fiori generator assuming it is the best way doing it.

tobiasqueck: Fixed with https://github.com/SAP/open-ux-tools/pull/629/commits/fcb0770d22aee090410919d20ee8e270b497a7bc.
@Klaus-Keller @devinea please close the conversation if this is acceptable for you.

packages/fiori-freestyle-writer/test/__snapshots__/index.test.ts.snap
IainSAP: This may be a mandatory property and should contain a value. Please see: https://ui5.sap.com/#/topic/be0cf40f61184b358b5faedaec98b2da

devinea: > Why? If no minUI5Version is provided then the Fiori tools will use the latest version which would be the expected behavior if no min version or version is provided.

Preview technically works.  But `minUI5Version` is mandatory as Iain mentioned and I think is checked during building or deploying the app. So better to keep a value here IMO.

packages/ui5-application-writer/src/data/defaults.ts
devinea: @tobiasqueck  thanks for https://github.com/SAP/open-ux-tools/commit/fcb0770d22aee090410919d20ee8e270b497a7bc

Klaus-Keller (approved):
Source code and test coverage looks good, but I did not build and run it locally.

packages/fiori-freestyle-writer/test/__snapshots__/index.test.ts.snap
tobiasqueck: @petermuessig we need your guidance here. @IainSAP is correct, the doc says that the `minUI5Version` is mandatory, so leaving it empty like I planned to seems not correct, however, I took your recent change to `ui5-language-assistant` (https://github.com/SAP/ui5-language-assistant/pull/472/) as guidance, where you also use the absence of the `minUI5Version` as an indicator to use the latest version.

petermuessig:
Just FYI: regarding version handling, for the open-source UI5 freestyle templates we provide a selection option for the framework (`openui5` or `sapui5`) and second we provide a version input which defaults to the latest version, by using the `package-json` script (just some pseudo code follows):

```js
const packageJson = require('package-json');
const npmPackage = <"@openui5/ts-types-esm" || "@sapui5/ts-types-esm">;
const latestVersion = (await packageJson(npmPackage)).version;
```

The npm package could also be something else like `@sapui5/distribution-metadata` for SAPUI5 and `@openui5/sap.ui.core` for OpenUI5 as above the ts-types are used. Or you read the `https://ui5.sap.com/version.json`.

packages/fiori-freestyle-writer/test/__snapshots__/index.test.ts.snap
tobiasqueck: Ok, I have removed the fix and we keep generating `minUI5Version = 1.60` if nothing is provided. I am still fully convinced it is not at all what a caller would expect but changing it is a bit more than just leaving it empty. This really needs attention soon but for now I have just enhanced the example generator to show to consumers how to fetch the latest version during the prompting phase.

packages/ui5-application-writer/test/data.test.ts
```diff
@@ -292,31 +305,46 @@ describe('Setting defaults', () => {
...
+                ...baseInput,
+                sourceTemplate
+            });
+            expect(merged.sourceTemplate).toEqual(sourceTemplate);
+        });
+    });
+
+    describe('getSpecTagVersion', () => {
+        test('get latest version', async () => {
+            expect(await getSpecTagVersion('')).toBe('latest');
```
devinea, line 339: await can be removed as it is no longer async

packages/ui5-application-writer/src/data/version-to-descriptor-mapping.json
```diff
@@ -1,5 +1,10 @@
 {
-    "latest": "1.40.0",
+    "latest": "1.99.0",
```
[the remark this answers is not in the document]
tobiasqueck: That must have been a copy and paste error. The file is correct now, it is a copy of https://github.com/SAP/ui5-manifest/blob/master/mapping.json

packages/ui5-application-writer/test/data.test.ts
tobiasqueck: done

Klaus-Keller (approved):
Focused my review on handling spec `dist-tags` in `getSpecTagVersion()` and `mergeWithDefaults()` and code looks good.

devinea (approved):
Thanks for the updates @tobiasqueck

PR #633, opened 2022-08-23
fix(fe-fpm-writer): building block generator does not add macros ns

kranthie-sap:
## Issue 
The building block generator in the `fe-fpm-writer` module does not add the `macros` namespace to the xml view or fragment file if the namespace doesn't already exist.

## Description
* Update building block generator to add the `xmlns:macros='sap.fe.macros'` namespace if the xml view or fragment file doesn't already contain it.
* Cleanup `.vscode/launch.json` to use the correct module names as prefixes and sort the configurations alphabetically.

### DIFF
--- a/packages/fe-fpm-writer/test/unit/building-block.test.ts
+++ b/packages/fe-fpm-writer/test/unit/building-block.test.ts
@@ -168,7 +168,7 @@ describe('Building Blocks', () => {
         ).toThrowError(/Unable to parse xml view file/);
     });
 
-    describe.only('Generate with just ID and xml view without macros namespace', () => {
+    describe('Generate with just ID and xml view without macros namespace', () => {
         const testInput = [
             {
                 buildingBlockData: {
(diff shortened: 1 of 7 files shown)


packages/fe-fpm-writer/test/unit/building-block.test.ts
```diff
@@ -168,6 +168,60 @@ describe('Building Blocks', () => {
         ).toThrowError(/Unable to parse xml view file/);
     });
 
+    describe.only('Generate with just ID and xml view without macros namespace', () => {
```
tobiasqueck, line 171: I am sure the `only` is not supposed to stay :)

kranthie-sap: Fixed.

tobiasqueck (dismissed):
Clean and simple change. Test cases nicely enhanced.
Thanks for the cleanup of the launch.json

kranthie-sap:
@tobiasqueck: Please re-review. I had to push an update to the snapshot file. Thanks.

tobiasqueck (approved):

PR #642, opened 2022-08-26
Fix XML escaping issue in deploy function

### LINKED ISSUE #641: BUG - SAP Fiori tools deployment to ABAP environment fails when there is special char in deploy description
### Related Feature

### Description
SAP Fiori tools deployment to ABAP environment fails when there is special char in deploy description

### Steps to Reproduce
1. In either VSCode or SBAS running SAP Fiori tools, generate a LROP project and generate deploy config for target environment `ABAP`
2. While generating deploy config in Fiori generator, provide a `deploy description` having a special character like `&` (See snapshot below)
3. Deploy
image

### Expected results

Deployment is success

### Actual results
Deploy error with status code 400 from backend

### Screenshots
image

OS:                      
- [X] Mac OS
- [ ] Windows
- [ ] Other

### Problem
The flow described above is using the `@sap-ux/axios-extension` to call the backed. The faulty functiona is at https://github.com/SAP/open-ux-tools/blob/main/packages/axios-extension/src/abap/ui5-abap-repository-service.ts#L62. 

### Fix
The description is used as part of the XML body, so we most likely need to encode the description so that it is valid XML value. However, for this issue we need to do manual tests with a real system to be sure that it is not a different layer e.g. something in ABAP that has issues with the special characters.

### Why was it missed
Corner-case that was simply overlooked.

### How can we avoid this

tobiasqueck:
Fixes #641


IainSAP (approved):

devinea (approved):
Thanks @tobiasqueck 
Changes and tests look good.

PR #649, opened 2022-08-31
Improving the simple generator example

tobiasqueck:
Improving the simple generator example
* fix: correctly handle url params
* feature: added usage of `store` module to securely store credentials

### DIFF
--- a/examples/simple-generator/src/app/service.ts
+++ b/examples/simple-generator/src/app/service.ts
@@ -84,15 +84,19 @@ async function storeCredentials(generator: Generator, provider: AbapServiceProvi
         }
     ]);
     if (storeCreds) {
-        const systemService = await getService<BackendSystem, string>({ entityName: 'system' });
-        const system = new BackendSystem({
-            name,
-            url: provider.defaults.baseURL!,
-            client: provider.defaults.params?.['sap-client'],
-            username: provider.defaults.auth?.username,
-            password: provider.defaults.auth?.password
-        });
-        await systemService.write(system);
+        try {
+            const systemService = await getService<BackendSystem, string>({ entityName: 'system' });
+            const system = new BackendSystem({
+                name,
+                url: provider.defaults.baseURL!,
+                client: provider.defaults.params?.['sap-client'],
+                username: provider.defaults.auth?.username,
+                password: provider.defaults.auth?.password
+            });
+            await systemService.write(system);
+        } catch (error) {
+            generator.log(`Couldn't store credentials. ${error}`);
+        }
     }
 }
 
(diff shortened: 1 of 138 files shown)


examples/simple-generator/src/app/ui5.ts
```diff
@@ -39,6 +39,9 @@ export type UI5SupportVersions = {
 
 let ui5VersionSupport: UI5SupportVersions;
 
+/**
+ *
+ */
```
unseen1980, line 44: Is this a todo?

examples/simple-generator/src/app/service.ts
```diff
@@ -39,6 +48,54 @@ export async function getServiceInfo(generator: Generator): Promise<ServiceInfo>
...
+    if (storeCreds) {
+        const systemService = await getService<BackendSystem, string>({ entityName: 'system' });
+        const system = new BackendSystem({
+            name,
+            url: provider.defaults.baseURL!,
+            client: provider.defaults.params?.['sap-client'],
+            username: provider.defaults.auth?.username,
+            password: provider.defaults.auth?.password
+        });
+        await systemService.write(system);
```
unseen1980, line 95: If for some reason the ```write``` fails, will the store return an error or should we handle it here as well?

tobiasqueck: I am not fully sure to be honest, however, you make a very good point here, we need to handle the possibility that something goes wrong. Storing the credentials is not critical for the generation process, therefore, I wrapped it in a `trycatch` and log the error and continue. https://github.com/SAP/open-ux-tools/pull/649/commits/7518ecc7b8cdfcf579312c8da1734ef823abcf73
Ok for you?

examples/simple-generator/src/app/ui5.ts
tobiasqueck: https://github.com/SAP/open-ux-tools/pull/649/commits/0931d94d9973bbdcac725c10ac85a80dff8fb61a

examples/simple-generator/src/app/service.ts
unseen1980: 👍

examples/simple-generator/src/app/ui5.ts
unseen1980: Thanks!

unseen1980 (approved):
I have tested locally but the code looks clean and easy to follow. Thanks, Tobias.

ullasholla (changes requested):
Thanks @tobiasqueck, looks good.
There's one minor change you might want to consider. Cheers

ullasholla (approved):
Thanks @tobiasqueck, looks good to me

tobiasqueck:
@unseen1980 @ullasholla could you please re-review?

ullasholla (approved):
Looks good, thanks

unseen1980 (approved):
👍

Issue #650, opened 2022-08-31
TBI - Improve FPM for consumers


IainSAP:
### Description (include screenshots)

Improves how FPM is integrated with consuming code and builds.

-  Pass the min ui5 version through to FPM when consuming from `@sap-ux/fiori-elements-writer`. This was missing.
-  Adds a new function `getTemplatePath` to FPM to maintain the relative path to the templates root when FPM is bundled and flattened ( specifically by webpack ). This simplifies bespoke bundling configuration for FPM.

### Value

Adds missing functionality and improves usability.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

PR #651, opened 2022-08-31
tbi(fe-fpm-writer): Updates to improve consumption

IainSAP:
#650 

Adds 2 improvements:

- Template handling is now relative to the root (./dist) folder to help when bundling
- Pass the ui5 version through when creating FPM from FE writer

### DIFF
--- a/packages/fe-fpm-writer/src/templates.ts
+++ b/packages/fe-fpm-writer/src/templates.ts
@@ -3,9 +3,8 @@ import { join } from 'path';
  * Locates template files relative to the dist folder.
  * This helps to locate templates when this module is bundled and the dir structure is flattened, maintaining the relative paths.
  *
- *
  * @param relativeTemplatePath - optional, the path of the required template relative to the ./templates folder. If not specified the root templates folder is returned.
- *
+ * @returns the path of the template specified or templates root folder
  */
 export function getTemplatePath(relativeTemplatePath?: string): string {
     return join(...[__dirname, '../templates'].concat(relativeTemplatePath || []));
--- a/packages/fe-fpm-writer/src/templates.ts
+++ b/packages/fe-fpm-writer/src/templates.ts
@@ -6,6 +6,6 @@ import { join } from 'path';
  * @param relativeTemplatePath - optional, the path of the required template relative to the ./templates folder. If not specified the root templates folder is returned.
  * @returns the path of the template specified or templates root folder
  */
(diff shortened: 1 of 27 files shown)


packages/fe-fpm-writer/src/templates.ts
```diff
@@ -0,0 +1,11 @@
+import { join } from 'path';
+/**
+ * Locates template files relative to the dist folder.
+ * This helps to locate templates when this module is bundled and the dir structure is flattened, maintaining the relative paths.
+ *
+ * @param relativeTemplatePath - optional, the path of the required template relative to the ./templates folder. If not specified the root templates folder is returned.
+ * @returns the path of the template specified or templates root folder
+ */
+export function getTemplatePath(relativeTemplatePath?: string): string {
+    return join(...[__dirname, '../templates'].concat(relativeTemplatePath || []));
```
tobiasqueck, line 10: That line looks more complicated then it has to be. Wouldn't the snippet below do the same?
```
export function getTemplatePath(relativeTemplatePath: string = ''): string {
    return join(__dirname, '../templates', relativeTemplatePath);
```

devinea (approved):
Thanks @IainSAP 
Changes look good.

packages/fe-fpm-writer/test/unit/section.test.ts
```diff
@@ -21,11 +21,11 @@ describe('CustomSection', () => {
             { version: 1.86, expected: join(root, 'section', '1.86') }
         ];
         test.each(testInput)('get root path of template', ({ version, expected }) => {
-            expect(getManifestRoot(root, version)).toEqual(expected);
+            expect(getManifestRoot(version)).toEqual(expected);
         });
         test('invalid version', () => {
             try {
-                getManifestRoot(root, 1.8);
+                getManifestRoot(1.8);
```
tobiasqueck, line 28: It is not a problem but why did you change this test?

tobiasqueck (approved):
One question, otherwise, everything looks good. Clean code and and easy to read.

packages/fe-fpm-writer/test/unit/section.test.ts
IainSAP: `getManifestRoot`, which gets the manifest template root, no longer requires the templates root path to be passed in as it internally uses the new `getTemplatePath` function.

tobiasqueck: Thanks!

PR #653, opened 2022-09-01
feat: Controller Extension support in fe-fpm-writer

815are:
Support for controller extension creation using `fe-fpm-writer`.

Some special parts to mention:
1. In `manifest.json` - for controller extension we can have:
1.1. Single controller using property `controllerName` - string;
1.2. Multiple controllers using property `controllerNames` - string array;
In result we need handle cases when `controllerName` is defined in `manifest.json` and we try to add new `controllerName`, then we need apply transition from `controllerName` to `controllerNames`.
Also additional property `overwrite` is available in case user wants to replace existing with new.
2. Property `extension` - https://github.com/SAP/open-ux-tools/blob/feat/controllerExtensions/packages/fe-fpm-writer/src/controller-extension/types.ts#L35:
```
extension: ControllerExtensionPageTarget | string;


export interface ControllerExtensionPageTarget {
    /**
     * The page type for which controller extension should be triggered.
     */
    pageType: ControllerExtensionPageType;
    /**
     * The unique page id for which controller extension should be triggered.
     */
    pageId?: string;
}
```
We can use object and provide `ListReport` or `ObjectPage` as `pageType`, but I noticed there is also other variants are valid like:
```
"sap.ui.controllerExtensions": {
   "project4.ext.view.Test": {
      "controllerName": "project4.ext.OPExtend3"
    }
}
...
"BookingSupplementTest": {
    "type": "Component",
    "id": "BookingSupplementTest",
    "name": "sap.fe.core.fpm",
    "options": {
        "settings": {
            "viewName": "project4.ext.view.Test",
            "entitySet": "BookingSupplement"
        }
    }
}
```
In result providing option to pass `extension` as string. In such case config object would look:
```
{
  extension: 'project4.ext.view.Test',
  name: 'NewExtension',
  folder: 'ext/controller',
}
```

If we use page type:
```
{
  extension: {
      pageType: ControllerExtensionPageType.ListReport
  },
  name: 'NewExtension',
  folder: 'ext/controller',
}
```

### DIFF
--- a/packages/fe-fpm-writer/src/controller-extension/index.ts
+++ b/packages/fe-fpm-writer/src/controller-extension/index.ts
@@ -200,7 +200,7 @@ export function generateControllerExtension(
     const ext = controllerConfig.typescript ? 'ts' : 'js';
     const viewPath = join(internalConfig.path, `${internalConfig.name}.controller.${ext}`);
     if (!fs.exists(viewPath)) {
-        fs.copyTpl(join(root, 'controller-extension/Controller.js'), viewPath, internalConfig);
+        fs.copyTpl(join(root, `controller-extension/Controller.${ext}`), viewPath, internalConfig);
     }
 
     return fs;
--- a/packages/fe-fpm-writer/src/controller-extension/index.ts
+++ b/packages/fe-fpm-writer/src/controller-extension/index.ts
@@ -172,16 +172,16 @@ export function generateControllerExtension(
     const root = join(__dirname, '../../templates');
 
     // merge with defaults
-    const completeController = enhanceConfig(controllerConfig, manifestPath, manifest);
+    const internalConfig = enhanceConfig(controllerConfig, manifestPath, manifest);
 
     // enhance manifest with view definition
-    const filledTemplate = render(fs.read(join(root, 'controller-extension', `manifest.json`)), completeController, {});
-    fs.extendJSON(manifestPath, JSON.parse(filledTemplate), getManifestReplacer(completeController));
+    const filledTemplate = render(fs.read(join(root, 'controller-extension', `manifest.json`)), internalConfig, {});
+    fs.extendJSON(manifestPath, JSON.parse(filledTemplate), getManifestReplacer(internalConfig));
 
     // add controller js file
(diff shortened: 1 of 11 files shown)


tobiasqueck:
@815are don't forget to also add a typescript template for the controller extension.

815are:
> @815are don't forget to also add a typescript template for the controller extension.

@tobiasqueck 
I checked but did not find valid usage of ui5 `override`, there is also issue reported for UI5 team regarding `Controller extension` in typescript - https://github.com/SAP/ui5-typescript/issues/332
In result I have interface as:
```
export interface ControllerExtension extends CustomElement {
    ...
    /**
     * Typescript generation is not supported for control extensions.
     */
    typescript?: false;
}
```

nlunets:
The link from @815are is correct, currently there is no official support for this.
You can still write them in TS i think but not with class syntax .

packages/fe-fpm-writer/src/controller-extension/index.ts
```diff
@@ -0,0 +1,207 @@
...
+    // Validate the base and view paths
+    if (!fs) {
+        fs = create(createStorage());
+    }
+    validateBasePath(basePath, fs);
+
+    const manifestPath = join(basePath, 'webapp/manifest.json');
+    const manifest = fs.readJSON(manifestPath) as Manifest;
+
+    const root = join(__dirname, '../../templates');
```
tobiasqueck, line 190: With https://github.com/SAP/open-ux-tools/commit/a1dc0698b66fa8fbfcf7523488c0b5f1afc440f8 we have a common function to access the template folders (makes it easier to integrate it when webpacking). Please merge from main and use it instead of determining root here.

tobiasqueck: Thanks for [6071127](https://github.com/SAP/open-ux-tools/pull/653/commits/6071127d76c268b0938e66a3550eaa2b63586172)

tobiasqueck (approved):
* code adheres code quality standards
* test coverage is good
* manual tests successful

zdravko-georgiev (approved):

PR #661, opened 2022-09-05
feat(sap-ux/ui-components): ui-components for web

Jimmy-Joseph19:
Added new package @sap-ux/ui-components


package.json
```diff
@@ -52,7 +52,10 @@
             "follow-redirects": ">=1.14.8",
             "tmpl": "^1.0.5",
             "vm2@<3.9.6": "^3.9.6",
-            "minimist": "^1.2.6"
+            "minimist": "^1.2.6",
+            "glob-parent@<5.1.2": ">=5.1.2",
```
tobiasqueck, line 56: Please provide an explanation for every single override that you want to add. Explain what the issue is and why it is not possible to use a newer version of dependency directly.
The goal is to have zero overrides. Each override is considered an ugly exception.

packages/ui-components/jest.config.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
```
tobiasqueck, line 1: The jest config is significantly different from all other projects e.g. `fe-fpm-writer/jest.config.js`. In some cases, it would make sense to move some of the settings to the root `tsconfig` e.g. adding tsx to `collectCoverageFrom` and other cases I'd like to understand why the settings are needed.

packages/ui-components/package.json
```diff
@@ -0,0 +1,73 @@
...
+        "@types/react-virtualized": "9.21.11",
+        "@types/sanitize-html": "2.3.2",
+        "@types/uuid": "3.4.7",
+        "babel-jest": "^24.8.0",
+        "babel-loader": "8.1.0",
+        "copyfiles": "2.1.1",
+        "css-loader": "^5.2.6",
+        "enzyme": "3.11.0",
+        "enzyme-adapter-react-16": "1.15.2",
+        "eslint": "8.21.0",
```
tobiasqueck, line 55: `eslint`, `typescript` and friends are already listed as devDependencies in the root, please remove them

packages/ui-components/src/components/UIButton/UIIconButton.tsx
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Initializes component properties.
+     * @param {ButtonProps} props
+     */
+    public constructor(props: ButtonProps) {
+        super(props);
+    }
+
+    /**
+     * Method which returns button interaction background with including fallback for olde VSCode or BAS versions.
```
tobiasqueck, line 34: Is this still required? please check which exact versions require the fallback and check if we still support them

packages/ui-components/src/utilities/Focus.ts
```diff
@@ -0,0 +1,5 @@
+import { getFirstFocusable, getLastFocusable, setFocusVisibility } from '@fluentui/react';
```
tobiasqueck, line 1: Why do we need the re-export under a different name? It is especially confusing to me that the functions are renamed so that they sound like objects.

tobiasqueck (changes requested):
I added a bunch of comments on a general level. I haven't looked closer at the actual components

packages/ui-components/src/components/UIButton/UIIconButton.tsx
815are: as I checked, looks like we still need it. `--vscode-toolbar-hoverBackground` was presenter at VSCode version about ~`1.56`.
But as I checked variable `--vscode-toolbar-hoverBackground` still not presenter in BAS environment:
image
In result fallback is required for BAS

packages/ui-components/src/utilities/Focus.ts
Jimmy-Joseph19: It was to avoid import from @fluentui in extensions. Would renaming to `getUIFirstFocusable` or exporting same with same name `getFirstFocusable ` fine?

tobiasqueck: Gotcha, so you are kind of facading it. Yes, I'd prefer `getUIFirstFocusable`

packages/ui-components/src/components/UIButton/UIIconButton.tsx
tobiasqueck: Thanks

packages/ui-components/package.json
Jimmy-Joseph19: when eslint and typscript are removed we get following error in `pnpm install`
image

Jimmy-Joseph19: when eslint and typscript are removed we get following error in `pnpm install`
image

tobiasqueck: ugh. Ok, then keep them but, please, make sure that the versions are the same as in the root `package.json`

package.json
Jimmy-Joseph19: We have 3 overrides 

1.` glob-parent(3.1.0)`
2.` trim: 0.0.1`
3.`trim-newlines 3.0.1`
Which happens due to the dependency of storybook packages. We are currently using latest(stable) version of storybook available. So without overrides we have folowing issues.

image

Would it be fine to maintain the issue in image in file [version-overrides.md](https://github.com/SAP/open-ux-tools/blob/main/docs/version-overrides.md)

Jimmy-Joseph19: We have 3 overrides 

1.` glob-parent(3.1.0)`
2.` trim: 0.0.1`
3.`trim-newlines 3.0.1`
Which happens due to the dependency of storybook packages. We are currently using latest(stable) version of storybook available. So without overrides we have folowing issues.

image

Would it be fine to maintain the issue in image in file [version-overrides.md](https://github.com/SAP/open-ux-tools/blob/main/docs/version-overrides.md)

Jimmy-Joseph19: sure thanks

packages/ui-components/jest.config.js
Jimmy-Joseph19: I have adapted jest-config based on the other config file. Could you review it again. tsx file added to collect to get the coverage for react components test files.

Jimmy-Joseph19: I have adapted jest-config based on the other config file. Could you review it again. tsx file added to collect to get the coverage for react components test files.

Jimmy-Joseph19:
Closing this PR as New PR created https://github.com/SAP/open-ux-tools/pull/671

Issue #372, opened 2022-03-04
BUG - ui5-application-writer: setting `typesVersion` has no effect on outputs


IainSAP:
### Description
The `UI5` interface property `typesVersion` is set but not referenced in any outputs and so has no bearing on the value written in the template file here : https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-application-writer/templates/optional/codeAssist/package.json#L5

### Steps to Reproduce
Set the ui5 `localVersion` when calling `ui5-application-writer` to be <= '1.76' and enable the option `codeAssist`. The resulting package.json should contain an entry:
```
"@sapui5/ts-types": "1.71.18"
``` 
but instead always contains version "1.92.2"

The same problem happens when using the option `typescript`. There always `latest` is used for module `@sapui5/ts-types-esm`.

### Expected results
The correct types version is set.

### Actual results
Its hardcoded to "1.92.2"

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #670, opened 2022-09-08
Fix: typeVersion has no effect on outputs

tobiasqueck:
This PR contains
* #372 with the addition of generating the `sap.fe.d.ts` file depending on the types version
* fix of FE types for lower version (`<1.102`)
* better typed generated FPM code snippets


tobiasqueck:
@nlunets would you please review the improved ts templates and type enhancements for v4 in `fe-fpm-writer`

@devinea @IainSAP would you please review.

packages/ui5-application-writer/src/data/defaults.ts
```diff
@@ -62,6 +62,7 @@ export const enum UI5_DEFAULT {
     OPENUI5_CDN = 'https://openui5.hana.ondemand.com',
     TYPES_VERSION_SINCE = '1.76.0',
     TYPES_VERSION_PREVIOUS = '1.71.18',
+    TYPES_VERSION_BEST = '1.102.7',
```
IainSAP, line 65: Could we have some comments to explain what `BEST` means as its a qualitative term, better than ...?

IainSAP (dismissed):
Thanks for the fix. One small query.

- Fix for types version looks good
- Snapshots updated as expected

packages/ui5-application-writer/src/data/defaults.ts
tobiasqueck: Good point, I am also happy to rename the entry to something that fits better. Any suggestions?

What I mean with `BEST` here is, that this is the latest version that will receive downports. Let's assume we have an issue in UI5 version 1.106, then we will fix it with the next available version e.g. 1.108 but if the issue is important, we will also downport it to 1.102 as long as it is supported (Q1 2023). Afterwards, we switch this variable to the new `BEST` version. This is also aligned with our fix strategy for `@sap/ux-specification`

nlunets (dismissed):

packages/ui5-application-writer/src/data/defaults.ts
```diff
@@ -110,20 +111,37 @@ export function mergeUi5(ui5: Partial<UI5>): UI5 {
...
+    merged.typesVersion = ui5.typesVersion ?? getTypesVersion(merged.minUI5Version);
     merged.ui5Theme = ui5.ui5Theme ?? 'sap_fiori_3';
     merged.ui5Libs = getUI5Libs(ui5.ui5Libs);
 
     return Object.assign({}, ui5, merged) as UI5;
 }
 
+/**
+ * Get the best types version for the given minUI5Version.
+ * For the latest versions the lates S/4 on-premise version (1.102.x) is used, for anything before we match the versions as far back as available.
```
devinea, line 125:
```suggestion
 * For the latest versions the LTS S/4 on-premise version (1.102.x) is used, for anything before we match the versions as far back as available.
```

packages/ui5-application-writer/src/data/defaults.ts
```diff
@@ -110,20 +111,37 @@ export function mergeUi5(ui5: Partial<UI5>): UI5 {
...
+ * @returns semantic version representing the types version.
+ */
+export function getTypesVersion(minUI5Version?: string) {
+    const version = semVer.coerce(minUI5Version);
+    if (!version) {
+        return `~${UI5_DEFAULT.TYPES_VERSION_BEST}`;
+    } else {
+        if (semVer.gte(version, UI5_DEFAULT.TYPES_VERSION_BEST)) {
+            return `~${UI5_DEFAULT.TYPES_VERSION_BEST}`;
+        } else {
```
devinea, line 137: What about if the version is greater than 1.108.0. Shouldn't that version be used rather than 1.102.x?

packages/ui5-application-writer/templates/optional/codeAssist/package.json
```diff
@@ -2,6 +2,6 @@
     "devDependencies": {
         "eslint": "7.32.0",
         "@sap/eslint-plugin-ui5-jsdocs": "2.0.5",
-        "@sapui5/ts-types": "1.102.x"
+        "@sapui5/ts-types": "<%- ui5.typesVersion %>"
```
devinea, line 5: It is intentional to update codeAssist? 
I thought the types version calculation related to `@sapui5/ts-types-esm`

devinea (approved):
Generally all looks good. 
Few small questions.

packages/ui5-application-writer/templates/optional/codeAssist/package.json
tobiasqueck: Both are in sync and are released at the same time based on the same sources.

packages/ui5-application-writer/src/data/defaults.ts
tobiasqueck: the idea was that if in 1.109 something is missing, then we also downport it to the `BEST` version, so for the types, the `BEST` version should be your best shot.

devinea: OK. Interesting. Fine for now but I assume that logic will need to reviewed at some point

devinea (approved):

packages/ui5-application-writer/src/data/defaults.ts
IainSAP: Thx I see you added a comment here: https://github.com/SAP/open-ux-tools/pull/670/files#diff-0f207248944a1732a5b7343b9d1bb29e22bbae5fa2acce7a573a344c04c856d2R123

nlunets (approved):

Issue #673, opened 2022-09-12
BUG -  @sap-ux/fiori-elements-writer : Both `.ts` and `.js` files included by copy template


IainSAP:
### Description

The issue occurs due to . (dot) in the path to the template. This occurs under certain conditions depending on the path location where the module (`@sap-ux/fiori-elements-writer`) is installed. Adding the `dot: true`, `globOption` to the `copyTpl` call addresses the issue by including these files in `globOption` patterns (inc. negative patterns such as ignore option : **/*.ts).

The issue typically occurs when used to copy templates from installed locations using nvm (.nvm) or, in Business Application Studio , where the manual install location is `.node_modules_global`

### Steps to Reproduce
Install the module to a location that includes a dot (`.`) in the path.

### Expected results
Only one of `.ts` or `.js` template files are included, depending on passed options.

### Actual results
Both `.ts` and `.js` template files are included

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

tobiasqueck:
Wow. That is an interesting finding. Do you have a proposal on how to solve it? I played a little with it at https://www.digitalocean.com/community/tools/glob?comments=true&glob=%2A%2A%2F%2A.js&matches=false&tests=%2F%2F%20This%20will%20match%20as%20it%20ends%20with%20%27.js%27&tests=%2Fhello%2Fworld.js&tests=%2F%2F%20it%20should%20match%20but%20doesn%27t&tests=%2Fhello%2F.dotdir%2Fword.js and I didn't find one.

PR #674, opened 2022-09-12
https://github.com/SAP/open-ux-tools/issues/673

IainSAP:
Fix for: https://github.com/SAP/open-ux-tools/issues/673

### DIFF
--- a/.changeset/quiet-gifts-taste.md
+++ b/.changeset/quiet-gifts-taste.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fiori-freestyle-writer': patch
+---
+
+Fix for https://github.com/SAP/open-ux-tools/issues/673. Include dot files in negative pattern when copying templates.


tobiasqueck (changes requested):
Could you please apply the same fix at https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-freestyle-writer/src/index.ts#L36

tobiasqueck (dismissed):
Fix looks good, however, I don't know how to properly test it. 
Any idea how to put this in a unit test? 

If not, I am also ok with it.

IainSAP:
> Fix looks good, however, I don't know how to properly test it. Any idea how to put this in a unit test?
> 
> If not, I am also ok with it.

Possibly by mounting the test `fs` to include a path with a dot. Will create a follow up PR if thats ok since this fix is needed as priority?

devinea (commented):
@IainSAP Does this need to be applied to all `copyTpl` calls to be safe?

IainSAP:
> @IainSAP Does this need to be applied to all `copyTpl` calls to be safe?

I dont know as that requires more extensive testing and the fix is urgent. The bug is fixed by this PR for now. A follow-up may be required with additional testing.

devinea (approved):
I am OK to merge this much for now.

IainSAP:
Follow up item created: https://github.com/SAP/open-ux-tools/issues/675

tobiasqueck (approved):

Issue #676, opened 2022-09-12
BUG - fiori-elements-writer generates typescript app with missing files (`changes_loader.js/changes_preview.js`)


IainSAP:
### Description
Generating an LROP typescrtipt app setting the ui5 version to '1.78.0' or less. The generated app should allow local flex changes to be applied. The required files `changes_loader.js` and `changes_preview.js` are missing but referenced from the `flpSandbox.html` file.

### Steps to Reproduce
Run f-e-w unit test `lropV2_flex_changes` adding the option to generate a typescritp app. Notice that the files are missing from the output.

### Expected results

Local changes can be applied by using the test sandbox.

### Actual results

The required files are missing and an error is generated loading the sandbox file.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #677, opened 2022-09-12
fix(fe-writer) Add missing JS files(changes_loader and changes_preview ) as ts files

devinea:
fix for #676

Because of the ignore here 
https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-elements-writer/src/index.ts#L102
the js files are not copied. Adding a TS copy of the files so that the copy works. 
transpiling middleware removes the comments but otherwise seems ok. 
Added a test to cover the case

### DIFF
--- a/packages/fiori-elements-writer/test/lrop.test.ts
+++ b/packages/fiori-elements-writer/test/lrop.test.ts
@@ -100,13 +100,13 @@ describe(`Fiori Elements template: ${TEST_NAME}`, () => {
         {
             name: 'lropV2_set_toolsId',
             config: {
-                ...Object.assign(feBaseConfig('felrop5'), {
+                ...Object.assign(feBaseConfig('lropV2_set_toolsId'), {
                     template: {
                         type: TemplateType.ListReportObjectPage,
                         settings: v2TemplateSettings
                     },
                     app: {
-                        ...feBaseConfig('felrop4').app,
+                        ...feBaseConfig('lropV2_set_toolsId').app,
                         sourceTemplate: {
                             version: '1.2.3-test',
                             id: 'test-fe-template',
@@ -120,13 +120,13 @@ describe(`Fiori Elements template: ${TEST_NAME}`, () => {
         {
             name: 'lropV2_set_toolsId_only',
             config: {
-                ...Object.assign(feBaseConfig('felrop5'), {
+                ...Object.assign(feBaseConfig('lropV2_set_toolsId_only'), {
                     template: {
                         type: TemplateType.ListReportObjectPage,
                         settings: v2TemplateSettings
                     },
                     app: {
-                        ...feBaseConfig('felrop4').app,
+                        ...feBaseConfig('lropV2_set_toolsId_only').app,
                         sourceTemplate: {
                             toolsId: 'toolsId:1234abcd'
                         }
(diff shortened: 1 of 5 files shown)


IainSAP (approved):
Thx @devinea

tobiasqueck (changes requested):
Please fix the type errors (`Found 46 errors in 3 files.`). To see them just run in `packages/fiori-elements-writer`
```
UX_DEBUG=x pnpm test
cd test/test-output/lropTemplates/lropV2_ts
npm i
npm run ts-typecheck
```

packages/fiori-elements-writer/test/lrop.test.ts
```diff
@@ -135,6 +135,26 @@ describe(`Fiori Elements template: ${TEST_NAME}`, () => {
                 service: v2Service
             } as FioriElementsApp<LROPSettings>
         }
+        ,
+        {
+            name: 'lropV2_ts',
+            config: {
+                ...Object.assign(feBaseConfig('felrop4'), {
```
tobiasqueck, line 142: Can you change this one to 'lropV2_ts' as well. (not critical but confused me when looking at the generated project).

packages/fiori-elements-writer/test/lrop.test.ts
```diff
@@ -135,6 +135,26 @@ describe(`Fiori Elements template: ${TEST_NAME}`, () => {
...
+                    template: {
+                        type: TemplateType.ListReportObjectPage,
+                        settings: v2TemplateSettings
+                    },
+                    ui5: {
+                        ...feBaseConfig('felrop4'),
```
tobiasqueck, line 148: Line 148 is not clear to me.

packages/fiori-elements-writer/test/lrop.test.ts
devinea: Done

packages/fiori-elements-writer/test/lrop.test.ts
devinea: Removed

devinea:
> Please fix the type errors (`Found 46 errors in 3 files.`). To see them just run in `packages/fiori-elements-writer`
> 
> ```
> UX_DEBUG=x pnpm test
> cd test/test-output/lropTemplates/lropV2_ts
> npm i
> npm run ts-typecheck
> ```

Screenshot 2022-09-14 at 20 54 34

It shows one error. I think you mentioned this is missing in the types module so we need an updated version with that added?
OK to merge in the meantime or we should add ts ignore / ts no check in that case for now?

tobiasqueck:
> It shows one error. I think you mentioned this is missing in the types module so we need an updated version with that added?
> OK to merge in the meantime or we should add ts ignore / ts no check in that case for now?

Yes, I have reached out to the corresponding team already and hope it get fixed soon.

tobiasqueck (approved):
I have to admit, I expected a different fix than `// @ts-nocheck` but since these files are rather static, I am ok with it.

IainSAP (approved):
Thx @devinea 

- New files added as expected
- Tests to cover

devinea:
> I have to admit, I expected a different fix than `// @ts-nocheck` but since these files are rather static, I am ok with it.

I looked into adding `exclude` to the `tsconfig` for `webapp/test` like we discussed and this worked ok for `npm run ts-typecheck`. But in VS Code it showed various TS errors in these files. `// @ts-nocheck`  gives a consistent result.

I also looked into converting these files to full TS, but there was various errors with types not exposed so seem safer to ignore them for now. I think these will be removed eventually when middleware takes care of this for V2 projects like in V4.

Issue #683, opened 2022-09-19
TBI - FPM: add required macro namespace


IainSAP:
### Description (include screenshots)
In order for custom code to be added to a generated FPM app, this namespace should be in the `main.view.xml` for the generated FPM app.  Although guided development will add the NS if not present, it was agreed that it makes sense that the generator adds it.

`View.xml` would be updated:

https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/templates/page/custom/1.94/ext/View.xml

with the entry

`xmlns:macros="sap.fe.macros"`

### Value
Supports FPM app building

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Add namespace

PR #684, opened 2022-09-19
tbi(fe-fpm-writer):683 Adds required namespace

IainSAP:
Addresses: #683


devinea (approved):
Changes look good and tests updated

tobiasqueck (approved):

Issue #293, opened 2022-01-29
FEATURE - Provide modules for project constants and project access


Klaus-Keller:
_As a **developer**, I want to **have API access to SAP Fiori projects**, so that I **can read and update project resources**_

### Description
Multiple tools and modules require functionality to 
- find SAP Fiori projects in a folder/workspace
- access project resources, such as the AppDescriptor (`manifest.json`), local annotations, etc. 

There should be a module that provides a common way to find projects and access resources of them programmatically. Constants and types should also be available as typescript module, which can be same as the module that provides access to a project, or provided as separate module.

### Technical Design
Create module(s) for project types and constants and access to project resources. Other tools or apps can add these modules as dependencies and use exposed functionality.

### Acceptance Criteria
I have a workspace and execute search using the new project access module then the modules returns me a list of SAP Fiori projects and allows accessing resources of them.

### Tasks
- [x] Add module for project constants and project access 
- [x] Unit Tests added
- [x] Test Cases defined

Klaus-Keller:
Published: https://www.npmjs.com/package/@sap-ux/project-access
Closing this ticket.

PR #685, opened 2022-09-19
feat:project access, project types

Klaus-Keller:
## Issue https://github.com/SAP/open-ux-tools/issues/293
Adding new module

### `@sap/ux-project-access`
Functions to find and read SAP UX projects.


packages/project-types/src/constants.ts
```diff
@@ -0,0 +1,6 @@
+export enum FileName {
+    Manifest = 'manifest.json',
+    Package = 'package.json',
+    Ui5Yaml = 'ui5.yaml',
+    Ui5LocalYaml = 'ui5-local.yaml'
```
tobiasqueck, line 5: `ui5-mock.yaml`?

packages/project-types/src/package.ts
```diff
@@ -0,0 +1,19 @@
...
+    dependencies?: { [dependencyName: string]: string };
+    devDependencies?: { [dependencyName: string]: string };
+    optionalDependencies?: object;
+    scripts?: { [scriptName: string]: string };
+    ui5?: object;
+    remarkConfig?: object;
+    version?: string;
+}
+
+export const enum UI5FlexLayer {
```
tobiasqueck, line 16: What is the advantage of using an enum instead of `type UI5FlexLayer = 'VENDRO' | 'CUSTOMER_BASE'`?

packages/project-types/src/package.ts
```diff
@@ -0,0 +1,19 @@
+export interface Package {
```
tobiasqueck, line 1: It's odd that we define the package json type. Should we use something existing e.g. https://www.npmjs.com/package/type-fest and only extend it with UI5 / Fiori tools specific things?

packages/project-access/src/project/dependencies.ts
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Helper to check for dependency/devDependency.
+ *
+ * @param packageJson - content of package.json to check
+ * @param dependency - name of the dependency
+ * @returns - true: has dependency; false: no dependency
+ */
+export const hasDependency = (packageJson: Package, dependency: string): boolean =>
+    (typeof packageJson.dependencies === 'object' && packageJson.dependencies[dependency] !== undefined) ||
```
tobiasqueck, line 11: What are we doing here? Wouldn't `packageJson.dependencies?.[dependency]` do the very same?

packages/project-access/src/project/cap.ts
```diff
@@ -0,0 +1,118 @@
...
+    return fileExists(join(projectRoot, 'srv', 'src', 'main', 'resources', 'application.yaml'));
+}
+
+export interface CapCustomPaths {
+    app: string;
+    db: string;
+    srv: string;
+}
+
+interface CdsEnvFolders {
```
tobiasqueck, line 54: Shouldn't that also move to the `project-types` project?

packages/project-access/src/i18n.ts
```diff
@@ -0,0 +1,35 @@
+import type { TOptions } from 'i18next';
```
tobiasqueck, line 1: What does this have to do with a module called `project-access`?

tobiasqueck:
Aside from my other comments, there is one major problem with the `project-access` module as it is: the file access. All modules in the `open-ux-tools` use `mem-fs` to access the file system so that we can also work on changes that are not yet committed to the filesystem which is very important when you use the writers in a generator like module. 
As result, I would like to see the same abstraction for the `project-access` module that it is also usable for the writers.

packages/project-access/src/project/getWebapp.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Get path to webapp.
+ *
+ * @param projectRoot - root path, where package.json or ui5.yaml is
+ * @returns - path to webapp folder
+ */
+export async function getWebappPath(projectRoot: string): Promise<string> {
+    let webappPath = join(projectRoot, 'webapp');
+    const ui5YamlPath = join(projectRoot, FileName.Ui5Yaml);
+    if (await fileExists(ui5YamlPath)) {
+        const yamlDoc = await readYAML<PartialUi5Yaml>(ui5YamlPath);
```
tobiasqueck, line 19: please use the `ui5-config` module to access `ui5*.yaml` files. If some functionality is missing, please add it there.

packages/project-access/src/project/getWebapp.ts
```diff
@@ -0,0 +1,26 @@
...
+        const relativeWebappPath = yamlDoc?.resources?.configuration?.paths?.webapp;
+        if (relativeWebappPath) {
+            webappPath = join(projectRoot, ...relativeWebappPath.split(posix.sep));
```
tobiasqueck, line 22: `webappPath = join(projectRoot, relativeWebappPath);` should result in exactly the same

packages/project-access/src/translations/project-access.i18n.json
```diff
@@ -0,0 +1,6 @@
+{
+    "error": {
+        "projectRootNotFound": "Could not find any project root for {{-path}}. Parameter 'sapuxRequired' was {{sapuxRequired}}",
```
tobiasqueck, line 3: `sapuxRequired` is a property in your function that is not known to a human being using anything consuming this module, so I don't think that this is a proper error message that a user should ever see. And, if it isn't intended to be one, why does it need to be translated?

packages/project-access/src/translations/project-access.i18n.json
```diff
@@ -0,0 +1,6 @@
...
+        "moduleNotInstalled": "Module {{moduleName}} not installed in project {{-projectRoot}}"
```
tobiasqueck, line 4: Similar to the other translatable string: is this a message that is ever intended to be shown in a user interface? If it is just for logging, then not translation is required and if it is to be handled by the consumer, maybe a designated Error type would be a better option.

packages/project-access/src/i18n.ts
tobiasqueck: I think you misunderstood my comment. I meant why do we need this here? `project-access` is a low level module to be used by others modules to simplify project access, so why would we have any translatable strings in here?

packages/project-access/src/project/cap.ts
Klaus-Keller: While `CdsEnvFolders` is used local in this file only, you are right for `CapCustomPaths`. I move it to `project-types`.

packages/project-access/src/project/dependencies.ts
Klaus-Keller: I'll change it

packages/project-access/src/project/getWebapp.ts
Klaus-Keller: My understanding was that `ui5-config` is used to modify yaml files. In this case I just read it. Another thought I had was: it might be that at some point `ui5-config` requires some functionality from this module and in this case we would create circular dependencies. What do you think?

packages/project-access/src/translations/project-access.i18n.json
Klaus-Keller: I guess it is a bit of a grey zone. How would you classify the OUTPUT view in VSCode:
image

All texts in there used to be translated. You want to change this?

packages/project-access/src/translations/project-access.i18n.json
Klaus-Keller: same as above, we need a decision on whether text in OUTPUT need translation. But keep in mind the previously decision was to translate.

packages/project-types/package.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "@sap-ux/project-types",
```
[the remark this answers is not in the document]
Klaus-Keller: No, it provides `enum` values that can't be imported as `devDependency`

packages/project-types/src/constants.ts
Klaus-Keller: Added the ones that are used only and plan to extend them as we go.

packages/project-access/src/translations/project-access.i18n.json
tobiasqueck: This is logging information - quite technical information, so I would not expect it to be translated.

packages/project-types/src/webapp/manifest.ts
```diff
@@ -0,0 +1,2960 @@
+/* tslint:disable */
```
tobiasqueck, line 1: We have the same file already at https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-config/src/types/manifest.ts.
I fully agree that it should be here, but, please also delete the old one, and make sure that all consumers use your new `project-types` module

Klaus-Keller: @tobiasqueck, agree, I'd like to align this with you. There is also an attempt to get this directly from AppDescriptor: https://github.wdf.sap.corp/AppDescriptor/manifest/pull/424

packages/project-types/src/package.ts
ullasholla: * The enum values are meant to be opaque to the consumer, so they don't really care what `UI5FlexLayer.VENDOR` actually resolves to
* We can iterate through the value of an Enum if we want to use that in a validator, for example
* The enum name `UI5FlexLayer` acts as sort of a namespace compared to the unadorned string `VENDOR`
* The transpiled code is going to be slightly more verbose compared to string types. If performance is a problem, it's probably better to profile first to see if this transpilation is a problem. The other considerations probably have more weight

In the end, I think choosing one over the other comes down to coding style

packages/project-types/src/package.ts
Klaus-Keller: @tobiasqueck, I changed this to use type from `PackageJson` `type-fest` and extend it with our custom properties. As I have it as a `devDependency` and import it using 

```
import type { PackageJson } from 'type-fest';
```

it should not add anything it terms of size to `@sap-ux/project-types`. However, I've checked `type-fest` by installing it into an empty folder and it takes ~400KB, what concerns me slightly, as we replace 7 lines of code in our type definition. We could think of just adding a file like https://github.com/sindresorhus/type-fest/blob/main/source/package-json.d.ts which contains just the type for `package.json`, but then as I said, it doesn't really affect the resulting module size, it just uses some resources for developers.

packages/project-access/src/project/getWebapp.ts
Klaus-Keller: Changed this to use `ui5-config`

packages/project-access/src/project/getWebapp.ts
Klaus-Keller: fixed

packages/project-types/src/webapp/manifest.ts
Klaus-Keller: Changed it and updated all consumers

packages/ui5-test-writer/package.json
```diff
@@ -31,16 +31,16 @@
         "!dist/**/*.map"
     ],
     "dependencies": {
+        "@sap-ux/project-types": "workspace:*",
```
tobiasqueck, line 34: It is just a type import, therefore, it should move to the `dev-dependencies`

packages/ui5-proxy-middleware/package.json
```diff
@@ -31,6 +31,7 @@
     ],
     "dependencies": {
         "@sap-ux/logger": "workspace:*",
+        "@sap-ux/project-types": "workspace:*",
```
tobiasqueck, line 34: devDependency please

packages/ui5-application-writer/package.json
```diff
@@ -32,18 +32,19 @@
         "!dist/**/*.map"
     ],
     "dependencies": {
+        "@sap-ux/project-types": "workspace:0.0.1",
```
tobiasqueck, line 35: devDependency please

packages/fe-fpm-writer/package.json
```diff
@@ -29,6 +29,7 @@
         "!dist/**/*.map"
     ],
     "dependencies": {
+        "@sap-ux/project-types": "workspace:*",
```
tobiasqueck, line 32: devDependency please

packages/project-types/src/constants.ts
tobiasqueck: The `odata-service-writer` and `ui5-test-writer` check for `ui5-mock.yaml`, so we have a use case here.

packages/project-types/package.json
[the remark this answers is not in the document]
tobiasqueck: If the project does more than types then I find the project name misleading and we need to discuss it again.
Btw, is there any `enum` that you cannot export as `const enum`? Wouldn't this solve the problem?

packages/project-types/src/package.ts
tobiasqueck: > In the end, I think choosing one over the other comes down to coding style

I agree with you here because I doubt that perfomance is an issue. I have started to use simple types if I know the strings are constant, I only use an enum if I know the value can be changed in the future. In this case, I doubt that the values change, so I would switch to the simple type BUT as you said, it is coding style, so no need to change, just look at my argument again, and then decide :)

packages/project-types/src/package.ts
tobiasqueck: I am also ok with copying (but don't forget to list it in https://github.com/SAP/open-ux-tools/blob/feat/293/project-access/.reuse/dep5. 
Otherwise, I think the `type-fest` module is pretty interesting, and I am sure we find my uses for it.

packages/project-access/src/translations/project-access.i18n.json
tobiasqueck: @kranthie-sap tried using `Change Display Language` in the command palette to change the display language. It looks like VS Code also doesn’t translate the log messages in the output channels, so I'd recommend we align with VSCode here.

.github/CODEOWNERS
```diff
@@ -15,6 +15,12 @@
 # Module: logger
 /packages/logger/ @SAP/ux-tools-logger
 
+# Module: project-access
+/packages/project-access/ @SAP/ux-tools-project-access
```
tobiasqueck, line 19: Could we have one team for both modules? We have started doing this also for other modules where the same persons will be responsible for multiple modules.

I would recommend `@SAP/ux-tools-project-access` for both

tobiasqueck (changes requested):
Thanks for the changes. Here is the 2nd round of reviews

packages/project-types/package.json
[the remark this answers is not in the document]
Klaus-Keller: I did not find any way to do an `import type` of an `enum` or `const enum` and use it's value. If you find a way I'm very happy to use it. 
If we do not find a way and want to use the module as pure type only, we need to remove the enums.

packages/project-types/src/constants.ts
Klaus-Keller: added it

packages/project-types/src/package.ts
Klaus-Keller: Given the fact that 
a) the value here is very very unlikely to be changed
b) `const enum` are more problematic than string literals 

I've decided to change this to

```
export type UI5FlexLayer = 'VENDOR' | 'CUSTOMER_BASE';
```

Klaus-Keller:
> Aside from my other comments, there is one major problem with the `project-access` module as it is: the file access. All modules in the `open-ux-tools` use `mem-fs` to access the file system so that we can also work on changes that are not yet committed to the filesystem which is very important when you use the writers in a generator like module. As result, I would like to see the same abstraction for the `project-access` module that it is also usable for the writers.

I propose that `@sap-ux/project-access` does not provide means to write directly to projects on filesystem. Whenever write is required, we should use or provide an appropriate `@sapux/*-writer` module. I support the idea to use package `@sap-ux/project-access` to retrieve information from a project that partially or completely exists in `mem-fs`.

.github/CODEOWNERS
Klaus-Keller: I really liked the fact that we had a 1:1 mapping of team to module. Many things can change and to me 1:1 seems to be the most flexible, clean, and future proof choice. 
However, we can also start with one team for both and split it as needed. 
Changed it for now.

packages/project-types/package.json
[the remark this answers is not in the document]
tobiasqueck: My understanding is that you need a `const enum`, still need to do `import` (without type) but you can have the module as `devDependency` because during the transpilation the enums are inlined.

This is a topic for which we need the knowledge of @ullasholla

packages/project-access/src/i18n.ts
tobiasqueck: Resolving this question because it is discussed at https://github.com/SAP/open-ux-tools/pull/685#discussion_r975145025

packages/project-access/src/translations/project-access.i18n.json
tobiasqueck: Discussed at https://github.com/SAP/open-ux-tools/pull/685#discussion_r975145025.
Note: the text was not translated in the Fiori tools, it was ready for translation.

packages/ui5-proxy-middleware/package.json
tobiasqueck: Thanks for also removing the other not needed dependency

packages/project-types/package.json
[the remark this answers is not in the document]
ullasholla: > My understanding is that you need a const enum, still need to do import (without type) but you can have the module as devDependency because during the transpilation the enums are inlined.

This is my understanding as well. You may have to play around with the bundling to check if it actually works as we expect it.

If we do stick with const enums, does this mean that we'll need ban non-const enums in `project-types`? Maybe using something like this in `project-types`: https://github.com/typescript-eslint/typescript-eslint/issues/561#issuecomment-593059472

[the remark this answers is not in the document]
tobiasqueck: I like the idea of adding that rule here.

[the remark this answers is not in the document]
Klaus-Keller: **It works!** thanks @tobiasqueck  and @ullasholla 

_Here is what I've done_
Changed to `const enum` in module `@sap-ux/project-types`  
-> build `@sap-ux/project-types`

Changed `@sap-ux/project-access` to consume `@sap-ux/project-types` as `devDependecy`, then imported enum with

```
import { FileName } from '@sap-ux/project-types';
..
const manifest = await readJSON<Manifest>(join(manifestPath, FileName.Manifest));
```

-> build and check in `dist`, the const enum values have been replaced with string, e.g.

```
const manifest = yield file_1.readJSON(path_1.join(manifestPath, "manifest.json" /* Manifest */));
```

Now I use this const enum value from `@sap-ux/project-types` in other module like `@sap/ux-project-access`, which is bundled in extension.  
-> add code to output on error console

Build vsix

Install and run in VSCode Insiders
-> error console shows
```
ERR [Extension Host] JustToProof.Works: 'thisworks'
ERR [Extension Host] JustToProof.MakeSure: 'thisstillworks'
```

I've also added eslint rule `no-restricted-syntax`

packages/project-access/src/translations/project-access.i18n.json
Klaus-Keller: removed usage of `i18next`

packages/fe-fpm-writer/src/app/index.ts
```diff
@@ -3,7 +3,7 @@ import type { Editor } from 'mem-fs-editor';
 import { create } from 'mem-fs-editor';
 import { join } from 'path';
 import { lt, valid } from 'semver';
-import type { Manifest } from '@sap-ux/ui5-config';
+import type { Manifest } from '@sap-ux/project-types';
```
stefanschreck, line 6: With https://github.wdf.sap.corp/ux-engineering/tools-suite/pull/21282 I just try to get rid of the redundant Manifest definitions ux-spec vs. ux-specification-types. Now we have a third definition here (and a fourth, partial one at sap.fe). Which of the definitions shall be the final one? I hope, this one, as it is open source?

packages/project-access/test/project/dependecies.test.ts
```diff
@@ -0,0 +1,22 @@
+import { Package } from '@sap-ux/project-types';
```
stefanschreck, line 1: typo: dependecies.test.ts

packages/fe-fpm-writer/src/app/index.ts
Klaus-Keller: @stefanschreck, perhaps there will be even one more, but I hope that will be the final one. I've contacted maintainers of https://github.com/SAP/ui5-manifest and asked to also publish the typescript types along with the JSON spec.

packages/project-access/test/project/dependecies.test.ts
Klaus-Keller: fixed, thanks!

packages/project-access/src/file/fileAccess.ts
```diff
@@ -0,0 +1,36 @@
+import { promises as fs } from 'fs';
```
tobiasqueck, line 1: Do we have a use for these helper functions? We use `mem-fs` everywhere in the `open-ux-tools` we are not going directly to the filesystem.

Klaus-Keller: I removed the file access functions from top level export, but still need it internally, e.g. to find all apps in a given list of workspace root files.

For other functions, we need to check whether they should offer to read via `mem-fs`. Currently exported functions:

```
findAllApps()
getCapModelAndServices()
getWebappPath() 
loadModuleFromProject()
```

out of this list, these functions access the file system:
`findAllApps()`: accesses `package.json` and `manifest.json` to find apps 
`getWebappPath()`: accesses `ui5.yaml` to pass content to ui5-config


other candidates of functions, that access the file system, but are currently not exported are:
```
isCapProject()
isCapNodeJsProject()
isCapJavaProject()
```

Proposal to enable project access with mem-fs:
Once we have a proper app or project access structure in place, we can have a factory method that allows to pass optionally a `mem-fs` instance. All access methods on returned object would then use the `mem-fs` instance in case is was provided. E.g. in future something like

```typescript
function createProjectAccess(root:string, fs?: Editor): ProjectAccess
```

tobiasqueck: @Klaus-Keller I gave `findFiles` and `findFileUp` a shot since these seem to be the lower level functions used for the functions above. My experiment is at https://github.com/SAP/open-ux-tools/pull/705/files. Opinions?

packages/project-access/src/project/cap.ts
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Return the CAP model and all services.
+ *
+ * @param projectRoot - CAP project root where package.json resides
+ */
+export async function getCapModelAndServices(
+    projectRoot: string
+): Promise<{ model: csn; services: { name: string; urlPath: string }[] }> {
+    const cds = await loadModuleFromProject<any>(projectRoot, '@sap/cds');
+    const capProjectPaths = await getCdsEnvFolders(projectRoot);
```
zdravko-georgiev, line 96: Can't we use `getCapCustomPaths`?

packages/project-access/src/project/findApps.ts
```diff
@@ -0,0 +1,225 @@
...
+ * Find root folder of the project containing the given file.
+ *
+ * @param path path of a project file
+ * @param sapuxRequired if true, only find sapux projects
+ */
+export async function findProjectRoot(path: string, sapuxRequired = true): Promise<string> {
+    const packageJson = await findFileUp(FileName.Package, path);
+    if (!packageJson) {
+        throw new Error(
+            `Could not find any project root for '${path}'. Search was done for ${
```
zdravko-georgiev, line 61: i18n?

zdravko-georgiev (commented):
Why so many empty test files? Can't we just mock the requests when testing?

packages/project-access/src/project/cap.ts
Klaus-Keller: Not sure if I understand, we call here `getCdsEnvFolders()` which calls `getCapCustomPaths()`?

packages/project-access/README.md
```diff
@@ -0,0 +1,20 @@
...
+`yarn add @sap-ux/project-access`
+
+Pnpm   
+`pnpm add @sap-ux/project-access`
+
+## Usage
+See usage in `test/` folder.
+
+## Keywords
+SAP Project Fiori Freestyle
```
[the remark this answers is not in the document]
Klaus-Keller: added

Klaus-Keller:
> Why so many empty test files? Can't we just mock the requests when testing?

I think 4 files to validate find functionality should be fine. Don't like the idea of mocking `fs` access for find functionality, in case we change to other libs I think it is safest to have the files physically existing.

packages/project-access/src/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const FileName = {
+    Manifest: 'manifest.json',
+    Package: 'package.json',
+    Ui5Yaml: 'ui5.yaml',
+    Ui5LocalYaml: 'ui5-local.yaml',
+    Ui5MockYaml: 'ui5-mock.yaml'
+} as const;
```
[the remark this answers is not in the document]
zdravko-georgiev: OK

zdravko-georgiev:
> > Why so many empty test files? Can't we just mock the requests when testing?
> 
> I think 4 files to validate find functionality should be fine. Don't like the idea of mocking `fs` access for find functionality, in case we change to other libs I think it is safest to have the files physically existing.

I understand, but I count 14 empty files across the whole PR...Anyway I was just surprised.

packages/project-access/src/project/cap.ts
zdravko-georgiev: I was just wondering if we can't use `getCapCustomPaths()` directly, without using `getCdsEnvFolders()` as "wrapper".

packages/project-access/src/file/fileAccess.ts
tobiasqueck: The low level fs functions are supporting mem-fs. The optional mem-fs parameter needs to be exposed by its using functions as well.

packages/project-access/src/project/cap.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+/**
+ * Returns true if the project is a CAP Node.js project.
+ *
+ * @param projectRoot - the root path of the project
+ * @param [packageJson] - the parsed package.json object
+ * @returns - true if the project is a CAP Node.js project
+ */
+export async function isCapNodeJsProject(projectRoot: string, packageJson?: Package): Promise<boolean> {
```
tobiasqueck, line 28: What would be a nice interface for supporting mem-fs? I see two options but I am unsure which to recommend:
- adding it as a 3rd parameter
- making the 1st parameter of type `string | { root: string; fs: Editor }`

tobiasqueck: Actually, I don't think the function is well designed (same for `isCapProject`). If you already have a `package.json` then you want to know if it is node with or without cap. If you don't have the package.json yet, then you want to know if it is CAP Java, Cap Node, or sth. else (the function will not tell you if it is a non-cap node project).
Please think about the exact usage of the functions and maybe split them e.g. 
- check for an already loaded package.json if CAP is enabled or not
- get the project type for a folder: cap-node, cap-java, ...

tobiasqueck (changes requested):
I have reviewed

- all related changes to moving the manifest type, and they are a exactly as expected
- project access root files as well as changes to the project root and they are as expected
- added comments about two functions with not ideal signature
- usage of mem-fs required for many of the exposed functions

packages/project-access/src/project/cap.ts
Klaus-Keller: done

packages/project-access/src/project/cap.ts
Klaus-Keller: Changed this with https://github.com/SAP/open-ux-tools/pull/685/commits/601fd71425e1fc4cc3dc3c6fac51bedb17ef83af. 

We have now one new type:
```
type CapProjectType = 'CAPJava' | 'CAPNodejs';
```

and 3 exported functions:

```
isCapNodeJsProject(packageJson: Package): boolean
isCapJavaProject(projectRoot: string): Promise<boolean>
getCapProjectType(projectRoot: string): Promise<CapProjectType | undefined>
```

packages/project-access/src/project/cap.ts
```diff
@@ -0,0 +1,98 @@
...
+import { loadModuleFromProject } from './moduleLoader';
+
+/**
+ * Returns true if the project is a CAP Node.js project.
+ *
+ * @param packageJson - the parsed package.json object
+ * @returns - true if the project is a CAP Node.js project
+ */
+export function isCapNodeJsProject(packageJson: Package): boolean {
+    return !!(packageJson?.cds || packageJson?.dependencies?.['@sap/cds']);
```
tobiasqueck, line 14: packageJson is defined based on the function signature, so you can remove the check, or is the signature incorrect?

Klaus-Keller: fixed

Klaus-Keller:
@ullasholla, could I get a review on the file and folder structure of project-access, especially this part:

```
src
├── file
│   ├── file-access.ts
│   ├── file-search.ts
│   └── index.ts
├── project
│   ├── cap.ts
│   ├── dependencies.ts
│   ├── index.ts
│   ├── module-loader.ts
│   ├── search.ts
│   └── ui5-config.ts
├── types
├── constants.ts
└── index.ts
```

You think this is fine or do you have any change requests?

tobiasqueck (approved):
* code adheres to coding guidelines and is well structured, is easy to read and understand
* very good test coverage
* thanks for the removal of duplicated code in other modules

packages/project-access/src/file/file-access.ts
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @param path - path to JSON file
+ * @param memFs - optional mem-fs-editor instance
+ * @returns - file content as object of type T
+ */
+export async function readJSON<T>(path: string, memFs?: Editor): Promise<T> {
+    if (memFs) {
+        return memFs.readJSON(path) as unknown as T;
+    } else {
+        return JSON.parse(await readFile(path)) as T;
```
ullasholla, line 30: Both these functions can throw errors if JSON parsing fails (mal-formed file, for example). Is it the responsibility of the caller to handle these? If so, can we add that to the JS Doc?

packages/project-access/src/file/file-search.ts
```diff
@@ -0,0 +1,91 @@
+import type { Editor } from 'mem-fs-editor';
+import { basename, dirname, join, sep } from 'path';
+import { default as find } from 'findit2';
+import { fileExists } from './file-access';
+
+/**
+ * Add missing dump properties
+ */
+declare module 'mem-fs-editor' {
```
ullasholla, line 9: Add a brief comment about the need to do this?

packages/project-access/src/file/file-search.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    export interface Editor {
+        dump(cwd: string): FileMap;
+    }
+}
+
+/**
+ * Search for 'filename' starting from 'root' in the list of virtual (not committed) files. Returns array of paths that contain the file.
+ *
+ * @param files - - array of paths that contain the filename found on the filesystem
```
ullasholla, line 20:
```suggestion
 * @param files - array of paths that contain the filename found on the filesystem
```

packages/project-access/src/file/file-search.ts
```diff
@@ -0,0 +1,91 @@
...
+ * @param filename - filename to search
+ * @param root - root folder to start search
+ * @param stopFolders - list of folder names to exclude (search doesn't traverse into these folders)
+ * @param fs - mem-fs-editor instance
+ * @returns - enhanced array of paths that contain the filename
+ */
+function checkVirtualFiles(files: string[], filename: string, root: string, stopFolders: string[], fs: Editor) {
```
ullasholla, line 27: Consider converting the parameters to a destructured object? This function accepts a lot of params.

Does `exclude` sound better than `stopFolders`?

packages/project-access/src/file/file-search.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Search for 'filename' starting from 'root'. Returns array of paths that contain the file.
+ *
+ * @param filename - filename to search
+ * @param root - root folder to start search
+ * @param stopFolders - list of folder names to exclude (search doesn't traverse into these folders)
+ * @param fs - optional mem-fs-editor instance
+ * @returns - array of paths that contain the filename
+ */
+export function findFiles(filename: string, root: string, stopFolders: string[], fs?: Editor): Promise<string[]> {
```
ullasholla, line 52: Does `exclude` sound better than `stopFolders`?

packages/project-access/src/file/file-access.ts
Klaus-Keller: This is a low level functionality and consumers need to handle errors. I've added this to JS Doc.

packages/project-access/src/file/file-search.ts
Klaus-Keller: `mem-fs-editor` type do not expose `dump`, added it as comment.

packages/project-access/src/file/file-search.ts
Klaus-Keller: done

packages/project-access/src/file/file-search.ts
```diff
@@ -0,0 +1,91 @@
...
+import { fileExists } from './file-access';
+
+/**
+ * Add missing dump properties
+ */
+declare module 'mem-fs-editor' {
+    type FileMap = { [key: string]: { state: 'modified' | 'deleted' } };
+
+    export interface Editor {
+        dump(cwd: string): FileMap;
```
ullasholla, line 13: JS Doc for `dump` may be helpful

packages/project-access/src/file/file-search.ts
Klaus-Keller: Changed it to `excludeFolders`, in JS Doc we also use the word exclude so your proposal makes it consistent.

packages/project-access/src/file/file-search.ts
Klaus-Keller: Copied docu from https://github.com/SBoudrias/mem-fs-editor#dumpcwd-filter

packages/project-access/src/file/file-search.ts
Klaus-Keller: changed `stopFolders` to `excludeFolders`. 
@tobiasqueck, any comments on the other points?

ullasholla:
@Klaus-Keller:
> @ullasholla, could I get a review on the file and folder structure of project-access,
> ...
> You think this is fine or do you have any change requests?

The folder structure looks good to me, cheers.

packages/project-access/test/project/ui5-config.ts
```diff
@@ -0,0 +1,24 @@
+import { join } from 'path';
```
zdravko-georgiev, line 1: File name is wrong. It should be `ui5-config.test.ts` 😉

packages/project-access/src/project/cap.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Returns the CAP project type, undefined if it is not a CAP project.
+ *
+ * @param projectRoot - root of the project, where the package.json resides.
+ * @returns - CAPJava for Java based CAP projects; CAPNodejs for node.js based CAP projects; undefined if it is no CAP project
+ */
+export async function getCapProjectType(projectRoot: string): Promise<CapProjectType | undefined> {
+    if (await isCapJavaProject(projectRoot)) {
+        return 'CAPJava';
```
zdravko-georgiev, line 36: Does it make sense to define a constant/enum for this one?

packages/project-access/src/project/cap.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+    let packageJson;
+    try {
+        packageJson = await readJSON<Package>(join(projectRoot, FileName.Package));
+    } catch {
+        // Ignore errors while reading the package.json file
+    }
+    if (packageJson && isCapNodeJsProject(packageJson)) {
+        return 'CAPNodejs';
```
zdravko-georgiev, line 45: Same as above

packages/project-access/src/project/cap.ts
tobiasqueck: @zdravko-georgiev we went away from enums to use basic types since I create less overhead when exporting. 
While the code now has a basic string, TS will throw errors in case you mistype.

packages/project-access/src/project/cap.ts
tobiasqueck: same as above :)

packages/project-access/src/file/file-search.ts
tobiasqueck: What!? This function is a piece of art! Joke aside, I will rework it based on the very good feedback.

tobiasqueck: > Instead of modifying the original array being passed in, does returning a new array make more sense? The caller can choose to extend the original array if they want

The incoming array is not being modified in the current code.

tobiasqueck: Anyways, please review https://github.com/SAP/open-ux-tools/pull/685/commits/a4918f6c32f5b5f2685a385891642baf1ec09f32. I have split the functionality into two functions with a single purpose each. This also reduced the amount of parameters.

packages/project-access/test/project/ui5-config.ts
zdravko-georgiev: thanks

packages/project-access/src/project/cap.ts
zdravko-georgiev: thanks for the info

zdravko-georgiev (approved):
Requested changes incorporated, code is covered with unit tests, looks good

packages/project-access/src/file/file-search.ts
```diff
@@ -0,0 +1,106 @@
...
+            }
+        });
+        finder.on('file', (file: string) => {
+            if (file.endsWith(sep + filename)) {
+                results.push(dirname(file));
+            }
+        });
+        finder.on('end', () => {
+            if (fs) {
+                const changes = fs.dump('');
```
Klaus-Keller, line 77: @tobiasqueck according to https://github.com/SBoudrias/mem-fs-editor#dumpcwd-filter
the `cwd` argument of `dump()` is optional. Should we remove `''`?

tobiasqueck: No :) Using `''` makes the returned paths absolute, like paths `findit2` returns

ullasholla (changes requested):
Had a few more comments

packages/project-access/src/file/file-search.ts
```diff
@@ -0,0 +1,97 @@
+import type { Editor } from 'mem-fs-editor';
+import { basename, dirname, join, sep } from 'path';
+import { default as find } from 'findit2';
```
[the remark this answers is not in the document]
Klaus-Keller: @ullasholla, we've tried `glob` which was 5x-10x slower than `findit2`. As we use `findit2` since quite some time without issues, I'd like to keep it in order to reduce risks of regression. Not saying that we should never swap the find module, but it should be done as a separate step with deeper analysis, which I consider as done for `findit2` due to long term experience.

packages/project-access/test/file/file-access.test.ts
```diff
@@ -0,0 +1,69 @@
...
+            expect(content).toContain('file-access.test.ts');
+        });
+
+        test('Read non existing file, should throw error', async () => {
+            try {
+                await readFile('NOT_EXISTING_FILE');
+                fail(`readFile() should have thrown error but did not`);
+            } catch (error) {
+                expect(error.toString()).toContain('NOT_EXISTING_FILE');
+            }
```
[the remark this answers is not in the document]
Klaus-Keller: Rather a matter of 'debugability'. I find it way easier to inspect the error object while writing/checking test cases this way. If there are valid reasons to use one over the other let me know, I can change it.

packages/project-access/test/file/file-access.test.ts
```diff
@@ -0,0 +1,69 @@
...
+            expect(fileExists(join(__dirname, 'file-access.test.ts'))).resolves.toBe(true);
+        });
+
+        test('Check non existing file, should return false', () => {
+            expect(fileExists('DOES_NOT_EXIST')).resolves.toBe(false);
+        });
+
+        test('Check existing file in memf-fs, should return true', () => {
+            expect(fileExists(memFilePath, memFs)).resolves.toBe(true);
+        });
```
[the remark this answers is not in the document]
Klaus-Keller: Good catch! I've tested cases when expect fails and the issue is visible on console only, test case doesn't fail.
Changed it.

zdravko-georgiev (approved):

kranthie-sap (approved):
The updates to codeowners and sonar project config file look good.

tobiasqueck (approved):
I have re-reviewed the usage of the new types in `fe-fpm-writer` and `ui5-application-writer` as well as the removal of them in `ui5-config`. I don't see any issue, on the contrary, `project-access` is a better location for type `Manifest`.

Issue #680, opened 2022-09-16
TBI - Allow generation of OPA tests with `@sap-ux/fiori-elements-writer`


IainSAP:
### Description (include screenshots)
This TBI will integration the recently added `@sap-ux/ui5-test-writer` into the `@sap-ux/fiori-elements-writer` module. Generation of tests will initially only be supported for v4 applications where a mock server is also available (any apps that have a valid metadata xml available). 

### Value
Provides additional functionality.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Integrate `@sap-ux/ui5-test-writer` into the `@sap-ux/fiori-elements-writer` module
- [ ] Add tests

PR #686, opened 2022-09-20
Tbi/680/integrate ui5 test writer to fiori elements writer1

IainSAP:
https://github.com/SAP/open-ux-tools/issues/680

- Move new `addTests` option to avoid future breaking changes


tobiasqueck (approved):
Thank you! That should keep the API stable for the future.

devinea (approved):

PR #688, opened 2022-09-20
fix(fe-fpm-writer): 'enableFPM' method throws 'Invalid Version' exception if unknown UI5 version stored in '["sap.ui5"]["dependencies"]["minUI5Version"]'

815are:
If we call `enableFPM`, when we have following entry in `manifest.json`:
image

we get exception `TypeError: Invalid Version: ${sap.ui5.dist.version}` in `..\node_modules\semver\classes\semver.js`.

Changes:
1. Applying additional check if version is valid by calling `semver.valid`;
2. Additional minor check if 'dependencies' property is defined;

### DIFF
--- a/.changeset/little-foxes-mate.md
+++ b/.changeset/little-foxes-mate.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fe-fpm-writer': patch
+---
+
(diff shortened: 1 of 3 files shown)


tobiasqueck (approved):
good catch!

johannes-kolbe (approved):

IainSAP (approved):

Issue #694, opened 2022-09-23
FEATURE - EnvCheck - Enable environment check for VSCode


cianmSAP:
**As a developer developing Fiori application using SAP Fiori tools in VSCode, I would like a quick better way to gather basic information about my environment so that I can send it to SAP while reporting incident to minimize the time it takes going back n forth to support asks about installed component versions**

### Description

- SAP Fiori tools Environment Check for SBAS has been beneficial for reducing the support roundtrip and is being used by SAP Fiori tools team as well as SBAS support team and slowly growing.
- Now, with this US, we want to introduce first set of a similar environment check tailored towards VSCode which will help both internally and externally with customers who are running into issues related to running SAP Fiori tools in VSCode.
- This feature will help Fiori tools team to gather most relevant info. Of course, this is initial set, can be expanded later. 
- We will have follow up US to have saved `SAP systems check`, similar to what we do for destination in SBAS
- If during grooming, team is ok, then we can enable similar for SBAS too.

Environment Check in VSCode will report/check the following:
1. node version
2. Cloud CLI tools installed or not
3. Application Wizard version
4. SAP Fiori tools - Application generator version
5. SAP Fiori tools - Application Modeler version
6. SAP Fiori tools - Guided Development version
7. SAP Fiori tools - Service Modeler version
8. SAP Fiori tools - XML Annotation Language Server version
9. XML Toolkit version
10. SAP CDS Language Support version for CAP
11. UI5 Language Assistant Support extension

### Design
Interaction
Functional Flow 1
- User in VSCOde enter command `Fiori: Open Environment Check`
- From the quick pick, user selects `Gather environment information`
- A `envirnoment_info.md` file opens with title `SAP Fiori tools - Environment Details` and is saved under `envcheck-<timestamp>.zip`

### Acceptance Criteria
- Given i run command `Fiori: Open Environment Check` and select `Gather environment information`, I should be shown environment info in a .md file and should also be saved in a zip file.

Sample Output:

| Tools/Extensions                                 | Version |
|--------------------------------------------------|---------|
| Node.js                                          |         |
| Cloud CLI tools                                  |         |
| Application Wizard                               |         |
| SAP Fiori tools - Fiori generator                |         |
| SAP Fiori tools - Application Modeler            |         |
| SAP Fiori tools - Guided Development             |         |
| SAP Fiori tools - Service Modeler                |         |
| SAP Fiori tools - XML Annotation Language Server |         |
| XML Toolkit                                      |         |
[...]

PR #695, opened 2022-09-23
feat(env-check): add environment check for vscode

cianmSAP:
#694

Add env check for vscode


packages/environment-check/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/environment-check",
-    "version": "0.1.4",
+    "version": "0.1.5",
```
tobiasqueck, line 3: Please leave the version number as-is. Once you are done, you add a changeset and the version number will be updated automatically.

cianmSAP: Updated, thanks @tobiasqueck

packages/environment-check/src/cli/index.ts
```diff
@@ -150,7 +151,7 @@ export async function cli(): Promise<void> {
         }
         const options = getOptions(cliArgs) || {};
         options.credentialCallback = credentialCallback;
-        const result = await checkEnvironment(options);
+        const result = isAppStudio() ? await checkBASEnvironment(options) : await checkVSCodeEnvironment();
```
Klaus-Keller, line 154: as mentioned above, I'd keep `checkEnvironment()` as is and move the check done here inside `checkEnvironment`.

packages/environment-check/src/checks/index.ts
```diff
@@ -1,2 +1,2 @@
-export { checkEnvironment, getEnvironment } from './environment';
+export { checkBASEnvironment, getBASEnvironment, checkVSCodeEnvironment, getVSCodeEnvironment } from './environment';
```
Klaus-Keller, line 1: same as above, please keep the public interface 
```
export { checkEnvironment, getEnvironment } from './environment';
```
stable

packages/environment-check/src/checks/install.ts
```diff
@@ -0,0 +1,64 @@
...
+    return versions;
+};
+
+/**
+ * Read the version of the cloud foundry CLI.
+ *
+ * @param module - module id
+ * @returns version
+ */
+export const getCFCliToolVersion = async (module: string): Promise<string> => {
```
Klaus-Keller, line 32: The name suggests that this function returns the version of the "Clound Foundry CLI (cf)", but you still need to pass parts of the CLI command into it. I'd say, either this function returns the version of cf and doesn't require the consumer to know about the command name (`cf`), or we rename this function to indicate, that you can pass any command line command and it will return the version (as long as argument -v returns the version. I'm in favor of the first and made the suggestions accordingly.

```suggestion
export const getCFCliToolVersion = async (): Promise<string> => {
```

packages/environment-check/src/checks/install.ts
```diff
@@ -0,0 +1,64 @@
...
+    let cfVersion: string;
+    try {
+        const ouput = await spawnCommand(module, ['-v']);
```
Klaus-Keller, line 35:
```suggestion
        const ouput = await spawnCommand(NpmModules.CloudCliTools, ['-v']);
```
Requires also to add 
```
import { NpmModules } from '../types';
```
to the import section. 

Also there is a type: ouput instead of output

packages/environment-check/src/checks/install.ts
```diff
@@ -0,0 +1,64 @@
+import { existsSync, promises } from 'fs';
+import { t } from '../i18n';
+import { join } from 'path';
+import { spawnCommand, npmCommand } from '../command';
+
```
Klaus-Keller, line 5: Can we use format
[export] function <FUNCTION_NAME>() {
<FUNCTION_BODY>
}
for top level functions to be consistent with the rest of this module? 
Usually we use arrow function expressions for anonymous functions or functions that do not require a body.

Additionally, the name `install.ts` is a bit misleading.

packages/environment-check/src/checks/install.ts
```diff
@@ -0,0 +1,64 @@
...
+};
+
+/**
+ * Read the version of the fiori generator.
+ *
+ * @param module - module id
+ * @returns version
+ */
+export const getFioriGenVersion = async (module: string): Promise<string> => {
+    let fioriGenVersion: string;
```
Klaus-Keller, line 50: similar comment like for `getCFCliToolVersion()`, I'd either remove the need to add the `module` name or make this function more generic so it can return the version of any globally installed module and rename it.

packages/environment-check/src/output/markdown.ts
```diff
@@ -3,12 +3,14 @@ import { Severity, UrlServiceType } from '../types';
...
-    Environment,
+    BASEnvironment,
     EnvironmentCheckResult,
     MarkdownWriter,
-    ResultMessage
+    ResultMessage,
+    VSCodeEnvironment
 } from '../types';
 import { t } from '../i18n';
+import { isAppStudio } from '@sap-ux/btp-utils';
```
Klaus-Keller, line 13: Ideally, the markdown writer doesn't have any knowledge about the environment. We should try to find a result structure that can be converted to markdown without knowing which environment we currently running in.

packages/environment-check/src/output/zip.ts
```diff
@@ -25,8 +25,13 @@ function byteNumberToSizeString(byteNumber: number): string {
  *
  * @param results - environment check results
  * @param targetFile - path and filename of target zip archive. Default is 'envcheck-results.zip'.
+ * @param markdownOnly - only include markdown in the zip
  */
-export function storeResultsZip(results: EnvironmentCheckResult, targetFile = 'envcheck-results.zip'): void {
+export function storeResultsZip(
+    results: EnvironmentCheckResult,
+    targetFile = 'envcheck-results.zip',
+    markdownOnly = false
```
Klaus-Keller, line 33: In which case do we need `markdownOnly`? I think if we store check results as zip it is always better to include json and markdown. @theStiggySpd was there a specific request?

Klaus-Keller (changes requested):

packages/environment-check/src/types.ts
```diff
@@ -14,11 +14,41 @@ export enum OutputMode {
...
+    AppMod = 'SAPSE.sap-ux-application-modeler-extension',
+    Help = 'SAPSE.sap-ux-help-extension',
+    ServiceMod = 'SAPSE.sap-ux-service-modeler-extension',
+    CDS = 'SAPSE.vscode-cds'
+}
+
+interface Environment {
     platform: NodeJS.Platform;
-    versions: NodeJS.ProcessVersions;
+}
```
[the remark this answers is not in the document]
cianmSAP: It wasn't removed as such but moved into BASEnviornment and there was no request to include the additional `NodeJS.ProcessVersions` (`v8`, `uv`, `zlib` etc ) in the VSCode results.

packages/environment-check/src/output/zip.ts
cianmSAP: I did raise this in grooming and it was requested to provide markdown only.

cianmSAP:
With the ask to include `Gather environment information` for BAS also as per https://github.wdf.sap.corp/ux-engineering/tools-suite/issues/21564#issue-5803089 I've exposed the `getToolsExtensions` function and also reverted the checkEnvironment and getEnvironment functions as requested.

The only slight troube with keeping the markdown file unaware of the dev environment is/was the title. I've added the markdownTitle property to the `EnvironmentCheckResult` interface and exposed the `getMarkdownTitle` function to retreieve it. It may be over engineering a simple part so please suggest an appropriate or alternative way. 
Thanks @Klaus-Keller

packages/environment-check/src/output/markdown.ts
cianmSAP: See comment

packages/environment-check/src/checks/install.ts
cianmSAP: Updated

packages/environment-check/src/checks/install.ts
cianmSAP: Updated

packages/environment-check/src/checks/install.ts
cianmSAP: Updated as suggested

packages/environment-check/src/checks/install.ts
cianmSAP: Updated as suggested

packages/environment-check/src/cli/index.ts
cianmSAP: Done

packages/environment-check/src/checks/environment.ts
```diff
@@ -8,31 +8,33 @@ import type {
...
 
 /**
- * Return the environment.
+ * Returns the BAS environment.
  *
- * @returns environment, including ide, versions, ...
+ * @returns environment, including platform, versions ...
  */
-export async function getEnvironment(): Promise<{ environment: Environment; messages: ResultMessage[] }> {
+export async function getBASEnvironment(): Promise<{ environment: BASEnvironment; messages: ResultMessage[] }> {
```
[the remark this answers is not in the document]
cianmSAP: Done

packages/environment-check/src/types.ts
```diff
@@ -22,6 +36,20 @@ export interface Environment {
     basDevSpace?: string;
 }
 
+export interface ToolsExtensions {
```
Klaus-Keller, line 39: Interface `ToolsExtensions` looks good, but I would have an optional property in `Environment` with this interface. Logically extension and module versions belong to the environment. Right now we read and fill this information when running in VSCode only, but at some point we might collect the same information in BAS.

packages/environment-check/src/output/markdown.ts
```diff
@@ -259,11 +296,15 @@ function writeMessages(writer: MarkdownWriter, messages: ResultMessage[] = []):
 export function convertResultsToMarkdown(results: EnvironmentCheckResult): string {
     const writer = getMarkdownWriter();
 
-    writer.addH1(t('markdownText.envCheckTitle'));
-    writeEnvironment(writer, results.environment);
-    writeDestinationResults(writer, results.destinationResults, results.destinations);
-    writeDestinations(writer, results.destinations);
-    writeMessages(writer, results.messages);
+    writer.addH1(results.markdownTitle);
+    if (results.toolsExtensions) {
```
Klaus-Keller, line 300: Can we avoid having an `if` here and output whatever is there in the `EnvironmentCheckResult`. We might fill certain properties only in certain cases, but the writer should just traverse over the results and output markdown, not aware of other logic. This keeps the writer simple. I assume at some point we might also check certain module versions in BAS and destinations in VSCode.

packages/environment-check/src/checks/environment.ts
```diff
@@ -137,31 +204,44 @@ async function getDestinationsResults(
...
+        const destResults = await getDestinationsResults(
+            deepDiveDestinations,
+            destinations,
+            options?.credentialCallback
+        );
+        destinationResults = destResults.destinationResults;
+        logger.push(...destResults.messages);
+    }
+
+    const toolsExtensionResults = await getToolsExtensions();
```
Klaus-Keller, line 236: I've tested in BAS and when I enter this function I get an error:

```
Error checking environment: Error: spawn code ENOENT
```

Because command `code` doesn't exist in BAS. This indicates also that we need a better error handling, if getting one installed extension/module fails, it should continue with the others and not completely exit.

packages/environment-check/src/checks/environment.ts
```diff
@@ -44,6 +48,69 @@ export async function getEnvironment(): Promise<{ environment: Environment; mess
...
+ *
+ * @returns environment, including tools, extensions ...
+ */
+export async function getToolsExtensions(): Promise<{
+    toolsExtensions: ToolsExtensions;
+    messages: ResultMessage[];
+}> {
+    const logger = getLogger();
+
+    const extensions = await getInstalledExtensions();
```
Klaus-Keller, line 62: this step could fail, but should not cause the whole check to fail. Perhaps wrap it in `try/catch` or add try catch inside `getInstalledExtensions`

Klaus-Keller (changes requested):
Hi @cianmSAP, 
See my comments with further review. I did not review the whole code yet, will continue tomorrow.

packages/environment-check/src/checks/environment.ts
cianmSAP: Fixed

packages/environment-check/src/types.ts
cianmSAP: Updated

packages/environment-check/src/output/markdown.ts
```diff
@@ -254,15 +294,18 @@ function writeMessages(writer: MarkdownWriter, messages: ResultMessage[] = []):
  * Converts the envcheck results to markdown report.
  *
  * @param results - envcheck results
+ * @param isDestinationsChecked - has destinations been checked, default is true
  * @returns - markdown report
  */
-export function convertResultsToMarkdown(results: EnvironmentCheckResult): string {
+export function convertResultsToMarkdown(results: EnvironmentCheckResult, isDestinationsChecked = true): string {
```
cianmSAP, line 300: I've added this as I don't want to implement a check for the existence of destinations in the results, as I've seen in support where customers have no destinations in the results due to configuration issues. 
Therefore I think it might be better to pass the param from app-mod when there has actually been no checked performed.

Klaus-Keller: Good consideration, we need a way to differentiate between:
- 0 destinations due to configuration issues
- 0 destinations because we just do an environment check

We could even extend this concept a bit, instead of manually passing this into `convertResultsToMarkdown()` function, could we add a property on the result state that indicates which check were requested? 

E.g. extend type  `EnvironmentCheckResult`
https://github.com/SAP/open-ux-tools/blob/e026fc2f82327df8068c7c4e29f874bf05528eab/packages/environment-check/src/types.ts#L65-L70
with

```typescript
    requestedChecks: Set<Check>
}

type Check = 'environment' | 'destination' | 'destinationResults';

```
We also allow to skip `destinationResults` and the same concept could be applied there. 
That way it would be clear from the results which check was executed and the markdown writer could skip results for not executed checks.

What do you think?

Klaus-Keller (changes requested):

Klaus-Keller:
> The only slight troube with keeping the markdown file unaware of the dev environment is/was the title. I've added the markdownTitle property to the `EnvironmentCheckResult` interface and exposed the `getMarkdownTitle` function to retrieve it. It may be over engineering a simple part so please suggest an appropriate or alternative way. Thanks @Klaus-Keller

@theStiggySpd, why do we need different title for VSCode or BAS results? This information is already included in "Environment", see:

image

Personally, I think we should try to have the checks as development environment agnostic as possible, what do you think?

Klaus-Keller:
@cianmSAP, another information, there is in new version of `@sap/bas-sdk` available: `2.1.1`. The new version includes functionality to get the devspace information. 
We can delete our custom function `getSbasDevspace()`:

https://github.com/SAP/open-ux-tools/blob/e026fc2f82327df8068c7c4e29f874bf05528eab/packages/environment-check/src/checks/environment.ts#L169-L190


and retrieve the information directly from `bas-sdk`, like:

```
import { core as basCoreApi, devspace } from '@sap/bas-sdk';

..
      if (isAppStudio()) {
          environment.basDevSpace = (await devspace.getDevspaceInfo()).packDisplayName;

```

packages/environment-check/src/output/markdown.ts
cianmSAP: Implemented this, I think it will keep the markdown clean when the check has not been performed. Thanks @Klaus-Keller

packages/environment-check/src/checks/environment.ts
```diff
@@ -27,23 +29,113 @@ export async function getEnvironment(): Promise<{ environment: Environment; mess
...
     };
 }
 
+/**
+ * Returns extensions and versions.
+ *
+ * @param extensions to be checked
+ * @returns extension with version numbers
+ */
+function __getExtVersions(extensions: { [id: string]: { version: string } }): { [id: string]: string } {
```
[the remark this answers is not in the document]
cianmSAP: Updated

packages/environment-check/src/checks/getInstalled.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+    return false;
+}
+
+/**
+ * Reads the list of extensions installed in BAS.
+ *
+ * @returns list of extension ids and versions
+ */
+async function _getExtensionsBAS(): Promise<{ [id: string]: { version: string } }> {
```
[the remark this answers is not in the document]
cianmSAP: Updated

packages/environment-check/src/checks/getInstalled.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+    return cfVersion;
+}
+
+/**
+ * Read the version of the fiori generator in BAS.
+ *
+ * @returns version
+ */
+async function _getFioriGenBAS(): Promise<string> {
```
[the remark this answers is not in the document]
cianmSAP: Updated

packages/environment-check/src/checks/getInstalled.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+    return version;
+}
+
+/**
+ * Read the version of the fiori generator in VSCode.
+ *
+ * @returns version
+ */
+async function _getFioriGenGlobalRoot(): Promise<string> {
```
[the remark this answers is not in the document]
cianmSAP: Updated

packages/environment-check/src/checks/environment.ts
```diff
@@ -137,26 +219,32 @@ async function getDestinationsResults(
  */
 export async function checkEnvironment(options?: CheckEnvironmentOptions): Promise<EnvironmentCheckResult> {
     const logger = getLogger();
+    let destinations: Destination[];
+    let destinationResults: { [dest: string]: DestinationResults };
 
     const { environment, messages } = await getEnvironment();
```
Klaus-Keller, line 225: we now need to set the ``requestedChecks` otherwise they are not part of the markdown if you run it from command line:

```suggestion
    const requestedChecks: Set<Check> = new Set();
    
    requestedChecks.add(Check.Environment);
    const { environment, messages } = await getEnvironment();
```

packages/environment-check/src/checks/environment.ts
```diff
@@ -137,26 +219,32 @@ async function getDestinationsResults(
...
+            workspaceResults.destinations.forEach((dest) => deepDiveDestinations.add(dest));
+        }
 
-    const { messages: destDetailsMessages, destinationResults } = await getDestinationsResults(
-        deepDiveDestinations,
-        destinations,
-        options?.credentialCallback
-    );
-    logger.push(...destDetailsMessages);
+        const basDestResults = await checkBASDestinations();
```
Klaus-Keller, line 236: similar here, when we check destinations we need to add it to requested checks

```suggestion
        requestedChecks.add(Check.Destinations);
        const basDestResults = await checkBASDestinations();
```

packages/environment-check/src/checks/environment.ts
```diff
@@ -137,26 +219,32 @@ async function getDestinationsResults(
...
+        destinations = basDestResults.destinations;
+        logger.push(...basDestResults.messages);
+
+        const destResults = await getDestinationsResults(
```
Klaus-Keller, line 240: and here, proposal:

```suggestion
        if (deepDiveDestinations.size > 0) {
            requestedChecks.add(Check.DestResults);
        }
        const destResults = await getDestinationsResults(
```

Klaus-Keller (changes requested):
@cianmSAP, looks pretty good. I found one small issue when running the check from command line now, see comments, should be easy fix.

packages/environment-check/src/output/zip.ts
```diff
@@ -20,6 +21,32 @@ function byteNumberToSizeString(byteNumber: number): string {
     return `${parseFloat((byteNumber / Math.pow(1024, i)).toFixed(2))} ${units[i]}`;
 }
 
+/**
+ * Convert results to JSON.
+ *
+ * @param results - environment check results
+ * @returns json string
+ */
+function convertResultsToJson(results: EnvironmentCheckResult): string {
```
Klaus-Keller, line 30: not sure whether we need to massage the result json. I think I would keep it raw, it contains the information of `requestedChecks` so when checking the JSON it is should be clear what checks have been executed

packages/environment-check/src/checks/environment.ts
cianmSAP: Thanks Klaus, slipped my mind that we could generate the markdown from the CLI. Updated now

packages/environment-check/src/output/zip.ts
cianmSAP: Ok. I've reverted

packages/environment-check/src/checks/environment.ts
cianmSAP: Updated

packages/environment-check/src/checks/environment.ts
cianmSAP: Updated

Klaus-Keller:
Hi @cianmSAP, sorry, I forgot that a set is not serialized when running `JSON.stringify()`. This is why the `requestedChecks` where not shown in the resulting json. 
There are two ways to overcome this:
1. Write an own serializer for `Set`, e.g. by passing a function as second argument to `JSON.stringify`
2. Use array instead of Set

I think it easier to switch to array, so I pushed a commit: https://github.com/SAP/open-ux-tools/pull/695/commits/08a15c5e9ca93e301d446c174d9db0c470447d91

Klaus-Keller (approved):
- Checked code and it looks good to me now
- Test coverage is excellent
- Build and tested via CLI in VScode and SBAS, no issues found

**Approved**. Thanks a lot @cianmSAP

Klaus-Keller:
@tobiasqueck, could you please also check again.

tobiasqueck (approved):
Not required to be changed for this PR, however, you should think about an object oriented approach with a base class and extensions for VSCode and BAS. That would remove requiring `isAppStudio` checks in e.g. `getFioriGenVersion` or `getInstalledExtensions`.

Also the interfaces e.g. `ToolsExtensions` would be easier to read if the properties had a short explanation each.

All in all, the code is easy to read, and the tests are covering the functionality, therefore, please go ahead and merge. Nevertheless, please consider my above comments in a follow up PR.

cianmSAP:
> Not required to be changed for this PR, however, you should think about an object oriented approach with a base class and extensions for VSCode and BAS. That would remove requiring `isAppStudio` checks in e.g. `getFioriGenVersion` or `getInstalledExtensions`.
> 
> Also the interfaces e.g. `ToolsExtensions` would be easier to read if the properties had a short explanation each.
> 
> All in all, the code is easy to read, and the tests are covering the functionality, therefore, please go ahead and merge. Nevertheless, please consider my above comments in a follow up PR.

Noted, thanks @tobiasqueck

Klaus-Keller:
> Not required to be changed for this PR, however, you should think about an object oriented approach with a base class and extensions for VSCode and BAS. That would remove requiring `isAppStudio` checks in e.g. `getFioriGenVersion` or `getInstalledExtensions`.

@tobiasqueck, let's have a chat about this before adding classes into this functional component.

tobiasqueck:
> let's have a chat about this before adding classes into this functional component.

Sure, we can discuss. It see it as a simple "is there a state or not" question, and by looking at the code, I see more than a handful `isAppStudio` calls leading to `if-else` conditions that strongly indicate that using polymorphism here would reduce the cognitive complexity of the code. It would also cleanly separate environment specific code into one place instead of mixing it everywhere.

Klaus-Keller:
> Sure, we can discuss. It see it as a simple "is there a state or not" question, and by looking at the code, I see more than a handful `isAppStudio` calls leading to `if-else` conditions that strongly indicate that using polymorphism here would reduce the cognitive complexity of the code. It would also cleanly separate environment specific code into one place instead of mixing it everywhere.

Yes, fully agree to improve the separation `isAppStudio` or not, ideally the differences should be minimal and clearly visible in the structure of the code. We can certainly improve this.

PR #696, opened 2022-09-27
fix(fe-fpm-writer): Add sap.se.macros namespace to custom section's Fragment.xml when ui5 version >= 1.90

815are:
Add `sap.se.macros` namespace to custom section's `Fragment.xml` when ui5 version >= 1.90
I am not fully sure about `1.90`, but it is listed here as `Available since` - https://sapui5.hana.ondemand.com/sdk/#/api/sap.fe.macros.FormElement

### DIFF
--- a/.changeset/many-onions-hammer.md
+++ b/.changeset/many-onions-hammer.md
@@ -2,4 +2,4 @@
 '@sap-ux/fe-fpm-writer': patch
 ---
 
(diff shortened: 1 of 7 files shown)


tobiasqueck (approved):
Reviewed both sources and tests and they meat the code quality requirements. Additionally, the tests cover both pre-macro and macro support cases

johannes-kolbe (approved):
lgtm

PR #698, opened 2022-09-27
move from dev_deps to deps

Jimmy-Joseph19:
Following deps are required at runtime so moving from devDeps =>  deps
`react-movable`, `react-virtualized`, `sanitize-html`


devinea (approved):
Thanks @Jimmy-Joseph19 
Changes look good.

815are (approved):
such version was working ok when it was tested using generated `tgz`

PR #689, opened 2022-09-21
chore - linting changes

devinea:
Changes 
- `pnpm pretty-quick --staged --ignore-path .prettierignore` to 
`pnpm pretty-quick --ignore-path .prettierignore` using alias `pnpm run:pretty-quick` as many developers commit without staging.
- Run linting on test folder (static checks and ensures prettier has been applied to the code)
   - remove `test` from `.eslintignore`
   - add `test` to  `tesconfig.json`
   - created `tsconfig-build.json` to exclude `test` from the build command so that test is not emitted to `dist`
   - Apply fixes from lint
- Apply prettier formatting. 

Verified `eslint-disable-next-line` statements or corrections. Each module needs two people as verifiers.

- [x] axios-extension @ullasholla @tobiasqueck 
- [ ] environment-check @tobiasqueck  
- [x] fiori-elements-writer @unseen1980 @IainSAP 
- [x] fiori-freestyle-writer @unseen1980 @IainSAP 
- [ ] logger - @tobiasqueck 
- [ ] store - @tobiasqueck 
- [ ] ui5-config @tobiasqueck 
- [x] ui5-proxy-middleware - @tobiasqueck @zdravko-georgiev 

Following modules don't have changes in `src` or `test`

- [x] backend-proxy-middleware
- [x] btp-utils
- [x] fe-fpm-writer (#699)
- [x] odata-service-writer (#702)
- [x] ui-components (#700)
- [x] ui5-application-writer (#700)
- [x] ui5-test-writer
- [x] yaml

PR #702, opened 2022-09-28
linting fixes for base writers

tobiasqueck:
Carve out from #689 for `odata-service-writer` and `ui5-application-writer`
Main focus was on null-checks when using semver


Klaus-Keller (approved):
Thanks @tobiasqueck, lint fixes look good and lead to better readable code. Didn't build and run locally, just visual code review.
Approved.

PR #705, opened 2022-09-29
proposal for usage of mem-fs

tobiasqueck:
As discussed, all the base function offer optional support for `mem-fs-editor` instances, however, if not provided, the file system is directly access as before, therefore, there shouldn't not be any overhead for the existing functionality but it also is not possible to use these functions in the context of the writers.

Not done in this PR is to also expose the optional parameter in the consuming functions.


tobiasqueck:
@Klaus-Keller please go ahead and merge it into your branch if the changes are acceptable for you

packages/project-access/src/file/findFiles.ts
```diff
@@ -39,8 +78,27 @@ export async function findFiles(filename: string, root: string, stopFolders: str
...
+ * @param pathName - path for start searching up
+ * @param fs - mem-fs-editor instance
+ * @param fs.exists - function to be used to check the file existence
  * @returns - path to file name if found, otherwise undefined
  */
-export async function findFileUp(fileName: string, startPath: string): Promise<string | undefined> {
-    return findUp(fileName, { cwd: startPath });
+function findUp(fileName: string, pathName: string, fs: { exists: Editor['exists'] }): string | undefined {
+    const filePath = join(pathName, fileName);
+    if (fs.exists(filePath)) {
```
Klaus-Keller, line 99: @tobiasqueck, is there a reason why we have this not as `async function` and use function 
```
fileExists(path: string, memFs?: Editor)
```
from `fileAccess.ts`? I see you've even prepared the function to use with mem-fs.

packages/project-access/src/file/findFiles.ts
```diff
@@ -1,20 +1,59 @@
...
+ * Search for 'filename' starting from 'root' in the list of virtual (not committed) files. Returns array of paths that contain the file.
+ *
+ * @param files - - array of paths that contain the filename found on the filesystem
+ * @param filename - filename to search
+ * @param root - root folder to start search
+ * @param stopFolders - list of folder names to exclude (search doesn't traverse into these folders)
+ * @param fs - mem-fs-editor instance
+ * @returns - enhanced array of paths that contain the filename
+ */
+function checkVirtualFiles(files: string[], filename: string, root: string, stopFolders: string[], fs: Editor) {
```
Klaus-Keller, line 27: Could we just return the result of search on virtual file system and do the merge of 
`[...physicalFSresults, ...virtualFSresults]` outside in the consuming function `findFiles()`. I think this would make the scope of this function cleaner.

packages/project-access/src/file/findFiles.ts
tobiasqueck: I prepared it after doing that one and totally forgot to consolidate the code.
Here it is now https://github.com/SAP/open-ux-tools/pull/705/commits/f8db29e7fbcb419f0cfad59b5b3f28dcb7d7379f

packages/project-access/src/file/findFiles.ts
tobiasqueck: I am not in favor of separating it, since we have that patter of the option `fs` parameter all over the code, I would also like to consistently keep it here.

Klaus-Keller: Guess there is a misunderstanding, Intention was not to remove the `fs` argument. What I meant is changing the function to:

```typescript
function findVirtualFiles(filename: string, root: string, stopFolders: string[], fs: Editor): string[]
```

So we would return the result array and not pass it into the function and fill it, same signature as `findFiles()`.

tobiasqueck: The function does a little more than just finding virtual files, it also removes files that were found on the filesystem but are removed in the `mem-fs`. How about renaming it to `applyChangesFromMemory`?

Klaus-Keller: Ah, got it. In this case it is ok. I thought the find functionality on virtual files might also be a candidate for standalone usage at some point. We can also keep it like this and extract find in case there is demand later on.

Klaus-Keller (approved):

PR #706, opened 2022-09-29
fix(ui-components) minor fix Storybook bundle

devinea:


devinea:
With nx caching this is not so important anymore.

PR #707, opened 2022-09-30
chore(ui-components) cherrypick UITable fix tools-suite/pull/21766

devinea:
Fix to correct cell editing state


Jimmy-Joseph19 (approved):
Approving as i have reviewed the code but did not test locally As code was already tested with internal version in ts

Klaus-Keller (approved):
- Code looks good
- Test cases look good
- Did not build and test locally, but fix should have been verified in tools-suite repo

Approved

PR #708, opened 2022-09-30
Chore - Upgrade to "typescript": "4.9.4" and adapt code

devinea:
Chore - Upgrade to "typescript": "4.9.4" and adapt code


packages/axios-extension/test/auth/uaa.test.ts
```diff
@@ -70,6 +70,7 @@ describe('UAA', () => {
         });
 
         it('returns the access token without refresh token', async () => {
+            jest.setTimeout(10000);
```
ullasholla, line 73: Is the default value of 5s not enough for this test? Does it fail often?

packages/backend-proxy-middleware/test/base/proxy.test.ts
```diff
@@ -474,9 +474,9 @@ describe('proxy', () => {
             const backend = { url: '', path: '/my/path' } as LocalBackendConfig;
             try {
                 await generateProxyMiddlewareOptions(backend);
-            } catch (error) {
+            } catch (error: any) {
```
ullasholla, line 477: This test confuses me. Will it pass if no error is thrown as expected?

packages/store/test/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
     "extends": "../tsconfig.json",
     "compilerOptions": {
-        "strict": false
+        "strict": false,
+        "experimentalDecorators": true
```
ullasholla, line 5: Don't think we need this, the `tsconfig.json` at the project root already sets this property to `true`.

packages/logger/package.json
```diff
@@ -34,7 +34,8 @@
         "@types/debug": "4.1.7",
         "@types/lodash": "4.14.176",
         "@types/vscode": "1.63.1",
-        "jest-extended": "1.2.0",
+        "jest-extended": "2.0.0",
+        "jest": "27.5.1",
```
ullasholla, line 38: I know we get a complaint about missing peer dependencies here, but can we get away with not having a specific jest version for this package (and store)?

packages/store/package.json
```diff
@@ -42,7 +42,8 @@
         "fast-check": "2.19.0",
         "jest-extended": "3.2.3",
         "memfs": "3.3.0",
-        "typescript": "4.3.5",
+        "jest": "27.5.1",
```
ullasholla, line 45: Same comment as for the logger module: https://github.com/SAP/open-ux-tools/pull/708/files#r1113574214

packages/axios-extension/test/auth/uaa.test.ts
devinea: It is very close to 5s so intermittent, but regular enough.

packages/backend-proxy-middleware/test/base/proxy.test.ts
devinea: Added `expect.assertions(2)` in 111a307c9819a04795cd33565b9807f9d770cfa2

packages/logger/package.json
devinea: `"jest": ">=27.2.5",`
Seems to work better as the version from the root will be used and the warning is gone. 
Updated in 05539429acb5bcbd25dbe45d7fb97a177de7bb51

packages/store/package.json
devinea: `"jest": ">=27.2.5",`
Seems to work better as the version from the root will be used and the warning is gone. 
Updated in 05539429acb5bcbd25dbe45d7fb97a177de7bb51

packages/store/test/tsconfig.json
devinea: Removed in dbac36fcf5c5e9752be2f18a8961439665680bdc

devinea (commented):
@ullasholla Thanks for the review. 
I've addressed comments

packages/backend-proxy-middleware/test/base/proxy.test.ts
ullasholla: Thanks Austin. I know this PR only added the type declaration in the catch clause but this test could have been written differently to start with. We're having to touch this test in the PR and it makes sense to clean it up.

@zdravko-georgiev, tagging you as this may have been an oversight.

Something like this works better?
```typescript
        test('throw an error if proxyOptions.target is not defined', async () => {
            const backend = { url: '', path: '/my/path' } as LocalBackendConfig;
            await expect(() => generateProxyMiddlewareOptions(backend)).rejects.toThrow(
                `Unable to determine target from configuration:\n${JSON.stringify(backend, null, 2)}`
            );
        });
```
Adding `expect.assertions(2)` works but if it's two places to change if the number of assertions change in the future - makes the test a little fragile.

packages/axios-extension/test/auth/uaa.test.ts
ullasholla: Ouch 🙈  
Is there an issue to look at this test?

devinea: Not sure, I am not familiar with this code. 
@tobiasqueck What do you think?

packages/backend-proxy-middleware/test/base/proxy.test.ts
devinea: Thanks @ullasholla updated in 357015f2a22918a31c9f7e50d5568fd9fecf8411

ullasholla (approved):
Thanks @devinea.

We have the one outstanding issue with the brittle test - https://github.com/SAP/open-ux-tools/pull/708#discussion_r1113556629

Flaky tests are a pain and we do need to look at it, not necessarily in this PR

Jimmy-Joseph19 (approved):
Approving upgrade for ui-components package.

tom-sap (approved):

donal-tobin-sap (approved):
Looks good to me.

PR #714, opened 2022-10-05
fix(backend-proxy-middleware): make onProxyReq handler more robust

zdravko-georgiev:
In some rare cases it could happen that the `proxyReq` in https://github.com/SAP/open-ux-tools/blob/main/packages/backend-proxy-middleware/src/base/proxy.ts#L41 is missing the `path` property. If this the case then the proxy request will throw an exception. This PR introduces a check for the `path` property to make the `onProxyReq` handler more robust.

### DIFF
--- a/packages/backend-proxy-middleware/src/base/proxy.ts
+++ b/packages/backend-proxy-middleware/src/base/proxy.ts
@@ -38,7 +38,7 @@ export const ProxyEventHandlers = {
      * @param _options (not used)
      */
     onProxyReq(proxyReq: ClientRequest, _req?: IncomingMessage, _res?: ServerResponse, _options?: ServerOptions) {
-        if (proxyReq.path.indexOf('Fiorilaunchpad.html') !== -1 && !proxyReq.headersSent) {
+        if (proxyReq.path && proxyReq.path.indexOf('Fiorilaunchpad.html') !== -1 && !proxyReq.headersSent) {
             proxyReq.setHeader('accept-encoding', 'br');
         }
     },
--- a/packages/backend-proxy-middleware/src/base/proxy.ts
+++ b/packages/backend-proxy-middleware/src/base/proxy.ts
@@ -38,7 +38,7 @@ export const ProxyEventHandlers = {
      * @param _options (not used)
      */
     onProxyReq(proxyReq: ClientRequest, _req?: IncomingMessage, _res?: ServerResponse, _options?: ServerOptions) {
-        if (proxyReq.path && proxyReq.path.indexOf('Fiorilaunchpad.html') !== -1 && !proxyReq.headersSent) {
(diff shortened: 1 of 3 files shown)


packages/backend-proxy-middleware/src/base/proxy.ts
```diff
@@ -38,7 +38,7 @@ export const ProxyEventHandlers = {
      * @param _options (not used)
      */
     onProxyReq(proxyReq: ClientRequest, _req?: IncomingMessage, _res?: ServerResponse, _options?: ServerOptions) {
-        if (proxyReq.path.indexOf('Fiorilaunchpad.html') !== -1 && !proxyReq.headersSent) {
+        if (proxyReq.path && proxyReq.path.indexOf('Fiorilaunchpad.html') !== -1 && !proxyReq.headersSent) {
```
tobiasqueck, line 41: Can't you merge it to `proxyReq.path?.indexOf('Fiorilaunchpad.html') !== -1

packages/backend-proxy-middleware/test/base/proxy.test.ts
```diff
@@ -107,6 +107,9 @@ describe('proxy', () => {
         test('onProxyReq', () => {
             const mockSetHeader = jest.fn();
 
+            onProxyReq({ setHeader: mockSetHeader as unknown } as ClientRequest);
```
tobiasqueck, line 110: Can you explain this change, please?

tobiasqueck: Oh wait, sorry, I get it. Minor suggestion, move the type case to line 108, so you don't need to do it in 110 and 113

packages/backend-proxy-middleware/src/base/proxy.ts
zdravko-georgiev: Nope, `proxyReq.path?.indexOf('Fiorilaunchpad.html') !== -1` will evaluate to `true` and this will not be correct for this scenario.

packages/backend-proxy-middleware/test/base/proxy.test.ts
zdravko-georgiev: Done with https://github.com/SAP/open-ux-tools/pull/714/commits/69bcf476dcf0e0569f56a9082d0193dd36225e67

packages/backend-proxy-middleware/src/base/proxy.ts
tobiasqueck: Good point, `proxyReq.path?.includes('Fiorilaunchpad.html')` should do the trick.

longieirl (approved):

zdravko-georgiev:
Thanks @tobiasqueck @longieirl

Issue #716, opened 2022-10-05
TBI - exclude pnpm-lock.yaml from prettier formatting


Klaus-Keller:
### Description
To reduce conflicts and allow for easier comparison of changes in `dependencies` and `devDependencies`, we should consider excluding file [pnpm-lock.yaml](https://github.com/SAP/open-ux-tools/blob/main/pnpm-lock.yaml) from prettier formatting.
This could be achieved by adding the file to [.prettierignore](https://github.com/SAP/open-ux-tools/blob/main/.prettierignore).

### Value
- Less merge conflicts
- Simpler comparison of new or modified dependencies

### Architecture Elaboration
n/a

tobiasqueck:
Fixed with #717

PR #717, opened 2022-10-05
chore: exclude pnpm lock from formatting

Klaus-Keller:
https://github.com/SAP/open-ux-tools/issues/716 exclude file [pnpm-lock.yaml](https://github.com/SAP/open-ux-tools/blob/main/pnpm-lock.yaml) from prettier formatting.


devinea (approved):

kranthie-sap (approved):
Looks good to me.

PR #718, opened 2022-10-06
refactor(yaml): rm i18next & add func to interpolate strings

ullasholla:
* Remove dependencies to `i18next` and supporting libraries
* Add a function to interpolate strings. This function can reuse the i18 messages, if required. This function is based on Douglas Crockford's `supplant()` method from Remedial Javascript (http://crockford.com/javascript/remedial.html)
* Increase test coverage 77% -> 96.x% (overall for module, changed code has 91.x% covered)
* Ensure YAML doc modification is only done if there are no errors thrown at a later stage in the method
* Export error codes and new custom error (consumed in odata-service-writer)


ullasholla:
The build breaks because of a test failure.

Line in question: https://github.com/SAP/open-ux-tools/blob/b9322c2a6c49608254459879fa8fb869634d3cc5/packages/odata-service-writer/src/index.ts#L88

We may need to formalize on an error code in the YAML library

packages/yaml/src/errors/index.ts
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Create a new error with `code` correctly set.
+ *
+ * @param options - Options object
+ * @param {YamlError} options.error - YAML error
+ * @param {string} options.message  - optional message for Error.message
+ * @param {StringMap} options.templateReplacements -  values for message template interpolation
+ * @returns {Error} - Error with correct code set
+ */
+export function newError({
```
Klaus-Keller, line 47: @ullasholla, if we need to work with different error types, would it make sense to create an own set of [custom errors](https://javascript.info/custom-errors) extending from `Error` and throw them accordingly?

ullasholla: I'm undecided about that. Subclassing vs error codes were the clear options here. I picked error codes in the end because it gives me finer granularity, I didn't want to have to create multiple error classes to deal with all the different failure modes. In Nodejs, there seems to be a trend towards error codes as well: https://nodejs.org/dist/latest/docs/api/errors.html#errorcode

packages/ui5-config/src/index.ts
```diff
@@ -13,3 +13,5 @@ export {
     UI5ProxyConfig,
     UI5ProxyConfigTarget
 } from './types';
+
+export { errors as yamlErrors } from '@sap-ux/yaml';
```
ullasholla, line 17: This is another place that could use some discussion. Here I re-export YAML errors so that consumers of ui5-config don't need to declare a dependency to `@sap-ux/yaml` just to use the error codes (like `odata-service-writer`)

ullasholla: The question is still valid, even though the original comment is shown as outdated

packages/yaml/src/errors/index.ts
ullasholla: @Klaus-Keller, take another look? I'm used both approaches - a custom error and error codes (for more granular handling)

ullasholla: Link to updated code: https://github.com/SAP/open-ux-tools/blob/f34071a5d3752a6d6b7b65621c2f1c7ff6337381/packages/yaml/src/errors/index.ts#L44-L58

Usage: https://github.com/SAP/open-ux-tools/blob/f34071a5d3752a6d6b7b65621c2f1c7ff6337381/packages/odata-service-writer/src/index.ts#L88-L92

packages/ui5-config/src/index.ts
tobiasqueck: I think it makes sense to do that. I am currently looking at the `abap-deploy` task that uses the `axios-extension`. In my code I needed the type for `AxiosError` from `axios` and since it is part of the signature of one of the `axios-extension` functions, it made sense to me to also re-export it.

packages/yaml/src/errors/index.ts
tobiasqueck: Not a big thing, but why do you need the `newError` function? Cant you do the same in the constructor of `YAMLError`?

ullasholla: > Not a big thing, but why do you need the `newError` function? Cant you do the same in the constructor of `YAMLError`?

I guess I could do that. I needed that `newError()` wrapper before as I had to create the error, set the code and then throw it. I'll take care of it

ullasholla: Fixed in commit: https://github.com/SAP/open-ux-tools/pull/718/commits/a2b3f0661c8550dabf936bf0cb50be209527e2f3

tobiasqueck: Thanks, I would have moved the execution of `interpolate` into the constructor, but this is really just me saving a few lines of code

tobiasqueck (approved):
* removal of unnecessary i18n is cleanly done
* introduced error codes help when using the module
* great improvement of test coverage
* linting only has `Too many nested callbacks` issues which obsolete with #689

kranthie-sap (approved):
The code looks very clean.

PR #719, opened 2022-10-06
Update UILink.tsx

gpeuc:
Swapped underline logic


gpeuc:
@devinea Hey, I did it here as well

PR #720, opened 2022-10-07
Swapped underline logic on text links

gpeuc:
Reversed logic for underline in plaintext links. It's now underlined by default, and on hover underline is removed.


devinea (approved):
Changes look good. I haven't tested locally.

Jimmy-Joseph19 (approved):
Changes looks good. Checked with storybook
image

PR #721, opened 2022-10-07
[CHORE] Replaced `json-merger` with `lodash/mergewith`

tobiasqueck:
Reasoning:
* nested dependency of `json-merger` caused problems when bundling one of the SAP Fiori tools extensions
* `lodash/mergewith` has a significant lower installation size than `json-merger`
* `lodash/mergewith` has a significant higher usage than `json-merger`
* `json-merger` had a dependency to vulnerable version of `vm2` that previously required an override

The PR contains
* replacement of the dependencies
* additional tests for the touched functionality
* fixed incorrect snapshot that was detected when replacing the JSON merging logic
* consistent usage of `lodash` across all modules in the repository


packages/ui5-application-writer/src/data/defaults.ts
```diff
@@ -4,7 +4,7 @@ import { getUI5Libs } from './ui5Libs';
 import semVer from 'semver';
 import type { SemVer } from 'semver';
 import { t } from '../i18n';
-import { mergeObjects } from 'json-merger';
+import merge from 'lodash.mergewith';
```
devinea, line 7:
```suggestion
import { merge } from 'lodash/mergewith';
```

packages/ui5-application-writer/package.json
```diff
@@ -34,16 +34,17 @@
     "dependencies": {
         "@sap-ux/ui5-config": "workspace:*",
         "ejs": "3.1.7",
+        "i18next": "21.6.11",
+        "lodash.mergewith": "4.6.2",
```
devinea, line 38:
```suggestion
        "lodash": "4.17.21",
```

devinea (dismissed):
@tobiasqueck Can you use the main lodash package as per their recommendations ?
https://lodash.com/per-method-packages
`use of these packages is discouraged`
`Although they may seem more lightweight, they will usually increase the size of node_modules and webpack/rollup bundles in a project that transitively depends on multiple per method packages and/or the main lodash package. Whereas many methods in the main lodash package share code, the per method packages internally bundle copies of any code they depend on.`

donal-tobin-sap (approved):
👍🏽 Looks good to me.

Nice to remove json merger,.

devinea (commented):
Thanks @tobiasqueck. Changes look good and tests included. 

@IainSAP - Can you review?

IainSAP (approved):
Thx @tobiasqueck look good. 

- No unexpected snapshot changes in other modules, so assumption is the functionality is identical

It's a pity we couldn't find another 3rd party instead. Next time we need this, again we will have to replicate the code or spend time once again looking for another module (or creating one)

tobiasqueck:
> It's a pity we couldn't find another 3rd party instead. Next time we need this we will have to replicate the code or spend time once again looking for another module (or creating one)

Yeah, however, even with `json-merger` we had to write our own deduping.

PR #729, opened 2022-10-11
Correct UI components dependencies

tom-sap:
I believe this is causing some monorepo-related build issues.
Because these were specified as `devDependencies` they were not installed correctly in the consuming components (and they were also not bundled correctly with `ui-component`).
With that - `ui-components` were using these packages with different versions (as the modules consuming `ui-components` used them but in different versions)

By moving these `devDependencies` to `dependencies` we're forcing `yarn` (in the consumers) to also install the versions of these dependencies that the `ui-components` requires.

@devinea


packages/ui-components/package.json
```diff
@@ -37,15 +37,16 @@
         "test-u": "jest --ci --forceExit --colors  --silent -u"
     },
     "dependencies": {
+        "@fluentui/react": "8.66.2",
         "react-movable": "2.4.0",
         "react-virtualized": "9.22.3",
-        "sanitize-html": "2.7.1"
+        "sanitize-html": "2.7.1",
+        "uuid": "3.4.0"
     },
```
devinea, line 45: Per https://stackoverflow.com/questions/30451556/what-is-the-correct-way-of-adding-a-dependency-to-react-in-your-package-json-for we should probably add this. What do you think?

```suggestion
    },
    "peerDependencies": {
        "react": ">16.14.0",
        "react-dom": ">16.14.0"
        },
```

devinea (approved):
Thanks @tom-sap
Changes look good.

Jimmy-Joseph19 (approved):
Locally built PR. Tested with storybook.

Issue #724, opened 2022-10-07
BUG (Windows) - Gather environment information is not detecting installed extensions


cianmSAP:
### Related Feature
Feature request: https://github.com/SAP/open-ux-tools/issues/694

### Description
When ran in VSCode the environment check is failing to detect the extensions as the spawn command is exting. 

### Steps to Reproduce
1. Go to VSCode
2. Open Command pallete : Fiori : Open Environment Check 
3. Click Gather Environment Information and then View Results

### Expected results
That extension versions are shown

### Actual results
Extension versions show Not installed

### Screenshots
![972b6c58-d997-4ef3-b805-548b4e4d947a](https://user-images.githubusercontent.com/99677697/194594368-2eb6f412-6fb8-4485-9152-a6dd8ca71e0a.png)

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

cianmSAP:
Issue is resolved.

PR #733, opened 2022-10-12
fix(env-check): fix npm root command

cianmSAP:
#724


devinea (commented):
@cianmSAP There is a windows build running on the PR. Is there a test that can be added to check for windows issue like this?

devinea (approved):
As discussed, fine for now. 
Please follow up with TBI for integration test. Scope of that can be discussed later. 
E.g. Testing getting the npm global dir or installing the generator and getting the full path to that. 

Thanks @cianmSAP

PR #736, opened 2022-10-12
feat(ui-components): new property "secondary" for UILink component

815are:
Adding requested new property "secondary" for UILink component.
Difference between primary and secondary is only coloring:
image


Jimmy-Joseph19 (approved):
Changes looks good.tested and verified with Storybook.

devinea (approved):
Changes look good. Tests included. Haven't tested locally.

Issue #739, opened 2022-10-17
BUG - Implicit conversion of file names at FPM writer


stefanschreck:
### Description
When creating new custom columns or custom sections, the FPM writer does an implicit file name conversion: the first letter gets converted to lower case.
This leads to problems at our consuming application: we do a check against existing files to prevent using the same fragment twice, by different extensions. But e.g. Frg2 and frg2 mean two different  files, so if the user enters Frg2 for the new filename it is accepted if only frg2 exists. But the FPM writer then translates Frg2 to frg2 and thus overwrites the content of an already existing file.

### Steps to Reproduce
Call `generateCustomColumn` with name = Frg2 and check which file gets generated.

### Expected results
No implicit conversion takes place, the file gets generate with the same name as specified by the using application. 

### Actual results
Implicit conversion for the file name, as well as for the folder name if not explicitly specified.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [X] Mac OS
- [ ] Windows
- [ ] Other

## Root Cause Analysis 
![image](https://user-images.githubusercontent.com/22190895/196116269-03054807-8fd7-445b-9b88-9aa2c8c51728.png)

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #740, opened 2022-10-17
fix: implicit filename conversion

stefanschreck:
Fix for issue: https://github.com/SAP/open-ux-tools/issues/739

Disables the implicit conversion of file and folder names at the FPM writer.


tobiasqueck (changes requested):
* correction is straight forward and easy to understand, no new tests needed
* thanks for fixing the typo in the tests

So, the change is good, but a patch changeset is required

tobiasqueck (approved):
Thanks for the changeset.

815are (approved):
looks ok, but did not test locally with real app.
My small concern is about unit tests - I do not see any test which would check expected behaviour or am I missing it?
For example if I revert code(in 'common/defaults.ts') back to `ext/${firstChar.toLocaleLowerCase()`, then still all tests are passing.

tobiasqueck (changes requested):
@815are makes a good point with https://github.com/SAP/open-ux-tools/pull/740#pullrequestreview-1148984487. Please add a test covering the renaming scenario.

stefanschreck:
> @815are make a good point with [#740 (review)](https://github.com/SAP/open-ux-tools/pull/740#pullrequestreview-1148984487). Please add a test covering the renaming scenario.

Yes, a good point. I will add further tests

stefanschreck:
When trying to commit I face the following vulnerability:
image
==> I had to add an override on root level.

815are (dismissed):

tobiasqueck:
> When trying to commit I face the following vulnerability: image ==> I had to add an override on root level.

This needs to got to a separate PR also including an update of the override doc. Working on it with #750

packages/fe-fpm-writer/test/unit/action.test.ts
```diff
@@ -275,6 +275,19 @@ describe('CustomAction', () => {
...
 
+            test('"eventHandler" is "object", action with lowercase first letter', () => {
+                generateCustomActionWithEventHandler(name, {
+                    fnName: 'dummyOnAction'
+                });
+
+                const manifest = fs.readJSON(join(testDir, 'webapp/manifest.json'));
+                const action = getActionByName(manifest, name);
+                expect(action['press']).toEqual('my.test.App.ext.myCustomAction.MyCustomAction.dummyOnAction');
+                expect(
```
tobiasqueck, line 286: To we need to store a snapshot here? Isn't the purpose of the test to check the existence of the file at the correct path?

packages/fe-fpm-writer/test/unit/column.test.ts
```diff
@@ -150,6 +150,66 @@ describe('CustomAction', () => {
...
+            const testCustomColumn: CustomTableColumn = {
+                ...customColumn,
+                name: 'newCustomColumn',
+                folder: 'extensions/custom'
+            };
+
+            const testFS = generateCustomColumn(testDir, { ...testCustomColumn, minUI5Version: '1.102' });
+            const updatedManifest: any = testFS.readJSON(join(testDir, 'webapp/manifest.json'));
+
+            const settings = updatedManifest['sap.ui5']['routing']['targets']['sample']['options']['settings'];
```
tobiasqueck, line 163: Same comment as before, do we really need to test the snapshots here? This test is about the correct naming, so let's focus on that in the test.

tobiasqueck (changes requested):

packages/fe-fpm-writer/test/unit/column.test.ts
stefanschreck: @tobiasqueck Sorry, I don't get the point. I'm testing the outcome of the function call, same as the other tests are doing. Should be the mem-fs content.

tobiasqueck: @stefanschreck let me try again. The purpose of this bug fix is that the filename patter (lower/upper case) is not enforced anymore. We already have sufficient tests to check that the generated outcome (content of file) is correct. What we asked you to add, was a test verifying the changed functionality. Technically, you could also do a snapshot test of the whole webapp folder but you don't because it would be harder to maintain and review. 
So, please focus in your additional test on the functionality that actually needs to be tested, and not functionality that is already covered in other tests.

stefanschreck: Ok - as the filename is just part of the  file path, a simple existence check will be sufficint.

packages/fe-fpm-writer/test/unit/section.test.ts
```diff
@@ -124,6 +124,88 @@ describe('CustomSection', () => {
...
+
+        test('no handler, no fs, no folder, section name uppercase', () => {
+            const testCustomSection: CustomSection = {
+                ...customSection,
+                folder: undefined,
+                name: 'NewCustomSection'
+            };
+
+            const testFS = generateCustomSection(testDir, { ...testCustomSection });
+            const updatedManifest: any = testFS.readJSON(join(testDir, 'webapp/manifest.json'));
```
tobiasqueck, line 170: We discourage the use of `any`. The `project-access` module has a type for `Manifest` that you can use here. Please use it.

packages/fe-fpm-writer/test/unit/section.test.ts
```diff
@@ -124,6 +124,88 @@ describe('CustomSection', () => {
...
+            expect(settings.content).toMatchSnapshot();
+
+            const fragmentPath = join(
+                testDir,
+                `webapp`,
+                `${testCustomSection.folder}`,
+                `${testCustomSection.name}.fragment.xml`
+            );
+
+            expect(testFS.read(fragmentPath)).toMatchSnapshot();
```
tobiasqueck, line 206: Same comment as before, isn't `expect(testFS.exists(fragmentPath)).toBe(true);` what we actually care about in this test?

tobiasqueck (changes requested):
Looks better. Please don't use `any` if we know the expected type and if we have a type definition for it. Also, only use snapshots in tests if the content is what you are testing against, don't use it for file existence tests.

packages/axios-extension/src/abap/adt-catalog/handlers/transport.ts
```diff
@@ -102,6 +102,7 @@ function getLockedTransport(doc: Document): TransportRequest[] {
 
 /**
  * Convert transport request in XML element of ADT response to typed object.
+ *
```
tobiasqueck, line 105: Of course it is only a comment change (most likely done automatically with `lint --fix`) but still I would prefer not to have it in this PR, so that all your changes are limited to one module.

tobiasqueck:
This PR doesn't contain any changes to a `package.json` therefore, it shouldn't contain a change to the `pnpm-lock.yaml`.

packages/fe-fpm-writer/test/unit/section.test.ts
stefanschreck: Not so easy, as the Manifest definition is not detailed enough when it comes to routing.targets (target details = unknown). But I give it a try.

packages/fe-fpm-writer/test/unit/section.test.ts
stefanschreck: done

packages/fe-fpm-writer/test/unit/section.test.ts
stefanschreck: Was not so easy, as there is no dedicated definition for routing.targets, but I find some workaround. Please check.

packages/fe-fpm-writer/test/unit/section.test.ts
stefanschreck: done

packages/axios-extension/src/abap/adt-catalog/handlers/transport.ts
stefanschreck: Not relevant anymore, file was deleted, done by merge from main

stefanschreck:
> This PR doesn't contain any changes to a `package.json` therefore, it shouldn't contain a change to the `pnpm-lock.yaml`.

I restored from main

tobiasqueck (approved):
Thanks for the improvements of the tests.

packages/fe-fpm-writer/test/unit/__snapshots__/column.test.ts.snap
```diff
@@ -291,20 +291,6 @@ exports[`CustomAction generateCustomColumn version 1.85, no handler, all propert
...
-        "header": "col header",
-        "position": Object {
-          "anchor": "DataField::BooleanProperty",
-          "placement": "After",
-        },
-        "template": "sapux.fe.fpm.writer.test.extensions.custom.NewCustomColumn",
-        "width": "150px",
-      },
-    },
-  },
```
815are, line 307: @stefanschreck 

test `version 1.85, no handler, no fs, all properties`.
In test we are adding new custom column `NewCustomColumn`:
image

but in test expectation `NewCustomColumn` is not appearing. I think `NewCustomColumn` should appear in test snapshot for `version 1.85, no handler, no fs, all properties` like it was before

stefanschreck: Thanks, Andis. Yes, one of the tests needed a correction.

815are (approved):

zdravko-georgiev (approved):

815are (approved):

zdravko-georgiev (approved):

Issue #746, opened 2022-10-19
FEATURE - Enable TS support for SAPUI5 application


longieirl:
**As a Fiori developer using the basic SAPUI5 template in the Fiori generator, I'd like the option to generate the project with TypeScript support.**

### Description

With the delivery of `@sap/generator-fiori#21585` (internal issue), TypeScript is now an advanced option during generation. In order to support TechEd workshops, we need to enable this TypeScript option for the SAPUI5 Basic template only.

This is a requirement for the TechEd workshops, demonstrating Typescript support inside Fiori Freestyle applications.

### Technical Design
Core changes to `ui5_application_writer` and should only impact Fiori Freestyle writer, all other writers should not be affected.

The generated package.json for the Fiori Freestyle application should support typescipt build step for both the `npm run start` and `npm run build` commands.

### Acceptance Criteria
Given I launch the generator and choose the `Freestyle` -> `SAPUI5 Application` as the template
when the Fiori App is generated
then the application will support Typescript 

Given I have a Freestyle application
when  I run the Freestyle `start` and `build` commands
then Typescript command should also be executed but only for SAPUI5 template

Given I have a Freestyle application that is dependent on npmjs modules `@sapui5/ts-types-esm` and `@sapui5/ts-types`
when  the minimium UI5 version is specified
then the install phase should install versions that are available on npmjs and should not fail the install phase

### Tasks
Required
- [x] Ensure`ts-typecheck` is run as part of `npm run start` and `npm run build` steps, restricted only for Basic SAPUI5 template
- [x] Validate all npm modules added to `package.json` are available on npm.js for both `@sapui5/ts-types-esm` and `@sapui5/ts-types` otherwise the Fiori app install will fail 
- [x] Unit Tests

Nice to have
- [ ] Include fallbacks for offline usage so if user is restricted from accessing npmjs, the generation should not fail

longieirl:
There is a change in the behaviour, if a user selects TS and code assist during the project generation flow, then the types version specified for both will be aligned, using the [ESM types version](https://www.npmjs.com/package/@sapui5/ts-types-esm) as the min version. However, if a user generates a SAPUI5 app with only code assist, then the types min version will be from [@sapui5/ts-types](https://www.npmjs.com/package/@sapui5/ts-types)

PR #747, opened 2022-10-19
feature(ui5-app-writer): Support TS for SAPUI5 apps

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/746

- better support of `@sapui5/ts-types-esm` to ensure a valid package is aways chosen during install phase of Fiori app
- unit tests added and updated
- `pre` ts steps added to esnure tsc compile is run before `npm run start` and `npm run build`


tobiasqueck (changes requested):
Please rework #746 so that it is clear why these changes should be accepted.

packages/ui5-application-writer/templates/optional/typescript/package.json
```diff
@@ -1,5 +1,7 @@
 {
     "scripts": {
+        "prestart": "npm run ts-typecheck",
+        "prebuild": "npm run ts-typecheck",
```
devinea, line 4: @jlongie as discussed, can we limit this to the FF SAPUI5 floorplan only for now?

packages/ui5-application-writer/src/data/defaults.ts
```diff
@@ -120,22 +120,20 @@ export function mergeUi5(ui5: Partial<UI5>): UI5 {
...
 /**
- * Get the best types version for the given minUI5Version.
+ * Get the best types version for the given minUI5Version within a selective range, starting at 1.90.0 as the minimum
  * For the latest versions the LTS S/4 on-premise version (1.102.x) is used, for anything before we match the versions as far back as available.
  *
- * @param minUI5Version the mininum UI5 version that needs to be supported
+ * @param minUI5Version the minimum UI5 version that needs to be supported
  * @returns semantic version representing the types version.
  */
 export function getTypesVersion(minUI5Version?: string) {
```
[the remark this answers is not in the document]
longieirl: Addressed in latest PR.

packages/ui5-application-writer/templates/optional/typescript/package.json
longieirl: Addressed in latest PR.

packages/ui5-application-writer/src/data/defaults.ts
```diff
@@ -113,29 +114,54 @@ export function mergeUi5(ui5: Partial<UI5>): UI5 {
...
     merged.ui5Libs = getUI5Libs(ui5.ui5Libs);
 
     return Object.assign({}, ui5, merged) as UI5;
 }
 
 /**
- * Get the best types version for the given minUI5Version.
+ * Get the best types version for the given minUI5Version for https://www.npmjs.com/package/@sapui5/ts-types where specific versions are missing
  * For the latest versions the LTS S/4 on-premise version (1.102.x) is used, for anything before we match the versions as far back as available.
  *
```
devinea, line 127:
```suggestion
```
this relates to the esm module not the one for JS.

packages/fiori-elements-writer/test/__snapshots__/fpm.test.ts.snap
```diff
@@ -3669,7 +3669,7 @@ archive.zip
     \\"@ui5/cli\\": \\"^2.14.1\\",
     \\"@sap/ux-ui5-tooling\\": \\"1\\",
     \\"@sap/ux-specification\\": \\"latest\\",
-    \\"@sapui5/ts-types-esm\\": \\"~1.96.11\\",
+    \\"@sapui5/ts-types-esm\\": \\"~1.96.0\\",
```
devinea, line 3672: why is the patch dropped? 
more or less the same....

packages/ui5-application-writer/templates/optional/typescript/package.json
```diff
@@ -3,7 +3,7 @@
         "ts-typecheck": "tsc --noEmit"
     },
     "devDependencies": {
-        "@sapui5/ts-types-esm": "<%- ui5.typesVersion %>",
+        "@sapui5/ts-types-esm": "<%- ui5.esmTypesVersion %>",
```
devinea, line 6: 👍

packages/ui5-application-writer/test/data.test.ts
```diff
@@ -263,7 +275,8 @@ describe('Setting defaults', () => {
                 localVersion: '1.76.0',
                 minUI5Version: '1.28.0',
                 descriptorVersion: '1.12.0',
-                typesVersion: UI5_DEFAULT.TYPES_VERSION_PREVIOUS,
+                typesVersion: `~1.71.18`,
+                esmTypesVersion: `~${UI5_DEFAULT.ESM_TYPES_VERSION_SINCE}`,
```
devinea, line 279: Bug?

packages/ui5-application-writer/test/data.test.ts
```diff
@@ -140,7 +145,8 @@ describe('Setting defaults', () => {
                 localVersion: '1.95.6',
                 minUI5Version: '1.80.0',
                 descriptorVersion: '1.12.1',
-                typesVersion: '1.95.5',
+                typesVersion: '1.95.0',
+                esmTypesVersion: `1.90.0`,
```
devinea, line 149: this is a bit un expected. Why a lower version for esm here?

devinea (commented):
Thanks @jlongie 
Tested locally and works much better for `packages/fiori-freestyle-writer/test/test-output/basicTemplate/basic_typescript` now. 
A few comments below.

packages/fiori-elements-writer/test/__snapshots__/fpm.test.ts.snap
longieirl: With the latest changes to both types method, if a version is specified that is within the correct range, the patch is dropped, this will ensure the latest of the minor release version is used. There is no guarntee that the patch version supplied will match what is available in `@sapui5/ts-types-esm` or `@sapui5/ts-types` version tree.

packages/ui5-application-writer/test/data.test.ts
longieirl: Updated now.

packages/ui5-application-writer/test/data.test.ts
longieirl: This test is ensuring the passed in values are used rather than relying the minUi5Version to generate them.

packages/fiori-elements-writer/test/__snapshots__/fpm.test.ts.snap
devinea: Good point. Thanks

devinea (approved):
Thanks @longieirl 
Changes and tests look good.

packages/ui5-application-writer/templates/optional/typescript/package.json
tobiasqueck: @devinea could you please explain why you would only do it for the basic template?

packages/ui5-application-writer/src/data/defaults.ts
```diff
@@ -120,22 +120,20 @@ export function mergeUi5(ui5: Partial<UI5>): UI5 {
...
-        return `~${UI5_DEFAULT.TYPES_VERSION_BEST}`;
-    } else if (semVer.gte(version, UI5_DEFAULT.TYPES_VERSION_BEST)) {
+    if (!version || semVer.gte(version, UI5_DEFAULT.TYPES_VERSION_BEST)) {
         return `~${UI5_DEFAULT.TYPES_VERSION_BEST}`;
     } else {
         return semVer.gte(version, UI5_DEFAULT.TYPES_VERSION_SINCE)
-            ? `~${semVer.major(version)}.${semVer.minor(version)}.${semVer.patch(version)}`
-            : UI5_DEFAULT.TYPES_VERSION_PREVIOUS;
+            ? `~${semVer.major(version)}.${semVer.minor(version)}.0`
+            : `~${UI5_DEFAULT.TYPES_VERSION_PREVIOUS}`;
```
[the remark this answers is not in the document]
tobiasqueck: Any reason why we are using 0 and not the patch version provided?

packages/ui5-application-writer/src/types.ts
```diff
@@ -47,6 +47,7 @@ export interface UI5 {
     version: string;
     localVersion: string;
     typesVersion: string;
+    esmTypesVersion: string;
```
tobiasqueck, line 50: When using typescript, we use the `esm` types, otherwise we are using the old experimental types, however, there is never a case where we need both, therefore, please use the property `typesVersion` and don't introduce a new one here.

tobiasqueck (changes requested):

packages/fiori-freestyle-writer/src/packageConfig.ts
```diff
@@ -63,6 +66,7 @@ export function getPackageJsonTasks({
             ? {
                   'start-mock': mockTask
               }
-            : {}
+            : {},
+        addTypeScript ? { prestart: 'npm run ts-typecheck', prebuild: 'npm run ts-typecheck' } : {}
```
tobiasqueck, line 70: Don't add the scripts programmatically but add them to the typescript template at https://github.com/SAP/open-ux-tools/blob/feat/746/support-ts-sapui5-apps/packages/ui5-application-writer/templates/optional/typescript/package.json#L3. Please also don't use `npm...` but use the same script as use for the `ts-typecheck` script so that generated apps can also be used with `yarn` or `pnpm` or ...

packages/ui5-application-writer/src/data/defaults.ts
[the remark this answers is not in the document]
longieirl: Since we have no guarantee the minUI5version will match the corresponding types versions, its a safety net and future patch versions will be used since the tilda is also specified.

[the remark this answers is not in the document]
tobiasqueck: Ok, good point, using internal knowledge, we would know that the release of a types version is bound to a release of a UI5 version, so they always exist together, however, as you said, due to the ~ it doesn't make a difference.

packages/fiori-freestyle-writer/src/packageConfig.ts
tobiasqueck: @devinea which ones do have problems? Should we fix those instead?

devinea: @tobiasqueck I haven't checked today. But from what I recall type checking is failing on FE V2 templates as we are waiting for more types to be exposed for `sap.ui.generic.AppComponent` 
Also, we probably need more tests to validate the types for the templates. 
The plan is for now to enable the type checking for the simple template. Then add the others in another iteration.
Obviously the `ts-typecheck` is still available to run manually in the meantime.

devinea: I don't think i've seen problems using `npm run` with yarn or pnpm. 
Are you suggesting that we have this instead?
```JSON
{
"ts-typecheck": "tsc --noEmit",
"prestart": "tsc --noEmit",
"prebuild": "tsc --noEmit"
}
```

packages/ui5-application-writer/templates/optional/typescript/package.json
devinea: See https://github.com/SAP/open-ux-tools/pull/747#discussion_r1000661047

packages/fiori-freestyle-writer/src/packageConfig.ts
tobiasqueck: Yes, that was my suggestion. I actually assumed it doesn't work if you'd do in a monorepo where the modules might not be in `./node_modules` but to be honest I didn't check.

packages/ui5-application-writer/src/data/defaults.ts
tobiasqueck: Starting with `1.90` both are always generated whenever a new UI5 version is generated. @petermuessig correct me if I am stating something wrongly here.

packages/ui5-application-writer/src/options.ts
```diff
@@ -5,13 +5,13 @@ import type { Ui5App } from './types';
...
 import { ui5NPMSupport, ui5TSSupport } from './data/ui5Libs';
-import { mergeObjects, UI5_DEFAULT } from './data/defaults';
+import { getEsmTypesVersion, mergeObjects, UI5_DEFAULT } from './data/defaults';
 
 /**
  * Input required to enable optional features.
  */
 export interface FeatureInput {
-    ui5App: { app: { id: string; baseComponent?: string } };
+    ui5App: { app: { id: string; baseComponent?: string }; ui5?: { esmTypesVersion?: string; minUI5Version?: string } };
```
longieirl, line 14: Using the FeatureInput to handle the change locally. The input object is being shared across the different factories so a new variable introduced to handle this particular flow for the specific TS template.

packages/fiori-freestyle-writer/src/packageConfig.ts
devinea: There is duplication. But, it is shorter than using `npm run ts-typecheck`  so 👍 .

packages/ui5-application-writer/src/data/defaults.ts
```diff
@@ -120,22 +121,43 @@ export function mergeUi5(ui5: Partial<UI5>): UI5 {
...
+ * Get the best types version for the given minUI5Version for https://www.npmjs.com/package/@sapui5/ts-types where specific versions are missing.
  *
- * @param minUI5Version the mininum UI5 version that needs to be supported
+ * @param minUI5Version the minimum UI5 version that needs to be supported
  * @returns semantic version representing the types version.
  */
 export function getTypesVersion(minUI5Version?: string) {
     const version = semVer.coerce(minUI5Version);
-    if (!version) {
+    if (!version || semVer.gte(version, UI5_DEFAULT.TYPES_VERSION_BEST)) {
```
tobiasqueck, line 131: This or condition `|| semVer.gte(version, UI5_DEFAULT.TYPES_VERSION_BEST)` was left out on purpose. The reasoning is that it is not possible (with reasonable effort) to ship a type fix for a none LTS version once it is shipped, therefore, the best option for a project that e.g. uses UI5 version `1.103` is to use `1.02.x`, therefore, please remove it again.
Btw, we go a similar approach with our versions for the @sap/ux-specification.

packages/ui5-application-writer/src/index.ts
```diff
@@ -115,7 +114,7 @@ async function enableTypescript(basePath: string, fs?: Editor): Promise<Editor>
     const ui5App = {
         app: manifest['sap.app'],
         ui5: {
-            typesVersion: getTypesVersion(manifest['sap.ui5']?.dependencies?.minUI5Version)
+            minUI5Version: manifest['sap.ui5']?.dependencies?.minUI5Version
```
tobiasqueck, line 117: I am not sure I understand this change and the changes required down the line. If you just change this line back to what it was, with a minor fix : `typesVersion: getEsmTypesVersion(manifest['sap.ui5']?.dependencies?.minUI5Version)` then you don't need any changes in
packages/ui5-application-writer/src/options.ts and
packages/ui5-application-writer/templates/optional/typescript/package.json

tobiasqueck (changes requested):

packages/fiori-freestyle-writer/src/packageConfig.ts
tobiasqueck: >But from what I recall type checking is failing on FE V2 templates as we are waiting for more types to be exposed for sap.ui.generic.AppComponent
Also, we probably need more tests to validate the types for the templates.

Ok, I give in here, but create a follow up issue with the requirement to enable this for all templates with the tasks
- remove conditional code added in this PR
- enhancing the template in `ui5-application-writer`
- validate and fix all provided ts templates in `fiori-*-writer` to throw no errors when running a typecheck

packages/ui5-application-writer/src/index.ts
longieirl: So there are a few things I've noticed here;
1. `enableTypescript` is an exposed method by the ui5-app-writer, used when adding TS to an existing Fiori app
2. It calls `enableTypescriptOption` which in turn calls`enableTypescript` from the factories flow which adds the actual templates
3. Your change will work for the `adding to an existing Fiori app` but the factories flow is used when TS is added as part of the options flow, used when creating the app for the first time
4. The optional flow, is passing the FeatureInput object as pass by reference thus if I overwrite the typesVersion in the factories `enableTypescript` method it will use that version also for the `@sapui5/ts-types` version.

One solution is to clone the FeatureInput object in the `enableTypescript` factories method but the minUI5Version still needs to be exposed.

tobiasqueck: I need to try it myself with https://github.com/SAP/open-ux-tools/pull/751. The only problem that my proposal caused was in the test case where we enable `typescript` and `codeAssistant` in the same config. However, the test that my change broke was actually a problem.
Was your current code, the test results in the generation of a project with a types version of 1.71 and an esm types version of 1.90 which is not a good idea. Both should have the same version. If it is a ts project, then both should use the esm version. 
Please check https://github.com/SAP/open-ux-tools/pull/751 as a proposal to fix that inconsistency.

devinea (approved):
Thanks @longieirl 
Change in behaviour and tests look good. 
Thanks for creating follow-up TBI for other templates.

tobiasqueck (approved):
* new test case covers the new types version assignment
* old snapshot tests are updated correctly with this change
* follow issue to enable this for all templates has been created (#754)
* additional typos in comments fixed (thanks!)
* changeset is as expected

PR #749, opened 2022-10-20
fix(ui5-proxy-middleware): change setting of proxy env variables

zdravko-georgiev:
Follow up of #727
Refactor the `ui5-proxy-middleware` so that `getProxyFromUrl` is used directly in the `proxy.ts` and to avoid duplicating code when consuming in `@sap/ux-ui5-tooling`.
Changes the way proxy env variables are set:
- command line arguments should be set to process.env and always win
- config should also be set to  process.env, not overwriting existing process.env settings
- process.env is to be used


Klaus-Keller (approved):
Hi @zdravko-georgiev,
- code looks good
- test coverage is excellent
- did not test locally

Thanks!

tobiasqueck (approved):
As discussed in the chat, code looks good, tests make sense, so we are good to go.
In general, I think we have too many options for proxy / no proxy variables that makes the code overly complicated. We should discuss reducing some options to simplify all this.

PR #751, opened 2022-10-21
Proposal feat/746

tobiasqueck:


longieirl:
Those changes look good. My only question is, `@sapui5/ts-types` is missing a number of versions! Thus if a user selects UI5 version 1.83 and enables TS and codeassit, the project generation install phase will fail since there is a dependency missing.

longieirl:
@tobiasqueck to understand this change, it means if a user selects TS and codeassit, minimium type will be that of the esm version for both. If the user selects just codeassit then it will use the `@sapui5/ts-types` min version? 

If my comment is correct and that is the expected behaviour, then Im happy.

longieirl (approved):

tobiasqueck:
> If my comment is correct and that is the expected behaviour, then Im happy.

Correct. Feel free to merge it into your branch

PR #752, opened 2022-10-21
chore(open-ux-tools) Enable eslint import/no-extraneous-dependencies

devinea:
chore(open-ux-tools) Enable eslint import/no-extraneous-dependencies

Ensure that packages imported the a module are properly included in the package.json.


.eslintrc
```diff
@@ -264,6 +264,7 @@
         //		"sap-no-jquery": 2
         // turn on errors for missing imports
         "import/no-unresolved": "error",
+        "import/no-extraneous-dependencies": ["error", {"devDependencies": true, "optionalDependencies": true, "peerDependencies": true}],
```
Klaus-Keller, line 267: @devinea, I think the rule makes sense. I'm just wondering why we set options?
According to https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md#options the default for all 4 types of dependencies 
```
devDependencies
optionalDependencies
peerDependencies
bundledDependencies
```
is `true`. What is the reason for setting it for 
```
devDependencies
optionalDependencies
peerDependencies
```
manually to `true` and skip option for `bundledDependencies`?

ullasholla (dismissed):
Thanks @devinea

kranthie-sap:
@devinea: Since we use `pnpm` is this rule relevant?
https://pnpm.io/pnpm-vs-npm#installation

devinea:
> @devinea: Since we use `pnpm` is this rule relevant? https://pnpm.io/pnpm-vs-npm#installation

That was might assumption as well. But enabling it found a package.json that was missing declared devDeps.

.eslintrc
Klaus-Keller: I agree we should avoid `bundledDependencies` at all. The documentation of this rule says:

> `bundledDependencies`: If set to `false`, then the rule will show an error when `bundledDependencies` are imported. Defaults to `true`.

In this case, should not set `"bundledDependencies": false` then?

devinea: @Klaus-Keller Good point. Updated in https://github.com/SAP/open-ux-tools/pull/752/commits/4af69f48b264f005427f87be8953d43851074e90

Klaus-Keller (approved):
New eslint rule helps to spot missing/extraneous dependencies and improves code quality. 
Approved.

ullasholla (approved):

devinea:
Thanks @Klaus-Keller @ullasholla

PR #753, opened 2022-10-21
[SECURITY] override for minimatch for two modules that use the vulnerable version

devinea:
More info at https://github.com/advisories/GHSA-f8q6-p94x-37v3


tobiasqueck (approved):
Thank you! Looks better than my attempt with #750

stefanschreck (approved):

PR #759, opened 2022-10-25
chore(uie): replace vscuie prefix in css vars with vscode

tom-sap:
Cleanup, as requested by design team.


Klaus-Keller:
@tom-sap, what about the style var definitions in https://github.com/SAP/open-ux-tools/blob/main/packages/ui-components/src/styles/_color.scss? Can they be removed? @815are can you confirm?

tom-sap:
@Klaus-Keller thanks for that. I'm pretty sure that this file is not used anymore (and hasn't been for a while now).
Will ask guys to confirm.

815are:
> @tom-sap, what about the style var definitions in https://github.com/SAP/open-ux-tools/blob/main/packages/ui-components/src/styles/_color.scss? Can they be removed? @815are can you confirm?

@Klaus-Keller 
I think idea is to remove it. But we still use those variables in out extensions - so we will need adapt(replace vscuie) in our webview extension code as well. If we will run extension webview code against latest `ui-components` - there would be ui issues

815are (dismissed):
looks ok - did not find any reference to `vscuie`
Tested with storybook - did not notice issues

815are (approved):

Klaus-Keller:
> @Klaus-Keller I think idea is to remove it. But we still use those variables in out extensions - so we will need adapt(replace vscuie) in our webview extension code as well. If we will run extension webview code against latest `ui-components` - there would be ui issues

Thanks for looking into this @815are. I'll create a follow up ticket to cleanup.

Issue #748, opened 2022-10-19
BUG - Environment check Node Version inconsistencies


cianmSAP:
### Related Feature
Feature request: https://github.com/SAP/open-ux-tools/issues/694

### Description
When environment check is consumed by the Application modeler in VSCode and BAS, the Node version displayed is not consistent with the version installed on the system.

### Steps to Reproduce
1. Run environment check on VSCode or BAS and check Node version
2. Open a terminal and run `node -v`
3. See error (node versions are possibly different)

### Expected results
Node version in environment check should be consistent with the version installed and enabled on the system or environment.

### Actual results
They can be different. 

### Screenshots
<img width="920" alt="Screenshot 2022-10-19 at 14 59 14" src="https://user-images.githubusercontent.com/99677697/196712399-dc62050f-e858-4b26-a90a-1ccc3732453e.png">

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

cianmSAP:
Issue is resolved

PR #760, opened 2022-10-25
fix(env-check): fix node process versions

cianmSAP:
https://github.com/SAP/open-ux-tools/issues/748 

The environment check when consumed in VSCode and BAS is returning the node process versions of the environment as per https://github.wdf.sap.corp/ux-engineering/tools-suite/blob/master/docs/ci-cd/supported-node-versions.md

Fix is to run the command to check for the versions installed on the system / environment.


packages/environment-check/src/checks/get-installed.ts
```diff
@@ -152,3 +152,22 @@ export async function getFioriGenVersion(): Promise<string> {
...
 }
+
+/**
+ * Returns the versions of node.js modules.
+ *
+ * @returns modules and versions
+ */
+export async function getProcessVersions(): Promise<{ [id: string]: string }> {
+    const output = await spawnCommand('node', ['-p', 'process.versions']);
+    const versions = output.split('\n');
```
Klaus-Keller, line 163: When I execute following command in terminal
```
$ node -p process.versions
```
I get as result
```
{
  node: '14.20.0',
  v8: '8.4.371.23-node.87',
  uv: '1.42.0',
  zlib: '1.2.11',
  brotli: '1.0.9',
  ares: '1.18.1',
  modules: '83',
  nghttp2: '1.42.0',
  napi: '8',
  llhttp: '2.1.5',
  openssl: '1.1.1q',
  cldr: '40.0',
  icu: '70.1',
  tz: '2021a3',
  unicode: '14.0'
}
```

Would it not be easier to do a `JSON.parse()` of the output instead of applying string operations like `split`, `trim`, `indexOf`?

packages/environment-check/src/types.ts
```diff
@@ -38,7 +38,7 @@ export enum Extensions {
 export interface Environment {
     developmentEnvironment: DevelopmentEnvironment;
     platform: NodeJS.Platform;
-    versions: NodeJS.ProcessVersions;
+    versions: { [id: string]: string };
```
Klaus-Keller, line 41: The return interface of the new way: 
```
node -p process.versions
```
should be equivalent to what we previously did, like 
```
node -e "console.log(process.versions)"
```
I understand that the new approach returns the versions for installed node and the previous approach returns it for the current process, which in case we execute from VSCode returns the versions from VSCode's node, but the return structure should be the same, correct?

Do we need to change the type here?

packages/environment-check/src/checks/get-installed.ts
```diff
@@ -152,3 +152,22 @@ export async function getFioriGenVersion(): Promise<string> {
     }
     return fioriGenVersion;
 }
+
+/**
+ * Returns the versions of node.js modules.
+ *
+ * @returns modules and versions
+ */
+export async function getProcessVersions(): Promise<{ [id: string]: string }> {
```
Klaus-Keller, line 161: Can we return `NodeJS.ProcessVersions` here?

```suggestion
export async function getProcessVersions(): Promise<NodeJS.ProcessVersions> {
```

packages/environment-check/src/checks/environment.ts
```diff
@@ -98,7 +100,7 @@ async function getToolsExtensions(): Promise<{
     const cloudCli = await getCFCliToolVersion();
 
     let toolsExtensions: ToolsExtensions = {
-        nodeVersion: process.versions.node,
+        nodeVersion: nodeVersion,
```
Klaus-Keller, line 103: We have the node version in two places of the result structure:
`EnvironmentCheckResult->environment->versions->node` and
`EnvironmentCheckResult->environment->toolsExtensions->nodeVersion`

I would avoid having it in two places and delete it here in interface `ToolsExtensions`. If we need to output the node version in multiple places in the markdown report, I would just take it from `EnvironmentCheckResult->environment->versions->node` multiple times to add to the report, but not have it twice in JSON results. That avoids also to pass parameter `nodeVersion` into this function just to write it 1:1 into the result structure.

Klaus-Keller (changes requested):

packages/environment-check/src/checks/get-installed.ts
cianmSAP: I had hoped it would be easy to parse the output without much manipulation. But the string returned cannot be parsed as keys are not quoted (`SyntaxError: Unexpected token n in JSON at position 3)`. 

The returned string contains new lines - see after using `JSON.stringify()`

`"{\n  node: '16.17.0',\n  v8: '9.4.146.26-node.22',\n  uv: '1.43.0',\n  zlib: '1.2.11',\n  brotli: '1.0.9',\n  ares: '1.18.1',\n  modules: '93',\n  nghttp2: '1.47.0',\n  napi: '8',\n  llhttp: '6.0.7',\n  openssl: '1.1.1q+quic',\n  cldr: '41.0',\n  icu: '71.1',\n  tz: '2022a',\n  unicode: '14.0',\n  ngtcp2: '0.1.0-DEV',\n  nghttp3: '0.1.0-DEV'\n}\n"`

Am I missing a trick here?

packages/environment-check/src/checks/get-installed.ts
cianmSAP: With the current method we don't know the properties returned, so I've used a type assertion on the returned object

packages/environment-check/src/checks/environment.ts
cianmSAP: Updated, markdown.ts is not quite as dynamic now if we want to leave the Node.js property in the results table

packages/environment-check/src/checks/get-installed.ts
donal-tobin-sap: The string contains '\n' not a new line.  The new line characters have already been escaped into the string.

Klaus-Keller: @cianmSAP, what about running the command as `node -p 'JSON.stringify(process.versions)'` and parsing the output?

cianmSAP: thanks @Klaus-Keller , that works

packages/environment-check/src/types.ts
cianmSAP: Reverted

Klaus-Keller (approved):
Thanks @cianmSAP 
- code is readable and looks good
- test coverage is great
- I've build and tested locally with CLI

Approved.

devinea (approved):
Thanks. Changes look good.

PR #762, opened 2022-10-25
Fix/ui5 app writer/761 handle ns in tsconfig

longieirl:
https://github.com/SAP/open-ux-tools/issues/746

An additional path is added to support `app.id which contains dots i.e. `myname.test.namespace` thus two paths are added to tsconfig;

```
    "paths": {
        "myname/test/namespace/*": [ "webapp/*" ],
      "myname.test.namespace/*": ["webapp/*"]
```


packages/ui5-application-writer/test/options.test.ts
```diff
@@ -129,4 +129,28 @@ describe('UI5 templates', () => {
             }
         });
     });
+
```
tobiasqueck, line 132: Instead of adding a new test, could we add a namespace to the app id of the `baseAppConfig` so that we are checking this case for all options?

packages/ui5-application-writer/templates/optional/typescript/tsconfig.json
```diff
@@ -13,7 +13,11 @@
     "rootDir": "webapp",
     "outDir": "./dist",
     "baseUrl": "./",
-    "paths": {
+    "paths": {<% if (app.id.includes('.')) { %>
```
tobiasqueck, line 16: To simplify the code, I would remove the if, and just always replace `.` with `/`. There is no harm in running on a string without dots.

packages/ui5-application-writer/templates/optional/typescript/tsconfig.json
```diff
@@ -13,7 +13,11 @@
...
+      "<%=app.id.replace(/\./g, '/')%>/*": [
+        "webapp/*"
+      ],
+      <% } %>
       "<%- app.id %>/*": [
```
tobiasqueck, line 21: Why did you leave the old one? Shouldn't this one be replaced with your new code?

longieirl: The idea behind introducing a second path with the IF statement, was to ensure that other references who require `myname.test.namespace` continue to work. I didnt want to do a full replace on the chance this introduced a regression. 
The `component.ts` file in Freestyle writer is the only TS that does the replace;
```
import { createDeviceModel } from "<%=app.id.replace(/\./g, '/')%>/model/models";
```
However, there are other TS files that just reference the app.id;
```
name: "<%- app.id %>.view.ViewSettingsDialog",
```

packages/ui5-application-writer/test/options.test.ts
longieirl: Happy to make this change. The new test was introduced as there were alot of changes to the snapshots if the baseAppConfig ID was changed, wanted to keep this PR to a minimum amount of changes.

packages/ui5-application-writer/templates/optional/typescript/tsconfig.json
tobiasqueck: But that is inconsistent, and we shouldn't do that. To be honest, within the same project the references should all be relative e.g. `import { createDeviceModel } from "./model/models";`
Please clean this up properly, `"<%=app.id.replace(/\./g, '/')%>/*": [ "webapp/*" ]` is the correct line for all projects (it was simply wrong before) an also make sure that in our templates we always use relative paths.

Btw, where is `name: "<%- app.id %>.view.ViewSettingsDialog"`?

tobiasqueck: @petermuessig please check my statements above and let us know if you disagree.

longieirl: @tobiasqueck the reference file https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-freestyle-writer/templates/listdetail/add/webapp/controller/List.controller.ts#L147

tobiasqueck: That is a property that we are giving to the fragment loader. That is not related to typescript. UI5 needs the ID of the fragment and does its own resolution of the file based on the registered path for the corresponding namespace/module.

tobiasqueck: @jlongie I checked the code,  `import { createDeviceModel } from "<%=app.id.replace(/\./g, '/')%>/model/models";` seems the only one that needs fixing and should be `import { createDeviceModel } from "./model/models";`.
The rest is good to go.

tobiasqueck (dismissed):
* changes are clean and easy to understand
* new tests are very useful

I did not generate a project and do a manual test, if you did this, then I am ok with the PR.

tobiasqueck (dismissed):

longieirl:
> * changes are clean and easy to understand
> * new tests are very useful
> 
> I did not generate a project and do a manual test, if you did this, then I am ok with the PR.

@tobiasqueck Yes, I tested these changes and the generated SAPUI5 works as expected both building and in local preview. Thanks.

.changeset/proud-lions-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/ui5-application-writer': patch
```
devinea, line 2: Add freestyle writer here or I suppose it will be bumped anyway as app writer is a dependency?

packages/fiori-freestyle-writer/test/__snapshots__/basic.test.ts.snap
```diff
@@ -1024,8 +1024,8 @@ builder:
     "state": "modified",
   },
   "webapp/Component.ts": Object {
-    "contents": "import BaseComponent from \\"sap/ui/core/UIComponent\\";
-import { createDeviceModel } from \\"nods1/model/models\\";
+    "contents": "import BaseComponent from 'sap/ui/core/UIComponent';
```
devinea, line 1027: Not sure the double to single quotes should be applied here. At least for JS projects double quotes should be used. TS should be the same.

packages/ui5-application-writer/templates/optional/typescript/tsconfig.json
```diff
@@ -14,7 +14,7 @@
     "outDir": "./dist",
     "baseUrl": "./",
     "paths": {
-      "<%- app.id %>/*": [
+      "<%=app.id.replace(/\./g, '/') %>/*": [
```
devinea, line 17: Should we keep `<%-` here? @IainSAP wasn't that the recommendation for the templates?
https://stackoverflow.com/a/11025348

devinea (commented):
Thanks @longieirl 
Changes look good. Few small comments.

devinea (approved):
Thanks @longieirl 
Changes look good.

tobiasqueck (approved):

PR #766, opened 2022-10-27
Update Icons.tsx

aislingnoone:
Update to info icon - icon made bigger


devinea (dismissed):
Thanks @aislingnoone 
Changes look good
There is no change set in the pr so release won’t be immediate. @gpeuc mentioned this is ok

tom-sap (dismissed):

815are (dismissed):

tobiasqueck (changes requested):
Changeset is missing

gpeuc:
@devinea @tobiasqueck So, we don't really have any idea what changeset is. If I go to that "Learn what changeset is" link, it basically starts with

`Run the command line script npx changeset or yarn changeset.`

and this is where you lost us. Especially because we are not doing these changes through any local tools, rather, we are doing all this through GitHub code editor directly here. So it's rather pointless to do command line stuff, when our changes are directly in the browser.

Is there a way for us to do these things using just GitHub.com as a tool?

tobiasqueck:
> Is there a way for us to do these things using just GitHub.com as a tool?

@gpeuc  Yes, just click on `Click here if you're a maintainer who wants to add a changeset to this PR` and then update the ext in line 5++. Whatever you type there will be the text added to the changelog.

devinea (approved):
Thanks @aislingnoone

gpeuc (approved):

PR #768, opened 2022-10-28
Add linting action with auto-fix

devinea:
Automatically run lint fix and commit the changes back to the current branch as a new action on each PR into main or commit to main. 
This is useful for designers who edit files directly in github.com and devs who might forget to run lint fix before pushing the branch.
Linting is removed from the regular pipeline action so it runs once on linux and not on each platform matrix PR build. 
Saving about 1.5 minutes per build on each platform.


Klaus-Keller (approved):
Hi @devinea,
The workflow locks good to me und helps for online editing of sources and might also safe time for developers in case they forget to lint before commit. I'm not sure if this could cause conflicts with the change set workflow, as it also commits back to the `main` branch, but as this runs on each PR the likelihood of conflicts should be very low or even zero if we do not commit directly to main (which should never happen).

I'll approve this, but would appreciate if some workflow expert could review, too.

devinea:
> I'm not sure if this could cause conflicts with the change set workflow, as it also commits back to the `main` branch, but as this runs on each PR the likelihood of conflicts should be very low or even zero if we do not commit directly to main (which should never happen).

Sure. In case we run into issues we can remove these lines so it only commits in pull requests.
https://github.com/SAP/open-ux-tools/blob/lint_bot/.github/workflows/lint.yml#L6-L8

tobiasqueck (approved):
> devs who might forget to run lint fix before pushing the branch

Thank you!

devinea:
Thanks @Klaus-Keller @tobiasqueck

PR #772, opened 2022-11-02
fix(env-check): fix retrieval of process versions

cianmSAP:
#748 

Run all commands from os.homedir() in order to avoid directory issues in BAS. 
Also introduce better error handling.

### DIFF
--- a/packages/environment-check/src/checks/get-installed.ts
+++ b/packages/environment-check/src/checks/get-installed.ts
@@ -3,8 +3,8 @@ import { t } from '../i18n';
 import { join } from 'path';
 import { spawnCommand, npmCommand } from '../command';
 import { Extensions, NpmModules } from '../types';
+import type { ILogger } from '../types';
 import { isAppStudio } from '@sap-ux/btp-utils';
-import { getLogger } from '../logger';
 
 const pluginsDirBAS = join('/extbin/plugins');
 const globalNpmBAS = join('/extbin/npm/globals/lib');
@@ -69,9 +69,10 @@ async function getExtensionsVSCode(): Promise<{ [id: string]: { version: string 
 /**
  * Reads the list of extensions installed and returns the id and version.
  *
+ * @param logger - logger to report errors
  * @returns list of extension ids and versions
  */
-export async function getInstalledExtensions(): Promise<{ [id: string]: { version: string } }> {
+export async function getInstalledExtensions(logger?: ILogger): Promise<{ [id: string]: { version: string } }> {
     let versions;
     try {
         if (isAppStudio()) {
@@ -80,7 +81,7 @@ export async function getInstalledExtensions(): Promise<{ [id: string]: { versio
             versions = await getExtensionsVSCode();
         }
     } catch (e) {
-        console.error(t('error.retrievingExtensions', { error: e.message }));
+        logger?.error(t('error.retrievingExtensions', { error: e.message }));
     }
     return versions;
 }
@@ -157,15 +158,15 @@ export async function getFioriGenVersion(): Promise<string> {
 /**
  * Returns the versions of node.js modules.
  *
+ * @param logger - logger to report errors
  * @returns modules and versions
  */
-export async function getProcessVersions(): Promise<NodeJS.ProcessVersions> {
-    const logger = getLogger();
+export async function getProcessVersions(logger?: ILogger): Promise<NodeJS.ProcessVersions> {
     try {
         const output = await spawnCommand('node', ['-p', 'JSON.stringify(process.versions)']);
         return JSON.parse(output);
     } catch (e) {
-        logger.error(t('error.retrievingProcessVersions', { error: e.message }));
+        logger?.error(t('error.retrievingProcessVersions', { error: e.message }));
         return {} as NodeJS.ProcessVersions;
     }
 }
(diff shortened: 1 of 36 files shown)


packages/environment-check/test/output/__snapshots__/markdown.test.ts.snap
```diff
@@ -228,3 +228,39 @@ No messages<br>
...
+|UI5 Language Assistant Support|2|
+|XML Toolkit|2|
+|SAP Fiori tools - XML Annotation Language Server|2.2|
+|SAP Fiori tools - Application Modeler|2|
+|SAP Fiori tools - Guided Development|2|
+|SAP Fiori tools - Service Modeler|2.4|
+|SAP CDS Language Support|2|
+<details><summary>Versions</summary>
+<pre>
+undefined
```
devinea, line 257: is `undefined` expected here?

devinea (approved):
@cianmSAP Changes look good. One small query on the test output.

devinea (dismissed):
Thanks for the update @cianmSAP

packages/environment-check/src/checks/get-installed.ts
```diff
@@ -159,6 +159,11 @@ export async function getFioriGenVersion(): Promise<string> {
  * @returns modules and versions
  */
 export async function getProcessVersions(): Promise<NodeJS.ProcessVersions> {
-    const output = await spawnCommand('node', ['-p', 'JSON.stringify(process.versions)']);
-    return JSON.parse(output);
+    try {
+        const output = await spawnCommand('node', ['-p', 'JSON.stringify(process.versions)']);
+        return JSON.parse(output);
+    } catch (e) {
+        console.error(t('error.retrievingProcessVersions', { error: e.message }));
```
Klaus-Keller, line 166: @cianmSAP, can you please pass the logger into this function and log the error there, instead of the `console` so it is visible in the results? I think for an environment check it would be helpful to know why execution of command 
```
node -p 'JSON.stringify(process.versions)'
``` 
fails.

Klaus-Keller (changes requested):

packages/environment-check/src/checks/get-installed.ts
cianmSAP: Updated

packages/environment-check/src/checks/get-installed.ts
```diff
@@ -159,6 +160,12 @@ export async function getFioriGenVersion(): Promise<string> {
  * @returns modules and versions
  */
 export async function getProcessVersions(): Promise<NodeJS.ProcessVersions> {
```
Klaus-Keller, line 162:
```suggestion
export async function getProcessVersions(logger?: ReturnType<typeof getLogger>): Promise<NodeJS.ProcessVersions> {
```

Instead of `ReturnType<typeof getLogger>` you could also define a logger interface in `types.ts` that the class `Logger` implements, but you don't have to.

packages/environment-check/src/checks/get-installed.ts
```diff
@@ -159,6 +160,12 @@ export async function getFioriGenVersion(): Promise<string> {
...
-    const output = await spawnCommand('node', ['-p', 'JSON.stringify(process.versions)']);
-    return JSON.parse(output);
+    const logger = getLogger();
```
Klaus-Keller, line 163:
```suggestion
```
Sorry @cianmSAP, perhaps I phrased it not precisely. Whenever you do `getLogger()`, you get a new instance where you would need to do `logger.getMessages()` and return the messages. In this case, I would just pass the logger from the consumer here:
https://github.com/SAP/open-ux-tools/blob/87138cd93b7ea7f7a3f46cdb10547133720dcec1/packages/environment-check/src/checks/environment.ts#L25-L26

like 
```typescript
const logger = getLogger();
const processVersions = await getProcessVersions(logger);
```

the logger can be an optional parameter. In case of an error we use the passed logger instance (if set) to register the error. See other proposal in `catch`

packages/environment-check/src/checks/get-installed.ts
```diff
@@ -159,6 +160,12 @@ export async function getFioriGenVersion(): Promise<string> {
...
+    try {
+        const output = await spawnCommand('node', ['-p', 'JSON.stringify(process.versions)']);
+        return JSON.parse(output);
+    } catch (e) {
+        logger.error(t('error.retrievingProcessVersions', { error: e.message }));
```
Klaus-Keller, line 168:
```suggestion
        logger?.error(t('error.retrievingProcessVersions', { error: e.message }));
```

Klaus-Keller (changes requested):

packages/environment-check/src/checks/get-installed.ts
cianmSAP: Implemented the ILogger interface

packages/environment-check/src/checks/get-installed.ts
cianmSAP: Added

devinea (approved):

Klaus-Keller (approved):
Thanks @cianmSAP 
- thank for adding the interface for the logger
- test coverage is excellent
- installed, build and tested locally using CLI on VSCode, works

Approved.

cianmSAP:
Thanks @Klaus-Keller & @devinea

PR #773, opened 2022-11-03
TBI(ui5-application-writer) disable code assist with Typescript

devinea:
Disable codeAssist when Typescript is also selected. Since Typescript includes it's own code assist config.


longieirl (approved):

tobiasqueck (approved):
Thanks, the change makes a lot of sense. Existing tests are covering the case and the snapshots are updated correctly. So, all good.

packages/ui5-application-writer/src/data/index.ts
```diff
@@ -18,6 +18,10 @@ export function mergeWithDefaults(ui5App: Ui5App): {
     validate(ui5App);
     ui5App.app = mergeApp(ui5App.app);
     ui5App.appOptions = ui5App.appOptions || {};
+    // if typescript and codeAssist is enabled disable codeAssist
+    if (ui5App.appOptions.typescript && ui5App.appOptions.codeAssist) {
```
IainSAP, line 22: Should we document this behaviour via the types ts docs: 
https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-application-writer/src/types.ts#L57

IainSAP (approved):
Thx @devinea 

- Tests updated as expected
- No behavioural changes

packages/ui5-application-writer/src/data/index.ts
IainSAP: What if I set `codeAssist` to false and 'typescript' to true? I assume we still get code assist...

devinea: Yes that would be the same result as both being true. You get the Typescript code assist enabled.

devinea: Updated in https://github.com/SAP/open-ux-tools/pull/773/commits/daff85d11c822dec010b2b6ef3f006356d0472c4

IainSAP (approved):
Thx for the doc update, helps to clarify behaviour.

tobiasqueck (approved):
Thanks for updating the comments.

longieirl (approved):

devinea:
Thanks all

PR #779, opened 2022-11-07
Increased best (recommended) version of UI5 types modules to latest LTS version

tobiasqueck:


devinea (approved):
Thanks @tobiasqueck 
Changes and test updates look good. 
Manifest descriptor validation in tools-suite also needs to be updated. 
I'll create a PR for that.

devinea (approved):

nlunets (commented):
I don't 100% agree with the way ui5 describe their type but this is at least accurate usage :)

nlunets (approved):
Forgot to approve :)

PR #785, opened 2022-11-07
Enhance flex table control

vadson71:


Klaus-Keller (changes requested):
- Could you please check the code smell: https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=785&id=SAP_open-ux-tools&open=AYRSlgGn3tk3uEmNkSY9
- There is no change set defined

packages/ui-components/src/components/UIFlexibleTable/UIFlexibleTable.scss
```diff
@@ -318,9 +324,24 @@ $readonly-labels-opacity: 0.4;
...
         }
+
+        &-table.dragged {
+            .flexible-table-content-table-row.highlight-drop-warning {
+                .flexible-table-content-table-row-wrapper::before {
+                    position: absolute;
+                    width: calc(100% - 2px);
+                    height: calc(100% - 2px);
+                    content: " ";
+                    border: 1px solid $warning-highlight-color;
```
815are, line 335: do we need calc?
I think with following it should works same but without calc:
```
width: 100%;
height: 100%;
box-sizing: border-box;
border: 1px solid $warning-highlight-color;
```

packages/ui-components/src/components/UIFlexibleTable/UIFlexibleTable.scss
```diff
@@ -318,9 +324,24 @@ $readonly-labels-opacity: 0.4;
                 }
             }
         }
+
+        &-table.dragged {
+            .flexible-table-content-table-row.highlight-drop-warning {
+                .flexible-table-content-table-row-wrapper::before {
+                    position: absolute;
+                    width: calc(100% - 2px);
+                    height: calc(100% - 2px);
```
815are, line 333: I think we can avoid 'calc'
I think it would be cleaner with following and it would automatically adjust to border size increasement:
```
width: 100%;
height: 100%;
box-sizing: border-box;
border: 1px solid $warning-highlight-color;
```

vadson71: Update is done

815are (approved):
tested with storybook - did not find any regression

Klaus-Keller (approved):
- code changes look good
- test coverage is good
- I've installed and built locally and started Storybook, works.

Approved

Issue #780, opened 2022-11-07
Prototype pollution in webpack loader-utils


tobiasqueck:
https://github.com/advisories/GHSA-76p3-8jx3-jpfq

ui-components maintainers (@eouin @devinea @Klaus-Keller @815are @Jimmy-Joseph19), please address this one asap.

@donal-tobin-sap fyi

Klaus-Keller:
With draft PR https://github.com/SAP/open-ux-tools/pull/786 I update dependencies of `@sap-ux/ui-components` to latest versions. It installs, builds, and tests run fine. Remaining outdated modules are:

```
$ pnpm outdated                                                               
┌────────────────────────┬─────────┬─────────┐
│ Package                │ Current │ Latest  │
├────────────────────────┼─────────┼─────────┤
│ eslint (dev)           │ 8.23.0  │ 8.27.0  │
├────────────────────────┼─────────┼─────────┤
│ typescript (dev)       │ 4.3.5   │ 4.8.4   │
├────────────────────────┼─────────┼─────────┤
│ @types/react (dev)     │ 16.14.0 │ 18.0.25 │
├────────────────────────┼─────────┼─────────┤
│ @types/react-dom (dev) │ 16.8.5  │ 18.0.8  │
├────────────────────────┼─────────┼─────────┤
│ @types/uuid (dev)      │ 3.4.7   │ 8.3.4   │
├────────────────────────┼─────────┼─────────┤
│ babel-jest (dev)       │ 24.8.0  │ 29.3.0  │
├────────────────────────┼─────────┼─────────┤
│ react (dev)            │ 16.14.0 │ 18.2.0  │
├────────────────────────┼─────────┼─────────┤
│ react-dom (dev)        │ 16.14.0 │ 18.2.0  │
├────────────────────────┼─────────┼─────────┤
│ react-movable          │ 2.4.0   │ 3.0.4   │
├────────────────────────┼─────────┼─────────┤
│ uuid                   │ 3.4.0   │ 9.0.0   │
└────────────────────────┴─────────┴─────────┘
```

Did not touch these for different reasons:
- `eslint`, `typescript` are consistent with root level dependencies.
- `react`, `react-dom`, `react-movable`, `@types/react`, `@types/react-dom` could cause breaking changes and are not using `loader-utils`
- `uuid`, `@types/uuid` are not using `loader-utils`, but I think they could be updated as well

[...]

PR #786, opened 2022-11-07
fix(ui-components): update dependencies

Klaus-Keller:
Issue: https://github.com/SAP/open-ux-tools/issues/780


tobiasqueck (approved):
version updates, resolutions and resulting changes to the `pnpm-lock.yaml` look as expected.
I did not check if these changes have an impact on the `ui-components` functionality.

tobiasqueck (approved):

devinea (approved):
That is a lot of updates. Tools-suite would be best to get the same updates when consuming UI components.
Quick check on storybook locally looks OK.

Klaus-Keller:
> That is a lot of updates. Tools-suite would be best to get the same updates when consuming UI components. Quick check on storybook locally looks OK.

@devinea my bad. I've also asked @815are to double check consumption.

815are (approved):
As tested works ok for me,

only faced one build error in consumption:
image
but it is adjustable
otherwise my tests were ok

PR #787, opened 2022-11-08
[CHORE] Aligning ts-config and jest.config across modules

tobiasqueck:
The following has been consolidated across all modules:

`ts-config`

* removed unnecessary properties from the root config
* moved properties from specific config into root config that we anyhow always set
* simplified include/exclude patterns

`jest.config`

* created a base config with all common settings
* simplified project specific configs to only have the changed settings

No changeset required because neither code has been change nor are the build results affected.


tobiasqueck:
@Klaus-Keller @devinea @ullasholla opinions? Does it make sense to align the configs? 

I believe yes, because whenever I create a new module I copy paste the configs from an existing module and wonder why there are differences. I would like to consolidate the settings once, and only keep differences if really necessary.

Klaus-Keller:
> @Klaus-Keller @devinea @ullasholla @kranthie-sap (and anyone else) opinions? Does it make sense to align the configs?
> 
> I believe yes, because whenever I create a new module I copy paste the configs from an existing module and wonder why there are differences. I would like to consolidate the settings once, and only keep differences if really necessary.

@tobiasqueck, yes, I think it makes sense to align the configs. I assume this is why we inherit from root config. There might be module-specific special cases in which we overwrite root configurations, but the root config should serve as a foundation. Makes it also easier to change things centrally.

stefanschreck:
@tobiasqueck merge from main please

ullasholla (approved):
Thanks @tobiasqueck, looks much better now.

Checked:
* Base config files have correct values
* All the config files use the base files correctly
* Build scripts use the correct build config files

Klaus-Keller (approved):
Thanks @tobiasqueck for unifying tsconfig and jest config files. 
- Changes look good
- Thanks for cleaning `package.json` in `ui-components`
- I focused on `environment-check`, `project-access`, and `ui-components`, but also saw others
- Did a local install, build, test, and lint
- Tried storybook in `ui-components`

No issues found, approved.

Issue #788, opened 2022-11-08
BUG - EnvCheck - Destinations with OAuth2UserTokenExchange AuthenticationType are not listed


cianmSAP:
### Description

Destinations with OAuth2UserTokenExchange as the AuthenticationType are not being returned from `@sap/bas-sdk`

### Steps to Reproduce
Create a destination with OAuth2UserTokenExchange.
Go to Environment check and select `Check destination`

### Expected results
To see created destination

### Actual results
It is not listed

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [x] SAP Business Application Studio

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

cianmSAP:
Fix is merged

PR #791, opened 2022-11-09
fix(env-check): bump bas-sdk version to support oauth2usertokenexchange auth type

cianmSAP:
#788

Consume bas-sdk with OAuth2UserTokenExchange auth type support.


Klaus-Keller (approved):
Thanks @cianmSAP,
- consumption of new version looks good
- thanks for the compare cleanup
- did not build and run it locally
 
Approved.

Issue #711, opened 2022-10-03
FEATURE - Use ADT service to create transport request


slin-sap:
### Description

As the consumer of axios-extension AbapServiceProvider, SAP Fiori tools - Deployment config generator is planing to support a new feature, which allows its users to create a new transport request and use the new transport request in the generated configuration.

### Technical Design

axios-extension: Need to add createTransportRequest implementation to AbapServiceProvider. Expose this implementation from ServiceProvider to be consumed by UX tools suite Deployment config generator.

### Acceptance Criteria

**Given** consumer needs to create a new transport request
**when** the new createTransportRequest is called with input parameter description: string.
**then** ADT service for creating transport request should be executed and ABAP backend should return the newly created transport request ID.

### Tasks
- [x] PoC to test ADT service
- [x] Implement createTransportRequest and expose interface to axios-extension consumer
- [x] Unit Tests added

slin-sap:
This issue depends on previous PR: https://github.com/SAP/open-ux-tools/pull/627

tobiasqueck:
Finished with #793

PR #793, opened 2022-11-09
fix: update create transport number api

slin-sap:
#711 

- Changed TransportRequestService implementation and API for creating new transport request number.
- New TransportRequestService API now requires two extra input parameters `packageName` and `appName`.


packages/axios-extension/src/abap/adt-catalog/services/transportrequest-service.ts
```diff
@@ -26,24 +23,34 @@ export class TransportRequestService extends AdtService {
...
     /**
+     * TransportRequestService API function to create a new transport number.
      *
+     * @param packageName Package name for deployment
+     * @param appName Fiori project name for deployment. A new project that has not been deployed before is also allowed
      * @param description Description of the new transport request to be created
      * @returns Newly created transport request number
      */
-    public async createTransportRequest(description: string): Promise<string> {
+    public async createTransportRequest(packageName: string, appName: string, description: string): Promise<string> {
```
ullasholla, line 33: With this PR we no longer support any ABAP object but very specifically support only UI5 apps?

ullasholla (changes requested):
Looks like we're using the correct service now, thanks @slin-sap 

However, looks like this PR removes support for _any_ ABAP object and only specifically supports UI5 apps? Is that what we want?

theStiggySpd:
> However, looks like this PR removes support for _any_ ABAP object and only specifically supports UI5 apps? Is that what we want?

For the current consumption scope this is at least what we need. We can always expand it in future.
As of today, we know only Fiori tools is the consumer so at this early stage, IMO its fine to make this change for UI5 apps specifically. Of course please document it @slin-sap 
But if you have concern that in future we will again have this backward compatibility issue, then I agree, we should minimize that. If so, please see how we can make Shen's changes robust.

packages/axios-extension/src/abap/adt-catalog/services/transportrequest-service.ts
ullasholla: Thanks @theStiggySpd for the clarification here: https://github.com/SAP/open-ux-tools/pull/793#issuecomment-1309224867

Resolving

ullasholla, line 33: @slin-sap, please consider using a destructured object for the parameters here - the number is starting to get unwieldy.

Also, can the parameter `appName` be called `ui5AppName` or something for clarity?

slin-sap: Updated param name

ullasholla: Thank you for renaming that.

Can you do the destructured object for the parameters too? Positional arguments are a hidden and subtle form of tight coupling - the order of the parameters being the main cause. It's easy to transpose two parameters of the same type too, for example. The problem gets worse with increasing number of arguments. If you look at the other code in `axios-extension`, most of the methods take 0 or 1 arguments.

Typescript does not have named parameters like Python does and the using destructured parameters is a good alternative.

This is available as an automated refactoring in VS Code (it's also the only automated refactoring offered. Food for thought?). VS Code takes care of changing the code in the invocation locations as well.

tobiasqueck:
> But if you have concern that in future we will again have this backward compatibility issue, then I agree, we should minimize that. If so, please see how we can make Shen's changes robust.

We are still at a `0.x` version so it is ok to do an incompatible change. In this case, the change is required because it fixes the underlying API that was incorrectly used, and cannot be assumed to be available in future releases as-is.

packages/axios-extension/README.adoc
```diff
@@ -65,7 +65,11 @@ const provider = createForAbap({
     params: { 'sap-client': client }
 });
 const transportRequestSerivce = provider.getAdtService<TransportRequestService>(TransportRequestService);
-const newTrNumber = await transportRequestSerivce.createTransportRequest('A new transport request number for deployment');
+const newTrNumber = await transportRequestSerivce.createTransportRequest({
+    packageName: 'Z_PACKAGE',
```
tobiasqueck, line 69: What does the backend service actually do with the package and BSP name? Does it validate that it exists? If yes, what do I do when I want to create a transport for a new application that does not yet exist?

tobiasqueck (changes requested):
Code and tests look very good. Just one open question about the functionality and what the parameters do. Please update the readme with this information, and then it is good to go from my perspective.

devinea (dismissed):
Code and tests look good.

packages/axios-extension/README.adoc
slin-sap: README is to provide example of using ADT service. Instead, I updated the jsdoc for ui5AppName parameter in transportrequest-service.ts

ullasholla (approved):
Thanks @slin-sap

packages/axios-extension/README.adoc
tobiasqueck: I'd prefer getting a bit more information in the `README` as well but not just for this one, so for this PR, I am ok, and will approve, however, please consider a follow up given a little more guidance (like you explanation above) so that consumers have all information where it belongs.

tobiasqueck (approved):
Previously requested changes addressed, thank you.

PR #795, opened 2022-11-10
fix(ui-components): apply lost background to splitter

815are:
vscode color regression after - https://github.com/SAP/open-ux-tools/pull/759
after we removed `vscuie` variables - splitter lost its background because `vscuie` variable name was not matching vscode variable:
image

Adapted code to use previous variable `--vscode-debugToolBar-background`

### DIFF
--- a/.changeset/twelve-poets-shave.md
+++ b/.changeset/twelve-poets-shave.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui-components': patch
+---
+
+fix: transparent splitter issue, apply lost background color variable for splitter component
--- a/packages/ui-components/src/components/UISection/UISplitter.scss
+++ b/packages/ui-components/src/components/UISection/UISplitter.scss
@@ -16,7 +16,7 @@ $splitter-hover-contrast-border: 1px dashed var(--vscode-contrastActiveBorder);
 $splitter-focus-border: 1px solid var(--vscode-focusBorder);
 
 .splitter {
-    background-color: var(--vscode-debugToolBar-background);
+    background-color: var(--vscode-panelSection-border);
     position: absolute;
     outline: none;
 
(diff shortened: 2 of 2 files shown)


gpeuc:
@815are Hold on, splitter background and debugtoolbar background don't sound semantically the same. I believe there is a variable in vscode that should fit better here.

Where are we using this variable? What is the component? Is it the area splitter?

Screenshot 2022-11-10 at 09 13 20

Because that is var(--vscode-editorWidget-border) and on hover var(--vscode-focusBorder)

In any case, there should be a better match for this as I don't think splitter should go into debugtoolbar. It sounds too random. 😄

815are:
> Where are we using this variable? What is the component? Is it the area splitter?

Your screenshot is correct.
It looks following:
1. `--vscode-debugToolBar-background`:
image

2. `--vscode-editorWidget-border`
image

Should we still apply `--vscode-editorWidget-border`? Or we need adjust color for inner dots? currently it is '--vscode-icon-foreground'

815are:
@gpeuc 
Regarding debug menu, I am not sure if I was author of that usage or it was passed by someone else, but maybe idea was that debug toolbar have splitter dots as well:
image

gpeuc:
The dots might be ok, since they have an exact match, but the separator background is strange.

Native vscode is using this:

Screenshot 2022-11-10 at 09 43 18

for various separators

Screenshot 2022-11-10 at 09 43 33

can you please see if there is something that's a good match in our webview for this?

815are:
> can you please see if there is something that's a good match in our webview for this?

--vscode-panel-border: rgba(128, 128, 128, 0.35);
--vscode-panelSection-border: rgba(128, 128, 128, 0.35);
--vscode-settings-headerBorder: rgba(128, 128, 128, 0.35);
--vscode-settings-sashBorder: rgba(128, 128, 128, 0.35);
--vscode-terminal-border: rgba(128, 128, 128, 0.35);
--vscode-notebook-cellToolbarSeparator: rgba(128, 128, 128, 0.35);

If I use `--vscode-panel-border`:
Light:
image
Dark:
image

Should we use `--vscode-panel-border`?

gpeuc:
`--vscode-panelSection-border`

Because it is a border that splits sections. And it seems RGBA values match as well.

815are:
> `--vscode-panelSection-border`
> 
> Because it is a border that splits sections. And it seems RGBA values match as well.

ok, pushed change to use `--vscode-panelSection-border`

gpeuc (approved):
panelSection-border 

all good

Issue #687, opened 2022-09-20
FEATURE - Publish deploy-to-abap task from the @sap/ux-ui5-tooling module


tobiasqueck:
_As a **tools developer**,
I want to **reuse the deployment task that is today only available in the `@sap/ux-ui5-tooling`**,
so that I **can benefit from it in more different setups**_

### Description
The idea of the feature request is to publish the deployment functionality of the `deploy-to-abap` task (https://www.npmjs.com/package/@sap/ux-ui5-tooling#deployment-to-abap) so that it can be used in other setups as well e.g. standalone in a CI/CD setup. 
Furthermore, the task needs to handle lots of different setups e.g. handle different flavors of corporate proxies, therefore, publishing the sources and allowing developers that have setups in which the `@sap/ux-ui5-tooling` doesn't work, to inspect and maybe even contribute to it, will help the overall quality of the task.

### Technical Design
New module `@sap-ux/abap-deploy-task`
* configurable as UI5 custom task
* deploy functionality can also be called programmatically independent of the ui5 tooling
* simple CLI exposed so that it can also be called directly in a terminal

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

Given I have only the need to deploy to an ABAP system
when I add the new extracted ui5 task module to my project
then I can do that by adding it to the ui5.yaml

Given I have an issue with deploy-to-abap task
when I look at the open-ux-tools repository
then I can understand exactly how the deployment works.

Given I have specific needs for the deploy-to-abap task
when I use and enhance the published module
then I can build my own deployment module

### Tasks
- [x] create new module
- [x] Unit Tests added
- [ ] Test Cases defined

longieirl:
Issue found when further testing
https://github.com/SAP/open-ux-tools/issues/989

tobiasqueck:
The requested functionality has been added and integrated back into the Fiori tools

PR #797, opened 2022-11-10
Changes required for implementing the abap-deploy task

tobiasqueck:
For the implementation of #687, it is required to enhance the `axios-extension` module to be a little more flexible. This PR is carved out of PR #690 to be able to separately review/discuss/release these changes.

Changes:
* input parameter `archivePath` of type `string` for `Ui5AbapRepositoryService.deploy` has been changed to `archive` of type `Buffer` to allow providing an archive that does not exist on the file system
* optional parameter `alias` has been added to `AbapServiceProvider.getUi5AbapRepository` allowing to connect to the service if only exposed via alias
* generally handle `4xx` http statuses as error and not as success 
* added parameter `safeMode` to `Ui5AbapRepositoryService.deploy`


packages/axios-extension/src/abap/ui5-abap-repository-service.ts
```diff
@@ -73,27 +85,33 @@ export class Ui5AbapRepositoryService extends ODataService {
...
      *
-     * @param archivePath path to a zip archive containing the application files
+     * @param archive zip archive containing the application files as buffer
      * @param app application configuration
      * @param testMode if set to true, all requests will be sent, the service checks them, but no actual deployment will happen
+     * @param safeMode if set then the SafeMode url parameter will be set. SafeMode is by default active, to activate provide false
      * @returns the Axios response object for futher processing
      */
-    public async deploy(archivePath: string, app: ApplicationConfig, testMode = false): Promise<AxiosResponse> {
+    public async deploy(
```
tobiasqueck, line 94: @ullasholla since I am anyhow doing a breaking change here, should I also switch to a destructured object for the parameters

ullasholla: Yes, please

tobiasqueck: https://github.com/SAP/open-ux-tools/pull/797/commits/02a067037dc3e012ae8bcc3b3f817eb9b190c18b

packages/axios-extension/src/abap/ui5-abap-repository-service.ts
```diff
@@ -1,8 +1,62 @@
...
+    /**
+     * app application configuration
+     */
+    bsp: BspConfig;
+    /**
+     * if set to true, all requests will be sent, the service checks them, but no actual deployment will happen
+     */
+    testMode?: boolean;
+    /**
+     * if set then the SafeMode url parameter will be set. SafeMode is by default active, to activate provide false
```
devinea, line 47:
```suggestion
     * if set then the SafeMode url parameter will be set. SafeMode is by default active, to de-activate provide false
```
typo?

devinea (commented):
@tobiasqueck 
Changes look good and tests covered. Few small comments.
I am not so familiar with this code so not sure how much the API change will impact consumers. 
FYI @longieirl @slin-sap

packages/axios-extension/src/abap/ui5-abap-repository-service.ts
tobiasqueck: Thanks (https://github.com/SAP/open-ux-tools/pull/797/commits/a00092606af43db6e0ff74bd24c5af729212b193)

packages/axios-extension/src/abap/index.ts
```diff
@@ -1,4 +1,4 @@
-export { Ui5AbapRepositoryService, AppInfo, ApplicationConfig } from './ui5-abap-repository-service';
+export * from './ui5-abap-repository-service';
```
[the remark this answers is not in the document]
tobiasqueck: https://github.com/SAP/open-ux-tools/pull/797/commits/a00092606af43db6e0ff74bd24c5af729212b193

tobiasqueck:
> @tobiasqueck Changes look good and tests covered. Few small comments. I am not so familiar with this code so not sure how much the API change will impact consumers. FYI @longieirl @slin-sap

It isn't consumed yet. The planned consumption comes with https://github.com/SAP/open-ux-tools/tree/feat/687/deploy

nx.json
```diff
@@ -0,0 +1,53 @@
+{
```
ullasholla, line 1: Should this file be part of this PR?

packages/axios-extension/src/abap/abap-service-provider.ts
```diff
@@ -130,16 +130,15 @@ export class AbapServiceProvider extends ServiceProvider {
...
      * @returns an instance of the UI5 ABAP repository service.
      */
-    public getUi5AbapRepository(): Ui5AbapRepositoryService {
-        if (!this.services[Ui5AbapRepositoryService.PATH]) {
-            this.services[Ui5AbapRepositoryService.PATH] = this.createService<Ui5AbapRepositoryService>(
-                Ui5AbapRepositoryService.PATH,
-                Ui5AbapRepositoryService
-            );
+    public getUi5AbapRepository(alias?: string): Ui5AbapRepositoryService {
+        const path = alias ?? Ui5AbapRepositoryService.PATH;
```
ullasholla, line 137: Are we ok with `''` as an input for alias?

packages/axios-extension/src/base/odata-request-error.ts
```diff
@@ -1,3 +1,15 @@
+import type { AxiosError } from 'axios';
+
+/**
+ * Casts an unknown error to an AxiosError.
+ *
+ * @param e unknown error
+ * @returns exception casted to AxiosError if it is one
+ */
+export function isAxiosError(e: unknown): e is AxiosError {
```
ullasholla, line 9: The `isAxiosError()` function is exposed by `axios@1.x`. Guess this will can be replaced when we upgrade?

packages/axios-extension/src/base/odata-request-error.ts
```diff
@@ -1,3 +1,15 @@
...
+    return typeof e === 'object' && e !== null && 'isAxiosError' in e;
+}
+
```
ullasholla, line 12: It's not related to this PR specifically, so I'm not sure it makes sense to change it here (on the other hand, it's a small change).

It's a good idea to set the value of `this.name` in the error class's constructor: https://github.com/SAP/open-ux-tools/blob/da9f6c6d50ee2bf39ac01e3b84a6a9003a79b310/packages/axios-extension/src/base/odata-request-error.ts#L43-L47
Something like this:
```typescript
        this.name = this.constructor.name;
```
Doc: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name

packages/axios-extension/src/auth/connection.ts
```diff
@@ -144,25 +143,33 @@ export function attachConnectionHandler(provider: ServiceProvider) {
     });
 
     // throw error if the user is unauthorized otherwise, remove interceptor if successfully connected
-    const oneTimeRespInterceptorId = provider.interceptors.response.use((response: AxiosResponse) => {
-        if (response.status === 401) {
-            throw new ConnectionError(response.statusText, response);
-        }
```
ullasholla, line 150: AFAIK, client code (for example, Fiori tools app generator), relies on catching this error to show a login prompt. I know Fiori tools has not completely moved to using this library. Does it make sense to keep this behavior?

packages/axios-extension/src/base/odata-request-error.ts
```diff
@@ -1,3 +1,15 @@
+import type { AxiosError } from 'axios';
+
+/**
+ * Casts an unknown error to an AxiosError.
+ *
+ * @param e unknown error
+ * @returns exception casted to AxiosError if it is one
```
ullasholla, line 7:
```suggestion
 * @returns exception cast to AxiosError if it is one
```

packages/axios-extension/src/base/odata-request-error.ts
```diff
@@ -1,3 +1,15 @@
+import type { AxiosError } from 'axios';
+
+/**
+ * Casts an unknown error to an AxiosError.
```
ullasholla, line 4: Isn't it more accurate to say that the this type guard narrows the type down to an AxiosError _if_ it is actually one? Not very pithy...

ullasholla:
This branch seems to have changes from https://github.com/SAP/open-ux-tools/pull/799, is that intentional?

packages/axios-extension/src/factory.ts
```diff
@@ -41,6 +41,14 @@ function createInstance<T extends ServiceProvider>(
     delete providerConfig.ignoreCertErrors;
     providerConfig.withCredentials = providerConfig?.auth && Object.keys(providerConfig.auth).length > 0;
 
+    /**
+     * Make axios throw an error for 4xx errors as well.
+     *
+     * @param status - http response status
+     * @returns success (true) or error (false)
+     */
+    providerConfig.validateStatus = (status) => status < 400;
```
ullasholla, line 50: Where is this used? Also 3xx codes (redirection) need the client to take additional steps, so getting a 3xx code does not indicate success. Do we not expect 3xx codes?

ullasholla (changes requested):
Thanks @tobiasqueck, the changes look good.
Had a few minor comments. Could you please check?

Also, this PR seems to include changes from https://github.com/SAP/open-ux-tools/pull/799

packages/axios-extension/src/abap/abap-service-provider.ts
tobiasqueck: I have to admit, I hadn't thought about it but, yes, we are ok with it, we should/could even use it for BAS scenarios, where a destination points to a full service url and not just a system.

packages/axios-extension/src/base/odata-request-error.ts
tobiasqueck: Wow, I missed that there is a version `1.x` of `axios`. However, my implementation is better for `TypeScript` usage :).
Let's re-evaluate it when we switch to `axios@1`

packages/axios-extension/src/base/odata-request-error.ts
tobiasqueck: Done https://github.com/SAP/open-ux-tools/pull/797/commits/991e4ef5c8d7dd7433b43a432b8e6fc908699e53

packages/axios-extension/src/auth/connection.ts
tobiasqueck: It is still thrown. Before my change `4xx` was handled as success, and the code above was throwing it. Now, `4xx` is thrown by `axios` without us having to check and throw, so the behavior hasn't changed here for the consumer

packages/axios-extension/src/base/odata-request-error.ts
tobiasqueck: https://github.com/SAP/open-ux-tools/pull/797/commits/73a5f5dd77aa97088071fae6b445d6786bc65561 better?

packages/axios-extension/src/abap/ui5-abap-repository-service.ts
```diff
@@ -295,8 +374,12 @@ export class Ui5AbapRepositoryService extends ODataService {
      */
     protected logError({ error, host }: { error: AxiosError; host?: string }): void {
         this.log.error(error.message);
-        if (error.isAxiosError && error.response?.data?.['error']) {
-            prettyPrintError({ error: error.response.data['error'], host, log: this.log });
+        if (error.isAxiosError && error.response?.data) {
```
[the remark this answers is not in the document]
tobiasqueck: Done https://github.com/SAP/open-ux-tools/pull/797/commits/72b31771127f95e9f03147a12a9aeb485d7fa705

packages/axios-extension/src/factory.ts
tobiasqueck: We don't have any special handling for `3xx` in our code, and as far as I can see, all the redirects are handled by `axios` transparently except you disable it using `maxRedirects: 0`.
By default `axios` only considers `>=500` as errors which lead to many parts in the code where we check a successful response for a `4xx` response and if we find it, we throw an error. This line makes `axios` to also throw `4xx` as errors right away which fits the usage of our library.

ullasholla (approved):
Thanks @tobiasqueck, the changes look good. Cheers for the clarifications and the changes

devinea (approved):

Issue #796, opened 2022-11-10
CHORE - Align modules as much as possible


tobiasqueck:
### Description (include screenshots)
The repository has grown and we have seventeen modules. Interestingly, while the projects should have a pretty similar setup, they have unnecessary differences.

### Value
By aligning configurations and structures across all projects, it is easy for people to work on multiple modules without having to handle different behaviors and setups.

### Architecture Elaboration
Not in general, but I expect that some topics will be discussed in this ticket

### Tasks
- [ ] #787 
- [ ] Consider aligning the structure of the test folders e.g. where to put mocks, test fixtures, output for debugs etc.

Issue #495, opened 2022-05-17
TBI - Use Typescript Project References for build better performance


devinea:
### Description (include screenshots)
Use Typescript Project References for better performance in the build.
https://www.typescriptlang.org/docs/handbook/project-references.html

### Value
Optional - details on how this technical backlog item provides value to end-users or dev team eg. Performance, Quality, Maintainability. This will help determine the ranking in relation to feature development

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Generate TS project references in the tsconfig.json files base on pnpm workspace dependencies
- [ ] Add script/hook to auto update the TS project references
- [ ] Add `compilerOptions` to tsconfig.json
 ```
    "composite": true,
    "declaration": true,
    "declarationMap": true,
    "incremental": true,
    "outDir": "dist",
    "rootDir": "./src",
```
- [ ] Change build/watch scripts to use `--build` instead of `--project` for `tsc` cli.
- [ ] Update ignore files to ignore `*.tsbuildinfo`
- [ ] Adjust tsconfig include/exclude as needed based on `rootDir` setting
- [ ] As necessary add a tsconfig file for tests files and configure jest/eslint etc.. to use it.
- [ ] Bench mark the performance changes.

devinea:
This would likely not add much benefit now that NX has been implemented on the build.
Closing

PR #798, opened 2022-11-11
Next round of simplifications of the project structure

tobiasqueck:
Relevant for #495 and #796

* deleted unnecessary project specific `.eslintrc` files, only kept the ones that actually have module specific rules
    * makes it easier to see where we have project specific rules - which we shouldn't have
* deleted `tsconfig-build.json` and use `tsconfig.json` for builds again
* added `references` to `tsconfig.json` to allow builds to also trigger builds of dependencies i.e. execute `pnpm build --build`
* added `test/tsconfig.json` for modules with less strict settings for the unit tests than for the sources

Results after the changes
* `pnpm test`, `pnpm lint` and `pnpm build` have the same results as before
* less config files with less duplication across modules


packages/store/tsconfig.json
```diff
@@ -1,13 +1,14 @@
 {
     "extends": "../../tsconfig.json",
-    "include": ["src", "test"],
+    "include": ["src"],
     "compilerOptions": {
         "baseUrl": "./src",
```
ullasholla, line 5:
```suggestion

```
Shall we remove this and the `paths`? Looks like we decided against using them. We don't need the config anymore

packages/store/tsconfig.json
```diff
@@ -1,13 +1,14 @@
...
+        "rootDir": "src",
         "outDir": "dist",
-        "strict": false,
         "paths": {
             "types/*": ["types/*"],
             "utils/*": ["utils/*"]
         }
```
ullasholla, line 11:
```suggestion

```
Related to this: https://github.com/SAP/open-ux-tools/pull/798/files#r1022164523

packages/store/tsconfig.json
tobiasqueck: Done

packages/store/tsconfig.json
tobiasqueck: removed

devinea (approved):
Thanks @tobiasqueck 
Good clean-up and much more consistent.
Tested locally and works well.

packages/backend-proxy-middleware/package.json
```diff
@@ -14,13 +14,13 @@
     "author": "@SAP/ux-tools-team",
     "main": "dist/index.js",
     "scripts": {
-        "build": "pnpm clean && tsc -p tsconfig-build.json",
-        "clean": "rimraf dist",
+        "build": "tsc --build",
```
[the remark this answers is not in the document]
tobiasqueck: That was a request by @devinea because we are trying to optimize the builds and only build if something changed, but if we delete the `dist` folder before every build then, well, it always changed.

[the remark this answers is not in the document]
zdravko-georgiev: Thanks!

zdravko-georgiev (approved):
Reviewed `logger`, `backend-proxy-middleware` & `ui5-proxy-middleware`.

devinea (approved):
Thanks for the update

PR #799, opened 2022-11-11
chore(open-ux-tools) Enable NX to speed up build leveraging cache.

devinea:
Setup NX caching for the following commands to speed up the build massively leveraging NX caching and using NX Cloud for cache.
```JSON
"build",
"lint",
"test",
"build:extension",
"build:storybook",
"build:storybook:bundle"
```
https://nx.dev/recipes/adopting-nx/adding-to-monorepo

Switch cacheable operations to `nx` instead of `pnpm` in root package.json

| Command        | Time without NX           | Time with NX  |
| ------------- |:-------------:| -----:|
| build     | 23,000ms | 40ms |
| lint     |    37,000ms   |  37ms  |
| test |      40,000ms |   45ms |

PR pipeline - `~9.0` minutes -> `~3.6`  minutes


nx.json
```diff
@@ -0,0 +1,53 @@
...
+    "build": {
+      "dependsOn": [
+        "^build"
+      ],
+      "outputs": [
+        "{projectRoot}/dist"
+      ]
+    },
+    "test": {
+      "inputs": ["default", "^prod", "{workspaceRoot}/jest.config.ts"],
```
tobiasqueck, line 36: Shouldn't that be `{projectRoot}/jest.config.ts`

devinea: Updated in https://github.com/SAP/open-ux-tools/pull/799/commits/974edf6288eb5b96dae4b151e2670972f298f959

.github/workflows/lint.yml
```diff
@@ -33,7 +33,9 @@ jobs:
         env:
             cache-name: cache-pnpm-modules
         with:
-            path: ~/.pnpm-store
+            path:  |
+                    ~/.pnpm-store
+                    **/node_modules/.cache/nx/
```
StalkAltan, line 38: Hi there!

It's likely that with Nx Cloud enabled you won't actually want to do this -- there isn't any sort of age-based cache pruning in Nx so this folder will just swell in size and slow down your pipelines on cache retrieval.

nx.json
StalkAltan: You may want to consider leaving `{workspaceRoot}/jest.config.ts`. If configuration in this file changes, those changes will propagate to project-specific `jest.config.ts` files, but only at runtime. That means that changing the root configuration will not invalidate the cache of individual projects.

.github/workflows/lint.yml
devinea: Thanks. It was a temporary approach.  
Reverted in https://github.com/SAP/open-ux-tools/pull/799/commits/55cb1edd79e69741e810fddce4fb50dd8634ca34

tobiasqueck (dismissed):
Local build performance is very nicely improved now.

Klaus-Keller (approved):
Thanks @devinea, tested it locally and install and build works.

tobiasqueck (approved):

PR #806, opened 2022-11-16
feat: add/remove features to project, start with mockserver config

Klaus-Keller:
This PR adds modules to add, update, or remove features to an app or project, starting with mockserver configuration.

**New modules**
`@sap-ux/create-fiori`
Provides CLI to add remove features. It is also implemented to leverage command [npm init <package-spec>](https://docs.npmjs.com/cli/commands/npm-init).

`@sap-ux/mockserver-config-writer`
Contains the code to add/update/remove mockserver configuration to an app. This involves adding script `start-mock` to `package.json`, make sure correct dependency to `@sap-ux/ui5-middleware-fe-mockserver` is present, and adding or updating `ui5-mock.yaml`

**To test**
On this branch run 
```
pnpm build
pnpm install
```

In the root folder of the Fiori app/project you want to test the command run 
```
pnpm link <PATH_TO_OPEN_UX_TOOLS>/packages/create-fiori
npx create-fiori
```
This will show the help for CLI. You can simulate adding a mockserver configuration using:
```
npx create-fiori add mockserver-config -s
```
Without `-s` the changes are written to the project.

**Open**
- [ ] Collect early feedback as it should serve as a blueprint to add/update/remove for other features
- [ ] Unit tests
- [ ] Documentation


packages/mockserver-config-writer/src/prompt/index.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @param params.webappPath - optional path to webapp folder, where manifest is
+ * @param params.fs - optional memfs editor instance
+ * @returns - array of questions that serves as input for prompt module
+ */
+export function getMockserverConfigQuestions(params?: { webappPath?: string; fs?: Editor }): PromptObject[] {
+    let path = '';
+    if (params?.webappPath) {
+        const fs = params.fs || create(createStorage());
+        const manifest: Manifest = JSON.parse(fs.read(join(params.webappPath, 'manifest.json')));
+        path = getMainServiceDataSource(manifest)?.uri || '';
```
tobiasqueck, line 22: Why do you limit it to the main service? Wouldn't this also be useful for an `OVP` app that might have multiple services?

packages/mockserver-config-writer/src/mockserver-config/index.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @param basePath - the base path where the package.json and ui5.yaml is
+ * @param fs - the memfs editor instance
+ * @returns Promise<Editor> - memfs editor instance with updated files
+ */
+export function removeMockserverConfig(basePath: string, fs?: Editor): Editor {
+    if (!fs) {
+        fs = create(createStorage());
+    }
+    removeFromPackageJson(fs, basePath);
+    removeUi5MockYaml(fs, basePath);
```
tobiasqueck, line 37: Does it make sense to also allow removing a mockdata folder here if it existed and was configured in the mock config?

packages/mockserver-config-writer/src/mockserver-config/package-json.ts
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * Adds start-mock script to package.json, do nothing if script is up to date. If there is a
+ * custom start-mock script, copy it to start-mock_.
+ *
+ * @param fs - mem-fs reference to be used for file access
+ * @param packageJson - path to package.json
+ * @param webappPath - path to webapp folder
+ */
+function enhanceScripts(fs: Editor, packageJson: Package, webappPath: string): void {
```
tobiasqueck, line 114: I would prefer a different way of adding the script. With your current you don't take into account what I as a user actually want as my script. I might have added additional things to my normal start script to e.g. start a server in the background or do xyz, therefore, I'd prefer is you copy the existing `start` script and find the existing `ui5 serve` or `fiori run` and then replace or add the `--config` parameter.

packages/mockserver-config-writer/src/mockserver-config/ui5-mock-yaml.ts
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * Create a new ui5-mock.yaml based on existing ui5.yaml.
+ *
+ * @param fs - Editor instance to read existing information
+ * @param basePath -
+ * @param [path] - optional path for mockserver config
+ */
+async function generateUi5MockYamlBasedOnUi5Yaml(fs: Editor, basePath: string, path?: string): Promise<UI5Config> {
+    const ui5MockYamlConfig = await UI5Config.newInstance(fs.read(join(basePath, 'ui5.yaml')));
+    ui5MockYamlConfig.updateCustomMiddleware(await getMockserverMiddleware(path));
```
tobiasqueck, line 75: What does update do if it doesn't exist? Does it add it?

packages/mockserver-config-writer/src/types/index.ts
```diff
@@ -0,0 +1 @@
+export * from './types';
```
tobiasqueck, line 1: Do we really need that level of abstraction for a small component like this or could you just move all your types directly in to this file?

packages/create/src/validation/validation.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @param basePath - base path of the app, where package.json and ui5.yaml resides
+ */
+export function validateBasePath(basePath: string): void {
+    const packageJsonPath = join(basePath, 'package.json');
+    if (!existsSync(packageJsonPath)) {
+        throw Error(`Required file '${packageJsonPath}' does not exist.`);
+    }
+    const ui5YamlPath = join(basePath, 'ui5.yaml');
+    const webappPath = join(basePath, 'webapp');
+    if (!existsSync(ui5YamlPath) && !existsSync(webappPath)) {
```
tobiasqueck, line 18: The `ui5.yaml` is required based on this validation function, however, you have code for creating the `ui5-mock.yaml` that handles the case of no `ui5.yaml`.
So is it required or not?

packages/create/src/validation/validation.ts
```diff
@@ -0,0 +1,33 @@
...
+import type { Editor } from 'mem-fs-editor';
+import { join } from 'path';
+import type { EditorWithDump } from '../types';
+
+/**
+ * Validate base path of app, throw error if file is missing.
+ *
+ * @param basePath - base path of the app, where package.json and ui5.yaml resides
+ */
+export function validateBasePath(basePath: string): void {
```
tobiasqueck, line 11: Is it possible that this one is outdated e.g. hardcoded `webapp` path but the consumer of this function, searches after validation for the webapp path

packages/create/src/types/types.ts
```diff
@@ -0,0 +1,8 @@
+import type { Editor } from 'mem-fs-editor';
+
+/**
+ * mem-fs-editor types do not expose dump
+ */
+export type EditorWithDump = Editor & {
```
tobiasqueck, line 6: While I don't remember when, but I defined that type before somewhere. I think we need it for testing all over the place, maybe it makes sense to introduce a `global.d.ts` in our repo and define that one there.

packages/create/src/tracing/logger.ts
```diff
@@ -0,0 +1,64 @@
...
+    return logger;
+}
+
+/**
+ * Set custom formatting to the logger. Needs better solution, perhaps proxy transport in @sap-ux/logger.
+ *
+ * @param logger - instance of the logger
+ */
+function setCustomFormatter(logger: ToolsLogger): void {
+    const transports = (logger as any)?._logger?.transports;
```
tobiasqueck, line 32: That look like something you shouldn't be doing...
Wouldn't it be cleaner if you create a subclass of `ToolsLogger` here and then add/overwrite the things you need? This very same class could also offer something like an `getInstance` to achieve the singleton pattern that implemented.

packages/create/src/cli/index.ts
```diff
@@ -0,0 +1 @@
+export { handleCreateFioriCommand } from './create-fiori';
```
tobiasqueck, line 1: abstraction question again - would it make sense to move the `create-fiori` code directly in the `index.ts` since (if I understand the structure correctly) will always only export the content from it, and nothing else.

packages/mockserver-config-writer/src/mockserver-config/ui5-mock-yaml.ts
Klaus-Keller: Yes, it adds it. I've added this recently  `@sap-ux/ui5-config`

https://github.com/SAP/open-ux-tools/blob/46b8a2fe8db734da1b6f713d2bcca55310a59d77/packages/ui5-config/src/ui5config.ts#L352-L360

packages/mockserver-config-writer/package.json
```diff
@@ -0,0 +1,46 @@
...
+        "@sap-ux/ui5-config": "workspace:*",
+        "mem-fs": "2.1.0",
+        "mem-fs-editor": "9.4.0"
+    },
+    "devDependencies": {
+        "@sap-ux/project-access": "workspace:*",
+        "@types/mem-fs": "1.1.2",
+        "@types/mem-fs-editor": "7.0.1",
+        "@types/prompts": "2.0.14",
+        "prompts": "2.4.2"
```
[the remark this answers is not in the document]
Klaus-Keller: Yes, we discussed following approach:
- `*-writer` modules provide the questions in a certain format. Candidates have been `inquirer` and `prompts`
- `@sap-ux/create` fetches the questions and executes the user interaction

packages/create/src/validation/validation.ts
Klaus-Keller: No, there is an `&&`. File `ui5.yaml` is only required if there is no `webapp` path.

packages/create/src/types/types.ts
Klaus-Keller: We have a declaration in project access:
https://github.com/SAP/open-ux-tools/blob/46b8a2fe8db734da1b6f713d2bcca55310a59d77/packages/project-access/src/file/file-search.ts#L9-L21

How do you want to make this available cross modules?
https://www.typescriptlang.org/tsconfig#typeRoots
Or an include in the root `tsconfig.json`?

packages/create/src/tracing/logger.ts
Klaus-Keller: Yes, I'm well aware of this ;). I discussed with @ullasholla, he plans to come up with a concept to provide custom formatters to the logger. I will change to the new approach as soon as it is available.

packages/create/package.json
```diff
@@ -0,0 +1,50 @@
...
+    },
+    "files": [
+        "LICENSE",
+        "dist",
+        "!dist/*.map",
+        "!dist/**/*.map"
+    ],
+    "dependencies": {
+        "@sap-ux/logger": "workspace:*",
+        "@sap-ux/mockserver-config-writer": "workspace:*",
```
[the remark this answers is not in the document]
Klaus-Keller: Good point, this is also the reason why I like to move CLI commands to configure out of `@sap/ux-ui5-tooling`. To avoid slowing down the initial install after project generation.

For `@sap-ux/create` I see it in several stages:
- If you using it only occasionally, I guess it is fine if it takes bit longer.
- If you use it more often and want to improve performance, you can install `@sap-ux/create` either in a project or globally and use the CLI command.

In case it grows to an extend these two approaches are not sufficient, I think dynamic loading of modules makes a lot of sense.

packages/create/src/cli/index.ts
Klaus-Keller: I'm not a big fan of having too much code in `index.ts`, but in this case you are correct: intention is to export nothing else than exports of `create-fiori.ts` so I'll move the content.

packages/mockserver-config-writer/src/mockserver-config/index.ts
Klaus-Keller: Good question, but I'd say we should be be consistent with what we create. Currently we do not create mockdata folder so I wouldn't delete it. But we could enhance the writer to allow also choosing the mockdata folder and create it if it doesn't exits. In that case I would also delete it.

packages/create/src/validation/validation.ts
Klaus-Keller: function `validateBasePath()` just checks whether we have a `webapp` path or a `ui5.yaml` file, which could contain a custom mapping for webapp. It doesn't resolve the webapp path. I saw we use a similar approach in other writers.
I would not combine the two functionalities get webapp path and validate base path together, I have cases where I do not need the webapp path (remove mockserver config), but I still would like to do a basic validation to check whether the folder is a valid basePath aka app root.

packages/create/src/types/types.ts
Klaus-Keller: I played with different ways and could not find a good solution. I also question now the approach we have in project-access. It works, but does redefine the `Editor` declaration, so other functions like `commit()` are missing. I think a better declaration would be:

```typescript
import type { Editor as MemFsEditor } from 'mem-fs-editor';
/**
 * Add missing dump function declaration as mem-fs-editor types do not expose this.
 */
declare module 'mem-fs-editor' {
    type FileMap = { [key: string]: { contents: string; state: 'modified' | 'deleted' } };

    export interface Editor extends MemFsEditor {
        /**
         * Dump files to compare expected result. Provide a cwd for relative path.
         * See also https://github.com/SBoudrias/mem-fs-editor#dumpcwd-filter for further details.
         *
         * @param [cwd] - optional, relative path
         */
        dump(cwd?: string): FileMap;
    }
}
```

But if we would have this in a central place in this monorepo, we to have a `devDependency` to `mem-fs-editor` on root level. 

II don't see a good and clean solution, ideas?

packages/create/src/validation/validation.ts
tobiasqueck: Thanks for explaining and and or conditions to me :/ . I missed that one.

packages/create/package.json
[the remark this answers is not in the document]
tobiasqueck: Sounds like a solid plan.

packages/create/src/types/types.ts
tobiasqueck: How about doing it like https://github.com/SAP/open-ux-tools/pull/815? I branched from `main` and tried reusing it in `project-access` and in the tests of `fe-fpm-writer`.

IainSAP: Why not simply update to https://github.com/SBoudrias/mem-fs-editor/releases/tag/v9.0.0 it now exposes `dump`?

tobiasqueck: @IainSAP we are already using `v9.40` but the latest `@types/mem-fs-editor` module is `7.x` and is missing the `dump` function. So, we are not using a private function, just augment an incorrect type.
Or am I misunderstanding something?

IainSAP: Hmm shouldn't we update the public types in that case and use any cast in the meantime, rather than managing types for third-party ourselves.

tobiasqueck: Yes, check https://github.com/DefinitelyTyped/DefinitelyTyped#how-can-i-contribute. The first step is to do exactly what I did to make sure it works as expected. Once we have that proven, we contribute.

Klaus-Keller: > How about doing it like #815? I branched from `main` and tried reusing it in `project-access` and in the tests of `fe-fpm-writer`.

I like it, having a own private types module makes sense to me.

packages/create/package.json
```diff
@@ -0,0 +1,50 @@
+{
+    "name": "@sap-ux/create",
+    "description": "SAP Fiori tools module to add or remove features",
```
ullasholla, line 3: Why call the module `create` when it can add or remove features? `modify-feature` or something works better?

packages/create/src/cli/add/mockserver-config.ts
```diff
@@ -0,0 +1,96 @@
...
+import { relative } from 'path';
+
+/**
+ * Add the "add mockserver config" command to a passed command.
+ *
+ * @param cmd - commander command for adding mockserver config command
+ */
+export function addAddMockserverConfigCommand(cmd: Command): void {
+    cmd.command('mockserver-config [path]')
+        .option('-i, --interactive', 'ask for config options, otherwise use defaults')
```
ullasholla, line 18: Use i18n texts?

packages/create/src/cli/index.ts
```diff
@@ -0,0 +1,74 @@
+import { readFileSync } from 'fs';
+import { join } from 'path';
+import { Command } from 'commander';
+import { getLogger } from '../tracing';
+import { getAddCommands } from './add';
+import { getRemoveCommands } from './remove';
+
+/*
+ * If chosen 'commander' over 'minimist' and 'yargs' for this CLI implementation. Reasons:
```
ullasholla, line 9: `If` -> `I've` -> `We've`?

packages/create/src/cli/index.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Run the command line interface.
+ *
+ * @param argv - arguments, typically 'process.argv'
+ */
+export function handleCreateFioriCommand(argv: string[]): void {
+    const logger = getLogger();
+    if (!Array.isArray(argv) || argv.length < 2) {
+        throw Error(`This function must be called from command line interface (cli). Or provide meaningful argv.`);
```
ullasholla, line 24: i18n text? `argv` -> `arguments`?

packages/create/src/cli/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+    try {
+        const program = getCommanderProgram();
+        program.parse(argv);
+    } catch (error) {
+        logger.error(`Error while executing ${argv[0]} ${argv[1]}`);
+        if (error instanceof Error) {
```
ullasholla, line 31: Why narrow the instance type to `Error`? It's a pity that any object can be `throw`n in JS/TS but the serialized value may still be useful

packages/create/src/cli/index.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Returns an instance of the command line program (commander command) ready to be executed (parsed), like 'program.parse(argv)'.
+ *
+ * @returns - commander program
+ */
+function getCommanderProgram(): Command {
+    const program = new Command();
+    const version = getVersion();
+    program.description(`Configure features for Fiori applications and projects. (${version})`);
```
ullasholla, line 45: The verbosity can be annoying if the command is used in a pipeline (It won't play nice with something like this: `npx create-fiori [args] | grep <something> | cut <something else>`). Support a `silent` option?

packages/mockserver-config-writer/src/mockserver-config/package-json.ts
```diff
@@ -0,0 +1,157 @@
...
+ */
+function enhanceDependencies(
+    packageJson: Package,
+    mockserverModule = '@sap-ux/ui5-middleware-fe-mockserver',
+    version = '2'
+) {
+    packageJson.devDependencies = packageJson.devDependencies || {};
+    delete packageJson.devDependencies['@sap/ux-ui5-fe-mockserver-middleware'];
+    packageJson.devDependencies[mockserverModule] = version;
+    packageJson.ui5 = packageJson.ui5 || {};
```
ullasholla, line 43:
```suggestion
    packageJson.ui5 ||= {};
```
?

packages/create/src/types/types.ts
tobiasqueck: Perfect, then I will get the PR ready for review asap.

packages/mockserver-config-writer/src/types/index.ts
Klaus-Keller: With the change to the central type for `Editor` this file is gone now.

packages/create/src/types/types.ts
Klaus-Keller: Changed to include `@sap-ux/types`.

packages/create/package.json
ullasholla: Thanks for reminding me about this. It is handy to be able to do that with `npm init`.

My concern was with running with `npx` or `npm exec`. Then the command will look something like:
`npx @sap-ux/create remove <feature>`

packages/mockserver-config-writer/src/prompt/index.ts
Klaus-Keller: This is the initial default value when prompting for the path to mock, but you are right, when we have an OVP or multiple OData services as data sources, we should allow to select from those. I've changed this now.

packages/create/src/cli/index.ts
Klaus-Keller: Fixed

packages/create/src/cli/index.ts
Klaus-Keller: Changed to `arguments`.

packages/create/src/cli/add/mockserver-config.ts
Klaus-Keller: I18N is on my list to do.

packages/create/src/cli/index.ts
Klaus-Keller: changed it to check for type `error` or `string`, as this is what the logger can digest.

packages/mockserver-config-writer/src/mockserver-config/package-json.ts
Klaus-Keller: fixed

packages/create/package.json
Klaus-Keller: We still have a `bin` entry in `package.json`:
```json
"bin": {
    "create-fiori": "./dist/index.js"
},
```
https://github.com/SAP/open-ux-tools/pull/806/files#diff-84505d75821385613d447e5d9c00802c2b657d2fbed9b9ccc2303a1e751dcc3bR14-R15

If you have `@sap-ux/create` globally installed on your machine you can do something like
```
npx create-fiori add mockserver-config 
```
or, to follow your concern
```
npx create-fiori remove mockserver-config
```

I just named it `create-fiori`, but I think we could name it differently. Any proposals? I guess best would be `@sap-ux` to be consistent, but I'm not sure if a terminal command starting with `@` is the best idea. Opinions?

Klaus-Keller: Just did a test, `@sap-ux` can't be used for `bin`, but `sap-ux` would work.

packages/mockserver-config-writer/src/mockserver-config/package-json.ts
Klaus-Keller: Changed the logic:
- copy start -> start-mock script and add/replace `--config`
- fallback to basic script in case there is no start script, which should be an exceptional case

packages/create/src/cli/index.ts
ullasholla: I think the type definitions of the `error` method in `ToolsLogger` is not accurate. We're basically re-exporting Winston's typedefs. 
The only requirement is that the argument to `error()` must be a string or any type that has a `toString()` method (which all types do).
This check is not required:
https://github.com/SAP/open-ux-tools/blob/b54f1b44870338306df7e89a8d4db33021419fb7/packages/create/src/cli/index.ts#L31-L33
Did you see a problem with not having that check before calling `logger.error()`?

Klaus-Keller: > Did you see a problem with not having that check before calling `logger.error()`?

Yes:

image

There are several ways to overcome this, none seems completely clean:

```typescript
} catch (error: any) {
        logger.error(`Error while executing ${argv[0]} ${argv[1]}`);
        logger.error(error);
```

```typescript
    } catch (error) {
        logger.error(`Error while executing ${argv[0]} ${argv[1]}`);
        logger.error(error as object | string);
```

For now I go with `logger.error(error as object | string);`. What do you think?

packages/create/src/tracing/logger.ts
ullasholla: Will be tracked in issue - https://github.com/SAP/open-ux-tools/issues/831

packages/create/src/cli/add/mockserver-config.ts
Klaus-Keller: i18n is used now in `mockservice-config-writer`. For cli we do not add i18n.

ullasholla (approved):
Thanks @Klaus-Keller, the PR looks good. Code quality and test coverage are good. Thanks for addressing my review comments as well

packages/mockserver-config-writer/README.md
```diff
@@ -0,0 +1,33 @@
+# @sap-ux/mockserver-config-writer
+
+Adds or removes configuration for mockserver module `@sap-ux/ui5-middleware-fe-mockserver` to SAP UX project.
```
kranthie-sap, line 3: `to SAP UX project` -> `to an SAP UX project`

packages/mockserver-config-writer/README.md
```diff
@@ -0,0 +1,33 @@
...
+    basePath,
+    {
+        webappPath: join(basePath, 'webapp'),
+});
+
+fs.commit();
+```
+
+See more complex example in [`/test/unit`](./test/unit)
+## Keywords
```
kranthie-sap, line 31: New empty line before `##Keywords`

Klaus-Keller: @kranthie-sap, I'll create a follow up for this.

packages/mockserver-config-writer/README.md
Klaus-Keller: @kranthie-sap, I'll create a follow up for this.

kranthie-sap (approved):
Code looks clean and well organized.

devinea (commented):
@Klaus-Keller 
Testing this to add mock server to an old project the wrong data folder is configured. 
The `mockdataPath` is hardcoded below but should be configured to match the current project. for example `mockdata` was a common folder name used on other projects using sapui5  mockserver.
https://github.com/SAP/open-ux-tools/blob/b6d0c67fb3d24e6cb23a94ada6ad052d5235e6d5/packages/ui5-config/src/middlewares.ts#L91

For apps using sapui5  mockserver have you considered removing `mockserver.js` , `flpSandboxMockServer.html` for example?

Also, I don't really like the name `@sap-ux/create` and bin `sap-ux`. 
In tools-suite will this be included in the `fiori` binary or still separate?

devinea (commented):
@Klaus-Keller Can you update https://github.com/SAP/open-ux-tools/blob/main/.github/CODEOWNERS for these new modules ?

Klaus-Keller:
@kranthie-sap  
> https://github.com/SAP/open-ux-tools/pull/806#discussion_r1047478855
to SAP UX project -> to an SAP UX project

-> https://github.com/SAP/open-ux-tools/pull/832

> @kranthie-sap https://github.com/SAP/open-ux-tools/pull/806#discussion_r1047480466
New empty line before ##Keywords

-> https://github.com/SAP/open-ux-tools/pull/832

@devinea 
> Testing this to add mock server to an old project the wrong data folder is configured.
The mockdataPath is hardcoded below but should be configured to match the current project. for example mockdata was a common folder name used on other projects using sapui5 mockserver.
[open-ux-tools/packages/ui5-config/src/middlewares.ts](https://github.com/SAP/open-ux-tools/blob/b6d0c67fb3d24e6cb23a94ada6ad052d5235e6d5/packages/ui5-config/src/middlewares.ts#L91)

> Line 91 in [b6d0c67](https://github.com/SAP/open-ux-tools/commit/b6d0c67fb3d24e6cb23a94ada6ad052d5235e6d5)

> mockdataPath: './webapp/localService/data', 

>For apps using sapui5 mockserver have you considered removing mockserver.js , flpSandboxMockServer.html for example?

Good points, perhaps we can make the path to mockdata configurable and also return it as question from here:
https://github.com/SAP/open-ux-tools/blob/0c9a3b0cf9152bc64ebfc53c1dda2a48153cf451/packages/mockserver-config-writer/src/prompt/index.ts#L18
Also when adding without questions we can try to see if an a mockdata path is present and use it, otherwise fall back to the default. Can you create a ticket for this?

>Also, I don't really like the name @sap-ux/create and bin sap-ux.
In tools-suite will this be included in the fiori binary or still separate?

The name `@sap-ux/create` was chosen due to the naming convention given by [`npm init`](https://docs.npmjs.com/cli/commands/npm-init), which allows us now to do something like
```
npm init @sap-ux help
```
without installing the module. I would have loved to use exactly the same as binary command, but the terminal doesn't like commands starting with `@`, so `sap-ux` was chosen.

`@sap/ux-ui5-tooling` offers following commands to add configuration to a project:

```
fiori add deploy-config 
fiori add flp-config 
fiori add flp-embedded-config
fiori add variants-config
```

There are plans to add these commands here using a *-writer module and then call them from `@sap/ux-ui5-tooling` to be backward compatible.

PR #807, opened 2022-11-16
chore: fix sonar config to exclude tests

ullasholla:
This PR fixes the exclusions in `sonar-project.properties` so that test files and mocks are correctly excluded from coverage calculation and reporting.

Currently, for example:
image

https://sonarcloud.io/component_measures?id=SAP_open-ux-tools&metric=new_coverage&view=list

It also removes project specific configuration in favor of general configuration.

No changeset is required as this only affects coverage reporting.


Klaus-Keller (approved):
Updated sonar exclusions look good to me. Thanks @ullasholla.

devinea (approved):
Thanks @ullasholla

PR #808, opened 2022-11-17
chore: add sonarjs for linting

Klaus-Keller:
Add `eslint-plugin-sonarjs` and enable rules:
```
"sonarjs/cognitive-complexity": "warn",
"sonarjs/no-nested-template-literals": "warn",
```

To match all rules exactly to what we see in https://sonarcloud.io/project/issues?resolved=false&severities=CRITICAL&id=SAP_open-ux-tools takes longer, as some rules are not covered by `eslint-plugin-sonarjs`. Sometimes they can be found in other eslint plugins, like `@typescript-eslint/eslint-plugin`.

While doing this I realized, there was a formatting issue with Icons, caused by: https://github.com/SAP/open-ux-tools/pull/782

I fixed, this, but it should have happened automatically @devinea.


tobiasqueck (approved):
Thank you. Let's keep cognitive complexity as warning for a few weeks and then switch it to error

zdravko-georgiev (approved):

PR #809, opened 2022-11-17
feat(fe-fpm-writer): add building block prompts and sample cli generator

kranthie-sap:
## Description
* Add prompts to capture the information required to generate chart and filter bar building block guides
* Add sample cli generator that uses the prompts and writer functions to generate chart and filter bar building blocks


packages/project-access/src/constants.ts
```diff
@@ -1,7 +1,9 @@
 export const FileName = {
+    Fragment: 'fragment.xml',
```
tobiasqueck, line 2: Are these really file names and not suffices that maybe should go to a different object?

kranthie-sap:
Closing this PR as it is being worked on here to support a new prompting UI module - https://github.com/SAP/open-ux-tools/pull/1564

Issue #769, opened 2022-11-02
BUG - China BAS update Generation and migration


cianmSAP:
### Description
Updates required to templates with new trusted host for BAS China.

### Steps to Reproduce
1. Generate or migration an app to support Fiori tools.
2. Inspect `changes_loader.js` and `changes_preview.js`,  trusted hosts need to be updated with BAS China.

### Expected results

The Fiori app should be fully usable in BAS China.

### Actual results

The BAS China host is not trusted.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

cianmSAP:
Fix implemented

PR #810, opened 2022-11-17
fix(fiori-elements-writer): removes unnecessary trusted host validation

zdravko-georgiev:
#769 
Removes the host validation from the `changes_preview/chages_loader` files. The `$.ajax` calls are made now without the host, which makes the host validation obsolete.


.changeset/violet-geckos-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/fiori-elements-writer': minor
+---
+
+removes trusted host validation
```
tobiasqueck, line 5: Could you change this changelog comment to "removes unnecessary trusted host validation". In addition, I think it is more of a patch then a minor change because you are not adding or changing functionality, you are removing a faulty limitation

tobiasqueck (changes requested):

.changeset/violet-geckos-lick.md
zdravko-georgiev: Yes, done with https://github.com/SAP/open-ux-tools/pull/810/commits/e3c04dd73ac86b21e2316a925c5db9e8eaaa0b56

zdravko-georgiev:
> Thank you for the cleanup. I just reviewed, I haven't tested it locally but know that you did as part of the testing for the Fiori tools.

Thank you.

devinea (approved):
Thanks @zdravko-georgiev for making these changes.
Looks good.

PR #812, opened 2022-11-18
feat(ui-components) - ui-treedropdown add error message handling

eouin:
- [x] Add an error message to the UITreeDropdown if any

PR #813, opened 2022-11-18
fix(ui-components) - align lint:fix script with others in monorepo

devinea:
fix(ui-components) - align lint:fix script with others in monorepo

Notice issues on:
https://github.com/SAP/open-ux-tools/pull/782
https://github.com/SAP/open-ux-tools/pull/812


tobiasqueck:
>  some lint:fix scripts behave differently, resulting in missing lint errors

Do you have an example, and shouldn't we fix those?

tobiasqueck:
> I think it is safer to use lint script with --fix parameter

I disagree. When you work, you use the scripts within the module, and I expect that when we run a script for the whole repo that it does for each module what the script in the module does.

tom-sap (dismissed):

package.json
```diff
@@ -36,7 +36,7 @@
         "build": "nx run-many --target=build --all",
         "format": "pnpm recursive run format",
         "lint": "nx run-many --target=lint --all",
-        "lint:fix": "nx run-many --target=lint:fix --all",
+        "lint:fix": "nx run-many --target=lint --fix --all",
```
tobiasqueck, line 39: Please consider removing this change again, and make sure that all modules work as expected, to keep running something within a module consistent with running it from root.

devinea: Done in https://github.com/SAP/open-ux-tools/pull/813/commits/394d782198507cc923adec228708c1cc05244333

tobiasqueck (dismissed):

tobiasqueck (approved):
Is the formatting change of the `Icons.tsx` on purpose? If yes, then all is good.

unseen1980 (approved):

devinea:
Thanks all

PR #814, opened 2022-11-21
fix(ui-components): sonar issues

815are:
Multiple fixes for sonar issues - https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&id=SAP_open-ux-tools
Issues for following rules was considered/fixed(but not all issues was fixed):
1. Prefer using nullish coalescing operator (`??`) instead of a logical or (`||`), as it is a safer operator.
2. Refactor this code to not use nested template literals.
3. Remove this property or method or refactor "Class", as "method" is not used inside component body
4. Move this component definition out of the parent component and pass data as props.
5. The signature '(from: number, length?: number | undefined): string' of 'text.substr' is deprecated.
6. Mark the props of the component as read-only.
7. Make this public static property readonly.
8. Prefer using an optional chain expression instead, as it's more concise and easier to read.
9. '{property}' is deprecated.
10. Unnecessary use of boolean literals in conditional expression.
11. useState call is not destructured into value + setter pair
12. Remove this commented out code.

Please note:
1. I skipped mulltiple `Prefer using nullish coalescing operator (`??`) instead of a logical or (`||`), as it is a safer operator` and did not apply fix, like:
```
title={tooltip || column.tooltip}
```
because we need also consider empty string.
2. I skipped fixes for `UITable.tsx` - I would suggest to handle it as separate PR, because there is multiple issues about complexity;
3. Skipped some issues which I was not sure or it was risky to fix;


tobiasqueck:
@815are what happened to this one? Obsolete or still being worked on?

815are:
> @815are what happened to this one? Obsolete or still being worked on?

@tobiasqueck 
I will recheck - I applied some sonar fixes, but I remember there was also other sonar issues left, that's why it was not published

Klaus-Keller (approved):
Thanks @815are,
- Fixes for Sonar code smells look good
- Did a brief test with Storybook

Approved from my side.

vadson71 (approved):

vadson71 (approved):

Klaus-Keller (approved):
re-approving

PR #815, opened 2022-11-21
Type augmentation for mem-fs-editor

tobiasqueck:
Proposal for discussion at https://github.com/SAP/open-ux-tools/pull/806#discussion_r1026951441.

@Klaus-Keller how about doing it like this?
@nlunets Thanks for the help!

Next steps

- [x] get opinion if this is a good option to augment types for all projects
- [x] finalize type enhancement
- [x] replace all unnecessary casts to `any`

No changeset required since it is only an internal type change.

### DIFF
--- a/.changeset/kind-pianos-act.md
+++ b/.changeset/kind-pianos-act.md
@@ -1,5 +1,0 @@
----
-"@sap-ux/ui-components": patch
----
-
-feat(ui-components) - ui-treedropdown add error message handling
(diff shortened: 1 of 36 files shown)


Klaus-Keller:
@tobiasqueck, I like this proposal. Should we try to merge it separate before https://github.com/SAP/open-ux-tools/pull/806?

tobiasqueck:
> @tobiasqueck, I like this proposal. Should we try to merge it separate before #806?

Yes, that would be my preference.

stefanschreck (approved):

devinea (approved):
Thanks @tobiasqueck 
Looks like a good approach.

Klaus-Keller (approved):
Adding missing type declaration in a private module makes sense to me. 
Including local type via `tsconfig.json` looks good.
I've focused review on project-access, but installed, build, and ran tests on the branch locally and it works.

Issue #811, opened 2022-11-18
[SECURITY] - Vulnerability in `loader-utils`


tobiasqueck:
### Description
https://github.com/advisories/GHSA-3rfm-jhwj-7488

### Expected results
1. `pnpm audit`
2. no issues

### Actual results
`pnpm audit`
```
┌─────────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────┐
│ high                │ loader-utils is vulnerable to Regular Expression Denial of Service (ReDoS) via url variable │
├─────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┤
│ Package             │ loader-utils                                                                                │
├─────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┤
│ Vulnerable versions │ >=2.0.0 <2.0.4                                                                              │
├─────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┤
│ Patched versions    │ >=2.0.4                                                                                     │
├─────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┤
│ More info           │ https://github.com/advisories/GHSA-3rfm-jhwj-7488                                           │
└─────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────┘
┌─────────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────┐
│ high                │ loader-utils is vulnerable to Regular Expression Denial of Service (ReDoS) via url variable │
├─────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┤
│ Package             │ loader-utils                                                                                │
├─────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┤
│ Vulnerable versions │ >=1.0.0 <1.4.2                                                                              │
├─────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┤
│ Patched versions    │ >=1.4.2                                                                                     │
├─────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┤
│ More info           │ https://github.com/advisories/GHSA-3rfm-jhwj-7488                                           │
└─────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────┘
┌─────────────────────┬─────────────────────────────────────────────────────────────────────────────┐
│ high                │ loader-utils is vulnerable to Regular Expression Denial of Service (ReDoS)  │
├─────────────────────┼─────────────────────────────────────────────────────────────────────────────┤
[...]

PR #816, opened 2022-11-22
fix(security): upgrade loader-utils

donal-tobin-sap:
- use `pnpm upgrade -r loader-utils` to upgrade lock file after updating the overrides
- updated overrides line comes from audit fix

Issue: https://github.com/SAP/open-ux-tools/issues/811


donal-tobin-sap:
needs to be recreated

PR #817, opened 2022-11-22
fix(security): upgrade loader-utils

donal-tobin-sap:
- use `pnpm upgrade -r loader-utils` to upgrade lock file after updating the overrides

Issue: https://github.com/SAP/open-ux-tools/issues/811


devinea (approved):
Thanks @donal-tobin-sap

zdravko-georgiev (approved):

PR #818, opened 2022-12-01
fix(ui-components): UIComboBox. Existing property "openMenuOnClick". Default value should be "true" and always open dropdown on first key press.

815are:
Regarding property 'openMenuOnClick':
1. If  `openMenuOnClick` is 'false', then if we click on `input` element, then dropdown menu will not be opened;
2. If `openMenuOnClick` is 'true', then if we click on `input` element, then dropdown menu will be opened;

PR contains three corrections regarding `UIComboBox` component and property `openMenuOnClick`:
1. Default value should be 'true';
2. There was condition if `openMenuOnClick=true` and we press any key on keyboard while input focused - we do not open dropdown menu. But it was faulty condition and we should open dropdown menu on first key press. 
3. Fix for sonar issue - property `which` is considered as deprecated. Replaced property `which` to property `key`. In our code it is used to detect which key press should trigger open of dropdown menu and which key should be ignored. Commit c4cb5ad8a32a90827f8af569a35588419f63e6d6


Klaus-Keller (approved):
- Code changes look good to me
- Didn't test locally
- Test coverage is great

Approved from my side.

Jimmy-Joseph19 (approved):
Code change looks good. Tested with storybook and works as described.

PR #820, opened 2022-12-03
fix(security): upgrade decode-uri-component

donal-tobin-sap:
Issue: tools-suite #22655 and #22663


zdravko-georgiev (approved):

tobiasqueck (approved):
Approved. I have checked whether it is required or if we could update the top level dependencies leading to this nested dependency but it, there are no valid versions available yet.

Issue #821, opened 2022-12-05
FEATURE - Add UI Component to support context sensitive help


IainSAP:
_As a UI Developer,
I want to show context sensitive help in-situ,
so that I provide contextually relevant help links and actions to users _

### Description

Adds a new UI component that can be used to display help links or trigger callbacks wit specific command configuration for integration to ide environments such as VSCode.

### Technical Design

- Adds new module `UIGuidedAnswersBox`
- Can be configured with url links
- Can be configured with command actions (passed as args to the callback on click) in the format:
- Can be configured with link text
- Can be configured with sub-text
- Can be configured with a  `UIIcon` other than the default provided
- Should re-use existing fluent-ui and/or UI Components to ensure consistent behaviour, look & feel.

Example: 
```
export interface IActionCalloutDetail {
    linkText: string;
    subText: string;
    /**
     * Command to be executed and parameters passed to the command
     */
    command?: {
        id: string;
        params: Object | string;
    };
    /**
     * A http url string, command takes precedence if provided
     */
    url?: string;
}
```

UI Design:

<img width="1476" alt="image" src="https://user-images.githubusercontent.com/46536134/205698277-cb0e5a43-1b09-44a6-a32b-2a589f3f96c8.png">

### Acceptance Criteria

Sufficent tests are provided to prove feature requirements.

### Tasks
- [x] Add new component:  `UIGuidedAnswersBox`
- [x] Add to `storybook`
- [x] Add unit tests

IainSAP:
Done

PR #822, opened 2022-12-05
feat(ui-components): Adds UIActionCallout

IainSAP:
Adds new component defined by: https://github.com/SAP/open-ux-tools/issues/821

Example:

image


packages/ui-components/test/unit/components/UIGuidedAnswerBox.test.tsx
```diff
@@ -0,0 +1,118 @@
+import { Callout } from '@fluentui/react';
+import type { ICalloutContentStyles, IStyle } from '@fluentui/react';
+import * as Enzyme from 'enzyme';
```
devinea, line 3: Please use [react testing library](https://testing-library.com/docs/react-testing-library/intro/) instead of enzyme as it appears to no longer be maintained and does not support newer react versions.

devinea (changes requested):
Control looks good and story is provided. 
New tests should use react-testing-library instead of enzyme.

packages/ui-components/test/unit/components/UIGuidedAnswerBox.test.tsx
IainSAP: I used what is already in use in the other tests, was that wrong?

devinea (approved):
Thanks @IainSAP for the updates.

packages/ui-components/package.json
```diff
@@ -57,6 +57,8 @@
         "@types/react-virtualized": "9.21.21",
         "@types/sanitize-html": "2.6.2",
         "@types/uuid": "3.4.7",
+        "@testing-library/react": "12.1.5",
+        "@testing-library/jest-dom": "5.16.5",
```
unseen1980, line 61: Is the plan to have both enzyme and testing-library for some time?

.changeset/lazy-points-turn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/ui-components': minor
+---
+
+Adds UIGuidedAnswerComponent
```
Klaus-Keller, line 5:
```suggestion
Adds UIActionCallout
```

Klaus-Keller (changes requested):

packages/ui-components/package.json
unseen1980: ok thanks!

packages/ui-components/src/components/Icons.tsx
```diff
@@ -122,7 +122,8 @@ export enum UiIcons {
     Chevron = 'Chevron',
     GoToKey = 'GoToKey',
     File = 'File',
-    NoResults = 'NoResults'
+    NoResults = 'NoResults',
+    actionDetail = 'actionDetail'
```
unseen1980, line 126: I see ```UiIcons``` is alphabetically sorted, and it follows the pascal case convention, maybe we can keep it like this.

unseen1980 (approved):
Only a small comment, otherwise looks good. Thanks Iain

packages/ui-components/src/components/Icons.tsx
IainSAP: Thanks, its the wrong name entirely. Will rename.

devinea: Still not sorted (most others are alphabetical as @unseen1980 mentioned).

devinea (dismissed):

unseen1980 (approved):
The code looks good. Thanks 👍

devinea (dismissed):
Thanks @IainSAP

unseen1980 (approved):

Klaus-Keller (approved):
- test coverage looks great
- built and tested locally with storybook

Approved from my side, thanks @IainSAP!

IainSAP:
Thanks for the reviews @unseen1980 @devinea @Klaus-Keller

PR #824, opened 2022-12-05
fix(security): upgrade qs

donal-tobin-sap:
Issue: tools-suite #22655

supertest and express upgraded to get later version of qs


zdravko-georgiev (approved):
looks good to me

tobiasqueck (approved):
Tests in `axios-extension` run smoothly as before.

PR #825, opened 2022-12-06
fix(*-writer): remove the incomplete JSDoc annotations

petermuessig:
To lower the expectations for JS code completion, all related JSDoc annotations have been removed. This ensures that for all UI5 modules do not declare the types of the required modules.

The addition of the `tsconfig.json` and the usage of the ESM types is ok and could be a started for interested developers.

If developers are interested in using code completion for JS they should best take a look into the following documentation: https://github.com/SAP-samples/ui5-cap-event-app/tree/js-with-typescript-support#applying-typescript-benefits-to-a-javascript-application

This explains in detail what is possible with JSDoc annotations and what are the limitations. But developers should decide whether to use it or not as we cannot provide a proper tooling support for this.

If developers want to benefit from full code completion support the best option is to use TypeScript:
https://sap.github.io/ui5-typescript/


tobiasqueck (approved):
* tests run through
* while old but the changes still make sense because they make the life easier for js developers that want to use types (instead of using typescript)
* changeset ok

devinea (approved):
Code changes and snapshot updates as expected. 
Extended tests running locally. 
Changeset ✅

petermuessig:
Thanks for merging, just noticed it today...

PR #826, opened 2022-12-07
chore(ui-components): update storybook dependency, remove overrides

Klaus-Keller:
Update of storybook devDependency in module `@sap-ux/ui-components`
```
"@storybook/builder-webpack5": "6.5.14"
"@storybook/manager-webpack5": "6.5.14"
"@storybook/react": "6.5.14"
```

The enables us to remove the `override` of `loader-utils`.


tobiasqueck (dismissed):
Only reviewed the overrides doc and the package.jsons, and they look as expected. I didn't check if anything is broken in the ui-components.
Not sure though if you need a changeset for the ui-components, since we ship the generated storybook, don't we?

815are (approved):
Tested storybook - works ok

tobiasqueck (approved):

Issue #792, opened 2022-11-09
FEATURE - Environment Check: Introduce ability to perform checks for saved SAP systems in VSCode


cianmSAP:
As a developer using `@sap-ux/environment-check` in VSCode, I would like to run checks specific to my saved destination, so that I can get a detailed report useful for checking the saved destination in one go and also to be shared with SAP product support

### Description

- Environment Check for checking destination was introduced for SBAS which has helped SAP support to get preliminary information about destination setting from the customers
- Along the same lines, for saved system in VSCode, this US will check for various health checks which are applicable for saved system in order to make sure there are no system specific issue related to app generation, preview or deployment.
- Environment check for a saved system can check the following OData services:

1. OData V2 catalog service
2. OData V4 catalog service
3. ATO catalog (`/sap/bc/adt/ato/settings`)
4. SAPUI5 repository service (for deploy & undeploy) : `/sap/opu/odata/UI5/ABAP_REPOSITORY_SRV`
5. ADT service to fetch Transport Requests

### Design

**Functional Flow 1**
- User runs command `Fiori: Open Environment Check` and select `Check SAP Systems`.
- User is presented with a list of saved SAP systems
- User selects a system
- User is presented with options `View the results` and `View and save results`
- When user selects `View and save results` , this will save the results JSON and MD file in the workspace
- When user selects `View the results`, this will show the results in an MD file

**Results output:**

### Details for `SYSTEM X [user]`
✅ V2 catalog call returned 2434 services
🚫  V4 catalog service not available
✅  ATO catalog available
✅  SAPUI5 repository service for deployment available
✅  Ability to retrieve available Transport Requests

### Acceptance Criteria
- Given I am using `@sap-ux/environment-check` in VSCode, I should be able to run the environment check and get results of checks for a particular saved system in a .md and .json files with the ability to view or save the generated output file

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

cianmSAP:
Feature implemented

PR #829, opened 2022-12-12
fix(env-check): remove credentials from zip file

cianmSAP:
#792 

Remove unencrypted credentials from zip file
Fix escaping of system name in markdown


devinea (dismissed):
Changes and tests look good.

packages/environment-check/src/output/zip.ts
```diff
@@ -20,6 +20,27 @@ function byteNumberToSizeString(byteNumber: number): string {
     return `${parseFloat((byteNumber / Math.pow(1024, i)).toFixed(2))} ${units[i]}`;
 }
 
+/**
+ * Convert results to JSON and removes credentials from (stored system) endpoints.
+ *
+ * @param results - environment check results
+ * @returns json string
+ */
+function convertResultsToJson(results: EnvironmentCheckResult): string {
```
Klaus-Keller, line 29: @cianmSAP, any chance we remove the credential at the time we collect the data so they are never part of the results?

cianmSAP: Made updates to never return credentials from the exposed functions

packages/environment-check/src/checks/stored-system.ts
```diff
@@ -23,6 +24,21 @@ export async function checkStoredSystem(storedSystem: Endpoint): Promise<{
 }> {
     const logger = getLogger();
 
+    // retrieve system credentials
+    const storeService = await getService({ logger, entityName: 'system' });
```
Klaus-Keller, line 28:
```suggestion
    const storeService = await getService<BackendSystem, BackendSystemKey>({ logger, entityName: 'system' });
```

packages/environment-check/src/checks/stored-system.ts
```diff
@@ -23,6 +24,21 @@ export async function checkStoredSystem(storedSystem: Endpoint): Promise<{
...
+    const backendKey = new BackendSystemKey({ url: storedSystem.Url, client: storedSystem.Client });
+    const sapSystem = (await storeService.read(backendKey)) as BackendSystem;
```
Klaus-Keller, line 30:
```suggestion
    const sapSystem = await storeService.read(backendKey);
```

Klaus-Keller (changes requested):
This
```
const url = new URL(s.url).origin;
```
https://github.com/SAP/open-ux-tools/pull/829/files#diff-6ebfe3548141e9d4ea15436c74a18fe19e941f3fe4f0c08336a548f6c61e9023R122
leads to problems when you have a stored system with url that contains get parameters.

packages/environment-check/src/checks/stored-system.ts
cianmSAP: Updated

packages/environment-check/src/checks/stored-system.ts
cianmSAP: Updated

Klaus-Keller (approved):
- code looks good
- test coverage is ok
- I've built and tested locally using CLI and no user passwords are in the results


Approved from my side, thanks @cianmSAP!

devinea (approved):
Thanks @cianmSAP

Issue #634, opened 2022-08-24
FEATURE - `axios-extension` enhancement requests


tobiasqueck:
This is a collection of small enhancement requests for the `axios-extensions` received through private channels by consumers. 

_As a **consumer of the `axios-extension`**,
I want to **be able to inspect the axios errors when using the catalog service**,
so that I **can better analyze why something went wrong.**_

_As a **consumer of the `axios-extension`**,
I want to **be able to use service aliases for all main service implementations**,
so that I **can also use it with backends that block the standard service urls.**_

### Description

These enhancement requests, while received for specific scenarios, are valid for a broader range of scenarios that will also be required by other consumers including the SAP Fiori tools.

### Technical Design
All changes will be in the `axios-extension` module. The alias handling should mainly require to change the service factory functions. Exposing the error object will require changes to the catalog service, however, it makes sense to also check other convenience functions in our services.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I call the catalog service (v2 or v4)
**when** there is an error
**then** I will have access to the low-lever axios error object

**Given** I have a service not available at the default path but an alias
**when** when I call it
**then** I can provide the alias but still have full access to the convenience functions

### Notes
fyi @Klaus-Keller @cianmSAP 

### Tasks
- [ ] expose error objects
- [ ] enhance interface for aliases
- [ ] Unit Tests added

Klaus-Keller:
@tobiasqueck, with https://github.com/SAP/open-ux-tools/pull/830 we implemented  access to the full axios error object in `@sap-ux/environment-check`. There were no changes required in `axios-extension` and I propose to close this ticket.

PR #830, opened 2022-12-13
fix(env-check): details for catalog service errors

Klaus-Keller:
This addresses part 
_As a consumer of the `axios-extension`,
I want to be able to inspect the axios errors when using the catalog service,
so that I can better analyze why something went wrong._

of https://github.com/SAP/open-ux-tools/issues/634 for environment check. It seems that now the full error object is available to the consumer of `axios-extension`. This PR uses this object and logs detailed information in case an error occurs.


tobiasqueck (dismissed):

Klaus-Keller:
FYI: @cianmSAP

packages/environment-check/src/checks/service-checks.ts
```diff
@@ -143,7 +143,9 @@ async function catalogRequest(
         if (errorJson?.config?.auth?.password) {
             delete errorJson.config.auth.password;
         }
-        logger.debug(t('error.urlRequestFailure', { url, error: error.message, errorObj: error }));
+        logger.debug(
+            t('error.urlRequestFailure', { url, error: error.message, errorObj: JSON.stringify(errorJson, null, 4) })
```
cianmSAP, line 147: I'm still seeing the error 
`'Converting circular structure to JSON\n    --> starting at object with constructor 'TLSSocket'\n    |     property '_httpMessage' -> object with constructor 'ClientRequest'\n    --- property 'socket' closes the circle'`

To reproduce, any system which only has v2 or v4 (for me EC1 system `https://ccwdfgl9773.devint.net.sap:44300` I only have auth for v2) should throw an error in the catalog request.

Klaus-Keller: Hi @cianmSAP, 

I'm still not able to reproduce the issue you are facing, said system has for me both catalog services. However, I did some investigation and found this: https://github.com/axios/axios/issues/4486.

Looks like the issue is solved in `axios@1.x.x`, but unfortunately our module `@sap-ux/axios-extension` uses currently version `axios@0.24.0`.
@tobiasqueck @ullasholla any plans to update `axios` dependency in `@sap-ux/axios-extension`?

For now I'll look into a way to ignore circular structures, like done in [`circular-json`](https://www.npmjs.com/package/circular-json) or  [`json-stringify-safe`](https://www.npmjs.com/package/json-stringify-safe).

cianmSAP: Updates have resolved the error. Thanks @Klaus-Keller

cianmSAP (approved):
Comeplete error object is now visible ✅

packages/environment-check/src/formatter.ts
```diff
@@ -23,3 +23,36 @@ export function countNumberOfServices(catalogResult?: ODataServiceInfo[]): numbe
...
 }
+
+/**
+ * Returns replacer function that can be used with JSON.stringify to detect
+ * and replace circular structures.
+ *
+ * @example JSON.stringify(object, getCircularReplacer());
+ * @returns - replacer that replaces circular structures
+ */
+export function getCircularReplacer(): (key: string, value: any) => any {
```
ullasholla, line 34: This looks very handy. Do you mind adding a note in the PR description why this is needed?

It's good to have something written down for traceablity

packages/environment-check/src/formatter.ts
```diff
@@ -23,3 +23,36 @@ export function countNumberOfServices(catalogResult?: ODataServiceInfo[]): numbe
...
+            }
+            if (stack.includes(value)) {
+                value =
+                    stack[0] === value
+                        ? '|CIRCULAR STRUCTURE|'
+                        : '|CIRCULAR STRUCTURE .' + keys.slice(0, stack.indexOf(value)).join('.') + '|';
+            }
+        } else {
+            stack.push(value);
+        }
```
ullasholla, line 55: Use an early return and save one indentation level?

ullasholla: Reduces the cognitive complexity too as an upshot

packages/environment-check/src/formatter.ts
Klaus-Keller: Reason is there in the comments already, please refer to: https://github.com/SAP/open-ux-tools/pull/830#discussion_r1068731727

ullasholla: It's hard to find comments, especially in resolved conversations. Don't want to nitpick though, resolving

packages/environment-check/src/formatter.ts
```diff
@@ -23,3 +23,36 @@ export function countNumberOfServices(catalogResult?: ODataServiceInfo[]): numbe
...
+ *
+ * @example JSON.stringify(object, getCircularReplacer());
+ * @returns - replacer that replaces circular structures
+ */
+export function getCircularReplacer(): (key: string, value: any) => any {
+    const stack: any[] = [];
+    const keys: string[] = [];
+    return function replacer(key, value) {
+        if (stack.length > 0) {
+            const thisPos = stack.indexOf(this);
```
ullasholla, line 39: Use `pos` as a name instead of `thisPos`?

packages/environment-check/src/formatter.ts
```diff
@@ -23,3 +23,36 @@ export function countNumberOfServices(catalogResult?: ODataServiceInfo[]): numbe
...
+
+/**
+ * Returns replacer function that can be used with JSON.stringify to detect
+ * and replace circular structures.
+ *
+ * @example JSON.stringify(object, getCircularReplacer());
+ * @returns - replacer that replaces circular structures
+ */
+export function getCircularReplacer(): (key: string, value: any) => any {
+    const stack: any[] = [];
```
ullasholla, line 35: It's not really being used as a stack but as a storage for already seen objects. `seen` or something sound better?

ullasholla:
The example on MDN for replacing circular references is very readable but it does not nicely capture the path like yours does @Klaus-Keller.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value#circular_references

```javascript
const getCircularReplacer = () => {
  const seen = new WeakSet();
  return (key, value) => {
    if (typeof value === "object" && value !== null) {
      if (seen.has(value)) {
        return;
      }
      seen.add(value);
    }
    return value;
  };
};

JSON.stringify(circularReference, getCircularReplacer());
// {"otherData":123}
```

Klaus-Keller:
> The example on MDN for replacing circular references is very readable but it does not nicely capture the path like yours does @Klaus-Keller.
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value#circular_references
> 
> ```js
> const getCircularReplacer = () => {
>   const seen = new WeakSet();
>   return (key, value) => {
>     if (typeof value === "object" && value !== null) {
>       if (seen.has(value)) {
>         return;
>       }
>       seen.add(value);
>     }
>     return value;
>   };
> };
> 
> JSON.stringify(circularReference, getCircularReplacer());
> // {"otherData":123}
> ```

Thanks @ullasholla, that simplifies the algorithm and I don't think we really need the path to keys where the circular structure occurs. I committed the changes, please re-check.

packages/environment-check/src/formatter.ts
Klaus-Keller: changed it to the simplified algorithm

packages/environment-check/src/formatter.ts
Klaus-Keller: changed it to the simplified algorithm

packages/environment-check/src/formatter.ts
Klaus-Keller: changed it to the simplified algorithm

ullasholla (approved):
Thanks @Klaus-Keller.

The code is readable and the test coverage is good.

zdravko-georgiev (approved):

cianmSAP (approved):

Issue #837, opened 2022-12-14
TBI - Avoid duplicate custom views


stefanschreck:
### Description (include screenshots)
The given implementation for the creation of custom views at the FPM writer allows adding a new custom view to the existing list of views in the manifest. 
Indeed, it is not checked if the new custom view is already registered. As a consequence, duplicate view entries get generated.

### Value
The enhancement is required due to a problem that we have with asynchronous execution of manifest changes from different sides: as a consequence, the custom view may or may not exist already. If it already exists, its default content shall get overwritten by the new values passed to the the FPM writer.

### Architecture Elaboration
There are two options how to solve this: either merge the new custom view into the list of all given views (`existingViews` at https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/view/index.ts), or to remove the old entry from this list if the key is found. 
The second option has the advantage that the template can stay as it is - but the significant disadvantage that the new view is appended at the end, and this changes the view order. Thus the first option is preferred.

### Tasks
- [ ] Task 1

schreckstefan:
Solved by https://github.com/SAP/open-ux-tools/pull/838

PR #838, opened 2022-12-15
feat: consider existing custom view

stefanschreck:
TBI: https://github.com/SAP/open-ux-tools/issues/837

Solution:
- Instead of just appending a possibly duplicate custom view, the list of existing views is analyzed; if the key is already present, the existing entry is taken and gets overwritten.
- A new API parameter `viewUpdate` is  introduced, with default value `true`; if set, the fragment will not be (over)written.


packages/fe-fpm-writer/test/integration/index.test.ts
```diff
@@ -147,6 +147,89 @@ describe('use FPM with existing apps', () => {
...
+                    key: 'CustomViewKey',
+                    label: 'Custom View',
+                    name: 'NewCustomView',
+                    eventHandler: true,
+                    ...config.settings
+                },
+                fs,
+                false
+            );
+        });
```
815are, line 232: In case of integration tests - I think by adding `fs.extendJSON(join(config.path, 'webapp/manifest.json'), {` on each test we are losing progress from previous.

For example if I remove test `generateCustomView in ListReport, second custom view`, then integration test will not fail.

Can we use `generateCustomView in ListReport, custom view to be overwritten` right after `generateCustomView in ListReport` - so that we do not need use `fs.extendJSON(join(config.path, 'webapp/manifest.json'` and then trigger 'generateCustomView in ListReport, second custom view' - so that 3rd view will be added to snapshot?

stefanschreck: @815are Of course I can change the order of tests - this slightly adapts the snapshot, as you indicated. See following commit.
Indeed, the really important tests come with  packages/fe-fpm-writer/test/unit/view.test.ts

815are: @stefanschreck 

I applied additional correction to integration tests - d94e9e306443864516f9882c82345c6cabf44623
my main point was about `fs.extendJSON(join(config.path, 'webapp/manifest.json'), {`.
I mean we have 3 tests regarding views update and in each we overwrite result of previous by `fs.extendJSON(join(config.path, 'webapp/manifest.json'), {`.

For example:
1. Test `generateCustomView in ListReport` - it created 2 views `LineItemView` and `CustomViewKey`
2. Test 'generateCustomView in ListReport, custom view to be overwritten' - overwrites manifest with views  `LineItemView` and `CustomViewKey`. In case if 1st test failed, then end expectation will not fail, because overwrite happened. At then end of test we have `LineItemView` and `CustomViewKey`
3. Test `generateCustomView in ListReport, second custom view`. Old result is overwritten by views `LineItemView`, `FirstCustomView` and then test runs by adding additional view.

In result at then end snapshot 1st and 2nd test will not be checked, because of overwrites in 2nd and 3rd.

815are (dismissed):
Tested with integration and unit tests - works ok.
But I did not test with real app

packages/fe-fpm-writer/src/view/index.ts
```diff
@@ -51,9 +79,10 @@ function enhanceConfig(fs: Editor, data: CustomView, manifestPath: string, manif
  * @param {string} basePath - the base path
  * @param {CustomView} customView - the custom view configuration
  * @param {Editor} [fs] - the mem-fs editor instance
+ * @param {boolean} viewUpdate - indicates if th view fragment shall be created or overwritten
  * @returns {Promise<Editor>} the updated mem-fs editor instance
  */
-export function generateCustomView(basePath: string, customView: CustomView, fs?: Editor): Editor {
+export function generateCustomView(basePath: string, customView: CustomView, fs?: Editor, viewUpdate = true): Editor {
```
815are, line 85: Just small observation, I am only not sure if we should introduce new API parameter?
I think we usually use 2nd param to pass configuration, for example `overwrite` for `controller extension`:
https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/controller-extension/types.ts#L39

devinea (approved):
Changes and tests look good. 
I haven' tested locally.

zdravko-georgiev (changes requested):
@stefanschreck Could we please what @815are is suggesting? I think it is better if don't introduce additional params, but just extend the definition of the second param

815are (dismissed):

devinea (approved):

packages/fe-fpm-writer/src/view/types.ts
```diff
@@ -21,6 +21,11 @@ export interface CustomView extends CustomElement, EventHandler {
      * If set to true, a table macro control will be generated.
      */
     control?: string | true;
+
+    /**
+     * Indicates that the view shall be crated or updated
```
zdravko-georgiev, line 26: typo it should be `created`

stefanschreck: done

zdravko-georgiev: thx

zdravko-georgiev (approved):
LGTM

PR #843, opened 2022-12-16
chore(logger) update winston deps

devinea:
Align deps with versions in tools-suite


longieirl (approved):

tobiasqueck (approved):
I am assuming they have be intensively tested in the FIori tools, therefore, the version change should be risk-free.

devinea:
Thanks all

Issue #844, opened 2022-12-20
TBI - TS - Add local type definition for Fiori Elements V2 on UI5 version less than 1.109.1


devinea:
### Description (include screenshots)
What will this TBI achieve, and how will it be measured as done. Is it documentation, increase of test coverage, refactoring

TBI - TS - Add local type definition for Fiori Elements V2 on UI5 version less than 1.109.1 as the type definition is not available in lower versions of ui5. 

### Value
Optional - details on how this technical backlog item provides value to end-users or dev team eg. Performance, Quality, Maintainability. This will help determine the ranking in relation to feature development

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

tobiasqueck:
@devinea looks to me like it is done, so I am closing it, please reopen and explain what is missing if I was wrong.

Issue #754, opened 2022-10-21
TBI - Better way to handle Typescript commands across all template types


longieirl:
### Description (include screenshots)
As part of the changes introduced by [TBI 746](https://github.com/SAP/open-ux-tools/issues/746), there needs to be further changes made. The changes introduced were to support the basic template only. TS support and what was introduced in for the Simple template needs to support all template types.

- remove conditional code added in this PR https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-freestyle-writer/src/packageConfig.ts#L66
- enhance the template in ui5-application-writer to support adding TS commands
- validate and fix all provided ts templates in fiori-*-writer to not throw errors when running a typecheck

### Value
Optional - details on how this technical backlog item provides value to end-users or dev team eg. Performance, Quality, Maintainability. This will help determine the ranking in relation to feature development

### Architecture Elaboration
Does it requires architecture elaboration?

### Notes
There have been issues with other template types and Typescipt which is why the change was restricted to Simple only.

### Tasks
- [x] Remove conditional code added in this PR https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-freestyle-writer/src/packageConfig.ts#L66
- [x] Enhance the template in ui5-application-writer to support adding TS commands
- [x] Validate and fix all provided ts templates in fiori-*-writer to not throw errors when running a typecheck  https://github.com/SAP/open-ux-tools/pull/845

devinea:
Closing. As tasks where completed

PR #845, opened 2022-12-20
TBI - Add local type definition for Fiori Elements V2 on UI5 version less than 1.108.0

devinea:
https://github.com/SAP/open-ux-tools/issues/844
https://github.com/SAP/open-ux-tools/issues/754

Changes included
- Add local type definition for Fiori Elements V2 on UI5 version less than `~1.108.0`
- For Fiori Elements V2 on UI5 version  `>=1.108.0` use the provided types in `~1.108.0`


packages/ui5-application-writer/test/defaults.test.ts
```diff
@@ -15,12 +15,13 @@ describe('Defaults', () => {
...
+        ['1.109.1', '~1.109.0'],
         [UI5_DEFAULT.TYPES_VERSION_BEST, typesVersionBest],
-        ['1.109-snapshot', typesVersionBest],
+        ['1.109-snapshot', '~1.109.0'],
         ['1.80-snapshot', esmTypesVersionSince],
-        ['1.102-snapshot', '~1.102.0'],
-        ['1.91.0', '~1.91.0']
+        ['1.102-snapshot', typesVersionBest],
+        ['1.91.0', '~1.91.0'],
+        [UI5_DEFAULT.UI5_VERSION_VARIABLE, 'latest']
```
devinea, line 24: Currently 'latest' for [@sapui5/ts-types-esm](https://www.npmjs.com/package/@sapui5/ts-types-esm?activeTab=versions) is `1.108.4` and not the available `1.109.3`
Screenshot 2022-12-20 at 11 58 52

@petermuessig - does `latest` point to latest LTS version only and not the highest available version normally?

devinea: I think we can opt for `>=1.109.1` if the minUI5Version is `${sap.ui5.dist.version}`

tobiasqueck (changes requested):
Could you please enhance the description of the PR and explain what is included here. I understand the part of the additional `ui5.d.ts` if we have an FEv2 app and use a UI5 version smaller than `1.109.1`, however, I struggle to understand the part about the `TYPES_VERSION_BEST_MIN`

devinea:
> Could you please enhance the description of the PR and explain what is included here. I understand the part of the additional `ui5.d.ts` if we have an FEv2 app and use a UI5 version smaller than `1.109.1`, however, I struggle to understand the part about the `TYPES_VERSION_BEST_MIN`

@tobiasqueck 
Update description https://github.com/SAP/open-ux-tools/pull/845#issue-1504437319

tobiasqueck:
> Previously the logic limited esm version to TYPES_VERSION_BEST (i.e. ~1.108.0) in too many cases

That was done on purpose because we cannot update the types of a UI5 version that is not the latest anymore and also not a LTS version e.g. once `1.110` is released, fixing a type for `1.109` is (practically) impossible. However, we can downport to the latest LTS version e.g. in this case `1.108`, therefore, we should use the latest LTS version as much as possible.

packages/fiori-elements-writer/src/index.ts
```diff
@@ -53,7 +53,12 @@ async function generate<T>(basePath: string, data: FioriElementsApp<T>, fs?: Edi
     const rootTemplatesPath = join(__dirname, '..', 'templates');
     // Add templates common to all template types
     // Common files
-    const ignore = [feApp.appOptions?.typescript ? '**/*.js' : '**/*.ts'];
+    const isV2FETypesAvailable = feApp.ui5?.version
```
IainSAP, line 56: What does V2 refer to here please? There is no check for odata version. 

Also the same coercion of ui5 version is being done multiple times and could be consolidated. Probably this type script  specific code should be moved to another function so it can be tested in isolation. I think all this TS related code should only be done if TS is requested but I dont se that check being done either?

devinea: Add comment regarding isV2FETypesAvailable and add check for typescript flag before executing the logic.

packages/fiori-elements-writer/src/index.ts
```diff
@@ -42,18 +42,36 @@ async function generate<T extends {}>(basePath: string, data: FioriElementsApp<T
...
     const rootTemplatesPath = join(__dirname, '..', 'templates');
     // Add templates common to all template types
-    // Common files
-    const ignore = [feApp.appOptions?.typescript ? '**/*.js' : '**/*.ts'];
+    const jsIgnoreGlob = ['**/*.ts'];
+
+    let ignore = jsIgnoreGlob;
+    if (feApp.appOptions?.typescript === true) {
+        // isV2FETypesAvailable - Boolean to indicate if V2 Fiori Element types were available in the UI5 version
+        const isV2FETypesAvailable = feApp.ui5?.version ? semVer.gte(coercedUI5Version, '1.108.0') : false;
```
tobiasqueck, line 59: You are using a magic string `1.108.0` however, you also declared a constant in the `ui5-application-writer` with `1.108.0` for the FE types available version. Shouldn't you refer to the same constant in both cases?

tobiasqueck: Continued thinking about that one: having a variable `V2_FE_TYPES_AVAILABLE` in `ui5-application-writer` that is only required for FE is not ok.
Please rethink, how you could keep the FE specifics completely in the fe writer.

tobiasqueck (dismissed):
Please see my comments about the FE types variable

packages/ui5-application-writer/src/data/defaults.ts
```diff
@@ -57,10 +57,13 @@ export const enum UI5_DEFAULT {
     SAPUI5_CDN = 'https://ui5.sap.com',
     OPENUI5_CDN = 'https://openui5.hana.ondemand.com',
     TYPES_VERSION_SINCE = '1.76.0',
+    TYPES_VERSION_BEST_MIN = '1.102.0',
     TYPES_VERSION_PREVIOUS = '1.71.15',
     TYPES_VERSION_BEST = '1.108.0',
+    V2_FE_TYPES_AVAILABLE = '1.108.0',
     ESM_TYPES_VERSION_SINCE = '1.94.0',
     MANIFEST_VERSION = '1.12.0',
+    UI5_VERSION_VARIABLE = '${sap.ui5.dist.version}',
```
IainSAP, line 66: This string is S/4 pipeline specific. It should never be "seen" by the open ux writers. Instead the value should be replaced by the caller of the writer.

devinea: 👍 removed in https://github.com/SAP/open-ux-tools/pull/845/commits/53dc7bccc45149585893580868b322ed87039676

packages/fiori-elements-writer/src/index.ts
devinea: Fixes applied in https://github.com/SAP/open-ux-tools/pull/845/commits/53dc7bccc45149585893580868b322ed87039676

packages/ui5-application-writer/src/index.ts
```diff
@@ -126,4 +126,4 @@ async function enableTypescript(basePath: string, fs?: Editor): Promise<Editor>
 }
 
 export { Ui5App, generate, enableTypescript, isTypescriptEnabled };
-export { App, Package, UI5, AppOptions };
+export { App, Package, UI5, AppOptions, UI5_DEFAULT };
```
tobiasqueck, line 129: Do we need the new export here? If I didn't miss anything then you only import it in a test and there you refer to the source file directly.

devinea: Removed in https://github.com/SAP/open-ux-tools/pull/845/commits/30c4337bcd33c2dff5e3e8ae4abd167fe26be480

packages/fiori-elements-writer/src/index.ts
```diff
@@ -42,18 +44,36 @@ async function generate<T extends {}>(basePath: string, data: FioriElementsApp<T
...
 
     // Add new files from templates e.g.
     const rootTemplatesPath = join(__dirname, '..', 'templates');
     // Add templates common to all template types
-    // Common files
-    const ignore = [feApp.appOptions?.typescript ? '**/*.js' : '**/*.ts'];
+    const jsIgnoreGlob = ['**/*.ts'];
+
+    let ignore = jsIgnoreGlob;
+    if (feApp.appOptions?.typescript === true) {
```
IainSAP, line 59: Should this block be refactored (move to a new function `generateIgnoreGlob`) to reduce complexity as highlighted by Sonar (16 > 15)?

IainSAP (commented):
Comment re. code complexity (as reported by Sonar)

packages/ui5-application-writer/test/defaults.test.ts
```diff
@@ -15,11 +15,11 @@ describe('Defaults', () => {
         ['1.80-snapshot', esmTypesVersionSince],
         ['metadata', typesVersionBest],
         [undefined, typesVersionBest],
-        ['1.109.1', typesVersionBest],
+        ['1.109.1', '~1.108.0'],
```
ullasholla, line 18: Doesn't `typesVersionBest` evaluate to the same value? Why replace the constant with the literal value?

packages/ui5-application-writer/test/defaults.test.ts
```diff
@@ -15,11 +15,11 @@ describe('Defaults', () => {
...
         [UI5_DEFAULT.TYPES_VERSION_BEST, typesVersionBest],
-        ['1.109-snapshot', typesVersionBest],
+        ['1.109-snapshot', '~1.108.0'],
```
ullasholla, line 20: Doesn't `typesVersionBest` evaluate to the same value? Why replace the constant with the literal value?

packages/ui5-application-writer/test/defaults.test.ts
```diff
@@ -15,11 +15,11 @@ describe('Defaults', () => {
...
         ['1.80-snapshot', esmTypesVersionSince],
-        ['1.102-snapshot', '~1.102.0'],
+        ['1.102-snapshot', '~1.108.0'],
```
ullasholla, line 22: Use `typesVersionBest`?

.vscode/launch.json
```diff
@@ -56,6 +66,16 @@
...
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "fiori-elements-writer: Debug Current Jest File",
+            "program": "${workspaceFolder}/node_modules/jest/bin/jest.js",
+            "args": ["${file}", "--config", "jest.config.js", "--coverage=false"],
+            "console": "integratedTerminal",
+            "internalConsoleOptions": "neverOpen",
+            "cwd": "${workspaceFolder}/packages/fiori-elements-writer"
+        },
```
ullasholla, line 78: Do we still need these configurations? The instructions here work well: https://github.com/SAP/open-ux-tools/blob/c77e420bf59ad83bef809b44bfc1e0690da99052/README.md?plain=1#L96

packages/fiori-elements-writer/src/index.ts
```diff
@@ -17,6 +17,8 @@ import { changesPreviewToVersion, escapeFLPText } from './data/templateAttribute
 import { extendManifestJson } from './data/manifestSettings';
 import semVer from 'semver';
 
+export const V2_FE_TYPES_AVAILABLE = '1.108.0';
```
ullasholla, line 20: Is this the same as `UI5_DEFAULT.TYPES_VERSION_BEST`? If so, can we reuse it?

ullasholla: Is this because of this conversation? https://github.com/SAP/open-ux-tools/pull/845#discussion_r1150058145

packages/fiori-elements-writer/src/index.ts
```diff
@@ -42,18 +44,36 @@ async function generate<T extends {}>(basePath: string, data: FioriElementsApp<T
...
+    let ignore = jsIgnoreGlob;
+    if (feApp.appOptions?.typescript === true) {
+        // isV2FETypesAvailable - Boolean to indicate if V2 Fiori Element types were available in the UI5 version
+        const isV2FETypesAvailable = feApp.ui5?.version ? semVer.gte(coercedUI5Version, V2_FE_TYPES_AVAILABLE) : false;
+        const tsIgnoreGlob = ['**/*.js'];
+        ignore = tsIgnoreGlob;
+        // Add local ui5.d.ts if types are missing in UI5 version for V2 Odata services
+        // OR template is OVP
+        if (feApp.service.version === OdataVersion.v2) {
+            if (isV2FETypesAvailable) {
```
ullasholla, line 67: Should this be `!isV2FETypesAvailable` instead?
Also can the scope of `isV2FETypesAvailable` be reduced by moving closer to the usage?

.vscode/launch.json
devinea: I think it is more convenient tbh.

packages/fiori-elements-writer/src/index.ts
devinea: Yes, moved here because of https://github.com/SAP/open-ux-tools/pull/845#discussion_r1150058145

But `V2_FE_TYPES_AVAILABLE` and `UI5_DEFAULT.TYPES_VERSION_BEST` have different semantic meaning. 
`V2_FE_TYPES_AVAILABLE` won't change over time. But `TYPES_VERSION_BEST` will be updated at some point to a newer UI5 LTS version.

packages/ui5-application-writer/test/defaults.test.ts
devinea: Done in e7992de982214fffc7fbc028ab8933b4530df2c6

packages/ui5-application-writer/test/defaults.test.ts
devinea: Done in e7992de982214fffc7fbc028ab8933b4530df2c6

packages/ui5-application-writer/test/defaults.test.ts
devinea: Done in e7992de982214fffc7fbc028ab8933b4530df2c6

packages/fiori-elements-writer/src/index.ts
devinea: done in d28610cad53b024aa586e2df6e9c5c1210a39b47

IainSAP (approved):
Thx @devinea 

- Comments addressed

packages/fiori-elements-writer/src/index.ts
```diff
@@ -17,6 +17,33 @@ import { changesPreviewToVersion, escapeFLPText } from './data/templateAttribute
...
+function getTypeScriptIgnoreGlob<T extends {}>(feApp: FioriElementsApp<T>, coercedUI5Version: semVer.SemVer): string[] {
+    let ignore = [];
+    // isV2FETypesAvailable - Boolean to indicate if V2 Fiori Element types were available in the UI5 version
+    const isV2FETypesAvailable = feApp.ui5?.version ? semVer.gte(coercedUI5Version, V2_FE_TYPES_AVAILABLE) : false;
+    const tsIgnoreGlob = ['**/*.js'];
+    ignore = tsIgnoreGlob;
+    // Add local ui5.d.ts if types are missing in UI5 version for V2 Odata services
+    // OR template is OVP
+    if (feApp.service.version === OdataVersion.v2) {
+        if (isV2FETypesAvailable) {
```
ullasholla, line 37: Should this check be reversed? Should we ignore `ui5.d.ts` if the types are missing and not the other way around?

packages/fiori-elements-writer/src/index.ts
ullasholla: The code location has moved, hence the `Outdated` label.
I'm still wondering about this: https://github.com/SAP/open-ux-tools/pull/845#discussion_r1158804536

ullasholla (changes requested):
Thanks for addressing the review comments @devinea.
Was wondering about this one: https://github.com/SAP/open-ux-tools/pull/845#discussion_r1158804536

packages/fiori-elements-writer/src/index.ts
ullasholla: @devinea clarified to me that the check is correct - we're not copying the types if they're available. Thanks!

ullasholla (approved):
Thanks @devinea, the PR looks good to me.
All the previous review comments have been addressed.

PR #847, opened 2022-12-20
TBI - refactor pre* scripts for start and build for typescript projects

devinea:
tbi #754 

TBI - refactor pre* scripts for start and build for typescript projects
- Update ESM_TYPES_VERSION_SINCE = '1.94.0'  https://github.com/SAP/open-ux-tools/pull/847/commits/74798d0f94fc4bac1b8712f8a666529e8d5b6eaa to solve issues below

Open issues:
- [x] type error in `packages/fiori-freestyle-writer/test/test-output/worklistTemplate/worklist_ts`
- [x] type error in `packages/fiori-freestyle-writer/test/test-output/listDetailTemplate/listdetail-ts`
Screenshot 2022-12-20 at 18 21 44
using newer `@sapui5/ts-types-esm` `~1.94.0` or `~1.96.0` for example solves the issue.


tobiasqueck (changes requested):
TL;TR: please re-discuss pre-scripts

I am not sure I like the idea of the `pre-` scripts. Can you explain why we need them? 
We are in a bit of an odd situation here, we use `tsc` for the typechecks but `babel` for the actual transpiling, so the build might work well even with `tsc` failing. 
My recommendation is to not the `prestart` script since it isn't required and might become more annoying than useful. About `prebuild` I am torn. I am just afraid that it leads to lots of `@ts-ignore` to just get past the `prebuild` knowing that the actual build works.

packages/ui5-application-writer/src/data/index.ts
```diff
@@ -24,6 +24,11 @@ export function mergeWithDefaults(ui5App: Ui5App): {
     }
     ui5App.ui5 = mergeUi5(ui5App.ui5 || {}, ui5App.appOptions);
     ui5App.package = mergeObjects(packageDefaults(ui5App.package.version, ui5App.app.description), ui5App.package);
+    // if typescript is enabled and pre scripts for start and build scripts
```
tobiasqueck, line 27: Independent of my comment about not adding them at, if we add them, shouldn't they be added here: https://github.com/SAP/open-ux-tools/blob/tbi/754/type_check_scripts/packages/ui5-application-writer/templates/optional/typescript/package.json#L3

devinea: Oh yeah, definitely. Thanks. Done in 6ee2ff2994b682f6eb18d4e9137844cce291c257

devinea:
> TL;TR: please re-discuss pre-scripts
> 
> I am not sure I like the idea of the `pre-` scripts. Can you explain why we need them? We are in a bit of an odd situation here, we use `tsc` for the typechecks but `babel` for the actual transpiling, so the build might work well even with `tsc` failing. 

Without enforcing the type checking in build scripts then it is only shown in VSC/BAS Editor which can be easily missed / ignored. If the type checking is not enforced then you are not really benefiting (apart from Code Assist) from TypeScript. Better to ensure type checking in enabled from the project generation for cleaner code IMO.

Since babel does not check types we needs `tsc` to do this as mentioned in 
https://www.typescriptlang.org/docs/handbook/babel-with-typescript.html#babel-for-transpiling-tsc-for-types

> My recommendation is to not the `prestart` script since it isn't required and might become more annoying than useful. 

I think `prestart` is useful to give earlier feedback to the developer. Another option I was considering was to use [npm-run-all](https://www.npmjs.com/package/npm-run-all) to run the preview script and tsc with watch option in parallel (we have similar in tools-suite) . 
This would continuously give type checking feedback to the developer in the terminal while running the preview, which is a much better experience IIMO. Just a little reluctant to add another devDependency.

> About `prebuild` I am torn. I am just afraid that it leads to lots of `@ts-ignore` to just get past the `prebuild` knowing that the actual build works.
 
I think this is a risk especially if preview does not run type checking with `prestart` or similar. 
If developers don't want type checking then maybe TypeScript was not the right choice for their project!!!

tobiasqueck (approved):
Thanks for the reasoning and moving the scripts to the template folder.

IainSAP (approved):
Thx @devinea.

Issue #848, opened 2022-12-20
TBI - Stop adding a dummy datasource to a project with no datasource selected during generation


devinea:
### Description (include screenshots)
Stop adding a dummy datasource to a project with no datasource selected during generation

### Value
Stop adding miss-leading datasource that is not valid and cannot be sync'd. 
A new service can be added via Service Manager and mockserver via `npx @sap-ux/create add mockserver-config`

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

devinea:
done in https://github.com/SAP/open-ux-tools/pull/849

PR #849, opened 2022-12-20
tbi - stop adding a dummy service when no datasource supplied.

devinea:
https://github.com/SAP/open-ux-tools/issues/848

### DIFF
--- a/packages/fiori-freestyle-writer/src/index.ts
+++ b/packages/fiori-freestyle-writer/src/index.ts
@@ -67,7 +67,7 @@ async function generate<T>(basePath: string, data: FreestyleApp<T>, fs?: Editor)
 
     packageJson.scripts = Object.assign(packageJson.scripts, {
         ...getPackageJsonTasks({
-            localOnly: !ffApp.service?.url,
+            localOnly: !!ffApp.service && !ffApp.service?.url,
             addMock: !!ffApp.service?.metadata,
             sapClient: ffApp.service?.client,
             flpAppId: ffApp.app.flpAppId,
--- a/packages/fiori-freestyle-writer/src/index.ts
+++ b/packages/fiori-freestyle-writer/src/index.ts
@@ -9,6 +9,7 @@ import cloneDeep from 'lodash/cloneDeep';
 import type { BasicAppSettings } from './types';
 import { FreestyleApp, TemplateType } from './types';
 import { setDefaults, escapeFLPText } from './defaults';
+import { UI5Config } from '@sap-ux/ui5-config';
 
 /**
  * Generate a UI5 application based on the specified Fiori Freestyle floorplan template.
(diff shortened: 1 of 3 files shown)


tobiasqueck (dismissed):
Thanks for removing the unnecessary configs. Changes look good, the updated snapshots as well.

zdravko-georgiev (dismissed):
looks good

IainSAP (changes requested):
@devinea This is a non-trivial change in behaviour for consumers who will now need to explicitly add a datasource where none is provided. Please adjust the version accordingly so this is clearly signposted. Thanks.

tobiasqueck:
> @devinea This is a non-trivial change in behaviour for consumers who will now need to explicitly add a datasource where none is provided. Please adjust the version accordingly so this is clearly signposted. Thanks.

@IainSAP can you explain this a little more? In what scenario is the dummy data source required?

.changeset/early-dancers-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/fiori-freestyle-writer': patch
```
IainSAP, line 2: Should be minor at least since the behaviour is changed and a feature removed.

IainSAP:
> > @devinea This is a non-trivial change in behaviour for consumers who will now need to explicitly add a datasource where none is provided. Please adjust the version accordingly so this is clearly signposted. Thanks.
> 
> @IainSAP can you explain this a little more? In what scenario is the dummy data source required?



> > @devinea This is a non-trivial change in behaviour for consumers who will now need to explicitly add a datasource where none is provided. Please adjust the version accordingly so this is clearly signposted. Thanks.
> 
> @IainSAP can you explain this a little more? In what scenario is the dummy data source required?

This behaviour is currently in use by `@sap/generator-fiori` at least. Needs a minor version bump to flag a change in behaviour.

tobiasqueck:
> it was implemented by design originally rather than by accident 

Can you explain the reasoning for the design? I just want it for the records in case we get regressions reported later to easily pinpoint where it comes from.

IainSAP:
> > it was implemented by design originally rather than by accident
> 
> Can you explain the reasoning for the design? I just want it for the records in case we get regressions reported later to easily pinpoint where it comes from.

I believe (its a long time ago!) its a placeholder to allow a dev to follow the updates manually required (before we had tooling support) - and I think its needed to run preview where no datasource was provided. This can of course be added after writing now but would require additional effort on the consumer side?

.changeset/early-dancers-eat.md
devinea: Done in 0f829edc6fa15e3886af6ccfd3642f25e16ec461 . Thanks.

devinea:
> > > it was implemented by design originally rather than by accident
> > 
> > 
> > Can you explain the reasoning for the design? I just want it for the records in case we get regressions reported later to easily pinpoint where it comes from.
> 
> I believe (its a long time ago!) its a placeholder to allow a dev to follow all the entries required to add a datasource (before we had tooling support) - and I think its needed to run preview where no datasource was provided. This can of course be added after writing now but would require additional effort on the consumer side and so probably not a patch bump?

Yes there were some gaps before that required this entry. 
I have tested: 
- Generate project with no datasource. 
- Appears in App Info. 
- Preview works (start, start-local, start-noflp)
- Service Manager Add Service works
- Guided Dev detects no datasource 

Changes for tools-suite
- Service Modeler  - `An unexpected error occurred, Unable to find relevant SAP Fiori tools project..`. Return more appropriate message. 
- App Info - `Main Service`  field is blank. Maybe show `None` instead.

So there is no obvious blockers to merge this and the small updates above will be dealt with in Tools-Suite.

tobiasqueck (approved):

.changeset/early-dancers-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/fiori-freestyle-writer': minor
```
IainSAP, line 2: 👍

IainSAP (approved):
Thx @devinea 

- Version bump minor as expected
- Tests updated
- Expected updates required by consumers (and tests)

devinea:
Thanks all

Issue #764, opened 2022-10-26
BUG - errors shown in preview log when using typescript code


cfg74:
### Description
When previewing a typescript enabled SAP Fiori elements app with LR and OP (created with Fiori generator, all standard settings), I see errors in log (see screenshot below). It does appear to have no influence on the preview function afterwards.

### Expected results
No (irrelevant) typescript related errors in preview log

### Actual results
Typescript related errors are shown which don't influence the preview workings

### Screenshots
![image](https://user-images.githubusercontent.com/66320096/198076351-d3765675-da8d-4f77-aa6f-a200e8f4c6a9.png)

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
https://github.com/ui5-community/ui5-ecosystem-showcase/issues/650

### Fix
{describe the fix} 

### Why was it missed
Our testing validated that both build and preview work, the log output was not checked, and the wrong errors where not considered as a problem.

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

cfg74:
fyi @tobiasqueck

tobiasqueck:
@cfg74 could you please add the following to your `ui5.yaml` and let me know if the errors are gone? If yes, then we can generate this as part of the writers because the `Component-preload.js` is something that never exists in your local dev scenario.
```
    - name: ui5-tooling-transpile-middleware
      afterMiddleware: compression
      configuration:
        excludePatterns:
        - /Component-preload.js
```

@petermuessig I think the documentation (or the implementation) of the `ui5-transpile-module` is incorrect. Based on the doc, I'd expected that I need to exclude `Component-preload.js`, however, it required a leading slash.

cfg74:
@tobiasqueck sorry for the late answer, yes this removes the error. There is another file `/utils/locate-reuse-libs.ts` which would need to be excluded as well. But I'm wondering, what happens if you have a mixture  of `.js` and `.ts` custom code in your app, which is supposed to work, right?

devinea:
https://github.com/ui5-community/ui5-ecosystem-showcase/issues/650 is done.
Fix with latest version: https://www.npmjs.com/package/ui5-tooling-transpile `0.2.1`

PR created to update templates https://github.com/SAP/open-ux-tools/pull/846

PR #850, opened 2022-12-21
bug - hide transpile warnings for `/Component-preload.js`

devinea:
#764 

- hide transpile warnings for `/Component-preload.js` per https://github.com/SAP/open-ux-tools/issues/764#issuecomment-1293114225

Screenshot 2022-12-21 at 15 06 08
Screenshot 2022-12-21 at 15 06 13


tobiasqueck (approved):
Thanks for adding the exclude for the Component-preload so that we get rid of the warning.

IainSAP (approved):
Thx @devinea

Issue #851, opened 2022-12-21
BUG (axios-extension) - TypeError: Cannot read properties of undefined (reading 'headers')


cianmSAP:
### Description
When retrieving Transport Requests for a given system, the following error is being thrown 

`TypeError: Cannot read properties of undefined (reading 'headers')`

### Steps to Reproduce
1. Create an instance of an ABAP service provider using `createForAbap`
2. Use the service provider to retrieve an instance of AdtService subclass e.g `abapServiceProvider.getAdtService(TransportChecksService);`
3. Use this to retrieve the transport requests for a given package e.g `adtService.getTransportRequests('package_name', 'app_name')`

### Expected results
Transport requests to be retrieved succesfully 

### Actual results
Error is thrown (see screenshot)

### Screenshots

<img width="945" alt="Screenshot 2022-12-21 at 16 22 48" src="https://user-images.githubusercontent.com/99677697/208954094-e89c9780-030f-4c00-9829-9e7d0e712d88.png">

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

 FYI @tobiasqueck

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #852, opened 2022-12-21
fix(axios-ext): make response property on error object optional  

cianmSAP:
https://github.com/SAP/open-ux-tools/issues/851

Reponse property is not always present on error object and throws an unwanted error - making it optional

### DIFF
--- a/.changeset/tiny-horses-act.md
+++ b/.changeset/tiny-horses-act.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/axios-extension': minor
+---
+
+minor bug fix
--- a/packages/axios-extension/src/auth/connection.ts
+++ b/packages/axios-extension/src/auth/connection.ts
@@ -163,7 +163,7 @@ export function attachConnectionHandler(provider: ServiceProvider) {
         },
         (error: AxiosError) => {
             // remember xsrf token if provided even on error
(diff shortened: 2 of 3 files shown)


tobiasqueck (changes requested):
The fix is of course good, and covers unexpected situations. However, looking at the code it it covers with a catch, there should only be a `AxiosError` (we even typed it this way), so could you please investigate what is happening in your scenario and why a different error is thrown and maybe fix the root cause for it (in addition to your null check)

cianmSAP:
> The fix is of course good, and covers unexpected situations. However, looking at the code it it covers with a catch, there should only be a `AxiosError` (we even typed it this way), so could you please investigate what is happening in your scenario and why a different error is thrown and maybe fix the root cause for it (in addition to your null check)

I came across this error when investigating another issue in app-gen, the error is thrown when ignoreCertErrors is set to false. The error is thrown as seen below, so while the retrieval on transport requests will ultimately fail, I'd expect to see this error rather than the `TypeError :  cannot read properties of undefined`

image

@tobiasqueck

tobiasqueck (approved):
Thanks for the explanation, I didn't realize that response can be undefined for an `AxiosError` but you proved that I was wrong.

devinea (approved):
Thanks @cianmSAP 
changes and tests look good.

Issue #853, opened 2022-12-22
BUG - Default FluentUI dialog animations are not being applied to UI-components


sanat-thukral:
### Related Feature
[Feature request:  GA #330](https://github.com/SAP/guided-answers-extension/issues/330)

### Description

By default Dialogs in FluentUI dialogs have transitions on them. However, when we use UIDialog, the animations only work on exit and not both.

### Steps to Reproduce

Create a new Dialog using the UIDialog component and the actual output should be reflected.

This is the code from Guided Answers, where the problem was first seen. Could be a good reference to see the problem  [(GA code reference)](https://github.com/SAP/guided-answers-extension/pull/335/commits/5cf2d06a72e82428354aebea743cbb2618e6a281#:~:text=%3CUIDialog,/UIDialog%3E)

### Expected results

[From FluentUI GH](https://github.com/microsoft/fluentui/pull/530 )

### Actual results

https://user-images.githubusercontent.com/104602674/209118078-4c2fff8f-a680-4ad5-a17e-d659f670bd08.mov

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

815are:
Referenced solution https://github.com/microsoft/fluentui/pull/530 was removed/changed within - https://github.com/microsoft/fluentui/pull/1567

Also if we check dialog example in fluent then there no open animation - https://developer.microsoft.com/en-us/fluentui#/controls/web/dialog

815are:
@sanat-thukral 

I created PR with fix - https://github.com/SAP/open-ux-tools/pull/854


https://user-images.githubusercontent.com/90789422/209174658-0bddb86c-b9a7-4b35-9bda-7f40fb15e350.mp4

815are:
@sanat-thukral 

Fix is resolved with version - `1.1.1`.
Please try to use latest version.

sanat-thukral:
@815are Looks great, thank you soo much 😄

PR #854, opened 2022-12-22
fix(ui-components): Apply FluentUI dialog animations for dialog open

815are:
Fix for issue #853

Enabling animation for dialog open.
Property `isOpenAnimated` - default value is `true`.

There is build in logic in fluent-ui which contains transition css for `opacity` added in
https://github.com/microsoft/fluentui/pull/530
https://github.com/microsoft/fluentui/pull/1567

but such logic does not work anymore - https://developer.microsoft.com/en-us/fluentui#/controls/web/dialog
Solution is to set `opacity: 0` before open and then set to default opacity(`1`) when dialog is opened.

https://user-images.githubusercontent.com/90789422/209174658-0bddb86c-b9a7-4b35-9bda-7f40fb15e350.mp4


Jimmy-Joseph19 (approved):
Approving PR. Tested with storybook.

vadson71 (approved):

PR #857, opened 2022-12-27
fix(fe-fpm-writer): Writing into manifest.json always provides 2 spaces for tab size.

### LINKED ISSUE #856: BUG - fe-fpm-writer. Writing into `manifest.json` always provides 2 spaces for tab size.
### Description
`fpm-writer` methods like `generateCustomColumn`, `generateCustomSection`, etc. updates and reformats `manifest.json` with 2 spaces for tab size. 

### Steps to Reproduce
1. Use app and format `manifest.json` in way that 4 spaces are used for tab size;
2. Trigger creation of custom column or other extension like using method `generateCustomColumn`

Problem - `manifest.json` is reformatted and uses 2 spaces for tab sizing.

image

### Expected results
Provide option to set tab size.

### Actual results
Problem - `manifest.json` is reformatted and in result uses 2 spaces for tab size.

### Screenshots

OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

815are:
Fix #856

Problem:
`fs.extendJSON` is used without 4th param `space`:
```
 extendJSON(filepath: string, contents: any, replacer?: WriteJsonReplacer, space?: WriteJsonSpace): void;
```
In result default values is `2`:
image

Solution:
1. New configuration property 'tabInfo' is added, where we an pass tab sizing configuration. Currently it is used for 'manifest.json' file updates;
2. If `tabInfo` is not passed, then logic in [file.ts](https://github.com/SAP/open-ux-tools/blob/fix/856/manifestTabSizing/packages/fe-fpm-writer/src/common/file.ts#L47) tries to calculate current file's tab sizing info. It finds first line with spaces or `\t` and detect if it should use white spaces or `\t` and calculates single tab step.

Additional changes:
1. [list.ts](https://github.com/SAP/open-ux-tools/blob/fix/856/manifestTabSizing/packages/fe-fpm-writer/src/page/list.ts), [object.ts](https://github.com/SAP/open-ux-tools/blob/fix/856/manifestTabSizing/packages/fe-fpm-writer/src/page/object.ts) 
small refactoring to avoid reported sonar duplication issues

### DIFF
--- a/.changeset/fifty-schools-hug.md
+++ b/.changeset/fifty-schools-hug.md
@@ -2,4 +2,4 @@
 '@sap-ux/fe-fpm-writer': patch
 ---
 
-Fix: writing into manifest.json always provides 2 spaces for tab size. Addition property 'tabInfo' can be passed in config for generatION functions or tab sizing will calculated before saving
+Fix: writing into manifest.json always provides 2 spaces for tab size. Additional property 'tabInfo' can be passed in config for generation functions or tab sizing will be calculated before saving
--- a/.changeset/fifty-schools-hug.md
+++ b/.changeset/fifty-schools-hug.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fe-fpm-writer': patch
+---
+
+Fix: writing into manifest.json always provides 2 spaces for tab size. Addition property 'tabInfo' can be passed in config for generatION functions or tab sizing will calculated before saving
--- a/packages/fe-fpm-writer/src/common/file.ts
+++ b/packages/fe-fpm-writer/src/common/file.ts
@@ -17,8 +17,8 @@ interface ExtendJsonParams {
 /**
  * Method returns tab info for passed line.
  *
- * @param {string} line Line with tab spacing.
- * @returns {TabInfo | undefined} Tab size information.
+ * @param line - line with tab spacing
+ * @returns tab size information
  */
 function getLineTabInfo(line: string): TabInfo | undefined {
     let tabSize: TabInfo | undefined;
@@ -40,8 +40,8 @@ function getLineTabInfo(line: string): TabInfo | undefined {
 /**
  * Method calculates tab space info for passed file content.
  *
- * @param {string} content File content.
- * @returns {TabInfo | undefined} Tab size information.
+ * @param content - file content.
+ * @returns tab size information.
  */
 export function detectTabSpacing(content: string): TabInfo | undefined {
     let tabSize: TabInfo | undefined;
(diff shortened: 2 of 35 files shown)


.changeset/fifty-schools-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/fe-fpm-writer': patch
+---
+
+Fix: writing into manifest.json always provides 2 spaces for tab size. Addition property 'tabInfo' can be passed in config for generatION functions or tab sizing will calculated before saving
```
stefanschreck, line 5: Addition ==> Additional
generatION ==> generation
tab sizing will BE calculated

packages/fe-fpm-writer/src/common/file.ts
```diff
@@ -0,0 +1,98 @@
...
+ * @returns tab size information.
+ */
+export function getJsonSpace(fs: Editor, filePath: string, tabInfo?: TabInfo | undefined): WriteJsonSpace | undefined {
+    let space: WriteJsonSpace | undefined;
+    // Use passed tab info or read from existing content
+    if (!tabInfo) {
+        const content = fs.read(filePath);
+        tabInfo = detectTabSpacing(content);
+    }
+    if (tabInfo !== undefined) {
```
stefanschreck, line 74: `else` would be sufficient

stefanschreck (changes requested):

packages/fe-fpm-writer/src/common/file.ts
815are: @stefanschreck 

I applied some correction in code, but `else` will break logic - result of first if block can affect second if condition. Idea is:
1. 1st condition. if `tabInfo` is not passed - we calculate value for `tabInfo` based on passed file;
2. 2nd condition. It checks if `tabInfo` is defined and calculates value for `space` variable;

We can pass `tabInfo` as custom tab info configuration, but it is optional param. If it is not passed, then first 'if' calculates `tabInfo` by checking target file.

If we add `else`, then calculated value in 1st condition will never be used in second condition, but it should be:
image

I also added some additional tests with new commit to catch such case - https://github.com/SAP/open-ux-tools/pull/857/commits/b234f10d48967c9e5f84bffb7298ffa77c995c00

Please rereview

.changeset/fifty-schools-hug.md
815are: thanks, corrected in edc31bd4a0cb4a5059f18d900c6a747848d09ab4

zdravko-georgiev (approved):
Code looks good
Very good test coverage
Didn't tested manually

stefanschreck (approved):

PR #858, opened 2022-12-29
fix(fe-fpm-writer): controller-extension for TS application. Error 'Subsequent property declarations must have the same type. Property 'base' must be of type '{ getExtensionAPI():...'

### LINKED ISSUE #841: BUG - fe-fpm-writer. controller-extension for TS application. Subsequent property declarations must have the same type.  Property 'base' must be of type '{ getExtensionAPI():
### Steps to Reproduce
1. Use ts application. For example - [project2-2022-12-160721.zip](https://github.com/SAP/open-ux-tools/files/10243513/project2-2022-12-160721.zip)
2. Generate two controller extension, but with different page types - 1st `ObjectPage` and 2nd `ListReport`

In result we have two ts controllers - `OP.controller.ts` and `LR.controller.ts`

In each `controller.ts` we have `declare module 'sap/ui/core/mvc/ControllerExtension'`, but return type for `getExtensionAPI()` differs:
1. `sap/fe/templates/ObjectPage/ExtensionAPI`
2. `sap/fe/templates/ListReport/ExtensionAPI`

Such state gives typescript error:
image

Attached app for reference - 
[project2-2022-12-160721.zip](https://github.com/SAP/open-ux-tools/files/10243513/project2-2022-12-160721.zip)

### Expected results
Generation of  `OP.controller.ts` and `LR.controller.ts` gives typescript error

### Actual results
Generation of  `OP.controller.ts` and `LR.controller.ts` without typescript errors

### Screenshots
image

OS:                      
- [ ] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

815are:
Fix #841

## Description:

We are facing following error when we create both controller extensions for ListReport and ObjectPage:
image

As correction - applying proposal from @tobiasqueck - https://github.com/tobiasqueck/proposal-controller-extension/pull/2

Changes:
1. `fpm-writer`. Add `ControllerExtension.d.ts` to `ext/controller` folder if such file does not exist - should be added on first creation of controller extension;
2. `fpm-writer`. Change template of [Controller.ts](https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/templates/controller-extension/Controller.ts) to use `Class` approach;
3. `ui5-application-writer`. Increase `ui5-tooling-transpile` generated dev dependency for typescript project from `0.2.1` to `0.3.7`;
4.  `ui5-application-writer`. Update generated `.babelrc` to use `addControllerStaticPropsToExtend: true`  to support controller extension class syntax
5. Snapshot updates for `fiori-elements-writer`, `fiori-freestyle-writer`


packages/ui5-application-writer/templates/optional/typescript/package.json
```diff
@@ -6,7 +6,7 @@
     },
     "devDependencies": {
         "@sapui5/ts-types-esm": "<%- ui5.typesVersion %>",
-        "ui5-tooling-transpile": "^0.2.1",
+        "ui5-tooling-transpile": "^0.3.1",
```
tobiasqueck, line 9: There is a `0.3.7` by now which has a few minor improvements.

tobiasqueck:
I tried testing the generated app of the `integration/ts` test and it didn't work. However, it wasn't because of your changes, but because of an issue in our test setup. I added a fix to your branch with https://github.com/SAP/open-ux-tools/pull/858/commits/9409fdb3426ba3c49085986e76d743d5cd71dfcd

tobiasqueck (approved):
* changeset properly created
* code adheres to coding standards and is easy to understand
* test cases properly updated
* manual test of the generated `integration/ts` test app works

packages/fe-fpm-writer/src/controller-extension/index.ts
```diff
@@ -243,6 +243,11 @@ export function generateControllerExtension(
 
     if (controllerConfig.typescript) {
         addExtensionTypes(basePath, controllerConfig.minUI5Version, fs);
+        // Typescript - add declaration ts file for 'sap/ui/core/mvc/ControllerExtension', which allows access to `this.base.getExtensionAPI()` within Controller extension classes
+        const declarationFilePath = join(basePath, '/webapp/ext/controller/ControllerExtension.d.ts');
```
nlunets, line 247: Just wondering, do we want to put the type in ext/controller/ControllerExtension or in another place only for type extensions ? 
Shouldn't we otherwise offer this definition in a type module that we deliver. (and that we can easily tweak ?)

tobiasqueck: This is a workaround since there is no such types module. Let's have a discussion with the SAP Fiori elements architects and find out if we can release something like that.

tobiasqueck (approved):

zdravko-georgiev (approved):
Code looks good, I haven't tested manually.

Issue #833, opened 2022-12-14
TBI - Replace Enzyme with React testing library


devinea:
### Description (include screenshots)
[Enzyme](https://enzymejs.github.io/enzyme/) does not support newer versions of react greater than 16.
As a result it is necessary to use an alternative testing library.
Recommended is https://testing-library.com/docs/react-testing-library/intro
There is a migration guide here. 
https://testing-library.com/docs/react-testing-library/migrate-from-enzyme/

**Any new react tests should be written using `react-testing-library` and NOT `enzyme`.**
Existing tests will need to be written converted over time to `react-testing-library`

### Value
Optional - details on how this technical backlog item provides value to end-users or dev team eg. Performance, Quality, Maintainability. This will help determine the ranking in relation to feature development

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

devinea:
FYI @SAP/ux-tools-ui-components

tobiasqueck:
> Existing tests will need to be written converted over time to react-testing-library

@devinea how far along are with that and what is the timeline for remaining tests?

devinea:
> > Existing tests will need to be written converted over time to react-testing-library
> 
> @devinea how far along are with that and what is the timeline for remaining tests?

Not much progress on this and currently no fix timeline. We should review this.

PR #863, opened 2023-01-10
fix(uic): fixes jumping cursor in editable table input

tom-sap:
Fixes jumping cursor in UITable inputs with validation.


packages/ui-components/src/components/UITable/UITable.tsx
```diff
@@ -127,23 +129,50 @@ export class UITable extends React.Component<UITableProps, UITableState> {
      * @param prevState
      */
     componentDidUpdate(prevProps: UITableProps, prevState: UITableState): void {
-        const scrollContainer = document.querySelector('.ms-ScrollablePane--contentContainer');
-        if (scrollContainer) {
-            const left = scrollContainer.scrollLeft || 0;
-            requestAnimationFrame(() => (scrollContainer.scrollLeft = left));
-        }
+        const newState: any = {};
```
815are, line 132: maybe we can avoid usage of `any`?

815are (dismissed):
Code looks ok(small comment regarding `any` usage).
Tested with storybook - did not notice issues.

@tom-sap 
Test coverage - can we improve test coverage?

packages/ui-components/src/components/UITable/UITable.tsx
tom-sap: fixed, thanks!

tom-sap:
as for the coverage, it was already low, and I only added 4 lines physically. The rest is just moving lines out of 1 function to another function, because another smart-bot complained that it was too complex. 
This issue already took more time than necessary, and we just don't have the capacity to write all the tests required for the component, as a part of this fix.

Klaus-Keller:
> as for the coverage, it was already low, and I only added 4 lines physically. The rest is just moving lines out of 1 function to another function, because another smart-bot complained that it was too complex. This issue already took more time than necessary, and we just don't have the capacity to write all the tests required for the component, as a part of this fix.

Hi @tom-sap,

1. Coverage is measured on [new lines](https://sonarcloud.io/component_measures?metric=new_coverage&selected=SAP_open-ux-tools%3Apackages%2Fui-components%2Fsrc%2Fcomponents%2FUITable%2FUITable.tsx&view=list&pullRequest=863&id=SAP_open-ux-tools) of code only, not the whole module.
2. If you move code, e.g. to reduce cognitive complexity, it would be a great opportunity to add missing tests for the small part of the code that has moved.

tom-sap:
Hi @Klaus-Keller,

I've spent several hours trying to make the tests work for this new feature, but it looks like it's not possible. Jest & enzyme just don't have the ability to test such low-level functionality. They can fake-trigger events (by calling the event listeners) but that doesn't actually update the DOM in the same way as the user interactions would.
I can update the input value, but that's not sufficient to test the interaction.

Basically the interaction works like this:
- user starts editing the field in a column that has validation
- user moves the caret inside the input to the beginning of the value text
- user removes a character from the value which triggers the validation error
- the caret stays in the same position after component is re-rendered

Unless there's a way to access the HTML Element of the input (which I haven't found) and trigger events directly to this element - I see no way of unit-testing this.

devinea:
> I've spent several hours trying to make the tests work for this new feature, but it looks like it's not possible. Jest & enzyme just don't have the ability to test such low-level functionality. 

@tom-sap Not sure if it helps here. But might be worth checking [react-testing library](https://testing-library.com/docs/react-testing-library/intro/) as we want to move away from enzyme anyway. See https://github.com/SAP/open-ux-tools/issues/833

tom-sap:
Hi @devinea, 
The testing framework looks better but looks like it's not possible after all to test the low level user interactions due to the browser behaviour.

Events created programmatically have [isTrusted](https://developer.mozilla.org/en-US/docs/Web/API/Event/isTrusted) set to false which changes the behaviour of the event, i.e. 
dispatching `keydown` event in an input does not affect the input value (which is what we need to test the feature).

Unless there are some hidden react hacks that would allow that, I have no more ideas :-)

tom-sap:
@Klaus-Keller @devinea can you review please?

Klaus-Keller:
Hi @tom-sap, 

Could you please add some information on what this PR exactly fixes and how to reproduce the issue? I've checked out this branch, built locally and tested table with validation in storybook, but I still see jumpy cursors:



For instance when I delete the `e` in hello cursor jumps to the end or when I insert an `e` the same, cursor jumps to the end.

By the way, I think lines like e.g.:
https://sonarcloud.io/code?id=SAP_open-ux-tools&pullRequest=863&selected=SAP_open-ux-tools%3Apackages%2Fui-components%2Fsrc%2Fcomponents%2FUITable%2FUITable.tsx&line=135
https://sonarcloud.io/code?id=SAP_open-ux-tools&pullRequest=863&selected=SAP_open-ux-tools%3Apackages%2Fui-components%2Fsrc%2Fcomponents%2FUITable%2FUITable.tsx&line=153
https://sonarcloud.io/code?id=SAP_open-ux-tools&pullRequest=863&selected=SAP_open-ux-tools%3Apackages%2Fui-components%2Fsrc%2Fcomponents%2FUITable%2FUITable.tsx&line=174

could be covered by test cases without technical limitations.

tom-sap:
Hi @Klaus-Keller,

1. The description of the user interaction is in my previous comment: https://github.com/SAP/open-ux-tools/pull/863#issuecomment-1380550452

More details in the issue internal issue #21529.

2. If you were able to actually test this in Storybook, than you got further than me. For me Storybook was crashing whenever I wanted to edit the text. I tested the fix in DataEditor table and it works fine there (for me at least). I cannot say for storybook, as I cannot test this in storybook.

3. As I said before - this is a bug-fix-PR, rather than a back-filling-missing-unit-tests-TBI-PR, and something that should take 5 minutes already took (probably) 5 days. 
 
I will revert the refactoring so the only changes to the code is the lines that I've added.

packages/ui-components/src/components/UITable/UITable.tsx
```diff
@@ -126,6 +128,7 @@ export class UITable extends React.Component<UITableProps, UITableState> {
      * @param prevProps
      * @param prevState
      */
+    // eslint-disable-next-line sonarjs/cognitive-complexity
```
Klaus-Keller, line 131: We should not ignore this as the cognitive complexity exists and should be fixed.
If you do not want to address this with your fix, you could make sure that you do not increase the complexity.
See https://github.com/SAP/open-ux-tools/pull/869

packages/ui-components/src/components/UITable/UITable.tsx
```diff
@@ -620,6 +640,17 @@ export class UITable extends React.Component<UITableProps, UITableState> {
             errorMessage = column.validate(value);
         }
         if (editedCell && editedCell.errorMessage !== errorMessage) {
+            if (editedCell.rowIndex && editedCell?.column?.key) {
```
Klaus-Keller, line 643: this condition causes row 0 to be ignored
https://github.com/SAP/open-ux-tools/pull/869

Klaus-Keller (changes requested):
Hi @tom-sap,

I found the reason why the fix did not work in storybook. I've tested with row 0 and this row was ignored due this condition twice in the code
```
if ( ... editedCell.rowIndex ...)
```
which resolves to `false` for row 0.

I also left a comment regarding cognitive complexity and added a test case in a separate PR.
Please check PR https://github.com/SAP/open-ux-tools/pull/869, if you agree we can merge it  into branch `jumping-cursor-in-edi-table`.

packages/ui-components/src/components/UITable/UITable.tsx
```diff
@@ -181,6 +184,21 @@ export class UITable extends React.Component<UITableProps, UITableState> {
                 this.props.showRowNumbers
             );
         }
+
+        const editedCell = this.state.editedCell;
+        if (this.caretPosition !== -1 && editedCell?.rowIndex && editedCell.column?.key) {
```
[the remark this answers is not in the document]
tom-sap: good catch, thanks!

tom-sap:
Hi @Klaus-Keller,
The `rowIndex` condition was a good catch!
Somehow I never tested the first row 😊.

As for the test - this is an interesting approach, I must say.
It tests that the functions are called (so the coverage goes up) without actually testing the exact interaction :-)
(the issue didn't happen when the value was changed, but how it was changed)

Anyways, linters are happy, so all is fine now 😄 
Thanks for your help!

815are (approved):

oliversberzs (approved):

Klaus-Keller (approved):
- Code looks good and is easy to understand
- Test coverage looks good
- Build and tested locally with Storybook

Approved from my side.

PR #870, opened 2023-01-19
fix(ui-components): UIActionButton. Remove forced color for inner icon

815are:
Problem reported by @aislingnoone , that color of icons in UIActionButton is always with `var(--vscode-foreground)`:
image

But expected is:
image

Updated storybook with:
image

### DIFF
--- a/.changeset/smooth-ways-smoke.md
+++ b/.changeset/smooth-ways-smoke.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui-components': patch
+---
+
+UIActionButton. Remove forced color for inner icon - icon will be taken from individual icon styles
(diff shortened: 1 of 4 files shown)


815are:
@gpeuc @aislingnoone 

In previous version we forcedly set icon color in UIAction button to `var(--vscode-foreground)`(most likely was introduced on very first versions). I removed such style.

I updated storybook with extended example:
image
Could you please approve if it is a valid state?

gpeuc:
Yes this is correct.

Sometimes we should forcefully push the colour into icons, for example, when we have active state of a button, so we have to forcefully paint the icon white so that it shows OK on the blue background.

But more often than not, we should just leave icons the way they are painted originally in the SVG. 

Screenshot 2023-01-19 at 12 55 13

PR #874, opened 2023-01-27
fix(ui-components): UITextInput. Background color issue for readonly fields

815are:
Scenario is when we overwrite css with style like(maybe more cases with other style variations):
```
.ms-TextField-fieldGroup { align-items: center; }
```

then readonly field looks following:
image

Applying same read-only background for container as well

### DIFF
--- a/.changeset/loud-phones-kick.md
+++ b/.changeset/loud-phones-kick.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui-components': patch
+---
+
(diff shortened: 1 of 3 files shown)


Klaus-Keller (approved):
- Code looks good
- Did install, build, and test locally
- Test case looks good

Approved from my side.

Jimmy-Joseph19 (approved):
Code looks good. Did not test locally. Approving from my side.

Issue #873, opened 2023-01-27
TBI - Add `UITranslationInput` to 'ui-components' extended from `UITextField`


815are:
### Value
Add `UITranslationInput` to 'ui-components' extended from `UITextField`:
![image](https://user-images.githubusercontent.com/90789422/215051141-4020ae43-a781-4c66-8a0b-357579f3869f.png)

Proposed interface(for now):
```
export interface UITranslationInputProps {
    entries: I18nBundle;
    // When entry exists in passed i18n entries and user clicked on show entry button
    // ToDo - param
    onShowExistingEntry?: () => void;
    // When creation of new i18n entry is requested
    onCreateNewEntry?: () => void;
    // PascalCase or camelCase to be used for key generation
    // Default value is -> ???
    namingConvention?: TranslationKeyGenerator;
    // Loader indicator
    busy?: boolean;

    defaultPattern:
    allowedPattern: []
}
```
Interface is not final - just quick draft and could differ

### Architecture Elaboration
Topics was:
1. interface with properties ` defaultPattern`, `allowedPattern`
2. No localization/i18n support in `ui-components` - all texts passed as props with default hardcoded text like it is done in fluentui - https://github.com/microsoft/fluentui/blob/master/packages/react/src/components/DatePicker/defaults.ts#L6-L13

### Tasks
- [ ] Component
- [ ] Storybook
- [ ] Unit tests

PR #877, opened 2023-01-30
feat(ui-components)L UITranslationInput component

815are:
Issue #873

New component `UITranslationInput` - provides UI to generate i18n key for entered text in input:
image

Also storybook example with table and localstorage is added:
translationInput


Klaus-Keller (approved):
- code looks clean and easy to understand
- code coverage is good
- built and tested locally and it works fine

Approved from my side.

tom-sap (approved):

Issue #713, opened 2022-10-05
FEATURE - switch archive command to use a deny list and include dot files


cfg74:
_As a **developer**,
I want to **be able to zip/download all files of a project in BAS including configurations in dot files**,
so that I **can upload and run it in a different devspace**_

### Description

_Currently the archive command excludes node modules, git configurations and `.env` files but also excludes all dot files. In some cases vital configurations are stored in dot files, and losing them on transfer makes the project not work at the destination, e.g. the `.cdsrc.json` file in CAP projects roots_

Proposal is to include all dot files into the archive, but exclude specifically the ones mentioned above.

### Technical Design
Requires update of module `@sap-ux/environment-check` to include dot files.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a zip archive created from a project with dot file configurations
**when** I execute the "Archive Project" command
**then** the project copy is fully functional in the target

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

cfg74:
@Klaus-Keller fyi.

Klaus-Keller:
The code to archive a project is in module `@sap-ux/environment-check`. The glob that controls what is included and excluded is here:

https://github.com/SAP/open-ux-tools/blob/fdd0d0c3ef5650ba62908a4e565160cc701fb77f/packages/environment-check/src/output/zip.ts#L90-L95

The `globOptions` can be enhanced by `dot: true` to include hidden files starting with `.`, see https://www.archiverjs.com/docs/archiver#glob -> https://github.com/yqnn/node-readdir-glob#options.
However, all hidden files would be included then and it is likely, that we would archive files that contain secrets. We exclude already `.env`, but there might be more files that we are not aware of. 

**Proposal**: if we find a `.gitignore` file, read it and apply the glob rules from it to exclude folders and files from the archive. If we do not find `.gitignore`, fall back to a deny list to exclude known files, like currently `.env`.

cfg74:
@tobiasqueck @Klaus-Keller any update on this topic?

Klaus-Keller:
@cfg74 I think this is ready to be assigned, technical proposal for this feature was discussed and seems to be agreed on:

- if file `.gitignore` exists, use it to define what is excluded from the archive
- if not, use current approach. To be discussed: any changes in current approach like including dot-files?

Klaus-Keller:
[...]

PR #878, opened 2023-01-31
feat(env-check): consider .gitignore when archiving project

Klaus-Keller:
### Issue
https://github.com/SAP/open-ux-tools/issues/713

### Description
This PR fixes the issue, that some hidden/dot files are excluded from the archive that should be included. The archiving strategy is now as follows:

If the root folder of the project that is to be archived contains a `.gitignore` file, this is used to decide which files and folders should be included in the archive.

f the root folder does not contain file `.gitignore` following rules apply:
- `node_modules` are excluded
- all non-hidden files are included
- files `.cdsrc.json` and  `.extconfig.json` are included

To test this feature, you can checkout branch `feat/713/env-check-dot-files`, and run following commands:
```
pnpm install
pnpm build
packages/environment-check
node -e 'require("./dist/archive/index.js").archiveProject("<ABSOLUTE_PATH_TO_PROJECT_ROOT>")'
```
where `<ABSOLUTE_PATH_TO_PROJECT_ROOT>` is the path to the root of any project. The resulting `.zip` file is stored in the parent of the project root.

This PR contains also a cleanup of the archive functionality. It has been moved from folder `output` to `archive`, as it does not provide and output format for environment check.


cianmSAP (approved):

packages/environment-check/src/archive/index.ts
```diff
@@ -0,0 +1,75 @@
...
+ * Get list of files to be added to archive. Depending on the existence of file {cwd}/.gitignore, we use either a default filter
+ * or the .gitignore content.
+ *
+ * @param cwd - working directory, usually the project root
+ * @returns - list of files to add to the archive, relative to cwd
+ */
+async function getFileList(cwd: string): Promise<string[]> {
+    const gitignorePath = join(cwd, '.gitignore');
+    const hasGitignore = existsSync(gitignorePath);
+    const globPattern = hasGitignore ? ['**'] : ['**', '.cdsrc.json', '.extconfig.json'];
```
devinea, line 63: thanks for including `.extconfig.json`

devinea (approved):
Approach and tests look good. New libs look ok.
Not sure if `.env` should always be added to the ignore in case it is missing in the .gitignore file?

packages/environment-check/src/archive/index.ts
Klaus-Keller: > Not sure if `.env` should always be added to the ignore in case it is missing in the .gitignore file?

Hi @devinea, thanks for looking into this. When no `.gitignore` is present, all dot-files are excluded, so `.env` is also excluded by default. We only include the well known dot files for SAP projects, currently `.cdsrc.json` and `.extconfig.json`.

devinea (approved):

cianmSAP (approved):

Issue #883, opened 2023-02-02
TBI - Consume App Descriptor (manifest.json) types from module @ui5/manifest


Klaus-Keller:
### Description
Due to the former lack of TypeScript types for UI5 App Descriptor (schema for `manifest.json`), we maintain our own types in module `@sap-ux/project-access`:
https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/types/webapp/manifest.ts

The types have recently been release with module https://www.npmjs.com/package/@ui5/manifest and we should consume them from there.

### Value
No double maintenance of types.

### Architecture Elaboration
N/A

### Tasks
- [ ] Use types from `@ui5/manifest` and remove it here

PR #884, opened 2023-02-02
fix(project-access): consume types from @ui5/manifest

Klaus-Keller:
Issue https://github.com/SAP/open-ux-tools/issues/883

Consume App Descriptor (manifest.json) types from [`@ui5/manifest`](https://www.npmjs.com/package/@ui5/manifest)


devinea (approved):
nice clean up.

PR #886, opened 2023-02-03
feat(21552): make UITable's combobox more generic

oliversberzs:
Changed UITable's boolean select into a more generic combobox

### DIFF
--- a/.changeset/selfish-gorillas-confess.md
+++ b/.changeset/selfish-gorillas-confess.md
@@ -1,0 +1,6 @@
+---
+"@sap-ux/ui-components": major
+"@sap-ux/types": minor
+---
+
+feat(21552): make UITable's combobox more generic (ColumnControlType.UIBooleanSelect -> ColumnControlType.UICombobox)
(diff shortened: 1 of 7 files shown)


tom-sap (approved):

.changeset/selfish-gorillas-confess.md
```diff
@@ -0,0 +1,6 @@
+---
+"@sap-ux/ui-components": major
+"@sap-ux/types": minor
```
devinea, line 3: `@sap-ux/types` is not changed here? `types/package.json`

For `@sap-ux/ui-components` Wouldn't patch / minor be more appropriate?
Is it technically a breaking change?
`UIBooleanSelect` is only used in data editor as far as I am aware.

oliversberzs: `@sap-ux/types` was changed cause both `UIColumn` and `ColumnControlType` was changed, added a optional field.

Yes, it is only used in the data editor, but shouldn't that then still be marked as a breaking change, cause previous api has been removed and this is an open source project?

oliversberzs: @Klaus-Keller @devinea removed the change marker from `@sap-ux/types` and made the other changes a minor bump by adding support for both control types

devinea (approved):
Thanks for the updates @oliversberzs

tom-sap (approved):

PR #894, opened 2023-02-10
fix(env-check): add optional parameter 'targetFolder' for the archiveProject method

### LINKED ISSUE #893: BUG - project archive is created in the wrong folder
### Description
For the case when the project root is opened as workspace in SAP Business Application, then project archive is created in the parent folder of the project. As a result the archive is not visible in the workspace and a wrong file is downloaded.

### Steps to Reproduce
Steps to reproduce the behaviour:
1. Open/generate project in BAS directly without enclosing folder (e.g. not projects)
1. Open command palette and run Fiori: Archive Project
1. Choose project, wait for successful information message
1. Click on Download.

### Expected results
Archive is created and it is visible in the workspace, and it can be downloaded

### Actual results
Archive is not visible and wrong file is downloaded

### Screenshots

OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

zdravko-georgiev:
Fixes #893


packages/environment-check/src/archive/index.ts
```diff
@@ -35,14 +44,14 @@ export async function archiveProject(
                     .replace(':', '')
                     .substring(0, 14)}.zip`;
             }
-            const targetPath = join(dirname(projectRoot), targetName);
-            const writeStream = createWriteStream(targetPath);
+            const targetFolder = targetPath ? join(targetPath, targetName) : join(dirname(projectRoot), targetName);
```
Klaus-Keller, line 47: @zdravko-georgiev, the content of the changes looks good, but I'm a bit confused about the naming of variables:
- `targetPath` sounds like the complete path with filename
- `targetFolder` sounds like the folder the archive should be stored in

as far as I can see, the names are used the opposite here

packages/environment-check/src/archive/index.ts
```diff
@@ -6,16 +6,25 @@ import ignore from 'ignore';
 import { t } from '../i18n';
 import { byteNumberToSizeString } from '../formatter';
 
+interface ArchiveProjectOptions {
+    projectRoot: string;
+    targetPath?: string;
```
Klaus-Keller, line 11: shouldn't it be the folder or dir?
```suggestion
    targetFolder?: string;
```

packages/environment-check/src/archive/index.ts
zdravko-georgiev: Thanks Klaus, done with https://github.com/SAP/open-ux-tools/pull/894/commits/b94dddd6cba9aafa1b686c0a6539604a0b609ba0

Klaus-Keller (approved):
Thanks @zdravko-georgiev,
- code changes look good and are easy to understand
- test coverage is excellent
- did just a visual review and did not build and test the code


Approved from my side.

zdravko-georgiev:
Thanks guys!

Issue #892, opened 2023-02-09
BUG - Missing support for multiple yaml documents in @sap-ux/yaml


Klaus-Keller:
### Description
Module `@sap-ux/yaml` does not support [multiple documents](https://www.yaml.info/learn/document.html) in yaml files. This leads to follow up issues for consumers, see detailed issue report for UI5 Language Assistant here: https://github.com/SAP/ui5-language-assistant/issues/555.

### Steps to Reproduce
1. Download and extract sample project: [fe_lrop_v4_odata_none-2023-02-092233.zip](https://github.com/SAP/open-ux-tools/files/10702093/fe_lrop_v4_odata_none-2023-02-092233.zip)
2. Open `packages/sap.demo.ui5typescripttutorialapp/src/view/App.view.xml`, this should start UI5 Language Assistant

### Expected results
UI5 Language Assistant is correct initialized

### Actual results
UI5 Language Assistant crashes:
```
YAMLError: Error parsing YAML document
Source contains multiple documents; please use YAML.parseAllDocuments()
    at new YamlDocument (/Users/d045154/.vscode-insiders/extensions/saposs.vscode-ui5-language-assistant-4.0.3/node_modules/@ui5-language-assistant/language-server/dist/server.js:7099:19)
    at YamlDocument.<anonymous> (/Users/d045154/.vscode-insiders/extensions/saposs.vscode-ui5-language-assistant-4.0.3/node_modules/@ui5-language-assistant/language-server/dist/server.js:7112:20)
    at Generator.next (<anonymous>)
    at /Users/d045154/.vscode-insiders/extensions/saposs.vscode-ui5-language-assistant-4.0.3/node_modules/@ui5-language-assistant/language-server/dist/server.js:7071:71
    at new Promise (<anonymous>)
    at __webpack_modules__.17547.__awaiter (/Users/d045154/.vscode-insiders/extensions/saposs.vscode-ui5-language-assistant-4.0.3/node_modules/@ui5-language-assistant/language-server/dist/server.js:7067:12)
    at YamlDocument.newInstance (/Users/d045154/.vscode-insiders/extensions/saposs.vscode-ui5-language-assistant-4.0.3/node_modules/@ui5-language-assistant/language-server/dist/server.js:7111:16)
[...]

PR #906, opened 2023-02-14
fix(yaml): do not throw errors for multi doc yaml

Klaus-Keller:
Issue https://github.com/SAP/open-ux-tools/issues/892

## Description
This PR fixes errors thrown in module `@sap-ux/yaml` when parsing multi document yaml strings. While it prepares for working with multiple documents, it does not provide full multi document capabilities. Full support of multi document yaml requires more technical evaluation and planing.

## Background
When running 
https://github.com/SAP/open-ux-tools/blob/41e234c0ba70425b42139a42218ece5a3562c032/packages/yaml/src/yaml-document.ts#L41-L47

with a `serializedYaml` containing [mulitple documents](https://www.yaml.info/learn/document.html), `this.document.error` contains at least one error with `code: 'MULTIPLE_DOCS'`. However, the document would still be parsed. Thus another possible fix for this would be to add an error filter to the constructor, like
`this.document.errors = this.document.errors?.filter((error) => error.code !== 'MULTIPLE_DOCS');`

Resulting constructor looks then like this:

```typescript
private constructor(serializedYaml: string) {
    this.document = yaml.parseDocument(serializedYaml);
    this.document.errors = this.document.errors?.filter((error) => error.code !== 'MULTIPLE_DOCS');
    if (this.document.errors?.length > 0) {
        throw new YAMLError(
            errorTemplate.yamlParsing + '\n' + this.document.errors.map((e) => e.message).join(''),
            errorCode.yamlParsing
        );
    }
}
```

However, this solution would have a flaw: 
When providing a yaml with multiple documents and calling `doc.toString()` at some point, only the first document is serialized.

The proposed solution on this PR also prepares for full support of multiple document yamls. For instance, the index of the document (currently fixed to `0`) could be optional parameter for function calls allowing to reference a specific document in the yaml.

### DIFF
--- a/packages/yaml/src/yaml-document.ts
+++ b/packages/yaml/src/yaml-document.ts
@@ -17,7 +17,7 @@ export interface NodeComment<T> {
  * @class YamlDocument
  */
 export class YamlDocument {
-    private document: Document;
+    private documents: Document[];
 
     /**
      * Returns a new instance of YamlDocument.
@@ -38,10 +38,16 @@ export class YamlDocument {
      * @memberof YamlDocument
      */
     private constructor(serializedYaml: string) {
-        this.document = yaml.parseDocument(serializedYaml);
-        if (this.document.errors?.length > 0) {
+        this.documents = yaml.parseAllDocuments(serializedYaml);
+        if (this.documents.length === 0) {
+            // this is done for backward compatibility when only single document yaml was supported.
+            // yaml.parseDocument('') creates a new document
+            // yaml.parseAll('') does not create a new document
+            this.documents.push(yaml.parseDocument(serializedYaml));
+        }
+        if (this.documents[0].errors?.length > 0) {
             throw new YAMLError(
-                errorTemplate.yamlParsing + '\n' + this.document.errors.map((e) => e.message).join(''),
+                errorTemplate.yamlParsing + '\n' + this.documents[0].errors.map((e) => e.message).join(''),
                 errorCode.yamlParsing
             );
         }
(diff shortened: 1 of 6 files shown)


packages/yaml/src/yaml-document.ts
```diff
@@ -38,10 +38,16 @@ export class YamlDocument {
...
-        this.document = yaml.parseDocument(serializedYaml);
-        if (this.document.errors?.length > 0) {
+        this.documents = yaml.parseAllDocuments(serializedYaml);
+        if (this.documents.length === 0) {
+            // this is done for backward compatibility when only single document yaml was supported.
+            // yaml.parseDocument('') creates a new document
+            // yaml.parseAll('') does not create a new document
+            this.documents.push(yaml.parseDocument(serializedYaml));
+        }
+        if (this.documents[0].errors?.length > 0) {
```
ullasholla, line 48: Consider errors in the other documents as well?

ullasholla (changes requested):
Thanks @Klaus-Keller, the code looks good. The test coverage is very good. Had one question about error handling

packages/yaml/src/yaml-document.ts
Klaus-Keller: added it with https://github.com/SAP/open-ux-tools/pull/906/commits/ce25fc781ab51ad569d1dfd679865a49d4d199c3

ullasholla (approved):

tobiasqueck (approved):
* functionality is as expected
* tests properly extended for the new case

PR #907, opened 2023-02-14
feat: add silent mode

voicis:
In some cases we need to perform multiple `findProjectRoot` to correctly identify project root (CDS project with multiple apps containing `package.json`) and having to wrap each call in a try-catch block makes it more difficult to use than it needs to be.

I've added another parameter `silent` which allows to suppress thrown errors. By default it is `false` and would not affect current users.

### DIFF
--- a/.changeset/thirty-walls-repair.md
+++ b/.changeset/thirty-walls-repair.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': minor
+---
+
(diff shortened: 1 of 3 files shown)


devinea (approved):
New option looks fine and covered by tests. 
Thanks.
@voicis Can you add a change set? `pnpm cset`

devinea (approved):
Thanks

zdravko-georgiev (approved):
looks good

Issue #887, opened 2023-02-08
FEATURE - Support query available package names using ADT service


slin-sap:
As a consumer of axios-extension library,
I want to support ADT service that queries list of available package names,
so that I can call this API to show list of package names in Fiori deployment configuration tool.

### Technical Design
Same implementation approach as other ADT services that have been implemented in axios-extensions.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a ABAP onprem system with ADT service enabled
**when** I call listPackages API
**then** A list of package names are returned by ABAP ADT service

### Tasks
- [x] Implementation
- [x] Unit Tests added
- [x] odata-cli example is updated with test case to load list of packages.

tobiasqueck:
@slin-sap this seems to be done, so I am closing it, if that is wrong, please reopen.

PR #910, opened 2023-02-14
feat: adt list package

slin-sap:
#887


packages/axios-extension/src/abap/types/adt-types.ts
```diff
@@ -95,3 +95,9 @@ export type AdtTransportStatus = 'S' | 'E';
 // the <DLVUNIT/> element contain text that indicate it is local package.
 // No transport number required for deploying to local pacakge.
 export const LocalPackageText = ['LOCAL_PACKAGE', 'LOCAL'];
+
+export interface PackageInfo {
```
ullasholla, line 99: JSDoc for this? Maybe include the purpose of this interface?

packages/axios-extension/src/abap/types/adt-types.ts
```diff
@@ -95,3 +95,9 @@ export type AdtTransportStatus = 'S' | 'E';
...
+    uri: string;
+    type: 'DEVC/K';
```
ullasholla, line 101: Why a hardcode value for `type`? Maybe a constant would do? What's `DEVC/K`?

packages/axios-extension/test/abap/abap-adt-service.test.ts
```diff
@@ -120,7 +122,7 @@ describe('Create new transport number', () => {
             .replyWithFile(200, join(__dirname, 'mockResponses/transportRequest-1.txt'));
         const transportRequestService = await provider.getAdtService<TransportRequestService>(TransportRequestService);
         expect(
-            await transportRequestService.createTransportRequest({
+            await transportRequestService?.createTransportRequest({
```
ullasholla, line 125: Why use the optional chaining operator? When could `transportRequestService` be falsy? Is it an error in that case?

packages/axios-extension/src/abap/adt-catalog/services/list-package-service.ts
```diff
@@ -0,0 +1,81 @@
...
+     * @returns A list of package names that can be used for deploy
+     */
+    public async listPackages(phrase: string): Promise<string[]> {
+        const config = {
+            headers: {
+                Accept: 'application/xml'
+            },
+            params: {
+                operation: 'quickSearch',
+                query: `${phrase}*`,
```
ullasholla, line 40: A few comments here:
1. You may need to ensure `phrase` is not falsy. If it's `undefined` the query will be `undefined*`. Not what you expect.
2. How are the results sorted?
3. As the max results is being limited to 50, what is the usage pattern going to look like? Will the user keep entering text until the search results are narrowed down? That'll be multiple calls obviously.
4. Does the input case matter for search?

packages/axios-extension/src/abap/adt-catalog/services/list-package-service.ts
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * ListPackageService implements ADT requests for fetching a list of available package names
+ * from ABAP backend system.
+ */
+export class ListPackageService extends AdtService {
+    /**
+     * @see AdtService.getAdtCatagory()
+     */
+    private static AdtCategory = {
```
ullasholla, line 13: Start name with lowercase? It's not a type

packages/axios-extension/src/abap/types/adt-types.ts
ullasholla: Where is this interface used?

slin-sap: Only used internally when processing response data from the ADT service. Updated code to avoid exposing this interface to axios-extension consumer.

packages/axios-extension/src/abap/types/adt-types.ts
slin-sap: Used string type instead.

packages/axios-extension/test/abap/abap-adt-service.test.ts
slin-sap: The assertion in all the 4 test cases expect the output to be array or at least empty array. If transportRequestService is null, the actual output with '?' would be undefined, and thus fail the expect assertion.

packages/axios-extension/src/abap/adt-catalog/services/list-package-service.ts
slin-sap: - Make maxResults configurable input parameter rather than fixing it to 50 records.
- added default value for phrase (empty string) and maxResults.
- Sorting will be done by the consumer of the API.

packages/axios-extension/src/abap/adt-catalog/services/list-package-service.ts
```diff
@@ -0,0 +1,81 @@
...
+
+        const response = await this.get('', config);
+        return this.parsePackageListResponse(response.data);
+    }
+
+    /**
+     * Parse an XML document for ATO (Adaptation Transport Organizer) settings.
+     *
+     * @param xml xml document containing ATO settings
+     * @returns parsed ATO settings
```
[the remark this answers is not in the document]
slin-sap: Fixed jsdoc

packages/axios-extension/src/abap/adt-catalog/services/list-package-service.ts
slin-sap: I forgot there is already a static getter method for accessing adtCategory and the default impl in the parent class throws error if it is not implemented. I will just rename to small case and not apply the bigger change we discussed.

packages/axios-extension/src/abap/adt-catalog/services/list-package-service.ts
```diff
@@ -0,0 +1,81 @@
...
+        };
+
+        const response = await this.get('', config);
+        return this.parsePackageListResponse(response.data);
+    }
+
+    /**
+     * Parse an XML document for ATO (Adaptation Transport Organizer) settings.
+     *
+     * @param xml xml document containing ATO settings
```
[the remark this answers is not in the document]
slin-sap: Fixed the jsdoc

packages/axios-extension/src/abap/adt-catalog/services/list-package-service.ts
```diff
@@ -0,0 +1,106 @@
...
+            attributeNamePrefix: '',
+            ignoreAttributes: false,
+            ignoreNameSpace: true,
+            parseAttributeValue: true
+        };
+        const obj = XmlParser.getTraversalObj(xml, options);
+        const parsed = XmlParser.convertToJson(obj, options);
+
+        let packageArray: PackageInfo[] = [];
+        if (parsed?.objectReferences?.objectReference) {
```
slin-sap, line 97: I am not able to write test case to cover the branch that `parsed` is undefined. This is because XmlParser.validate has been used for validation. Undefined value is not possible here. I leave this ? as an extra harmless safe guard.

ullasholla (approved):
Thanks @slin-sap.

PR looks good:
* code is clean and well-structured
* test coverage is very good

kranthie-sap:
Merging using admin privileges due to an issue with the CLA signing app.

PR #912, opened 2023-02-15
fix(project-access): do not throw error in silent mode

voicis:
Follow up to #907

The `silent` parameter was not correctly passed in recursion call.


devinea (dismissed):
changes look good and covered by test.

zdravko-georgiev (approved):

tobiasqueck (approved):
Small, straight forward change that is properly covered with a test.

zdravko-georgiev (approved):

PR #913, opened 2023-02-15
feat: introduce app-config-writer with smartlinks-config

johannes-kolbe:
#916 

### Description
Introduce app-config-writer to support smart link configuration in Fiori apps
- Adds SmartLinks configuration to local project 
  - Find smartlinks configuration target from `ui5-deploy.yaml/ui5.yaml`
  - Add pre-configured targets for the included smart links

### Changes
- `@sap-ux/project-access`: introduce `readUi5Yaml` to get contents of Ui5 yaml file
- `@sap-ux/ui5-config`: export and extend types, add `FioriToolsServeStaticConfig` middleware type
- `@sap-ux/add-config-writer`: introduce app-config writer to add smartlinks config
- `@sap-ux/create`: add `add smartlinks-config` command

### How to test
In test project, install:
- [sap-ux-app-config-writer-0.0.1.tgz](https://github.com/SAP/open-ux-tools/files/11050667/sap-ux-app-config-writer-0.0.1.tgz)
- [sap-ux-create-0.0.12.tgz](https://github.com/SAP/open-ux-tools/files/10981314/sap-ux-create-0.0.12.tgz)
- [sap-ux-project-access-1.2.0.tgz](https://github.com/SAP/open-ux-tools/files/10981312/sap-ux-project-access-1.2.0.tgz)

Then you should be able to run `npx @sap-ux/create add smartlinks-config`


packages/app-config-writer/src/prompt/smartlinks-config.ts
```diff
@@ -0,0 +1,145 @@
...
+                name: 'authNeeded',
+                message: `${cyan(t('info.authNeeded'))}`
+            }
+        ]);
+        if (!authNeeded) {
+            return undefined;
+        }
+    } else {
+        log?.info(yellow(t('info.credentialsRequired')));
+    }
```
zdravko-georgiev, line 62: I don't get this part. The check if authentication is needed should be independent if we are running in AppStudio or VSCode. In general authentication is then needed, when it is not provided by other means, e.g. from destination configuration or secure store.

packages/app-config-writer/src/smartlinks-config/generateSmartLinks.ts
```diff
@@ -0,0 +1,49 @@
...
+    fs?: Editor
+): Promise<Editor> {
+    if (!fs) {
+        fs = create(createStorage());
+    }
+    if (!config.service) {
+        const service = (await getServices(basePath, logger, fs))[0];
+        const credentials = await getCredentials(service, logger);
+        config.service = { ...service, credentials };
+    }
```
zdravko-georgiev, line 29: If you are at this stage, then you should already have all the data. In which case `config.service` is not present?

packages/app-config-writer/src/smartlinks-config/service.ts
```diff
@@ -0,0 +1,88 @@
...
+ * @param logger logger to report info to the user
+ * @returns response from service
+ */
+export async function sendRequest(
+    service: ServiceConfig,
+    credentials?: AxiosBasicCredentials,
+    logger?: ToolsLogger
+): Promise<SystemDetailsResponse> {
+    const serviceUrl = getSmartLinksServiceUrl(service);
+    const httpsAgent = new HttpsAgent({ rejectUnauthorized: service.ignoreCertError });
```
zdravko-georgiev, line 41: Why do you need `new HttpsAgent`? Is direct usage of `axios.get` not working?

packages/app-config-writer/src/smartlinks-config/generateSmartLinks.ts
```diff
@@ -0,0 +1,49 @@
...
+    basePath: string,
+    config: SmartLinksConfig,
+    logger?: ToolsLogger,
+    fs?: Editor
+): Promise<Editor> {
+    if (!fs) {
+        fs = create(createStorage());
+    }
+    if (!config.service) {
+        const service = (await getServices(basePath, logger, fs))[0];
```
zdravko-georgiev, line 26: It is not good idea just to take the first service. If there are multiple defined, then we need to ask the user to pick one.

zdravko-georgiev (commented):
Still reviewing, this is what I've found so far.

packages/app-config-writer/src/smartlinks-config/service.ts
```diff
@@ -0,0 +1,88 @@
...
+    const axiosConfig = { auth: credentials, httpsAgent };
+    try {
+        logger?.info(`${cyan(t('info.connectTo'))} ${serviceUrl}`);
+        const response = await axios.get(serviceUrl, axiosConfig);
```
tobiasqueck, line 45: Why don't we use the `@sap-ux/axios-extension` module here? It has been create for simplifying access to SAP services

tobiasqueck:
Could you please create or link a feature request (issue) for this PR that describes exactly what the expected behavior of this module is to be?

packages/app-config-writer/src/smartlinks-config/generateSmartLinks.ts
johannes-kolbe: This case is for the usage without prompts, if not chosen yet, the service still has to be determined.

packages/app-config-writer/src/smartlinks-config/service.ts
johannes-kolbe: I ran into authorization issues without agent, but will look into @tobiasqueck comment about using `@sap-ux/axios-extension` on how to address this

packages/app-config-writer/src/smartlinks-config/generateSmartLinks.ts
johannes-kolbe: OK, will rework this

packages/app-config-writer/src/smartlinks-config/service.ts
johannes-kolbe: Thanks, yes will look into that

johannes-kolbe:
> Could you please create or link a feature request (issue) for this PR that describes exactly what the expected behavior of this module is to be?

Thanks for pointing this out, feature request #916 created - would that work?

packages/app-config-writer/src/smartlinks-config/generateSmartLinks.ts
zdravko-georgiev: @johannes-kolbe There is no usage without prompts. In order to create the correct configuration we always need to prompt the user for the service and the credentials (if the credentials are not provided).

packages/app-config-writer/src/i18n.ts
```diff
@@ -0,0 +1,35 @@
+import type { TOptions } from 'i18next';
+import i18next from 'i18next';
+import translations from './translations/smartlinks-config.json';
+
+const NS = 'odata-service-writer';
```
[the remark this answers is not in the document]
johannes-kolbe: Done

packages/app-config-writer/src/prompt/smartlinks-config.ts
johannes-kolbe: Reworked authentication for local store and BAS

packages/app-config-writer/src/smartlinks-config/service.ts
johannes-kolbe: Now using `@sap-ux/axios-extension` for call. In case `ignoreCertError` is available in `ui5-deploy.yaml` it will be considered

packages/app-config-writer/src/smartlinks-config/generateSmartLinks.ts
johannes-kolbe: Approach reworked and as discussed, if no `ui5-deploy.yaml` is found, user input is expected. No fallback to `ui5.yaml` for now.

packages/app-config-writer/src/smartlinks-config/service.ts
johannes-kolbe: Using `@sap-ux/axios-extension` now. Very basic approach taken for first iteration.

packages/app-config-writer/README.md
```diff
@@ -0,0 +1,41 @@
+# @sap-ux/app-config-writer
+
+The `@sap-ux/app-config-writer` adds smartlinks configuration for a SAP UX project.
```
tobiasqueck, line 3: `app-config-writer` sounds like a more generic component than something that adds `smartlinks` to a project.
If the intention is to add more functionality here in the future, then please write a few words about what the app-config-writer is supposed to be in the future.

Please also add a few words about what a smartlink is and what situation, a Fiori developer would need/use them

packages/app-config-writer/README.md
```diff
@@ -0,0 +1,41 @@
...
+`npm install --save @sap-ux/app-config-writer`
+
+Yarn
+`yarn add @sap-ux/app-config-writer`
+
+Pnpm
+`pnpm add @sap-ux/app-config-writer`
+
+## Usage
+
```
tobiasqueck, line 16: Could you please also give an example in our to use it together with the `@sap-ux/create` module e.g. `npm init @sap-ux app-config add smartlinks` ?

packages/create/test/fixtures/ui5-deploy-config/ui5-deploy.yaml
```diff
@@ -0,0 +1,22 @@
...
+        excludes:
+            - /test/**
+            - /localService/**
+    customTasks:
+        - name: deploy-to-abap
+          afterTask: generateCachebusterInfo
+          configuration:
+              target:
+                  destination: ABC123
+                  url: https://abc.abap.stagingaws.hanavlab.ondemand.com
```
tobiasqueck, line 16: Please make all mock/test urls end with `.example` as domain just to make sure that we don't by accident have a real url here

packages/app-config-writer/test/unit/prompt/smartlinks-config.test.ts
```diff
@@ -0,0 +1,300 @@
...
+
+    beforeEach(() => {
+        jest.resetAllMocks();
+        isAppStudioMock = jest.spyOn(btp, 'isAppStudio');
+        getSystemCredentialsSpy = jest.spyOn(utils, 'getLocalStoredCredentials');
+        getServiceMock = jest.spyOn(store, 'getService').mockImplementation(() => serviceMock as any);
+        listDestinationsMock = jest.spyOn(btp, 'listDestinations');
+    });
+
+    describe('Prompt for target url and client: ', () => {
```
tobiasqueck, line 52: I am not sure `.matchSnapshot` is a good choice here. Please think about checking against the specific object properties that are relevant in each test. It makes it easier for a reader of the test to understand what you are verifying.

packages/app-config-writer/src/prompt/smartlinks-config.ts
```diff
@@ -0,0 +1,173 @@
...
+const getTargetPromptQuestions = (target?: DeployTarget) => {
+    const validator = (value: string, error: string): boolean | string => {
+        if (!value || !value.trim()) {
+            return t(error);
+        } else {
+            return true;
+        }
+    };
+    const questions: PromptObject[] = [];
+    if (isAppStudio()) {
```
tobiasqueck, line 26: It is also possible to connect to systems via url in AppStudio. Destination should be the most common case but it is not the only.

packages/app-config-writer/src/smartlinks-config/utils.ts
```diff
@@ -0,0 +1,202 @@
...
+    return undefined;
+}
+
+/**
+ * @description Combines the target parameters and returns the target url
+ * @param baseUrl destination or url
+ * @param client optional sap-client
+ * @returns target url to be called for smartlinks config
+ */
+function getSmartLinksTargetUrl(baseUrl: string, client?: string) {
```
tobiasqueck, line 54: That is not exactly how the `axios-extension` is supposed to be used. You shouldn't build full urls here. I will create a PR with a different way and link it here.

.changeset/shy-ravens-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/project-access': minor
+---
+
+Introduce `readUi5Yaml` to read ui5.yaml files
```
tobiasqueck, line 5: add might be a better choice than introduce here

tobiasqueck: Maybe a native speaker needs to comment here. It is not wrong to use `introduce` but it sounds off (to me)

tobiasqueck (changes requested):
- `pnpm-lock.yaml` - I don't understand why your newly added module causes so many version changes in references but I also didn't see an issue (however, it would be good if one of the other admins checks it as well)
- ui5-config - all changes make sense and adhere to our coding standards
- project-access - not sure if the helper function is required but I don't see any problems with it
- create: happy to see the usage of `__mocks__` and `fixtures`. Just one small suggestion for improvements: remove irrelevant content from the fixture files so that it is easier for the person looking at the tests to understand what is relevant for the test. 
- app-config-writer
    - same comment about the fixtures, please think about remove irrelevant content, to make it easier to understand the different test fixtures
    - some inline comments about the readme and the `axios-extension` usage
- changesets look good

devinea (approved):
Use of package versions aligned with other modules. 
lock file updates seem ok.
I haven't reviewed code changes.

packages/app-config-writer/README.md
johannes-kolbe: Thanks, added code example `npm init @sap-ux add smartlinks-config` [R25-28](https://github.com/SAP/open-ux-tools/pull/913/files#diff-e422e89a638c2b8978be2545eace180c17a2a35a0659610979f9e840b5a1dd50R25-R28)

packages/app-config-writer/README.md
johannes-kolbe: Thanks, added a few words to the general app-config-writer and a small description for the smartlinks configuration - would that work?

packages/app-config-writer/test/unit/prompt/smartlinks-config.test.ts
johannes-kolbe: Reworked to check for the expected properties directly.

.changeset/shy-ravens-prove.md
johannes-kolbe: Changed to `add`. Only part is `Introducing` for app-config-writer, but can be changed too, if necessary.

packages/app-config-writer/src/prompt/smartlinks-config.ts
cfg74: Hi @tobiasqueck as we are talking about a deploy "destination" and not the service source URL, I think it may be an even less common case to use blank URLs (not sure, doesn't BAS convert URLs into destinations anyway?). We could think about it in a next round, but to complete a first scope I would stick to destination support only.

tobiasqueck: Otherwise, yes, I would be ok with just supporting destination, however, I think doing it right is actually not much more effort.

packages/app-config-writer/src/smartlinks-config/generateSmartLinks.ts
```diff
@@ -0,0 +1,27 @@
+import { create as createStorage } from 'mem-fs';
+import { create } from 'mem-fs-editor';
+import type { Editor } from 'mem-fs-editor';
+import type { ToolsLogger } from '@sap-ux/logger';
+import type { TargetConfig } from '../types';
+import { writeSmartLinksConfig } from './utils';
+
+/**
+ * @description Add smartlinks configuration to a UI5 application.
```
tobiasqueck, line 9: Just noticed the `@description` tag you used. We never use it and put the main description as plain text at the beginning. Could you please to the same for your jsdoc comments.

johannes-kolbe: Removed `description` tag - initially added because of local eslint issues

johannes-kolbe: Hmm - I am running into linting issues also on the pipeline. Seems like [`jsdoc/match-description`](https://github.com/SAP/open-ux-tools/blob/main/.eslintrc#L126) throws errors (except in tests).
Should that be changed?

packages/app-config-writer/src/smartlinks-config/utils.ts
tobiasqueck: Here we go: https://github.com/SAP/open-ux-tools/pull/953. I haven't tested it but this should be doing the same as your previous code just leaving the url generation to the `axios-extension`.
(and adding support for url usage in BAS ;) )

packages/app-config-writer/src/smartlinks-config/generateSmartLinks.ts
tobiasqueck: Why would this happen just in your change and nowhere else?  Please check the linting issues closely, maybe you are not matching the description regex correctly.

johannes-kolbe: You are right, closing dot missing. Working on commit

packages/app-config-writer/test/unit/prompt/smartlinks-config.test.ts
```diff
@@ -54,17 +54,29 @@ describe('Test function getSmartLinksTargetFromPrompt', () => {
...
+            expect(urlPrompt.message).toContain('url');
+            expect(urlPrompt.name).toEqual('url');
+            expect(urlPrompt.type).toEqual('text');
+            expect(urlPrompt.validate).toBeDefined();
+            const clientPrompt = promptMock.mock.calls[0][0][1];
+            expect(clientPrompt.initial).not.toBeDefined();
+            expect(clientPrompt.message).toEqual('SAP client ');
+            expect(clientPrompt.name).toEqual('client');
+            expect(clientPrompt.type).toEqual('text');
+            expect(clientPrompt.format).toBeDefined();
```
tobiasqueck, line 68: Thanks.  Follow up suggestion to get it more concise (but just a suggestion, no need to change it)
```
            const [urlPrompt, clientPrompt] = promptMock.mock.calls[0][0];
            expect(urlPrompt.initial).not.toBeDefined();
            expect(urlPrompt.message).toContain('url');
            expect(urlPrompt.validate).toBeDefined();
            expect(urlPrompt).toMatchObject({
                name: 'url',
                type: 'text'
            });
            expect(clientPrompt.initial).not.toBeDefined();
            expect(clientPrompt).toMatchObject({
                message: 'SAP client ',
                name: 'client',
                type: 'text'
            });
            expect(clientPrompt.format).toBeDefined();
```

packages/app-config-writer/README.md
```diff
@@ -1,6 +1,12 @@
 # @sap-ux/app-config-writer
 
-The `@sap-ux/app-config-writer` adds smartlinks configuration for a SAP UX project. 
+The `@sap-ux/app-config-writer` adds or updates configurations to a SAP UX project.
```
tobiasqueck, line 3: SAP Fiori tools project is better aligned with branding ;)

packages/app-config-writer/README.md
```diff
@@ -1,6 +1,12 @@
...
+
+## Features
+
+### Smart Links configuration
+Smart Link controls can be defined in an application to provide navigation links to other Fiori Applications in a standardized way.
```
tobiasqueck, line 8: SAP Fiori applications 
(SAP added and applications lower case)

packages/app-config-writer/README.md
```diff
@@ -1,6 +1,12 @@
...
+To deliver an application with pre-configured targets for the included smart links, it is necessary to add a smartlinks configuration to the project.
```
tobiasqueck, line 9: This makes it sound that it is required for the productive use of the app, however, based on the changes, I assumed, it is only required for local testing. Please clarify.

packages/app-config-writer/README.md
```diff
@@ -0,0 +1 @@
+# @sap-ux/app-config-config-writer
```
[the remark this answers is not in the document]
tobiasqueck: Resolving it for @zdravko-georgiev

packages/app-config-writer/src/smartlinks-config/service.ts
tobiasqueck: Modules is used well.

packages/app-config-writer/src/prompt/smartlinks-config.ts
tobiasqueck: Resolving this conversation since it is covered implictely with the other discussion about the usage of the `axios-extension`

packages/app-config-writer/src/smartlinks-config/utils.ts
johannes-kolbe: Hi @tobiasqueck thanks for the axios-extension suggestion. Locally this works great, but on BAS I run into 404 providing either url or destination. Is there some approach I am missing?

packages/app-config-writer/test/unit/prompt/smartlinks-config.test.ts
johannes-kolbe: Thanks, since there was a bit of overhaul for the tests, the suggestions were included too

packages/app-config-writer/README.md
johannes-kolbe: Taken over

packages/app-config-writer/README.md
johannes-kolbe: Thanks, clarified and updated the readme with the changes

packages/app-config-writer/README.md
johannes-kolbe: Thanks for the suggestion, changed

packages/app-config-writer/src/smartlinks-config/utils.ts
tobiasqueck: Please have a look at https://github.com/SAP/open-ux-tools/pull/913/commits/cf42b02c84071400ad8c55d3ca7f8d95f09291be. It should fix it. The actual issue is in the `axios-extension` but that needs to be addressed separately, so here is a local fix for you.

tobiasqueck:
@johannes-kolbe as discussed on the phone, please have a look at https://github.com/SAP/open-ux-tools/pull/913/commits/cfe7418284913ff0ace6495e44c7f8793762240c. I have modified your mocking so that the http calls are mocked with `nock` but the `axios-extension` is real. This makes it easier to find issues like the one with BAS when testing.
Not required to be changed but I think you don't need to test the logging that meticulously.

tobiasqueck (approved):
All my review comments have been addressed.
I am not fully happy that we generate the file into the project root into the `appconfig` folder but after discussion with @cfg74, we agreed that we address it later.

Klaus-Keller (approved):
- Focused my review more on `@sap-ux/create` and the calls to `@sap-ux/app-config-writer`, but also checked briefly other parts
- Build and tested locally, works
- Test coverage is excellent

Approved from my side

packages/app-config-writer/src/smartlinks-config/ui5-yaml.ts
```diff
@@ -0,0 +1,90 @@
...
+    for (const ui5Yaml of ui5Yamls) {
+        let ui5YamlConfig: UI5Config;
+        try {
+            ui5YamlConfig = await readUi5Yaml(basePath, ui5Yaml);
+        } catch (error) {
+            logger?.debug(`File ${ui5Yaml} not existing`);
+            continue;
+        }
+        if (!ui5YamlConfig.findCustomMiddleware<FioriToolsProxyConfig>(DeployConfig.FioriToolsProxy)) {
+            ui5YamlConfig.addFioriToolsProxydMiddleware({ ui5: {} });
```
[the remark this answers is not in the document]
johannes-kolbe: @zdravko-georgiev There was a dependency for fioritoolsproxy in servestatic, so in case there is no fioritoolsproxy we add it - or is that not expected?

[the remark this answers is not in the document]
zdravko-georgiev: @johannes-kolbe This is not correct. The dependency to the `fiori-tools-proxy` is only for the case when reuse libs needs to be loaded from the local workspace. In this case we need to ensure that the `fiori-tools-servestatic` is called before the `fiori-tools-proxy`. For the case where there is no `fiori-tools-proxy`, then we can just add:
```
  - name: fiori-tools-servestatic
    afterMiddleware: compression
```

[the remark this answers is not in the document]
johannes-kolbe: Ok, thanks for the clarification! Updated PR with handling of non-existing `fiori-tools-proxy`

packages/app-config-writer/README.md
johannes-kolbe: Hi @tobiasqueck, would the description work like this? Was the only open point still to be resolved.

PR #914, opened 2023-02-15
Various fixes in UI5, FF and FE writers

devinea:
changes for https://github.com/SAP/open-ux-tools/issues/915 
- Align FF ListDetail template between JS and TS (impacts unit tests) 
- Add `projectChecks` called in FE and FF tests when env `UX_DEBUG_FULL=true`
    - runs `npm run install` , `npm run lint` or `npm run ts-typecheck` as appropriate to speed up manual regression testing
- Add `root: true`   to all template `.eslintrc` for consistency (and allow tests above to run with inheriting eslint config from above.
- "fix" ui5 version used in TS FE V2 tests to use a version that contains the types for FE V2.
- - fix TYPES_VERSION_PREVIOUS from `1.71.18` to `1.71.15` as `1.71.18` does not exist in https://www.npmjs.com/package/@sapui5/ts-types?activeTab=versions

### DIFF
--- a/packages/fiori-elements-writer/test/common.ts
+++ b/packages/fiori-elements-writer/test/common.ts
@@ -5,17 +5,21 @@ import { create as createStore } from 'mem-fs';
 import { create } from 'mem-fs-editor';
 import { join } from 'path';
 import type { FEOPSettings, FioriElementsApp, LROPSettings, WorklistSettings } from '../src/types';
+import { promisify } from 'util';
+import { exec as execCP } from 'child_process';
+const exec = promisify(execCP);
 
 export const testOutputDir = join(__dirname, 'test-output');
 
 export const debug = prepareDebug();
 
-export function prepareDebug(): { enabled: boolean; outputDir: string } {
+export function prepareDebug(): { enabled: boolean; outputDir: string; debugFull: boolean } {
     const debug = !!process.env['UX_DEBUG'];
+    const debugFull = !!process.env['UX_DEBUG_FULL'];
     if (debug) {
         console.log(testOutputDir);
     }
-    return { enabled: debug, outputDir: testOutputDir };
+    return { enabled: debug, outputDir: testOutputDir, debugFull };
 }
 
 const sampleTestStore = create(createStore());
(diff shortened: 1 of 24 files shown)


packages/fiori-elements-writer/test/common.ts
```diff
@@ -5,17 +5,21 @@ import { create as createStore } from 'mem-fs';
...
+const exec = promisify(execCP);
 
 export const testOutputDir = join(__dirname, 'test-output');
 
 export const debug = prepareDebug();
 
-export function prepareDebug(): { enabled: boolean; outputDir: string } {
+export function prepareDebug(): { enabled: boolean; outputDir: string; debugFull: boolean } {
     const debug = !!process.env['UX_DEBUG'];
+    const debugFull = !!process.env['UX_DEBUG_FULL'];
```
IainSAP, line 18: Can we add some documentation about the use of these 2 debug env vars, since the naming is a little obscure, please.

packages/fiori-elements-writer/test/common.ts
```diff
@@ -115,3 +119,39 @@ export const v2TemplateSettings: LROPSettings | WorklistSettings = {
         }
     }
 };
+
+export const projectChecks = async (
+    rootPath: string,
+    config: FioriElementsApp<unknown>,
+    debugFull = false
+): Promise<void> => {
+    if (debugFull && (config.appOptions?.typescript || config.appOptions?.eslint)) {
```
IainSAP, line 128: Seems odd to have the `debugFull` condition wrapping the entire code block. Would be clearer to have this test at the point of call?

packages/fiori-elements-writer/test/feop.test.ts
```diff
@@ -56,6 +57,8 @@ describe(`Fiori Elements template: ${TEST_NAME}`, () => {
             } else {
                 resolve(true);
             }
+        }).then(async () => {
+            await projectChecks(testPath, config, debug?.debugFull);
```
IainSAP, line 61: Shouldnt wrap each call to `projectChecks` like so for clarity :
```suggestion
    if (debug.debugFull) {
            await projectChecks(testPath, config);
     }
```

packages/fiori-elements-writer/test/lrop.test.ts
```diff
@@ -161,7 +163,7 @@ describe(`Fiori Elements template: ${TEST_NAME}`, () => {
                         settings: v2TemplateSettings
                     },
                     ui5: {
-                        version: '1.77.2' // flex changes preview should be included with this version
+                        version: '1.108.0'
```
IainSAP, line 166: Why was this changed?

IainSAP (commented):
@devinea Couple minor comments. It might be time to look at splitting out the snapshot files, perhaps using the file snapshot matcher from the internal repo and making this open source also. The snapshot files are too unwieldy and difficult to review now IMO since they contain multiple test cases.

IainSAP:
@devinea Probably should be a linked backlog item for traceability?

packages/fiori-elements-writer/test/feop.test.ts
devinea: I had it like that first locally and it seemed like more duplication than to do the check inside projectChecks() once.

packages/fiori-elements-writer/test/lrop.test.ts
devinea: Because the FE V2 types are only available in `~1.108.0` and `>=1.110.0`.

devinea:
@IainSAP Yeah using the file snapshot matcher is something we should look to bring into the open source. That would be a separate TBI.

packages/fiori-elements-writer/test/feop.test.ts
IainSAP: I dont think its usual to encapsulate the entire function internals behind a passed options. Seem like a smell to me and decoupling and clarity would trump duplication here?

packages/fiori-elements-writer/test/lrop.test.ts
IainSAP: But it appears that the version chosen is to test the switching on of a feature (`flex changes`) not TS types. But maybe this is already covered in a non-TS test?

devinea: Yeah that's why I removed the flex changes comment as it is covered by the JS version.

packages/fiori-elements-writer/test/common.ts
IainSAP: Actually we already have docs for `UX_DEBUG` so maybe we can add `UX_DEBUG_FULL` also : https://github.com/SAP/open-ux-tools/blob/9d4e3bcfd4a1d45e93ba603ecf345e34896cedf8/README.md

devinea:
> Probably should be a linked backlog item for traceability?
Done. #915

packages/fiori-elements-writer/test/common.ts
devinea: Done in https://github.com/SAP/open-ux-tools/pull/914/commits/26fac5bc8aef6ebcbb2ba993fe3ccbd9431780d1

IainSAP (approved):
Thx @devinea

zdravko-georgiev (approved):
Looks good

devinea:
Thanks all

PR #917, opened 2023-02-17
Feat/687/deploy with cli

longieirl:


packages/axios-extension/src/abap/message.ts
```diff
@@ -38,6 +38,22 @@ export interface ErrorMessage {
     };
 }
 
+const logLevel = (severity: string, msg: string, log: Logger, error = false): void => {
```
tobiasqueck, line 41: Could we use function syntax `function logLevel(severity ...)` since we use it in all other functions in this file.
And then, please add a JSDod comment

packages/axios-extension/src/abap/ui5-abap-repository-service.ts
```diff
@@ -154,14 +154,13 @@ export class Ui5AbapRepositoryService extends ODataService {
             // An app can be successfully deployed after a timeout exception, no value in showing exception headers
             if (response?.headers?.['sap-message']) {
                 prettyPrintMessage({ msg: response.headers['sap-message'], log: this.log, host: frontendUrl });
+                // log url of created/updated app
+                const path = '/sap/bc/ui5_ui5' + (!bsp.name.startsWith('/') ? '/sap/' : '') + bsp.name.toLowerCase();
+                const query = this.defaults.params?.['sap-client']
+                    ? '?sap-client=' + this.defaults.params['sap-client']
+                    : '';
+                this.log.info(`App available at ${frontendUrl}${path}${query}`);
             }
```
tobiasqueck, line 157: Why do we only need to show the url if there is a `sap-message` in the header? Shouldn't we show it always when deployment is successful?

packages/axios-extension/src/abap/ui5-abap-repository-service.ts
```diff
@@ -226,13 +225,14 @@ export class Ui5AbapRepositoryService extends ODataService {
             charset: 'UTF8'
         };
         const params: { [key: string]: string | boolean } = {
-            CodePage: `'UTF8'`,
+            CodePage: "'UTF8'",
```
tobiasqueck, line 228: I don't mind the change, just curious about the reason.

packages/axios-extension/src/auth/connection.ts
```diff
@@ -22,7 +22,7 @@ export class Cookies {
      * @returns cookies object
      */
     public setCookies(response: AxiosResponse): Cookies {
-        if (response.headers?.['set-cookie']) {
+        if (response?.headers?.['set-cookie']) {
```
tobiasqueck, line 25: `response` shouldn't be optional here. The function must be called with a `response` object. If not, then there is a bug in the caller code, so I would prefer to remove the `?` here and if there are cases with `response` not being defined, then lets fix them on caller side.

packages/axios-extension/src/auth/connection.ts
```diff
@@ -167,6 +167,7 @@ export function attachConnectionHandler(provider: ServiceProvider) {
                 provider.defaults.headers.common[CSRF.RequestHeaderName] =
                     error.response.headers[CSRF.ResponseHeaderName];
             }
+            provider.cookies.setCookies(error.response);
```
tobiasqueck, line 170: Could we change 
```
if (error.response?.headers?.[CSRF.ResponseHeaderName]) {
                provider.defaults.headers.common[CSRF.RequestHeaderName] =
                    error.response.headers[CSRF.ResponseHeaderName];
}
provider.cookies.setCookies(error.response);
```
to
```
if (error.response) {
                if (error.response.headers?.[CSRF.ResponseHeaderName]) {
                    provider.defaults.headers.common[CSRF.RequestHeaderName] =
                        error.response.headers[CSRF.ResponseHeaderName];
                }
                provider.cookies.setCookies(error.response);
}
```

packages/deploy-tooling/src/cli/config.ts
```diff
@@ -16,7 +16,7 @@ export async function getDeploymentConfig(path: string): Promise<AbapDeployConfi
     const ui5Config = await UI5Config.newInstance(content);
     const config = ui5Config.findCustomTask<AbapDeployConfig>(NAME)?.configuration;
     if (!config) {
-        throw new Error('The deployment configuration is missing.');
+        throw new Error('The deployment configuration is missing. 01');
```
tobiasqueck, line 19: Is the `01` on purpose here?

packages/deploy-tooling/test/unit/cli/index.test.ts
```diff
@@ -49,13 +64,6 @@ describe('cli', () => {
             errorMock.mockClear();
         });
 
-        test('missing mandatory parameter --config', () => {
```
tobiasqueck, line 52: Could you please add a test case with a command containing all required parameters if the `--config` is not provided? 
I would like to have such a test case for documentation purposes.

packages/axios-extension/src/abap/message.ts
tobiasqueck: Fixed in https://github.com/SAP/open-ux-tools/pull/929/files

packages/axios-extension/src/abap/ui5-abap-repository-service.ts
tobiasqueck: reverted back in https://github.com/SAP/open-ux-tools/pull/929/files

packages/axios-extension/src/auth/connection.ts
tobiasqueck: Reverted back in https://github.com/SAP/open-ux-tools/pull/929/files

packages/axios-extension/src/auth/connection.ts
tobiasqueck: Thanks for fixing in https://github.com/SAP/open-ux-tools/pull/929/files

packages/deploy-tooling/test/unit/cli/index.test.ts
tobiasqueck: Thanks for adding to https://github.com/SAP/open-ux-tools/pull/929/files

Issue #941, opened 2023-03-07
TBI - Provide detection of app using TypeScript


Klaus-Keller:
### Description (include screenshots)
With support for Fiori apps using TypeScript there is also a need to detect such apps. This technical backlog item aims to define and provide a central functionality for other modules to detect TypeScript applications.

### Value
One central functionality to avoid code duplication.

### Architecture Elaboration
Good candidate for a module to provide this functionality is [`@sap-ux/project-access`](https://github.com/SAP/open-ux-tools/tree/main/packages/project-access)

Proposal to detect TypeScript app:
- presence of  `tsconfig.json`
- presence of `*.ts` files

Klaus-Keller:
Done with https://github.com/SAP/open-ux-tools/pull/942

PR #942, opened 2023-03-08
feat(project-access: detect app programming language

Klaus-Keller:
### Issue
https://github.com/SAP/open-ux-tools/issues/941

Enhance `@sap-ux/project-access` by functionality to detect the used programming language of an application:
`JavaScript` | `TypeScript`, or `''` if not detected.

While adding support to detect language, I've also did following improvements and enhancements:

**Finding files**
Previous implementation of `findFiles()` allowed to search for file names only. To check for files with extension, e.g `*.ts`, I had to enhance the search. I've extracted the search logic into internal function `findBy()` and enhanced it to allow also search for file extensions. Existing `findFiles()` calls now `findBy()` with argument `by: 'name'`. 
Added new function `findFilesByExtension()` which allows to search for files with a specific file extension. It returns an array of file paths matching the queried extension. Internally, it calls also `findBy()` with argument `by: 'extension'`
While doing this I also had to adjust the merging of files from mem-fs into the search results, as the previous implementation worked with folders only (there can only be one file with a specific filename), but for search by extensions there can be multiple hits in one folder. See logic around internal function `getMemFsChanges()`.

**Detect used programming language**
Added new
```typescript
export async function getAppLanguage(appRoot: string, memFs?: Editor): Promise<'JavaScript' | 'TypeScript' | ''>
``` 
that detects the used programming language in application located at `appRoot`. Same as for other functions in project-access, it allows to optionally pass a mem-fs editor instance, which will be considered in case it is provided. 

**Detection of custom webapp path**
While looking into this feature I've recognized, that function `getWebappPath` did not allow to pass mem-fs editor instance, so I've extended it to also support mem-fs.

### DIFF
--- a/packages/project-access/src/index.ts
+++ b/packages/project-access/src/index.ts
@@ -3,7 +3,7 @@ export {
     findAllApps,
     findProjectRoot,
     getAppRootFromWebappPath,
-    getAppLanguage,
+    getAppProgrammingLanguage,
     getCapModelAndServices,
     getCapProjectType,
     getWebappPath,
--- a/packages/project-access/src/index.ts
+++ b/packages/project-access/src/index.ts
@@ -3,6 +3,7 @@ export {
     findAllApps,
     findProjectRoot,
     getAppRootFromWebappPath,
+    getAppLanguage,
     getCapModelAndServices,
     getCapProjectType,
     getWebappPath,
--- a/packages/project-access/src/project/info.ts
+++ b/packages/project-access/src/project/info.ts
@@ -2,6 +2,7 @@ import type { Editor } from 'mem-fs-editor';
 import { join } from 'path';
 import { FileName } from '../constants';
 import { fileExists, findFilesByExtension } from '../file';
+import type { AppProgrammingLanguage } from '../types';
 import { getWebappPath } from './ui5-config';
 
 /**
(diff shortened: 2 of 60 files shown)


packages/project-access/src/index.ts
```diff
@@ -3,6 +3,7 @@ export {
     findAllApps,
     findProjectRoot,
     getAppRootFromWebappPath,
+    getAppLanguage,
```
kranthie-sap, line 6:
```suggestion
    getAppProgrammingLanguage,
```

I am wondering if app language could mean natural language. I can live with it though.

packages/project-access/src/project/info.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Get the used programming language of an application.
+ *
+ * @param appRoot - root folder of the application
+ * @param [memFs] - optional mem-fs editor instance
+ * @returns - used language, JavaScript or TypeScript
+ */
+export async function getAppLanguage(appRoot: string, memFs?: Editor): Promise<'JavaScript' | 'TypeScript' | ''> {
+    const ignoreFolders = ['node_modules', '.git'];
+    let appLanguage: 'JavaScript' | 'TypeScript' | '' = '';
```
kranthie-sap, line 16: Can we use and export this as a type? I am sure we will need to use something similar in the consumed tools code.

kranthie-sap (changes requested):
The helper function and changes to file search module look good and has good tests. 👍 
I have a suggestion for the helper and a request to export the type for the programming language.

kranthie-sap (approved):

tobiasqueck (approved):
On first glance, it looked like quite a big change to solve the linked problem, however, keeping in mind that this generic `findByExtension` function is also usable for other cases in our tools e.g. to find all `xml` files, I am ok with this complex change
* code is well structured
* tests are easy to read and make clear what functionality is expected
* did not test locally 
* changeset looks good

Issue #937, opened 2023-03-06
BUG - FLP Embedded preview fails to load


johannes-kolbe:
Issue: tools-suite #23945

Adding a FLP Embedded Configuration to an application and starting preview might lead to an error 
`Page load fails with net::ERR_CONTENT_DECODING_FAILED 200` with preview not loading.

johannes-kolbe:
Hi @tobiasqueck, we are encountering an issue with setting the `accept-encoding` header to `br` ([backend-proxy-middleware](https://github.com/SAP/open-ux-tools/blob/main/packages/backend-proxy-middleware/src/base/proxy.ts#L43)) for an application. Would you know if that is still always required?

shahzeb79:
@johannes-kolbe Thanks alot for raising it. Just to clarify, its issue with our all fiori apps trying to load launchpad in embedded mode.

tobiasqueck:
> Hi @tobiasqueck, we are encountering an issue with setting the `accept-encoding` header to `br` ([backend-proxy-middleware](https://github.com/SAP/open-ux-tools/blob/main/packages/backend-proxy-middleware/src/base/proxy.ts#L43)) for an application. Would you know if that is still always required?

I don't understand the question.

johannes-kolbe:
> I don't understand the question.

@tobiasqueck Sorry about the formulation, and please feel free to redirect me. 
The question (also being investigated on tools-suite issue) is if the special setting of `accept-encoding` to `br` is still necessary in case of flp embedded preview for `Fiorilaunchpad.html`. It seems like something changed in that case.
I naively saw your name with the code here, but I assume this was related to the move from tools-suite. 🙂

tobiasqueck:
@johannes-kolbe thanks for the explanation. You are correct, I move some code there but the brain behind that one is @zdravko-georgiev

johannes-kolbe:
Fix applied in PR #944, consumption and final verification will happen on tools-suite side. Closing issue.

PR #944, opened 2023-03-10
fix(backend-proxy-middleware): Enable FLP Embedded preview to load in case of encoding failure

johannes-kolbe:
Fix for #937

### DIFF
--- a/packages/backend-proxy-middleware/src/base/proxy.ts
+++ b/packages/backend-proxy-middleware/src/base/proxy.ts
@@ -40,7 +40,7 @@ export const ProxyEventHandlers = {
      */
     onProxyReq(proxyReq: ClientRequest, _req?: IncomingMessage, _res?: ServerResponse, _options?: ServerOptions) {
         if (proxyReq.path?.includes('Fiorilaunchpad.html') && !proxyReq.headersSent) {
-            proxyReq.setHeader('accept-encoding', 'br');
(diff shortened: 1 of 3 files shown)


zdravko-georgiev (approved):
Good looks good
Simple change
100% coverage

packages/backend-proxy-middleware/src/base/proxy.ts
```diff
@@ -40,7 +40,7 @@ export const ProxyEventHandlers = {
      */
     onProxyReq(proxyReq: ClientRequest, _req?: IncomingMessage, _res?: ServerResponse, _options?: ServerOptions) {
         if (proxyReq.path?.includes('Fiorilaunchpad.html') && !proxyReq.headersSent) {
-            proxyReq.setHeader('accept-encoding', 'br');
+            proxyReq.setHeader('accept-encoding', '*');
```
tobiasqueck, line 43: The previous code had this specific `accept-encoding` header, and now you change it to accept everything. That makes me wonder, if you need it at all, i.e. would it also work if you remove the complete if block?

johannes-kolbe: @tobiasqueck good point - I had tried removing the whole block, but ran into the same encoding error `net::ERR_CONTENT_DECODING_FAILED 200` if it wasn't set at all. Seems like it has to be set.

Issue #922, opened 2023-02-22
FEATURE - FPM writer shall support contextPath for custom pages


schreckstefan:
_As a **developer**,
I want to **use contextPath instead of entitySet for the definition of Fiori pages in newer UI5 releases**,
so that I **can use SAP Graph and containments**_

### Description

Fiori Elements uses a new  way to define pages in manifest, using a contextPath instead of entitySet.
As clarified with the responsible developers, the contextPath shall also be used when defining custom pages. 
Indeed, the template of the FPM writer for custom pages does not provide this option yet.

### Technical Design
_What modules need to be created or enhanced?_ 
https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/templates/page/custom/1.94/manifest.json

_How would the functionality be exposed and how can it be consumed?_
The API shall provide both options.  contextPath is supported for UI5 versions from 1.90 onwards, thus it is sufficient to enhance the existing template for 1.94. 

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have an API call to crate a custom page AND the UI5 version is 1.94 or higher
**when** I supply the call with a contextPath
**then** the target settings of the new page in manifest gets generated with contextPath instead of entitySet

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

tobiasqueck:
@schreckstefan is this done or is there anything still open?

PR #952, opened 2023-03-16
feat: contextPath support for FPM writer

schreckstefan:
issue: https://github.com/SAP/open-ux-tools/issues/922

The PR adds the contextPath to the API for creating custom pages, list pages or object pages.


.changeset/contextPath.md
```diff
@@ -0,0 +1,9 @@
+---
+"@sap-ux/fe-fpm-writer": patch
+"@sap-ux/types": patch
+---
+
+feat: contextPath for new pages
+
+Feature enhacement: in addition to the entity (set) you now can define a contextPath that gets written instead of entitySt then into the `options > settings` of the new page.
```
hitesh-parmar, line 8:
```suggestion
Feature enhancement: in addition to the entity (set) you now can define a contextPath that gets written instead of entitySet then into the `options > settings` of the new page.
```

devinea (dismissed):
webpack upgrade and lock file changes look good.

packages/fe-fpm-writer/src/page/types.ts
```diff
@@ -45,6 +45,11 @@ export interface ListReport extends WriterConfig {
      */
     entity: string;
 
+    /**
+     * With template or UI5 version 1.94 you can alternatively specify a contextPath that shall be insterted as page option, insted of the entitySet.
```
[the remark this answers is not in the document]
schreckstefan: done

packages/fe-fpm-writer/src/page/types.ts
```diff
@@ -60,6 +65,11 @@ export interface ObjectPage extends WriterConfig {
...
```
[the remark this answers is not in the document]
schreckstefan: done

packages/fe-fpm-writer/src/page/types.ts
```diff
@@ -77,9 +87,15 @@ export interface ObjectPage extends WriterConfig {
 export interface CustomPage extends CustomElement {
     /**
      * Name of the entity used for the custom page.
+     * Take the entitySet that shall be inserted as page option.
      */
     entity: string;
 
+    /**
+     * With template or UI5 version 1.94 you can alternatively specify a contextPath that shall be insterted as page option, insted of the entitySet.
```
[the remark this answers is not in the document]
schreckstefan: done

.changeset/contextPath.md
schreckstefan: done

devinea (dismissed):
lg. 
no changeset for ui-components should be fine.

hitesh-parmar (dismissed):
LGTM ✅

tobiasqueck (changes requested):
Why does the changeset contain a patch for types but no changes for the module?

packages/ui-components/package.json
```diff
@@ -78,6 +78,6 @@
         "style-loader": "3.3.1",
         "ts-loader": "9.4.1",
         "typescript": "4.9.4",
-        "webpack": "5.74.0"
+        "webpack": "5.76.0"
```
tobiasqueck, line 81: This change does not belong into this PR. Please put this in a separate one and describe that this is fixing a security issue

packages/fe-fpm-writer/src/page/list.ts
```diff
@@ -14,7 +14,11 @@ function enhanceData(data: ListReport, manifest: Manifest): InternalListReport {
     const config: InternalListReport = { settings: {}, ...data, name: 'ListReport', ...getFclConfig(manifest) };
 
     // move settings into correct possition in the manifest
-    config.settings.entitySet = data.entity;
+    if (data.contextPath) {
+        config.settings.contextPath = data.contextPath;
+    } else {
+        config.settings.entitySet = data.entity;
```
tobiasqueck, line 20: @nlunets I need your opinion here. Does it make sense to write either into `contextPath` or `entitySet` or should we consolidate this and always convert an `entitySet` `abc` into `contextPath` `/abc`?

tobiasqueck:
> In addition I was asked by the commit husky to upgrade webpack to 5.76.0

No idea who asked but no, semantically different changes are to be separated into different commits.

schreckstefan:
> > In addition I was asked by the commit husky to upgrade webpack to 5.76.0
> 
> Even if the script "asked", no, semantically different changes are to be separated into different commits.

@tobiasqueck Then the pipeline must be enhanced. As already described, I was not able to commit my original change, the husky interrupted and did to allow to commit without the webpack upgrade.

packages/ui-components/package.json
schreckstefan: Had not chance to do so, sorry.

tobiasqueck:
> > > In addition I was asked by the commit husky to upgrade webpack to 5.76.0
> > 
> > 
> > Even if the script "asked", no, semantically different changes are to be separated into different commits.
> 
> @tobiasqueck Then the pipeline must be enhanced. As already described, I was not able to commit my original change, the husky interrupted and did to allow to commit without the webpack upgrade.

I understand that, still it needs to be a separate change/PR that you need to commit/merge first.

packages/fe-fpm-writer/src/page/list.ts
schreckstefan: @tobiasqueck Contrary to the custom page we here to not have different templates for the different UI5 versions.
Thus we _have_ to provide here two different options, for the different UI5 versions. In older UI5 versions contextPath is not supported.

tobiasqueck: I get that point. My question suggestion was that if someone with e.g. UI5 v1.112 provides the `entity` we still convert it to `contextPath`.

schreckstefan:
> > > > In addition I was asked by the commit husky to upgrade webpack to 5.76.0
> > > 
> > > 
> > > Even if the script "asked", no, semantically different changes are to be separated into different commits.
> > 
> > 
> > @tobiasqueck Then the pipeline must be enhanced. As already described, I was not able to commit my original change, the husky interrupted and did to allow to commit without the webpack upgrade.
> 
> I understand that, still it needs to be a separate change/PR that you need to commit/merge first.

@tobiasqueck 
I opened https://github.com/SAP/open-ux-tools/pull/954 now, needs to get merged first. 
I will merge from main afterwards.

packages/fe-fpm-writer/src/page/list.ts
schreckstefan: @tobiasqueck with the given template for list and object it would be a failure to implicitly assume that `entity` is filled with a contextPath, starting from a certain UI5 release.   Problem is that `entity` is also used as the new target's key, but contextPath definitions normally are very long strings /<entitySet>/ <navigationProperty1>/<navigationProperty2>/...
Thus I would keep both parameters separately. 

What we could do is: _only_  insert contextPath starting from a given UI5 version, for instance 1.94.
This would mean I 
- eliminate options>settings>entitySet from the custom page template for 1.94.
- introduce versioning also for list and object page templates and apply the same here.

nlunets: @tobiasqueck sorry missed that one.

In general it is better to go for contextPath as long as the version supports it, and yes i would even go to the point of replacing when they write entitySet...

schreckstefan:
@tobiasqueck @hitesh-parmar @devinea 
Please rwview again

packages/fe-fpm-writer/src/page/list.ts
```diff
@@ -12,9 +13,14 @@ import type { ListReport, InternalListReport } from './types';
  */
 function enhanceData(data: ListReport, manifest: Manifest): InternalListReport {
     const config: InternalListReport = { settings: {}, ...data, name: 'ListReport', ...getFclConfig(manifest) };
+    const minVersion = coerce(data.minUI5Version);
 
     // move settings into correct possition in the manifest
-    config.settings.entitySet = data.entity;
+    if (minVersion && minVersion.minor >= 94) {
```
Klaus-Keller, line 19: @schreckstefan, shouldn't we also check the major version of UI5? Just because there is no version higher than 1 does not really mean there never can be.

packages/fe-fpm-writer/src/page/object.ts
```diff
@@ -12,9 +13,14 @@ import type { ObjectPage, InternalObjectPage } from './types';
  */
 function enhanceData(data: ObjectPage, manifest: Manifest): InternalObjectPage {
     const config: InternalObjectPage = { settings: {}, ...data, name: 'ObjectPage', ...getFclConfig(manifest) };
+    const minVersion = coerce(data.minUI5Version);
 
     // move settings into correct possition in the manifest
-    config.settings.entitySet = data.entity;
+    if (minVersion && minVersion.minor >= 94) {
```
Klaus-Keller, line 19: same here, missing check for major version. Perhaps: 
```suggestion
    if (minVersion && (minVersion.major > 1 || minVersion.minor >= 94)) {
```

packages/fe-fpm-writer/src/page/list.ts
schreckstefan: Hi @Klaus-Keller you are basically right - I just had used the same logic as in other places of FPM writer. 
Then we should adjust them all.

packages/fe-fpm-writer/src/page/object.ts
schreckstefan: I have used gt(e) now

.changeset/contextPath.md
```diff
@@ -0,0 +1,9 @@
+---
+"@sap-ux/fe-fpm-writer": patch
+"@sap-ux/types": patch
```
tobiasqueck, line 3: the `@sap-ux/types` module has not been touched in your PR

.changeset/contextPath.md
```diff
@@ -0,0 +1,9 @@
...
+---
+
+feat: contextPath for new pages
```
tobiasqueck, line 6: I'd remove line six, the long text in line 8 is good by itself for the changelog.

packages/fe-fpm-writer/test/unit/column.test.ts
```diff
@@ -70,7 +70,7 @@ describe('CustomAction', () => {
             expect(fs.read(expectedFragmentPath)).toMatchSnapshot();
         });
 
-        test('version 1.86, with new handler, all properties', () => {
+        test('version 1.86.3, with new handler, all properties', () => {
```
tobiasqueck, line 73: Why does this test case need to be changed?

packages/fe-fpm-writer/test/unit/column.test.ts
```diff
@@ -79,7 +79,47 @@ describe('CustomAction', () => {
...
+            const settings = (
+                updatedManifest['sap.ui5']?.['routing']?.['targets']?.['sample']?.['options'] as Record<string, any>
+            )['settings'];
+            expect(settings.controlConfiguration).toMatchSnapshot();
+
+            expect(fs.read(expectedFragmentPath)).toMatchSnapshot();
+            expect(fs.read(expectedFragmentPath.replace('.fragment.xml', '.js'))).toMatchSnapshot();
+        });
+
+        test('version 2.0.0, with new handler, all properties', () => {
```
tobiasqueck, line 93: What is the reason for this test case? It looks like copy and paste from the last one just with a different version? Could we slim this one down to actually only verify the version relevant aspects? or consider a `test.each` if we need to run the very same test with multiple versions.

.changeset/contextPath.md
schreckstefan: done

packages/fe-fpm-writer/src/page/list.ts
```diff
@@ -12,9 +13,14 @@ import type { ListReport, InternalListReport } from './types';
  */
 function enhanceData(data: ListReport, manifest: Manifest): InternalListReport {
     const config: InternalListReport = { settings: {}, ...data, name: 'ListReport', ...getFclConfig(manifest) };
+    const minVersion = coerce(data.minUI5Version);
 
     // move settings into correct possition in the manifest
-    config.settings.entitySet = data.entity;
+    if (minVersion && gte(minVersion, '1.94.0')) {
+        config.settings.contextPath = data.contextPath ? data.contextPath : `/${data.entity}`;
```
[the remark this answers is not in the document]
schreckstefan: Thanks for the hint. Didn't know that.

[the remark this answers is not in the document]
schreckstefan: Done

packages/fe-fpm-writer/test/unit/column.test.ts
schreckstefan: Just to show that is also works with a different patch level.

packages/fe-fpm-writer/test/unit/column.test.ts
schreckstefan: Yes, I can introduce test.each here.  2.0.0 as I wanted to have at least one test case for the scenario described by Klaus before, i.e. a potential major version bump of UI5.

hitesh-parmar: Hi @schreckstefan , you could also create unique titles by using parameters https://jestjs.io/docs/api#testeachtablename-fn-timeout
See example here: https://dev.to/bgord/simplify-repetitive-jest-test-cases-with-test-each-310m

schreckstefan: Thus the question is if I should not replace test.each...

schreckstefan:
> * please check my comments about the changeset
> * additionally a minor suggestion for a code improvement (suggestion, not requirement)
> * questions about the tests
> * otherwise
>   
>   * code changes are easy to read and adhere to our coding standards
>   * new properties are properly documented
>   * changes match the description of the PR
>   * better usage of `semver` improves the code

@tobiasqueck Please review again. Consider my comment concerning test.each, please

packages/fe-fpm-writer/test/unit/column.test.ts
schreckstefan: Thanks @hitesh-parmar , I applied this to all test.each now where I could see identical texts.

schreckstefan: ok, not to all: weird, but in some cases the use of %p or &p led to falsification of the parameter data, and to wrong snapshot results. Made undone, and the snapshots look fine. In other tests this falsification does not occur. I have no explanation for this, but only used the %p or &p syntax now where it did not lead to falsifications.

schreckstefan:
> I am still not happy with this one and it took me a while understand what needs improvements.
> 
> My main problem started with the newly created tests and the amount of additional snapshots that we are now comparing for a change that is supposed be rather simple. So, when looking at the tests and how to improve them, I noticed, that there is not much improvement possible which is usually a very good indicator that something is wrong with the code structure. Now focussing on the code, the issue is that we have three different page types with three separate implementations for the very same thing. If that gets fixed and consolidated then it also becomes easier to test and the amount of tests and test snapshots for this PR will be drastically reduced. Here is my proposal https://github.com/SAP/open-ux-tools/compare/feat/922/contextPath4customPage...feat/922/contextPath4customPage-consolidation?expand=1 for consolidating the sources (I did not consolidate the tests accordingly that I would like to see being done here).
> 
> Btw, interesting is that with my change one of the custom page tests fails. First, I assumed I did something wrong but then it turned our the fragmented version coding actually has a flaw: for `contextPath` vs. `entitySet` we use the code for older versions than 1.94 as fallback if we don't know the version BUT for the usage of the `view` in a custom page, we fallback to the coding for version 1.94 and newer. That is a bad inconsistency.

Hi @tobiasqueck
Looks like it always ends like this, right?  Just a small enhancement needed, then some room for improvement detected during review, resulting in a lot of. more code changes. But Ok, lt's start improving. 
I crated https://github.com/SAP/open-ux-tools/pull/961 based on your proposal, will add some comments and further commits over there.

tobiasqueck (changes requested):
Almost all my review requests have been correctly applied, there is only open problem:
Check 
https://github.com/SAP/open-ux-tools/blob/feat/922/contextPath4customPage/packages/fe-fpm-writer/src/page/common.ts#L168
and
https://github.com/SAP/open-ux-tools/blob/feat/922/contextPath4customPage/packages/fe-fpm-writer/src/page/custom.ts#L52

If you create a customPage and you do **not** provide a `minVersion` then we will use `entitySet` but we will use the template valid for `1.94` and above. So, our handling of the absence of a minVersion is inconsistent. Please consistently handle the missing minVersion.

schreckstefan:
> Almost all my review requests have been correctly applied, there is only open problem: Check https://github.com/SAP/open-ux-tools/blob/feat/922/contextPath4customPage/packages/fe-fpm-writer/src/page/common.ts#L168 and https://github.com/SAP/open-ux-tools/blob/feat/922/contextPath4customPage/packages/fe-fpm-writer/src/page/custom.ts#L52
> 
> If you create a customPage and you do **not** provide a `minVersion` then we will use `entitySet` but we will use the template valid for `1.94` and above. So, our handling of the absence of a minVersion is inconsistent. Please consistently handle the missing minVersion.

@tobiasqueck
 The difference between the 1.84 and the 1.94 template for custom pages is not w.r.t. `options>settings`, but there is a difference in the view definition that needs to be considered. 
The contextPath vs. entitySet decision is orthogonal  to that. 
Yes, the handling of missing `minVersion` is different: 
- in case of custom page, the assumption is for `latest`, resulting in the newer template version which would be invalid for older UI5 versions. Here, the best solution would be to make `minVersion` mandatory. But this would be an incompatible change, calling applications could break.
- in case of `options>settings`, the defaullt should stay with entitySet if no UI5 version is given, as Fiori V4 supports `entitySet` throughout all versions. Here we have no problem.
In the end, I vote for not changing anything on both sides.

hitesh-parmar (dismissed):
LGTM ✅

tobiasqueck:
> LGTM ✅

@hitesh-parmar please provide more information about what you reviewed and what you didn't do

Klaus-Keller:
There is a new [code smell](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=952&id=SAP_open-ux-tools&open=AYdLvE9Km5Fb-3wyeV3K):

image

schreckstefan:
> There is a new [code smell](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=952&id=SAP_open-ux-tools&open=AYdLvE9Km5Fb-3wyeV3K):
> 
> image

@Klaus-Keller 
That's not new. But the pattern that Tobias has proposed, thus I do not want to change it.

tobiasqueck:
> > There is a new [code smell](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=952&id=SAP_open-ux-tools&open=AYdLvE9Km5Fb-3wyeV3K):
> > image
> 
> @Klaus-Keller That's not new. But the pattern that Tobias has proposed, thus I do not want to change it.

Can't you simply fix it with
```
config.settings.controlConfiguration ??= {};
 const controlConfig = config.settings.controlConfiguration  as Record<string, Record<string, unknown>>;
```

schreckstefan:
> > > There is a new [code smell](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=952&id=SAP_open-ux-tools&open=AYdLvE9Km5Fb-3wyeV3K):
> > > image
> > 
> > 
> > @Klaus-Keller That's not new. But the pattern that Tobias has proposed, thus I do not want to change it.
> 
> Can't you simply fix it with
> 
> ```
> config.settings.controlConfiguration ??= {};
> const controlConfig = config.settings.controlConfiguration  as Record<string, Record<string, unknown>>;
> ```
> 
> PS: no matter what I propose, Sonar has the authority to overrule me ;)

fixed

schreckstefan:
> > The difference between the 1.84 and the 1.94 template for custom pages is not w.r.t. `options>settings`, but there is a difference in the view definition that needs to be considered. The contextPath vs. entitySet decision is orthogonal to that. Yes, the handling of missing `minVersion` is different:
> > 
> > * in case of custom page, the assumption is for `latest`, resulting in the newer template version which would be invalid for older UI5 versions. Here, the best solution would be to make `minVersion` mandatory. But this would be an incompatible change, calling applications could break.
> > * in case of `options>settings`, the defaullt should stay with entitySet if no UI5 version is given, as Fiori V4 supports `entitySet` throughout all versions. Here we have no problem.
> >   In the end, I vote for not changing anything on both sides.
> 
> Sorry but no. We are consistently handling the absence of a `minVersion` as an indicator to use the latest version e.g. https://github.com/search?q=repo%3ASAP%2Fopen-ux-tools+minVersion+fe-fpm-writer&type=code. https://github.com/SAP/open-ux-tools/blob/feat/922/contextPath4customPage/packages/fe-fpm-writer/src/page/common.ts#L168 would be the very only exception, therefore, please stay consistent to the existing code and take the absence of a `minVersion` as an indicator for the latest version.

OK, I adapted the logic now.  Quite some  snapshots have been actualized, as the minimalInput did not pass the minUI5Version, resulting in a switch to contextPath .

schreckstefan:
ERROR: 

The version of node.js (12) you have used to run this analysis is deprecated and we stopped accepting it.
Please update to at least node.js 14. You can find more information here: https://docs.sonarcloud.io/appendices/scanner-environment/

@kranthie-sap

schreckstefan:
@tobiasqueck @Klaus-Keller 
Ready for re-review

Klaus-Keller (approved):
I've only checked the part of the code smell that I've pointed out. That part is fixed.
Approved from my side, but this needs additional review for the rest of the solution.

tobiasqueck (approved):
Reviewed https://github.com/SAP/open-ux-tools/pull/952/commits/cc1e19a0dbdc8540ea62af2772e04057b2c39ade and https://github.com/SAP/open-ux-tools/pull/952/commits/92573c1bd633d6a37eca24e49f3ee8961d13a31f and now all my requested changes are applied.

hitesh-parmar (approved):

nlunets (dismissed):

nlunets (commented):
Checked the part of the generated code and it looks fine.
The logic on how to apply this and how it's applied also seems ok.
However i don't see any test snapshot for LR > OP > SubOP only one for LR > OP > OtherOP

Can you also provide one such test ?

schreckstefan:
> Checked the part of the generated code and it looks fine. The logic on how to apply this and how it's applied also seems ok. However i don't see any test snapshot for LR > OP > SubOP only one for LR > OP > OtherOP
> 
> Can you also provide one such test ?

@nlunets Sorry, I've seen your comment too late. Maybe a missing refresh.

nlunets:
Thanks

PR #953, opened 2023-03-16
Feat/app config writer suggestion

tobiasqueck:
Suggestion for a better usage of the `axios-extension`

### DIFF
--- a/packages/app-config-writer/src/types/smartLinks.ts
+++ b/packages/app-config-writer/src/types/smartLinks.ts
@@ -7,13 +7,12 @@ export enum DeployConfig {
     'FioriToolsServestatic' = 'fiori-tools-servestatic'
 }
 
-export enum UrlParameters {
-    UI2 = '/sap/bc/ui2/start_up?so=%2A',
-    Action = 'action=%2A',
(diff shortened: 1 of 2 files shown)


packages/app-config-writer/src/types/smartLinks.ts
```diff
@@ -7,13 +7,12 @@ export enum DeployConfig {
...
 }
 
-export enum UrlParameters {
-    UI2 = '/sap/bc/ui2/start_up?so=%2A',
-    Action = 'action=%2A',
-    Alias = 'systemAliasesFormat=object',
-    Language = 'sap-language=EN',
-    ShellType = 'shellType=FLP',
-    Depth = 'depth=0'
+export const UrlParameters = {
```
tobiasqueck, line 10: btw, I would move this constant into the code where you use it since it is only used once and makes it easier to see directly in the code what we are adding to the request

johannes-kolbe: Ok, will move that

johannes-kolbe (approved):
Thanks @tobiasqueck for the code overhaul!

PR #956, opened 2023-03-21
feat: column ids added to flexible table header row cells

vadson71:


devinea (commented):
@vadson71 Can you add a changeset please ?
https://github.com/SAP/open-ux-tools#create-changesets-for-feature-or-bug-fix-branches

vadson71:
> @vadson71 Can you add a changeset please ? https://github.com/SAP/open-ux-tools#create-changesets-for-feature-or-bug-fix-branches

Sure. Will do in a moment

devinea (commented):
Code changes lg.
Could add a test that the id's are set as expected. 
There should be an issue created and linked to the PR.

vadson71:
@devinea 
I've adapted unit test to cover new changes. Issue is also created and linked.
Please review once again. Thanks!

devinea (approved):

packages/ui-components/src/components/UIFlexibleTable/UIFlexibleTableRow.tsx
```diff
@@ -134,7 +142,11 @@ export function renderTitleRow<T>(props: UIFlexibleTableProps<T>, paddingRight:
                 column.className
             ]);
             rowCellsData.push(
-                <div key={`title-cell-${key}-${i}`} className={className} title={column.tooltip}>
+                <div
+                    key={`title-cell-${key}-${i}`}
+                    className={className}
+                    title={column.tooltip}
+                    id={`columnKey-${key}`}>
```
815are, line 149: could be issue with unique ids. If we have two tables with different table ids, but columns has same keys, then we will get duplicate ids. I think in such case usually table id is used to build child ids with concatenation

vadson71: fixed

vadson71:
@devinea 
Please approve once again. One more commit is done - work on Andis comments.

815are (approved):
Checked:
1. Code looks ok - ids for column header uses table id as prefix. Should produce unique id if table id is maintained well from consumer;
2. Test coverage is great;
3. Tested with storybook - ids looks as expected;

But did not test with extensions

PR #960, opened 2023-03-22
chore(all) jest 29 upgrade

devinea:
Jest 29 upgrade to align with tools-suite
 - includes better support for ESM modules.


tobiasqueck (changes requested):
What is the reason for the update? I don't mind it, but the PR should contain some information about why we are upgrading.

packages/logger/package.json
```diff
@@ -35,7 +35,7 @@
         "@types/lodash": "4.14.176",
         "@types/vscode": "1.63.1",
         "jest-extended": "2.0.0",
-        "jest": ">=27.2.5",
+        "jest": ">=29.5.0",
```
Klaus-Keller, line 38: Why don't we use fixed version as usual?
```suggestion
        "jest": "29.5.0",
```

packages/store/package.json
```diff
@@ -42,7 +42,7 @@
         "fast-check": "2.19.0",
         "jest-extended": "3.2.3",
         "memfs": "3.3.0",
-        "jest": ">=27.2.5",
+        "jest": ">=29.5.0",
```
Klaus-Keller, line 45: same here
```suggestion
        "jest": "29.5.0",
```

packages/ui-components/package.json
```diff
@@ -59,15 +59,15 @@
         "@types/uuid": "3.4.7",
         "@testing-library/react": "12.1.5",
         "@testing-library/jest-dom": "5.16.5",
-        "babel-jest": "^24.8.0",
+        "babel-jest": "^29.5.0",
```
Klaus-Keller, line 62: and here 
```suggestion
        "babel-jest": "29.5.0",
```

packages/logger/package.json
devinea: Because I want to use the version from the root rather than installing another version (if they diverge) from the root.

devinea:
> What is the reason for the update? I don't mind it, but the PR should contain some information about why we are upgrading.

Added description

packages/logger/package.json
kranthie-sap: @devinea: Shouldn't the jest dev dependencies in the root be sufficient? We should remove them from the individual packages, no?

devinea: @kranthie-sap ideally yes, but needed for peer dependency.
See https://github.com/SAP/open-ux-tools/pull/708#discussion_r1113600956

kranthie-sap: OK,thanks.
@ullasholla: what does jest-extended bring in? Is it worth bubbling the dev dep up?

ullasholla: `jest-extended` Array and Object matchers are very useful. 
Especially something like this: https://jest-extended.jestcommunity.dev/docs/matchers/Array#toincludesamemembersmembers
This particular matcher is used to assert that the returned array contains the same elements as the expected array, in any order. We have APIs where the order of the array returned is not part of the contract (it cannot be enforced by TypeScript types).

Example usages in our repo: https://github.com/search?q=repo%3ASAP%2Fopen-ux-tools+toIncludeSameMembers&type=code

Technically jest@29.5.0 should work well with jest-extended@2.0.0 (the version we use)

tobiasqueck: @ullasholla the four examples of where we use it could also be done with 
```typescript
expect(yourArray).toEqual(expect.arrayContaining([expectedMember1, expectedMember2]);
```
Which doesn't seem to be more complicated are harder to understand then the helper from `jest-extended`. So, I'd also prefer removing the dependency.

ullasholla: @tobiasqueck, it's not an exact replacement. 
From jest [documentation](https://jestjs.io/docs/expect#expectarraycontainingarray):
> expect.arrayContaining(array) matches a received array which contains all of the elements in the expected array. That is, the expected array is a subset of the received array.

To make them equivalent, we'll need to assert the length of the arrays are equal as well.

The replacement is not complicated but more verbose.

I'm in favor of reducing our dependencies as well. It seemed genuinely useful to me, hence the inclusion.

> the four examples of where we use it could also be done with...

To nit pick, it's used in 4 files. 15 actual instances of usage 😛
(I'm not sure if we use other matchers from jest-extended, `toIncludeSameMembers` came to mind readily).

@devinea, maybe you know the answer to this:
> Technically jest@29.5.0 should work well with jest-extended@2.0.0 (the version we use)

tobiasqueck: > To make them equivalent, we'll need to assert the length of the arrays are equal as well.

I am ok with keeping it if it makes sense. 
Btw, if you want to check that the array has exactly the elements you expect wouldn't it become a simple `expect(yourArray).toEqual([expectedMember1, expectedMember2]);` ... but again, I am ok if it makes code easier to maintain

ullasholla: We'll need to `sort()` the arrays to use `toEqual()`
Something like this:
```typescript
  const actual = [1, 2];
  const expected = [2, 1];

  // Will fail
  expect(actual).toEqual(expected);

  // Sorting arrays immutably will make it pass
  expect([...actual].sort()).toEqual([...expected].sort());

  // Can also use Array.prototype.sort(). Will modify original arrays
  expect(actual?.sort()).toEqual(expected.sort());

  // Same as
  expect(actual).toIncludeSameMembers(expected);
```
I'm ok with either. `toIncludeSameMembers()` is very explicit about what to expect and does not need modifying the arrays (or use of destructuring to avoid modifying them).

Opinions, @tobiasqueck, @kranthie-sap, @Klaus-Keller & @devinea?

devinea: I don't have a strong view on it. But I am fine with keeping `jest-extended`.
Any change like that should be dealt with in a separate PR.

tobiasqueck (approved):
- no changeset - as expected
- pipeline shows no errors, so build & test works
- version entries (fixed in root, range in project) makes sense
- lock-file changes as expected

packages/logger/package.json
Klaus-Keller: @ullasholla, I did a search for `jest-extended` and saw usage in `@sap-ux/logger` and `@sap-ux/store`. Do you know what exact features from `jest-extended` we use? Is it only `toIncludeSameMembers()`?

As usual, to get a better feeling about cost/benefit of using a certain module, I ran `npm i jest-extended` in an empty folder and check the newly create `node_modules` folder using `du -h -d 1 node_modules` afterwards. Here is the result:

```terminal
$ du -h -d 1 node_modules 
 56K	node_modules/diff-sequences
 40K	node_modules/react-is
360K	node_modules/jest-extended
 20K	node_modules/@jest
 20K	node_modules/color-name
 52K	node_modules/chalk
 20K	node_modules/has-flag
 20K	node_modules/supports-color
 44K	node_modules/color-convert
 32K	node_modules/ansi-styles
508K	node_modules/@sinclair
 16K	node_modules/jest-get-type
116K	node_modules/pretty-format
104K	node_modules/jest-diff
1.4M	node_modules
```

While the size doesn't matter too much for `devDependencies`, it still adds install time for every developer and every build run. Also the amount of transitive dependencies is a bit high, considered what we are using it for. 

**My proposal:** if we use `jest-extended` solely for `toIncludeSameMembers()`, I would change the tests and remove it. If we use other functionalities in multiple places, I propose to keep it.

devinea: @Klaus-Keller FYI: I believe all those dependencies are common with jest so it is not a big impact on developer install size.

Klaus-Keller: Even if we decide to remove `jest-extended`, it should be done in a separate PR. @ullasholla, just let us know your preference.

devinea:
Thanks all

packages/logger/package.json
ullasholla: @Klaus-Keller, like Austin mentioned, `jest-extended` uses a lot of libraries that are common with `jest`. I don't think the cost to the developer is that high. I'd prefer to leave it alone for now

PR #963, opened 2023-03-27
Feat/962/export file exists function

korotkovao:


slin-sap (approved):

Klaus-Keller (changes requested):
Hi @korotkovao, 

Could you please add some background information that tells us why function 

https://github.com/SAP/open-ux-tools/blob/e40fff755ec01b2ea711b7f64bab2d27c54b05b5/packages/project-access/src/file/file-access.ts#L34-L52

should be exported in public API? 

The reason I ask is, we initially thought about whether or not to make the [file access functions](https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/file/file-access.ts) public available. Eventually, we've decided against it: The functions provide low level file access, which can be easily achieved using functions from built-in module `fs`. However, we've also added support `mem-fs` editor, so you can pass an instance of `Editor` which then will be also use to do file operations (e.g. a file exists in mem-fs only). If it's the `mem-fs` capabilities you are looking to use we could consider to add these functions to public interface.

tobiasqueck (changes requested):
The PR has no description or any reasoning why this change is required.

korotkovao:
Thank you for having a look @Klaus-Keller et all
I will close the PR.

tobiasqueck:
@korotkovao, since the PR is closed, could the corresponding branch be deleted?

Issue #964, opened 2023-03-28
BUG - When deploying to an ABAP system, the BAS destination URL is being referenced instead of the SAP BTP destination URL


longieirl:
### Description
Found when testing https://github.com/SAP/open-ux-tools/issues/687

Sample SAP BTP destination;
```
#
Description=OData Serivce URL
Type=HTTP
HTML5.DynamicDestination=true
Authentication=NoAuthentication
WebIDEUsage=odata_abap
Name=northwind
WebIDEEnabled=true
ProxyType=Internet
URL=https\://services.odata.org
```
__These destination properties are only for display purposes__

When deploying the ABAP app using verbose mode, the payload is being sent as;

```XML
<entry xml:base="https://northwind.dest" xmlns="http://www.w3.org/2005/Atom" xmlns:d="http://schemas.microsoft.com/ado/2007/08/dataservices" xmlns:m="http://schemas.microsoft.com/ado/2007/08/dataservices/metadata">
	<id>https://northwind.dest/Repositories('ZTRAJL2403')</id>
	<title type="text">Repositories('ZTRAJL2403')</title>
	<updated>2023-03-27T20:14:07.950Z</updated>
	<category scheme="http://schemas.microsoft.com/ado/2007/08/dataservices/scheme" term="/UI5/ABAP_REPOSITORY_SRV.Repository"/>
	<link href="Repositories('ZTRAJL2403')" rel="edit" title="Repository"/>
	<content type="application/xml">
		<m:properties>
			<d:Name>ZTRAJL2403</d:Name>
			<d:Package>--TRANSPORT</d:Package>
			<d:Description>Travel App</d:Description>
			<d:ZipArchive/>
			<d:Info/>
		</m:properties>
	</content>
</entry>
```

The `xml:base` and `id` are showing the BAS destination host `northwind.dest` instead of the destination URL property `https://services.odata.org`

Also, if the ABAP app is successfully deployed, the frontend URL shown to the user is;
```
App available at https://northwind.dest/sap/bc/ui5_ui5/sap/mypp
```
instead of 
```
App available at https://services.odata.org/sap/bc/ui5_ui5/sap/mypp
```

### Expected results
The payload to create or update an existing ABAP app should show the SAP BTP destination URL, similarly for the frontend log as well.

### Actual results
Showing the BAS destination URL which is only applicable for BAS and shouldnt be referenced.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
[...]

PR #967, opened 2023-03-29
Fix/964/bas destination host

longieirl:
Fix for [#964](https://github.com/SAP/open-ux-tools/issues/964)

Summary of changes;

1. For BAS destination flows, the SAP BTP destination URL needs to be referenced as part of the POST/PUT payload when creating/updating the ABAP and when showing the URL path to the deployed ABAP app
2. Change to `createUi5Archive` to take `projectNamespace` if present, as the calling client can use a `projectNamespace` when creating the workspace dist space
3. Additional unit tests
4. Additional logging added based on testing ABAP deployment, these additional logging statements provide more information when in verbose mode


packages/axios-extension/src/abap/ui5-abap-repository-service.ts
```diff
@@ -99,18 +99,21 @@ function encodeXmlValue(xmlValue: string): string {
...
+    private readonly publicUrl: string;
 
     /**
      * Extension of the base constructor to set preferred response format if not provided by caller.
      *
      * @param config optional base configuration for Axios
+     * @param publicUrl optional string representing the host exposing the ABAP repository
      */
-    public constructor(config?: AxiosRequestConfig) {
+    public constructor(config?: AxiosRequestConfig, publicUrl?: string) {
```
tobiasqueck, line 110: Could we keep it in one object in case we want to add more in the future

```
public constructor(config?: AxiosRequestConfig & { publicUrl?: string }) {
```

packages/axios-extension/src/abap/ui5-abap-repository-service.ts
```diff
@@ -99,18 +99,21 @@ function encodeXmlValue(xmlValue: string): string {
...
         config = config ?? {};
         config.headers = config.headers ?? {};
         // @see https://axios-http.com/docs/config_defaults
         config.headers['Accept'] = config.headers['Accept'] ?? 'application/json,application/xml,text/plain,*/*';
         super(config);
+        this.publicUrl = publicUrl ?? this.defaults.baseURL;
```
tobiasqueck, line 116: Wait ... why do we need this here? Didn't we have the very same logic in the getter function on the provider

packages/axios-extension/src/abap/abap-service-provider.ts
```diff
@@ -62,6 +67,24 @@ export class AbapServiceProvider extends ServiceProvider {
         return this.atoSettings;
     }
 
+    /**
+     * Set the public facing URL, typically used for a destination related flows.
+     *
+     * @param url
+     */
+    public setPublicUrl(url: string) {
```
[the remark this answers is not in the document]
tobiasqueck: Or is this a problem when overwriting it in a subclass?

packages/axios-extension/test/abap/ui5-abap-repository-service.test.ts
```diff
@@ -109,7 +113,7 @@ describe('Ui5AbapRepositoryService', () => {
                     `${Ui5AbapRepositoryService.PATH}/Repositories(%27${validApp}%27)?${updateParams}`,
                     (body) => body.indexOf(archive.toString('base64')) !== -1
                 )
-                .replyWithError('Deployment failed');
+                .reply(401, 'Deployment failed');
```
tobiasqueck, line 116: What is the reason for this change?

tobiasqueck (changes requested):
A few comments added.

packages/axios-extension/test/abap/ui5-abap-repository-service.test.ts
longieirl: This actually provides the AxiosError object populated the error.response.data is populdated with `Deployment failed` which is required to cover some of the logging changes made.

packages/axios-extension/src/abap/ui5-abap-repository-service.ts
longieirl: So the `AbapServiceProvider` is aware of the `publicUrl` when its created as part of the `createForDestination` factory flow;
```
if (isAbapSystem(destination)) {
        provider = createInstance<AbapServiceProvider>(AbapServiceProvider, providerConfig);
        // For an ABAP destination flow, need to show the destination host URL property instead of the BAS host URL i.e. https://mydest.dest
        (provider as AbapServiceProvider).publicUrl = destination.Host;
    } else {
        provider = createInstance<ServiceProvider>(ServiceProvider, providerConfig);
    }
```
Thus, when `getUi5AbapRepository` is called from `AbapServiceProvider` the `publicUrl` is passed to the service instance which is created by the base service-provider.

tobiasqueck: My thinking was that the snippet above is only called by `provider.createService` and it uses `provider.publicUrl` which in its getter has a similar check for undefined. However, assuming this constructor could also be called from somewhere else, you are right, the check needs to be there.

Minor thing: in the getter you use `||`, here you use `??`. I always forget which minor difference the operators have, therefore, please check which one makes more sense here, and then use the same in both cases. Or was it a conscious decision?

tobiasqueck (approved):
* changeset is fine
* implemented changes are documented in the PR description
* implementation adheres to coding standards
* test coverage is sufficient
* I did NOT manually test it and assume its functional correctness based on the review and the test cases

longieirl:
Sample verhose output from BAS destination ABAP deployment;
```
2023-03-30 14:13:45 debug   abap-deploy-task: Retrieving application ZTRAVELJL3003, error found Error: Request failed with status code 401 
2023-03-30 14:13:45 debug   abap-deploy-task: Payload:
<entry xmlns="http://www.w3.org/2005/Atom"       xmlns:m="http://schemas.microsoft.com/ado/2007/08/dataservices/metadata"       xmlns:d="http://schemas.microsoft.com/ado/2007/08/dataservices"       xml:base="http://internalhost:50000">  <id>http://internalhost.net.sap:50000/Repositories('ZTRAVELJL3003')</id>  <title type="text">Repositories('ZTRAVELJL2903')</title>  <updated>2023-03-30T14:13:45.768Z</updated>  <category term="/UI5/ABAP_REPOSITORY_SRV.Repository" scheme="http://schemas.microsoft.com/ado/2007/08/dataservices/scheme"/>  <link href="Repositories('ZTRAVELJL3003')" rel="edit" title="Repository"/>  <content type="application/xml">    <m:properties>      <d:Name>ZTRAVELJL2903</d:Name>      <d:Package>Z_I313149_PACKAGE</d:Package>      <d:Description>Travel App</d:Description>      <d:ZipArchive>UEsDBBQACAgIACqDfVYAAAAAAAAAAAAAAAAQAAAAQ29tcG9uZW50LWRiZy5qc12NywqDMBBF935FyEpBzF5XpZ9RukjNWEaSSTATsBT/vdFC+ri7OWfmTtShS9gZmJCgrkTORUYdVEzIoBKqOxAsOCoGF6zOzOJNnUI4exc8AbG8tsfdlGhk9CTqohrxPNQemSKIyLmK5VAVvACnhUQ5
```
using the command;
```
./bin/deploy --destination MyDest --name ZTRAVELJL3003 --description 'Travel App' --package Z_I313149_PACKAGE --transport Y11K90044 --verbose true --archive-folder /home/user/projects/jltestdeployment/dist
```
Where `MyDest` exposes URL `http://internalhost:50000`

devinea (approved):
changeset, code and test changes look good per description.
I haven' tested locally.

Issue #973, opened 2023-04-04
BUG - Sonar: Unsafe usage of optional chaining


kranthie-sap:
### Description
The following bugs with 'major' priority are reported in sonar cloud `Unsafe usage of optional chaining. If it short-circuits with 'undefined' the evaluation will throw TypeError.`:
* [packages/axios-extension/src/auth/connection.ts](https://sonarcloud.io/code?id=SAP_open-ux-tools&selected=SAP_open-ux-tools%3Apackages%2Faxios-extension%2Fsrc%2Fauth%2Fconnection.ts&line=39)
* [packages/app-config-writer/src/smartlinks-config/ui5-yaml.ts](https://sonarcloud.io/code?id=SAP_open-ux-tools&selected=SAP_open-ux-tools%3Apackages%2Fapp-config-writer%2Fsrc%2Fsmartlinks-config%2Fui5-yaml.ts&line=29)

### Why is this an issue?
<img width="1001" alt="Screenshot 2023-04-04 at 09 01 24" src="https://user-images.githubusercontent.com/84985584/229850727-f901e18e-0bed-4ad3-b1ee-17fc2f9dfd07.png">

### How to fix the issue?
<img width="995" alt="Screenshot 2023-04-04 at 09 01 32" src="https://user-images.githubusercontent.com/84985584/229850746-dadc1c63-734e-4e4a-8701-896fa21e4dc9.png">

tobiasqueck:
@kranthie-sap it looks to me like this fixed with #974 so I am closing it. If it is still an issue then please reopen and explain what is missing.

PR #974, opened 2023-04-04
fix(sonar): unsafe usage of optional chaining

kranthie-sap:
# Issue

#973

# Description
* Update `axios-extension` and `app-config-writer` packages with fixes for the sonar bug `Unsafe usage of optional chaining. If it short-circuits with 'undefined' the evaluation will throw TypeError.`


devinea (approved):
changes and tests lg. 
Sonar smells are gone.

packages/axios-extension/src/auth/connection.ts
```diff
@@ -35,8 +35,8 @@ export class Cookies {
      * @returns cookies object
      */
     public addCookie(cookieString: string): Cookies {
-        const cookie = cookieString.split(';');
-        const [key, ...values] = cookie[0]?.split('=');
+        const cookie = cookieString?.split(';') || [];
```
tobiasqueck, line 38: Based on the function signature `cookieString` is always set, why do we need to do the null check then in line 38?

tobiasqueck: And Sonar agrees with me ;)
It only complain about line 39 https://sonarcloud.io/code?id=SAP_open-ux-tools&selected=SAP_open-ux-tools%3Apackages%2Faxios-extension%2Fsrc%2Fauth%2Fconnection.ts&line=39

kranthie-sap: @tobiasqueck: I agree that it is not required if used like in the tests. However, since `Cookies` is a public type and a consumer could directly use `addCookie`, this additional safety may be required.

kranthie-sap: Actually, I take that back. It looks like `split` works fine with an empty string. The result of `''.split[';']` results in `['']` and the check in line 39 should then be sufficient. I'll update the code.

kranthie-sap: Reverted changes in line 38.

tobiasqueck (dismissed):
changeset is good
code changes adhere to coding standards
test added verifying that the change does not cause any surprises

tobiasqueck (approved):

Issue #975, opened 2023-04-06
BUG - project access api does not identify library projects


slin-sap:
### Related Feature
Feature request: _issue number_

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

slin-sap:
@Klaus-Keller

The following code cannot identity the programming language when applying it to the example project attached in this issue, since the example project doesn't have webapp folder. If language type needs to be detected for this type of migrated ui5 library projects, this method also needs to be updated to support this.

https://github.com/SAP/open-ux-tools/blob/6789f09bf97ba0557d42ea39a45115cfbb0f2059/packages/project-access/src/project/info.ts#L19

Klaus-Keller:
Hi @slin-sap,
https://github.tools.sap/fiori-tools-qa/se.mi.plm.attachmentservice is not a Fiori application project, it is a project containing a library.
I think there needs to be further discussions on general support of library development using SAP Fiori tools.
FYI @ashleytung @devinea @kranthie-sap @schreckstefan

PR #976, opened 2023-04-06
fix: find library projects

slin-sap:
#975 

Provide option to also include 'library' type UI5 projects when finding Fiori projects in a workspace.


devinea (approved):
Thanks @slin-sap 
Code changes, tests and changeset look good

packages/project-access/src/project/search.ts
```diff
@@ -204,19 +204,26 @@ async function findRootsForPath(path: string): Promise<{ appRoot: string; projec
...
+): Promise<AllAppResults[]> {
     const result: AllAppResults[] = [];
     const manifestPaths = await findAllManifest(wsFolders);
 
     for (const manifestPath of manifestPaths) {
         try {
             // All UI5 apps have at least sap.app: { id: <ID>, type: "application" } in manifest.json
             const manifest = await readJSON<Manifest>(join(manifestPath, FileName.Manifest));
-            if (!manifest['sap.app'] || !manifest['sap.app'].id || manifest['sap.app'].type !== 'application') {
+            // With default `includeLibraries` set to false, apps that do not have type: "application" in manifest.json are excluded.
```
Klaus-Keller, line 221:
```suggestion
            // With default `includeLibraries` set to false, only apps that do have type: "application" in manifest.json are included.
```

.changeset/bright-suits-fold.md
```diff
@@ -0,0 +1,13 @@
+---
+'@sap-ux/project-access': minor
+---
+
+Updated API of search function `findAllApps()`. It now takes a 2nd argument to
+include UI5 library projects in the returned Fiori app list.
+
+E.g.
+```typescript
+const appsAndLibraries = findAllApps(wsFolders, true);
```
Klaus-Keller, line 10: if we change to `findOptions` object this example needs to be updates as well.

PR #977, opened 2023-04-13
docs: add description to use tools offline

Klaus-Keller:
Add documentation on how to use tools in an isolated environment.

### DIFF
--- a/docs/tools-offline-instructions/dependencies.json
+++ b/docs/tools-offline-instructions/dependencies.json
@@ -1,0 +1,26 @@
+{
+    "vscode_extensions": {
+        "sapse.sap-ux-fiori-tools-extension-pack": "latest"
+    },
+    "node_modules": {
+        "@sap/abap-deploy": "latest",
+        "@sap/cds-compiler": "latest",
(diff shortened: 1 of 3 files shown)


docs/tools-offline-instructions/dependencies.json
```diff
@@ -0,0 +1,27 @@
+{
+    "vscode_extensions": {
+        "sapse.sap-ux-fiori-tools-extension-pack": "latest"
+    },
+    "node_modules": {
```
voicis, line 5: `@sap/ux-cds-odata-language-server-extension` is also required for CAP CDS projects. CAP CDS VSCode extension loads this dynamically from NPM registry and the registry URL can be overwritten with VSCode setting  `cds.contributions.registry`.

tobiasqueck:
@Klaus-Keller is this one still being worked on?

docs/tools-offline-instructions/dependencies.json
Klaus-Keller: Added it to `docs/tools-offline-instructions/info/globalModules.json`

tobiasqueck:
@Klaus-Keller any plan to merge this?

Issue #980, opened 2023-04-17
FEATURE - Add configuration option for generation of standalone index.html


unseen1980:
As a **developer**, I should have the option to toggle the generation of index.html. If I choose not to generate the ```index.html``` then there should not be a ```start-noflp``` target available in the package.json.

### Description
For some scenarios, the index.html file is not required, so there is no point to generate it.

### Technical Design
Add a ~~```generateStandaloneIndexHtml```~~  generateIndex property for toggling ```start-noflp```  and index.html generation.

### Acceptance Criteria
Given I generate a Fiori application using the generator and setting the generateIndex to false, I will not get a webapp/index.html file generated, nor will there be a ```start-noflp``` target created in the package.json of the generated project.

@ianquigley-sap @devinea

unseen1980:
This has been completed.

PR #981, opened 2023-04-17
Feat/app gen/remove standalone index html

unseen1980:
https://github.com/SAP/open-ux-tools/issues/980 

Add configuration option ```generateIndex```to toggle the generation of index.html


packages/fiori-elements-writer/src/packageConfig.ts
```diff
@@ -60,9 +63,13 @@ export function getPackageJsonTasks({
     return Object.assign(
```
tobiasqueck, line 63: I know it wasn't introduced in your change but why do we use `Object.assign` here? Wouldn't this be better readable with something like 
```
const scripts = {
    start: startCommand,
    'start-local': startLocalCommand
};
if (generateIndex) {
    scripts['start-mock'] = localOnly
        ? `echo \\"${t('info.mockOnlyWarning')}\\"`
        : `fiori run --open "${'index.html'}${searchParam}"`;
}
if (addMock) {
    scripts['start-mock'] = `fiori run --config ./ui5-mock.yaml --open "test/flpSandbox.html${params}"`;
}
...
```

packages/fiori-freestyle-writer/src/packageConfig.ts
```diff
@@ -56,9 +59,9 @@ export function getPackageJsonTasks({
     return Object.assign(
         {
             start: startCommand,
-            'start-local': startLocalCommand,
-            'start-noflp': startNoFlpCommand
+            'start-local': startLocalCommand
```
tobiasqueck, line 62: same comment here as before

tobiasqueck (changes requested):
- changeset correct and meaningful
- code adheres to coding guidelines and is easy to understand
- added test covers the new feature
- did NOT locally test but reviewed the new test snapshot
- minor comment about an existing modified code section
   - once clarified, I would approve the PR

packages/fiori-elements-writer/test/lrop.test.ts
```diff
@@ -208,6 +208,22 @@ describe(`Fiori Elements template: ${TEST_NAME}`, () => {
                 }),
                 service: v2Service
             } as FioriElementsApp<LROPSettings>
+        },
```
tobiasqueck, line 211: The test is fine however for test simplification, this could be a separate test in which we do not dump the whole project into a snapshot but just check that the index file is not generated and the `package.json` does not contain the script

packages/fiori-elements-writer/src/packageConfig.ts
unseen1980: I think readability would be more or less the same, but I like that with your suggestion, we don't create new variables if the ```generateIndex``` or the ```addMock``` is false. 

Please review again, I made the requested changes, thanks

packages/fiori-freestyle-writer/src/packageConfig.ts
unseen1980: Please review again, I made the requested changes, thanks

packages/fiori-elements-writer/test/lrop.test.ts
unseen1980: No objection to this, to be honest I followed the test pattern here. Would it be ok to create a tbi for this? I need this PR merged if possible today.  I think we can do the same for other tests as well.

tobiasqueck: Yes, that doesn't need to be part of this PR. We need a separate task to review our tests and make sure that they are still maintainable.

tobiasqueck (dismissed):
Previously requested changes have been addressed and the rest of the previous review is still valid.

packages/fiori-elements-writer/src/index.ts
```diff
@@ -153,7 +153,8 @@ async function generate<T extends {}>(basePath: string, data: FioriElementsApp<T
             sapClient: feApp.service?.client,
             flpAppId: feApp.app.flpAppId,
             startFile: data?.app?.startFile,
-            localStartFile: data?.app?.localStartFile
+            localStartFile: data?.app?.localStartFile,
+            generateIndex: feApp.appOptions?.generateIndex
```
IainSAP, line 157: Thanks for the better name :) 👍

IainSAP (approved):
Thx @unseen1980 

- Reviewed code changes
- Tests updated
- Previous property name improved

IainSAP (approved):
Re-approving based on expectation that follow-up TBI will be created to remove duplication excluded by Sonar config.

Thx @unseen1980

PR #985, opened 2023-04-19
feat(fe-fpm-writer): generate custom filter

AlinaGovoruhina:
Generate custom filter with fpm-writer
- Add fragment.xml file template, manifest.json template, 2 controller file templates: JS and TS
- Use apply event handler configuration
- Add unit tests


packages/fe-fpm-writer/src/common/event-handler.ts
```diff
@@ -51,17 +52,18 @@ export function applyEventHandlerConfiguration(
     eventHandler: EventHandlerConfiguration | true | string,
     controllerSuffix = false,
     typescript?: boolean,
-    parameters: EventHandlerTypescriptParameters = defaultParameter
+    parameters: EventHandlerTypescriptParameters = defaultParameter,
+    defaultConfig = { templatePath: 'common/EventHandler', fnName: 'onPress', fileName: config.name }
```
kranthie-sap, line 56: Should this be a part of the `config` parameter already defined?

packages/fe-fpm-writer/templates/filter/Controller.js
```diff
@@ -0,0 +1,21 @@
+sap.ui.define(["sap/ui/model/Filter", "sap/ui/model/FilterOperator"], function(Filter, FilterOperator) {
+	"use strict";
```
kranthie-sap, line 2: It looks like there is a mix of tabs and spaces in all the templates in the module. For the templates in this PR, can you switch to spaces?
@tobiasqueck, @nlunets, @devinea, @Klaus-Keller: What is the guidance on whitespaces in the templates?

kranthie-sap (changes requested):
Hi @AlinaGovoruhina, the code is well structured and has good code coverage with tests. Please see some minor comments.

Since this is an enhancement, please generate a `minor` update changeset using `pnpm changeset` and add an appropriate message for the change. Also, please update the PR description with what has been changed with this PR.

packages/fe-fpm-writer/src/common/event-handler.ts
AlinaGovoruhina: I removed the fileName property since it is available in config.

packages/fe-fpm-writer/templates/filter/Controller.js
```diff
@@ -0,0 +1,21 @@
+sap.ui.define(["sap/ui/model/Filter", "sap/ui/model/FilterOperator"], function(Filter, FilterOperator) {
+    "use strict";
+    return {
+        <%- (typeof eventHandlerFnName !== 'undefined' && eventHandlerFnName) || 'itemsFilter' %>: function(sValue) {
+            switch (sValue) {
+                case "0":
```
schreckstefan, line 6: I know this is only an example implementation that gets overwritten by the user later on, but it will only work for filter properties of numeric  type, I guess (it would not work for the property from https://sapui5untested.int.sap.eu2.hana.ondemand.com/test-resources/sap/fe/core/fpmExplorer/index.html#/customElements/customFilterContent).  I wonder if we could not provide other templates, for other property types.

AlinaGovoruhina: Yes, this is used as an example controller file. In Guided Development any input can be used as a filtered property and the type is not checked. It is not in the scope of the current task and can be added afterwards with templates appropriate for the property types.

packages/fe-fpm-writer/templates/filter/Controller.js
nlunets: I think the recommended ui5 rules are spaces but i don't know what you've applied so far

kranthie-sap: Thanks @nlunets. I see that @AlinaGovoruhina has switched to spaces now. I will mark this as resolved.

kranthie-sap (approved):

kranthie-sap:
The changeset update and the PR description looks good. Rest of the my comments are resolved. Thanks.

packages/fe-fpm-writer/src/filter/types.ts
```diff
@@ -0,0 +1,13 @@
+import type { CustomElement, InternalCustomElement, Position, EventHandler } from '../common/types';
+
+export interface CustomFilter extends CustomElement, EventHandler {
```
tobiasqueck, line 3: Please add JSDoc comments for each property. It is not fully clear to what e.g. `ControlId` would refer to.

packages/fe-fpm-writer/templates/filter/Controller.ts
```diff
@@ -0,0 +1,24 @@
+import Filter from 'sap/ui/model/Filter';
+import FilterOperator from 'sap/ui/model/FilterOperator';
+
+/**
+ * Custom filter
+ * @param sValue selected filter item
+ * @returns new Filter
+ */
+export function <%- (typeof eventHandlerFnName !== 'undefined' && eventHandlerFnName) || 'itemsFilter' %>(sValue: string) {
```
tobiasqueck, line 9: Same as before: Don't you set the value all the time in the `applyEventHandlerConfiguration` function? If yes, please just use `eventHandlerFnName` and don't do the additional checks and fallbacks.
(the same for the `Controller.js` file)

packages/fe-fpm-writer/src/filter/types.ts
AlinaGovoruhina: JSDoc added.

packages/fe-fpm-writer/templates/filter/Controller.js
```diff
@@ -0,0 +1,21 @@
+sap.ui.define(["sap/ui/model/Filter", "sap/ui/model/FilterOperator"], function(Filter, FilterOperator) {
+    "use strict";
+    return {
+        <%- (typeof eventHandlerFnName !== 'undefined' && eventHandlerFnName) || 'itemsFilter' %>: function(sValue) {
```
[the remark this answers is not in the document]
AlinaGovoruhina: `eventHandler` is an optional parameter in the custom filter config but adding function name to be set in `enhanceConfig` allowed to remove these checks.

packages/fe-fpm-writer/templates/filter/Controller.ts
AlinaGovoruhina: Same as above

packages/fe-fpm-writer/src/common/event-handler.ts
```diff
@@ -51,17 +52,18 @@ export function applyEventHandlerConfiguration(
     eventHandler: EventHandlerConfiguration | true | string,
     controllerSuffix = false,
     typescript?: boolean,
-    parameters: EventHandlerTypescriptParameters = defaultParameter
+    parameters: EventHandlerTypescriptParameters = defaultParameter,
+    defaultConfig = { templatePath: 'common/EventHandler', fnName: 'onPress' }
```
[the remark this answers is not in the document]
AlinaGovoruhina: Changed. `enhanceConfig` sets the function name, templatePath parameter left in `applyEventHandlerConfiguration`

packages/fe-fpm-writer/src/filter/index.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ * Enhance the target object in the manifest with the required nested objects and return a reference to it.
+ *
+ * @param {Manifest} manifest - the application manifest
+ * @param {CustomFilter} customFilter - the custom filter configuration
+ * @returns Filters object of the given page
+ */
+function enhanceManifestAndGetFiltersReference(manifest: any, customFilter: CustomFilter): any {
```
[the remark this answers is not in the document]
tobiasqueck: You have another `any` in line 98. Please ensure that you do not use any. If you don't know the type `unknown` is the correct choice because it forces you to handle the variable like a variable you don't anything about.

packages/fe-fpm-writer/src/common/event-handler.ts
```diff
@@ -51,17 +52,18 @@ export function applyEventHandlerConfiguration(
...
+    parameters: EventHandlerTypescriptParameters = defaultParameter,
+    templatePath = 'common/EventHandler'
 ): string {
     if (typeof eventHandler === 'string') {
         // Existing event handler is passed - no need for file creation/update
         return eventHandler;
     }
     // New event handler function name - 'onPress' is default
-    let eventHandlerFnName = 'onPress';
+    let { eventHandlerFnName } = { eventHandlerFnName: 'onPress', ...config };
```
tobiasqueck, line 63: Isn't this line doing nothing else then
```
let eventHandlerFnName = config.eventHandlerFnName ?? 'onPress'
```

packages/fe-fpm-writer/src/filter/types.ts
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Existing filter XML fragment file name without '.fragment.xml' suffix.
+     * If not set, is equal to the name specified in the config.
+     */
+    fragmentFile?: string;
+    /**
+     * Name of the created filter function.
+     * If function name is not set in the eventHandler, is equal to 'itemsFilter'.
+     */
+    eventHandlerFnName?: string;
```
tobiasqueck, line 28: Why do we have this on on the public interface again, shouldn't it be already possible to set it with `eventHandler.fnName`?

packages/fe-fpm-writer/src/filter/types.ts
```diff
@@ -0,0 +1,43 @@
...
+     */
+    label: string;
+    /**
+     * Property used to filter by.
+     */
+    property: string;
+    /**
+     * ID of the embedded filter XML view.
+     */
+    controlID: string;
```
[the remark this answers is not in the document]
tobiasqueck: Wait, I think your JSDoc comment is misleading, it is not the viewId but the id of the combobox in your generated fragment.

packages/fe-fpm-writer/src/filter/index.ts
```diff
@@ -0,0 +1,107 @@
...
+    // clone input
+    const config: CustomFilter & Partial<InternalCustomFilter> = data;
+
+    setCommonDefaults(config, manifestPath, manifest);
+
+    // set default values for requirement, language, the fragment file name
+    config.required = config.required || false;
+    config.typescript = !!config.typescript;
+    config.fragmentFile = config.fragmentFile || config.name;
+    config.eventHandlerFnName =
```
tobiasqueck, line 33: I am wondering if we really need this property, couldn't you do this with the existing `config.eventHandler.fnName`? E.g.
```
if (config.eventHandler === true) {
   config.eventHandler = {};
}

if (typeof config.eventHandler === 'object' && !config.eventHandler.fnName) {
   config.eventHandler.fnName = 'itemsFilter;
}
```
As result `applyEventHandlerConfiguration` should later behave exactly is it does now.

packages/fe-fpm-writer/templates/filter/Controller.js
```diff
@@ -1,7 +1,7 @@
 sap.ui.define(["sap/ui/model/Filter", "sap/ui/model/FilterOperator"], function(Filter, FilterOperator) {
     "use strict";
     return {
-        <%- eventHandlerFnName %>: function(sValue) {
+        <%- (typeof eventHandlerFnName !== 'undefined' && eventHandlerFnName) || 'filterItems' %>: function(sValue) {
```
tobiasqueck, line 4: Why is this one back? Aren't you still ensuring in your code that `eventHandlerFnName` is always defined?

AlinaGovoruhina: There are 2 cases where there is no `eventHandlerFnName` set: 
- eventHandler is not set/undefined, then event handler configuration is not applied,
- eventHandler is a string, then it is used as provided to link an existing event handler.

packages/fe-fpm-writer/src/filter/index.ts
```diff
@@ -0,0 +1,112 @@
...
+            fs,
+            config,
+            config.eventHandler,
+            false,
+            config.typescript,
+            contextParameter,
+            'filter/Controller'
+        );
+    } else {
+        // create a controller file
```
tobiasqueck, line 74: This `else` block is incorrect (i.e. should be deleted). The specification of the `eventhandler` property says "If not set (i.e. undefined) then no event handler is linked", so you shouldn't create anything

packages/fe-fpm-writer/test/unit/filter.test.ts
```diff
@@ -0,0 +1,304 @@
...
+
+        beforeEach(() => {
+            fs = create(createStorage());
+            fs.delete(testDir);
+            fs.write(join(testDir, 'webapp/manifest.json'), testAppManifest);
+        });
+        test('New custom filter (no eventhandler)', () => {
+            generateCustomFilter(testDir, filter, fs);
+            expect(fs.readJSON(join(testDir, 'webapp/manifest.json'))).toMatchSnapshot();
+            expect(fs.exists(getControllerPath(filter))).toBeTruthy();
```
tobiasqueck, line 58: See above comments. If `eventhandler` is undefined, then it shouldn't be created so this should be a `toBe(false)` check.

packages/fe-fpm-writer/templates/filter/Controller.ts
```diff
@@ -0,0 +1,24 @@
+import Filter from 'sap/ui/model/Filter';
+import FilterOperator from 'sap/ui/model/FilterOperator';
+
+/**
+ * Custom filter
+ * @param sValue selected filter item
+ * @returns new Filter
+ */
+export function <%- (typeof eventHandlerFnName !== 'undefined' && eventHandlerFnName) || 'filterItems' %>(sValue: string) {
```
tobiasqueck, line 9: we don't use Hungarian notation in typescript, please just use `value`

packages/fe-fpm-writer/test/unit/filter.test.ts
```diff
@@ -0,0 +1,304 @@
...
+                name: 'NewCustomFilter2',
+                label: 'Test Custom Filter 2',
+                controlID: 'testID2',
+                property: 'Testing'
+            };
+            generateCustomFilter(testDir, filter, fs);
+
+            generateCustomFilter(testDir, secondFilter, fs);
+            expect(fs.readJSON(join(testDir, 'webapp/manifest.json'))).toMatchSnapshot();
+            expect(fs.exists(getControllerPath(secondFilter))).toBeTruthy();
```
tobiasqueck, line 75: This test is incorrect since it expects a new controller despite the input being `undefined`

tobiasqueck (changes requested):
Please see the inline comments. I have also pushed https://github.com/SAP/open-ux-tools/pull/1027 containing my local changes when testing your code locally.

packages/fe-fpm-writer/test/unit/filter.test.ts
tobiasqueck: Fixed with #1027

packages/fe-fpm-writer/templates/filter/Controller.ts
tobiasqueck: Fixed with #1027

packages/fe-fpm-writer/test/unit/filter.test.ts
tobiasqueck: Fixed with #1027

packages/fe-fpm-writer/src/filter/index.ts
tobiasqueck: Fixed with #1027

packages/fe-fpm-writer/templates/filter/Controller.js
tobiasqueck: Fixed with #1027

packages/fe-fpm-writer/src/filter/index.ts
tobiasqueck: Fixed with #1027

tobiasqueck (changes requested):
Please fix the code smells: https://sonarcloud.io/project/issues?id=SAP_open-ux-tools&pullRequest=985&resolved=false&types=CODE_SMELL otherwise
* changeset is correct
* code adheres to coding standards (except the code smells, of course)
* good test coverage

tobiasqueck (approved):

kranthie-sap (approved):
The new changes to the event handler, the filter writer and the tests look good. Changeset  with the right version update exists.

PR #991, opened 2023-04-25
fix(security): upgrade yaml

donal-tobin-sap:
Fix possible unhandled exception when a subset of parse document APIs are called.

- https://github.com/SAP/open-ux-tools/security/dependabot/9

Issue: tools-suite #24585


Klaus-Keller (commented):
Changeset missing, otherwise looks good.

donal-tobin-sap:
> Changeset missing, otherwise looks good.

Should be fixed now.
Changeset was there - it just had an error because of a duplicate line.
(Not just direct dependencies to be published, but transitive as well)

tobiasqueck (approved):
* changeset is ok
* pnpm-lock changes match the package.json changes (no surprises)
* didn't test locally but assuming the tests run through all is good

donal-tobin-sap:
Security Incident #2370054043

schreckstefan (approved):
The upgrade is required by
https://github.com/SAP/open-ux-tools/security/dependabot/9
https://github.com/SAP/open-ux-tools/security/dependabot/10
https://github.com/SAP/open-ux-tools/security/dependabot/11

Issue #993, opened 2023-04-27
BUG - project archive contains empty folders if in .gitignore


Klaus-Keller:
### Related Feature
Feature request: https://github.com/SAP/open-ux-tools/issues/713

### Description
If a project is archived using `@sap-ux/environment-check` we check for a `.gitignore` file to exclude files or folders. If `.gitignore` contains folders with `/`, they are included as empty folder in the archive, if it contains folders without `/` they are completely excluded.

Example `.gitignore`:
```
node_modules
```
Resulting archive e.g.:
<img width="207" alt="image" src="https://user-images.githubusercontent.com/66327622/234729221-282633aa-3cb2-433c-87f3-967995da4dfb.png">

Example `.gitignore`:
```
node_modules/
```

Resulting archive e.g.:
<img width="196" alt="image" src="https://user-images.githubusercontent.com/66327622/234729197-bc6bdbf3-bf2f-43a0-bc66-a78bca8759e9.png">

### Steps to Reproduce
Steps to reproduce the behavior with SAP Fiori tools - Application Modeler extension:
1. Open VSCode with a Fiori project that has a filled `node_modules` folder and a `.gitignore` which contains `node_modules/` 
2. Open command palette and select `Fiori: Archive Project`
3. Extract the resulting zip archive

### Expected results
No `node_modules` folder should be in the archive.

### Actual results
An empty `node_modules` folder is in the archive.

### Screenshots
See above

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

Klaus-Keller:
Fixed with https://github.com/SAP/open-ux-tools/pull/994

PR #994, opened 2023-04-27
fix(env-check): ignore empty folders in archive

Klaus-Keller:
See https://github.com/SAP/open-ux-tools/issues/993.

Adding option `mark: true` to `glob` does the trick, it adds `/` to directories when comparing: https://github.com/kaelzhang/node-ignore#2-filenames-and-dirnames

### DIFF
--- a/.changeset/small-clouds-raise.md
+++ b/.changeset/small-clouds-raise.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/environment-check': patch
+---
+
+Remove empty folders if ignored
--- a/packages/environment-check/src/archive/index.ts
+++ b/packages/environment-check/src/archive/index.ts
@@ -78,6 +78,7 @@ async function getFileList(cwd: string): Promise<string[]> {
         cwd,
         dot,
         ignore: ignore().add(ignores),
+        mark: true,
         skip
     });
     return files;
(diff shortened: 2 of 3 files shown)


theStiggySpd (approved):
- Changeset looks good
- Other relevant code change is good based on the explanation/link

Please note that I have not tested locally

slin-sap (changes requested):
- Function signal changed only to an internal function. Thus CS level patch is correct. 
- Documentation of using mark: true is linked in the PR. 

This change to 3rd party library changes the behavior of getFileList function. But there is no unit test capture this behavior change. This might give opportunity to introduce regression.

Klaus-Keller:
> * Function signal changed only to an internal function. Thus CS level patch is correct.
> * Documentation of using mark: true is linked in the PR.
> 
> This change to 3rd party library changes the behavior of getFileList function. But there is no unit test capture this behavior change. E.g. If glob lib introduced a regression and mark: true doesn't take effect, there is no unit test can capture the issue.

@slin-sap, you are right, we mock `archiver` and `glob-gitignore` in the unit tests of this module. There are checks, whether `glob-gitignore` is called with correct parameters: 
- https://github.com/SAP/open-ux-tools/pull/994/files#diff-20d5dd809ed4daae5f2e90e7e1921b9bdd21e5acf4e53552fd3900330ea1236eR68
- https://github.com/SAP/open-ux-tools/pull/994/files#diff-20d5dd809ed4daae5f2e90e7e1921b9bdd21e5acf4e53552fd3900330ea1236eR118

but I agree, having no integration tests to check whether the correct files are selected using `glob-gitignore` and later archived in .zip file using `archiver` is a gap and should be added. I did a small tests, but adding integration tests blows up this PR and I'd prefer to address this as a separate task.

slin-sap (approved):
Agree testing the glob lib correctness is more appropriate in an integration test.

PR #995, opened 2023-04-27
FPM writer: additional unit tests for contextPath 

schreckstefan:
Issue: https://github.com/SAP/open-ux-tools/issues/922

This PR provides additional unit tests, as required in https://github.com/SAP/open-ux-tools/pull/952#pullrequestreview-1401595864


tobiasqueck:
@schreckstefan what happened to this one? Is it obsolete or will you continue working on it?

schreckstefan:
@tobiasqueck 
@nlunets asked for it in my previous enhancement, thus I added the test and asked for a review.
Waiting for this since weeks...

nlunets:
At which point did you tag me ? :)

packages/fe-fpm-writer/test/unit/page/object.test.ts
```diff
@@ -131,6 +150,25 @@ describe('ObjectPage', () => {
...
         });
 
+        test('simple nested navigation with contextPath', () => {
+            const target = join(testDir, 'with-nested-nav');
+            fs.write(join(target, 'webapp/manifest.json'), testAppManifest);
+            generate(
+                target,
+                {
+                    ...minimalInput,
+                    contextPath: '/RootEntity/OtherEntity',
```
nlunets, line 160: if the navigation is called to_OtherEntity the contextPath also needs to be /RootEntity/to_OtherEntity

nlunets (changes requested):
this unit test is not 100% correct and i'm actually more worried about the part which generates the input you use for your test, where it that done ?

schreckstefan:
> 

@nlunets 
I corrected the test.

The logic for calculating the contextPath is not part of the FPM writer, it resides at the tools-suite:
image
It could be transferred to open-ux-tools, but I decided against doing the calculation inside the existing API, as it would require an incompatible enhancement of the given API functions: in order to detect the "mixed mode" at existing apps, carrying the entitySet at the parent node, I have to know the parent page,  plus potentially further information like the route pattern.
I could try to set up a separate helper function for the same, if you like. But would open a separate feature request for that.

Issue #997, opened 2023-04-28
BUG - UITranslationButton. Parameter of 'onShowExistingEntry'(property) function is not generic type.


815are:
### Description
When we use `UITranslationButton` and we have bundle's single entry like:
```
  key: {...}.
  value: {...},
  filePath: string;
```

It meats interface `TranslationEntry`
https://github.com/SAP/open-ux-tools/blob/main/packages/ui-components/src/components/UITranslationInput/UITranslationButton.types.ts#L32-L35

But when we attach method to `onShowExistingEntry`, then type of param is strict `TranslationEntry`.
In result we need apply cast:
![image](https://user-images.githubusercontent.com/90789422/234951739-6c1b96d0-eb41-4197-ac57-19d041872323.png)

### Expected results
Generic type is used which allows to receive enhanced type in param when `onShowExistingEntry` called

### Actual results
Strict type used as param when `onShowExistingEntry` called

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

PR #998, opened 2023-04-28
fix(ui-components): Fix generic type for translation input entry

815are:
Issue #997

Trying to improve interface of `UITranslationInput` to avoid casting like:
image

Root idea is that we can pass objeect like:
```
    key: {...}.
    value: {...},
    dummyProperty: string;
```
to `entries` property(interface of [Bundle](https://github.com/SAP/open-ux-tools/blob/main/packages/ui-components/src/components/UITranslationInput/UITranslationButton.types.ts#L32-L37)) 

In result once existing entry is found in bundle and user clicks on show button -> callback `onShowExistingEntry` returns reference to original object(with custom properties).
Current problem is that `onShowExistingEntry` is defined with strict type `TranslationEntry` as `onShowExistingEntry?: (entry: TranslationEntry) => void;`, but we need more generic type like:
```
export interface UITranslationProps<T extends TranslationEntry = TranslationEntry> {
    // When entry exists in passed i18n entries and user clicked on show entry button
    onShowExistingEntry?: (entry: T) => void;
```

### DIFF
--- a/.changeset/fuzzy-spoons-chew.md
+++ b/.changeset/fuzzy-spoons-chew.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui-components': patch
+---
+
+UITranslationInput. Generic type for "entries" property("I18nBundle" type) - generic type is used as parameter when "onShowExistingEntry" is called.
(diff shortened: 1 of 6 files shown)


Jimmy-Joseph19 (approved):
Changes improving translation types looks good. Also see the storybook is updated.
I have not tested locally.

Issue #1002, opened 2023-05-05
BUG - User is not prompted for credentials when deploy an ABAP app in test mode


longieirl:
### Description
Found when testing https://github.com/SAP/open-ux-tools/issues/687

Usingh the following CLI command;
```
./bin/deploy --name ZJLFFLDEC1 --description 'EC1 Deployment JLong Cataog' --package '$TMP' --archive-folder /Users/Documents/packages/testjl/dist --url http://mytest.server:50000 --client 100 --verbose true --test true
```

The command is returning a notification the deployment has completed but is not showing any of the ABAP errors being returned nor is the flow prompting for a username and password. Removing the `--test` flag and re-run the steps prompts the user for their credentials.

The app was created as part of a deployment test where an app with a different ID in the same ABAP repository was already deployed.

![Screenshot 2023-05-05 at 13 11 06](https://user-images.githubusercontent.com/4520960/236455048-e37ad573-4a0b-41ff-9fd8-9e29ad4e7228.png)

### Expected results
User is prompted for their credentials and any ABAP errors to be shown in the console output.

### Actual results
User is not prompted for their credentials and none of the ABAP errors returned in the response are shown to the user.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Issue with credentials is now resolved, tested locally and is working as expected. 
The error logging for test mode needs to be handled better, additional PR created with changes required. 
Can be misleading when there is an error logged to the console during test mode. Test mode does return a 403 error but its the details of `errordetails` in the error response that needs to be displayed based on the log level specified in each error detail node.

PR #1004, opened 2023-05-05
Fix/1002/safemode not handling invalid credentials

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/1002

- additional test cases added and updated to reflect change in behaviour
- additional check in `deploy.ts` to handle credentials returning an empty object. This prevents the user from using ctrl+c the flow exist the process
- Based on local testing, `408` response code was appended as another HTTP code to retry deployment

Handles a host requesting credentials;
Screenshot 2023-05-05 at 14 47 07

Shows the ABAP errors in test mode;
Screenshot 2023-05-05 at 14 47 14

### DIFF
--- a/packages/axios-extension/src/abap/ui5-abap-repository-service.ts
+++ b/packages/axios-extension/src/abap/ui5-abap-repository-service.ts
@@ -1,5 +1,5 @@
 import type { AxiosResponse, AxiosRequestConfig } from 'axios';
-import { prettyPrintError, prettyPrintMessage } from './message';
+import { ErrorMessage, prettyPrintError, prettyPrintMessage } from './message';
 import { ODataService } from '../base/odata-service';
 import { isAxiosError } from '../base/odata-request-error';
 
@@ -122,11 +122,18 @@ export class Ui5AbapRepositoryService extends ODataService {
      * @returns application information or undefined
      */
     public async getInfo(app: string): Promise<AppInfo> {
+        console.log(`getInfo>>>> `);
         try {
             const response = await this.get<AppInfo>(`/Repositories('${encodeURIComponent(app)}')`);
+            console.log(`getInfo>>>> return data`);
             return response.odata();
         } catch (error) {
-            this.log.debug(`Retrieving application ${app}, error found ${error}`);
+            console.log(`getInfo>>>> error ${JSON.stringify(error)}`);
+            this.log.debug(`Retrieving application ${app}, ${error}`);
+            if (isAxiosError(error) && error.response?.status !== 404) {
+                throw error;
+            }
+            console.log(`getInfo>>>> return undefined`);
             return undefined;
         }
     }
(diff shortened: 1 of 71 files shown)


packages/axios-extension/src/abap/ui5-abap-repository-service.ts
```diff
@@ -126,7 +127,10 @@ export class Ui5AbapRepositoryService extends ODataService {
             const response = await this.get<AppInfo>(`/Repositories('${encodeURIComponent(app)}')`);
             return response.odata();
         } catch (error) {
-            this.log.debug(`Retrieving application ${app}, error found ${error}`);
+            this.log.debug(`Retrieving application ${app}, ${error}`);
+            if (isAxiosError(error) && error.response?.status !== 404) {
```
tobiasqueck, line 131: We only want to return `undefined` if it is a 404 otherwise, we don't know what to do with the error, therefore, it should be 
```
if (isAxiosError(error) && error.response?.status === 404) {
   return undefined;
} else {
   throw error;
}
```

packages/axios-extension/src/abap/ui5-abap-repository-service.ts
```diff
@@ -158,12 +162,20 @@ export class Ui5AbapRepositoryService extends ODataService {
...
-                : '';
-            this.log.info(`App available at ${frontendUrl}${path}${query}`);
+            if (!testMode) {
+                // log url of created/updated app
+                const path = '/sap/bc/ui5_ui5' + (!bsp.name.startsWith('/') ? '/sap/' : '') + bsp.name.toLowerCase();
+                const query = this.defaults.params?.['sap-client']
+                    ? '?sap-client=' + this.defaults.params['sap-client']
+                    : '';
+                this.log.info(`App available at ${frontendUrl}${path}${query}`);
+            } else if (testMode) {
```
tobiasqueck, line 172: the 2nd if is redundant (i.e no need to check `testmode` again, it must be truthy). 
what does prettyPrintError do if there is no error?

packages/axios-extension/src/abap/ui5-abap-repository-service.ts
```diff
@@ -326,7 +338,7 @@ export class Ui5AbapRepositoryService extends ODataService {
                 return response;
             }
         } catch (error) {
-            if (error?.response?.status === 504) {
+            if ([504, 408].includes(error?.response?.status)) {
```
tobiasqueck, line 341: Could you please add a concise comment in the code why we check for 504 and 408

tobiasqueck (changes requested):

packages/axios-extension/src/abap/ui5-abap-repository-service.ts
longieirl: Please refer to the 2nd screenshot above, it prints the error nodes returned.

tobiasqueck (approved):
* coding adheres to our standards
* open issues have been addressed
* test coverage is _barely_ acceptable, but added tests cover the changed functionality well
* changeset is ok

Issue #1007, opened 2023-05-08
TBI - project-access: Add support for multiple project types/artifacts


Klaus-Keller:
### Description
The current implementation of search functions in `@sap-ux/project-access` is restricted to Fiori elements and Fiori freestyle applications. There is also a demand to find extension projects, adaptation projects, and UI5 libraries/modules.

### Value
Central functionality to find different Fiori artifacts

### Architecture Elaboration
If we just add new functions to search for other Fiori artifacts, we might end up traversing the workspace multiple times. To avoid this, we should provide a find function that allows to collects multiple files, extensions, or folders. 

To classify `applications`, `adaptations`, `extensions`, and `libraries`, following criteria is used

| Artifact type | `manifest.json` | `['sap.app'].type` in manifest | `{approot}/.extconfig.json` | `manifest.appdescr_variant` |
| -- | :--: | :--: | :--: | :--: |
| Application | ✓ | `application` | ✗ | ✗ |
| Adaptation Project | ✗ | ✗ | ✗ | ✓ |
| Extension Project | ✓ | `application` | ✓ | ✗ |
| Library | ✓ | `library` | ✗ | ✗ |

### Tasks
- [ ] Provide function to search for multiple files, extensions, folders
- [ ] Provide functionality to search for Fiori artifacts

tobiasqueck:
> {approot}/.adp/config.json

That is design time file from WebIDE. So, it is not required that it exists in every adaption project. I would check for a `manifest.appdescr_variant` in the root of the `webapp` which is as required for an adaptation project as the manifest is for an application.

devinea:
> > {approot}/.adp/config.json
> 
> That is design time file from WebIDE. So, it is not required that it exists in every adaption project. I would check for a `manifest.appdescr_variant` in the root of the `webapp` which is as required for an adaptation project as the manifest is for an application.

`{approot}/.adp/config.json` is used in BAS and written during migration.
It looks the from the US https://github.wdf.sap.corp/ux-engineering/tools-suite/issues/14968 that `.che/project.json` was the old name in WebIDE.
Since the `{approot}/.adp/config.json` is added during migration we know it is a project that is supported by Fiori tools and not an old project that has not been migrated yet. Using the `manifest.appdescr_variant` alone does not help distinguish a project we support or not. Also it is presumably faster to find from the `{approot}` than searching the webapp folder name and checking for `manifest.appdescr_variant` in it.

tobiasqueck:
[...]

PR #1009, opened 2023-05-11
feat(project-access): introduce search for multiple Fiori artifacts

Klaus-Keller:
### Ticket: 
https://github.com/SAP/open-ux-tools/issues/1007

### Description
Introduce functionality to search folders, usually the workspace root folders, for Fiori artifacts like applications, adaptation projects, extension projects, and libraries.


packages/project-access/src/project/search.ts
```diff
@@ -227,3 +263,145 @@ export async function findAllApps(wsFolders: WorkspaceFolder[] | string[] | unde
...
+    }
+    if (options.artifacts.includes('adaptations')) {
+        results.adaptations = await filterAdaptations(pathMap);
+    }
+    if (options.artifacts.includes('extensions')) {
+        results.extensions = await filterExtensions(pathMap);
+    }
+    if (options.artifacts.includes('libraries')) {
+        results.libraries = await filterLibraries(pathMap);
+    }
```
devinea, line 405: May want to run these searches in parallel using ` Promise.all()` or similar.

devinea (dismissed):
Thanks @Klaus-Keller 
Code changes and tests look like they cover the require scenarios. 
I haven't tested locally. 
One nice to have suggestion

packages/project-access/src/project/search.ts
Klaus-Keller: @devinea 
While I did this in some other places for HTTP request and could significantly gain performance by running requests in parallel, I've tried it for the search for apps by running the search in each workspace root parallel. I couldn't measure significant performance improvements, my assumption is that parallelized disk access doesn't really help as the disk is the single resources and access is sequential. Do you have different experience? I can easily change the code.

devinea: @Klaus-Keller I don't have experience on parallel disk access. It just looked like code that could run in parallel.
But if there is no obvious benefit then I am fine with it as is. Cheers

.changeset/purple-phones-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/project-access': minor
+---
+
+Add findFioriArtifact() to allow searching for apps, adaptation projects, extension projects, and libraries
```
slin-sap, line 5: New API function name is findFioriArtifacts(). Missing a 's' in the cset.

slin-sap (changes requested):

.changeset/purple-phones-yawn.md
Klaus-Keller: @slin-sap, thanks, fixed it.

devinea (approved):
Update lg

slin-sap (approved):
- Since new API is added, cset with minor version bump looks good.
- New API allows consumer to flexibly filter Fiori projects list based on project types. 
- New API function has tsdoc and unit test with clear usage instructions
- Change is backward compatible to filter Fiori apps.

PR #1013, opened 2023-05-15
Update TypeScript to v5 in TS templates.

devinea:
Implements https://github.com/SAP/open-ux-tools/issues/1198
- Update TypeScript to v5 in TS templates
- Add outDir to tsconfig again to fix error shown inline in JS files in VSCode


IainSAP (approved):
OK for fe/ff writers no unexpected changes. Not tested locally.

zdravko-georgiev (approved):

Issue #1015, opened 2023-05-15
TBI - project-access: Add support for getting a CAP project environment config


voicis:
### Description (include screenshots)
In CAP project users can override defaults in project specific configuration files and the tools should honor this in a consistent way. To achieve this we should have a wrapper function for `@sap/cds` calls that retrieves environment config and could be reused across the tools.

### Value
A common way to get a CAP project environment configuration.

### Tasks
- [ ] Create a reuse function that alls `cds.env.for`

PR #1016, opened 2023-05-15
feat(project-access): get CAP environment config

voicis:
### Ticket: 
#1015 

### Description
Introduce new utility function for getting a CAP project environment.

### DIFF
--- a/packages/project-access/src/project/cap.ts
+++ b/packages/project-access/src/project/cap.ts
@@ -1,6 +1,6 @@
 import { join } from 'path';
 import { FileName } from '../constants';
-import type { CapCustomPaths, CapProjectType, csn, Package } from '../types';
+import type { CapCustomPaths, CapProjectType, CdsEnvironment, csn, Package } from '../types';
 import { fileExists, readJSON } from '../file';
 import { loadModuleFromProject } from './module-loader';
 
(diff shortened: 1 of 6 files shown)


Klaus-Keller (approved):
- code change looks good and easy to understand
- test coverage is great
- build locally and tested using
    ```Shell
    packages/project-access/dist$ node -e 'require("./index").getCapEnvironment("PATH_TO_CAP").then(env=>console.log(env))'
    ````
    and could not find any issue.

Approved from my side.

packages/project-access/src/project/cap.ts
```diff
@@ -96,3 +95,15 @@ export async function getCapModelAndServices(
...
 }
+
+/**
+ * Get CAP CDS project environment config for project root.
+ *
+ * @param capProjectPath - project root of a CAP project
+ * @returns - environment config for CAP project
+ */
+export async function getCapEnvironment(capProjectPath: string): Promise<CdsEnvironment> {
+    const cds = await loadModuleFromProject<any>(capProjectPath, '@sap/cds');
```
tobiasqueck, line 106: `unknown` would be the better type here.

tobiasqueck (changes requested):
small comment regarding type usage

packages/project-access/src/project/cap.ts
voicis: If I use `unknown` then we get compiler error in the next line. We would need a specific type instead, I suppose we could do that as well.

tobiasqueck: Do you know the type, then yes, please. Based on the next line, I assumed you don't know exactly what you get.

tobiasqueck (approved):
* use case and corresponding changeset is clear
* code adheres to coding standards and is properly tested
* did not manually test it

Klaus-Keller (approved):
I re-reviewed the newly added type, looks good to me.
Again approved from my side.

Issue #1018, opened 2023-05-17
BUG - archive name with timestamp producing archive with timestamp appended


longieirl:
### Description
Found when testing https://github.com/SAP/open-ux-tools/issues/687

Appending a timestamp to the archive name should be a env/config variable. Archives generated with the timestamp will cause issues for existing clients who are not expecting or wanting to generate archives with a new naming convention.

### Steps to Reproduce
Generate a project for ABAP deployment

```
./bin/deploy --name ZJLONG0205 --description 'Deployment Cataog' --package '$TMP' --archive-folder /Users/testprojectjl0205/dist --url http://testsite.com:50000 --client 100 --verbose true --test true
```

### Expected results
The archive generated is `archive.zip`.

### Actual results
Generated archive is `archive-1684321871711.zip`.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
## Root Cause Analysis 

### Problem
Then the keep flag was enabled, the generated archive appends a timestamp.

### Fix
Remove the timestamp from the name of the arhive name.

### Why was it missed
Issue with existing flow which assume the generated archive is always archive.zip. This will support existing behaviour in third party tools.

### How can we avoid this
Continue testing and ensuring existing behaviour is maintained.

PR #1019, opened 2023-05-17
fix(deploy-config): Remove timestamp from archive.zip

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/1018

- update name archive.zip
- applied test cases


devinea (approved):
@longieirl 
Changes and tests look good. Thanks

longieirl:
> What happens if `archive.zip` exists and `keep` is set to `true`? Will it be overwritten, will there be an error or ... ?

Yes, its ovewritten.

tobiasqueck (approved):
* changeset and code are ok
* I am not in favor of this change but understand it is required for backward compatibility with the Fiori tools

Issue #987, opened 2023-04-21
BUG - Generated Controller Extensions do not always work with TS


tobiasqueck:
### Related Feature
Feature request: #263

### Description
When generating a controller extension in typescript using the `fe-fpm-writer` then the static property `override` is used. This only works if you also add a “[addControllerStaticPropsToExtend](https://github.com/ui5-community/babel-plugin-transform-modules-ui5#class-conversion)” in the babel config which has some unwanted side-effects.
To solve this problem UI5 `1.112` and above supports defining the `overrides` property which is also supported with the ui5 babel plugin since https://github.com/ui5-community/babel-plugin-transform-modules-ui5/pull/82. In addition, https://github.com/ui5-community/babel-plugin-transform-modules-ui5/pull/89 allows to also make use of it in older UI5 versions.
Thus, to solve the issue, the `fe-fpm-writer` should change the template to use `overrides` and the `ui5-application-writer` should use the latest version of [babel-plugin-transform-modules-ui5](https://github.com/ui5-community/babel-plugin-transform-modules-ui5) and set `overridesToOverride: true` in the babel config. I would also recommend to update to the latest version of `ui5-tooling-transpile` in the same fix because in its later versions it makes directly use of the babel config so that preview and transpilation results are consistent.

### Steps to Reproduce
1. Create a FE typescript project
2. Remove `addControllerStaticPropsToExtend` 
3. Add a controller extension with the fiori tools
4. Preview your app
5. Your controller extension is not executed

### Expected results
 Your controller extension is executed

### Actual results
 Your controller extension is not executed

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
The problem could not be fixed at the time of the feature implementation.

### Fix
Update configurations and dependencies to 3rd party modules solving the issue

### Why was it missed
The problem could not be fixed at the time of the feature implementation.

### How can we avoid this
Impossible

PR #1020, opened 2023-05-17
BUG - Generated Controller Extensions do not always work with TS

devinea:
#987 BUG - Generated Controller Extensions do not always work with TS


packages/ui5-application-writer/templates/optional/typescript/.babelrc.json
```diff
@@ -4,7 +4,8 @@
         [
             "transform-ui5",
             {
-                "addControllerStaticPropsToExtend": true
+                "addControllerStaticPropsToExtend": true,
```
tobiasqueck, line 7: `addControllerStaticPropsToExtend` shouldn't be required anymore if we use `overridesToOverride`

tobiasqueck (changes requested):

packages/ui5-application-writer/templates/optional/typescript/.babelrc.json
devinea: Thanks. Updated in https://github.com/SAP/open-ux-tools/pull/1020/commits/9838a2cd8c9277431da31417abb374eedd3c1355

tobiasqueck (approved):
* changeset is fine
* only config changes but they are as expected
* didn't test manually but I assume that has been done

IainSAP (approved):
LGTM

- Updates as expected

Issue #983, opened 2023-04-18
FEATURE - create eslint plugin module


docirl:
_As a developer,
I want to have eslint config provided in a reusable plugin,
so that I can use it in my project to provide a standard and consistent config for eslint in my project_

### Description

Create a eslint plugin containing shared config that can be used in projects for eslint config. This approach is being used to limit the size and number of lines of the eslint file, and to make it easier to consume the config from a reusable package.

The initial version of this module will contain config for typescript projects. The shareable config could be expanded later to include eslint config for javascript projects also.

### Technical Design
Create a new plugin that wraps and exposes the eslint config

### Acceptance Criteria
Given I generate a new project with the generators, 
when I choose to enable typescript in the project 
then the generated project will have the new eslint config module included as dependancy and extended in the eslint file in the project.

### Tasks
- [ ] Create a new eslint plugin that wraps and exposes the eslint config
- [ ] Update typescript templates to consume this new module
- [ ] Unit Tests updated
- [ ] Test Cases defined

PR #1026, opened 2023-05-23
Use  @sap-ux/eslint-plugin-fiori-tools for JS projects

devinea:
https://github.com/SAP/open-ux-tools/issues/983

Use @sap-ux/eslint-plugin-fiori-tools for JS projects


tobiasqueck (approved):
* changes are fine
* changeset is ok (I approved) BUT it would be better to actually split it in two
   * eslint: the one you have
   * ui5-app-writer: fixing the incorrectly used export for ts and enabling it for js

tobiasqueck (approved):
* changes are fine
* changeset is ok (I approved) BUT it would be better to actually split it in two
   * eslint: the one you have
   * ui5-app-writer: fixing the incorrectly used export for ts and enabling it for js

IainSAP (approved):
LGTM

- Updated as expected

devinea:
Thanks @tobiasqueck @IainSAP

PR #1028, opened 2023-05-24
fix: remove superfluous checks in template

tobiasqueck:
Removing superfluous checks in the template that make it unnecessarily complicated.
The code was detected while reviewing https://github.com/SAP/open-ux-tools/pull/985

### DIFF
--- a/packages/fe-fpm-writer/templates/common/EventHandler.js
+++ b/packages/fe-fpm-writer/templates/common/EventHandler.js
@@ -4,7 +4,7 @@ sap.ui.define([
     'use strict';
 
     return {
-        <%- (typeof eventHandlerFnName !== 'undefined' && eventHandlerFnName) || 'onPress' %>: function(oEvent) {
+        <%- eventHandlerFnName %>: function(oEvent) {
             MessageToast.show("Custom handler invoked.");
         }
     };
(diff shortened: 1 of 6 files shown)


815are (dismissed):
Checked:
1. Code looks good - code in templates is simplified and tests adjusted;
2. I did not find regression, as checked tests - we still can pass custom event handler name
https://github.com/SAP/open-ux-tools/blob/fpm-writer-event-fix/packages/fe-fpm-writer/test/unit/action.test.ts#L257
https://github.com/SAP/open-ux-tools/blob/fpm-writer-event-fix/packages/fe-fpm-writer/test/unit/__snapshots__/section.test.ts.snap#L16
https://github.com/SAP/open-ux-tools/blob/fpm-writer-event-fix/packages/fe-fpm-writer/test/unit/__snapshots__/action.test.ts.snap#L131
 and default is `onPress` like https://github.com/SAP/open-ux-tools/blob/fpm-writer-event-fix/packages/fe-fpm-writer/test/unit/__snapshots__/section.test.ts.snap#L58
3. I did not test locally with extension

devinea (approved):
changeset, code changes and tests updates are appropriate for the change.
Thanks @tobiasqueck

devinea (approved):

815are (approved):

Issue #1024, opened 2023-05-22
TBI - Add onClick method to UIActionCallout


cianmSAP:
### Description (include screenshots)
Under the current implement of [UIActionCallout](https://github.com/SAP/open-ux-tools/blob/main/packages/ui-components/src/components/UIActionCallout/.tsx) the commandAction is used to open extensions in VSCode **or** open the link in the broswer. 

In addition to this, it would be valuable to have a callback in order to track when the click is event is fire in all cases. The proposal is to add an onClick method.

### Value
onClick method contains a callback (can be used to track telemetry events for example) 

### Architecture Elaboration
No

### Tasks
- [ ] Add onClick method to UIActionCallout

IainSAP:
@cianmSAP The proposed fix is a little confusing (having a flag to prevent multiple click events), can we investigate using `stopPropogation` as that is its intended use. Also it would be useful to pass the action detail in the click callback if available for consumers.

PR #1032, opened 2023-05-24
fix(ui-component): fix onclick logic for ui action callout component

cianmSAP:
#1024

Currently the `this.anchor.current?.click();` triggers the onClick twice in some cases.

Solution is to add the `anchorClicked` flag to track whether the anchor tag was clicked or not. When the onCalloutClick function is called, it checks the flag before triggering the click event on the anchor.


Klaus-Keller (approved):
- Code to prevent multiple invocation of click looks good
- Test coverage is great
- Did not build and test locally

Approved from my side.

devinea (approved):
Updates are expected to fix the bug. 
Thanks

PR #1036, opened 2023-05-25
Ui5 transpile updates

### LINKED ISSUE #1035: BUG - incorrect settings for ui5-middleware-transpile
### Related Feature
Feature request: #621

### Description
The latest version of the `ui5-middleware-transpile` is using the `.bablrc` if it exists and ignores settings from the `ui5.yaml`, however, our generated projects contain a mix of both.

### Steps to Reproduce
1. Generate a UI5 typescript application
2. Run `npm run build`
3. See warning in logs

### Expected results
No warnings and correctly transpiled files.

### Actual results
Warning in the log, and transpiled files contain `await` statements (amongst other issues)

### Screenshots

OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

tobiasqueck:
The latest version of the `ui5-middleware-transpile` is using the `.bablrc` if it exists and ignores settings from the `ui5.yaml`, however, our generated projects contain a mix of both.
This PR 
* moves all settings into the `.babelrc` and thereby fixes #1035
* updates the patch version of the `ui5-tooling-transpile`
* applies consistent indentation

### DIFF
--- a/.changeset/clever-pants-pull.md
+++ b/.changeset/clever-pants-pull.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui5-application-writer': patch
+---
+
+Fix: incorrect settings for ui5-middleware-transpile
--- a/packages/fiori-elements-writer/test/__snapshots__/feop.test.ts.snap
+++ b/packages/fiori-elements-writer/test/__snapshots__/feop.test.ts.snap
@@ -3754,8 +3754,6 @@ server:
       afterMiddleware: compression
       configuration:
         debug: true
-        transpileAsync: true
-        transpileTypeScript: true
         excludePatterns:
           - /Component-preload.js
     - name: fiori-tools-proxy
@@ -3781,9 +3779,6 @@ builder:
       afterTask: replaceVersion
       configuration:
         debug: true
-        removeConsoleStatements: true
-        transpileAsync: true
-        transpileTypeScript: true
 ",
     "state": "modified",
   },
@@ -3818,8 +3813,6 @@ server:
       afterMiddleware: compression
       configuration:
         debug: true
-        transpileAsync: true
-        transpileTypeScript: true
         excludePatterns:
           - /Component-preload.js
     - name: sap-fe-mockserver
@@ -3838,9 +3831,6 @@ builder:
       afterTask: replaceVersion
       configuration:
         debug: true
-        removeConsoleStatements: true
-        transpileAsync: true
-        transpileTypeScript: true
 ",
     "state": "modified",
   },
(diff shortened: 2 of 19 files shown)


petermuessig:
@tobiasqueck : from my POV this looks good, but the indentation drives me crazy!

Insider: and you also create configuration for Babel 😉

packages/fiori-elements-writer/test/__snapshots__/feop.test.ts.snap
```diff
@@ -3689,30 +3702,32 @@ archive.zip
...
-    \\"strictPropertyInitialization\\": false,
-    \\"moduleResolution\\": \\"node\\",
-    \\"rootDir\\": \\"webapp\\",
-    \\"outDir\\": \\"./dist\\",
-    \\"baseUrl\\": \\"./\\",
-    \\"paths\\": {
-      \\"fefeop2ts/*\\": [
-        \\"webapp/*\\"
-      ]
+    \\"compilerOptions\\": {
```
petermuessig, line 3705: We also polished the `tsconfig.json` in our latest example:
https://github.com/SAP-samples/ui5-typescript-helloworld/blob/main/tsconfig.json

* `es2022` supported by Tooling V3, 
* some config is useless when Babel is used, e.g. `inlineSourceMaps`, `outDir`
* if `strict` is `true` all strict rules are true and thus `strictNullChecks` is redundant defined here
* if `types` are used, you may also use `typeRoots` to point to the definitely typed types like in the hello world.
* the `composite` flag in the helloworld sample is not needed (still a rest from the cleanup)

tobiasqueck: @petermuessig https://github.com/SAP/open-ux-tools/pull/1036/commits/065699ecee2bafc20d6113da46b9edcdfa47f88a takes your suggestions into account

.changeset/clever-pants-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/ui5-application-writer': patch
+---
+
+Fix: incorrect settings for ui5-middleware-transpile
```
devinea, line 5: Since other modules are test changes it is fine not to release them

devinea (approved):
template updates look good and snapshots are expected. 
lock file change is not related but is ok to include. 

Thanks @tobiasqueck

.changeset/clever-pants-pull.md
tobiasqueck: I think they will be automatically released as well due to the dependency (even if not needed) I think

petermuessig (approved):

Issue #270, opened 2022-01-20
TBI - Clean up eslint warnings in generated projects


devinea:
### Description (include screenshots)
Improve quality of the generated projects. Add missing JSDoc and fix eslint warnings/errors on each floorplan.
Follow on from #254

### Value
Optional - details on how this technical backlog item provides value to end-users or dev team eg. Performance, Quality, Maintainability. This will help determine the ranking in relation to feature development

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [x] Add test for each floor plan with eslint and code assist enabled.
- [ ] Add missing JSDocs
- [ ] Fix eslint issues.

FYI @tobiasqueck

tobiasqueck:
@devinea any updates on this one?

tobiasqueck:
@devinea same question as in February :)

devinea:
@IainSAP You took this over and merged the PR. Is there anything remaining?

PR #1038, opened 2023-05-25
TBI - Clean up eslint warnings in generated projects

devinea:
https://github.com/SAP/open-ux-tools/issues/270


tobiasqueck (approved):
* reviewed changes:
   * single quotes to double
   * additional rule to ignore
   * snapshots
* no code changes to review only style and these are as expected
* didn't test locally if something broke (but I wouldn't know how)

PR #1043, opened 2023-05-26
fix: cds dynamic import

voicis:
There is an issue when loading `@sap/cds` module dynamically, because some of it's API's are exposed as getters on a class, which does not work well with transpiled import from typescript. It uses `Object.prototype.hasOwnProperty` which returns `false` for such getters and thus causes errors in runtime.

However, it also exposes the whole object as `default` property and this seems to be safe to use.

### DIFF
--- a/.changeset/lemon-feet-crash.md
+++ b/.changeset/lemon-feet-crash.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': patch
+---
+
+properly handle "@sap/cds" dynamic import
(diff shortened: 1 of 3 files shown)


Klaus-Keller (approved):
- Code change is easy to understand, reason is well explained in description
- Test coverage is great
- I did not build and test this locally

Approved from my side.

devinea (approved):
Code, tests and change set looks good. Thanks @voicis

PR #1046, opened 2023-05-30
fix: make icon positioning consistent in searchbox inputs

oliversberzs:
* Fixed the searchbox icon moving slightly before disappearing on focus


Klaus-Keller (changes requested):
Hi @oliversberzs,

I've checked the changes in storybook and after your change the search icon (magnifier) looks slightly misplaced to the top. See this animated gif, it shows after and then before the change. The one in which the magnifier is more to the top is after:



Perhaps the consuming app has styles that causes misalignment of the search icon?

Klaus-Keller (approved):
- style changes look good
- tested locally, no glitch on icon animation when search box get's focus after the fix

Approved from my side.

Klaus-Keller (approved):
- style changes look good
- tested locally, no glitch on icon animation when search box get's focus after the fix

Approved from my side.

unseen1980 (approved):

Issue #1051, opened 2023-06-01
BUG - ui5-proxy-middleware crashes on startup if the cwd is on the root of the project


tobiasqueck:
### Related Feature
Feature request: _issue number_

### Description
If you have a project with the `ui5-proxy-middleware` configured in the `ui5.yaml` and you start it from the outside e.g. `ui5 server --config ./my-project/ui5.yaml` then the middleware crashes when trying to find the `manifest.json`.

### Steps to Reproduce
1. Generate a project with the SAP Fiori generator
2. (install the ui5 cli globally)
3. Run `ui5 serve` outside the project directory but pointing to the config inside

### Expected results
The preview starts as expected,

### Actual results
The middleware crashes because it cannot find the `manifest.json` to read the `minUI5Version`

### Expected fix
This issue requires two fixes
1. Don't rely on the `cwd` but use the ui5 tooling provided `fs` to find project files
2. Even if a `manifest.json` is not found e.g. for an adaptation project, the middleware should not crash but just assume the latest UI5 version

### Additional Information
Conversation with @petermuessig 

> The problematic code is the following here which just relies on the current working directory as you also mentioned above:
https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-proxy-middleware/src/base/utils.ts#LL199C11-L199C22
> 
> If your middleware would work with the UI5 tooling version 3.0 only, you could run the following code to determine the project root in your custom middleware:
> ```js
> middlewareUtil.getProject().getRootPath()
> ```
> If you need to support the old and the newer version in parallel having a specVersion < 3.0 then you may do the same workaround as I did in the transpile middleware / task:
> https://github.com/ui5-community/ui5-ecosystem-showcase/blob/main/packages/ui5-tooling-transpile/lib/util.js#L291-L307
>
> This function requires the project root collection to be passed, e.g.:
https://github.com/ui5-community/ui5-ecosystem-showcase/blob/main/packages/ui5-tooling-transpile/lib/middleware.js#L31

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

[...]

PR #1054, opened 2023-06-01
Fix: middleware crashes when manifest.json is missing or not where it was expected

tobiasqueck:
Fix for #1051


packages/ui5-proxy-middleware/src/base/ui5.d.ts
```diff
@@ -0,0 +1,118 @@
...
+        getBuffer(): Promise<Buffer>;
+
+        /**
+         * Gets a string with the resource content.
+         */
+        getString(): Promise<string>;
+    }
+
+    /**
+     * https://sap.github.io/ui5-tooling/stable/api/@ui5_fs_ReaderCollection.html
```
ullasholla, line 23:
```suggestion
     * https://sap.github.io/ui5-tooling/v3/api/@ui5_fs_ReaderCollection.html
```
Broken link. This is the v3 link

packages/ui5-proxy-middleware/src/base/ui5.d.ts
```diff
@@ -0,0 +1,118 @@
...
+        byGlob(virPattern: string | string[], options?: object): Promise<Resource[]>;
+
+        /**
+         * Locates resources by matching a given path.
+         */
+        byPath(virPattern: string | string[], options?: object): Promise<Resource[]>;
+    }
+
+    /**
+     * https://sap.github.io/ui5-tooling/stable/api/module-@ui5_fs.DuplexCollection.html
```
ullasholla, line 38:
```suggestion
     * https://sap.github.io/ui5-tooling/v3/api/@ui5_fs_DuplexCollection.html
```
Broken link. Here is the v3 link

packages/ui5-proxy-middleware/src/base/ui5.d.ts
```diff
@@ -0,0 +1,118 @@
...
+     */
+    export class DuplexCollection extends ReaderCollection {}
+
+    /**
+     * https://sap.github.io/ui5-tooling/api/module-@ui5_fs.AbstractReader.html
```
ullasholla, line 43:
```suggestion
     * https://sap.github.io/ui5-tooling/v3/api/@ui5_fs_AbstractReader.html
```
Broken link. Replacement v3 link

packages/ui5-proxy-middleware/src/base/ui5.d.ts
```diff
@@ -0,0 +1,118 @@
...
+        /**
+         * Middleware utilities (not yet used in our middlewares)
+         */
+        middlewareUtils: unknown;
+    }
+}
+
+declare module '@ui5/builder' {
+    /**
+     * https://sap.github.io/ui5-tooling/api/module-@ui5_builder.tasks.TaskUtil.html
```
ullasholla, line 78:
```suggestion
     * https://sap.github.io/ui5-tooling/v3/api/@ui5_project_build_helpers_TaskUtil.html
```
Broken link. v3 replacement link

packages/ui5-proxy-middleware/src/base/ui5.d.ts
```diff
@@ -0,0 +1,118 @@
+declare module '@ui5/fs' {
```
ullasholla, line 1: I understand why we need to have our own types declaration file. @tobiasqueck, do you know if the UI5 libraries have plans to supply the types in the future? Does it make sense to contribute to their repos?

packages/ui5-proxy-middleware/src/base/utils.ts
```diff
@@ -195,36 +195,40 @@ export const setHtmlResponse = (res: any, html: string): void => {
...
  *
  * @param args list of runtime args
+ * @param manifest optional already loaded manifest.json
  * @returns The minUI5Version from manifest.json or undefined otherwise
  */
-export async function getUI5VersionFromManifest(args: string[]): Promise<string | undefined> {
-    const manifest = await getManifest(args);
-    return manifest['sap.ui5']?.dependencies?.minUI5Version;
+export async function getUI5VersionFromManifest(args: string[], manifest?: Manifest): Promise<string | undefined> {
+    manifest ??= await getManifest(args);
```
ullasholla, line 219: We should probably avoid mutating the passed in `manifest` parameter here. It doesn't actually change the `manifest` variable in the calling code and is confusing. If the intent is to fetch the manifest if it's missing, it's better to explicitly return it?

packages/ui5-proxy-middleware/src/base/utils.ts
```diff
@@ -236,7 +240,7 @@ export async function resolveUI5Version(version?: string, log?: ToolsLogger): Pr
         ui5Version = version ? version : '';
         ui5VersionLocation = getYamlFile(process.argv);
     } else {
-        const minUI5Version = await getUI5VersionFromManifest(process.argv);
+        const minUI5Version = await getUI5VersionFromManifest(process.argv, manifest);
         ui5Version = minUI5Version && !isNaN(parseFloat(minUI5Version)) ? minUI5Version : '';
```
ullasholla, line 244: I know this PR did not touch this line, but for such a simple piece of code this line is not very readable :)
`Number.isNan()` is a more robust version of the global `isNaN()` (source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN)

packages/ui5-proxy-middleware/src/ui5/middleware.ts
```diff
@@ -46,13 +49,42 @@ function createRequestHandler(routes: { route: string; handler: RequestHandler }
...
-module.exports = async ({ options }: MiddlewareParameters<Ui5MiddlewareConfig>): Promise<RequestHandler> => {
+/**
+ * Search the project for the manifest.json.
+ *
+ * @param rootProject @ui5/fs reader collection with access to the project files
+ * @returns manifest.json as object or undefined if not found
+ */
+async function loadManifest(rootProject: ReaderCollection): Promise<Manifest | undefined> {
+    const files = await rootProject.byGlob('**/manifest.json');
+    if (files.length > 0) {
```
ullasholla, line 60: From the documentaton of [`byGlob()`](https://sap.github.io/ui5-tooling/v3/api/@ui5_fs_ReaderCollection.html), it's not clear to me if it returns an empty array or undefined when there's no match. Also, do we need a `try/catch` block because of the `await` statement?

packages/ui5-proxy-middleware/test/ui5/middleware.test.ts
```diff
@@ -2,176 +2,234 @@ import * as utils from '../../src/base/utils';
...
+            ui5: {
+                path: '/resources',
+                url: 'http://ui5.example'
+            }
+        };
+        beforeEach(() => {
+            ui5ProxySpy.mockClear();
         });
-        expect(ui5ProxySpy).toBeCalledWith(expect.objectContaining({}), expect.objectContaining({ secure: false }));
-    });
```
ullasholla, line 209: Do we need a test for a mising `manifest.json`?

ullasholla (changes requested):
@tobiasqueck, thanks for the PR.

* `patch` changeset looks correct
* the code is readable and clean
* tests cover most of the changes

I had a few minor comments, could you please take a look when you can? Cheers.

packages/ui5-proxy-middleware/src/base/utils.ts
tobiasqueck: I am sure whether this is really required because of `parseFloat`. Based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN it is more robust when you provide strings, booleans etc, however, in our case we convert the input beforehand, therefore, it shouldn't be different.
So, it is something not to straight forward to change, therefore, I'd keep it our of here.

packages/ui5-proxy-middleware/src/base/utils.ts
tobiasqueck: Fixed with ea8223bf

packages/ui5-proxy-middleware/src/base/ui5.d.ts
tobiasqueck: Already requested at https://github.com/SAP/ui5-tooling/issues/651.

packages/ui5-proxy-middleware/src/base/ui5.d.ts
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * https://sap.github.io/ui5-tooling/api/module-@ui5_fs.Resource.html
```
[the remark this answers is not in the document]
tobiasqueck: I used the v2 types and only defined so far the ones that are also compatible with v3.
Fixed with https://github.com/SAP/open-ux-tools/pull/1054/commits/b92aa5605a142a43dc72da2d50b8bbd878d18786.

packages/ui5-proxy-middleware/src/ui5/middleware.ts
tobiasqueck: It is wrapped in a try/catch in line 80.
I tried `byGlob` and it always returned an empty array if it didn't find anything, but to be sure, I added the undefined check with 53b4ed73

packages/ui5-proxy-middleware/test/ui5/middleware.test.ts
tobiasqueck: Good point, missing test added with 4524f436

tobiasqueck:
@ullasholla please re-review, I believe, I have addressed all your concerns
@zdravko-georgiev would you be available for the 2nd review?

ullasholla (dismissed):
@tobiasqueck, thanks for addressing all the review comments.

tobiasqueck:
@Klaus-Keller thanks for your direct message. I have enhanced the `test/test-input` folder with a test setup for local testing. Just go to `test/test-input/easy-ui5-app` and then run `npm i && npm start`. 
With this fix it will work as expected, without the fix the middleware will crash.

packages/ui5-proxy-middleware/src/base/utils.ts
```diff
@@ -195,36 +195,40 @@ export const setHtmlResponse = (res: any, html: string): void => {
...
 /**
  * Determines which UI5 version to use when previewing the application.
  *
  * @param version ui5 version as defined in the yaml or via cli argument
  * @param log logger for outputing information from where ui5 version config is coming
+ * @param manifest optional already loaded manifest.json
  * @returns The UI5 version with which the application will be started
  */
-export async function resolveUI5Version(version?: string, log?: ToolsLogger): Promise<string> {
+export async function resolveUI5Version(version?: string, log?: ToolsLogger, manifest?: Manifest): Promise<string> {
```
Klaus-Keller, line 231: I propose to make `manifest.json` mandatory. Function `resolveUI5Version` is not in the public interface and used in one place in `middleware.ts` only. This would make two other functions obsolete and we can remove them:
1. `getUI5VersionFromManifest()`
2. `getManifest()` -> this is especially good as it contains the problematic `process.cwd()` statement and with removing this we minimize the risk a similar issue occurs again

Whenever we need the manifest again, we pass it from `middleware.ts`.

What do you think?

packages/ui5-proxy-middleware/src/base/utils.ts
```diff
@@ -195,36 +195,40 @@ export const setHtmlResponse = (res: any, html: string): void => {
...
  * @param args list of runtime args
+ * @param manifest optional already loaded manifest.json
  * @returns The minUI5Version from manifest.json or undefined otherwise
  */
-export async function getUI5VersionFromManifest(args: string[]): Promise<string | undefined> {
-    const manifest = await getManifest(args);
-    return manifest['sap.ui5']?.dependencies?.minUI5Version;
+export async function getUI5VersionFromManifest(args: string[], manifest?: Manifest): Promise<string | undefined> {
+    const mf = manifest ?? (await getManifest(args));
+    return mf?.['sap.ui5']?.dependencies?.minUI5Version;
```
Klaus-Keller, line 220: Could be deleted, see other comment

packages/ui5-proxy-middleware/src/base/utils.ts
```diff
@@ -195,36 +195,40 @@ export const setHtmlResponse = (res: any, html: string): void => {
...
-    const manifest: Manifest = JSON.parse(readFileSync(manifestPath, { encoding: 'utf8' }));
-
-    return manifest;
-};
+    if (existsSync(manifestPath)) {
+        return JSON.parse(readFileSync(manifestPath, { encoding: 'utf8' }));
+    } else {
+        return undefined;
+    }
+}
```
Klaus-Keller, line 209: Could be deleted, see other comment

packages/ui5-proxy-middleware/src/base/utils.ts
```diff
@@ -236,7 +240,7 @@ export async function resolveUI5Version(version?: string, log?: ToolsLogger): Pr
         ui5Version = version ? version : '';
         ui5VersionLocation = getYamlFile(process.argv);
     } else {
-        const minUI5Version = await getUI5VersionFromManifest(process.argv);
+        const minUI5Version = await getUI5VersionFromManifest(process.argv, manifest);
```
Klaus-Keller, line 243:
```suggestion
        const minUI5Version = manifest?.['sap.ui5']?.dependencies?.minUI5Version;
```

Klaus-Keller (commented):
Hi @tobiasqueck, 
I like the approach to use `resources` to look up `manifest.json`, but would then remove the code in `utils.ts` that reads the manifest. See my comments.

tobiasqueck:
> but would then remove the code in `utils.ts` that reads the manifest.

We tried to keep a clean separation of code specific to running this middleware in the ui5 tooling compared to code independent of the UI5 tooling. This new function is specific to the ui5 tooling therefore it should not move to the `utils`. 
We use this separation in all our middlewares.

packages/ui5-proxy-middleware/src/base/utils.ts
tobiasqueck: Agree, done with https://github.com/SAP/open-ux-tools/pull/1054/commits/e66599a445bf869d90bb045cd8e762f64fbb6128

packages/ui5-proxy-middleware/src/base/utils.ts
tobiasqueck: https://github.com/SAP/open-ux-tools/pull/1054/commits/e66599a445bf869d90bb045cd8e762f64fbb6128

packages/ui5-proxy-middleware/src/base/utils.ts
tobiasqueck: https://github.com/SAP/open-ux-tools/pull/1054/commits/e66599a445bf869d90bb045cd8e762f64fbb6128

packages/ui5-proxy-middleware/src/base/utils.ts
tobiasqueck: https://github.com/SAP/open-ux-tools/pull/1054/commits/e66599a445bf869d90bb045cd8e762f64fbb6128

815are (approved):
Checked:
1. Code is easy to understand - I did not notice issue;
2. Tests are maintained - scenarios listed in bug are covered

I did not test locally.

PR #1057, opened 2023-06-05
fix: remove version from pnpm/action-setup to use packageManager from package.json

Klaus-Keller:
The GitHub action `pnpm/action-setup@{version}` does not need a version property, if a `packageManager` field in `package.json` is present, see also https://github.com/pnpm/action-setup#version.

This PR removes all `version` fields for `pnpm/action-setup`. This allows to control the version of `pnpm` solely in one place:
https://github.com/SAP/open-ux-tools/blob/46eeb620786dde2150455ef0411146c29a42bbcc/package.json#L58


kranthie-sap (approved):
Awesome. This should reduce maintenance of the pipeline when we upgrade `pnpm`. Thanks.
No changeset required since it only affects the pipeline.

devinea (approved):
Great. Thanks.

PR #1058, opened 2023-06-06
fix(fpm-writer): update template of typescript Controller Extension after changes in 'transform-ui5' settings(.babelrc.json)

815are:
Root issue https://github.com/SAP/open-ux-tools/issues/987
Root PR was [#1020](https://github.com/SAP/open-ux-tools/pull/1020)
This PR fixes follow up issue in `fpm-writer`

Generated `controller.ts` by `fpm-writer` does not work using new configuration(#1020):
```
{
    "ignore": ["**/*.d.ts"],
    "presets": [
        [
            "transform-ui5",
            {
                "overridesToOverride": true
            }
        ],
        "@babel/preset-typescript"
    ]
}
```

Old version of `controller.ts` was:
```
export default class Global2 extends ControllerExtension<ExtensionAPI> {
        static override = {
            onInit(this: Global2) {
                const model = this.base.getExtensionAPI().getModel();
            }
        }
    }
```
Which gets transpiled into:
image
such combination does not works.

Working version with `overrides` instead of `override`:
```
export default class Global2 extends ControllerExtension<ExtensionAPI> {
        static overrides = {
            onInit(this: Global2) {
                const model = this.base.getExtensionAPI().getModel();
            }
        }
    }
```
Then it is transpiled to:
image

### DIFF
--- a/.changeset/new-bags-obey.md
+++ b/.changeset/new-bags-obey.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fe-fpm-writer': patch
+---
+
+Fix: generated typescript controller extension does not work with latest ui5-application-writer version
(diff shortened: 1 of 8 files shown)


hitesh-parmar (approved):
Thanks @815are 

- Reviewed change of "override" to "overrides" in new controller extensions

devinea:
Thanks @815are for the fix.

Issue #1029, opened 2023-05-24
TBI - Enhance the `--cloud-service-key` flow to support UAA credentials


longieirl:
### Description (include screenshots)
When running the ABAP deploy comamnd;
```
npx fiori deploy --name ZJLONG0205 --package '$tmp' --transport '' --url 'https://server.exmaple' --cloud true --cloud-service-key './service.json'
```
a new window tab is opened and I have login to authenticate the session. I would like to use the existing service.json file with my credentials to automate the flow, for example ina ci/cd pipeline flow.

The following slimmed down version of a service.json file has been appended with a username and password;

```json
{
 "uaa": {
  "username": "env:username",
  "password": "env"password",
  "clientid": "abapcp-test!b2274",
  "clientsecret": "JveKNuBq7WA",
  "url": "https://server.exmaple"
 },
 "url": "https://1304ac32-36df-4308-ae73-831943ee6763.server.exmaple"
}
```
There should be enough information provided to retrieve the access and refresh token without having to open a new browser window to authenticate the session and a allow a successful ABAP deployment of my Fiori UI app.

This will provide an automated way for ci/cd pipelines jobs as well and this is existng behaviour in Fiori Tools.

### Value
Improved way of deploying to a ABAP cloud.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Update axios-extension to handle the UAA credentials if present in the service.json file
- [ ] Update test cases

tobiasqueck:
@longieirl @cianmSAP is this one done?

tmackowski:
Hi @longieirl ,

Is this still open? I'm trying to deploy Fiori app to ABAP BTP environment from CI/CD pipeline and cannot configure it to bypass the browser popup.

longieirl:
@tmackowski this has been delivered since June '23, here is is a sample command;

```bash
npx fiori deploy --name ZJLONG0205 --package '$tmp' --transport '' --url 'https://mysystem.abap.us31.abap.int.sap' --cloud true --cloud-service-key '/users/jl/MyFioriProject/service.json' --verbose
```
`service.json` looks like the following;
```json
{
 "uaa": {
  "username": "env:UAA_USER",
  "password": "env"UAA_PASSWORD",
  "clientid": "abapcp-test!b2274",
  "clientsecret": "JveKNuBq7WA",
  "url": "https://server.exmaple"
 },
[...]

PR #1059, opened 2023-06-06
Fix/1029/support uaa credentials

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/1029

Set the following environment variables in your CI/CD environment (or create an .env file with them):
```
SERVICE_CLIENT_ID=sb-xs-1304ac32-36df-74
SERVICE_CLIENT_SECRET=JveKN
SERVICE_UAA_URL=https://abapcp.authentication.sap.hana.ondemand.com
SERVICE_USERNAME=my@email.com
SERVICE_PASSWORD=Mypassword
SERVICE_SYSTEM_ID=Y11
```
Run the command;
```
 ./bin/undeploy --url https://1304ac32-36df-4308-ae73-.abap.hanavlab.ondemand.com --name ZTRAJL01021015 --transport Y11K900443 --cloud --cloud-service-env
```

### DIFF
--- a/packages/deploy-tooling/README.md
+++ b/packages/deploy-tooling/README.md
@@ -53,7 +53,7 @@ Usage: deploy [options]
 Options:
   -c, --config <path-to-yaml>          Path to config yaml file
   -y, --yes                            yes to all questions (default: false)
-  -n, --no-retry                       do not retry if deploy fails for any reason (default: false)
+  -n, --no-retry                       do not retry if deploy fails for any reason, for CI/CD pipeline flows this option needs to be included (default: false)
   --verbose                            verbose log output (default: false)
   --destination  <destination>         Destination in SAP BTP pointing to an ABAP system
   --url <target-url>                   URL of target ABAP system
@@ -85,7 +85,7 @@ Usage: undeploy [options]
 Options:
   -c, --config <path-to-yaml>          Path to config yaml file
   -y, --yes                            yes to all questions (default: false)
-  -n, --no-retry                       do not retry if undeploy fails for any reason (default: false)
+  -n, --no-retry                       do not retry if undeploy fails for any reason, for CI/CD pipeline flows this option needs to be included (default: false)
   --verbose                            verbose log output (default: false)
   --destination  <destination>         Destination in SAP BTP pointing to an ABAP system
   --url <target-url>                   URL of target ABAP system
--- a/packages/deploy-tooling/src/cli/config.ts
+++ b/packages/deploy-tooling/src/cli/config.ts
@@ -155,7 +155,7 @@ export async function mergeConfig(taskConfig: AbapDeployConfig, options: CliOpti
     config.keep = mergeFlag(options.keep, taskConfig.keep);
     config.strictSsl = mergeFlag(options.strictSsl, taskConfig.strictSsl);
     config.yes = mergeFlag(options.yes, taskConfig.yes);
-    config.retry = options.retry;
+    config.retry = process.env.NO_RETRY ? !process.env.NO_RETRY : mergeFlag(options.retry, taskConfig.retry);
 
     if (!options.archiveUrl && !options.archivePath && !options.archiveFolder) {
         options.archiveFolder = 'dist';
--- a/packages/deploy-tooling/src/types/index.ts
+++ b/packages/deploy-tooling/src/types/index.ts
@@ -61,7 +61,7 @@ export interface CommonOptions {
     verbose?: boolean;
 
     /**
-     * If set to false then do not retry if a deployment fails, true by default
+     * '--no-retry' cli param negates the retry value which is true by default
      */
     retry?: boolean;
 }
--- a/packages/deploy-tooling/test/unit/cli/config.test.ts
+++ b/packages/deploy-tooling/test/unit/cli/config.test.ts
@@ -4,6 +4,12 @@ import { join } from 'path';
 import { readFileSync } from 'fs';
 
 describe('cli/config', () => {
+    const env = process.env;
+
+    afterAll(() => {
+        process.env = env;
+    });
+
     describe('getDeploymentConfig', () => {
         test('valid config path', async () => {
             expect(await getDeploymentConfig(join(__dirname, '../../test-input/ui5-deploy.yaml'))).toBeDefined();
@@ -71,7 +77,7 @@ describe('cli/config', () => {
             expect(merged.target.serviceKey).toEqual(JSON.parse(readFileSync(cloudServiceKey, 'utf-8')));
         });
 
-        test('service keys merged from env', async () => {
+        test('validate reading of environment variables', async () => {
             process.env.SERVICE_URL = 'http://service-url';
             process.env.SERVICE_UAA_URL = 'http://uaa-url';
             process.env.SERVICE_CLIENT_ID = 'MyClientId';
@@ -79,10 +85,12 @@ describe('cli/config', () => {
             process.env.SERVICE_SYSTEM_ID = 'Y11';
             process.env.SERVICE_USERNAME = 'MyUsername';
             process.env.SERVICE_PASSWORD = 'MyPassword';
+            process.env.NO_RETRY = 'true';
             const merged = await mergeConfig(config, {
                 cloud: true,
                 cloudServiceEnv: true
             } as CliOptions);
+            expect(merged.retry).toEqual(false);
             expect(merged.target.serviceKey).toMatchObject({
                 systemid: 'Y11',
                 uaa: {


packages/axios-extension/src/auth/uaa.ts
```diff
@@ -86,6 +86,24 @@ export class Uaa {
         return this.serviceInfo.systemid;
     }
 
+    /**
+     * Getter for resource owner.
+     *
+     * @returns username
+     */
```
ullasholla, line 93: Do we need doc comments for private getters? The comments are longer than the code

packages/axios-extension/src/auth/uaa.ts
```diff
@@ -86,6 +86,24 @@ export class Uaa {
...
+    private get username(): string {
+        return this.serviceInfo.uaa.username;
+    }
+
+    /**
+     * Getter for resource owner password.
+     *
+     * @returns password
+     */
```
ullasholla, line 102: Do we need doc comments for private getters? The comments are longer than the code.
I'd add that if the comments are needed to clarify the code then the getter name may need to be rethought.

packages/axios-extension/src/auth/uaa.ts
```diff
@@ -257,4 +299,14 @@ export class Uaa {
         this.log.info('Got access token successfully');
         return response?.data?.access_token;
     }
+
+    /**
+     * @returns an access token using the BTP UAA credentials
+     */
+    public async getAccessTokenUsingClientCredentials(): Promise<string> {
```
ullasholla, line 306:
```suggestion
    public async getAccessTokenWithClientCredentials(): Promise<string> {
```
`With` sounds a bit better?

packages/axios-extension/src/auth/uaa.ts
```diff
@@ -257,4 +299,14 @@ export class Uaa {
...
+        const tokenRequest = this.getTokenRequestForClientCredential();
+        const response: AxiosResponse = await axios.request(tokenRequest);
```
ullasholla, line 308: What are the failure modes for this? How should they be handled?

packages/deploy-tooling/README.md
ullasholla, line 1: Can we add a warning not to commit or check-in the `.env` file? We want to encourage users not to accidentally add sensitive information to their repos (maybe recommend adding the filename to `.gitignore` too)

packages/deploy-tooling/test/unit/cli/index.test.ts
```diff
@@ -77,6 +80,26 @@ describe('cli', () => {
         });
     });
 
+    describe('deploy | undeploy should handle exception', () => {
+        test.each([runDeploy, runUndeploy])(
+            'unsuccessful deploy | undeploy, help options is shown if no parameters are passed in $param',
+            async (method) => {
+                // Dont exit the jest process
```
ullasholla, line 87:
```suggestion
                // Don't exit the jest process
```
Typo

packages/deploy-tooling/test/unit/cli/index.test.ts
```diff
@@ -85,19 +108,21 @@ describe('cli', () => {
...
-            process.argv = ['node', 'test'];
+        test('successful undeploy with environment variable and no config file', async () => {
+            const target = 'https://target.example';
+            process.argv = [
+                'node',
+                'test',
+                '--test',
+                '--yes',
+                '--url',
+                target,
```
ullasholla, line 119: Why use a `const` and not a literal here?

packages/deploy-tooling/test/unit/cli/index.test.ts
```diff
@@ -85,19 +108,21 @@ describe('cli', () => {
...
 
-        test('unsuccessful undeploy, help options is shown if no parameters are passed in', async () => {
-            // Dont exit the jest process
-            const mockExit = jest.spyOn(process, 'exit').mockImplementation((number) => {
-                throw new Error('process.exit: ' + number);
-            });
-            const errorMock = jest.spyOn(Command.prototype, 'error').mockImplementation();
-            const helpMock = jest.spyOn(Command.prototype, 'help');
-            process.argv = ['node', 'test'];
+        test('successful undeploy with environment variable and no config file', async () => {
```
ullasholla, line 111: The test description makes me a little uncomfortable, it only successfully _calls_ `undeploy`, doesn't it?

packages/axios-extension/test/abap/abap-adt-service.test.ts
```diff
@@ -288,19 +306,62 @@ describe('Use existing connection session', () => {
...
+        { remove: 'url', errorStr: 'UAA URL missing' }
+    ])('Fail with error: $errorStr', async ({ remove, errorStr }) => {
+        try {
+            const cloneObj = cloneDeep(configForAbapOnCloud);
+            delete cloneObj.service.uaa[remove];
+            createForAbapOnCloud(cloneObj as any);
+            throw new Error('Should not pass');
+        } catch (error) {
+            expect(error).toEqual(new Error(errorStr));
+        }
```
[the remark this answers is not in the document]
slin-sap: I like separating test code and assertion statements. :D

packages/deploy-tooling/src/base/config.ts
```diff
@@ -31,13 +31,17 @@ export function getConfigForLogging(
     }
 }
 
+export function loadEnv(): void {
```
tobiasqueck, line 34: why do we wrap one function call into another function?

tobiasqueck: I am also not sure why we need to call it twice in our coding then. Shouldn't there be one place, somewhere at the very beginning for the CLI use case where call config and then we are good?

packages/axios-extension/test/abap/abap-adt-service.test.ts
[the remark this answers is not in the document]
ullasholla: > I like separating test code and assertion statements. :D

Is that a comment about the current code?
IMO,
```
expect(functionBeingTested()).toThrow('foo');
``` 
is much more readable than:
```
try {
  functionBeingTested();
  throw new Error('function did not throw');
} catch (error) {
  expect(error).toEqual(new Error('foo'));
}

File: packages/deploy-tooling/README.md
Comment (longieirl) [reply packages/deploy-tooling/README.md:1]: Any generated project will have .env already appended. 
But its good that we explivity state this on the readme as well.

File: packages/axios-extension/src/auth/uaa.ts (hunk 4 above)
Comment (longieirl) [reply packages/axios-extension/src/auth/uaa.ts:308]: When a client requires access without having to interact with a user but the application requires a user context for granting access, the `password` grant type is used. This requires sharing the user credentials with the client id and secret. Thus this works the same as the `getAccessToken` flow where the grant type is `refresh_token` if a refresh token is already passed in, or `authorization_code` grant type if a new refresh_tokn is required.
Not sure if that answers your question?

File: packages/axios-extension/src/auth/uaa.ts (hunk 2 above)
Comment (longieirl) [reply packages/axios-extension/src/auth/uaa.ts]: Lint requires jsdoc! Doesnt complain locally but fails remote build.

File: packages/axios-extension/src/auth/uaa.ts (hunk 4 above)
Comment (ullasholla) [reply packages/axios-extension/src/auth/uaa.ts:309]: Thanks for the explanation. My question is more about handling any failures in the code

File: packages/deploy-tooling/README.md
Comment (ullasholla) [inline packages/deploy-tooling/README.md:1]: @longieirl & @tobiasqueck, does it make sense to add a link to `dotenv` documentation. They support multiple environments and options CI/CD (also people may not want to name the file `.env`. This is configurable):
* https://www.npmjs.com/package/dotenv#-manage-multiple-environments
* https://www.npmjs.com/package/dotenv#-deploying
(Need to ensure `dotenv@16.0.0` supports them, latest version as of today is 16.1.4)

Comment (tobiasqueck) [reply packages/deploy-tooling/README.md:1]: To be honest, I wouldn't recommend using `dotenv` in CI/CD environments, I would set the env variables on the process or in the environment. `dotenv` is something I'd only use for local testing or as a workaround for existing modules that cannot use env variables but still want to call this new module.

**Review by tobiasqueck — APPROVED** [review]
* code enhancements adhere to coding standards and are easy to understand
* tests cover the new scenario
* did NOT test it locally but I trust the implementation (based on the review)
* changest looks good
* small suggestion but not critical for the approval: I would turn the description for the new CLI option around and focus on using env variables and then maybe in parenthesis mention the support for `dotenv`

File: packages/deploy-tooling/README.md
Comment (longieirl) [reply packages/deploy-tooling/README.md:1]: @ullasholla and @tobiasqueck I will add a note to the original story to cover this documentation update.

File: packages/axios-extension/src/auth/uaa.ts (hunk 4 above)
Comment (slin-sap) [reply packages/axios-extension/src/auth/uaa.ts:309]: @longieirl  I think Ullas is questioning about there is no try-catch error handling for the axios request. `getAccessTokenWithClientCredentials` is called in `attachUaaAuthInterceptor`, which also doesn't handle errors.

**Comment by longieirl** [comment]
@slin-sap please review when you get a chance? Thanks.

File: packages/axios-extension/src/auth/uaa.ts (hunk 4 above)
Comment (ullasholla) [reply packages/axios-extension/src/auth/uaa.ts:309]: Thanks for clarifying that Shen. That's exactly what I was trying to say as well

Comment (longieirl) [reply packages/axios-extension/src/auth/uaa.ts:309]: @ullasholla are you referring to there being no try/catch? It's the calling client that needs to handle this since its the interceptor that is setting up the appropiate headers for all the subsequent API calls to work. For example, we make a GET and then a POST or a PUT. On the initial GET, if this fails with the UAA properties provided when using either `getAccessToken` or `getAccessTokenWithClientCredentials`, the its deploy-tooling to decide how to proceed, either prompt for credentials or ignore and exit the process.

Comment (ullasholla) [reply packages/axios-extension/src/auth/uaa.ts:309]: Is the error handling in `handleAxiosError()`?

Do we try to prompt for creds for an HTTP 401?
image

If yes, in the pipeline, this will just sit there with the prompt and then time out

Comment (longieirl) [reply packages/axios-extension/src/auth/uaa.ts:309]: No, the `handleError` in ui5-tooling is configured with the following code;
```
if (!config.noRetry && isAxiosError(error)) {
```
So for a ci/cd pipepline flow, it will exit the process and log the error.

**Comment by longieirl** [comment]
@slin-sap can you approve and I will wait then until @ullasholla approves before merging.

File: packages/axios-extension/src/auth/uaa.ts (hunk 4 above)
Comment (ullasholla) [reply packages/axios-extension/src/auth/uaa.ts:309]: Quickly tested it out with the `deploy` command instead of `undeploy`. `config.noRetry` was undefined:
image

I forced the status code to be 401 and I get this:
image

Did not exit with an error for me.

Could you quickly test it once for me please? Please test `deploy` like I did

Comment (longieirl) [reply packages/axios-extension/src/auth/uaa.ts:309]: @ullasholla good catch! No idea how this was working before! So there is a bug with the commander module, using the param `--no-retry` does not map to `noRetry`. I had to change it to `--abort-retry`. From locally testing with Command v9, two letter words dont work as expected!

```
./bin/deploy --name ZJLONG0205 --package '$tmp' --transport '' --url 'https://abcd.ondemand.com' --cloud true --cloud-service-env  --safe -n
2023-06-14 22:05:23 info    abap-deploy-task: Creating archive from dist. 
2023-06-14 22:05:23 info    abap-deploy-task: Archive created from dist. 
2023-06-14 22:05:23 warn    abap-deploy-task: You chose not to validate SSL certificate. Please verify the server certificate is trustful before proceeding. See documentation for recommended configuration (https://help.sap.com/viewer/17d50220bcd848aa854c9c182d65b699/Latest/en-US/4b318bede7eb4021a8be385c46c74045.html). 
2023-06-14 22:05:23 info    abap-deploy-task: Starting to deploy. 
2023-06-14 22:05:23 error   abap-deploy-task: Deployment has failed. 
2023-06-14 22:05:23 error   abap-deploy-task: Change logging level to debug your issue
        (see examples https://github.com/SAP/open-ux-tools/tree/main/packages/deploy-tooling#configuration-with-logging-enabled) 
Request failed with status code 401
```
So now, the `-n` and  `--abort-retry` work as expected.

FYI @tobiasqueck

longieirl:
Need a fresh review from @slin-sap @ullasholla @tobiasqueck regards latest changes. I'm going to open a bug in the commander module around `--no-retry` flag.

tobiasqueck:
> fix(deploy-tooling): fix no-retry flag

Explain this one please.

packages/axios-extension/src/auth/uaa.ts
ullasholla: Passing `-n` or `--abort-retry` works fine now, thanks @longieirl.

For the particular usecase we're discussing - using deploy/undeploy in a pipeline and having trouble with credentials - the user needs to explicitly pass `-n/--abort-retry`

This is because we cannot assume using `--cloud-service-env` implies that it's being used in a CI env.

In our doco it's not clear that not retrying is the same as suppressing prompts.
This is what we have:
`  -n, --abort-retry                    do not retry if deploy fails for any reason (default: false)`

John, @theStiggySpd, @ianquigley-sap  & @tobiasqueck, I can think of a few things we can do to make this clearer to the user (not all mutually exclusive):
* mention it in the `--abort-retry` doco
* add a 'non-interactive' option to do the same thing
* add an example for the CI scenario in the doco

Thoughts?

longieirl: @ullasholla I agree, I've added another task to the story to update documentation around this flow.
As regards the non-interactive mode, I will add that as an additional task to the original story as its currently only being supported by ui5-tooling for the moment.

tobiasqueck (approved):
Previous review approval still valid

packages/axios-extension/src/auth/uaa.ts
longieirl: Further documentation to be appended.

Issue #1060, opened 2023-06-07
TBI - fpm-writer. Custom extension fragments. Wrap generated content into container for extensions like custom section, sub-section


815are:
### Description (include screenshots)
Improves usability of newly created extension, in way that it simpler to append new controls into fragments.

Problem with currentl approach, is that if we have following fragment for custom sub section:
```
<core:FragmentDefinition xmlns:core="sap.ui.core" xmlns="sap.m" xmlns:macros="sap.fe.macros">
	<Text id="testsub" text="testsub1" />
	<Text id="testsub2" text="testsub2" />
</core:FragmentDefinition>
```
It will be rendered as:
![image](https://github.com/SAP/open-ux-tools/assets/90789422/f66873b7-8fdb-40fa-8d58-46f62b0ec3fc)

If we apply wrapper:
```
<core:FragmentDefinition xmlns:core="sap.ui.core" xmlns="sap.m" xmlns:macros="sap.fe.macros">
	<VBox>
		<Text id="testsub" text="testsub1" />
		<Text id="testsub2" text="testsub2" />
	</VBox>
</core:FragmentDefinition>
```
Rendered as:
![image](https://github.com/SAP/open-ux-tools/assets/90789422/3b2f9911-1853-45f0-8553-2801c78ca11d)

Wrapper should be applied for following extensions:
1. section and sub-section

Wrapper should not be applied for:
1. column - does not makes much sense
2. view - it supports multi aggregation by default:
![image](https://github.com/SAP/open-ux-tools/assets/90789422/8a1747ad-7cef-46ac-a426-a27f7fa3024a)

### Tasks
- [ ] Apply wrapper for section and sub section

PR #1061, opened 2023-06-07
feat(fpm-writer): Custom extension fragments. Wrap generated content into VBox container for extensions like custom section, sub-section

815are:
Issue #1060

Some background:
We faced usability issue regarding custom section, when we tried to append generated fragment with new control, like:
```
<core:FragmentDefinition xmlns:core="sap.ui.core" xmlns="sap.m" xmlns:macros="sap.fe.macros">
	<Text id="testsub" text="testsub1" />
	<Text id="testsub2" text="testsub2" />
</core:FragmentDefinition>
``` 
It will be rendered as:
image

But same example works ok with custom views:
image

Reason is that custom section has single aggregation:
image

In result to improve usability - decision was to wrap content with VBox for custom sections and sub sections.
And ignore other extensions:
1. views - supports multi aggregtion and does not require wrapper
2. columns - does not make much sense to add container

Example of generation:
old:
```
<core:FragmentDefinition xmlns:core=\\"sap.ui.core\\" xmlns=\\"sap.m\\" xmlns:macros=\\"sap.fe.macros\\">
	<Text text=\\"DummySection\\" />
</core:FragmentDefinition>"
```
new:
```
<core:FragmentDefinition xmlns:core=\\"sap.ui.core\\" xmlns=\\"sap.m\\" xmlns:macros=\\"sap.fe.macros\\">
	<VBox>
		<Text text=\\"DummySection\\" />
	</VBox>
</core:FragmentDefinition>"
```


packages/fe-fpm-writer/templates/common/partials/AdditionalDependencies.xml
```diff
@@ -0,0 +1 @@
+<%- typeof dependencies !== 'undefined' ? Object.entries(dependencies).map((dependency) => ` ${dependency[0]}="${dependency[1]}"`).join('') : '' %>
```
tobiasqueck, line 1: That seems like an odd place or way to do it. If we need to reuse it, couldn't we do this already in the (debuggable) code and then give the result as part of the config to the template?

tobiasqueck (changes requested):
One suggestion for improvement, the rest looks good.

packages/fe-fpm-writer/templates/common/partials/AdditionalDependencies.xml
815are: @tobiasqueck 
thanks for pointing.
Some background:
1. map with additional dependencies was added some time ago and currently used by custom sections only to include `sap.fe.macros` based on ui5 version - https://github.com/SAP/open-ux-tools/blob/feat/1060/containerForSectionFragment/packages/fe-fpm-writer/src/section/index.ts#L41;
2. Idea was that probably in future similar will be added to custom views;

I simplified logic within commit d2a3c6b07ea34eb907b8b8cd5c89eae108ad6334
1. there is no more `partials`
2. instead of map - directly string is used

tobiasqueck (changes requested):
* test coverage is good
* changeset needs another round: is it a patch (for a fix) or is it the implementation of a new 'Feature' requiring a minor version update
     * for me it is a fix/patch but you can judge it better

815are:
> changeset needs another round: is it a patch (for a fix) or is it the implementation of a new 'Feature' requiring a minor version update
> * for me it is a fix/patch but you can judge it better

@tobiasqueck 
I changed it to `minor` - I think it more fits as backwards compatible feature

tobiasqueck (approved):

broksy (approved):
- Code looks good and is easily understandable.
- Changeset is understandable.

Approved from my side.

Issue #1062, opened 2023-06-08
BUG [SECURITY] - fast-xml-parser vulnerable to regex injection via Doctype entities


kranthie-sap:
### Description
The following errors are reported by `pnpm audit`:
```
┌─────────────────────┬────────────────────────────────────────────────────────┐
│ high                │ fast-xml-parser vulnerable to Regex Injection via      │
│                     │ Doctype Entities                                       │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Package             │ fast-xml-parser                                        │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Vulnerable versions │ <4.2.4                                                 │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Patched versions    │ >=4.2.4                                                │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Paths               │ examples/odata-cli >                                   │
│                     │ @sap-ux/axios-extension@link:../../packages/axios-     │
│                     │ extension > fast-xml-parser@3.12.20                    │
│                     │                                                        │
│                     │ examples/simple-generator >                            │
│                     │ @sap-ux/axios-extension@link:../../packages/axios-     │
│                     │ extension > fast-xml-parser@3.12.20                    │
│                     │                                                        │
│                     │ packages/app-config-writer >                           │
│                     │ @sap-ux/axios-extension@link:../axios-extension >      │
│                     │ fast-xml-parser@3.12.20                                │
│                     │                                                        │
│                     │ ... Found 14 paths, run `pnpm why fast-xml-parser` for │
│                     │ more information                                       │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ More info           │ https://github.com/advisories/GHSA-6w63-h3fj-q4vw      │
└─────────────────────┴────────────────────────────────────────────────────────┘
2 vulnerabilities found
Severity: 2 high
```

donal-tobin-sap:
Updated

PR #1063, opened 2023-06-08
fix(security): upgrade fast-xml-parser

donal-tobin-sap:
https://github.com/NaturalIntelligence/fast-xml-parser/security/advisories/GHSA-6w63-h3fj-q4vw 
https://github.com/SAP/open-ux-tools/security/dependabot/12 
https://github.com/SAP/open-ux-tools/security/dependabot/13

Issue: https://github.com/SAP/open-ux-tools/issues/1062
Issue: tools-suite #25082

### DIFF
--- a/.changeset/tough-scissors-tap.md
+++ b/.changeset/tough-scissors-tap.md
@@ -1,0 +1,13 @@
+---
+'@sap-ux/axios-extension': patch
+'@sap-ux/odata-service-writer': patch
+'@sap-ux/odata-cli': patch
+'@sap-ux/generator-simple-fe': patch
+'@sap-ux/app-config-writer': patch
+'@sap-ux/backend-proxy-middleware': patch
+'@sap-ux/deploy-tooling': patch
+'@sap-ux/environment-check': patch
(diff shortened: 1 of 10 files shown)


kranthie-sap (changes requested):
The changes to package.json file and updated usage of `fast-xml-parser` looks good. 
Do we need an entry in the `overrides` section just in case another dependency brings an older version in or rely on `pnpm audit` to take care of it when that happens? The latter one probably has less maintenance for us.

Please see minor comment about changeset.

donal-tobin-sap:
> The changes to package.json file and updated usage of `fast-xml-parser` looks good. Do we need an entry in the `overrides` section just in case another dependency brings an older version in or rely on `pnpm audit` to take care of it when that happens? The latter one probably has less maintenance for us.

The guidance is to only use overrides when specifically required and no other solution will work, not to protect in case of future changes.

Klaus-Keller (approved):
Update to new version of `fast-xml-parser` and resulting code changes look good, thanks @donal-tobin-sap!

Issue #1066, opened 2023-06-12
BUG -  fpm-writer. Not possible to create custom section/subsection without position.


815are:
### Description
`sap.fe` supports rendering of custom section/subsection without anchoring and positioning
custom sections and custom sub sections can be created without position

We can see such examples here - https://ui5.sap.com/test-resources/sap/fe/core/fpmExplorer/index.html#/customElements/customSubSectionContent

As tested  -same applies for both - sections and subsections

### Steps to Reproduce
1. use `fpm-writer`
2. create custom section without position;

Problem - it is not possible.

Interface expects `position` as required - https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/section/types.ts#L17

### Expected results
Creation of custom section/subsection should be possible without position

### Actual results
Creation of custom section/subsection is not posible without position

815are:
> @815are has this been fixed with #1068? If yes, please close, otherwise, please outline what is missing.

@tobiasqueck 
yes, sorry, fix was merged long time ago, but task was not maintained

PR #1068, opened 2023-06-12
fix(fpm-writer): allow to create custom section and subsection without 'position' property

815are:
Issue https://github.com/SAP/open-ux-tools/issues/1066

We noticed that it is not posible to create custom section without position, because it was considered as required property.
Making position property option similar like for custom actions.


tobiasqueck (approved):
* template changes are clear
* tests cover the different scenarios
* changeset is ok

tobiasqueck (approved):

Issue #1074, opened 2023-06-19
TBI - enhance check in hasMinCdsVersion


docirl:
### Description (include screenshots)
When checking the version of @sap/cds in `hasMinCdsVersion` the version installed can be higher and satisfy the min version requirement but will return false as part of this check.

Example:

Version in package.json
"@sap/cds": ">=6.7.0",

This will install (at time of writing) `6.8.4`. The current check will reduce the version to `6.7.0` and return false for min version.

Suggestion is to enhance the semver check, so that it will check if the min required @sap/cds version will `satisfy` the semver in the package.json.

### Value
Projects that specify ranges (carets, x, tilde, etc) that would install a valid version for workspaces feature will correctly return true.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Enhance logic of `hasMinCdsVersion` to correctly check possible ranges specified return true. Use semver `satisfy` as an additional check.

tobiasqueck:
Why `6.7.0`? `6.8.2` contains a fix that is required for monorepo-like setup that we need for typescript in CAP.

tobiasqueck:
We still need to change the entry though. It should not have `>=6.70` afterwards, so `false` is correct even if already a version `6.8.4` is installed.

tobiasqueck:
@docirl what is going on with this one? Is it done, could it be closed?

docirl:
Addressed by https://github.com/SAP/open-ux-tools/pull/1075

PR #1075, opened 2023-06-19
tbi(cap-config-writer): enhance @sap/cds check in hasMinCdsVersion

docirl:
#1074 

Enhance check for @sap/cds in `hasMinCdsVersion` to use semver `satisfy` so that checks of ranges return true when required min version would be installed.


packages/cap-config-writer/src/cap-config/package-json.ts
```diff
@@ -61,7 +61,11 @@ export function addCdsPluginUi5(packageJson: Package): void {
  * @returns - true: min cds version is present; false: cds version needs update
  */
 export function hasMinCdsVersion(packageJson: Package): boolean {
-    return gte(coerce(packageJson.dependencies?.['@sap/cds']) ?? '0.0.0', minCdsVersion);
+    const packageCdsVersion = packageJson.dependencies?.['@sap/cds'];
+    return (
+        satisfies(minCdsVersion, packageCdsVersion ?? '0.0.0') ||
```
Klaus-Keller, line 66: `satisifies` allows all versions within a range, we need a min version for the check.

packages/cap-config-writer/test/unit/cap-config/index.test.ts
```diff
@@ -119,4 +121,85 @@ describe('Test checkCdsUi5PluginEnabled()', () => {
...
         expect(await checkCdsUi5PluginEnabled(__dirname, memFs)).toBe(true);
     });
+
+    test('CAP project with valid @sap/cds version using caret(^)', async () => {
+        const memFs = create(createStorage());
+        memFs.writeJSON(join(__dirname, 'package.json'), {
+            dependencies: { '@sap/cds': '^6.7.0' }
+        });
+        const packageJson: Package = (memFs.readJSON(join(__dirname, 'package.json')) ?? {}) as Package;
+        expect(hasMinCdsVersion(packageJson)).toBe(true);
```
Klaus-Keller, line 131: This is wrong, `^6.7.0` should return `false`, as the ^-range allows e.g. for 6.7.2, which is lower than version 6.8.2 which we need.

packages/cap-config-writer/test/unit/cap-config/index.test.ts
```diff
@@ -119,4 +121,85 @@ describe('Test checkCdsUi5PluginEnabled()', () => {
...
+        expect(hasMinCdsVersion(packageJson)).toBe(false);
+    });
+
+    test('CAP project with valid @sap/cds version using x-range', async () => {
+        const memFs = create(createStorage());
+        memFs.writeJSON(join(__dirname, 'package.json'), {
+            dependencies: { '@sap/cds': '6.x' }
+        });
+        const packageJson: Package = (memFs.readJSON(join(__dirname, 'package.json')) ?? {}) as Package;
+        expect(hasMinCdsVersion(packageJson)).toBe(true);
```
Klaus-Keller, line 149: Should return false for `6.x`, because 6.0.0 is valid for it, but does not meet our requirements of minimum 6.8.2

packages/cap-config-writer/test/unit/cap-config/index.test.ts
```diff
@@ -119,4 +121,85 @@ describe('Test checkCdsUi5PluginEnabled()', () => {
...
+        expect(hasMinCdsVersion(packageJson)).toBe(false);
+    });
+
+    test('CAP project with valid @sap/cds version using greater than (>)', async () => {
+        const memFs = create(createStorage());
+        memFs.writeJSON(join(__dirname, 'package.json'), {
+            dependencies: { '@sap/cds': '>4.0.0' }
+        });
+        const packageJson: Package = (memFs.readJSON(join(__dirname, 'package.json')) ?? {}) as Package;
+        expect(hasMinCdsVersion(packageJson)).toBe(true);
```
Klaus-Keller, line 167: wrong, would also allow for e.g. 5.0.0

Klaus-Keller (changes requested):

docirl:
@Klaus-Keller I've updated my change to only include ranges when an optional param is passed. This will allow the as is flow to return false for the version check. The generator can pass the param as it needs to check if the version range is valid to ask the enable workspaces prompt, and we will run an install at the root so the latest (and valid) version will be installed.

Klaus-Keller:
> @Klaus-Keller I've updated my change to only include ranges when an optional param is passed. This will allow the as is flow to return false for the version check. The generator can pass the param as it needs to check if the version range is valid to ask the enable workspaces prompt, and we will run an install at the root so the latest (and valid) version will be installed.

Hi @docirl,

I think I understand now what you want to achieve: a check if the current dependency to `@sap/cds` is within the range of the minimum version `6.8.2`. This means the update would be possible without conflicts to the current range.

To me that sounds like a different check than `hasMinCdsVersion()` and therefore my proposal: keep code as it was and create an additional check `satisfiesMinCdsVersion()` which does check if the range satisfies the min version 6.8.2. When enabling cds plugin, we still need to ensure min version by setting it to `^6.8.2` in the `package.json`.

Klaus-Keller:
Hi @docirl, 
To illustrate what I'm talking about, I've create a draft PR: https://github.com/SAP/open-ux-tools/pull/1077
This slightly changes the process flow to: 

```mermaid
graph TD
    start((Generate app in CAP))-->hasMinCdsVersion["satisfiesMinCdsVersion()"]
    hasMinCdsVersion-- yes -->cdsUi5PluginEnabled["checkCdsUi5PluginEnabled()"]
    cdsUi5PluginEnabled-- yes -->newProjectSetup["New project setup, TypeScript apps possible"]
    cdsUi5PluginEnabled-- no -->askEnable["Ask: Should workspace and dev dependency cds-plugin-ui5 be enabled?"]
    askEnable-- yes --> enabledCdsUi5Plugin["enabledCdsUi5Plugin()"]
    enabledCdsUi5Plugin-->newProjectSetup
    askEnable-- no -->oldProjectSetup
    hasMinCdsVersion-- no -->oldProjectSetup["Old project setup, no TypeScript apps"]
```

-> `hasMinCdsVersion()` is replaced with `satisfiesMinCdsVersion()`

docirl:
Thanks @Klaus-Keller , those changes look good. I've merged them here.

Klaus-Keller (approved):
Hi @docirl,

- thanks for adding extensive tests for versions and version ranges
- thanks for adding vscode launch config
- test coverage is great

Approved from my side.

devinea (approved):
Code changes and tests look great.
Changeset description is not 100% accurate anymore (similar for PR description and text).

Issue #999, opened 2023-05-03
TBI - Switch to @sapui5/types with UI5 1.113 and newer


tobiasqueck:
### Description (include screenshots)
Starting with release 1.113, please use the identical package `@sapui5/types` instead
https://www.npmjs.com/package/@sapui5/ts-types-esm/v/1.113.0

### Value
Best practice

### Architecture Elaboration
n/a

### Notes
n/a

### Tasks
- [ ] Change the code
- [ ] update the snapshots

PR #1076, opened 2023-06-20
Removing types version workaround

tobiasqueck:
A few months ago we implemented an algorithm to detect the best possible UI5 types library version based on internal knowledge of fixes and downports to specific versions. The problematic versions are now out of support so that the downports are not required anymore and that we can use the latest available version.

This PR removes that algorithm and goes back to a simple version mapping.

### DIFF
--- a/.changeset/nice-ears-push.md
+++ b/.changeset/nice-ears-push.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui5-config': minor
+---
+
+Removed types determination workaround.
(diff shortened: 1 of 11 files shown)


devinea (approved):
Thanks @tobiasqueck 
Changes are as expected. Code, test, snapshots and changset 👍

c-kobo:
@tobiasqueck Would it make sense to switch in that course to generating @sapui5/types when using UI5 1.113 and newer, as @sapui5/ts-types-esm has a deprecation note set: https://github.com/SAP/open-ux-tools/issues/999

tobiasqueck:
> @tobiasqueck Would it make sense to switch in that course to generating @sapui5/types when using UI5 1.113 and newer, as @sapui5/ts-types-esm has a deprecation note set: https://github.com/SAP/open-ux-tools/issues/999

No, that should be a separate change.

nlunets (approved):

Issue #1078, opened 2023-06-23
BUG [SECURITY] - semver vulnerable to ReDos and upgrade for fast-xml-parser


hitesh-parmar:
### Description

https://github.com/advisories/GHSA-c2qf-rxjj-qqgw

https://github.com/advisories/GHSA-gpv5-7x3g-ghjv

### Expected results

1. `pnpm audit`
2. no issues

### Actual results

Run `pnpm audit`

![image](https://github.com/SAP/open-ux-tools/assets/24454497/bf53fafc-b794-4615-8449-1858b72853b4)

hitesh-parmar:
Solved with PR #1079

PR #1079, opened 2023-06-23
fix(security): upgrade semver and fast-xml-parser

hitesh-parmar:
Solves #1078 

https://github.com/advisories/GHSA-c2qf-rxjj-qqgw

https://github.com/advisories/GHSA-gpv5-7x3g-ghjv


donal-tobin-sap (dismissed):
Looks good to me.

schreckstefan (dismissed):
OK for me, should be guaranteed by the unit tests that everything works fine as before. No snapshots have been updated.

hitesh-parmar:
Pushed a new commit as meanwhile `semver` got a patch update to `7.5.3`

devinea (commented):
package updates and lock file updates look as expected. 
adding semver overrides is ok temporarily. But we should update dev dependencies soon.

Please update https://github.com/SAP/open-ux-tools/blob/main/docs/version-overrides.md

hitesh-parmar:
Thanks @devinea. Update `version-overrides` and also enhanced overrides entry. Ready for re-review.

devinea (approved):
@hitesh-parmar Thanks for the updates. 
All looks good now.

docs/version-overrides.md
```diff
@@ -1,11 +1,21 @@
 # Version Overrides
 This document lists the version overrides for vulnerable (nested) dependencies and the reason.
 
+## semver
+| Override:           | >=7.5.3 |
```
donal-tobin-sap, line 5: Minor documentation correction. 

The override is still only 7.5.2
But the latest 7.5.3 is consumed.
This is right to consume that latest version because it looks like a vulnerability fix as well, but is not reported as such yet.

schreckstefan (approved):

docs/version-overrides.md
hitesh-parmar: Thanks @donal-tobin-sap. Corrected with my next commit.

donal-tobin-sap (approved):

PR #1080, opened 2023-06-26
Chore - Update dev dependencies

devinea:
Chore - Update dev dependencies 
- github action
    - slackapi/slack-github-action@v1.24.0
- Root package.json
 ``` 
     "@changesets/cli": "2.26.1",
      "@nrwl/nx-cloud": "16.0.5",
      "@types/jest": "29.5.2",
      @typescript-eslint/eslint-plugin": "5.60.0",
      "@typescript-eslint/parser": "5.60.0",
      "eslint": "8.43.0",
      "eslint-config-prettier": "8.8.0",
      "eslint-import-resolver-typescript": "3.5.5",
      "eslint-plugin-jsdoc": "46.3.0",
      "eslint-plugin-promise": "6.1.1",
      "eslint-plugin-sonarjs": "0.19.0",
      "jest-sonar": "0.2.16",
      "nx": "16.4.0",
      "prettier": "2.8.8",
      "rimraf": "5.0.1", and add --glob option needed now
      "ts-jest": "29.1.0",
```

- ui-components dev Deps

```
      "@babel/core": "7.22.5",
      "babel-jest": "29.5.0",
      "css-loader": "6.8.1",
      "eslint-plugin-react": "7.32.2",
      "sass": "1.63.6",
      "sass-loader": "13.3.2",
      "style-loader": "3.3.3",
      "ts-loader": "9.4.3",
      "webpack": "5.88.0"
```

- Add .nxignore to ignore test projects
- Lint fix (removed extra line under jsdoc description
- lint jsdoc fixes (add @returns etc...)

### CODE REVIEW

File: examples/odata-cli/src/activities.ts [hunk 1]
```diff
@@ -12,7 +12,6 @@ import { logger } from './types';
 
 /**
  * Execute a sequence of test calls using the given provider.
- *
```
Klaus-Keller, line 15: Do we really need to delete the newline between jsdoc description and parameters? This creates a lot of noise on this PR and personally I don't think it increases the readability of the jsdoc comments.

devinea: @Klaus-Keller Yeah it is a lot of noise. I've adjusted the eslint config for the new version to keep the old behaviour for the newline after the jsdoc description
https://github.com/SAP/open-ux-tools/pull/1080/commits/4cc4eb32377ef5c056defa06f49f91dbcc369b26 and https://github.com/SAP/open-ux-tools/pull/1080/commits/29add5017080ea6a1f2148d6a6d671bf0b56019f

packages/deploy-tooling/src/base/deploy.ts
```diff
@@ -64,6 +65,7 @@ export async function getCredentials<T extends BasicAuth | ServiceAuth | undefin
  * @param target - url target configuration
  * @param noPrompt - only if not truthy prompt for anything
  * @param logger - reference to the logger instance
+ * @returns {*}  {(Promise<AbapServiceProvider>)} - ABAP Service Provider
```
tobiasqueck, line 68: Why do we add the return type in the doc? Isn't it sufficient to declare it in the signature.
We haven't done that before.

tobiasqueck (approved):
* not sure if we need the changeset, since no code was changed BUT I am also ok with a patch
* `.nxignore` is a good idea
* version updates make sense, build and tests are still running
* changes to the `rimraf` calls are ok
* small comment about the updated JSDoc but no reason for me to block this PR
* I hope I didn't miss anything in this PR

packages/deploy-tooling/src/base/deploy.ts
devinea: the new jsdoc eslint plugin version raises an error in case it is missing. We could disable it if we want, but simpler to keep the defaults IMO.

examples/odata-cli/src/activities.ts
Klaus-Keller: Thanks @devinea!

Klaus-Keller (approved):
- update of devDependencies looks good
- focused review on  cap-config-writer, create, environment-check, mockserver-config-writer, project-access, and ui-components, but had also a brief look at the rest
- did not locally build and test

Approved from my side.

Issue #1083, opened 2023-06-27
BUG - Getting list of transport and packages does not work when appName has namespace


cianmSAP:
### Related Feature
Feature request: [#518](https://github.com/SAP/open-ux-tools/issues/518)

### Description
Getting list of transport and packages does not work when app name has namespace, as it is not encoded

### Steps to Reproduce

1. Generate deploy config with AppName in the format /NAMESPACE/APPNAME
2. Expand TR or Package to get the list

### Expected results
Transport requests are retrieved

### Actual results
Failure to fetch transport requests, error `URI-Mapping cannot be performed due to invalid URI:`

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1085, opened 2023-06-27
fix(axios-ext): encode transport request payload

cianmSAP:
#1083

Encodes the path in case there is a namespace

### DIFF
--- a/.changeset/chilled-geese-sniff.md
+++ b/.changeset/chilled-geese-sniff.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/axios-extension': minor
+---
+
+encode app name for tr requests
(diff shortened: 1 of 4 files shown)


longieirl (approved):
Looks good to me.
- tests cover the change.

longieirl (approved):
Code smell addressed.

.changeset/chilled-geese-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/axios-extension': minor
```
tobiasqueck, line 2: patch should be fine since your fixing a bug

tobiasqueck (changes requested):
- change is clear and easy to understand
- tests cover the fix
- please change the changeset from minor to patch since it is a bug fix and not a new feature

.changeset/chilled-geese-sniff.md
cianmSAP: Updated changeset, thanks @tobiasqueck

tobiasqueck (approved):

longieirl (approved):

Issue #1084, opened 2023-06-27
TBI - Enhance ABAP deploy/undeploy flow to support `service` cli param


longieirl:
### Description (include screenshots)
When running the ABAP deploy or undeploy commands, for example
```
npx deploy --url 'https://server.exmaple' -name ZJLONG0205 --transport 'Y11'  --package  '$tmp' --cloud --cloud-service-env --verbose --no-retry --archive-folder ./dist --noConfig
```

```
npx undeploy --url 'https://server.exmaple' --name 'ZJLONG0205' --transport 'Y11' --cloud --cloud-service-env --verbose --no-retry --noConfig
```

I should be able to append the `--service` param so that the SAPUI5 Repositroy OData service specified is used instead of the default `/sap/opu/odata/UI5/ABAP_REPOSITORY_SRV`.

This should also be supported if defined in the ui5-deploy.yaml, for example;

```yaml
- name: abap-deploy-task
  configuration:
    log: 5
    app:
      name: Z_TEST
      package: $TMP
    target:
      url: https://target.example,
      service: 'bc/myservice'
```

### Value
Improved way of deploying to a ABAP cloud.

### Architecture Elaboration
No

### Notes
- axios-extension allows an alias to be passed to the `getUi5AbapRepository` method when generating the ABAP deploy service.

### Tasks
- [x] Update CLI params to support `--service`
- [x] Update readme.md
- [x] Update test cases

T1mey:
Could you please inform the project piper team to include this parameter into the step transportRequestUploadCTS?

See Params:
https://www.project-piper.io/steps/transportRequestUploadCTS/

longieirl:
Confirmed as working in tools-suite.

PR #1086, opened 2023-06-28
fix(deploy-tooling): Support new parameter 'service' 

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/1084

- tests updated
- changeset appended

### DIFF
--- a/packages/deploy-tooling/README.md
+++ b/packages/deploy-tooling/README.md
@@ -57,6 +57,7 @@ Options:
   --verbose                            verbose log output (default: false)
   --destination  <destination>         Destination in SAP BTP pointing to an ABAP system
   --url <target-url>                   URL of target ABAP system
+  --service                            Target SAPUI5 Respository OData Service
   --client <sap-client>                Client number of target ABAP system
   --cloud                              target is an ABAP Cloud system
   --cloud-service-key <file-location>  JSON file location with the ABAP cloud service key.
@@ -89,6 +90,7 @@ Options:
   --verbose                            verbose log output (default: false)
   --destination  <destination>         Destination in SAP BTP pointing to an ABAP system
   --url <target-url>                   URL of target ABAP system
+  --service                            Target SAPUI5 Respository OData Service
   --client <sap-client>                Client number of target ABAP system
   --cloud                              target is an ABAP Cloud system
   --cloud-service-key <file-location>  JSON file location with the ABAP cloud service key.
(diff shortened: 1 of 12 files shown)


packages/deploy-tooling/README.md
```diff
@@ -57,6 +57,7 @@ Options:
   --verbose                            verbose log output (default: false)
   --destination  <destination>         Destination in SAP BTP pointing to an ABAP system
   --url <target-url>                   URL of target ABAP system
+  --service                            Target SAPUI5 Respository OData Service
```
tobiasqueck, line 60: Could we indicated that this is to be used for an optional alias for the service and usually not required?

packages/deploy-tooling/test/test-input/ui5-deploy.yaml
```diff
@@ -21,3 +21,4 @@ builder:
         target:
           destination: EXAMPLE
           url: https://target.example
+          sersvice: /bc/my/service
```
tobiasqueck, line 24: That test shouldn't be working... or do you have a magic string auto-correction that converts `sersvice` to `service` ;)

packages/deploy-tooling/test/unit/cli/index.test.ts
```diff
@@ -2,20 +2,25 @@ import type { ParseOptions } from 'commander';
...
-import { mockedUi5RepoService } from '../../__mocks__';
+import { mockedUi5RepoService, mockedProvider } from '../../__mocks__';
 import { Command } from 'commander';
 import fs from 'fs';
 import { ToolsLogger } from '@sap-ux/logger';
 
 describe('cli', () => {
     const fixture = join(__dirname, '../../test-input/');
     const target = 'https://target.example';
+    const env = process.env;
```
tobiasqueck, line 13: Most likely no problem but I sometimes have weird behaviors with assignments like that if I don't put them into `beforeAll`. Could you please move it there, it would then also match the corresponding call in `afterAll`?

tobiasqueck (changes requested):
- changeset good
- code adheres to coding standards
- a few minor open questions that might require changes

packages/deploy-tooling/README.md
longieirl: Done.

packages/deploy-tooling/test/unit/cli/index.test.ts
longieirl: Done.

tobiasqueck (approved):

cianmSAP (approved):
- code changes as expected per bug description
- test updated / added appropriately

Issue #1089, opened 2023-06-30
TBI - Secure storage is not being read when creating a service provider instance for an on-premise ABAP system


longieirl:
### Description (include screenshots)
Extend the `createAbapServiceProvider` method to support both credentials and service keys if found in the system secure store.

Existing functionality;
https://github.com/SAP/open-ux-tools/blob/main/packages/deploy-tooling/src/base/deploy.ts#L106

### Value
Provides a secure way to deploy to ABAP using service keys.
Provides an alternative to the user to deploy to an on-premise ABAP system. 

### Architecture Elaboration
No

### Notes
Found when testing on-premise ABAP deplyment where the secure store had the service keys for the target URL.

### Tasks
- [ ] Update `createAbapServiceProvider` to determine if the secure store also has service keys, if so, then generate a provider sinstance using these keys
- [ ] Update test cases

tobiasqueck:
I don't think we should do that because `createAbapServiceProvider` is only called if the configuration defines that it is not a cloud system. If it is a cloud system then we execute `createAbapCloudServiceProvider` (which is also checking the store for credentials). 

What we could and should do instead, is to check in both functions if the returned system is of the incorrect type, and if yes, throw an exception informing the user that either the deployment configuration or the stored system information are wrong. We could even recommend how to fix it.

tobiasqueck:
If I am not mistaken, then this one is also done, correct @cianmSAP @longieirl ?

cianmSAP:
Yes, this is complete. Closing

PR #1090, opened 2023-06-30
fix(deploy-tooling): Update createAbapServiceProvider to support service keys

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/1089

- tested locally in vscode reading secure store
- updated test cases

Fiori UI project with scp: false but secure storage has service keys available for the onpremise ABAP system;
Screenshot 2023-06-30 at 12 44 03

### DIFF
--- a/packages/deploy-tooling/src/base/deploy.ts
+++ b/packages/deploy-tooling/src/base/deploy.ts
@@ -23,7 +23,7 @@ import type { AbapDeployConfig, UrlAbapTarget } from '../types';
 import { getConfigForLogging, isUrlTarget } from './config';
 import { promptConfirmation, promptCredentials, promptServiceKeys } from './prompt';
 
-type BasicAuth = Required<Pick<BackendSystem, 'username' | 'password'>>;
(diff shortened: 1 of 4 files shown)


packages/deploy-tooling/src/base/deploy.ts
```diff
@@ -101,26 +101,45 @@ async function createAbapCloudServiceProvider(
...
-        const storedOpts = await getCredentials<BasicAuth>(target);
-        if (storedOpts?.password) {
-            options.auth = {
-                username: storedOpts.username,
-                password: storedOpts.password
-            };
+        const storedOpts = await getCredentials(target);
+        if (storedOpts) {
+            let systemName = '';
+            if ((storedOpts as BasicAuth).password) {
```
tobiasqueck, line 121: Could you please create type guard so that you don't need to `as BasicAuth` so often (makes the code unnecessarily complex)

longieirl:
Closing for now based on feedback.

PR #1094, opened 2023-07-11
fix: cds v7 dynamic loading

voicis:
Similar change to #1043.

Fixed issue where `getCapModelAndServices` was crashing when used in project with `@sap/cds` version 7.

Starting with version `7.0.0` of `@sap/cds` functions are only available under `default` export. In order try avoiding similar issues I created a utility function that properly loads the CDS module.

### DIFF
--- a/.changeset/tricky-panthers-bathe.md
+++ b/.changeset/tricky-panthers-bathe.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': patch
+---
+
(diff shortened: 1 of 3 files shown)


mr25mr:
Issue: https://github.com/SAP/open-ux-tools/issues/1096

devinea (commented):
approach looks good. Compatible across CDS versions. 
Lint fix needed.

devinea (approved):
thanks. lg.

PR #1095, opened 2023-07-11
tbi(deploy-tooling): improve abap service provider error handling

cianmSAP:
https://github.com/SAP/open-ux-tools/issues/1089#issuecomment-1614595199

In the case the config does not specify `--cloud` and the target returns a abap cloud system, return an informative error.

### DIFF
--- a/packages/deploy-tooling/src/base/deploy.ts
+++ b/packages/deploy-tooling/src/base/deploy.ts
@@ -119,6 +119,11 @@ async function createAbapServiceProvider(
                 password: storedOpts.password
             };
         }
+        if ((storedOpts as unknown as ServiceAuth)?.serviceKeys) {
+            throw new Error(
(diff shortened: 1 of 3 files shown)


packages/deploy-tooling/src/base/deploy.ts
```diff
@@ -310,7 +315,7 @@ async function tryDeploy(
                 'Deployment in TestMode completed. A successful TestMode execution does not necessarily mean that your upload will be successful.'
             );
         } else {
-            logger.info('Successfully deployed.');
+            logger.info('Deployment Successful.');
```
cianmSAP, line 318: There was also an ask to revert to this statement, I can move to a separate PR if required but as it is so small ...

tobiasqueck (approved):
* clean and easy to read coding
* properly tested
* changeset fits the change
* did not test locally but the change looks straight forward

docirl (approved):

Issue #1098, opened 2023-07-13
TBI - Add functionality to project-access that allows converting a CDS model to EDMX


Klaus-Keller:
### Description
In order to enable tools to work with CAP projects, it is helpful to have a converter functionality that compiles CDS data to EDMX format. This is beneficial for tools, that do not work directly on CDS sources and allows them to work with CAP project in the same manner, as with RAP projects where EDMX is provided by a backend

### Value
A subset of tools do not need to handle EDMX and CDS data structures.

### Architecture Elaboration
No architecture elaboration required.

### Tasks
- [x] Provide functionality to convert CDS project to EDMX format

PR #1099, opened 2023-07-13
fix(project-access): function to convert cds to edmx

Klaus-Keller:
## Issue
https://github.com/SAP/open-ux-tools/issues/1098

## Description
Add 
```typescript 
async function readCapServiceMetadataEdmx(root: string, uri: string, version: 'v2' | 'v4' = 'v4'): Promise<string>
```
to module `@sap-ux/project-access`, which allows to convert a given service by `uri` from a CDS project into [EDMX](http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html).

This allows to have tools working in the same manner on project using EDMX and projects using CDS as data description format.

### DIFF
--- a/.changeset/grumpy-maps-press.md
+++ b/.changeset/grumpy-maps-press.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': minor
+---
+
+Add functionality to convert CDS into EDMX
--- a/packages/project-access/src/index.ts
+++ b/packages/project-access/src/index.ts
@@ -13,6 +13,7 @@ export {
     isCapJavaProject,
     isCapNodeJsProject,
     loadModuleFromProject,
+    readCapServiceMetadataEdmx,
     readUi5Yaml
 } from './project';
 export { getFilePaths } from './file';
--- a/packages/project-access/src/project/cap.ts
+++ b/packages/project-access/src/project/cap.ts
@@ -151,8 +151,8 @@ function findServiceByUri(
     services: { name: string; urlPath: string }[],
     uri: string
 ): { name: string; urlPath: string } | undefined {
-    const searchUri = uri.replace(/[\\\/]/g, '/'); // replace all \ and / in service uri due to issues on Windows with backslashes
-    return services.find((srv: { name: string; urlPath: string }) => srv.urlPath.replace(/[\\\/]/g, '/') === searchUri);
+    const searchUri = uri.replace(/[\\/]/g, '/'); // replace all \ and / in service uri due to issues on Windows with backslashes
+    return services.find((srv: { name: string; urlPath: string }) => srv.urlPath.replace(/[\\/]/g, '/') === searchUri);
 }
 
 /**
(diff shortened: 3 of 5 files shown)


devinea (approved):
Thanks @Klaus-Keller

This will be useful. 
Code changes, tests and changeset all look good.
What about exporting `loadCdsModuleFromProject` and related interfaces?

Klaus-Keller:
> What about exporting `loadCdsModuleFromProject` and related interfaces?

@devinea, I had the same idea and discussed it with @voicis. Our conclusion: to have some sort of control over what is used on consumer side it might be best, if we are a bit more conservative and expose the specific functionality only,  rather than the complete cds object. The restriction should help with backward compatibility later on, we can try to keep our functions stable and do not needs to modify consumers.

devinea:
> > What about exporting `loadCdsModuleFromProject` and related interfaces?
> 
> @devinea, I had the same idea and discussed it with @voicis. Our conclusion: to have some sort of control over what is used on consumer side it might be best, if we are a bit more conservative and expose the specific functionality only, rather than the complete cds object. The restriction should help with backward compatibility later on, we can try to keep our functions stable and do not need to modify consumers.

@Klaus-Keller yeah that is a reasonable approach. 
We probably need to expose a method for `cds.resolve`, `cds.parse`, `cds.link` and also compiler loading from the project.

Klaus-Keller:
>We probably need to expose a method for cds.resolve, cds.parse, cds.link and also compiler loading from the project.

@devinea 
Would be good if we find an abstraction in a way, that the signature and return values of function calls are not too complex cds interfaces, but more primitive types.

kranthie-sap (approved):
* New function to read CAP service metadata and finding service by URI looks good
* Good test coverage and sufficient test cases
* Change set maintained

IainSAP:
@Klaus-Keller Can we please expose the package directly also. Allow the consumer to decide if they want to access the package directly or not rather than only exposing a facade which needs to be maintained and documented.

Klaus-Keller:
> @Klaus-Keller Can we please also expose the package directly. Allow the consumer to decide if they want to access the package directly or not rather than only exposing a new API. Restricting access in this way is constraining consumers unnecessarily. We cannot predict how consumers will want to use the cds package from the project?

See discussion thread above. We intentionally want to have some sort of abstraction layer and NOT return the reference to the module due to possibly incompatible changes in `@sap/cds`. That way we can (at least try) to keep the exposed functions stable.

IainSAP:
> > @Klaus-Keller Can we please also expose the package directly. Allow the consumer to decide if they want to access the package directly or not rather than only exposing a new API. Restricting access in this way is constraining consumers unnecessarily. We cannot predict how consumers will want to use the cds package from the project?
> 
> See discussion thread above. We intentionally want to have some sort of abstraction layer and NOT return the reference to the module due to possibly incompatible changes in `@sap/cds`. That way we can (at least try) to keep the exposed functions stable.

Yes I have read the discussion. I think we should also return the underlying cds package instance as this allows access according to the cds apis, even if they change between versions.

Issue #1100, opened 2023-07-14
BUG - UAA url with trailing slash causes error in deployment 


cianmSAP:
### Related Feature
Feature request: https://github.com/SAP/open-ux-tools/issues/1029

### Description
In some instances, the UAA url may be passed to the axios-extension with a trailng slash, resulting in a double slash in the url and causing errors later (e.g in deployment)

### Steps to Reproduce
1. Attempt to deploy aplication using `deploy-tooling` and for the `--uaa-url` include a trailing slash
2. See error

### Expected results
Application deploys

### Actual results
Application fails to deploy

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1101, opened 2023-07-14
fix(axios-ext): remove trailing slash in uaa url

cianmSAP:
#1100

The token request config hardcodes a `/` at the end of the uaa-url, e.g ```url: `${this.url}/oauth/token```, with the issue arising when the uaa-url passed to the axios-extension already has a trailing slash


tobiasqueck (approved):
* changeset ok
* code ok
* test ok

Locally tested the changed line but not the whole project.

docirl (approved):
Code change and test looks good, not tested locally.

PR #1102, opened 2023-07-14
fix(project-access): handle backslash and leading/trailing slashes urlPath

Klaus-Keller:
## Issue
https://github.com/SAP/open-ux-tools/issues/1098

## Description
Small fix for internal function `findServiceByUri()`. When searching for service by URI we should ignore leading and trailing slashes.


devinea (approved):
Code changes, tests and changset looks as expected.

packages/project-access/src/project/cap.ts
```diff
@@ -151,8 +151,9 @@ function findServiceByUri(
     services: { name: string; urlPath: string }[],
     uri: string
 ): { name: string; urlPath: string } | undefined {
-    const searchUri = uri.replace(/[\\/]/g, '/'); // replace all \ and / in service uri due to issues on Windows with backslashes
-    return services.find((srv: { name: string; urlPath: string }) => srv.urlPath.replace(/[\\/]/g, '/') === searchUri);
+    const uniformUri = (uri: string) => uri.replace(/[\\/]/g, '/').replace(/(?:^\/)|(?:\/$)/g, '');
```
slin-sap, line 154: IMO, the original comment should be kept and extended to explain the 2nd replace is used to remove leading and trailing slashes.

Klaus-Keller: changed it and added comment

devinea (approved):
Thanks @Klaus-Keller 
Changes to code and tests look good. 
Changeset ✅

slin-sap (approved):
- Understand the proper fix will be required on cds side.
- Fix lgtm
- patch level is justified since there is no API level change.

Issue #1103, opened 2023-07-17
BUG - Unable to fetch transport list by using ADT service


slin-sap:
### Related Feature
Feature request:  #518
Related fix: #1083

### Description
When package name contains namespace e.g. "/A/B/NAME", ADT service API fails to fetch transport list.

### Steps to Reproduce

1. Enter a valid package name in the format "/A/B/NAME".
1. Expand to get the list

### Expected results
Transport list is fetched successfully.

### Actual results

Failure to fetch transport requests, error `URI-Mapping cannot be performed due to invalid URI:
`
### Screenshots

Package name can also have namespace, which requires uriencoding:
![Screenshot 2023-07-17 at 9 30 11 AM](https://github.com/SAP/open-ux-tools/assets/98066146/07dddfe9-6f4c-49d0-b75c-51fdbbce4813)

### Problem
- Customer didn't upgrade Fiori generator version for testing. Neither this fix nor #1083 was tested by customer. That's why list transport request never worked. Otherwise #1083 should fix the problem as validated during customer call.
- Customer also mentioned package name querying was not working and then started working. This is because customer uses destination with NoAuthentication. We had a bug that Deploy Config Generator fails to detect 401 Unauthorized error when running standalone and thus not showing username/password questions. As a result, user sometimes can query package name when there is existing authenticated http session, but fails otherwise since the ADT queries were not authenticated. This bug has been fixed in our internal tools-suite repo.

### Fix
After validating the scenario with customer, we can confirm that only fix introduced by #1083 is required. Thus the origin fix introduced by this GH issue needs to be removed by PR #1220.

### Why was it missed
Specific scenario that package name and app name contain namespace was not considered in the original requirements.

### How can we avoid this
Added unit test to make sure ADT request body has uri encode for app name and no uri encode for package name.

tobiasqueck:
@slin-sap it seems to be fixed, could you please check and close the issue.

slin-sap:
@tobiasqueck A fix has been applied based on the WebIDE implementation. But it didn't resolve the customer incident related to this bug. We are waiting for ADT colleague to further investigate the ADT service response data.

PR #1104, opened 2023-07-17
fix: encode package name

slin-sap:
#1103

Add uri encoding to package name


devinea (approved):
code, test and changeset look appropriate. 
I haven't tested locally.

Issue #1112, opened 2023-07-19
FEATURE - create transport request during deployment 


cianmSAP:
_As a **developer of fiori apps**,
I want to **create transport requests during deployment**,
so that I **only create them when necessary i.e the app is actually being deployed**_

### Description

In order to prevent users continually creating TR's during deployment configuration that they may not use. It was agreed to allow the user to specify that a TR should be created during deployment, without it being created before that.

### Technical Design
Updates to deploy tooling required. A new parameter `--create-transport` will be available.

### Acceptance Criteria

Given I choose to create my TR during deployment, when I try and deploy my application, the TR will get created as pat of this, and the following log message will be displayed which includes the new TR created:
Transport request XXXXXXX created for deployment.

Given I choose to create my TR during deployment, when I try and deploy my application, if the TR could not be created, I will see the following message and deployment will stop:
Transport request could not be created. Please create it manually and re-run deployment configuration for this project

### Tasks
- [ ] Update deploy tooling to create a new TR using axios-extension
- [ ] Unit Tests added
- [ ] Test Cases defined

cianmSAP:
Also needs to be enabled for undeployment

tobiasqueck:
Closing this one since it is done.

PR #1113, opened 2023-07-19
feat(deploy-tooling): create transport request during deployment

cianmSAP:
#1112

Adds new functionality to create transport requests during deployment. This can be done by passing the `--create-transport` flag to the deploy command
Also exposes the logic as a standalone function `createTransport`


packages/deploy-tooling/src/base/deploy.ts
```diff
@@ -242,18 +253,25 @@ async function handleAxiosError(
...
             const credentials = await promptCredentials(service.defaults.auth?.username);
             if (Object.keys(credentials).length) {
                 service.defaults.auth = credentials;
-                await deploymentCommands[command](service, config, logger, archive);
+                await deploymentCommands[command](provider, config, logger, archive, configPath);
                 return true;
             } else {
                 return false;
             }
         case 412:
```
slin-sap, line 264: 1. You might want to extract handler for each error code to a separate function. This handleAxiosError function might keep growing and having increased complexity. 

2. It is not from your change. But I want to ask why there is not log at all for unhandled errors (default case).

cianmSAP: I've extracted the functions for 401 & 412. I think the default returns false as this is more of retry handler and there is a general log in the calling function `handleError`. I've renamed to `axiosErrorRetryHandler` which may give it a better indication to the functions purpose

packages/deploy-tooling/src/base/deploy.ts
```diff
@@ -281,43 +353,123 @@ async function getAbapService(config: AbapDeployConfig, logger: Logger): Promise
...
+            const ui5DeployConfig = await UI5Config.newInstance(fs.read(deployYamlPath));
+            const abapTarget = {
+                ...config.target,
+                scp: config.target.cloud
+            } as AbapTarget;
+            ui5DeployConfig.updateAbapDeployTask(abapTarget, config.app);
+            fs.write(deployYamlPath, ui5DeployConfig.toString());
+            return new Promise((resolve) => {
+                fs.commit(resolve);
+                logger.info('Updated the ui5-deploy.yaml with the new transport request');
```
longieirl, line 378: Even though Fiori tools uses ui5-deploy.yaml, there is nothing stopping a developer from using any filename when deploying to ABAP.

packages/deploy-tooling/src/cli/index.ts
```diff
@@ -60,6 +60,7 @@ export function createCommand(name: 'deploy' | 'undeploy'): Command {
         command
             .option('--package <abap-package>', 'Package name for deploy target ABAP system')
             .option('--description <description>', 'Project description of the app')
+            .option('--create-transport <description>', 'Create transport request during deployment')
```
longieirl, line 63: I'm wondering if we need to change the behaviour;

if `--create-transport` is passed as a cli param, then a new TR is generated whether or not the config passed in contains the bookmark `CREATE_TR_DURING_DEPLOY`.

longieirl, line 63: Dont forget to update the readme with this param too.

cianmSAP: Yep just saw this, pushing update

packages/deploy-tooling/src/base/deploy.ts
```diff
@@ -281,6 +326,47 @@ async function getAbapService(config: AbapDeployConfig, logger: Logger): Promise
...
 
+/**
+ * Creates a new transport request using adt service.
+ *
+ * @param config - deployment configuration
+ * @param logger - reference to the logger instance
+ * @param provider - instance of the axios-extension abap service provider (it is created if not passed in)
+ * @returns transportRequest
+ */
+export async function createTransportRequest(
```
longieirl, line 337: This method can be consumed in Fiori deployment generator to replace the existing functionality around generating the TR value.
FYI @slin-sap

packages/deploy-tooling/src/cli/index.ts
longieirl: `--create-transport` should have a conflict added with `--transport` as both are mutually exclusive.

cianmSAP: Updated

longieirl (dismissed):
Tests look good
Changes include a cleanup of code
New TR functionality exposed and be consumed by other parties
Havent tested locally

packages/deploy-tooling/src/base/deploy.ts
```diff
@@ -328,7 +418,7 @@ async function tryDeploy(
             logger.info('Deployment Successful.');
         }
     } catch (error) {
-        await handleError(TryCommands.Deploy, error, service, config, logger, archive);
+        await handleError(TryCommands.Deploy, error, provider, config, logger, archive);
```
IainSAP, line 421: Do we need to rollback the TR creation if we get here. Or maybe the TR should not be created until after deployment? We dont seem to check or use the returned TR...

IainSAP (commented):
Just one query...thx @cianmSAP

packages/deploy-tooling/src/base/deploy.ts
cianmSAP: The newly created TR is assigned to the `config.app.transport` and used in the deployment. But it does raise a good point, if we have successfully created a TR but the deployment to the UI5 Abap Repo failed (due to auth issues for example), we shouldn't create a second TR. 
I've updated to set `config.createTransport` to false when a TR is created succesfully

packages/deploy-tooling/src/base/deploy.ts
```diff
@@ -281,6 +326,48 @@ async function getAbapService(config: AbapDeployConfig, logger: Logger): Promise
...
+        const createTransportParams = {
+            packageName: config.app.package ?? '',
+            ui5AppName: config.app.name,
+            description: 'Created by @sap-ux/deploy-tooling'
+        };
+
+        try {
+            const transportRequest = await adtService.createTransportRequest(createTransportParams);
+            if (transportRequest) {
+                config.app.transport = transportRequest;
```
IainSAP, line 357: Can we simply return the TR, it is not expected that the config is updated based on the jsdoc?

cianmSAP: I've updated to just return the TR and assign it to the config in the consuming function

packages/deploy-tooling/src/base/deploy.ts
```diff
@@ -295,29 +380,39 @@ async function runCommand(
...
+    provider: AbapServiceProvider,
     config: AbapDeployConfig,
     logger: Logger,
     archive: Buffer
 ): Promise<void> {
+    const service = getUi5AbapRepositoryService(provider, config, logger);
+
+    if (config.createTransport) {
+        const transportRequest = await createTransportRequest(config, logger, provider);
+        if (transportRequest) {
```
IainSAP, line 410: 👍 thx

IainSAP (dismissed):
Thx @cianmSAP 

-  Good test coverage
-  Did not test manually
-  Prevents multiple TR creation in same run

packages/deploy-tooling/src/cli/index.ts
```diff
@@ -60,6 +60,9 @@ export function createCommand(name: 'deploy' | 'undeploy'): Command {
         command
             .option('--package <abap-package>', 'Package name for deploy target ABAP system')
             .option('--description <description>', 'Project description of the app')
+            .addOption(
+                new Option('--create-transport', 'Create transport request during deployment').conflicts(['transport'])
```
longieirl, line 64: The `createCommand` unit test can be extended to ensure the conflict is validated.

cianmSAP: Updated unit test

packages/deploy-tooling/src/base/deploy.ts
longieirl: If a 401 error is returned then the TR wouldnt have been created either?

packages/deploy-tooling/test/unit/cli/index.test.ts
```diff
@@ -180,7 +180,8 @@ describe('cli', () => {
         test.each([
             { params: ['--url', '~url', '--destination', '~dest'] },
             { params: ['--client', '001', '--destination', '~dest'] },
-            { params: ['--cloud', '--destination', '~dest'] }
+            { params: ['--cloud', '--destination', '~dest'] },
+            { params: ['--transport', '~transport', '--create-transport', 'true'] }
```
longieirl, line 184: Thank you.

longieirl (dismissed):
Previous comment applies. Thanks

packages/deploy-tooling/src/base/deploy.ts
```diff
@@ -281,6 +326,46 @@ async function getAbapService(config: AbapDeployConfig, logger: Logger): Promise
...
+    logger: Logger,
+    provider?: AbapServiceProvider
+): Promise<string | void> {
+    if (!provider) {
+        provider = await getAbapServiceProvider(config, logger);
+    }
+
+    const adtService = await provider.getAdtService<TransportRequestService>(TransportRequestService);
+    if (adtService) {
+        const createTransportParams = {
```
slin-sap, line 348: Can you add tsdoc @throws (https://tsdoc.org/pages/tags/throws/) to indicate deploy is not triggered if createTransportRequest failed with error.

cianmSAP: Updated

packages/deploy-tooling/src/base/deploy.ts
```diff
@@ -281,6 +326,47 @@ async function getAbapService(config: AbapDeployConfig, logger: Logger): Promise
...
+    config: AbapDeployConfig,
+    logger: Logger,
+    provider?: AbapServiceProvider
+): Promise<string | void> {
+    if (!provider) {
+        provider = await getAbapServiceProvider(config, logger);
+    }
+
+    const adtService = await provider.getAdtService<TransportRequestService>(TransportRequestService);
+    if (adtService) {
```
slin-sap, line 348: Return type should be Promise<string> instead of Promise<string|void>

You have the void case because the full execution logic is in `if (adtService) {...}`. If this branch is not executed, void type is expected. IMO, you can throw error when adtService is not available, because it is the same scenario as create transport request cannot be succeeded and the subsequent deploy cannot continue.

slin-sap (changes requested):

packages/deploy-tooling/src/base/deploy.ts
longieirl: @slin-sap what is the reason why `getAdtService` would return null? Surely that should throw an exception if not found? As you cant continue if not found.

longieirl:
@slin-sap made some changes based on our conversation. I need to test locally that everything is working as expected.

packages/deploy-tooling/src/base/deploy.ts
longieirl: That should be handled now.

longieirl:
Unable to create TR, correctly handled and error shown; in this instance was a 503 exception thrown.
Screenshot 2023-07-25 at 15 05 55

longieirl:
Handling 401 correctly as well;
Screenshot 2023-07-25 at 15 43 44

@slin-sap I'm wondering if we need to change the error message since in this case, a TR is not required.

longieirl:
Command executed for a cloud ABAP target system, however, the ADT Service couldnt be retrieved;
```
const adtService = await provider.getAdtService<TransportRequestService>(TransportRequestService);
```
This is the command executed;
```
./bin/deploy --url https://myapi-abap.ondemand.com --cloud --name ZTRA20161633 --description 'Travel App' --create-transport --package 'MY_PACKAGE' --verbose true --archive-folder /Users/Documents/tools-suite-testmyapp/dist
```
Flow with verbose enabled
Screenshot 2023-07-25 at 16 36 36

longieirl:
End to end flow validated against a host exposing the ADT service;

Confirmed ADT service first before deploying;
```
curl -X GET -i -H 'X-Csrf-Token:Fetch' http://myabap-server.com:50000/sap/bc/adt/discovery > output.txt
```

Deploy command pointing to ABAP system with ADT service exposed;
Screenshot 2023-07-25 at 17 02 09

longieirl (approved):
I'm approving as I've tested locally and its working as expected.

I've made minor changes to the PR so it needs reviews from other developers.

slin-sap (approved):
- Code changes look good to me
- Previous raised change requests have been addressed

PR #1120, opened 2023-07-24
fix(project-access): add readonly for WorkspaceFolder

zdravko-georgiev:
Adds `readonly` for `WorkspaceFolder` in methods that are using `WorkspaceFolder` as input parameter. This is explicitly required when using newer versions of `@types/vscode`.


johannes-kolbe (approved):
- [x] Code changes look good
- [x] changeset provided

Issue #1121, opened 2023-07-24
BUG - no-strict-ssl not deactivating strict-ssl from the CLI


longieirl:
### Description
When appending the `--no-strict-ssl` flag to a CLI command, its throwing the error;

```
error: unknown option '--no-strict-ssl'
(Did you mean --strict-ssl?)
```
The readme.md is showing the CLI param should be supported;
```
.option('--strict-ssl', 'Perform certificate validation (use --no-strict-ssl to deactivate it)')
```

### Steps to Reproduce
1. Deploy Fiori UI application
2. Run the undeploy command;
```
"undeploy": "undeploy --config ui5-deploy.yaml --verbose --no-retry --no-strict-ssl",
```

### Expected results

strict-ssl is disabled

### Actual results

Command error shown and not able to disable SSL

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Merged locally and validated.

PR #1122, opened 2023-07-24
fix(deploy-tooling): handle toggling of strict-ssl

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/1121

- updated test cases

Negating options
https://github.com/tj/commander.js/blob/master/examples/options-negatable.js


packages/deploy-tooling/src/cli/index.ts
```diff
@@ -46,6 +46,7 @@ export function createCommand(name: 'deploy' | 'undeploy'): Command {
         .option('--transport <transport-request>', 'Transport number to record the change in the ABAP system')
         .option('--name <bsp-name>', 'Project name of the app')
         .option('--strict-ssl', 'Perform certificate validation (use --no-strict-ssl to deactivate it)')
+        .option('--no-strict-ssl', 'Deactivate certificate validation')
```
Klaus-Keller, line 49: Do we really need both, `--strict-ssl` and `--no-strict-ssl`? 
Should not one be the default and the other overwrites it? For security relevant things like this I'd say, strict ssl certificate validation is the default and by using option `--no-strict-ssl` you can deactivate it.

Klaus-Keller (commented):
Added a question.
Additionally, I guess we also need to update the `README.md`.

devinea (approved):
Changes, changeset, tests and doc updates look good.

Klaus-Keller (approved):
- changes and test looks good
- did a manual test of `--no-strict-ssl`, no issue found

Approved from my side.

Issue #1128, opened 2023-07-27
BUG - blocked from deploying to cloud target system


longieirl:
### Related Feature
Related to issue https://github.com/SAP/open-ux-tools/issues/1116

### Description
As part of Fiori tools, we support a UAA flow;
```bash
npx fiori deploy --url https://myapi.hanavlab.ondemand.com/ 
--name <name> 
--package <package-name> 
--transport <transport> 
--archive-path '<path-to-zip>' 
--uaa-url https://myauthapp.sap.hana.ondemand.com/oauth/token 
--uaa-username '<email-address>' 
--uaa-password '<invalid-password>' 
--uaa-clientid 'sb-xs-658b' 
--uaa-clientsecret '67ac1e4b' 
--scp --noConfig --verbose
```
Which is now failing with a 401;
```
2023-07-27 16:05:50 error   abap-deploy-task: Change logging level to debug your issue
        (see examples https://github.com/SAP/open-ux-tools/tree/main/packages/deploy-tooling#configuration-with-logging-enabled) 
Request failed with status code 401
```

The verbose logging is showing;
```
credentials: 'hidden',
```
which indicates the object is populated and this is blocking the request.

### Steps to Reproduce
1. Deploy a project to a cloud system 

Deploy tooling CLI command;
```
./bin/deploy --name ZJLONG0205 --package '$tmp' --transport '' --url 'https://myapi.hanavlab.ondemand.com/' --cloud true --cloud-service-env
```
where the .env file is;

```
SERVICE_CLIENT_ID=sb-xs-658b
SERVICE_CLIENT_SECRET=67ac1e4b
SERVICE_UAA_URL=https://myapi.hanavlab.ondemand.com/
SERVICE_USERNAME=<email-address>
SERVICE_PASSWORD=<invalid-password>
```

### Expected results

Deployment request to a cloud target system is successful.

### Actual results

Getting 401 due to invalid credentials.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

tobiasqueck:
@longieirl is this fixed with #1129

[...]

PR #1129, opened 2023-07-27
fix(deploy-tooling): Handle cli credentials

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/1128

- Updated test cases
- Applied changset
- Reverting the `process.env.SERVICE_USERNAME` code since the correct way to use dotenv is to explicity request it using the `--username` cli param and unblock cloud target deployments which use the SERVICE_USERNAME env variable
- Correct flow, which is currently supported and is not impacted;

Sample .env file
```
XYZ_USER=<my_username>
XYZ_PASSWORD=<my_password>
```

Sample CLI deploy command
```bash
./bin/deploy --name ZJLONG0205 --package '$tmp' --transport '' --url 'https://myapi.hanavlab.ondemand.com/' --username  env:XYZ_USER  --password  env:XYZ_PASSWORD
```

### DIFF
--- a/.changeset/poor-lions-serve.md
+++ b/.changeset/poor-lions-serve.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/deploy-tooling': patch
+---
+
+fix issue cloud target systems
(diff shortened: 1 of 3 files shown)


cianmSAP (approved):
- Code correction makes sense to me
- Test added
- Tested locally and resolves issue

Issue #1131, opened 2023-07-28
BUG - Cannot authenticate during deployment flow for cloud systems


cianmSAP:
### Related Feature
Feature request: https://github.com/SAP/open-ux-tools/issues/1029

### Description
When deploying to a cloud system using uaa credentials, if the credentials are missing or incorrect, the user is prompted with to enter their credentials. When the correct credentials are entered, they are not accepted.

### Steps to Reproduce

Deploy a project to a cloud system
Deploy tooling CLI command;
```
./bin/deploy --name ZJLONG0205 --package '$tmp' --transport '' --url 'https://myapi.hanavlab.ondemand.com/' --cloud true --cloud-service-env
```
where the .env file is;
```
SERVICE_CLIENT_ID=sb-xs-658b
SERVICE_CLIENT_SECRET=67ac1e4b
SERVICE_UAA_URL=https://myapi.hanavlab.ondemand.com/
SERVICE_USERNAME=<email-address>
SERVICE_PASSWORD=<invalid-password>
```

### Expected results
When correct credential are entered the deployment is successful 

### Actual results
Prompt for credentials is redeatedly shown

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1132, opened 2023-07-28
fix(deploy-tooling): handle 401 error for cloud target systems

cianmSAP:
#1131

When a 401 error is thrown in deployment, the user is prompted for credentials. 
In the case where a serviceKey is used, the uaa params need to be updated


packages/deploy-tooling/test/unit/base/deploy.test.ts
```diff
@@ -200,6 +200,31 @@ describe('base/deploy', () => {
...
+                        clientsecret: '~clientsecret',
+                        url: '~url'
+                    }
+                },
+                url: '~targetUrl'
+            } as AbapTarget;
+            mockedUi5RepoService.deploy.mockRejectedValueOnce(axiosError(401));
+            await deploy(archive, { app, target: cloudTarget, yes: true }, nullLogger);
+            prompts.inject(['~uaa-username', '~uaa-password']);
+            expect(mockedUi5RepoService.deploy).toBeCalledWith({
```
longieirl, line 220: Wondering if we should be more explicit what is actually changed here i.e. the serviceKey.uaa object with the updated values. Its not clear from the test?

cianmSAP: Good point. I've updated the test

longieirl: Thank you.

longieirl (approved):

Klaus-Keller (approved):
- code changes are straight forward and easy to understand
- Test coverage is excellent
- Did not locally build and test, just visual review

Approved from my side.

PR #1134, opened 2023-08-01
feat(deploy-tooling): create transport for undeployment

cianmSAP:
#1112

### DIFF
--- a/packages/deploy-tooling/README.md
+++ b/packages/deploy-tooling/README.md
@@ -100,12 +100,12 @@ Options:
   --cloud-service-env                  Load ABAP cloud service properties from either a .env file or your environment variables
   --username                           ABAP Service username
   --password                           ABAP Service password
-  --create-transport                   Create a transport request during deployment
   --transport <transport-request>      Transport number to record the change in the ABAP system
+  --create-transport                   Create a transport request during deployment
+  --package <abap-package>             Package name for deploy target ABAP system (only required when --create-transport is used)
   --name <bsp-name>                    Project name of the app
   --no-strict-ssl                      Deactivate SSL certificate validation, enabled by default
   --test                               Run in test mode. ABAP backend reports undeployment errors without actually undeploying (use --no-test to deactivate it).
-  --package <abap-package>             Package name for deploy target ABAP system (only required when --create-transport is used)
   -v, --version                        version of the deploy tooling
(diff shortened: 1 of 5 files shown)


packages/deploy-tooling/README.md
```diff
@@ -101,6 +101,8 @@ Options:
   --username                           ABAP Service username
   --password                           ABAP Service password
   --transport <transport-request>      Transport number to record the change in the ABAP system
+  --create-transport                   Create a transport request during deployment
+  --package <abap-package>             Package name for deploy target ABAP system (only required when --create-transport is used)
```
tobiasqueck, line 105: The `package` shouldn't be required for undeploy. The deployed app is already part of a package and undeployment doesn't need to know which on (I think).

packages/deploy-tooling/src/cli/index.ts
```diff
@@ -43,7 +43,11 @@ export function createCommand(name: 'deploy' | 'undeploy'): Command {
                 'Read ABAP cloud service properties from environment variables or .env file'
             ).conflicts(['cloudServiceKey', 'destination'])
         )
+        .option('--package <abap-package>', 'Package name for deploy target ABAP system')
```
tobiasqueck, line 46: Same as above, I don't think it is used.

tobiasqueck (changes requested):

packages/deploy-tooling/README.md
cianmSAP: The `package` is a required param for `createTransportRequest` 
https://github.com/SAP/open-ux-tools/blob/84579be8d6780b5133524cc66b3aab1dc68a2d83/packages/axios-extension/src/abap/adt-catalog/services/transportrequest-service.ts#L7-L12

tobiasqueck: @slin-sap you are the ADT expert here, is it an issue with the type or is a package name really required for creating a transport to undeploy an existing BSP?

tobiasqueck: Weird but ok, retracting my comment then.

tobiasqueck (approved):
* changes are consistent with the deploy command
* code is easy to read and maintain
* test cover the new functionality
* no manual tests but no doubts about it working based on the previous functionality

Issue #1135, opened 2023-08-01
BUG - Custom fragment in adaptation project preview is not loaded


tobiasqueck:
### Related Feature
Feature request: #971

### Description
I have a local ADP project and started the preview with `npm start`.
  
The local web server started and the browser was opened. But it didn’t load the files from the `changes` folder.
It tried to load it from the ABAP backend instead.
Error message:
Change 'id_1_1_addXML' could not be applied. Merge error detected while processing the JS control tree. - Error: resource xxx/changes/fragments/MyButton.fragment.xml could not be loaded from /sap/bc/ui5_ui5/sap/xxx/~C49E6D8E5E829BE93B54968D0ADEC4F6~5/changes/fragments/MyButton.fragment.xml
 
So it identified, that there are change files, but it didn’t load it from the local drive.

### Steps to Reproduce
1. Create an ADP project
2. Add a fragment or any other file (annotation, controller extension etc).
3. Start the preview

### Expected results
The fragment or other additional file is loaded

### Actual results
Error

### Screenshots
n/a

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [x] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

tobiasqueck:
Fixed with #1136

PR #1136, opened 2023-08-01
Fix for #1135

tobiasqueck:
Fix for #1135 

Local artifacts like controller extensions are not loaded because the requests are forwarded to the backend. This fix intercepts all requests that have a local file representation.

### DIFF
--- a/packages/preview-middleware/src/base/flex.ts
+++ b/packages/preview-middleware/src/base/flex.ts
@@ -6,7 +6,7 @@ import { join, parse } from 'path';
 /**
  * Structure of a flex change.
  */
-interface FlexChange {
+export interface FlexChange {
     [key: string]: string | object | undefined;
     changeType: string;
     reference: string;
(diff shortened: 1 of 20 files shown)


packages/preview-middleware/src/base/flex.ts
```diff
@@ -3,19 +3,60 @@ import type { ReaderCollection } from '@ui5/fs';
...
+        fragmentPath?: string;
+    };
+}
+
+/**
+ * Map of change type specific correction functions.
+ */
+const moduleNameContentMap: { [key: string]: (change: FlexChange) => string } = {
+    codeExt: (change) => (change.content?.codeRef ?? '').replace('.js', ''),
+    addXML: (change: any) => change.content.fragmentPath
```
zdravko-georgiev, line 25: From what I know `change.content` is always defined, even when empty. So we don't really need to check. But if you prefer to keep it, then maybe we can add the check also for the `addXML` use case?

zdravko-georgiev, line 25: `addXML: (change) => change.content.fragmentPath ?? ''`?

tobiasqueck: Yes, https://github.com/SAP/open-ux-tools/pull/1136/commits/d28d2418a8490994d4e6a6d36e4f03497cd1b281

tobiasqueck: I'd prefer the check since it is read from the filesystem and every 'expert' could modify it.

zdravko-georgiev: Thx

zdravko-georgiev: Good point 👍

zdravko-georgiev (approved):
Code changes look good and are easy to understand
Very good code coverage
Manual testing ✅

longieirl (approved):
Changes look good.
Havent tested functionality.
Good test coverage.

PR #1138, opened 2023-08-02
Correct code snippets in readme

tobiasqueck:
* correct wrong code snippets
* added another example of how to use the middleware with multiple backends or paths


zdravko-georgiev (approved):
Thanks for updating and adding the language to the code snippets.
Changes are looking good.

ullasholla (approved):
- Changes to the documentation look good
- Markdown file renders correctly
- No version bump is necessary and the PR correctly adds no changeset

Thank you

PR #1141, opened 2023-08-03
Improving the @sap-ux/create readme

tobiasqueck:
I have noticed that if you use `npm init @sap-ux` it might use a previous version that you used before, therefore, it is best to add `@latest` to ensure the latest version is used.
This PR is just an update of the `README` , however, a follow up investigation is needed to better understand why and how and maybe improve it differently.


IainSAP (approved):
- Docs only

zdravko-georgiev (approved):
Doc changes looks good. Thanks for updating

Issue #1150, opened 2023-08-08
BUG - username & password cli param should not conflict when using a destination cli param


longieirl:
### Related Feature
Related to https://github.com/SAP/open-ux-tools/issues/1116

### Description
When the destination and username cli params are used, the user is blocked from deploying.

```
./bin/deploy --name ZTRAJL01021015 --description 'Travel for JLong' --package Z_PACKAGE --transport Y11K900443 --verbose true --archive-folder ./jl01021015/dist --yes --test --destination MY_DESTINATION --username MYUSERNAME --password MYPASSWORD
```

The same flow works when there is a `.env` file and the username & password cli params are removed.

This also blocks undeploy as well.

### Steps to Reproduce
1. Generate an ABAP Fiori UI project
2. Try deploy using cli params only

### Expected results

destination and username cli params should be supported.

### Actual results
Command is blocked
```
Error on undeploy - option '--username ' cannot be used with option '--destination '
```
### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Validated locally and working as expected.

PR #1151, opened 2023-08-08
fix(deploy-tooling): Allow CLI username param with destination CLI param

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/1150

- removed username and password conflict
- cleanup test cases, they were not behaving as expected as mock was not being correctly reset

### DIFF
--- a/.changeset/cyan-mugs-compete.md
+++ b/.changeset/cyan-mugs-compete.md
@@ -1,5 +1,5 @@
 ---
-'@sap-ux/deploy-tooling': minor
+'@sap-ux/deploy-tooling': patch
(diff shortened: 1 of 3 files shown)


tobiasqueck (commented):
* removing the restriction is completely correct, I am confused that we had it in the first place
* change itself is simple an easy to understand
* thank a lot for the very thorough tests

One question: why a minor update? It is fixing a bug not adding new functionality, so I'd use `patch`

cianmSAP (approved):
- expected code changes as per bug description
- test updated / added appropriately

Issue #1153, opened 2023-08-08
BUG - Modules not loaded, even after installation


Klaus-Keller:
### Related Feature
Feature request: https://github.com/SAP/open-ux-tools/pull/1099

### Description
When reading service metadata from a CAP project, we load the `@sap/cds` module that is installed in the CAP project as node dependency. This leads to issues in case module `@sap/cds` is not installed initially. 

### Steps to Reproduce
The problem is reproducible only, if `@sap-ux/project-access` is used as part of a Visual Studio Code extension, because there is a certain cache used for `require.resolve()` (see details below). To reproduce this:
1. `@sap-ux/project-access` needs to be used in an environment, that runs and keeps a node process, like Visual Studio Code extension
2. https://github.com/SAP/open-ux-tools/blob/b827079c4baacefaede2749440cec7fb8a6548fd/packages/project-access/src/project/module-loader.ts#L8 needs to be called when the required module is not installed
3. Install the module
4. Call again `loadModuleFromProject()`

### Expected results
Module is loaded.

### Actual results
As the file is in cache marked as unavailable, module is not loaded.

### Screenshots
N/A

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

## Root Cause Analysis 

Root cause is the implementation of node's `require.resolve()`. It uses an internal `statCache` that is also filled with non-existing files, but never updated, see: https://github.com/nodejs/node/blob/d150316a8ecad1a9c20615ae62fcaf4f8d060dcc/lib/internal/modules/cjs/loader.js#L155

This issue was discussed in several tickets, but no change to node's module loading was done:
- https://github.com/nodejs/node/issues/31803
- https://github.com/nodejs/node/issues/12695

### Problem
Node module loader `cjs` has an internal file access cache that stores also non existing files and is never cleared, so installing the module and having the files available is not recognized by the cache.

### Fix
Several options:
- run `require.resolve()` in a separate process using `spawn` or `fork`
- check if the module files exist in `node_module` folder before loading, so the cache is not filled with non-existing files.

### Why was it missed
Problem occurs only if the node process that calls `loadModuleFromProject()` remains over several calls, before and after installation of respective node module. This is the case when running from a Visual Studio Code extension. 

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1154, opened 2023-08-08
fix(project-access): load module, avoid caching of non existing files

Klaus-Keller:
## Issue
https://github.com/SAP/open-ux-tools/issues/1153

## Description
Fix to avoid internal caching when calling `require.resolve()`. This pull requests implements the option, that check first if the module is installed before executing `require.resolve()` to get the module location. 
For details see description in ticket #1153

### DIFF
--- a/.changeset/lovely-plums-exist.md
+++ b/.changeset/lovely-plums-exist.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': patch
+---
+
+Fix for module loading after installation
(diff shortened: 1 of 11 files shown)


tobiasqueck (approved):
* code adheres to code quality
* long comment explaining why we need this is very helpful
* tests cover the scenarios
* did not manually test but I'd claim the function code is easy enough so that I am sure it works
* changeset looks good

815are (approved):
Checked:

1. code - easy to understand and do not see issue in code;
2. new tests are added;
3. Tested locally - reported scenario worked ok for me;

Issue #1160, opened 2023-08-09
BUG - dotenv not called when the UI5 task method is invoked


longieirl:
### Related Feature
Found when testing
https://github.wdf.sap.corp/ux-engineering/tools-suite/issues/25898

### Description
Environment variables are not loaded unless the calling client first loads the dotenv file; internally, the `task` [method](https://github.com/SAP/open-ux-tools/blob/main/packages/deploy-tooling/src/ui5/index.ts#L15) should load this without having to depend on the calling client.

### Steps to Reproduce
Call the task using the following code;
```
import { task } from '@sap-ux/deploy-tooling';
await task({ workspace, options });
```
The options has the credentials object which is not replaced with the environment variables unless dotenv is called beforehand.

This is not an issue when passing the credentials via the CLI since the dotenv is invoked using this flow.

### Expected results

dotenv is already loaded and the credentials are updated from env:my_username with the correct username

### Actual results

credentials remain as i.e. env:my_username

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1161, opened 2023-08-09
fix(deploy-tooling): Load dotenv when UI5 task is called

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/1160

- added tests to ensure dotenv is called when task is called


devinea (approved):
changeset, code and test updates look appropriate.
Thanks @longieirl

tobiasqueck (approved):
* changeset reflects the change
* change itself fixes an oversight
* tests cover the scenario

Issue #1164, opened 2023-08-10
BUG - Errors setting up the repository on Windows (for users developing on windows)


nikmace:
### Related Feature
Feature request: _issue number_

### Description
When _trying to set up the open-ux-tools_ on a **Windows machine** I am getting errors when installing with `pnpm i` and building with `pnpm run build`.

### Steps to Reproduce
1. Have a **Windows** computer.
2. Clone the **open-ux-tools repository** (start from the `main` branch).
3. **Install packages** with `pnpm i`. See error No. 1.
4. Run `pnpm run build`. See error No. 2.

### Expected results
The packages are _installed without errors_. The building of the packages gives no errors and generates no changes in the git.

### Actual results
Cannot install node_modules because of a problem with lock file.
Cannot build because of missing modules. 
Every tsconfig.json gets updated with \\\ instead of /.

### Screenshots
**Error No. 1**.
When running `pnpm i`:
![image](https://github.com/SAP/open-ux-tools/assets/39880364/91b49898-86a5-4151-bdc3-86885785ebe4)

**Error No. 2**.
Which then leads to some modules not being found. Thus, cannot build all modules and fail:
![image](https://github.com/SAP/open-ux-tools/assets/39880364/3e66d39e-5f7c-40a8-9eb6-3e3249b035c3)

Found a workaround to install and build the packages. First, delete the pnpm-lock file, `pnpm i` and finally build `pnpm run build`.
This way the packages are installed and build passes, but some files in the project get modified after the build has finished.
![image](https://github.com/SAP/open-ux-tools/assets/39880364/1655fdbd-cc35-40d7-8be4-0cd02bda7ea0)

#### Version/Components/Environment	
Add any other context about the problem here
`pnpm -v` : `8.1.0`
OS:                      
- [ ] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

devinea:
Hi @nikmace 
Thanks for letting us know about these issues. 
For Error no.2 I have created a PR with a fix.

I was not able to re-create error no.1. Can I check what version of pnpm are you using?

nikmace:
> 

Hello, as I've stated under `Version/Components/Environment` my **pnpm version** is `8.1.0`.

[...]

PR #1165, opened 2023-08-10
fix project reference postinstall windows

devinea:
fixes error no1. & no.2 in https://github.com/SAP/open-ux-tools/issues/1164

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "author": "@SAP/ux-tools-team",
     "private": true,
     "devDependencies": {
-        "@bessonovs/set-project-references": "0.0.10",
+        "update-ts-references": "2.6.1",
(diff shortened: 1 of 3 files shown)


tobiasqueck (dismissed):
* changes look as expected
* build on the mac still works as expected
* I did not test on windows but the feedback https://github.com/SAP/open-ux-tools/issues/1164#issuecomment-1673323460 is sufficient for me

Klaus-Keller (approved):
Thanks @devinea, update for the node module to update ts references looks good. Did not test on Windows. Approved from my side.

tobiasqueck (approved):
Approving again. Only change since last approval is a higher pnpm requirement which I am ok with

PR #1166, opened 2023-08-10
Allow loading RTA plugins for adaptation projects

tobiasqueck:
This PR does not contain a changeset on purpose because the change does not fix a bug or adds any new functionality that is exposed to the consumer. This is a preparation work required to be able to add new features to the `adp-tooling`. Once this is merged, PR #1183 will follow right away including a changeset.

* `adt-tooling`
    * added an `addApis` function allowing to define additional APIs for adaptation project e.g. to read and write fragments
    * moved content from `templates` to `templates/project` to allow adding other templates for RTA
* `preview-middleware`
    * enhanced the templates to support registering a callback script
    * enhanced the initialization for adaption projects to register additional routes implemented in `adt-tooling`
* `preview-middleware-client`
    * added an empty/dummy callback script (to be initially implemented in #1183)
    * enhanced the RTA types


packages/preview-middleware/templates/flp/sandbox.html
```diff
@@ -63,7 +63,7 @@
                 sap.ushell.Container.getServiceAsync('AppLifeCycle').then((serviceInstance) => {
                     serviceInstance.attachAppLoaded((event) => {
                         var oView = event.getParameter('componentInstance');
-                        sap.ui.require(["sap/ui/rta/api/startAdaptation"], async function (startAdaptation) {
+                        sap.ui.require(["sap/ui/rta/api/startAdaptation"<% if (locals.flex?.pluginModule) { %>, "<%- flex.pluginModule %>"<% } %>], async function (startAdaptation<% if (locals.flex?.pluginModule) { %>, loadPlugin<% } %>) {
```
zdravko-georgiev, line 66: In general, maybe we should introduce some general if statements and use those to differentiate between the different preview modes, hence normal, variant management, adaptation? We will have code duplication in the template, but the template will be more readable? Just an idea, what do you think?

packages/adp-tooling/package.json
```diff
@@ -28,6 +30,7 @@
     "files": [
         "LICENSE",
         "templates",
+        "rta",
```
zdravko-georgiev, line 33: Shouldn't this be `rta.d.ts`?

packages/preview-middleware/templates/flp/sandbox.html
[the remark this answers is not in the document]
tobiasqueck: if `flex` is `undefined` then `if (flex?. pluginModule)` crashes

tobiasqueck: What do you have in mind? I need an example to better understand it.

packages/adp-tooling/package.json
tobiasqueck: That is outdated, it shouldn't be there at all.

tobiasqueck: 8287c49

packages/preview-middleware/templates/flp/sandbox.html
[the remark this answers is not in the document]
zdravko-georgiev: OK, understood

packages/adp-tooling/package.json
zdravko-georgiev: Great

packages/preview-middleware/templates/flp/sandbox.html
zdravko-georgiev: This is what I have in mind
```JS
   <script>
    case 'NORMAL PREVIEW':
        sap.ui.getCore().attachInit(function (){
            sap.ushell.Container.createRenderer().placeAt("content")
        });
    break;
    case 'VARIANT':
        sap.ui.getCore().attachInit(function () {
            sap.ushell.Container.attachRendererCreatedEvent(function () {
                sap.ushell.Container.getServiceAsync('AppLifeCycle').then((serviceInstance) => {
                    serviceInstance.attachAppLoaded((event) => {
                        var oView = event.getParameter('componentInstance');
                        sap.ui.require(["sap/ui/rta/api/startAdaptation"], async function (startAdaptation) {
                            var options = {
                                rootControl: oView,
                                validateAppVersion: false,
                                flexSettings: {
                                    layer: '<%- flex.layer %>',
                                    developerMode: <%- flex.developerMode %>
								}
                            };
                            await startAdaptation(option);
                        });
                    })
                });
            });
    break;
    case 'ADAPTATION':
        sap.ui.getCore().attachInit(function () {
            sap.ushell.Container.attachRendererCreatedEvent(function () {
                sap.ushell.Container.getServiceAsync('AppLifeCycle').then((serviceInstance) => {
                    serviceInstance.attachAppLoaded((event) => {
                        var oView = event.getParameter('componentInstance');
                        sap.ui.require(["sap/ui/rta/api/startAdaptation", "<%- flex.pluginModule %>"], async function (startAdaptation, loadPlugin) {
                            var options = {
                                rootControl: oView,
                                validateAppVersion: false,
                                flexSettings: {
                                    layer: '<%- flex.layer %>',
                                    developerMode: <%- flex.developerMode %>
								}
                            };
                            await startAdaptation(options, loadPlugin);
                        });
                    })
                });
            });
        });
    break;
    </script>
```

tobiasqueck: Yes, that makes it easier to maintain. I will do it with the next commit

zdravko-georgiev: Thanks!

tobiasqueck: Now that I tried it, I am unsure whether it is a good idea or not. I didn't use a switch/case but an  if/if-else/else but I don't think we are gaining anything by having it in three blocks. On the contrary, the amount of duplication compared to the one simplified line seems to me not a good trade-off.
@zdravko-georgiev I am still open to be convinced otherwise though ...

```js
    <script><% if (!locals.flex) { /* simple preview */ %>
        sap.ui.getCore().attachInit(function (){
            <% if (!locals.locateReuseLibsScript) { %>
			sap.ushell.Container.createRenderer().placeAt("content");<% } %>
        });<% } else if (flex.pluginModule) { /* adaptation project */%>
        sap.ui.getCore().attachInit(function () {
            sap.ushell.Container.attachRendererCreatedEvent(function () {
                sap.ushell.Container.getServiceAsync('AppLifeCycle').then((serviceInstance) => {
                    serviceInstance.attachAppLoaded((event) => {
                        var oView = event.getParameter('componentInstance');
                        sap.ui.require(["sap/ui/rta/api/startAdaptation", "<%- flex.pluginModule %>"], async function (startAdaptation, loadPlugin) {
                            var options = {
                                rootControl: oView,
                                validateAppVersion: false,
                                flexSettings: {
                                    layer: '<%- flex.layer %>',
                                    developerMode: <%- flex.developerMode %>
								}
                            };
                            await startAdaptation(options, loadPlugin);
                        });
                    })
                });
            });
            <% if (!locals.locateReuseLibsScript) { %>
			sap.ushell.Container.createRenderer().placeAt("content");<% } %>
        });<% } else { /* variant management */ %>
        sap.ui.getCore().attachInit(function () {
            sap.ushell.Container.attachRendererCreatedEvent(function () {
                sap.ushell.Container.getServiceAsync('AppLifeCycle').then((serviceInstance) => {
                    serviceInstance.attachAppLoaded((event) => {
                        var oView = event.getParameter('componentInstance');
                        sap.ui.require(["sap/ui/rta/api/startAdaptation"], async function (startAdaptation) {
                            var options = {
                                rootControl: oView,
                                validateAppVersion: false,
                                flexSettings: {
                                    layer: '<%- flex.layer %>',
                                    developerMode: <%- flex.developerMode %>
								}
                            };
                            await startAdaptation(options);
                        });
                    })
                });
            });
            <% if (!locals.locateReuseLibsScript) { %>
			sap.ushell.Container.createRenderer().placeAt("content");<% } %>
        });
    <% } %>
    </script>
```

zdravko-georgiev: OK, let us keep it the way it is now.

packages/adp-tooling/tsconfig.eslint.json
```diff
@@ -1,4 +1,5 @@
 {
     "extends": "./tsconfig.json",
-    "include": ["src", "test", ".eslintrc.js"]
+    "include": ["src", "test", ".eslintrc.js"],
+    "exclude": []
```
zdravko-georgiev, line 4: why the empty `exclude`?

packages/adp-tooling/test/unit/preview/adp-preview.test.ts
```diff
@@ -161,4 +162,37 @@ describe('AdaptationProject', () => {
...
+                        url: backend
+                    }
+                },
+                mockProject as unknown as ReaderCollection,
+                logger
+            );
+
+            const app = express();
+            adp.addApis(app);
+            server = await supertest(app);
```
zdravko-georgiev, line 180: Please remove `await`, `supertest` is not async

packages/preview-middleware/src/ui5/middleware.ts
```diff
@@ -20,9 +21,20 @@ async function initAdp(rootProject: ReaderCollection, config: AdpPreviewConfig,
...
         const layer = await adp.init(JSON.parse(await appVariant.getString()));
-        flp.config.rta = { layer };
+        flp.config.rta = {
+            layer,
+            pluginModule: `${adp.extensionScript.namespace.replace('.', '/')}/${adp.extensionScript.module}`
+        };
+
+        const ns = adp.extensionScript.namespace.split('.');
+        const path = flp.config.path.split('/');
+        path.pop();
```
zdravko-georgiev, line 31:
```suggestion
        const path = flp.config.path.split('/').pop();
```

zdravko-georgiev:
Also I am getting the following error for `adp-preview.test.ts`
```
Cannot find module '@ui5/fs' or its corresponding type declarations.
```

devinea (commented):
What is the motivation to combine server, client and writer features in 1 module?
It does not seem an optimal separation. 
Also it deviates from standard module config for build etc..

voicis:
Is babel actually sufficient in this case? It only transpiles and usually all code that gets consumed by browser needs to be bundled for performance reasons.

tobiasqueck:
> Is babel actually sufficient in this case? It only transpiles and usually all code that gets consumed by browser needs to be bundled for performance reasons.

I'd say it is sufficient because it is a development module and while there might be a measurable performance issue I doubt that it is perceivable for a human being.

packages/adp-tooling/README.md
```diff
@@ -1,12 +1,23 @@
...
 The submodule writer contains functionality to generate the core project structure of an SAP UI5 adaptation project. It is not a standalone generator but designed to be integrated into `@sap-ux/create` or any kind of yeoman generator.
 
-## base
-The submodule contains functionality required in different scenarios, e.g. prompting for generation or when initializing the preview.
\ No newline at end of file
+### base
+The submodule contains functionality required in different scenarios, e.g. prompting for generation or when initializing the preview.
+
+## Development
+The standard `src` folder except the subfolder `src/preview/client` is transpiled into the dist folder. Additionally, the sources from `src/preview/client` are transpiled using babel into UI5 compatible client sources that can be used in previews to extend the runtime adaption with additional plugins.
```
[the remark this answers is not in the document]
tobiasqueck: I'd like to wait for the outcome of #1231. If the decision is to keep a flat structure, I'd really prefer to keep it together because it just confusing to have the client module standalone as a top level package. If we go for nested, then I am more than happy to split it apart.

tobiasqueck:
> What is the motivation to combine server, client and writer features in 1 module?

Combining server and writer is actually independent of this PR and happened a while ago. The main reason is the reuse of the functionality in both cases, e.g. adding artifacts like a fragment are planned to be enabled through the middleware and the client running the browser as well as through other interfaces e.g. in guided development or the page modeler. 
It is also a trend you can see in the `ui5-community` modules where middlewares and tasks are going into one `*-tooling` module since their shared functionality is significant. 

Combining client and server is done because the client is only hosted by this server, i.e. running the very same client by any other server, not offering the APIs is not useful.

packages/preview-middleware/src/ui5/middleware.ts
tobiasqueck: No :) Let's assume `flp.config.path = 'a/b/c'`
```ts
const path = flp.config.path.split('/');
path.pop();
```
results in `path = 'a/b'`
while
```ts
const path = flp.config.path.split('/').pop();
```
results in `path = 'c'`

packages/adp-tooling/test/unit/preview/adp-preview.test.ts
tobiasqueck: c92882d

packages/adp-tooling/test/unit/preview/adp-preview.test.ts
```diff
@@ -161,4 +162,37 @@ describe('AdaptationProject', () => {
             expect(next).toBeCalled();
         });
     });
+    describe('addApis', () => {
+        let server!: SuperTest<Test>;
```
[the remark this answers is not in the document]
tobiasqueck: Removed with c92882d.
If the same constellation was in `src` we would need it because I don't want to make null checks when using it in tests because `tsc` cannot know that `beforeAll` will always be executed before the test cases, however, since we don't call `tsc` on the test sources it seems to be no problem.
(to be honest, I am surprised that it isn't a problem in the tests - I'd still have expected a ts error when running the tests)

packages/adp-tooling/tsconfig.eslint.json
tobiasqueck: Because we are excluding something in `./tsconfig` that I don't want to be excluded here, so I am setting `exclude` to empty again.

tobiasqueck:
> Also I am getting the following error for `adp-preview.test.ts`
> 
> ```
> Cannot find module '@ui5/fs' or its corresponding type declarations.
> ```

that is the annoying ts server. It sometimes (sometimes not) crashes and then it cannot find the `.d.ts` files. I cannot explain to you why. I looked at the logs and at reported issues for it, and couldn't find anything helping. Restarting the language server helps often, restarting it a few times helps usually.

packages/adp-tooling/templates/project/gitignore.tmpl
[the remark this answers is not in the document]
tobiasqueck: I would say no as explained at https://github.com/SAP/open-ux-tools/pull/1166#issuecomment-1700409966

packages/adp-tooling/test/unit/preview/adp-preview.test.ts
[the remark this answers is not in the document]
zdravko-georgiev: Thanks

packages/preview-middleware/src/ui5/middleware.ts
zdravko-georgiev: Got it 😅

packages/adp-tooling/test/unit/preview/adp-preview.test.ts
zdravko-georgiev, line 166: You forgot this `!` 😉, please remove

packages/adp-tooling/test/unit/preview/adp-preview.test.ts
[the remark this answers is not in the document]
tobiasqueck: c56ee27

packages/adp-tooling/test/unit/preview/adp-preview.test.ts
tobiasqueck: c56ee27

packages/adp-tooling/README.md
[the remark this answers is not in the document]
tobiasqueck: @devinea would you be ok with https://github.com/SAP/open-ux-tools/compare/feat/1149/enable-enhancements...feat/1149/enable-enhancements-module?expand=1? I am moving the client source to a new module called `rta-client-extension`.
The ongoing work on the `control-properties-editor` will then also enhance that module and use it.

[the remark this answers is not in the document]
devinea: @tobiasqueck yes that is better thanks 👍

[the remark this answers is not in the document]
tobiasqueck: This is obsolete with the latest changes in `main`.

PR #1110, opened 2023-07-18
feat(ui5-writer): use @sapui5/types with UI5 1.113 and newer

MariusFreitag:
As described in the [@sapui5/ts-types-esm README file](https://www.npmjs.com/package/@sapui5/ts-types-esm), `@sapui5/types` should be preferred for newer SAPUI5 versions.

This closes #999.

PR #1169, opened 2023-08-11
chore(ui5-config) publish module

devinea:
publish to npm was not successful after #1110


devinea:
merging as there are no code changes here.

PR #1183, opened 2023-08-14
feat(adaptation-project): Fragment and Controller extension creation enhancements

nikmace:
_Enhancing_ the **preview-middleware** with new functionality as adding an **XML Fragment** (*creating "**addXML**" change) and adding a **Controller Extension** (*creating a "**codeExt**" change).


packages/adp-tooling/boilerplates/controller.ejs
```diff
@@ -0,0 +1,86 @@
+/*!
+ * OpenUI5
```
tobiasqueck, line 2: Please use the `templates` folder for templates. Right now it is only used for project generation, therefore, please restructure it to have `templates/project` and `templates/rta` (or some other better name)

packages/adp-tooling/package.json
```diff
@@ -39,22 +43,30 @@
         "@sap-ux/logger": "workspace:*",
         "@sap-ux/system-access": "workspace:*",
         "@sap-ux/ui5-config": "workspace:*",
+        "express-rate-limit": "6.9.0",
```
tobiasqueck, line 46: Any new module will require an admin review, please add a quick explanation to the PR for what you need these new modules.

tobiasqueck (commented):
I know it is WIP but I wanted to give some early feedback like no `any` and no `@ts-ignore` if you don't have a very VERY valid reason. :)

packages/adp-tooling/src/types.ts
```diff
@@ -39,3 +39,45 @@ export interface AdpWriterConfig {
...
+}
+
+export interface Content {
+    changeType: string;
+    content: object;
+    texts?: object;
+}
+
+export const enum FolderNames {
+    Webapp = 'webapp',
```
tobiasqueck, line 61: `Webapp` is never used in your code (and also shouldn't be used), therefore, please remove it.

packages/preview-middleware-client/src/adp/control-utils.ts
```diff
@@ -0,0 +1,484 @@
+/* eslint-disable consistent-return */
```
tobiasqueck, line 1: Please remove this and any other eslint-disable statements and fix the liniting issues properly

packages/preview-middleware-client/src/adp/control-utils.ts
```diff
@@ -0,0 +1,484 @@
...
+        name: string
+    ) {
+        let result = [];
+        const aggregation = (control ? control.getMetadata().getAllAggregations() : {})[name] as unknown as object & {
+            _sGetter: string;
+        };
+
+        if (aggregation) {
+            if (!aggregation._sGetter && !control.__calledJSONKeys) {
+                (control.getMetadata() as any).getJSONKeys();
```
tobiasqueck, line 96: no `any`. Additionally, this whole function needs to be revisited. Accessing private variables and attaching new ones does not look like maintainable software.

packages/preview-middleware-client/src/adp/control-utils.ts
```diff
@@ -0,0 +1,484 @@
...
+                    // var obj = {
+                    //    key1: value,
+                    //    key2: obj
+                    // }
+                    // and JSON.stringify can't handle that so we reach here.
+                    // however, postMessage can't handle that either, and throws:
+                    // "Failed to execute 'postMessage' on 'Window': An object could not be cloned".
+                    // so we need to check whether this is the failure and if so, don't return the rawValue,
+                    // but some default string to act as the property value.
+                    // (BCP: 1780025011)
```
tobiasqueck, line 246: please remove the BCP reference

packages/preview-middleware-client/src/adp/control-utils.ts
```diff
@@ -0,0 +1,484 @@
...
+     *
+     * @param control Control Managed Object
+     * @param controlOverlay Control overlay
+     * @param _includeDocumentation Toggle whether to include documentation
+     * @returns {BuiltRuntimeControl} Built runtime control
+     */
+    public static async buildControlData(
+        control: ManagedObject,
+        controlOverlay?: ElementOverlay,
+        _includeDocumentation = true
```
tobiasqueck, line 294: Why do you define this parameter but then don't use it?

tobiasqueck (changes requested):
First round of reviews

tobiasqueck (changes requested):
Thanks for addressing the most of the review comments.
Please have a look at the tests in `preview-middleware-client` next and try to put reusable mock functionality into `__mocks__` and only the specific setting of values into your tests.

I have taken two test files as example in https://github.com/SAP/open-ux-tools/pull/1183/commits/75aef4291de45a6178969e0fa659213723b7eee3. Please do the same for the rest.

packages/preview-middleware-client/src/adp/dialogs/fragment.ts
```diff
@@ -0,0 +1,47 @@
...
+    });
+};
+
+/**
+ * Handler for new context menu entry
+ *
+ * @param overlays Control overlays
+ * @param rta Runtime Authoring
+ */
+export async function handler(overlays: UI5Element[], rta: RuntimeAuthoring): Promise<void> {
```
tobiasqueck, line 33: I think I need help understanding this one.
You are creating an empty view referencing an instance of your controller. In your controller init you then add a fragment with your dialog.
Why don't you just create a view already containing the content of your fragment?

nikmace: Changed in https://github.com/SAP/open-ux-tools/pull/1183/commits/e0a55f9df1453f0624d0b2c1283d97d6c880374d
Now the view is not empty from the start and we do not call this.loadFragment in the onInit method.

tobiasqueck: The question stays, why not just having a view with all the content and completely getting rid of the fragment.

Your view at `src/adp/ui/AddFragment.view.xml` would be 
```xml
<mvc:View xmlns:mvc="sap.ui.core.mvc"
    xmlns="sap.m"
    xmlns:core="sap.ui.core"
    xmlns:f="sap.ui.layout.form">
    <Dialog id="addNewFragmentDialog" title="Add XML Fragment" contentWidth="500px" class="sapUiRTABorder">
        <!-- ... -->
    </Dialog>
</mvc:View>
```

And then you can simplify your handler function to 
```ts
export async function handler(overlays: UI5Element[], rta: RuntimeAuthoring): Promise<void> {
    const controller = new AddFragment('open.ux.preview.client.adp.controllers.AddFragment');
    controller.overlays = overlays;
    controller.rta = rta;

    await XMLView.create({
        viewName: 'open.ux.preview.client.adp.ui.AddFragment',
        controller
    });
}
```

PS: you might also want to add the variable assignments of `overlays` and `rta` to the constructor.

packages/adp-tooling/templates/rta/controller.ejs
```diff
@@ -0,0 +1,86 @@
+/*!
```
tobiasqueck, line 1: That one doesn't belong into this PR, correct? At least I didn't see any code generating a controller

packages/preview-middleware-client/src/adp/controllers/AddFragment.controller.ts
```diff
@@ -0,0 +1,404 @@
...
+
+/** sap.ui.dt */
+import OverlayRegistry from 'sap/ui/dt/OverlayRegistry';
+import type ElementOverlay from 'sap/ui/dt/ElementOverlay';
+
+import ControlUtils from '../control-utils';
+import CommandExecutor from '../command-executor';
+import { ManifestAppdescr, getFragments, getManifestAppdescr, writeFragment } from '../api-handler';
+
+type ExtendedEventProvider = EventProvider & {
```
tobiasqueck, line 30: What is an `ExtendedEventProvider`? It feels like a type definition that should not exist here. Is this an event provider for a specific type of events? Should this move into the `types` section with a more meaningful name?

packages/preview-middleware-client/src/adp/controllers/AddFragment.controller.ts
```diff
@@ -0,0 +1,404 @@
...
+        const fragmentName: string = source.getValue().trim();
+        const fragmentList: { fragmentName: string }[] = this.model.getProperty(
+            '/filteredFragmentList/unFilteredFragmentList'
+        );
+
+        const iExistingFileIndex = fragmentList.findIndex((f: { fragmentName: string }) => {
+            return f.fragmentName === `${fragmentName}.fragment.xml`;
+        });
+
+        switch (true) {
```
tobiasqueck, line 154: I have to admit that I have never seen `switch/case` being used like that, however, after a quick search I can see why it is nicer than `if/elseif/elseif/elseif`.
Now looking at your conditions, they shouldn't be done in this order and as a flat sequence. You should actually have nested logic like

```ts
if (fragmentName.length <= 0) {
 // case 2
} else {
    // check if the fragmentName is in the list
    if (your_name_check) {
        // case 3 
    } else if (in_the_list) {
       // case 1
    } else {
       // case 4
    }
}
```

Why should you do this: you only loop through the list if the name is valid, and you don't check length only once

packages/preview-middleware-client/src/adp/dialogs/fragment.ts
nikmace: Done in: https://github.com/SAP/open-ux-tools/pull/1183/commits/60916e4ac16947c1e325a6c4cf4474d79a5b7aec

packages/adp-tooling/templates/rta/controller.ejs
nikmace: Removed in: https://github.com/SAP/open-ux-tools/pull/1183/commits/4a2829c286592f1d845bdb93f987f25507c02025

packages/preview-middleware-client/test/unit/adp/controllers/AddFragment.controller.test.ts
```diff
@@ -0,0 +1,324 @@
...
+            addFragment.model = {
+                getProperty: jest
+                    .fn()
+                    .mockReturnValueOnce('Share')
+                    .mockReturnValueOnce('0')
+                    .mockReturnValueOnce('content')
+            } as unknown as JSONModel;
+
+            const createFragmentSpy = jest.fn();
+            // @ts-ignore
```
tobiasqueck, line 256: You shouldn't use `@ts-ignore` here, instead think about how to test without ts complaining.

The main issue here is that your tests assume the existence of a private function which is internal knowledge of the unit you are testing, so your test should not rely on that. Instead think about what the actual result of this test case should be.
If I read the code correctly then in a positive case it should be a call to the public backend API which for the client-side code is the call to `fetch`. So, instead of mocking internal knowledge, mock the call to fetch that is required to create a new fragment. At the end of the test you validate that the correct data is being sent to the backend (with fetch), don't just check if fetch is called.

Once you did this, you can also delete you test of private functions below/

packages/preview-middleware-client/src/adp/controllers/AddFragment.controller.ts
nikmace: Changed in: https://github.com/SAP/open-ux-tools/pull/1183/commits/f13674dc7d302b072082058c3dc36078dcc9adc3

packages/preview-middleware-client/src/adp/controllers/AddFragment.controller.ts
nikmace: Removed in: https://github.com/SAP/open-ux-tools/pull/1183/commits/07333f45e7c924ab2f3d32d4d86c57006fb1ffce

packages/preview-middleware-client/test/unit/adp/controllers/AddFragment.controller.test.ts
tobiasqueck: `@ts-ignore` is gone, thank you, however, I am not sure I like that you made private functions public to get rid of it. The test for `onCreateBtnPress` should cover everything you split into the tests for `createNewFragment` and `createFragmentChange`

nikmace: Made the methods private and reworked tests in: https://github.com/SAP/open-ux-tools/pull/1183/commits/1e3e4f2d1f48d5a510e8ad6dfe37f105cb2b0cd9

tobiasqueck (dismissed):
* changeset ok
* code is understandable
* test coverage is very good
   * there are few minor things that could be improved in the tests but testing the client-side code is something very new in this repo, so I expect that we will revisit the tests anyhow in the near future
* locally tested and works like a charm

tobiasqueck (approved):
Additional changes since last review are aligned.

testojs (approved):
Thanks, @nikmace, looks awesome! 🔥 
Tested locally, the main scenario is working as expected.

PR #1168, opened 2023-08-10
feat(project-access): fallback to load @sap/cds-dk

Klaus-Keller:
## Ticket 

https://github.com/SAP/open-ux-tools/issues/1167 

## Description
This PR implements a fallback if loading of `@sap/cds` for a CAP project fails, it tries to load `@sap/cds-dk` from any node module location, also global install node modules. This is required for CAP Java projects, that might or might not have a dependency to `@sap/cds`.

PR #1185, opened 2023-08-14
(project-access): fallback to load cds using cds --version

Klaus-Keller:
Alternative solution to https://github.com/SAP/open-ux-tools/pull/1168, 
Uses `cds --version` to find location of global `@sap/cds` installation.


.changeset/pretty-radios-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/project-access': minor
+---
+
+Add fallback to load @sap/cds-dk from any module location
```
devinea, line 5:
```suggestion
Add fallback to load @sap/cds from any module location
```

packages/project-access/src/project/cap.ts
```diff
@@ -184,13 +185,36 @@ export async function getCapEnvironment(capProjectPath: string): Promise<CdsEnvi
...
+    try {
+        // First approach, load @sap/cds from project
+        module = await loadModuleFromProject<CdsFacade | { default: CdsFacade }>(capProjectPath, '@sap/cds');
+    } catch (error) {
+        loadProjectError = error;
+    }
+    if (!module) {
+        try {
+            // Second approach, load @sap/cds from @sap/cds-dk
+            module = await loadGlobalCdsModule();
```
devinea, line 208: I don't think you need the fallback here.

packages/project-access/src/project/cap.ts
```diff
@@ -184,13 +185,36 @@ export async function getCapEnvironment(capProjectPath: string): Promise<CdsEnvi
...
  * @returns - CAP CDS module for a CAP project
  */
 async function loadCdsModuleFromProject(capProjectPath: string): Promise<CdsFacade> {
-    const module = await loadModuleFromProject<CdsFacade | { default: CdsFacade }>(capProjectPath, '@sap/cds');
+    let module: CdsFacade | { default: CdsFacade } | undefined;
+    let loadProjectError;
+    let loadError;
+    try {
+        // First approach, load @sap/cds from project
+        module = await loadModuleFromProject<CdsFacade | { default: CdsFacade }>(capProjectPath, '@sap/cds');
```
devinea, line 201:
```suggestion
       const versions = await getCdsVersionInfo(capProjectPath);
        if (!versions.home) {
        const CdsPathCache = versions.home;
        module = await loadModuleFromProject<CdsFacade | { default: CdsFacade }>(CdsPathCache, '@sap/cds');
```
Can't we rely `cds --version` to resolve the cds location (assuming home is returned in older versions of the cli ??)  and use that path for `loadModuleFromProject`

Klaus-Keller: We can rely on `cds --version`, but it is rather slow to execute the command each time we access the cds module. For instance, we use `getCapProjectType()` which calls `getCapCustomPaths()` -> `getCapEnvironment()` -> `loadCdsModuleFromProject()` many times when searching all projects in workspace, we use `getCapModelAndServices()` and `readCapServiceMetadataEdmx()` in design time after almost every step, so calling a terminal command each time before would have a very negative impact on performance. TI thought about creating a cache `Map<string, string>` where we store project root -> path to cds, but we also need to handle following scenario:
1. `@sap/cds` is not installed locally in the project
2. our tools access it and cache the path to global installation
3. `@sap/cds` is installed in the project
Not sure how we can detect this and invalidate the cache. 

This is why I went for the solution on this PR:
1. Try to load local `@sap/cds` from project
2. If it exists, it will be in `require` cache and loading is fast
3. If it does not exist, try to load global installed `@sap/cds-dk` (path is also cached)

Let me know if you find a better solution.

devinea: Sounds good. Thanks

devinea (dismissed):
Code, test and changeset looks good. Thanks @Klaus-Keller 
I haven't tested locally.

packages/project-access/src/project/cap.ts
```diff
@@ -295,3 +319,52 @@ async function readPackageNameForFolder(baseUri: string, relativeUri: string): P
...
+    return loadModuleFromProject<T>(globalCdsPathCache, '@sap/cds');
+}
+
+/**
+ * Get cds information, which includes versions and also the home path of cds module.
+ *
+ * @param [cwd] - optional folder in which cds --version should be executed
+ * @returns - result of call 'cds --version'
+ */
+function getCdsVersionInfo(cwd?: string): Promise<Record<string, string>> {
```
slin-sap, line 347: Adding async signature to the function?

slin-sap (changes requested):

slin-sap (approved):

packages/project-access/src/project/cap.ts
Klaus-Keller: Will be addressed in a separate PR, thanks @slin-sap

PR #1192, opened 2023-08-15
feat(project-access): async function definition

Klaus-Keller:
Small fix for https://github.com/SAP/open-ux-tools/pull/1185#discussion_r1295028942.
Does not change the behavior, but for consistency reasons we should use `async function` if a `Promise` is returned.


kranthie-sap (approved):
* Change to the signature looks good
* Changeset has been maintained

devinea (approved):
update and changeset look good

PR #1193, opened 2023-08-16
WIP: adding fragments to adaptation projects

tobiasqueck:
This PR is not to be merged but is to better review #1183.
#1183 consists of two sets of changes:
1. #1166
2. #1193


packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -0,0 +1,198 @@
...
+    };
+
+    /**
+     * Handler for reading the manifest.appdescr_variant contents
+     *
+     * @param req Request
+     * @param res Response
+     * @param next Next Function
+     */
+    public handleReadAppDescrVariant = async (req: Request, res: Response, next: NextFunction) => {
```
tobiasqueck, line 177: You don't need this additional handler. All files in the `webapp` folder are always hosted at `/` e.g. the app descriptor variant is at `/manifest.appdescr_variant`

packages/adp-tooling/src/preview/adp-preview.ts
```diff
@@ -91,7 +116,15 @@ export class AdpPreview {
         private readonly config: AdpPreviewConfig,
         private readonly project: ReaderCollection,
         private readonly logger: ToolsLogger
-    ) {}
+    ) {
+        this.cache = new NodeCache();
+        this.routesHandler = new RoutesHandler(project, logger, this.cache);
+        this.rateLimiter = rateLimit({
```
tobiasqueck, line 122: While this is very useful for a productive server, this middleware runs only in a dev environment, so you can remove the rate limiter code, and I will mark the corresponding sonar error as false positive.

packages/adp-tooling/package.json
```diff
@@ -14,9 +14,10 @@
     "author": "@SAP/ux-tools-team",
     "main": "dist/index.js",
     "scripts": {
-        "build": "pnpm build:server && pnpm build:client",
+        "build": "pnpm build:server && pnpm build:client && pnpm copy:templates",
         "build:server": "tsc --build",
         "build:client": "tsc --noEmit --project src/preview/client/tsconfig.json && babel --extensions .ts ./src/preview/client -d ./dist/preview/client",
+        "copy:templates": "rimraf ./dist/templates && shx cp -r ./templates/rta ./dist/templates",
```
tobiasqueck, line 20: We don't copy the templates into `dist`. The `templates` folder is part of the delivered package, so you can reference it there. You can check the coding in `writer/index.ts`

packages/adp-tooling/src/preview/adp-preview.ts
nikmace: Removed in https://github.com/SAP/open-ux-tools/pull/1183/commits/14175ce266a6695f01a40102dbf1a99c50c0b4e4
and in https://github.com/SAP/open-ux-tools/pull/1183/commits/b8b42a09ac4bf1a7e85e93d8903657b47187eb24

packages/adp-tooling/package.json
nikmace: Done in https://github.com/SAP/open-ux-tools/pull/1183/commits/48a9f41847b6978b602d2596b21102659e136323

packages/adp-tooling/src/preview/client/api-handler.ts
```diff
@@ -0,0 +1,114 @@
...
+export interface FragmentsResponse {
+    fragments: Fragments;
+    filteredFragments: Fragments;
+    message: string;
+}
+
+/**
+ * @description Class responsible for sending requests from client to the server
+ */
+export default class ApiRequestHandler {
```
tobiasqueck, line 26: You don't need the overhead of a (UI5) class for a bunch of static functions, you can just write them as separate functions in a module:

```typescript
export async function request<T>(endpoint: ApiEndpoints, method: RequestMethod, data?: unknown): Promise<T> {
 ...
}

export async function getFragments<T>(): Promise<T> {
    return request<T>(ApiEndpoints.FRAGMENT, RequestMethod.GET);
}

...
```
You existing imports of the functionality should work as they are.

packages/adp-tooling/src/preview/routes-handler.ts
tobiasqueck: @nikmace I think there is a misunderstanding. Not even what you have in 8abb0ba is required. You need no code at all. The content of the `webapp` folder is exposed by `ui5 serve` by default.

packages/adp-tooling/src/preview/client/api-handler.ts
nikmace: Done in https://github.com/SAP/open-ux-tools/pull/1183/commits/8abb0baa0fe0466e9049545b5d940d72c6955ded

packages/adp-tooling/src/types.ts
```diff
@@ -39,3 +39,52 @@ export interface AdpWriterConfig {
...
+    content: Content[];
+}
+
+export interface Content {
+    changeType: string;
+    content: object;
+    texts?: object;
+}
+
+export const enum FolderNames {
```
tobiasqueck, line 60: Are these folder really static? I know that by default an adaptation project uses these folders but technically, I can put in my change anything else and it will also work, correct? So, please think about how to not use them.

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -0,0 +1,200 @@
...
+    /**
+     * Generic caching function, caches the results of a callback
+     *
+     * @param key Cache key - unique identifier
+     * @param cb Callback used to return data
+     * @param ttlSeconds Time-to-live for cached item
+     * @template T
+     * @returns {T} Type or Interface of the return data
+     */
+    private withCache<T>(key: string, cb: () => T, ttlSeconds: number = 60): T {
```
tobiasqueck, line 41: I don't think we should use caching here because
1. Do we really need that little performance benefit in a dev middleware?
2. How do you know that nobody manually added/removed a file since you filled the cache?

tobiasqueck (commented):
Another set of review comments

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -0,0 +1,200 @@
...
+                    .map(async (file) => {
+                        const buffer = await file.getBuffer();
+                        const data: object & { content: { fragmentPath: string } } = JSON.parse(buffer.toString());
+                        // check if the fragment file exists in the workspace, return nothing if it does not
+                        const fragmentFileName = data.content.fragmentPath;
+                        const fragmentName = fragmentFileName.split('/').pop() as string;
+
+                        delete dict[fragmentName];
+                    })
+                    .map((f) => Promise.resolve(f))
```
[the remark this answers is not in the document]
nikmace: Because otherwise the `dict` object will be full and no keys will be deleted. That's why you need to wait for the promise to resolve. Also, I will be deleting this functionality since the new dialog doesn't require this.

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /**
+     * Handler for retrieving XML Fragments for templates used for Dialogs
+     *
+     * @param req Request
+     * @param res Response
+     * @param next Next Function
+     */
+    public handleGetXMLFragmentByName = async (req: Request, res: Response, next: NextFunction) => {
```
tobiasqueck, line 60: Can you explain for what we need this function? Why do you need expose the any of the fragments from the templates folder?

nikmace: Because as you can see I've removed the dialog from `dialogs/fragment.ts` and put it inside an XML file and it is now following the best practices of UI5. Now we have a UI folder with all the fragments, there will be dialog also for a controller extension and maybe other stuff later. We instantiate it in the fragment.ts file. And then it is looking for the fragment at /resources/adp/fragments/<name>.

packages/adp-tooling/src/preview/routes-handler.ts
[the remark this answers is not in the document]
tobiasqueck: Ok, if you delete this one, I won't argue further and mark this conversation as resolved the moment it is gone.

packages/adp-tooling/src/preview/routes-handler.ts
tobiasqueck: I am sorry, but no, "I cannot see". As far as I can see, you are working on a rather simple create and list all functionality. You client code needs a list of fragments in the workspace (simple `byGlob` for `.fragment.xml` files), and then your client either allows assigned one of these files to a change (no backend request needed) or to create a new fragment. For creating the new fragement, you client never needs to see the template, you just create a new file by copying the template when your post fragment request is triggered.

So, again, why do you need a function that exposes files from the template folder?

packages/preview-middleware/package.json
```diff
@@ -17,7 +17,7 @@
         "start:fixture": "ui5 serve --config test/fixtures/simple-app/ui5.yaml",
         "build": "tsc --build",
         "watch": "tsc --watch",
-        "clean": "rimraf dist test/test-output coverage *.tsbuildinfo",
```
tobiasqueck, line 20: Was that change required because something did not work?

packages/adp-tooling/src/preview/client/dialogs/fragment.ts
```diff
@@ -0,0 +1,436 @@
...
+     */
+    public init(contextMenu: ContextMenu) {
+        // We need this in order to keep the reference to the class
+        // we cannot use this keyword, because it is overshadowed by function scope
+        const that = this;
+
+        /**
+         * Controller for the action in the Dialog
+         */
+        const dummyController = {
```
tobiasqueck, line 108: The name should already indicate that this is not the best way of defining the controller logic for you dialog. Please consider moving this into a separate file e.g. `addFragmentDialog.controller.ts` file.

packages/adp-tooling/src/preview/routes-handler.ts
tobiasqueck: Wait a second. Why is the `add-fragment.fragment.xml` in the `templates` folder and not in `src/preview/client` (maybe a subfolder)? This file is static client code and shouldn't be templated.

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -0,0 +1,200 @@
...
+     * @param res Response
+     * @param next Next Function
+     */
+    public handleWriteFragment = async (req: Request, res: Response, next: NextFunction) => {
+        try {
+            const data = req.body as WriteFragmentBody;
+
+            const fragmentName = sanitize(data.fragmentName);
+
+            const projectPath = process.cwd();
```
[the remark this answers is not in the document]
nikmace: I do not need to access the files here, I need the path to where I will write the new fragment, and to know where I am at the moment C:\\something\\something\\adp.project\\webapp\\changes. `this.project` only finds files it does not give path to a folder for example

packages/adp-tooling/src/types.ts
nikmace: Yes, they are static and are predefined for adaptation projects. The structure is always the same. Maybe I didn't understand about you putting your 'change anything else' but you cannot put .change files anywhere else than the changes folder (not in any subfolder)

tobiasqueck: `changes` is the only one fixed (at least I haven't tried to change this one). `webapp` can be anything else, it is only relevant in your development projects to define the source root that will be used to build to `dist`. I can put my fragments in any folder under changes, only the reference fro the `.change` file matters. The same for i18n files, controller extensions etc.
So, don't use these hard coded folder names but you have to e.g. find your fragments using the project API (to not rely on `webapp` and then globs like `**/*.fragment.xml`.

packages/adp-tooling/src/preview/adp-preview.ts
```diff
@@ -37,6 +47,14 @@ export class AdpPreview {
      * Merged descriptor variant with reference app manifest
      */
     private mergedDescriptor: MergedAppDescriptor;
+    /**
+     * Descriptor variant from the project
+     */
+    private descriptorVariant: DescriptorVariant;
```
tobiasqueck, line 53: This value isn't used i.e. is set but never read. Please remove it

packages/adp-tooling/src/preview/client/ui/AddFragment.fragment.xml
```diff
@@ -0,0 +1,70 @@
+<core:FragmentDefinition 
+    xmlns="sap.m" 
+    xmlns:core="sap.ui.core" 
+    xmlns:l="sap.ui.layout"
+    xmlns:f="sap.ui.layout.form" >
+        <Dialog id="CreateSalesOrderDialog" title="Add XML Fragment" contentWidth="500px" class="sapUiRTABorder" >
```
tobiasqueck, line 6: I am sure that is not a good id ;)

packages/adp-tooling/src/preview/client/dialogs/fragment.ts
```diff
@@ -0,0 +1,51 @@
...
+/** sap.ui.dt */
+import type ContextMenu from 'sap/ui/dt/plugin/ContextMenu';
+
+import type AddFragment from '../controllers/AddFragment.controller';
+import type { BaseDialog } from './base';
+
+/**
+ * Handles creation of the dialog, fills it with data
+ */
+export default class FragmentDialog implements BaseDialog {
```
tobiasqueck, line 18: Seems like a bit of an overhead to have a UI5 class here since it only offers one function that is called exactly once.
Do you expect anything else here in the future that would require a class/instance keeping a state? Otherwise, I would recommend to just export the init function (with an additional parameter `rta`)

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -0,0 +1,200 @@
...
+
+            res.status(HttpStatusCodes.OK).contentType('application/xml').send(fragment);
+        } catch (e) {
+            res.status(HttpStatusCodes.INTERNAL_SERVER_ERROR).send(e.message);
+            next(e);
+        }
+    };
+
+    /**
+     * @description Handler for reading all fragment files from the workspace
```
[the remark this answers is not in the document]
tobiasqueck: Thanks for removing the filtering. Now remove `@description` and we are good to go

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -0,0 +1,114 @@
...
+                if (fs.existsSync(filePath)) {
+                    res.status(HttpStatusCodes.CONFLICT).send(`Fragment with name "${fragmentName}" already exists`);
+                    return;
+                }
+
+                // Copy the template XML Fragment to the project's workspace
+                const fragmentTemplatePath = path.join(__dirname, '../../templates/rta', TemplateFileName.Fragment);
+                fs.copyFileSync(fragmentTemplatePath, filePath);
+
+                const message = 'XML Fragment created';
```
tobiasqueck, line 102: Could you please add logging messages as you nicely did in `handleReadAllFragments`

packages/adp-tooling/src/types.ts
tobiasqueck: You are only using `Changes` and `Fragments` in your code. I understand that `Coding` and `Internationalization` will be used later, however, I would remove them for now and only add them when you start using them (in future PRs)

packages/adp-tooling/src/types.ts
```diff
@@ -39,3 +39,52 @@ export interface AdpWriterConfig {
...
+    Fragments = 'fragments',
+    Coding = 'coding',
+    Internalization = 'i18n'
+}
+
+export const enum TemplateFileName {
+    Fragment = 'fragment.xml'
+}
+
+export const enum ProjectFileNames {
```
tobiasqueck, line 72: I couldn't find an usage of this enum.

tobiasqueck (commented):
The code looks very nice already, I just had a few small comments.

packages/adp-tooling/jest.config.js
```diff
@@ -1,5 +1,5 @@
 const config = require('../../jest.base');
-config.coveragePathIgnorePatterns = ['src/preview/client'];
+// config.coveragePathIgnorePatterns = ['src/preview/client'];
```
tobiasqueck, line 2: If it is not needed anymore then just delete it.

Issue #1149, opened 2023-08-08
FEATURE - Enable control editing in adaptation


tobiasqueck:
This feature request is a follow up to #1105.

_As an **adaptation project developer**,
I want to **have full editing capabilities in both VSCode and BAS**,
so that I **can develop my apps anywhere**_

### Description
With #1105 it is possible to create an SAP UI5 adaptation project that can be previewed in VSCode. It is also possible to do simple editing of control properties, however, the more complex features like adding XML fragments or controller extensions are not yet supported and require to use SAP Business Application Studio. This feature request is to close the gap and support all editing features independent of the environment.

### Technical Design
* new module `control-properties-editor`
    * extracted from the Fiori tools
    * React based editing frame allowing to run a UI5 app as iframe
    * additional private module `control-properties-editor-common`
* enhance `adp-tooling`
    * serve additional functionality for the iframe
    * serve the `control-properties-editor` in edit-mode
* integrate everything in the `preview-middleware`
    * might require minor enhancements when initializing the `adp-tooling`
    * additional private module `preview-middleware-client` with sources that run in the browser

![Screenshot 2023-09-07 at 18 10 51](https://github.com/SAP/open-ux-tools/assets/44933199/937eeabf-3728-46da-b5fa-15123ee4fbb2)

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I create a valid adaptation project with `npm init @sap-ux generate adaptation-project` 
**when** I execute `npm run editor`
**then** I have access to all features e.g. adding XML fragments

### Notes
@testojs @zdravko-georgiev @ManfredLerner

### Tasks
- [x] Publish the `control-properties-editor` as new module (extract it from the SAP internal Fiori tools sources)
- [x] Enhance the `adp-tooling` module to also serve the required enhancements for the editing iframe
- [x] Integrate `control-properties-editor` into the `preview-middleware` (and `adp-tooling`) to load it for adaptation projects
- [x] Add sufficient unit test coverage

MN-public:
@Jimmy-Joseph19 FYI

PR #1199, opened 2023-08-17
Add SAPUI5 (client-side) types and enhancements required for #1149

tobiasqueck:
This PR adds the standard SAPUI5 types and enhancements for the dt and rta required for the work on #1149.

The change does not modify and published module, therefore, no changeset required.

Simple usage example: https://github.com/SAP/open-ux-tools/blob/4db5180d70e945c1dbabeaf728caac62b3c9ed3e/packages/adp-tooling/src/preview/client/tsconfig.json#L3


nikmace (approved):
All is good from my side for now.

voicis (approved):
- Tested types (not all) in another module locally and they worked as expected
- Types are neatly structured by namespaces and are easy to navigate

Note: for `ManagedObjectMetadata.getAllProperties` more precise types were needed in control property editor, but at the moment I don't have a good idea how to properly do it here and this can be done later.

Issue #1205, opened 2023-08-18
BUG - `FindFioriArtifacts` returns multiple libraries in case of multiple `manifest.json`


johannes-kolbe:
### Description
When using [`findFioriArtifacts`](https://github.com/SAP/open-ux-tools/blob/88e11c87ad94c05d6aeaa0c93ebef68a0b3a278a/packages/project-access/src/project/search.ts#L376) for libraries, multiple artifacts are being returned, if the library contains multiple `manifest.json`

### Steps to Reproduce

1. Download or clone example project https://[INTERNAL]/fiori-tools-qa/se.mi.plm.attachmentservice
2. Use `findFioriArtifacts` from `@sap-ux/project-access` to find `libraries`
3. See that result contains not 1 artifact, but 4 to the several `manifest.json` locations

### Expected results
- Only one library is returned, as other components are embedded by the main one.

### Actual results
- Multiple libraries are returned with each `manifest.json` location being different

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1207, opened 2023-08-21
fix(project-access): filter nested libraries

johannes-kolbe:
#1205 

- Filters nested libraries to only return main library, which might include components, but should not be displayed on their own.


devinea (approved):
Changeset, code and tests look appropriate for the fix. 
I haven't tested locally.

johannes-kolbe:
As proposed by @Klaus-Keller logic to further process the returned library/component will be done from consumer side. Closing issue

PR #1211, opened 2023-08-21
feat: proposal to get fiori artifact type

Klaus-Keller:
## Description

Proposal to return the Fiori Artifact type for a given path.


Klaus-Keller:
Found a different solution that uses the result of `findFioriArtifacts()`. Closing this PR.

PR #1212, opened 2023-08-21
feat: find all CAP projects

Klaus-Keller:
## Description

Add `findCapProjects()` to search for CAP projects in the workspace.

### DIFF
--- a/packages/project-access/src/project/search.ts
+++ b/packages/project-access/src/project/search.ts
@@ -407,3 +407,30 @@ export async function findFioriArtifacts(options: {
     }
     return results;
 }
+
+/**
+ * Find all CAP project roots by locating pom.xml or package.json in a given workspace.
+ *
+ * @param wsFolders - list of roots, either as vscode WorkspaceFolder[] or array of paths
+ * @returns - root file paths that may contain a CAP project
+ */
+export async function findCapProjects(wsFolders: WorkspaceFolder[] | string[]): Promise<string[]> {
+    const result = new Set<string>();
+    const excludeFolders = ['node_modules', 'dist', 'src', 'webapp', 'MDKModule', 'gen'];
+    const fileNames = [FileName.Pom, FileName.Package];
+    const wsRoots = wsFoldersToRootPaths(wsFolders);
+    for (const root of wsRoots) {
+        const filesToCheck = await findBy({
+            fileNames,
+            root,
+            excludeFolders
+        });
+        const foldersToCheck = Array.from(new Set(filesToCheck.map((file) => dirname(file)))); //only directories without duplicates
+        for (const folderToCheck of foldersToCheck) {
+            if ((await getCapProjectType(folderToCheck)) !== undefined) {
+                result.add(folderToCheck);
+            }
+        }
+    }
+    return Array.from(result);
+}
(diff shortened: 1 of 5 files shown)


packages/project-access/test/project/find-apps.test.ts
```diff
@@ -289,3 +289,18 @@ describe('Test findFioriArtifacts()', () => {
         ]);
     });
 });
+
+describe('Test findCapProjects()', () => {
+    test('Find CAP projects', async () => {
```
IainSAP, line 294: @Klaus-Keller Could you please add some nested projects to the test set also, for example see the `cloud-cap-samples` repo : https://github.com/SAP-samples/cloud-cap-samples

packages/project-access/src/project/search.ts
```diff
@@ -407,3 +407,33 @@ export async function findFioriArtifacts(options: {
...
+    const wsRoots = wsFoldersToRootPaths(options.wsFolders);
+    for (const root of wsRoots) {
+        const filesToCheck = await findBy({
+            fileNames,
+            root,
+            excludeFolders
+        });
+        const foldersToCheck = Array.from(new Set(filesToCheck.map((file) => dirname(file)))); //only directories without duplicates
+        for (const folderToCheck of foldersToCheck) {
+            if ((await getCapProjectType(folderToCheck)) !== undefined) {
```
IainSAP, line 433: Can we add additional validation, perhaps optionally, to the function `getCapProjectType` as the current checks are too rudimentary to be useful in some cases. We need to also validate that the services path exists and is accessible (correct file permissions), otherwise we will return non-cap projects that will not be compilable.

Klaus-Keller: @IainSAP, good call, I've ran against https://github.com/SAP-samples/cloud-cap-samples and it found the root of the repo as CAP project. Added check for `srv` folder: https://github.com/SAP/open-ux-tools/pull/1212/commits/d7045e3ee1ee582c2a89afc299888a66444e911d. Will add some more test cases and unit tests.

packages/project-access/test/project/find-apps.test.ts
Klaus-Keller: Added test with and without `srv` folder

packages/project-access/src/project/search.ts
Klaus-Keller: Tests modified/added

IainSAP: Could we add logs (and allow an external logger to be passed) so we can trace skipped projects and why they are skipped. This is needed so we can easily debug issues with CAP discovery.

IainSAP (approved):
Thanks, LGTM

- Consumed and tested locally, as expected

devinea (approved):
Changset ✅ 
New function to locate CAP Java projects looks good. 
There doesn't seem to be a test to coverage finding a CAP Java App with a pom.xml and no package.json (if that is valid?). This can be added later if needed.

slin-sap (approved):
I didn't test the changes. But code change looks good to me
Code changes has been covered by unit tests
minor cset looks good since there is API signature change.

kranthie-sap (approved):
* Changes to the CAP utility functions look good
* Good coverage with unit tests
* Changeset maintained

PR #1214, opened 2023-08-22
Expose generated preview html for usage in CAP projects with cds-plugin-ui5

tobiasqueck:
This enhancement makes use of the new hook of the `cds-plugin-ui5` https://github.com/ui5-community/ui5-ecosystem-showcase/issues/785.

I tested locally with https://github.com/tobiasqueck/ui5con2023/tree/preview-middleware by simply copying the `dist` folder of the `preview-middleware` into the `node_modules/@sap-ux/preview-middleware/dist` folder of my test project and everything worked as expected.


cfg74:
@tobiasqueck @zdravko-georgiev as discussed, we need to make sure the generated mount points have precedence over existing files if specified by the middleware.

tobiasqueck:
> @tobiasqueck @zdravko-georgiev as discussed, we need to make sure the generated mount points have precedence over existing files if specified by the middleware.

While I agree, that is unrelated to this fix ;)

packages/preview-middleware/src/types/index.ts
```diff
@@ -27,7 +27,7 @@ export interface App {
  */
 export interface FlpConfig {
     path: string;
-    intent?: Intent;
+    intent: Intent;
```
zdravko-georgiev, line 30: Why do we need to make the `Intent` mandatory? Just displaying the FLP page with the Apps is also a valid, right?

zdravko-georgiev, line 30: If this changes, then we need to update the README

tobiasqueck: It is not mandatory in the config: https://github.com/SAP/open-ux-tools/blob/poc/cds-plugin-ui5/packages/preview-middleware/src/types/index.ts#L44 but in the constructor I fill all undefined values with defaults because it is required later.

tobiasqueck: See above comment

zdravko-georgiev: OK, got it, thanks!

zdravko-georgiev: 👍

zdravko-georgiev (approved):
Code looks good
Excellent test coverage
Didn't tested manually

longieirl (approved):
Havent tested locally but code seems good.

PR #1220, opened 2023-08-23
fix: remove uri encoding on package name

slin-sap:
Root cause analysis detailed in #1103 

- Confirmed by ADT team no uri encoding required for package name.
- Checked with [WebIDE implementation](https://github.wdf.sap.corp/devx-archive/EclipseCheWebIDE/blob/9e27e01c221ee65515259cce2a0ff94ba038d685/com.sap.watt.ide.core/src/main/webapp/resources/sap/watt/uitools/plugin/abaprepository/service/Transport.js#L115), only application name requires uri encoding.
- Validated the scenario during customer meeting. List tr only works when uri encoding is not applied to pacakg name.

Changes in this PR:
- No uri encoding should be added to package name when constructing the request body
- Updated unit test


longieirl (approved):
Changes look good and testing is covering change.

devinea (approved):
Updates are expected. 
Changeset ✅ 
tests ✅

kranthie-sap (approved):
The changes look good. A nice to have feature would be add a client side validation to the package name and app name based on the naming conventions and allowed characters.
Changeset has been maintaine.

PR #1222, opened 2023-08-24
fix(project-access): usage of cds on windows

Klaus-Keller:
## Issue

The service details return from module `@sap/cds` looks different on Windows than on MacOS or Linux:

**Windows**
image

**MacOS, Linux**
image

While we did already a conversion from `\` -> `/` and removed double slashes, the leading slash was still in result of function call `getCapModelAndServices()`. The search for service removes leading and trailing slashes, so the mostly used function `readCapServiceMetadataEdmx()` was working as expected. This pull requests fixes the returned service of call `getCapModelAndServices()` and adds unit tests for the case with backslashes.

By the way the team of `@sap/cds` is informed and will provide a fix soon, goal is to return the service URL with forward slashes on any operating system.

### DIFF
--- a/packages/project-access/test/project/cap.test.ts
+++ b/packages/project-access/test/project/cap.test.ts
@@ -89,14 +89,15 @@ describe('Test getCapModelAndServices()', () => {
             load: jest.fn().mockImplementation(() => Promise.resolve('MODEL')),
             compile: {
                 to: {
-                    serviceinfo: jest.fn().mockImplementation(() => [{
-                        "name": "Forwardslash",
-                        "urlPath": "odata/service/with/forwardslash/",
-                    },
-                    {
-                        "name": "Backslash",
-                        "urlPath": "\\odata\\service\\with\\backslash/",
-                    },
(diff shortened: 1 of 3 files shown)


devinea (approved):
Changset ✅ 
Clean code and tests ✅ 

Thanks @Klaus-Keller

zdravko-georgiev (approved):
Code looks good
Excellent test coverage

packages/project-access/test/project/cap.test.ts
```diff
@@ -87,15 +87,40 @@ describe('Test getCapModelAndServices()', () => {
...
+        expect(capMS).toEqual({
+            model: 'MODEL',
+            services: [
+                {
+                    'name': 'Forwardslash',
+                    'urlPath': 'odata/service/with/forwardslash/'
+                },
+                {
+                    'name': 'Backslash',
+                    'urlPath': 'odata/service/with/backslash/'
```
schreckstefan, line 120: backslash???

@Klaus-Keller

PR #1224, opened 2023-08-27
fix(ui5-proxy-middleware): support usage inside connect servers

petermuessig:
When using the dev-approuter (https://www.npmjs.com/package/dev-approuter) to run UI5 and CAP applications inside the @sap/router from a monorepo then the server is using connect and not express. In those scenarios, similar like with livereload only the basic http response object from Node.js is available and the status function is not available which leads to an INTERNAL SERVER ERROR. By using the writeHead, write and end function of the response object the function is agnostic to the used server and works in both scenarios.

### DIFF
--- a/packages/ui5-proxy-middleware/src/base/utils.ts
+++ b/packages/ui5-proxy-middleware/src/base/utils.ts
@@ -185,7 +185,11 @@ export const setHtmlResponse = (res: any, html: string): void => {
         res.write(html);
         res.end();
     } else {
-        res.status(200).contentType('html').send(html);
+        res.writeHead(200, {
+            "Content-Type": "text/html"
+        });
+        res.write(html);
+        res.end();
     }
 };
 
--- a/packages/ui5-proxy-middleware/src/base/utils.ts
+++ b/packages/ui5-proxy-middleware/src/base/utils.ts
@@ -186,7 +186,7 @@ export const setHtmlResponse = (res: any, html: string): void => {
         res.end();
     } else {
         res.writeHead(200, {
-            "Content-Type": "text/html"
+            'Content-Type': 'text/html'
         });
         res.write(html);
         res.end();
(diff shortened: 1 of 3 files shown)


petermuessig:
Hi @zdravko-georgiev ,

please get in touch with @nicoschoenteich - he has the project setup for this new `dev-approuter` with which you can also validate the fix. I tested it locally by manually fixing it in the node_modules. 

BTW: when you search for `res.status` in this repo, you can find additional places which may have issues in such `connect` scenarios: `adp-tooling`, `preview-middleware`. They are not required to be fixed for now - but just FYI. It would be great to get this one fixed, so that the `proxy-middleware` can be used with the `dev-approuter`.

packages/ui5-proxy-middleware/src/base/utils.ts
```diff
@@ -185,7 +185,11 @@ export const setHtmlResponse = (res: any, html: string): void => {
         res.write(html);
         res.end();
     } else {
-        res.status(200).contentType('html').send(html);
+        res.writeHead(200, {
```
tobiasqueck, line 188: With this change, do we still need the if or could we also execute `res.writeHead(200, { ...` in a liveload scenario?

tobiasqueck (commented):
- code changes look good on first glance, just added a question
- changeset is missing
- didn't test locally the old or the new scenario

packages/ui5-proxy-middleware/src/base/utils.ts
petermuessig: You can also combine both scenarios - I wasn't sure about the content-type which is not set in the livereload scenario - that's why I kept it as is...

zdravko-georgiev (approved):
Code changes looks good
Excellent test coverage
Didn't tested locally

Thanks @petermuessig

packages/ui5-proxy-middleware/src/base/utils.ts
tobiasqueck: Ok, let's leave it as-is for now.

tobiasqueck (approved):
missing changeset was added and my questions answered since my last review

PR #1227, opened 2023-08-28
feat(ui-components) add UICreateSelect component

eouin:
Add ui-components UICreateSelect based on [https://react-select.com/home](https://react-select.com/home)

CleanShot 2023-08-28 at 17 52 28


gpeuc:
> @Edouard Ouin @Goran Peuc
> 
> why do we need another type of dropdown/combobox component? Which features are missing in existing components?Maybe we can add missing parts to existing components?

That was my initial thought as well. We have them already, we should just extend them a bit with missing functionality (creating new element through the input field). I am not clever enough to figure out should we create a new one, or extend existing one, I leave that to Devs. I just paint stuff. 😄

815are:
I checked combobox approach, my quick version was:
https://github.com/SAP/open-ux-tools/assets/90789422/04c1772f-b693-439c-92d5-37f992607c45

But it would require quite time to match same behaviour for combobox(keyboard navigation, loading) like it is in https://react-select.com/home - not sure if it worth it and not fully sure if fully behaviour could be matched(concern regarding keyboard navigation difference)

gpeuc:
@815are Ah yes, highlight with the select background, nice one.

815are:
@eouin 

I see sme styling issues like:
1. select `Two`
2. Search by `w`;
3. Hover over `Two`
4. Leave mouse from `Two`:

https://github.com/SAP/open-ux-tools/assets/90789422/4d6581f8-290e-4367-856c-0382bb52d94c

Also I noticed difference if we compare with combobox/dropdown.
If we hover selected entry using dropdown/combobox, then we do not see hover effect, but using new component hover effect is applied for selected entry.


Also I think there is issue with hcb theme:
Selected `Two:
image
Same with standard dropdown:
image

+ hover effect with hcb is also not same as with other dropdown menues

gpeuc:
@eouin @815are yea this is strange, but looks like a simple fix, just have proper highlight variables for mouse and keyboard controls. Can you guys tweak it?

815are (dismissed):
Checked:
1. tested manually - works ok and did not finds bugs. Styles are maintained similat to other dropdown components;
2. code of component is easy to read. But we have new variation of dropdown component with new dependency to third party library.

815are (commented):
@eouin @gpeuc 

One difference which mentioned before:
1. If item is selected and we hover, then:
1.1. UICreateSelect applies hover style on top of select:

https://github.com/SAP/open-ux-tools/assets/90789422/2df6ce58-1268-42a9-8843-d2c2c88b6a44

1.2. Combobox/Dropdown does not apply hover on top of selected item:

https://github.com/SAP/open-ux-tools/assets/90789422/58622ce1-9eb2-4bfc-938c-281fd20e43d7

devinea (changes requested):
react-select - MIT ✅ 
New component looks well structured
tests - Coverage of 55.1% is a bit low. Can you update in this?
Also it should be testing using react testing library instead of enzyme per https://github.com/SAP/open-ux-tools/issues/833
Please update this also.

devinea (approved):
Thanks @eouin for the updates.

PR #1229, opened 2023-08-29
Log a warning if the preview middleware is used with a path that also exists in the filesystem

tobiasqueck:
Detailed changes:

* `types/ui5.d.ts` - fix incorrect `byPath` signature
* `packages/ui5-proxy-middleware`
    * delete duplicated `ui5.d.ts` and change `tsconfig` to point to the `types` module
    * no changeset required
* `packages/preview-middleware`
    * use `byPath` instead of `byGlob` when we need to get exactly one file with a well defined path
    * log warning if the preview middleware is used with a path that also exists in the filesystem


zdravko-georgiev (approved):
Excellent test coverage
Didn't tested manually

nikmace (approved):
The code looks good.
Did not test locally.

Issue #1232, opened 2023-08-30
[moderate] @adobe/css-tools Regular Expression Denial of Service (ReDOS) while Parsing CSS


tobiasqueck:
### Description
Vulnerable version detected.

### Steps to Reproduce
1. run `pnpm audit`.'
2. See error

### Expected results
No error

### Actual results
Error

### Screenshots
![Screenshot 2023-08-30 at 10 02 59](https://github.com/SAP/open-ux-tools/assets/44933199/aa998cd2-c5fa-43ac-be2f-80f71015e4b6)

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1235, opened 2023-08-30
chore: updated dev dependencies of ui-components and verify overrides

Klaus-Keller:
Triggered by https://github.com/SAP/open-ux-tools/issues/1232 I had a look at updating the `devDependencies` of module `@sap-ux/ui-components` with following results:

**Update from `storybook@6` to `@storybook@7`**
There have been fundamental changes to storybook, I've followed migration guide https://storybook.js.org/docs/react/migration-guide. It required to modify the storybook scripts, added `.babelrc.json`, modified `.eslintrc.js` and `.storybook/main.js`. I also had do modify `.storybook/manager-head.html` as the `setStories` event is deprecated.
As a positive effect, overrides for storybook are not required anymore and dependency to `webpack` could be removed, storyboard uses now module `@storybook/react-webpack5`.

**Update `@testing-library/jest-dom`**
Unfortunately, update of `@testing-library/jest-dom` did not solve https://github.com/SAP/open-ux-tools/issues/1232. We also can't update to the very latest major version due to peer dependency to react 18.

**Other devDependencies**
I've also looked through other `devDependencies` of `@sap-ux/ui-components` and updated if possible.


devinea (approved):
Thanks @Klaus-Keller 
Storybook bundle size is good here  (opposed to my attempt in https://github.com/SAP/open-ux-tools/pull/1081 ) 
Package updates  ✅ 
Storybook file udpates ✅ 

Should probably add a changeset for ui5-components to release a new version.

Klaus-Keller:
> Should probably add a changeset for ui5-components to release a new version.

Was not sure about this, the only change I saw that could potentially alter the `dist` result is updated of `"sass": "1.66.1"`. But as you mentioned it, can't harm to increase the patch level. Will add changeset.

devinea:
> > Should probably add a changeset for ui5-components to release a new version.
> 
> Was not sure about this, the only change I saw that could potentially alter the `dist` result is updated of `"sass": "1.66.1"`. But as you mentioned it, can't harm to increase the patch level. Will add changeset.

storybook output folder `/storybook` is also published in the npm module.

devinea (approved):
Thanks for the cset

tobiasqueck (approved):
Thanks a lot for the cleanup of the overrides!

Issue #1248, opened 2023-09-01
BUG - Invalid manifest datasource key created


unseen1980:
### Description

The generator is using the datasource as a key name in the manifest.  In some cases this leads to an invalid property name, if the datasource contains forward slashes.

### Steps to Reproduce
1. Generate a Fiori elements application on BAS using E.g. Y05 with data source `/DMO/UI_FLIGHT_R_V2_VAN`
2. Open the generated `manifest.json` file.
3. Navigate the `dataSources` section.
4. Observe the error.

### Expected results

The dataSource key name is valid.

### Actual results

The data source key name is not valid, see https://github.com/SAP/ui5-manifest/blob/master/schema.json#L349 for the valid regex.

tobiasqueck:
@unseen1980 has this been fixed with the referenced PRs or is it still open?

unseen1980:
Completed

PR #1249, opened 2023-09-01
Fix/odata service writer/annotations key

unseen1980:
https://github.com/SAP/open-ux-tools/issues/1248 

Schema for manifest file does not allow ```/``` in key names. Only ```"^[a-zA-Z0-9_\\.\\-]*$"```


devinea (approved):
Thanks @unseen1980 
Changes and snapshot updates look appropriate. 
I haven't tested locally. 
changeset ✅

tobiasqueck (commented):
While it solves the problem, I am not in full favor of the solution because you are converting the string in two different places of the template with the same functionality. Aside from the duplicated coding, it is also error prone because both converted strings in the template must match therefore, if one is changed due to future issue, there is the risk that the other one is forgotten.
I'd prefer that you introduce an optional property `name` at https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-writer/src/types.ts#L26 and then if empty, you fill it with `technicalName.replace(/\//g, '_').replace(/^_/, '')` in https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-writer/src/data/defaults.ts#L10 and finally you refer to it in the template.

unseen1980:
> While it solves the problem, I am not in full favor of the solution because you are converting the string in two different places of the template with the same functionality. Aside from the duplicated coding, it is also error prone because both converted strings in the template must match therefore, if one is changed due to future issue, there is the risk that the other one is forgotten. I'd prefer that you introduce an optional property `name` at https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-writer/src/types.ts#L26 and then if empty, you fill it with `technicalName.replace(/\//g, '_').replace(/^_/, '')` in https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-writer/src/data/defaults.ts#L10 and finally you refer to it in the template.

Hi Tobias, please check again. Thanks

tobiasqueck (approved):
* fix is as expected
* the `defaults.ts` code is easier to understand now
* test snapshots show that the fix worked
* changeset correct

packages/odata-service-writer/templates/extend/manifest.json
```diff
@@ -6,15 +6,15 @@
                 "type": "OData",
                 "settings": {
                     "annotations": [<%if (locals.annotations && annotations.technicalName) {%>
-                        "<%- annotations.technicalName %>"<% if (locals.localAnnotationsName) {%>,<%}%><% } %><% if (locals.localAnnotationsName) { %>
+                        "<%- annotations.name %>"<% if (locals.localAnnotationsName) {%>,<%}%><% } %><% if (locals.localAnnotationsName) { %>
```
devinea, line 9: Would safer to change if to check `name` instead of `technicalName` as `name` is now referenced in the template. 
`"annotations": [<%if (locals.annotations && annotations.name) {%>`
Anyway from the code it looks like it always populated based on `setDefaultAnnotationsName`

devinea (approved):

packages/odata-service-writer/templates/extend/manifest.json
unseen1980: Thanks, I may change it in a future pr

Issue #1253, opened 2023-09-04
BUG - fpm-writer. Custom filter field. Fragment generation issues, when `eventHandler` is undefined or when existing event handler is passed 


815are:
### Steps to Reproduce
1st scenario.
1. Use `generateCustomFilter` API method -> https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/filter/index.ts#L52
2. Pass `eventHandler` property as `undefined`

Problem - generated xml still contains reference to js file, this snippet contains reference:
```
 core:require="{handler: 'dummy/project1/ext/fragment/id'}"
            selectedKey="{path: 'filterValues>', type: 'sap.fe.macros.filter.type.Value', formatOptions: { operator: 'dummy.project1.ext.fragment.Filename.filterItems' }}"
```

2nd scenario.
1. Use `generateCustomFilter` API method -> https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/filter/index.ts#L52
2. Pass `eventHandler` property as:
```
name: 'dummyName',
eventHandler: {
      fileName: 'Filename'
}
```

Problem - generated fragment contains reference to `dummyName` instead of `Filename`:
![image](https://github.com/SAP/open-ux-tools/assets/90789422/09420793-c8a1-4022-bf7a-e1cfc20a001a)

### Expected results
1st generated xml should not contain reference to js file if `eventHandler` is not passed
2nd generated fragment should contain reference to passed `eventHandler.fileName`

### Actual results
1st generated xml contains reference to js file if `eventHandler` is not passed
2nd generated fragment does not contain reference to passed `eventHandler.fileName`

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1254, opened 2023-09-04
fix: Custom filter field. Fragment generation issues, when `eventHandler` is undefined or when existing event handler is passed 

815are:
Issue #1253

Fixes for two scenarios
1st scenario.
1. Use `generateCustomFilter` API method -> https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/filter/index.ts#L52
2. Pass `eventHandler` property as `undefined`

Problem - generated xml still contains reference to js file, following snippet contains reference:
```
 core:require="{handler: 'dummy/project1/ext/fragment/id'}"
            selectedKey="{path: 'filterValues>', type: 'sap.fe.macros.filter.type.Value', formatOptions: { operator: 'dummy.project1.ext.fragment.Filename.filterItems' }}"
```

2nd scenario.
1. Use `generateCustomFilter` API method -> https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/filter/index.ts#L52
2. Pass `eventHandler` property as:
```
name: 'dummyName',
eventHandler: {
      fileName: 'Filename'
}
```

Problem - generated fragment contains reference to dummyName instead of Filename:
image

No new unit tests, because all scenarios are covered, but snapshots were not correct


AlinaGovoruhina (approved):
Snapshots are updated appropriately to the eventHandler test cases

schreckstefan (approved):
Tested OK:
- no negative influence in spec/app mod
- works fine in GD

PR #1256, opened 2023-09-05
[preview-middleware] cleanup of init script logic

tobiasqueck:
This PR changes the initialization of the sandbox by moving all script code into one better structured file that is configured as init script.

This is a preparation for further work on #1149. It will allow a cleaner way to extend the script.

## Details
* remove of old mockserver code from `load-reuse-libs` script that was never executed because the parameters triggering it were never set
* merging `load-reuse-libs` script and inline scripts from `sandbox.html` into one `sandboxInit.js` script
* keeping all templating logic in `sandbox.html` while `sandboxInit.js` is a static script that could be reused by multiple html files
* fix the static path of the `manifest.json` and while doing so, also added support for multiple apps registered in the sandbox FLP


packages/preview-middleware/src/base/flp.ts
```diff
@@ -147,6 +141,12 @@ export class FlpSandbox {
      * Add routes for html and scripts required for a local FLP.
      */
     private addStandardRoutes() {
+        // add route for the oninit script
+        this.router.get(`${dirname(this.config.path)}/sandboxInit.js`, (_req: Request, res: Response) => {
+            const initjs = readFileSync(join(__dirname, '../../templates/flp/sandboxInit.js'), 'utf-8');
+            res.status(200).contentType('text/javascript').send(initjs);
```
zdravko-georgiev, line 147:
```suggestion
           res.writeHead(200, {'Content-Type': 'text/html});
           res.write(html);
           res.end();
```

packages/preview-middleware/templates/flp/sandboxInit.js
```diff
@@ -0,0 +1,210 @@
...
+// Load custom library paths if configured
+var appUrls = bootstrapConfig.getAttribute("data-open-ux-preview-libs-manifests");
+var initPromise = appUrls ? registerComponentDependencyPaths(JSON.parse(appUrls)) : Promise.resolve();
+initPromise.catch(function (error) {
+    sap.ui.require(["sap/base/Log"], function (Log) {
+        Log.error(error);
+    });
+})
+.finally(function() {
+    // always set title, regiseter fonts and place the renderer
```
zdravko-georgiev, line 206: register 😉

zdravko-georgiev (commented):
I have just 2 small comments

packages/preview-middleware/templates/flp/sandboxInit.js
tobiasqueck: 30612d7

zdravko-georgiev:
@tobiasqueck `sandboxInit.js` fails to load
image

This is the configuration in my `ui5.yaml`

```
specVersion: "2.6"
metadata:
  name: project5
type: application
server:
  customMiddleware:
    - name: preview-middleware
      afterMiddleware: compression
      configuration:
        flp: 
          path: /zdravko.html
          libs: true
        debug: true
```

packages/preview-middleware/src/base/flp.ts
```diff
@@ -147,6 +141,14 @@ export class FlpSandbox {
      * Add routes for html and scripts required for a local FLP.
      */
     private addStandardRoutes() {
+        // add route for the oninit script
+        this.router.get(`${dirname(this.config.path)}/sandboxInit.js`, (_req: Request, res: Response) => {
```
zdravko-georgiev, line 145: Please change to `${dirname(this.config.path)}sandboxInit.js`. This will fix the issue that I reported.

tobiasqueck: No, because for a config path like `/my/path/index.html` it won't work but https://github.com/SAP/open-ux-tools/pull/1256/commits/cc5881a85b85e261d7bbff0cb4b6a0cfa4207efe should fix it

zdravko-georgiev (approved):
Very good test coverage
Manual testing with different configuration ✅

PR #1258, opened 2023-09-05
Fix/odata service writer/annotations key

unseen1980:
https://github.com/SAP/open-ux-tools/issues/1248


devinea (approved):
@unseen1980 
Thanks for the updates. Tested locally seems to work well.

tobiasqueck (approved):
* changeset ok
* change is clear
* tests are not affected

Issue #1257, opened 2023-09-05
BUG - Wrong link to error message displayed during deployment in BAS


mikicvi-SAP:
### Related Feature
Feature request: _issue number_

### Description
When deployment fails a link is displayed in the console, where more information about the deployment error can be found. But if deployment is done in BAS, then the link is using the virtual host from the destination, instead of the real host to the backend, e.g. link uses `http://esmclnt200:443`, where actually `https://ldciesm.abc.ade.domain:44300`should be used instead.

### Steps to Reproduce
Please check BCP issue 2280202476 

### Expected results
Working link is displayed in the console

### Actual results
Not working link is displayed in the console

### OS/Browser/Environment
OS:                      
- [x] Mac OS
- [x] Windows

Browser: 
- [x] Chrome
- [ ] Firefox
- [ ] Edge

Environment: 
- [x] Master
- [ ] Staging
- [ ] Production

--- 
### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue was missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Validated with latest release, working as expected.

PR #1265, opened 2023-09-06
fix(axios-extension): message bas deploy

mikicvi-SAP:
BUG - Wrong link to the error message displayed during deployment in BAS #1257

Adds additional note for a customer to replace the host in the URL.


packages/axios-extension/src/abap/message.ts
```diff
@@ -91,13 +102,29 @@ export function prettyPrintMessage({ msg, log, host }: { msg: string; log: Logge
...
+    isDest?: boolean;
+}): void {
     if (host && path) {
         const base = new URL(host).origin; // We only care for the origin value
         // Add this instruction to the user because of this bug in VS Code: https://github.com/microsoft/vscode/issues/144898
         // It undoes the encoding of special characters in the URL sent to the browser
         log.info('Please copy/paste this URL in a browser for more details:');
+        if (isDest) {
+            log.info(
+                '(Note: As the destination uses a virtual host, you will need to replace the host in the URL with the internal host that is configured with your SAP Cloud Connector)'
```
longieirl, line 125: @theStiggySpd this info message needs to be changed. Yes, we know its a destination but the URL defined in the destination properties, is not always a virtual host. Its only a virtual host when its on-premise. A suggestion might be;
```
(Note: You will need to replace the host in the URL with the internal host, if your destination is configured using an On-Premise SAP Cloud Connector)
```
In alot of cases, the URL defined is a direct mapping to the internal host mapping defined in the SCC but there is not always valid.

packages/axios-extension/test/abap/message.test.ts
```diff
@@ -20,9 +21,9 @@ describe('message helpers', () => {
 
             const infoMock = (log.info = jest.fn());
             const warningMock = (log.warn = jest.fn());
-            prettyPrintMessage({ msg: JSON.stringify(msg), log, host });
-            // log main message, two messages for the full url, and each detail
-            expect(infoMock).toBeCalledTimes(4);
+            prettyPrintMessage({ msg: JSON.stringify(msg), log, host, isDest });
```
tobiasqueck, line 24: I don't this test is valuable - not your change, the test itself already. I don't really think we need to check how often `log.info` is called.
Could we just change this one to check if infoMock is called (no matter how often) and have an additional test where we use a url matching `.dest` and check if the note has been logged?

tobiasqueck (commented):
* code looks good
* changeset is correct
* small comment about the test that could be improved - but I am ok if you reject that comment ;)

packages/axios-extension/src/abap/message.ts
mikicvi-SAP: Adjusted the message, if @theStiggySpd approves

packages/axios-extension/test/abap/message.test.ts
mikicvi-SAP: Thank you for the review.
Please review the new changes.

tobiasqueck (approved):
* changeset ok
* code is clean and easy to understand
* requested changes to the test have been applied

packages/axios-extension/src/abap/message.ts
theStiggySpd: Looks good @longieirl @mikicvi-SAP

devinea (approved):
changeset, msg text, code and tests are good.

PR #1282, opened 2023-09-13
fix(uitable): fix waitFor function

tom-sap:
fix sonar issue: https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=BUG&id=SAP_open-ux-tools&open=AYfrPj_2SN0lrgwATVOw

Fixes:
- The code smell: setTimeout doesn't accept async functions.
- Function updated to avoid going into an infinite loop, when the selector was not found.
- Delay for the rechecking for the selector changed from 100ms to 200ms to improve the performance a little bit.

@devinea 

fix https://github.com/SAP/open-ux-tools/issues/1283


devinea (approved):
Changeset ✅ 
Code changes look good. 
Quick test of ui table with storybook looks good.
Thanks @tom-sap

tobiasqueck (changes requested):
Could we please have a test for this one? The test coverage is a little low for our standards, or do you have a valid reason why we don't need it here?

Additionally, what is the reason for this change? First, I thought this is a refactoring because of a code smell but you changed the interface as well as the behavior (100ms vs. 200ms sleep time)

tobiasqueck:
> Hi Tobias,
> 
> 1. The code smell: setTimeout doesn't accept async functions.
> 2. I changed the interface to avoid the function to go into an infinite loop, when the selector was not found.
> 3. The 200 instead of 100 is due to performance: it will check 10 times, every 200ms, instead of checking every 100ms forever.
> 4. There was no test specific for this function, and I have no capacity to add it in this moment of time.
> 
> I can revert 2 and 3 if you think they are not necessary.

1-3 please add that to the description
4 Sorry, then the PR won't go in. Your work is not done until the code is tested.

devinea (approved):
Thanks for adding tests

tobiasqueck (approved):
Thank you for addressing the review feedback. Looks great now.

PR #1293, opened 2023-09-19
fix(preview-middleware): update README

zdravko-georgiev:
- Fixes some typos
- Remove documentation about not fully supported features


tobiasqueck (changes requested):
`developerMode` is fully supported, so why would you remove it?

packages/preview-middleware/README.md
```diff
@@ -61,15 +58,15 @@ server:
...
+With this configuration, the local FLP will be available at `/test/myFLP.html` and the log level is `debug`.
 ```Yaml
 server:
   customMiddleware:
   - name: preview-middleware
     afterMiddleware: compression
     configuration:
       flp: 
-        path: /preview.html
+        path: /test/myFLP.html
```
tobiasqueck, line 69: Ok, a different url, fine, but why is this required?

zdravko-georgiev:
> `developerMode` is fully supported, so why would you remove it?

It is fully supported yes, but only in combination with editing of adaptation projects. And since the editing is not fully supported, I also removed the `developerMode`, for now

packages/preview-middleware/README.md
zdravko-georgiev: Because `preview.html` is more or less currently taken. I also would like to emphasise that best practice is to have the HTML files in the `test` folder.

tobiasqueck:
> > `developerMode` is fully supported, so why would you remove it?
> 
> It is fully supported yes, but only in combination with editing of adaptation projects. And since the editing is not fully supported, I also removed the `developerMode`, for now

I am sorry but no. You can generate an adaptation-project using `@sap-ux/create` and that one will contain a setting `developerMode` true, so it is fully supported. The middleware is still a `0.x` version so things can change but in its current version it is supported.

zdravko-georgiev:
> > > `developerMode` is fully supported, so why would you remove it?
> > 
> > 
> > It is fully supported yes, but only in combination with editing of adaptation projects. And since the editing is not fully supported, I also removed the `developerMode`, for now
> 
> I am sorry but no. You can generate an adaptation-project using `@sap-ux/create` and that one will contain a setting `developerMode` true, so it is fully supported. The middleware is still a `0.x` version so things can change but in its current version it is supported.

OK, I added it back https://github.com/SAP/open-ux-tools/pull/1293/commits/0aae444aacdd33767328effbf28162ac1b754f0f

packages/preview-middleware/README.md
```diff
@@ -98,12 +96,12 @@ server:
...
         layer: CUSTOMER_BASE
         editors:
-          - path: /local/variant-editor.html
+          - path: /test/variant-editor.html
 ```
 
 
 ### Adaptation Project
-If you want to use the middleware in an adaption project, the additional `adp` object needs to be configured. This example would preview a local adaptation project merged with its reference application from the target system at `http://sap.example` and it will ignore certification validation errors. For adaptation projects, it is also recommended to add the `rta` configuration allowing to edit the project.
+If you want to use the middleware in an adaption project, the additional `adp` object needs to be configured. This example would preview a local adaptation project merged with its reference application from the target system at `http://sap.example` and it will ignore certification validation errors.
```
tobiasqueck, line 104: Please revert the changes in this section as well

zdravko-georgiev: Done with https://github.com/SAP/open-ux-tools/pull/1293/commits/33db76006778511bb50339eeb76fb9d8be910093

tobiasqueck (approved):
* review feedback has been applied
* readme is clearly improved
* no changeset needed, readme update will be shipped with the next patch/minor release

devinea (approved):
Readme updates seem appropriate 
No changeset is ok.

PR #1299, opened 2023-09-19
chore(uic): re-fix the sleep function tests

tom-sap:
Fix for: https://github.com/SAP/open-ux-tools/actions/runs/6237138309/job/16930023948?pr=1294
- add some wiggle room for measuring setTimeout execution time.


devinea (approved):
Thanks for the fix. lg.

Issue #1320, opened 2023-09-21
BUG - fiori-elements-writer `contextPath` issues


IainSAP:
### Description

Generate an app with ui5 ver > 1.94 and odata v4, specifying a navigation entity. Note that the object page context path should include the root entity.

Also note the ListReport page is incorrectly still using `entitySet`

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

IainSAP:
Closing as the original implementation will be reverted.

PR #1321, opened 2023-09-21
fix(fiori-elements-writer): Add root entity to object page context path

IainSAP:
https://github.com/SAP/open-ux-tools/issues/1320


packages/fiori-elements-writer/test/__snapshots__/lrop.test.ts.snap
```diff
@@ -14019,7 +14019,7 @@ EntityType=\\"com.sap.gateway.srvd.dmo.sd_travel_mduu.v0001.TravelAgencyType\\">
           \\"options\\": {
             \\"settings\\": {
               \\"editableHeaderContent\\": false,
-              \\"contextPath\\": \\"/Booking\\"
+              \\"contextPath\\": \\"/Travel/Booking\\"
```
schreckstefan, line 14022: @IainSAP That's not perfect, I guess: it should be the navigation property key, not the entitySet that the navigation points to: 
image

PR #1324, opened 2023-09-21
Revert "feat(fiori-elements-writer): Use `contextPath` for `entitySet`…

IainSAP:
This reverts commit 42d0cd9dbe7ffa695426cb135e701ea3bf70d605.

The introduction of this is not working as expected with some Fiori Tools and it may not be correctly implemented. More info and integration testing is required before this addition can be implemented.


devinea (approved):
Thanks @IainSAP i

tobiasqueck (approved):
* changeset correct
* explanation for reverting makes sense

schreckstefan (approved):

Issue #1336, opened 2023-09-26
BUG - fpm-writer. Custom filter field. Custom filter field overwrites existing extension fragment file


815are:
### Steps to Reproduce
1. Project which contains `ext/fragment/Test.fragment.xml` with content like;
```
<core:FragmentDefinition xmlns:core="sap.ui.core" xmlns="sap.m">
        <Input />
</core:FragmentDefinition>
```
2. Call `generateCustomFilter` and pass path to fragment file like:
```json
{
	"name": "id",
	"label": "label",
	"property": "CustomerName",
	"folder": "ext\\fragment",
	"target": "TravelList",
	"controlID": "id",
	"fragmentFile": "Test",
	"typescript": false,
	"required": false
}
```

Problem - 'ext/fragment/Test.fragment.xml' with be overwritten with new content

### Expected results
Existing extension files should not be overwritten

### Actual results
Existing extension file is overwritten

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1337, opened 2023-09-26
fix(fpm-writer): Custom filter field. Custom filter field overwrites existing extension fragment file

815are:
Issue #1336

Problem, if we have existing fragment.xml file in project and we pass path to such existing file using `generateCustomFilter`, then existing fragment file is overwritten with new content.

We avoid overwrite of existing fragments like codes here:
1. custom section -> https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/section/index.ts#L166 https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/section/index.ts#L115
2. custom page -> https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/page/custom.ts#L92
3. handler files -> https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/common/event-handler.ts#L90

### DIFF
--- a/.changeset/tidy-seahorses-move.md
+++ b/.changeset/tidy-seahorses-move.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fe-fpm-writer': patch
+---
+
+Custom Filter Field. Fix for issue, when creation of custom filter field overwrites existing extension fragment file.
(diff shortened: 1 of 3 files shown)


AlinaGovoruhina (approved):
Checking for existing fragment file looks good to me

schreckstefan (approved):
- Coding OK
- Unit test OK

PR #1338, opened 2023-09-26
fix(middlewares): make express peer dependecy

zdravko-georgiev:
Makes express peer dependency for `adp-tooling` and `preview-middleware`


tobiasqueck (approved):
Thank you, I think that is a better solution since 99% of the time it is used in environments where express is already present.

Klaus-Keller (approved):
Peer dependency for express is fine

Issue #1328, opened 2023-09-22
TBI - Improve unit tests in `preview-middleware-client`


tobiasqueck:
### Description (include screenshots)
The preview middleware client module is transpiled to UI5 code running in the browser. To test it we need to mock a lot of UI5 functionality. This lead to some tests not really testing functionality as they should but testing the execution of outgoing calls. 
We need to take some time to go through these tests and improve them by
* reusing as much of the mocking as possible across tests
* modify tests to test the expected output with given input as well as expected side-effects (calls into UI5) BUT not testing number of calls that depend on the internal implementation

### Value
Better test, likely better quality, and less regressions

### Architecture Elaboration
n/a

### Notes
This issues is created as part of a review of #1197

### Tasks
- [ ] Task 1

PR #1340, opened 2023-09-26
Fix/1328/improve unit test for middleware client

Jimmy-Joseph19:
Issue #1328

Tasks
- [x] Remove ui5 facade.
- [x] Reuse manual mocks
- [x] Test with arguments
- [x] remove node in tsconfig


voicis (changes requested):

packages/preview-middleware-client/test/unit/cpe/ui5Utils.test.ts
```diff
@@ -1,12 +1,10 @@
-import { createUi5Facade } from '../../../src/cpe/facade';
-import OverlayUtil from 'sap/ui/dt/OverlayUtil';
 import IconPool from 'mock/sap/ui/core/IconPool';
 import Component from 'sap/ui/core/Component';
 import { sapCoreMock } from 'mock/window';
-import OverlayRegistry, { mockOverlay } from 'mock/sap/ui/dt/OverlayRegistry';
 import type Element from 'sap/ui/core/Element';
+import { getComponent, getIcons } from '../../../src/cpe/ui5Utils';
```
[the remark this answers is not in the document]
Jimmy-Joseph19: Changed.

packages/preview-middleware-client/src/cpe/ui5Utils.ts
```diff
@@ -0,0 +1,45 @@
+import type { IconDetails } from '@sap-ux-private/control-property-editor-common';
```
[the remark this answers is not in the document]
Jimmy-Joseph19: Chnaged.

packages/preview-middleware-client/test/unit/cpe/changes/service.test.ts
```diff
@@ -336,10 +312,10 @@ describe('SelectionService', () => {
             })
         );
 
-        expect(window.fetch).toHaveBeenLastCalledWith('/preview/api/changes', {
-            body: '{"fileName":"id_1640106755570_203_propertyChange"}',
-            headers: { 'Content-Type': 'application/json' },
-            method: 'DELETE'
+        expect(fetchMock).toHaveBeenLastCalledWith('/preview/api/changes', {
+            'body': '{"fileName":"id_1640106755570_203_propertyChange"}',
```
tobiasqueck, line 316: Not a real problem but why did you put the property names in quotes?

packages/preview-middleware-client/test/unit/cpe/control-data.test.ts
```diff
@@ -197,9 +192,16 @@ describe('controlData', () => {
 
     test('buildControlData', async () => {
         // act
-        const result = await buildControlData(control as any, controlOverlay as any);
+        const result = await buildControlData(control as any, mockOverlay as any);
 
         // assert
         expect(result).toMatchSnapshot();
+
+        expect(control.getMetadata).toBeCalledWith();
```
tobiasqueck, line 200: If you don't check the calling parameter, just use `toBeCalled` please.

Now about the test itself: is it relevant that `control.getMetadata` has been called? Same for the other checks.
The question you should ask is what is `buildControlData` supposed to do and how can I verify it. Without looking at the function itself, I would assume that we need to ensure that the result is as expected, so I would assume that it would be a good idea to validate that the the provided control and overlay content matches the expectations of the result.

tobiasqueck (changes requested):
- changeset missing
- changes in `src` look good: facade removed and functions are called directly now (otherwise, I didn't see any change)
- a few comments about the tests added

packages/preview-middleware-client/test/unit/cpe/changes/service.test.ts
Jimmy-Joseph19: When i was trying out I thought matcher was not matching and I added quotes then forgot remove it.

packages/preview-middleware-client/test/unit/cpe/control-data.test.ts
Jimmy-Joseph19: adapted.

tobiasqueck (approved):
requested changes applied, good to go for me.

voicis (approved):
- refactoring of facade looks good
- requested changes done

Issue #1342, opened 2023-09-29
BUG - ui-components. UIFlexibleTable. Impossible to scroll within touch screen if touch over disabled row.


815are:
### Steps to Reproduce
1. Enable reordering for `UIFlexibleTable`, but disable drag for certain rows;
2. In chrome dev tools run with mobile device simulation;
3. Open storybook for `UIFlexibleTable.story.tsx` -> http://localhost:59041/?path=/story/tables-uiflexibletable--wrapping-table
4. Select for `Reorder restrictions`;
5. Make screen that scrollbar in table appears;
6. Try to scroll table while touching 2nd table;

### Expected results
Scroll should be working if user touches disabled row

### Actual results
Scroll does not work user touches disabled row

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1343, opened 2023-09-29
fix(ui-components): allow touch on disabled drag row

815are:
Issue #1342

Problem:
1. Wrong curos is applied when row is disabled for drag. Previously it `grab` cursor was applied;
2. Also scrolling with touch was not working if touch disabled row;


vadson71 (approved):

devinea (approved):
changeset| ✅ 
Code changes and test lg.

Issue #1326, opened 2023-09-22
BUG - ui-components. UITextField's input does not apply font family using vscode variable


815are:
### Steps to Reproduce
1. Check any textfield input:
![image](https://github.com/SAP/open-ux-tools/assets/90789422/d0c468e4-8c32-450b-944d-a41481443c00)

Problem - font-family is from `fluent-ui` instead of vscode `var(--vscode-font-family)`

### Expected results
Font-family for UITextField's input is from vscode `var(--vscode-font-family)`

### Actual results
Font-family is from `fluent-ui`

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1344, opened 2023-09-29
fix(ui-components): apply font family using vscode variable to UITextField input

815are:
issue #1326

Changes:
1. apply vscode font family to input of UITextField. Issue:
image
2. remove workaround from `control-property-editor`


Jimmy-Joseph19 (approved):
Approving.
Changes for removing workaround for CPE and adding the atttrubute to UITextInput.
Did not test locally.

broksy (approved):
Changes looks good and changeset is clear.

PR #1197, opened 2023-08-17
feat: Publish Control property editor

Jimmy-Joseph19:
This PR is for publishing control property editor from the SAP internal Fiori tools sources. 

Issue Associated: #1149 

The PR contains 2 new packages 

1. `@sap-ux/control-property-editor` - a web app based on react.
2. `@sap-ux-private/control-property-editor-common` - contains util methods and communication with react app and iframe

And enhancements to
1. `@sap-ux-private/preview-middleware`
2. `@sap-ux-private/preview-middleware-client`

PR #1347, opened 2023-10-02
add CPE FavIcon and GeneratorIssue for variant

Jimmy-Joseph19:
While testing after consuming the OSS CPE internally found that the generator for variant-management is missing. Was removed with [commit](https://github.com/SAP/open-ux-tools/pull/1197/commits/21294c6f1ee1e9a4d1cab8f32fb41f5b89aa6da5) PR #1197. Now generator property will be part of editors initialzed for flp.

Also added favicon and extended build to copyscript for favicon of the CPE into resources.


packages/preview-middleware-client/src/flp/WorkspaceConnector.ts
```diff
@@ -3,13 +3,33 @@ import ObjectStorageConnector from 'sap/ui/fl/write/api/connectors/ObjectStorage
...
+    };
+}
+
+function getFlexSettings() {
+    let result;
+    const bootstrapConfig = document.getElementById('sap-ui-bootstrap');
+    if (bootstrapConfig) {
+        result = bootstrapConfig.getAttribute('data-open-ux-preview-flex-settings');
+    }
+    return result;
```
voicis, line 19: Why not return parsed value already here?

packages/preview-middleware-client/src/flp/WorkspaceConnector.ts
```diff
@@ -3,13 +3,33 @@ import ObjectStorageConnector from 'sap/ui/fl/write/api/connectors/ObjectStorage
...
 
-const path = '/preview/api/changes'; 
+const path = '/preview/api/changes';
+interface Change {
+    support?: {
+        generator?: string;
+    };
+}
+
+function getFlexSettings() {
```
voicis, line 13: Add return type.

packages/control-property-editor/package.json
```diff
@@ -15,7 +15,7 @@
         "clean": "rimraf ./dist ./coverage *.tsbuildinfo",
         "build": "npm-run-all -l -s clean -p build:compile build:webapp",
         "build:compile": "tsc  --build ./tsconfig.json --pretty --dry",
-        "build:webapp": "node esbuild.js",
+        "build:webapp": "node esbuild.js && copyfiles  --up -1 \"./src/favicon.ico\" \"./dist\"",
```
voicis, line 18: Shouldn't this be part of the preview middleware? We were able to avoid copy tasks before and it would be good if we didn't need to add it here either.

voicis (changes requested):

packages/preview-middleware-client/src/flp/WorkspaceConnector.ts
Jimmy-Joseph19: adapted.

packages/control-property-editor/package.json
Jimmy-Joseph19: used https://www.npmjs.com/package/esbuild-plugin-copy?activeTab=readme for copying asset(favicon)

tobiasqueck (approved):
* changeset looks good
* sources adhere to coding standards
* test coverage is good
* successfully tested it locally

voicis (approved):
- code changes looks good
- tested locally
- test coverage is good

PR #1349, opened 2023-10-02
fix(fpm-writer): Stashed property of header sections

broksy:
In manifest templates `stashed` was defined as string type, but according to documentation, boolean property is required.


815are (approved):
Checked:
1. fix in template is valid - boolean property handled as boolean instead of string;
2. snapshots updated correctly;
3. changelog is maintained

AlinaGovoruhina (approved):
Conversion of stashed property to boolean and tests look appropriate

PR #1350, opened 2023-10-04
fix(ui-components): UIList. ListProps interface should be extended from fluent-ui IGroupedListProps

815are:
Enhance props type for UIList -> ListProps.
ListProps should be extended from IGroupedListProps, because it allows to pass any property from IGroupedListProps -> https://github.com/SAP/open-ux-tools/blob/main/packages/ui-components/src/components/UIList/UIList.tsx#L138

### DIFF
--- a/.changeset/serious-bobcats-listen.md
+++ b/.changeset/serious-bobcats-listen.md
@@ -2,4 +2,4 @@
 '@sap-ux/ui-components': patch
 ---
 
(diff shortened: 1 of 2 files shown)


vadson71 (approved):

Jimmy-Joseph19 (approved):
Approving the changes for extending the ListProps.

PR #1355, opened 2023-10-04
fix(ui-components): fix focus state for the split button

eouin:
Fix the UISplitButton focus state, fix the style


815are (approved):
Checked:
1. manually tested with storybook - focus border for split button appears same like for regular button;
2. changeset is maintained

devinea (approved):
changeset ✅ 
Code changes look good.

PR #1356, opened 2023-10-06
fix: build and editor script

tobiasqueck:
If you generate an adaptation project with the `npm init @sap-ux generate adpatation-project` command then you will have the following two problems:

Problem: The `npm run editor` is showing a 404
Reason: missing `/` in the `ui5.yaml`
Fix: add `/` to `ui5.yaml`

Problem: After adding a custom fragment to the project, `npm run build` will fail with an error when executing the `generateFlexChangesBundle` task
Reason: the task is not made for adaptation project
Fix: exclude task

Problem: After running `npm run build` the `dist` folder contains a useless `Component-preload.js` file
Reason: the `generateComponentPreload` task generates it
Fix: exclude task

### DIFF
--- a/.changeset/bright-fishes-fix.md
+++ b/.changeset/bright-fishes-fix.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/adp-tooling': patch
+---
+
+Fix: build and editor script for generated projects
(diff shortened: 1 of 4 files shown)


Klaus-Keller (approved):
Thanks @tobiasqueck, changes to the path for start and template for build script look good.
- did not test locally, just visual review
- test coverage looks good

Approved from my side.

nikmace (approved):
The changes look good.
Did not test locally.

PR #1357, opened 2023-10-06
fix - use patch version 0 for ui5 types semver for more reliable installs

devinea:
After the UI5 CDN's are updated, sometimes there is a small delay releasing the patch versions for new ui5 types module releases to npmjs.com.

This can lead to the npm install failures for newly created projects. 
This fix changes to use the 0 patch in the semver (`~x.x.0`) for the `@sapui5/types` and `@sapui5/ts-types-esm` modules
As a result the latest available patch is installed.
This reverts to this particular behaviour as it was before https://github.com/SAP/open-ux-tools/pull/1076.


longieirl (approved):

tobiasqueck (approved):
* changeset ok
* code changes are straight forward
* test snapshots updated as expected

devinea:
Thanks @longieirl @tobiasqueck

PR #1361, opened 2023-10-10
fix: createRenderer method to support maintenance versions

Jimmy-Joseph19:
createRenderer method argument order must be maintained for maintenance versions.


voicis (changes requested):
Please add a change set

zdravko-georgiev (approved):
Code looks good
Changeset added
Didn't tested manually

voicis (approved):
- code looks good
- test coverages is great
- tested with UI5 version 1.84

PR #1363, opened 2023-10-10
fix(preview-middleware): add sap.ui.rta to library preloads

zdravko-georgiev:
This fix adds the lib `sap.ui.rta` to the list of the libs of SAPUI5 libs that are "preloaded" when the application is started in Runtime Adaptation mode. It is needed to ensure the correct starting of RTA for older UI5 versions.


Klaus-Keller:
Hi @zdravko-georgiev, could you please have a look at the [code smell](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=1363&id=SAP_open-ux-tools&open=AYsZr8jAB3JCV_RVxKBx)?

zdravko-georgiev:
> f8addb4

@Klaus-Keller Done with https://github.com/SAP/open-ux-tools/pull/1363/commits/f8addb48d93a85eb039f6872311e734fe6453cc2. Please check again

packages/preview-middleware/src/base/flp.ts
```diff
@@ -174,6 +174,40 @@ export class FlpSandbox {
...
 
+    /**
+     * Gets the FLP sandbox for an editor.
+     *
+     * @param _req (not used) original request
+     * @param res response of the request
+     * @param rta runtime authoring configuration
+     * @param editor editor configuration
+     */
+    private async getSandboxForEditor(_req: Request, res: Response, rta: RtaConfig, editor: Editor) {
```
tobiasqueck, line 185: I assume the extraction of the code into a separate function was done to reduce the cognitive complexity, however, if we extract it, let's do it slightly differently to keep the functionality better encapsulated.
Suggestion: have a`private async function generateSandboxForEditor(rta: RtaConfig, editor: Editor): Promise<string>` that is returning the templated `html` and then keep in the calling function

```
this.router.get(previewUrl, (async (req: Request, res: Response) => {
  const html = await this.generateSandboxForEditor(rta, editor);
  res.status(200).contentType('html').send(html);
```

As a result you have nicely separated the templating and the request handling.

tobiasqueck (changes requested):
* changeset is good
* changes make sense
* not locally tested
* small change requested to reduce cognitive complexity

packages/preview-middleware/src/base/flp.ts
zdravko-georgiev: Thanks Tobias, will do.

zdravko-georgiev: Done with https://github.com/SAP/open-ux-tools/pull/1363/commits/5fde330ca07652a4719a9239001c3e336b7f8f34

tobiasqueck: Perfect. Btw, I just noticed that there is no `await` anymore, so we could actually remove the async from the handler and remove the type case, right?

zdravko-georgiev: Right, done with https://github.com/SAP/open-ux-tools/pull/1363/commits/f8d3e76fae834a14ab2c303dabd4845f5b558b77

tobiasqueck (approved):

Klaus-Keller (approved):
Code change to include `sap.ui.rta` for editor looks good, thanks @zdravko-georgiev.
- code is easy to understand
- test coverage is great
- no code smells
- did not test locally

Approved from my side

Issue #1366, opened 2023-10-10
BUG - `@sap-ux/app-config-writer` writes optional values when not provided


IainSAP:
### Description

If values are not provided for the optional navigation config properties `title` and `subTitle` empty strings are being written to the manifest. This is not the expected behaviour as consumers needs to be able to omit optional properties from being written.

### Steps to Reproduce
Update the unit tests omitting the optional properties `title` and/or `subTitle`. Observe that the manifest entries have been generated as empty strings.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1368, opened 2023-10-10
fix(@sap-ux/app-config-writer): dont write optional inbound config properties if not provided

IainSAP:
Fix for inbound-navigation config writer
- Fix for #1366 
- Update tests


.changeset/fluffy-turtles-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/create': minor
```
IainSAP, line 2: wrong cset

devinea (approved):
changeset minor ok
code changes and test updates lg.

zdravko-georgiev (approved):
Code changes looks good
Excellent test coverage
Didn't tested manually

slin-sap (approved):
- changeset looks good. It could also be considered as minor since the modified functions changed returned type definition. But patch level also seems ok since the new return type is less strict.  
- unit tests updated to cover the fixed bug scenario.
-  changes are easy to follow and reason to change has been well commented.

IainSAP:
> * changeset looks good. It could also be considered as minor since the modified functions changed returned type definition. But patch level also seems ok since the new return type is less strict.
> * unit tests updated to cover the fixed bug scenario.
> * changes are easy to follow and reason to change has been well commented.

Thanks @slin-sap it is a minor bump :)

PR #1311, opened 2023-09-20
feat(preview-middleware): add an option for setting the UI5 theme

zdravko-georgiev:
Adds a configuration option for setting the UI5 theme. This is useful when one wants to use the `preview-middleware` in an application with an older UI5 version that e.g. doesn't support the `sap_horizon` theme.

PR #1371, opened 2023-10-12
Property bindingString missing for getBindingInfo result

Jimmy-Joseph19:
1. ``bindingString` property was replaced with `path` which caused regression


packages/preview-middleware/src/base/flp.ts
```diff
@@ -30,7 +30,7 @@ export type EnhancedRouter = Router & {
 /**
  * Default theme
  */
-const DEFAULT_THEME = 'sap_horizon';
+const DEFAULT_THEME = 'sap_belize';
```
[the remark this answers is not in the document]
Jimmy-Joseph19: adapted

packages/preview-middleware/src/base/flp.ts
```diff
@@ -142,11 +142,12 @@ export class FlpSandbox {
         const supportedThemes: string[] = (manifest['sap.ui5']?.supportedThemes as []) ?? [DEFAULT_THEME];
         const ui5Theme =
             this.config.theme ?? (supportedThemes.includes(DEFAULT_THEME) ? DEFAULT_THEME : supportedThemes[0]);
+        const depsLibs = this.rta ? ['sap.ui.rta'] : [];
         this.templateConfig = {
             basePath: relative(dirname(this.config.path), '/') ?? '.',
             apps: {},
             ui5: {
-                libs: Object.keys(manifest['sap.ui5']?.dependencies?.libs ?? {}).join(','),
+                libs: [...Object.keys(manifest['sap.ui5']?.dependencies?.libs ?? {}), ...depsLibs].join(','),
```
zdravko-georgiev, line 150: Please remove, I added the missing `sap.ui.rta` with https://github.com/SAP/open-ux-tools/pull/1363 and it is added only for the cases when you want to start Runtime Adaptation.

packages/preview-middleware/src/base/flp.ts
```diff
@@ -30,7 +30,7 @@ export type EnhancedRouter = Router & {
 /**
  * Default theme
  */
-const DEFAULT_THEME = 'sap_horizon';
+const DEFAULT_THEME = 'sap_fiori_3';
```
zdravko-georgiev, line 33: Do we really need to change the default theme. Can we set the theme by using configuration https://github.com/SAP/open-ux-tools/pull/1311?

zdravko-georgiev (changes requested):

packages/preview-middleware/src/base/flp.ts
Jimmy-Joseph19: Is there any way to check ui5 version and set this? i did based on the below chart because 'sap_horizon' is not supported for maintenance version
https://sapui5.hana.ondemand.com/sdk/#/topic/38ff8c27b022475a92b591bcf6262551

zdravko-georgiev: You can get the UI5 version from the `manifest.json`. Additionally one needs to check also the `yaml` configuration of the `ui5-proxy-middleware`.  Hence we need something like https://github.com/SAP/open-ux-tools/blob/e392caa84142d80d69f474309c754f60d80e3698/packages/ui5-proxy-middleware/src/base/utils.ts#L204

Jimmy-Joseph19: then i will pass theme internally during flp initilization. Can i expose the function here
https://github.com/SAP/open-ux-tools/blob/e392caa84142d80d69f474309c754f60d80e3698/packages/ui5-proxy-middleware/src/index.ts#L1

zdravko-georgiev: No, please create a new function in the `preview-middleware`

packages/preview-middleware/src/base/flp.ts
Jimmy-Joseph19: removed

zdravko-georgiev: thanks

zdravko-georgiev:
@Jimmy-Joseph19 I see 100% coverage, but I do not see an unit test for this change. Could you please add also a corresponding unit test for this feature?

tobiasqueck:
The description of the PR and the actual change don't seem to match anymore ;)

packages/preview-middleware-client/src/cpe/selection.ts
```diff
@@ -219,8 +219,9 @@ export class SelectionService implements Service {
                 this.appliedChangeCache.delete(changeId);
                 return;
             }
-            const info: { path?: string } = runtimeControl.getBindingInfo(propertyName);
-            const newValue = info?.path ?? e.getParameter('newValue');
+            const info = runtimeControl.getBindingInfo(propertyName) as { path?: string; bindingString?: string };
+            // maintenance version 1.84.35 does not have 'path' property but 'bindingString'
+            const newValue = info?.path ?? info?.bindingString ?? e.getParameter('newValue');
```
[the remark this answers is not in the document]
Jimmy-Joseph19: I just tested with the latest actually there is no `path`. I have removed it now. i guess the confusion was that the return type from the @sapui5/types does not have 'bindingString' and was replaced with `path` without testing in earlier PR.

packages/preview-middleware/src/base/flp.ts
zdravko-georgiev: This will be handled in a different PR.

zdravko-georgiev (approved):
code changes looks good
review comments addressed
excellent test coverage

.changeset/nasty-candles-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux-private/preview-middleware-client': patch
+'@sap-ux/preview-middleware': patch
```
devinea, line 3:
```suggestion
```
No longer needed

devinea (commented):
Code and test ✅ 
changeset needs a small edit.

.changeset/nasty-candles-drop.md
tobiasqueck: I think we need both since `@sap-ux-private/preview-middleware-client` is not released but just packaged. If we remove the patch for the other one then we wouldn't get a meaningful changelog entry for the released module.

tobiasqueck (commented):
* changeset looks good
* code is straightforward and simple
* tested properly with jest
* I did NOT test with different UI5 versions manually

tobiasqueck (approved):

.changeset/nasty-candles-drop.md
devinea: ok

PR #1375, opened 2023-10-16
Same license reference in all modules

tobiasqueck:
The PR consolidates the usage of "Apache-2.0" as license reference in all `package.json` files instead of using "SEE LICENSE IN LICENSE".

This is **NOT** a change of the license because the `LICENSE` file included in all modules is the Apache 2.0 license, this is just a change of the reference in the `package.json` to be consistent.

I don't think this needs to be released but can simply get out with the next change of the affected modules, therefore, I didn't add a changeset.


devinea (commented):
Added some more changes to be consistent in all modules. 
Maybe best to add a patch Changeset so that the modules on npmjs show the correct license.
https://www.npmjs.com/package/@sap-ux/store

devinea (approved):
Changeset and updates look good now. 
thanks

tobiasqueck:
@zdravko-georgiev any chance you could review as well?

zdravko-georgiev (approved):
License changes looks good

PR #1376, opened 2023-10-16
fix(axios-extension): Apply additional info if endpoint is a destination

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/1257

- Additional flows need to have the info logger enabled, this helps the developer to understand the endpoint exposed needs to be updated locally to reflect their internal host
- Additiona tests added

### DIFF
--- a/packages/axios-extension/src/abap/ui5-abap-repository-service.ts
+++ b/packages/axios-extension/src/abap/ui5-abap-repository-service.ts
@@ -156,6 +156,7 @@ export class Ui5AbapRepositoryService extends ODataService {
         this.log.debug(`Payload:\n${payload}`);
         const config = this.createConfig(bsp.transport, testMode, safeMode);
         const frontendUrl = this.getAbapFrontendUrl();
+        const isDest = /\.dest\//.test(config.baseURL);
         try {
             const response: AxiosResponse | undefined = await this.updateRepoRequest(!!info, bsp.name, payload, config);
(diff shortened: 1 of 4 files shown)


packages/axios-extension/src/abap/ui5-abap-repository-service.ts
```diff
@@ -204,19 +206,25 @@ export class Ui5AbapRepositoryService extends ODataService {
         const config = this.createConfig(bsp.transport, testMode);
         const host = this.getAbapFrontendUrl();
         const info: AppInfo = await this.getInfo(bsp.name);
+        const isDest = /\.dest\//.test(config.baseURL);
```
tobiasqueck, line 209: Is there any other usage for the `isDest` flag except the usage when logging? If not, why don't we do the check in `prettyPrintMessage`? That would simplify the signature of the function and would remove the usage of the variable in multiple places.

longieirl: Yes, its only used for logging but the issue is, the host being passed down to the print methods is passing the public URL rather than the BAS destination URL.

tobiasqueck: ok

cianmSAP (approved):
- Updates to code as required
- Haven't tested
- Changeset ✅

Issue #1313, opened 2023-09-20
TBI - Add new APIs for validating package and transport request number in deploy-tooling 


slin-sap:
### Description (include screenshots)

In deploy-test mode, a new validation phase is going to be added before actual deployment. This phase validates if package and transport request number in deploy config file are valid. The detailed requirements are as follows.

- The validation will be implemented in open source repo. Validation is only triggered when running `deploy-test`. It will not be triggered for deploy or undeploy.
- Validation is executed after `create transport during deploy` and before running test deployment. Error found during validation do not terminate `deploy-test` execution. `deploy-test` executes regardlessly. The report is shown after test deployment completes. 
- Validation should check the inputs in deploy config (ui5-deploy.yaml) are valid. E.g. this involves checking app name, package name, transport number are valid (e.g. input length limit, regexp checks etc). Exactly same validation when user manually entering the inputs during deploy configuration.
- Invalid inputs in deploy config will be reported. If all inputs are valid, the validation report shows a single entry: 
        ✅ SAPUI5 ABAP Repository follows the rules of creating BSP application
   Otherwise, individual input format issues will be reported. E.g.
        ❌  Deploy configuration `app.name` is invalid.
        ❌  Deploy configuration `app.package` is invalid.
- Package name and transport number are validated against ADT service. Possible report statues as follows. Text in brackets is for explaining the requirements. It will not be shown in the report in terminal.
        ✅ Package is found (Backend system returned package)
        ❌ Package does not exist  (Backend system cannot find the package)
        ❓ Package could not be validated. Please check manually.   (Error connecting to backend)
        ✅ Transport Request is found (Backend system returned TR)
        ✅ Transport Request is not required for local package.
        ❌ Transport Request does not exist  (Backend system cannot find the package)
        ❓ Transport Request  could not be validated. Please check manually.
- This check might not be required since the report is already run in `deploy-test` only.   
        ✅ /UI5/ABAP_REPOSITORY_SRV in Test Mode
- The validation report status icons might not be exactly the same. But will provide similar icons for tick, cross, and question mark that are available in terminal logs. 

### Value
Support new pre-deploy validation feature mentioned above. Validation results help users of deploy-tooling to identify invalid inputs. 

#### axios-extension dependency

[...]

PR #1387, opened 2023-10-18
fix: deploy test mode validation 2

slin-sap:
#1313


packages/deploy-tooling/src/base/validate.ts
```diff
@@ -99,7 +99,7 @@ export function formatSummary(summary: SummaryRecord[]): string {
                     statusSymbol = yellow('?');
                     break;
             }
-            return `${statusSymbol} ${next.message}`;
+            return `${'    '}${statusSymbol} ${next.message}`;
```
devinea, line 102:
```suggestion
            return `    ${statusSymbol} ${next.message}`;
```
Should the `${}` be removed here?

packages/deploy-tooling/src/base/validate.ts
```diff
@@ -227,6 +227,12 @@ async function validatePackageWithAdt(
         return;
     }
 
+    // ADT expects input package
+    let inputPackage = input.package;
+    if (inputPackage.toUpperCase() === '$TMP') {
+        inputPackage = '$TMP';
+    }
```
devinea, line 234: I think it could make sense to upper case all package names for ABAP.
I remember correctly they are always upper case.

devinea (approved):
@slin-sap Code changes, tests and changeset look good.

packages/deploy-tooling/src/base/validate.ts
slin-sap: Thanks @devinea. I haven't got chance to test if small case package names other than '$tmp' would work for deploy. Currently we are treating it as case sensitive check and just give exception to '$tmp'. Otherwise, we report as an invalid input, which is not informative and non-blocking.

theStiggySpd (approved):
Thank @slin-sap 
Overall updated tests, changeset and code looks good.
I did not test locally.

For future, there are some good extensions which can help catch elusive typos :-)

tobiasqueck (approved):
* changesets are ok
* code is easy to understand and adheres to our guidelines
* tests properly enhanced

PR #1391, opened 2023-10-20
feat(preview-middleware): export initAdp function

zdravko-georgiev:
Exports the `initAdp` function so that it can be use in `@sap/ux-ui5-tooling`.


tobiasqueck (commented):
You didn't export the `initAdp` function from the module (`src/index.ts`). Isn't that what you need for a better integration into the Fiori tools?

zdravko-georgiev:
> You didn't export the `initAdp` function from the module (`src/index.ts`). Isn't that what you need for a better integration into the Fiori tools?

Yes, thanks 😅

tobiasqueck (changes requested):
* changeset ok
* changes are straight forward
* could you also move / copy a test case so that we test the `initAdp` function directly - it is public now so we need to ensure that it stays as it is

zdravko-georgiev:
> * changeset ok
> * changes are straight forward
> * could you also move / copy a test case so that we test the `initAdp` function directly - it is public now so we need to ensure that it stays as it is

@tobiasqueck Done with https://github.com/SAP/open-ux-tools/pull/1391/commits/06a8351bd4db179438817f7252ec12d6ec3cad83, please re-review

tobiasqueck (approved):
Thanks for the additional test

PR #1392, opened 2023-10-20
fix(env-check): stabilise destination unit tests

cianmSAP:
Mock axios get requests


tobiasqueck (approved):
- changeset is ok (but would have also been ok without changeset since it is just a unit test)
- tests are better because mocking happens on a lower layer - which is my personal prefernence

zdravko-georgiev (approved):
Unit test change looks good

Issue #1395, opened 2023-10-23
BUG - 'flexibilityServices' has a misconfiguration: Multiple Connector configurations were found to write into layer


nikmace:
### Related Feature
Feature request:  #1158 
### Description

When running Control Property Editor with Adaptation Project making a change and then trying to save the changes the application breaks. The following is the error (also see screenshot below): 

```
Sorry, a technical error occurred. Please try again later, or contact your administrator.
Reason:
Error: sap.ui.core.Configuration 'flexibilityServices' has a misconfiguration: Multiple Connector configurations were found to write into layer: VENDOR
at o (http://localhost:8080/resources/sap/ui/fl/library-preload.js:595:607)
```

### Steps to Reproduce
1. Go to https://github.tools.sap/ux/open-ux-tools-playground and run `npm run editor` inside apps/adp.v2app.
2. Make some changes and try to save them.
3. A dialog with the error will appear.

### Expected results
The changes are saved.

### Actual results
Changes are not saved, error is shown.

### Screenshots
![image](https://github.com/SAP/open-ux-tools/assets/39880364/5ddc74eb-a4bf-4bf6-9bd6-7ef7fed12f5a)

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [X] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1397, opened 2023-10-24
fix(preview-middleware): set layers configuration for the LocalStorageConnector

zdravko-georgiev:
Fixes: #1395 
When multiple connectors are used, then one needs to define also which connector is responsible for which layer. The `WorkspaceConnector` is responsible for the `VENDOR` and `CUSTOMER_BASE` layers by default. Since the `LocalStorageConnector` is responsible for all layers by default, the layers `VENDOR` and `CUSTOMER_BASE` needs to be excluded for the LocalStorageConnector. This done by setting `"layers": ["CUSTOMER", "USER"]` for the `LocalStorageConnector`

### DIFF
--- a/.changeset/beige-glasses-act.md
+++ b/.changeset/beige-glasses-act.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/preview-middleware': patch
+---
+
+Fixes the configuration of the LocalStorageConnector to avoid conflicts with the WorkspaceConnector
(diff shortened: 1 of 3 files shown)


zdravko-georgiev:
> The code looks good. Coverage is excellent. Tested manually for an Adaptation Project, changes are saved without a problem.

Thanks Nikita!

longieirl (approved):
Changes look small and isolated, covered by tests.

Klaus-Keller (approved):
@zdravko-georgiev, configuration for layers of LocalStorageConnector looks good, approved from my side.

Issue #1389, opened 2023-10-19
FEATURE - Add ui5 library prompting module


IainSAP:
### Description

As a developer I want to build a generator for UI5 libraries. As such in addition to the already available `@sap-ux/ui5-library-writer` I need a way to gather user inputs from the CLI or other user interfaces that supports Inquier prompting format

### Technical Design

New modules:

- `@sap-ux/ui5-library-inquirer` (main entry point)
- `@sap-ux/ui5-info` (supporting module - allow retrieval and processing of UI5 version info for use by UI5 version prompts)

### Tasks
- [x] Add `ui5-info` : provides UI5 version info
- [x] Add new validators to `project-input-validators`
- [x] Add `ui5-library-inquirer`: prompt module for UI5 library generation
- [x] Unit Tests added
- [ ] Test Cases defined

IainSAP:
Closing as completed

PR #1399, opened 2023-10-24
feature(ui5-info): Adds module for UI5 version info

IainSAP:
https://github.com/SAP/open-ux-tools/issues/1389

Adds new module required by `@sap-ux/ui5-library-inquirer`

- Adds new module `@sap-ux/ui5-info` to provide UI5 version info retrieval functionality
- Adds tests to cover


.changeset/afraid-dryers-happen.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ux/fiori-elements-writer': minor
+'@sap-ux/ui5-info': minor
```
devinea, line 3: ui5-config should have a entry here too

packages/ui5-info/README.md
```diff
@@ -0,0 +1,33 @@
+# @sap-ux/ui5-info
+
+Gets UI5 version info for available UI5 versions.
+
+## Installation
+Npm
+`npm install --save @sap-ux/ui5-library-writer`
```
devinea, line 7: Please update readme

packages/ui5-info/jest.config.js
```diff
@@ -0,0 +1,7 @@
+const config = require('../../jest.base');
+config.snapshotFormat = {
+    escapeString: false,
+    printBasicPrototype: false
+};
```
devinea, line 5:
```suggestion
```
I think false is the default so this can be removed.

packages/ui5-info/package.json
```diff
@@ -0,0 +1,50 @@
...
+    },
+    "files": [
+        "LICENSE",
+        "dist",
+        "templates",
+        "!dist/*.map",
+        "!dist/**/*.map"
+    ],
+    "dependencies": {
+        "@sap-ux/fiori-elements-writer": "workspace:*",
```
devinea, line 34: Won't this cause a circular dependency later 🤔

packages/ui5-info/src/commandRunner.ts
```diff
@@ -0,0 +1,37 @@
...
+            });
+            spawnedCmd.on('close', (errorCode) => {
+                if (errorCode !== 0) {
+                    reject(new Error(`Command failed, \`${cmd} ${args.join(' ')}\`, ${stack.join(', ')}`));
+                }
+                resolve(response);
+            });
+        });
+    }
+}
```
devinea, line 37: I think command runner should be in a separate module ideally

packages/ui5-info/src/ui5-info.ts
```diff
@@ -0,0 +1,432 @@
...
+/**
+ * Lowest UI5 version to return (not necessarily the min supported version)
+ */
+export const DEFAULT_MIN_UI5_VERSION = '1.65.0';
+const LatestVersionString = 'Latest';
+const DefaultVersion = LatestVersionString;
+
+export const DEFAULT_UI5_VERSIONS = [
+    DefaultVersion,
+    '1.104.0',
```
devinea, line 21: This should be extended with newer versions and remove no longer supported versions
https://sapui5.hana.ondemand.com/versionoverview.html

sonar-project.properties
```diff
@@ -3,7 +3,7 @@ sonar.projectKey=SAP_open-ux-tools
 sonar.sources=.
 sonar.inclusions=**/*.ts, **/*.tsx
 sonar.exclusions=**/*.test.ts, **/*.test.tsx, **/*.story.tsx, **/examples/**/*, **/templates/**/*, **/test/**/*
-sonar.cpd.exclusions=**/i18n.ts, **/fiori-elements-writer/src/packageConfig.ts, **/fiori-freestyle-writer/src/packageConfig.ts
+sonar.cpd.exclusions=**/i18n.ts, **/fiori-elements-writer/src/packageConfig.ts, **/fiori-freestyle-writer/src/packageConfig.ts, **/ui5-info/src/ui5VersionFallback.ts
```
devinea, line 6: Why is this excluded?

devinea (commented):
@IainSAP Overall looks great. I have a few comments.

packages/ui5-info/package.json
IainSAP: NO I dont think so. Writers should not be dependent on prompting related modules.

packages/ui5-info/src/commandRunner.ts
IainSAP: Which module please?

sonar-project.properties
IainSAP: Its essentially a json definition which inherently contains much replication. Please suggest how to remove otherwise as this seems appropriate to me.

.changeset/afraid-dryers-happen.md
IainSAP: Hmm, I removed 2 unused dev dependencies so there is no practical difference in the delivered module from npmjs so I think its not necessary?

packages/ui5-info/jest.config.js
IainSAP: The defaults are true in both cases : https://github.com/SAP/open-ux-tools/blob/main/jest.base.js#L25 so it required to change the behaviour. IMO the defaults are leading to hard to read and bloated snapshots.

packages/ui5-info/src/commandRunner.ts
```diff
@@ -0,0 +1,37 @@
+import { spawn } from 'child_process';
+
+/**
+ *
+ */
+export class CommandRunner {
```
Klaus-Keller, line 6: Instead of spawning a process we could also do an HTTP call to
https://registry.npmjs.org/@sapui5/distribution-metadata/ to list the versions, see https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md#package-endpoints.

In this case we would need to allow to pass the host to the registry (default to 'https://registry.npmjs.org') in case the consumer uses an own NPM registry.

If we stick to `spawn`, I would not create `CommandRunner` as a generic class that allows to execute any command, but provide a function to run the exact call we need, like

```typescript
export function getUi5NpmVersions(): Promise<string[]> {
    return new Promise((resolve, reject) => {
        const stack: any = [];
        const cmd = process.platform === 'win32' ? 'npm.cmd' : 'npm';
        const args = ['show', '@sapui5/distribution-metadata', 'versions', '--no-color'];
        const spawnedCmd = spawn(cmd, args, {});
        spawnedCmd.stdout.setEncoding('utf8');
        let response: string;
        spawnedCmd.stdout.on('data', (data: Buffer) => {
            response = data.toString();
        });
        spawnedCmd.stderr.on('data', (data) => {
            stack.push(data.toString());
        });
        spawnedCmd.on('error', (error) => {
            reject(new Error(`Command failed with error: ${error.message}`));
        });
        spawnedCmd.on('close', (errorCode) => {
            if (errorCode !== 0) {
                reject(new Error(`Command failed, \`${cmd} ${args.join(' ')}\`, ${stack.join(', ')}`));
            }
            resolve(
                response
                    .replace(/[\r?\n|[\] ']/g, '') // Remove all chars, new lines and empty space
                    .trim()
                    .split(',')
            );
        });
    });
}
```

Klaus-Keller (commented):
See my comments

packages/fiori-elements-writer/src/data/templateAttributes.ts
```diff
@@ -1,9 +1,9 @@
 import { OdataVersion, TemplateType } from '../types';
 
 // first version with SAP Fiori 3 theme
-export const oldestSupportedUI5Version = '1.65.0';
-
+export const oldestSupportedUI5VersionV2 = '1.65.0';
```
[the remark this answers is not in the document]
IainSAP: Thats seems burdensome on consumers to have to process the TemplateOptions no?

packages/fiori-elements-writer/src/index.ts
```diff
@@ -175,5 +180,11 @@ async function generate<T extends {}>(basePath: string, data: FioriElementsApp<T
     return fs;
 }
 
-export { generate, FioriElementsApp, App };
+export {
+    generate,
+    FioriElementsApp,
+    App,
+    oldestSupportedUI5VersionV2 as FE_MIN_UI5_VERSION_V2,
+    oldestSupportedUI5VersionV4 as FE_MIN_UI5_VERSION_V4
```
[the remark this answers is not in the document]
IainSAP: See above, I think we need to not force consumers to process the TemplateOptions ?

packages/ui5-info/src/commandRunner.ts
IainSAP: Lets create a new module later, will add TBI?

packages/ui5-info/src/types.ts
```diff
@@ -0,0 +1,79 @@
...
+     */
+    minSupportedUI5Version?: string;
+    /**
+     * Will use the cached versions from previous network calls if available otherwise will make network calls and populate cache
+     */
+    useCache?: boolean;
+    /**
+     * Some UI5 version services can return duplicates, set this option to make versions unique
+     */
+    removeDuplicateVersions?: boolean;
```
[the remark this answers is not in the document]
IainSAP: The idea is that you can return the raw version list.

packages/ui5-info/src/commandRunner.ts
IainSAP: Ill update so it only runs the specified command. I believe we don't use the http call for a reason internally.

packages/fiori-elements-writer/src/data/templateAttributes.ts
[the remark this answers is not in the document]
IainSAP: Will remove the dependency to `@sap-ux/fiori-elements-writer` and the associated `@sap-ux/ui5-info.getUI5Versions` option `fioriElementsVersion ` as there is already support for specifying the minimum ui5 version using `minSupportedUI5Version ` option instead. The consumer of the API will need to provide this, as discussed.

packages/ui5-info/package.json
IainSAP: Will remove this dependency based on other comment.

packages/ui5-info/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Usage
+
+```javascript
+import { getUI5Versions, type UI5VersionFilterOptions } from '@sap-ux/ui5-info';
+
+const filterOptions: UI5VersionFilterOptions = {
+    useCache: true,
+    removeDuplicateVersions: true,
+    groupUI5Versions: true
```
Klaus-Keller, line 24: These options do not exist on `UI5VersionFilterOptions` anymore

packages/ui5-info/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "files": [
+        "LICENSE",
+        "dist",
+        "templates",
+        "!dist/*.map",
+        "!dist/**/*.map"
+    ],
+    "dependencies": {
+        "@sap-ux/logger": "workspace:*",
+        "axios": "0.26.0",
```
Klaus-Keller, line 35: can we use version 1.4.0 to align with the rest of the modules?
```suggestion
        "axios": "1.4.0",
```

packages/ui5-info/package.json
```diff
@@ -0,0 +1,49 @@
...
+        "semver": "7.5.4",
+        "semver-compare": "1.0.0"
```
Klaus-Keller, line 37: Do we really need `semver-compare`. Looks like `semver` can also do comparison: https://www.npmjs.com/package/semver#comparison

packages/ui5-info/src/constants.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const UI5VersionRequestInfo = {
+    OfficialUrl: 'https://ui5.sap.com',
+    NeoAppFile: 'neo-app.json',
+    VersionsFile: 'version.json?sap-ui-config-patches=true&sap-ui-config-showall=true',
+    VersionsOverview: 'versionoverview.json',
+    VersionExternalFile: 'version.json'
+};
+
+export const DEFAULT_MIN_UI5_VERSION = '1.65.0';
```
Klaus-Keller, line 29: Can we align on one naming scheme for constants. Here we have 3 different ones: camelCase, PascalCase, UPPERCASE

Klaus-Keller (commented):
@IainSAP, thanks for the adjustments. I have some more comments, but looks already great.

packages/ui5-info/package.json
IainSAP: Unfortunately the behaviour of the compare functions from each module is different and will be some effort to swop. Adding the loose option and coercion using `semver` does not result in the same results.

packages/ui5-info/package.json
IainSAP: Done thx

packages/ui5-info/README.md
IainSAP: Done thx

Klaus-Keller (dismissed):
Thanks @IainSAP!
- text coverage is good
- tested local and could not find issues

Approved from my side.

devinea (approved):
Thanks for updates.
Code, tests and changeset looks good.

Klaus-Keller (approved):
Approved again, thanks for the removal of `server-compare`

PR #1401, opened 2023-10-24
Chore - Add madge to check for circular dependencies with a module code.

devinea:


Klaus-Keller (approved):
@devinea, thanks for adding detection of circular dependencies!
- Adding `madge` script to modules looks good
- Adding `madge` as devDependency and in `nx` looks good
- I leave the review for fixing existing circular dependencies to respective code owners

Approved from my side

PR #1402, opened 2023-10-26
fix: log validation summary regardless of validation result

slin-sap:
When validation result is successful, the successful log messages should still be printed on the console.


IainSAP (approved):
LGTM 

-  Tests updated as expected

longieirl (approved):
Looks good.

PR #1403, opened 2023-10-27
feat: move ux vocabulary module

nahmed22:
Move @sap/ux-odata-vocabularies module to the open-ux-tools repository in the public GitHub.


vinayhospete:
Please consider code smells and security hotspots

vinayhospete (approved):

packages/odata-vocabularies/types/csdl.d.ts
```diff
@@ -0,0 +1,625 @@
...
+         */
+        $EntitySetPath?: string;
+        $Parameter?: Parameter;
+        $ReturnType: ReturnType;
+    }
+}
+
+declare module '@sap-ux/vocabularies/CSDL' {
+    export = CSDL;
+}
```
Klaus-Keller, line 625: Why do we define this as a module with a name that looks like a package from us?

voicis: This is a copy of https://github.com/SAP/open-ux-odata/blob/main/packages/vocabularies-types/utils/csdl.d.ts with some minor modifications. In the future we'd like to check if this can somehow be reused between the packages.

Klaus-Keller: Ok, thanks for clarification

Klaus-Keller (commented):
As discussed, the code smells Remove this redundant type alias and replace its occurrences with "string" can be ignored, but with them we still have some. 
Can we do something about them?

voicis:
I think we already fixed the easy code smells. The remaining ones would take more time and we could have a look at them in a follow up PR as there are other tasks that depends on this code.

Klaus-Keller (approved):
- code looks good, except code smells that should be addressed as follow up
- code coverage is good
- did not test it locally

Approved from my side.

vinayhospete (approved):

Klaus-Keller (approved):
Approved again

nahmed22:
> Approved again

Thanks :-)

Issue #1404, opened 2023-10-30
FEATURE - Open existing controller from project files instead of creating a new one


nikmace:
_As a **developer**,
I want to **have new functionality for the adaptation projects**,
so that I **can reuse controller extension files for selected control**_

### Description

Visual Editor (for Adaptation Projects) can open the existing controller extension file (instead of creating a new one) when clicking 'Extend With Controller' from the context menu when right-clicking on a control.

### Technical Design
_What modules need to be created or enhanced?_ 
`@sap-ux/adp-tooling`,
`@sap-ux/preview-middleware`,
`@sap-ux-private/preview-middleware-client`

_How would the functionality be exposed and how can it be consumed?_
It can be installed through npm.
### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have an Adaptation Project
**when** I right-click on the context menu and choose 'Extend With Controller'
**then** ControllerExtension dialog will be opened with the text and a link: `Controller extension has already been created: Open file`. After clicking `Open file` the file will be opened in the VSCode application.

### Tasks
- [ ] An existing controller extension dialog is enhanced
- [ ] Existing controller extension file is opened in VSCode
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #1408, opened 2023-10-31
Open existing controller from project files instead of creating a new one

nikmace:
Feat for #1404.

- Includes changes to preview middleware client-side code.
- Existing Controller Extension dialog checks for existing controller extension file before opening and shows an appropriate message with a link to open that file locally in VSCode.


packages/adp-tooling/src/preview/adp-preview.ts
```diff
@@ -14,7 +14,8 @@ import type { AdpPreviewConfig, DescriptorVariant } from '../types';
 
 export const enum ApiRoutes {
     FRAGMENT = '/adp/api/fragment',
-    CONTROLLER = '/adp/api/controller'
+    CONTROLLER = '/adp/api/controller',
+    CODE_EXT = '/adp/api/code_ext/:controllerName'
```
zdravko-georgiev, line 18: Why not `/adp/api/code_ext/controllerName`?

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -170,6 +170,75 @@ export default class RoutesHandler {
...
+            const codeExtFiles = await this.readAllFilesByGlob('/**/changes/*_codeExt.change');
+
+            let controllerExists = false;
+            let controllerPath = '';
+            let controllerPathFromRoot = '';
+
+            const project = this.util.getProject();
+            const sourcePath = project.getSourcePath();
+            const rootPath = project.getRootPath();
+            const projectName = rootPath.split(/[\\/]/).slice(-1)[0];
```
zdravko-georgiev, line 193: Here you can use `project.getName()` 😉 See here https://sap.github.io/ui5-tooling/v3/api/@ui5_project_specifications_Project.html

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -170,6 +170,75 @@ export default class RoutesHandler {
...
+            for (const file of codeExtFiles) {
+                const fileStr = await file.getString();
+                const change = JSON.parse(fileStr) as {
+                    selector: {
+                        controllerName: string;
+                    };
+                    content: {
+                        codeRef: string;
+                    };
+                };
```
zdravko-georgiev, line 207: I would propose to extend https://github.com/SAP/open-ux-tools/blob/main/packages/adp-tooling/src/types.ts#L54 and use it here 😉

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -170,6 +170,75 @@ export default class RoutesHandler {
...
+            let controllerPath = '';
+            let controllerPathFromRoot = '';
+
+            const project = this.util.getProject();
+            const sourcePath = project.getSourcePath();
+            const rootPath = project.getRootPath();
+            const projectName = rootPath.split(/[\\/]/).slice(-1)[0];
+
+            const getPath = (projectPath: string, fileName: string) =>
+                path.join(projectPath, FolderNames.Changes, FolderNames.Coding, fileName).replace(/[\\/]/g, path.sep);
```
zdravko-georgiev, line 196: Why is the `replace` needed?

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -170,6 +170,75 @@ export default class RoutesHandler {
...
+            }
+
+            if (controllerExists && !fs.existsSync(controllerPath)) {
+                const errorMsg = `Controller extension file was not found at ${controllerPath}`;
+                res.status(HttpStatusCodes.NOT_FOUND).send({
+                    message: errorMsg
+                });
+                this.logger.debug(errorMsg);
+                return;
+            }
```
zdravko-georgiev, line 225: Can we rewrite this in
```
if (change.selector.controllerName === controllerName) {
    const fileName = change.content.codeRef.replace('coding/', '');
    controllerPath = getPath(sourcePath, fileName);
   if(fs.existSync) {
    // Read file
} else {
//Throw an error
}

}
```

packages/adp-tooling/src/preview/adp-preview.ts
nikmace: Because I am passing controllerName in the params and retrieving it like this:
image

And instead of using a POST request I am using GET request with a dynamic parameter. And if I use this `/adp/api/code_ext/controllerName` it means I have to send a body and use POST request, but the data in my case is just a string. And the dynamic parameter indicates that I am accessing a controller with a specific name.

packages/adp-tooling/src/preview/routes-handler.ts
nikmace: Because the path that is returned by path.join will look like this on windows: `C:\\Users\\projects\\adp.project`. And I am replacing those slashes with platform-specific path separator character for the current operating system, like:

Windows: `\` (backslash)
Unix-like systems (Linux, macOS): `/` (forward slash)

The path will be visible in the dialog, and this `C:\\Users\\projects\\adp.project` does not look good. :)

packages/adp-tooling/src/preview/routes-handler.ts
nikmace: This was not in the MiddlewareUtils interface in ui5.d.ts.
That is why I did not use it, thanks, will change this 👍🏻

packages/adp-tooling/src/preview/routes-handler.ts
nikmace: Fixed in: https://github.com/SAP/open-ux-tools/pull/1408/commits/43a23c11efe5bf1c3afb14575909c52931d8f0d1

packages/adp-tooling/src/preview/routes-handler.ts
nikmace: Fixed in: https://github.com/SAP/open-ux-tools/pull/1408/commits/43a23c11efe5bf1c3afb14575909c52931d8f0d1

packages/adp-tooling/src/preview/routes-handler.ts
nikmace: Changed in: https://github.com/SAP/open-ux-tools/pull/1408/commits/5ecb566fd4fb6f2a6d62ff1bceecce59d914acc7

packages/adp-tooling/src/preview/adp-preview.ts
zdravko-georgiev: Thanks, I overlooked that `:controllerName` is a dynamic parameter.

packages/adp-tooling/src/preview/routes-handler.ts
zdravko-georgiev: Thx

packages/adp-tooling/src/preview/routes-handler.ts
zdravko-georgiev: Thx

packages/adp-tooling/src/preview/routes-handler.ts
nikmace: Changed in: [1a92ecb](https://github.com/SAP/open-ux-tools/pull/1408/commits/1a92ecb1bdc70de9df0f9f07579f230c5f72f366)

zdravko-georgiev: Great, thanks!

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -170,6 +171,65 @@ export default class RoutesHandler {
...
+                if (change.selector.controllerName === controllerName) {
+                    const fileName = change.content.codeRef.replace('coding/', '');
+                    controllerPath = getPath(sourcePath, fileName);
+                    controllerPathFromRoot = getPath(projectName, fileName);
+                    controllerExists = true;
+
+                    if (!fs.existsSync(controllerPath)) {
+                        const errorMsg = `Controller extension file was not found at ${controllerPath}`;
+                        this.logger.debug(errorMsg);
+                        throw new Error(errorMsg);
```
zdravko-georgiev, line 211: Could you please replace `throw new Error(errorMsg)` with your previous implementation:
```
                res.status(HttpStatusCodes.NOT_FOUND).send({
                    message: errorMsg
                });
```
Throwing an error is too hard. Sorry for asking to change again.

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
```diff
@@ -120,28 +134,108 @@ export default class ControllerExtension extends BaseDialog {
...
+            source.setEnabled(false);
 
-        await this.createNewController(controllerName, viewId);
+            const controllerName = this.model.getProperty('/newControllerName');
+            const viewId = this.model.getProperty('/viewId');
+
+            await this.createNewController(controllerName, viewId);
+        } else {
+            const controllerPath = this.model.getProperty('/controllerPath');
+            window.open(`vscode://file/${controllerPath}`);
```
zdravko-georgiev, line 148: Please change to `vscode://file${controllerPath}`, otherwise opening in VSCode will not work, because of `//` after `file`.

zdravko-georgiev (changes requested):

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
nikmace: Opening in VSCode on Mac? Otherwise, on Windows, it works. And you mean those `//` should also be platform-specific?

zdravko-georgiev: On MAC the current implementation does NOT work. I needed to remove `/` after `file` to make it work. Does it work also without the `/` on Windows?

nikmace: I will test on both, since I still have both laptops :)

nikmace: Update: it does not work on Windows without slash `/`. For Windows the controllerPath is `C:\\Users` and for Mac it's`/Users/..`. So probably I need to add a slash when the OS is Windows, otherwise no slash.

zdravko-georgiev: @nikmace Can it be sorted out by just using `path.join`?

nikmace: We can use path.join on the server only. Just tested the last commit on Windows, doesn't work..

zdravko-georgiev: What URL is working for on Windows?

nikmace: For windows you need a slash after `file` like this: `"file/C:\\Users\\...\\open-ux-tools-playground\\apps\\adp.v2app\\webapp\\changes\\coding\\share.js"`. And for Mac you need to remove the slash after `file`

nikmace: Update 2: the last commit works ob both systems

packages/adp-tooling/src/preview/routes-handler.ts
nikmace: Changed in: [65646e7](https://github.com/SAP/open-ux-tools/pull/1408/commits/65646e72c170a4b0b2809ba12689bc4291c85c89)

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
zdravko-georgiev: @nikmace We can also use `/` on Windows. Hence a path like `vscode://file/C:/Users/git/open-ux-tools/package.json` will work. I just tested it on windows. So I would propose to convert the Windows paths to POSIX as done in here https://stackoverflow.com/a/63251716

nikmace: Again the same problem: For Windows the controllerPath is `C:/Users/..` and for Mac it's `/Users/..`. So for Windows I need to add a slash after file and for Mac no

IvoSG (approved):
The code looks good.
Coverage is good.
Manual testing ✅

packages/adp-tooling/src/preview/routes-handler.ts
zdravko-georgiev: Thanks

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
zdravko-georgiev: Thx, tested it works!

nikmace:
> @nikmace In case of an error. I do not see the full error message. I see only the status code error. Please make it so that the full error message is displayed. Furthermore what is the desired behaviour when a flex change is referencing a controller file that doesn't exist? Currently an error is thrown and one can no longer create a controller extension.
> 
> image

I think you should delete the respective change file (which points to a non-existent controller extension file) and create a new controller from scratch.

nikmace:
The existing functionality in this PR will be merged as it is and enhanced in #1460.

zdravko-georgiev (approved):
Code looks good and it is well structured
Review comments addressed
Very good test coverage
Manual testing ✅ 
Follow-up BLI created

IvoSG (approved):
The code looks good.
Coverage is good.
Manual testing ✅

PR #1415, opened 2023-11-01
fix: skip url validation if destination is present

slin-sap:
#1313

US #1313 aims to provide diagnosis message when deploy is running in test mode only. There might be rare cases when user mistakenly providing both url and destination when running deploy-test in CLI mode. This fix is to avoid showing incorrect diagnostic message "Invalid URL []" when deploy target is a destination. 

Some other minor fixes are also included in this PR as described in .changeset/lemon-walls-hope.md

### DIFF
--- a/packages/deploy-tooling/src/base/validate.ts
+++ b/packages/deploy-tooling/src/base/validate.ts
@@ -167,10 +167,11 @@ async function validateInputTextFormat(
 
 /**
  * A wrapper of validateUrl(). It uses same logic in system-access module's createAbapServiceProvider()
- * function to determine the deploy target. Only validate URL for on-prem ABAP deploy target.  
+ * function to determine the deploy target. Only validate URL for on-prem ABAP deploy target.
+ *
  * @param destination
- * @param url 
- * @returns 
+ * @param url
+ * @returns
  */
 function validateUrlForOnPremTargetOnly(destination: string, url: string): boolean | string {
     if (isAppStudio() && destination) {
--- a/packages/deploy-tooling/src/base/validate.ts
+++ b/packages/deploy-tooling/src/base/validate.ts
@@ -21,6 +21,7 @@ export type ValidationInputs = {
     transport: string;
     client: string;
     url: string;
+    destination: string;
 };
 
 export type ValidationOutput = {
@@ -76,7 +77,8 @@ export async function validateBeforeDeploy(
         package: config.app.package ?? '',
         transport: config.app.transport ?? '',
         client: config.target.client ?? '',
-        url: config.target.url ?? ''
+        url: config.target.url ?? '',
+        destination: config.target.destination ?? ''
     };
 
     // output is passed by reference and status updated during the internal pipeline below.
@@ -150,8 +152,10 @@ async function validateInputTextFormat(
     processInputValidationResult(result, output);
     result = validateClient(input.client);
     processInputValidationResult(result, output);
-    result = validateUrl(input.url);
-    processInputValidationResult(result, output);
+    if (!input.destination) {
+        result = validateUrl(input.url);
+        processInputValidationResult(result, output);
+    }
 
     // If all the text validation passed, only show the following success message.
     if (output.result) {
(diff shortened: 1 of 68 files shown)


packages/deploy-tooling/test/unit/base/validate.test.ts
```diff
@@ -4,6 +4,7 @@ import { mockedProvider, mockedAdtService } from '../../__mocks__';
 import { green, red, yellow } from 'chalk';
 import { TransportChecksService } from '@sap-ux/axios-extension';
 import { t } from '@sap-ux/project-input-validator/src/i18n';
+import exp from 'constants';
```
devinea, line 7:
```suggestion
import exp from 'constants';
```
Is this needed?

slin-sap: Removed

devinea (approved):
Thanks for the update. 
Changeset, code and tests updates look good.

packages/deploy-tooling/test/unit/base/validate.test.ts
```diff
@@ -76,6 +76,36 @@ describe('deploy-test validation', () => {
...
+                {
+                    app: { ...app, name },
+                    target: { ...target, destination, url: undefined }
+                },
+                mockedProvider as any,
+                nullLogger
+            );
+            expect(output.result).toBe(false);
+            const summaryStr = formatSummary(output.summary);
+            expect(summaryStr).toContain(`${red('×')} ${t('deploy.invalidAppNameMultipleReason')}`);
```
ullasholla, line 103: We're asserting for the error message already. Do we need to assert it's `red('x')` as well?

packages/deploy-tooling/src/base/validate.ts
```diff
@@ -150,8 +152,10 @@ async function validateInputTextFormat(
     processInputValidationResult(result, output);
     result = validateClient(input.client);
     processInputValidationResult(result, output);
-    result = validateUrl(input.url);
-    processInputValidationResult(result, output);
+    if (!input.destination) {
```
ullasholla, line 155: Do we have any code to sanitize the inputs? Something that will clear any URLs being passed in in this case. This code correctly skips the URL validation when a destination is available. However, there could be downstream code that could behave unpredictably if both a destination and (a potentially invalid) URL is being passed in.

ullasholla (changes requested):
@slin-sap, thanks for the PR.

Could you add a line or two about the problem this PR is fixing? The linked issue is big and it's not easy to find what exactly is being fixed and why it needs fixing.

Have a couple of questions as well. Could you please look at them when you can?

slin-sap:
> @slin-sap, thanks for the PR.
> 
> Could you add a line or two about the problem this PR is fixing? The linked issue is big and it's not easy to find what exactly is being fixed and why it needs fixing.
> 
> Have a couple of questions as well. Could you please look at them when you can?

Thanks @ullasholla  I added background info in the PR description. I think you are right. It would be more helpful to provide some insights. What I plan to do is to use [same logic here ](https://github.com/SAP/open-ux-tools/blob/main/packages/system-access/src/base/connect.ts#L136) to decide the deploy target from the input and isAppStudio() context. And report this insight to user. 

Does it make sense to extract the logic in [system-access connect](https://github.com/SAP/open-ux-tools/blob/main/packages/system-access/src/base/connect.ts#L136) to project-input-validator module? E.g.

`validateTarget(target: AbapTarget): 'url' | 'sbas' | undefined`

And reuse the logic to decide the deploy target environment in system-access, input validation in this US, and etc.... @devinea @longieirl @tobiasqueck

packages/deploy-tooling/src/base/validate.ts
slin-sap: This validate module is only called when deploy is run in test mode. It is more for providing some helpful validation/diagnostic messages in the console. 

For actual deploymenet, we have [similar logic](https://github.com/SAP/open-ux-tools/blob/290b6b598103d4614c3c215f9d982b04d3b463bd/packages/system-access/src/base/connect.ts#L136) to validate the deploy target and create the correct AbapProvider.

ullasholla (approved):
Thanks for addressing the review comments Shen.
The PR looks good.

Klaus-Keller (approved):
- Code looks good and easy to understand
- Test coverage is great
- Did not run a local test

Approved from my side

longieirl (approved):
Changes look good.
Havent tested locally.

Issue #1405, opened 2023-10-30
[Feature] Disable Control Property Editor for ADP projects with UI5 version lower than 1.71


GDamyanov:
As a **developer**,
I want to **disable the CPE for ADP projects with ui5 version lower than 1.71**,
so that I **cannot open CPE**

### Description

Control Property Editor should be disabled for ADP projects with UI5 version (set in the ui5.yaml in ui5-proxy-middleware) lower than 1.71 
### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a ADP project with UI5 version lower than 1.71 set in ui5.yaml in the ui5-proxy-middleware
**when** I start the Control Property Editor
**then** I should see error message with text "The current SAPUI5 version set for this Adaptation project is {some version}. The minimum version to use for SAPUI5 Adaptation Project and its SAPUI5 Visual Editor is 1.71" and a new browser tab with Control Property Editor should not be loaded.

### Tasks

- [ ] Unit Tests added
- [ ] Test Cases defined

tobiasqueck:
@GDamyanov what should happen in such case? Should an error be shown or ... ? Should the preview without CPE work?

GDamyanov:
> @GDamyanov what should happen in such case? Should an error be shown or ... ? Should the preview without CPE work?

Error message should be shown in the console and the CPE should not be loaded.

tobiasqueck:
> Error message should be shown in the console and the CPE should not be loaded.

Your implementation in #1423 is preventing also the preview (no editor). Is this also the goal or should the plain preview work?

GDamyanov:
> > Error message should be shown in the console and the CPE should not be loaded.
> 
> Your implementation in #1423 is preventing also the preview (no editor). Is this also the goal or should the plain preview work?

The preview should also be  stopped.

PR #1423, opened 2023-11-05
UI5 Version validation for Adaptation project

GDamyanov:
Feat for [#1405](https://github.com/SAP/open-ux-tools/issues/1405)

Warning message when the ui5 version of the adp project is less than 1.71


packages/preview-middleware/package.json
```diff
@@ -40,6 +40,7 @@
         "@sap-ux/logger": "workspace:*",
         "@sap-ux/adp-tooling": "workspace:*",
         "@sap-ux/control-property-editor-sources": "workspace:@sap-ux/control-property-editor@*",
+        "@sap-ux/ui5-config": "workspace:@sap-ux/ui5-config@*",
```
tobiasqueck, line 43: `"workspace:*"` is sufficient

tobiasqueck (changes requested):
The implementation only covers the case with the Fiori tools middleware being the one defining the UI5 version, however, the project has other ways to set the version e.g. a different proxy middleware (there are some) or by defining it in the yaml config as local version.
Therefore, I don't think you can find a working implementation that just looks at the config, if you want to cover all cases, then you need to move your check to the client-side, during the initialization because only then you know the effective UI5 version.

packages/preview-middleware-client/src/adp/init.ts
```diff
@@ -27,6 +29,15 @@ export default function (rta: RuntimeAuthoring) {
             }
         }
     });
+    
+    const ui5Version = (await VersionInfo.load()) as { version: string };
+    const ui5VersionValidationMessage = getUI5VersionValidationMessage(ui5Version.version);
+
+    if(ui5VersionValidationMessage) {
+        sendAction(showMessage('The current SAPUI5 version set for this Adaptation project is 1.70. The minimum version to use for SAPUI5 Adaptation Project and its SAPUI5 Visual Editor is 1.71'));
```
zdravko-georgiev, line 37: Please replace hardcoded test with the `ui5VersionValidationMessage` variable 😉

packages/ui5-proxy-middleware/src/ui5/middleware.ts
```diff
@@ -86,6 +87,12 @@ module.exports = async ({ resources, options }: MiddlewareParameters<UI5ProxyCon
         logger.debug(error);
     }
 
+    if (await isUI5VersionRemoved(ui5Version)) {
+        throw new Error(
+            `You are trying to load the application preview with the non existing version ${ui5Version} of SAPUI5. Please check and correct the version.`
+        );
+    }
+
```
zdravko-georgiev, line 95: Please don't do any changes to the `ui5-proxy-middleware`. Any specific UI5 version handling needs to be done in a package that integrates the `ui5-proxy-middleware`. OR in this case either in `adp-tooling` or in the `preview-middleware-client`.

zdravko-georgiev (changes requested):

packages/ui5-proxy-middleware/src/ui5/middleware.ts
GDamyanov: Removed as discussed.

packages/preview-middleware-client/src/adp/init.ts
GDamyanov: Fixed

packages/preview-middleware-client/src/adp/ui5-version-utils.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @param minVersion to check
+ *
+ * @returns boolean
+ */
+export function isLowerThanMinimalUi5Version(version: string, minVersion: string): boolean {
+    if (version !== undefined && version !== '' && minVersion !== undefined && minVersion !== '') {
+        const minVersionParsed = parseUI5Version(minVersion);
+        const ui5VersionParsed = parseUI5Version(version);
+        if (
+            !isNaN(ui5VersionParsed.major) &&
```
nikmace, line 14: For this big if statement could we split it more for readability? First, check if the version and minVersion are not empty. Second, check ui5VersionParsed.minor and ui5VersionParsed.major that they are valid numbers. Then check in the second if statement, which will be inside the first one, all the version stuff.

Something like this:

```typescript
if (version && minVersion) {
    const minVersionParsed = parseUI5Version(minVersion);
    const ui5VersionParsed = parseUI5Version(version);

    if (!isNaN(ui5VersionParsed.major) && !isNaN(ui5VersionParsed.minor)) {
        if (ui5VersionParsed.major < minVersionParsed.major) {
            return true;
        }
        if (ui5VersionParsed.major === minVersionParsed.major && ui5VersionParsed.minor < minVersionParsed.minor) {
            return true;
        }
    }
}
```

This also does not take into account the patch version (not sure if this is needed).

packages/preview-middleware-client/test/unit/adp/init.test.ts
```diff
@@ -42,12 +38,56 @@ describe('adp', () => {
         });
     });
 
+    beforeEach(() => {
+        rtaMock.getDefaultPlugins
```
nikmace, line 42: Why was this moved from the top, you mock the return value only once. You do not need to mock it again for each test.

packages/preview-middleware-client/test/unit/adp/init.test.ts
```diff
@@ -42,12 +38,56 @@ describe('adp', () => {
...
+            controlId: 'v2flex::sap.suite.ui.generic.template.ListReport.view.ListReport'
+        };
+
+        await callBackFn({
+            type: '[ext] add-extension-point',
+            payload
+        });
 
-        init(rtaMock);
+        expect(executeSpy).toHaveBeenCalledWith(payload.controlId, 'CTX_ADDXML_AT_EXTENSIONPOINT');
```
nikmace, line 81: This block of code is duplicated in two tests. It is testing the same thing two times, so maybe leave this block of code in the `init` test and remove it in the second test for the ui5 version notification.

packages/control-property-editor/src/slice.ts
```diff
@@ -219,6 +222,9 @@ const slice = createSlice<SliceState, SliceCaseReducers<SliceState>, string>({
                     state.changes.controls[key] = control;
                 }
             })
+            .addMatcher(showMessage.match, (state, action: ReturnType<typeof showMessage>): void => {
+                state.showDialogMessage = action.payload;
```
nikmace, line 226: It would be good to rename this variable to dialogMessage. And also in the initialState.

So here when you match showMessage you set the state like this: 
```typescript
state.dialogMessage = action payload;
```

Also in `App.tsx` when retrieving from the store change showDialogMessage to dialogMessage.

nikmace (changes requested):

packages/preview-middleware-client/test/unit/adp/init.test.ts
GDamyanov: This is done like that because for each call of init we should have these two specific mock return of getDefaultPlugins

packages/control-property-editor/src/App.tsx
```diff
@@ -84,28 +86,42 @@ export default function App(appProps: AppProps): ReactElement {
...
             </section>
             <section className="app-panel app-panel-right">
                 <PropertiesPanel />
             </section>
+            <UIDialog
+                hidden={!showDialogMessage}
+                dialogContentProps={{
+                    title: t('TOOL_DISCLAIMER_TITLE'),
+                    subText: dialogMessage
+                }}></UIDialog>
```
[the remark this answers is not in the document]
GDamyanov: Done

packages/preview-middleware-client/test/unit/adp/init.test.ts
```diff
@@ -42,12 +38,56 @@ describe('adp', () => {
...
+                }
+            })
+            .mockReturnValueOnce({});
+    });
+
+    afterEach(() => {
+        initOutlineSpy.mockClear();
+        addMenuItemSpy.mockClear();
+        setPluginsSpy.mockClear();
+        sendActionMock.mockClear();
```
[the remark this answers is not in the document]
GDamyanov: Done

packages/preview-middleware-client/src/adp/init.ts
```diff
@@ -27,6 +29,15 @@ export default function (rta: RuntimeAuthoring) {
             }
         }
     });
+    
+    const ui5Version = (await VersionInfo.load()) as { version: string };
+    const ui5VersionValidationMessage = getUI5VersionValidationMessage(ui5Version.version);
+
+    if(ui5VersionValidationMessage) {
```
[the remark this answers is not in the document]
GDamyanov: Accepted

packages/preview-middleware-client/src/adp/ui5-version-utils.ts
GDamyanov: Done

packages/preview-middleware-client/test/unit/adp/init.test.ts
GDamyanov: Done

packages/control-property-editor/src/slice.ts
GDamyanov: Done

nikmace (approved):
Code looks good
Good test coverage
Manual testing ✅

.changeset/lazy-roses-drive.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ux/preview-middleware': patch
+'@sap-ux/ui5-config': patch
+---
+
+Validate ui5 version defined in ui5-proxy-middleware for adp project
```
zdravko-georgiev, line 6: Please fix the description. This is not true. The goal is to validate the UI5 version, regardless from where it is coming.

.changeset/lazy-roses-drive.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ux/preview-middleware': patch
+'@sap-ux/ui5-config': patch
```
zdravko-georgiev, line 3: We do not need to change `ui5-config`. Revert any any changes to `ui5-config` and remove this please.

packages/ui5-config/src/ui5config.ts
zdravko-georgiev, line 1: Revert this change please

packages/ui5-proxy-middleware/tsconfig.json
zdravko-georgiev, line 1: revert this change please

zdravko-georgiev (changes requested):

zdravko-georgiev (changes requested):

packages/preview-middleware/package.json
zdravko-georgiev: Removed

packages/control-property-editor/src/App.tsx
```diff
@@ -27,6 +28,12 @@ export interface AppProps {
  */
 export default function App(appProps: AppProps): ReactElement {
     const { previewUrl } = appProps;
+    const urlWithoutHash = previewUrl.split('#')[0];
+    const queryStr = urlWithoutHash.split('?')[1];
+    const urlParams = new URLSearchParams(queryStr);
+    const scenario = urlParams.get('sap-ui-scenario') as Scenario;
+    const isAdpProject = scenario === 'ADAPTATION_PROJECT';
```
zdravko-georgiev, line 35: Please use the `flexSettings` for reading the scenario https://github.com/SAP/open-ux-tools/blob/063ef51d1e2f0d906f4f295bc7776c32ab197a08/packages/preview-middleware-client/src/cpe/outline/index.ts#L17

zdravko-georgiev (changes requested):

packages/control-property-editor/src/App.tsx
GDamyanov: This is react app and we don't have access to rta.

zdravko-georgiev: @GDamyanov There is a communication channel between the react app and the preview where RTA is running. Please read the scenario from the RTA and then send it to the react app.

nikmace: > @GDamyanov There is a communication channel between the react app and the preview where RTA is running. Please read the scenario from the RTA and then send it to the react app.

This way there is a huge delay for the post messaging between the iframe and react application (different loading times). This is really visible for the user. The most efficient way is to pass scenario in `previewUrl` to the `start` method, which is the entry point of the react application. This way when the react application is loading we already know the scenario and we do not have a huge delay.

zdravko-georgiev: @nikmace Good point. If this is the case, then please add the `scenario` to the start options here https://github.com/SAP/open-ux-tools/blob/main/packages/control-property-editor/src/index.tsx#L28

zdravko-georgiev: @nikmace @GDamyanov It is possible to read the `scenario` from the store. It is already there. Please see here https://github.com/SAP/open-ux-tools/blob/e9f1ea561af740d92b54f43bb13114be6b634d6e/packages/control-property-editor/src/index.tsx#L28

nikmace: > @nikmace @GDamyanov It is possible to read the `scenario` from the store. It is already there. Please see here
> 
> https://github.com/SAP/open-ux-tools/blob/a9de3d97f3678e8d54fe5525530a84e382b3e273/packages/control-property-editor/src/App.tsx#L42

Where is this scenario coming from? From post-messaging? If yes, that is exactly what I was talking about we will now be getting a big delay before we actually know in what scenario we are. In App.tsx you should instead pass the scenario to the URL and dispatch it in App.tsx. And then you can get it from the store by either using a useSelector or directly using the variable which you retrieved from previewUrl.

The result of the last commit with reverted changes will be is that first the react application will be loaded, second, the iframe will start being rendered (and we will see white container), then we receive a post message from it with the scenario and the redux store updates and we stop showing the iframe and show a dialog message.  The time between the react application load time and the iframe is noticeable not to say the least.

zdravko-georgiev: @nikmace When I start the application the time between loading the react application and then the iframe is good. There is no significant delay.

packages/control-property-editor/src/App.tsx
```diff
@@ -87,29 +89,45 @@ export default function App(appProps: AppProps): ReactElement {
         setWarningDialogVisibility(false);
     }
 
+    useEffect(() => {
+        if (dialogMessage && isAdpProject) {
+            setShouldShowDialogMessageForAdpProjects(true);
+        }
+    }, [dialogMessage]);
```
zdravko-georgiev, line 96: This is not working as expected. It is executed before actually the scenario is set to `ADAPTATION_PROJECT`. You can see it best when debugging. Please have a look.

zdravko-georgiev (changes requested):

zdravko-georgiev (approved):
Code looks good
Patchset is OK
Review comments addressed
Very good test coverage
Manual testing ✅ 

Considering that the only available version on UI5 CDN that is below `1.71` is `1.38`, users should not get to see this pop-up too often.

nikmace (approved):
Code looks good
Review comments addressed
Good test coverage
Did not test manually

nikmace:
Can someone from @SAP/ux-tools-control-property-editor do a review?

voicis (approved):
- CPE code looks good
- Good test coverage
- Manually tested CPE on FE projects.

PR #1428, opened 2023-11-06
fix: package capital case

slin-sap:
#1313

Move code for converting caplitaizing $tmp to $TMP. The previous version had this conversion locally in the function for ADT validation of package exists. This is also required to check if package is local package when running ADT query for transport requests.


packages/deploy-tooling/src/base/validate.ts
```diff
@@ -82,6 +82,10 @@ export async function validateBeforeDeploy(
         destination: config.target.destination ?? ''
     };
 
+    if (input.package.toUpperCase() === '$TMP') {
```
tobiasqueck, line 85: Why just `$TMP`, don't we need to capitalize any package name before communicating with the backend?

longieirl: During testing we noticed that the transport request flow only worked when the package name was in upper case! Unable to find any documentation around this but if submitted with lower case, the TR fails.

slin-sap: @tobiasqueck  The conversion here is for validation only. The user input for deploy remain low cases. My thought is to treat $tmp as exception since its more common used. I feel this conversion is based on the implementation details that lower case can be accepted for deploy, but ADT requires uppercase. This might change or our understanding might not be accurate. So I prefer ADT to validate on the raw user input and we don't perform this conversion behind the scene.

How about I can add a hint in the validation message to tell user to try uppercase, if I detect package name contains lower case and the package name was rejected by ADT?

tobiasqueck: @slin-sap I am not convinced that an exception is a good idea. I think it would be better to be consistent, i.e. we either allow lower case input and ensure that we always capitalize it before sending it to ADT (for deploy or any other use case using dev objects) or, we leave it as-as.
I'd prefer capitalizing before sending.
Btw, shouldn't we have similar issues when using lower case letters for transport requests?

slin-sap: > @slin-sap I am not convinced that an exception is a good idea. I think it would be better to be consistent, i.e. we either allow lower case input and ensure that we always capitalize it before sending it to ADT (for deploy or any other use case using dev objects) or, we leave it as-as. I'd prefer capitalizing before sending. Btw, shouldn't we have similar issues when using lower case letters for transport requests?

Thanks. I will apply uppercase conversion to both package name and transport req number. I will also add an extra log message indicating if uppercase conversion was applied for ADT validation

tobiasqueck (approved):
* changeset ok
* first review changes applied
* changes properly covered with tests

longieirl (approved):
Latest changes look good.

Issue #1438, opened 2023-11-08
BUG - Undeploy fails if app contains namespace


slin-sap:
Undeploying an application with a namespace (e.g. /SP2IT/ZROOMV2FSV1) fails with a 400 error, Invalid URI segment. 

### Description
A clear description of what the bug is.

### Steps to Reproduce

Generate an app with a repository name including a namespace, e.g. /SP2IT/ZROOMV2FSV1.
Deploy the app to, e.g. X02
Try to undeploy the app.

### Expected results

The app should undeploy successfully.

### Actual results

The app fails to undeploy, with error:

> fiori undeploy --config ui5-deploy.yaml

Confirmation is required to undeploy the app:

    Application Name: /SP2IT/ZROOMV2FSV1
    Transport Request: X02K900018
    Destination: abap-cloud-x02-abapcp-integrationtests-staging-provisioning-tests
    

✔ Start undeployment (Y/n)?

```
 … yes
info undeploy-abap Starting undeployment.
2023-11-06 10:17:54 info    abap-deploy-task: Starting to undeploy. 
2023-11-06 10:17:56 warn    abap-deploy-task: Warning: retry undeploy to handle a backend rejection... 
2023-11-06 10:17:56 error   abap-deploy-task: Request failed with status code 400 
2023-11-06 10:17:56 error   abap-deploy-task: Invalid URI segment 
2023-11-06 10:17:56 error   abap-deploy-task: SAP_Transaction: For backend administrators: use ADT feed reader "SAP Gateway Error Log" or run transaction /IWFND/ERROR_LOG on SAP Gateway hub system and search for entries with the timestamp above for more details 
2023-11-06 10:17:56 error   abap-deploy-task: SAP_Note: See SAP Note 1797736 for error analysis (https://service.sap.com/sap/support/notes/1797736) 
2023-11-06 10:17:56 error   abap-deploy-task: Undeployment has failed. 
2023-11-06 10:17:56 error   abap-deploy-task: Change logging level to debug your issue
        (see examples https://github.com/SAP/open-ux-tools/tree/main/packages/deploy-tooling#configuration-with-logging-enabled) 
```

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
URI encoding is missing for undeployment

### Fix
Added URI encoding

### Why was it missed
[...]

PR #1439, opened 2023-11-08
fix: encode appname with namespace

slin-sap:
#1438

Added uri encoding for appname that contains namespace in undeploy request url


longieirl (approved):
Changes look good and tests to cover change.

cianmSAP (approved):
Updates look good
Changeset ✅
Did not test

Issue #1441, opened 2023-11-09
BUG - CPE UI outline is empty for some applications until a change is created


IvoSG:
### Description
When the application is started the CPE outline pane is empty until a change is created

### Steps to Reproduce
1. Go to an adaptation project
2. Start the editor with `npm run editor`
3. The application is started but the outline pane left of CPE is empty
4. Make some changes and the outline pane will be populated

### Expected results
When the application is started the CPE outline pane should be populated 

### Actual results
When the application is started the CPE outline pane is empty

### Screenshots
before change:
![image](https://github.com/SAP/open-ux-tools/assets/52562399/2248eab6-2ef1-4fe3-b3e5-8c880e6b4480)
after change:
![image](https://github.com/SAP/open-ux-tools/assets/52562399/4e9ade0a-b740-4708-a4c2-d329bbbbfab0)

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1442, opened 2023-11-09
FIX - CPE UI outline is empty for some applications until a change is created

IvoSG:
Fix for [1441.](https://github.com/SAP/open-ux-tools/issues/1441)

-  call syncOutline in the initOutline because the attached event is not triggered for every case
(if the app is already loaded before the event is attached)


GDamyanov (changes requested):
Please add changeset

GDamyanov (dismissed):
Change is small and easy to understand
Manual testing ✅

.changeset/fifty-sloths-march.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ux-private/preview-middleware-client': patch
+'@sap-ux/preview-middleware': patch
+---
+
+call syncOutline in the initOutline because the attached event is not triggered for every case(if the app is already loaded before the event is attac
```
zdravko-georgiev, line 6: Please change to "fix outline initialisation for the case when application is loaded, but outline is empty"

zdravko-georgiev (changes requested):

.changeset/fifty-sloths-march.md
IvoSG: Done

zdravko-georgiev (approved):
Simple and straightforward change
Changeset is OK
Excellent test coverage
Manual testing ✅

nikmace (approved):
The code looks good.
Coverage is good.
Tested locally.

PR #1455, opened 2023-11-10
fix(ui-components): update `header` and `layout` icons

johannes-kolbe:
update `header` and `layout` icons


hitesh-parmar (approved):
updates to icons looks good

PR #1462, opened 2023-11-15
feat(adp): provide possibility to enter credentials via yaml config or .env

zdravko-georgiev:
This PR enables the user to provide credentials via `yaml` config or `.env` file when previewing and editing adaptation projects.

### DIFF
--- a/packages/adp-tooling/src/preview/adp-preview.ts
+++ b/packages/adp-tooling/src/preview/adp-preview.ts
@@ -84,6 +84,9 @@ export class AdpPreview {
                 username: process.env.FIORI_TOOLS_USER,
                 password: process.env.FIORI_TOOLS_PASSWORD
             };
+        } else if (config.auth && !process.env.FIORI_TOOLS_USER && !process.env.FIORI_TOOLS_PASSWORD) {
+            process.env.FIORI_TOOLS_USER = config.auth.username;
+            process.env.FIORI_TOOLS_PASSWORD = config.auth.password;
         }
     }
 
--- a/packages/adp-tooling/src/preview/adp-preview.ts
+++ b/packages/adp-tooling/src/preview/adp-preview.ts
@@ -1,3 +1,4 @@
+import dotenv from 'dotenv';
 import express from 'express';
 import ZipFile from 'adm-zip';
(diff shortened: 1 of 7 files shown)


packages/preview-middleware/README.md
```diff
@@ -39,6 +40,12 @@ Array of additional application configurations:
 | `client`      | `string` optional              | sap-client parameter                                                                                                                            |
 | `scp`         | `boolean` optional             | If set to true the proxy will execute the required OAuth routine for the ABAP environment on SAP BTP                                            |
 
+### `adp.auth`
```
tobiasqueck, line 43: I am against allowing to set user/password in the yaml. It is a file that is committed to git and should not contain such sensitive information.

packages/adp-tooling/src/preview/adp-preview.ts
```diff
@@ -75,7 +76,18 @@ export class AdpPreview {
...
     ) {
+        dotenv.config();
         this.routesHandler = new RoutesHandler(project, util, logger);
+
+        if (!config.auth && process.env.FIORI_TOOLS_USER && process.env.FIORI_TOOLS_PASSWORD) {
+            config.auth = {
+                username: process.env.FIORI_TOOLS_USER,
+                password: process.env.FIORI_TOOLS_PASSWORD
+            };
+        } else if (config.auth && !process.env.FIORI_TOOLS_USER && !process.env.FIORI_TOOLS_PASSWORD) {
```
tobiasqueck, line 87: This can be removed if we don't support providing the variables as part of the yaml.
Btw, if the user/password are missing, then it will already be prompted and stored in the variables (https://github.com/SAP/open-ux-tools/blob/main/packages/system-access/src/base/connect.ts#L105-L108)

tobiasqueck (changes requested):
No setting of secrets in the yaml, please remove that part. The rest looks good.

nikmace (changes requested):

tobiasqueck:
@zdravko-georgiev work on this one was discontinued, correct? Can we close it?

zdravko-georgiev:
Yes, we can close it.

Issue #1470, opened 2023-11-16
BUG - ui-components. UIFlexibleTable. Issue/conflict with touch events. Allow to choose between touch drag or touch scroll(allow to disable touch drag)


815are:
### Steps to Reproduce
1. If UIFlexibleTable rendered with scroll and drag enabled;
2. User tries to use touch to scroll;

Problem - drag is enabled.

Solution:
Allow to choose if `touch drag` should be disabled or enabled. If we disable touch drag, then touch should trigger scroll.

### Expected results
Allow to choose if `touch drag` should be disabled or enabled. If we disable touch drag, then touch should trigger scroll.

### Actual results
There no option to disable drag with touch events

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

PR #1471, opened 2023-11-16
fix(ui-components): allow to disable touch drag

815are:
Issue #1470

Change for `UIFlexibleTable`:
1. new property 'isTouchDragDisabled';
2. if property `isTouchDragDisabled` is true and table row is draggable - then drag should not be initialized if touch event were used.

Such setting is required to allow disable touch drag - in some cases container of tableflex is scrollable and if touch drag is enabled, then user is not allowed to scroll, because drag was triggered


vadson71 (approved):
Code changes look good to me. Tested them in storybook, - worked fine. Excellent code coverage. Approved

devinea (approved):
Changes look good.
I haven't tested locally

PR #1472, opened 2023-11-16
fix(project-access): load different cds versions

Klaus-Keller:
## Issue

Error:
```
TypeError: Cannot read properties of undefined (reading 'odata-v4')
```

### Description
When loading different versions of `@sap/cds` dynamically, this error can occur. Affected are versions >= `@sap/cds@7.3`. It seems `global.cds`, which might remains in global node environment space, causes this issue. Assigning `global.cds` to newly loaded cds versions fixes the issue.

### DIFF
--- a/packages/project-access/src/project/cap.ts
+++ b/packages/project-access/src/project/cap.ts
@@ -332,6 +332,13 @@ export async function getCapEnvironment(capProjectPath: string): Promise<CdsEnvi
 }
 
 /**
+ * To fix issues when switching different cds versions dynamically, we need to set global.cds, see end of function loadCdsModuleFromProject()
+ */
+declare const global: {
+    cds: CdsFacade;
+};
+
+/**
  * Load CAP CDS module. First attempt loads @sap/cds for a project based on its root.
  * Second attempt loads @sap/cds from global installed @sap/cds-dk.
  * Throws error if module could not be loaded.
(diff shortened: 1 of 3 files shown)


packages/project-access/src/project/cap.ts
```diff
@@ -362,7 +369,12 @@ async function loadCdsModuleFromProject(capProjectPath: string): Promise<CdsFaca
             `Could not load cds module. Attempt to load module @sap/cds from project threw error '${loadProjectError}', attempt to load module @sap/cds from @sap/cds-dk threw error '${loadError}'`
         );
     }
-    return 'default' in module ? module.default : module;
+    const cds = 'default' in module ? module.default : module;
+    // Fix when switching cds versions dynamically
+    if (global) {
+        global.cds = cds;
```
voicis, line 375: Switching global variables doesn't seem to fix all the cases, since someone else could load a different version in async code and I suspect we'll end up with having the same issue.

Klaus-Keller: @voicis, good point, but how else can we do it? 
First idea was to add `global.cds = cds` in CAP-related functions in [packages/project-access/src/project/cap.ts](https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/project/cap.ts), like `getCapModelAndServices()`, `getCapEnvironment()`, right before we do a call on cds instance. However, all those functions call `loadCdsModuleFromProject()` to get a cds instance, so `global.cds = cds` is done every time before we call functions on cds. Any other idea?

voicis: I looked at the relevant CDS coding and there is nothing else we can do in this case.

Since loading CDS module is not exposed and we do not keep the CDS module reference, but always load it this fix should be sufficient.

voicis (approved):
- Change set looks good
- Code changes are easy to understand
- Code coverage is good
- Tested locally with CDS 6.8.4 and 7.3.0

zdravko-georgiev (approved):
Excellent test coverage
Did not tested manually

devinea (approved):
changeset ✅ 
Code and tests look good.

Issue #1473, opened 2023-11-17
BUG -  UICallout. Focus next/previous element relative to target when `Tab` key is pressed using keyboard


815are:
### Steps to Reproduce
1. Callout example like:
```
const CalloutWithItems = (props: { id: string }): JSX.Element => {
    const { id } = props;
    const [isCalloutVisible, setCalloutVisible] = useState(false);
    const toggleCallout = (): void => {
        setCalloutVisible(!isCalloutVisible);
    };
    return (
        <div>
            <UIDefaultButton
                id={id}
                iconProps={{ iconName: UiIcons.Bulb }}
                checked={isCalloutVisible}
                onClick={toggleCallout}>
                Open
            </UIDefaultButton>
            {isCalloutVisible && (
                <UICallout
                    target={`#${id}`}
                    isBeakVisible={true}
                    setInitialFocus={true}
                    beakWidth={5}
                    directionalHint={4}
                    onDismiss={() => toggleCallout()}
                    calloutWidth={300}
                    calloutMinWidth={300}
                    contentPadding={UICalloutContentPadding.Standard}
                    focusTargetSiblingOnTab={true}>
                    <div>
                        <UIFocusZone>
                            <UIDefaultButton onClick={toggleCallout}>Option 1</UIDefaultButton>
                            <UIDefaultButton onClick={toggleCallout}>Option 2</UIDefaultButton>
                            <UIDefaultButton onClick={toggleCallout}>Option 3</UIDefaultButton>
                        </UIFocusZone>
                    </div>
                </UICallout>
            )}
        </div>
    );
};
```
2. Render tree components like:
```
<div>
            <CalloutWithItems id="callout-with-items-1-id" />
            <CalloutWithItems id="callout-with-items-2-id" />
            <CalloutWithItems id="callout-with-items-3-id" />
        </div>
```
3. Use keyboard and `Tab` and navigate to 2nd button;
4. Open callout - focus is set to inner button in callout;
5. Press `Tab`;

Problem - focus moves outside webview, because callout is rendered as last element in body.

### Expected results
Allow to configure callout with additional property, which allow to focus next element based on target. In such scenario it is 3rd button.
If `Tab + shift` pressed then 1st button shouldbe focused;

### Actual results
Focus moves outside webview, because callout is rendered as last element in body.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

PR #1474, opened 2023-11-17
fix(ui-components): UICallout. Focus next/previous element relative to target when `Tab` key is pressed using keyboard

815are:
#1473

Problem happens with Callout, when we focus any element in callout and use `Tab` key, then focus moves out of webview, because Callout is rendered as last element in `body`.

Changes:
`UICallout` component - new property `focusTargetSiblingOnTabPress`.

If `focusTargetSiblingOnTabPress` is passed and user pressed `Tab`(or `Tab + shift`) while focus is in Callut, then callout will calculate next/previous focusable element based on passed `target` and will set to calculated element.
For calculation of next/previous are used methods from fluent-ui - `{getNextElement, getPreviousElement} from '@fluentui/react';`

### DIFF
--- a/.changeset/forty-bottles-sparkle.md
+++ b/.changeset/forty-bottles-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui-components': patch
+---
+
+fix: UICallout. New property "focusTargetSiblingOnTabPress", which allows to focus next/previous element relative to target when Tab key is pressed using keyboard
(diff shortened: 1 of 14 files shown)


Jimmy-Joseph19 (approved):
Approving code changes 
Very good test coverage
Tested using storybook ✅

Klaus-Keller (approved):
Thanks @815are 
- code changes look good
- test coverage is good
- did not test it, just visual code review

Approved from my side.

Issue #1433, opened 2023-11-07
BUG - Some issues with FE and FE test writers


devinea:
### Related Feature
Feature request: tools-suite 24231 & S4 Extension 176

### Description
- fe-fpm-writer - Remove non-localised title in view.xml templates. Alternative would be to localise the text, but it is not visible to the user so does not add any value.
- OPA tests fail for worklist as filterbar and go button is hidden by default. Remove iExecuteSearch in V4 OPA test if filter bar is hidden.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

devinea:
Re-opening to localize xml view title in FPM

Issue #1365, opened 2023-10-10
FEATURE - i18n module


voicis:
### Description

Tools need to be able to read and write i18n entries for apps. There are various flavours of i18n files and we should have common way for handling all of them, that can be easily reused by tools. 

UI5 resource modules (properties files)
UI5 manifest.json translation entries (properties files)
CAP CDS services (properties, CSV and JSON files)

### Technical Design
_What modules need to be created or enhanced?_ 
New module `@sap-ux/i18n`
_How would the functionality be exposed and how can it be consumed?_

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a ...
**when** I execute ...
**then** ... will happen.

### Tasks
- [ ] Parsers
  - [ ] properties
  - [ ] csv
  - [ ] json 
- [ ] CDS support
- [ ] UI5 support
- [ ] Consider mem-fs support?
- [ ] Use common function to get i18n properties in https://github.com/SAP/open-ux-tools/pull/1475/files#diff-f2599904e616ac2446bef03eb16046f806b7c637da5b005a856be5abf0228a9aR97
- [ ] Unit Tests added
- [ ] Test Cases defined

voicis:
Added in #1655

PR #1475, opened 2023-11-17
fix fpm xml view localisation

devinea:
https://github.com/SAP/open-ux-tools/issues/1433


packages/fe-fpm-writer/src/page/custom.ts
```diff
@@ -91,6 +91,14 @@ export function generate(basePath: string, data: CustomPage, fs?: Editor): Edito
     if (!fs.exists(viewPath)) {
         fs.copyTpl(join(root, 'ext/View.xml'), viewPath, config);
     }
+    // i18n.properties
+    const i18nPath = join(basePath, 'webapp', 'i18n', 'i18n.properties');
+    const i18TemplatePath = join(root, 'i18n', 'i18n.properties');
+    if (fs.exists(i18nPath)) {
```
815are, line 97: @devinea 

One concern about hardcoded path to `i18n` file,
there can be scenario where we have `manifest.json` like:
```
"sap.ui5": {
        "models": {
            "i18n": {
                "type": "sap.ui.model.resource.ResourceModel",
                "uri": "custom-i18n2/i18n.properties"
            },
```

Then if there is such manifest and I create custom page using fpm-writer - in result new `i18n` entry or new file is created within `i18n/i18n.properties`. Maybe we could try detect path based on `manifest.json`?
I think custom i18n paths are rare scenario, not fully sure how critical it is

815are (commented):
@devinea 

I tested with app by triggering creation of custom pages.
I noticed following issues/questions:
1. Each time same entry `title` is created in `i18n.properties`
image

I think we should not create duplicate entry in `i18n` file.

2. Also if I create to continues custom pages - then there will be rendered same title for both custom page. Is it expected?

3.  Please check comment about hardcoded path to `i18n` file, there can be cases where `i18n` has different path configured in `manifest.json`,ut not sure how critical is that scenario

packages/fe-fpm-writer/src/page/custom.ts
devinea: @815are Thanks. Addressed in https://github.com/SAP/open-ux-tools/pull/1475/commits/0991f7d897b654b19de30916f8c99395b7347cfa

devinea:
> 1. Each time same entry `title` is created in `i18n.properties`

@815are Good point. Fixed in  https://github.com/SAP/open-ux-tools/pull/1475/commits/047e78f388e878913e4fedb82ce48b5c2ff901b9

IainSAP (approved):
Thx @devinea 

- Approving with the caveat that we should be using general API for updating i18n as there are cases that wont be covered.

-  Tests updates as expected

815are (commented):
@devinea 

New scenario while testing manually:
1. There is option to create custom page using existing view. In such case unnecessary/unused(or duplicate) i18n entry is created.
I think you can use test -> https://github.com/SAP/open-ux-tools/blob/fix_fpm_xml_view_localization/packages/fe-fpm-writer/test/unit/page/custom.test.ts#L135
I think we need create new i18n entry only when new view.xml is generated/created

devinea:
> 1. There is option to create custom page using existing view. In such case unnecessary/unused(or duplicate) i18n entry is created.

Thanks @815are Fixed in https://github.com/SAP/open-ux-tools/pull/1475/commits/31eb253db1d253a16e152626144b9510747a75db to only write i18nProperties when writing the view also.

815are (approved):
Checked `fe-fpm-writer` part:
1. manual test - reported issues are fixed;
2. Small concern about edge cases with `i18n` path(other variations on how i18n path resolved), but perhaps can be improved/reused after https://github.com/SAP/open-ux-tools/issues/1365

IainSAP (approved):
Re-approving as already reviewed

devinea:
Thanks @815are @IainSAP . I've made a note in https://github.com/SAP/open-ux-tools/issues/1365 to update this code with the common i18n function when available in open-ux-tools.

Issue #1482, opened 2023-11-21
BUG - ui-components. UIContextualMenu. Border bottom cut for UIContextualMenu in certain zoom level


815are:
### Steps to Reproduce
1. Apply zoom level `1.1`;
2. Open contextual menu from storybook 
![image](https://github.com/SAP/open-ux-tools/assets/90789422/18216b7e-e859-4059-aa68-df03dbb3c0a8)

Bottom border of contextual menu invisible

### Expected results
Bottom border should be visible on any zoom level

### Actual results
Bottom border is invisible on some zoom level

### Screenshots

![image](https://github.com/SAP/open-ux-tools/assets/90789422/18216b7e-e859-4059-aa68-df03dbb3c0a8)

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

PR #1483, opened 2023-11-21
fix: ui-components. UIContextualMenu. Border bottom cut for UIContextualMenu in certain zoom level

815are:
Issue #1482
PR fixes issue:
Before fix:
image

After fix:
image

Main issue happens because context menu has animation:
image
image

Animation result introduces issues with border(such borders are not present in default `fluent-ui` visualisation for context menu, where box-show is used)
Removing previous workaround with `translateZ(0) rotate(0.0001deg)` and applying more stable approach with box shadow(which works more smoother during animation and does not introduce reported cut effect)


vadson71 (approved):
Changes look good. Tested in storybook - works as expected. Approved

Jimmy-Joseph19 (approved):
Code Changes Looks Good
Change-set found
Covered with test
Approving

I have not tested locally

Jimmy-Joseph19 (approved):
Code Changes Looks Good
Change-set found
Covered with test
Approving

I have not tested locally

Issue #1468, opened 2023-11-16
BUG - Comp, control variant changes are not updating in the pending changes tab.


GDamyanov:
### Description
Comp, control variant changes are not updating in the pending changes tab

### Steps to Reproduce
1. Create comp or control variant adaptation project
2. Create some comp, control variant change
3. Open browser console

### Screenshots

![Screenshot 2023-11-16 100837](https://github.com/SAP/open-ux-tools/assets/27560551/3d2ecd22-cf73-4d06-928e-8b1d322b637d)

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1484, opened 2023-11-21
fix: comp/control variant changes update in pending changes tab

GDamyanov:
Fix for: #1468

Changes related to comp/control variants especially for creating new variants are updating in pending changes tab. These changes related to variant updates cannot be handled from the commands which are creating changes, because they are more specific.


nikmace:
> Fix for: #1468

Please add brief description, what changes this fix introduces

packages/preview-middleware-client/src/cpe/changes/service.ts
```diff
@@ -307,4 +322,44 @@ export class ChangeService {
 
         return result;
     }
+
+    private getCommandChangeType(command: FlexCommand): string | undefined {
```
nikmace, line 326: Here we have a duplication with the try-catch, I am suggesting making those operations more generic:

```typescript
private attemptOperations<T>(operations: Array<() => T>): T | undefined {
    for (const operation of operations) {
        try {
            return operation();
        } catch (error) {
            // Log error
            // Continue to the next operation
        }
    }
    return undefined;
}

private getCommandChangeType(command: FlexCommand): string | undefined {
    return this.attemptOperations([
        () => command.getChangeType(),
        () => command.getPreparedChange().getDefinition().changeType
    ]);
}

private getCommandSelectorId(command: FlexCommand): string | undefined {
    return this.attemptOperations([
        () => command.getSelector().id,
        () => command.getElement().getProperty('persistencyKey')
    ]);
}
```

This way we have less code and consistent return in case of an error.

packages/preview-middleware-client/src/cpe/changes/service.ts
```diff
@@ -307,4 +322,44 @@ export class ChangeService {
...
+        try {
+            const changeType = command.getChangeType();
+
+            return changeType;
+        } catch (error) {
+            return this.getCommandChangeTypeFromGetPreparedChange(command);
+        }
+    }
+
+    private getCommandChangeTypeFromGetPreparedChange(command: FlexCommand): string | undefined {
```
nikmace, line 336: Missing JSDocs for methods

nikmace (changes requested):

nikmace (approved):
Code looks good
Good test coverage
Comments addressed
Manual testing ✅

PR #1488, opened 2023-11-22
fix: update tesconfig to generate correct files order in dist

nahmed22:
changed the rootDir path to "src" instead of ".", to generate the correct files order in dist folder


voicis (approved):
Changes looks good and tested locally.

vadson71 (approved):

Issue #1489, opened 2023-11-22
BUG - Missing delete icon for new comp/control variant views


IvoSG:
### Description
Missing delete icon for new comp/control variants views

### Steps to Reproduce
1. Create a comp or control variant adaptation project
2. Go to variant menu => Save View As   
3. Refresh CPE 
4. Go to variant manu => Manage views

### Expected results
There should be a delete button on the right end for the newly created views

### Actual results
There is no way to delete already created views

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1491, opened 2023-11-22
Fix: show delete button for control variant views

IvoSG:
Fix for: [1489](https://github.com/SAP/open-ux-tools/issues/1489)

Changes related to control variants especially for creating new variant views for the adaptation project. 
These changes are related to the ability to delete variant views through the UI. The sap.ui.fl API for variant management needs the layer passed as url parameter to enable variant views for deletion


packages/preview-middleware/src/base/flp.ts
```diff
@@ -227,9 +227,13 @@ export class FlpSandbox {
                 previewUrl = `${previewUrl}.inner.html`;
                 editor.pluginScript ??= 'open/ux/preview/client/cpe/init';
                 this.router.get(editor.path, (_req: Request, res: Response) => {
+                    let templatePreviewUrl = `${previewUrl}?sap-ui-xx-viewCache=false&fiori-tools-rta-mode=forAdaptation&sap-ui-rta-skip-flex-validation=true&sap-ui-xx-condense-changes=true#${this.config.intent.object}-${this.config.intent.action}`;
+                    if (rta.options?.scenario === 'ADAPTATION_PROJECT') {
+                        templatePreviewUrl = templatePreviewUrl.replace('?', `?sap-ui-layer=${rta.layer}&`);
```
GDamyanov, line 232: Suggestion, can't we just append 'sap-ui-layer' query param at the end of the url, since the order of the query params does not matter

IvoSG: The parameters after the anchor(#) are not handled and depend on that I had two options to put it at the beginning or at the end before the anchor(#), but I chose the beginning because I think the anchor(#) can be more than one because of the app navigation

GDamyanov: Got it.

GDamyanov (approved):
Code looks good
Good test coverage
Manual testing done

packages/preview-middleware/src/base/flp.ts
zdravko-georgiev, line 232: I don't get it. Why do you need to put the layer as an URL parameter? Please elaborate.

zdravko-georgiev: Clarified.

zdravko-georgiev (approved):
Small and simple change
Code looks good
I believe that we can do a small refactoring in regards to setting the URL parameters. But this can be handled in a separate PR.
Did Not tested manually

Issue #1495, opened 2023-11-23
FEATURE - Validate UI5.yaml for adaptation projects


Dannysht:
_As a **developer**,
I want to **validate the UI5.yaml file of projects**,
so that I **can achieve some benefit**_

### Description

_If the ui5.yaml file is missing a node and any sub generator is started a technical error appears and shows quite uninformative technical error. Such exception should be handled and a user friendly Notification Error should show up explaining that the UI5.yaml file is corrupted and cannot be read (or something similar)_

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have an adaptation project
**when** I attempt to run a sub generator 
**then** a validation of its ui5.yaml file will be executed

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #1496, opened 2023-11-23
Feature - Validate ui5.yaml file for adaptation projects

Dannysht:
feature for #1495 

- Added adaptation project properties validation for ui5 yaml files


packages/adp-tooling/src/base/validate.ts
```diff
@@ -0,0 +1,70 @@
...
+     * Assert ui5 for its validity. Throws error if a property of middlewares is missing.
+     *
+     * @param properties array of property name
+     * @param target object which will be checked
+     */
+    private static assertProperties<T extends object>(properties: Properties<T>[], target: T): void {
+        for (const property of properties) {
+            const value = target[property];
+            if (value === null || value === undefined) {
+                throw new Error(`Missing ${property} in the ui5 yaml file`);
```
longieirl, line 42: Would `throw new Error(`Missing ${property} in the ui5.yaml file`);` make more sense?

zdravko-georgiev (changes requested):
This is not a good approach. This works only for  the oss middlewares. If you want to validate the configuration of adp, then you can do it e.g. in the constructor of the `ADPPreview`. Also it doesn't make any sense to validate the configuration of the other middlewares. Each middleware is exchangeable. You can have n-different middlewares that are proxying e.g. the UI5 resources.

packages/adp-tooling/src/base/validate.ts
Dannysht: Changed in https://github.com/SAP/open-ux-tools/pull/1496/commits/6dc65595d09bb26940c13a19d9de92f9b87e56ce

packages/adp-tooling/src/base/validate.ts
```diff
@@ -45,19 +47,25 @@ export default class UI5Validator {
...
         }
         this.assertProperties(['environment'], customConfiguration);
-        this.assertProperties(['configuration'], previewMiddleware);
-        this.assertProperties(['adp'], previewMiddleware.configuration);
-        this.assertProperties(['target'], previewMiddleware.configuration.adp);
-        this.assertProperties(['url', 'client'], previewMiddleware.configuration.adp.target);
+        this.assertProperties(['configuration'], fioriPreview);
+        this.assertProperties(['adp'], fioriPreview.configuration);
+        this.assertProperties(['target'], fioriPreview.configuration.adp);
+        this.assertProperties(['url', 'client'], fioriPreview.configuration.adp.target);
```
nikmace, line 65: When the user creates an ADP in BAS with the new structure, he will not have a url and client properties, which will throw an error and fail. The same will happen with the destination property on VS Code.

packages/adp-tooling/src/base/validate.ts
```diff
@@ -45,19 +47,25 @@ export default class UI5Validator {
...
         customConfiguration: AdpCustomConfiguration | undefined,
-        previewMiddleware: CustomMiddleware<PreviewMiddlewareConfiguration> | undefined
+        fioriPreview: CustomMiddleware<FioriToolsPreviewConfig> | undefined,
+        fioriProxy: CustomMiddleware<FioriToolsProxyConfig> | undefined
     ): void {
-        if (!customConfiguration || !previewMiddleware) {
+        if (!customConfiguration || !fioriPreview || !fioriProxy) {
             throw new Error('Missing required custom middleware or custom configuration in ui5.yaml');
         }
         this.assertProperties(['environment'], customConfiguration);
```
nikmace, line 61: We have removed `environment` from customConfiguration. It is ABAP by default. We have added safeMode property to customConfiguration, this one probably needs to be checked.

nikmace (changes requested):

packages/adp-tooling/src/base/validate.ts
Dannysht: There is still no BAS implementation, and we are not populating destination in ui5.yaml if a project is generated from BAS

zdravko-georgiev:
Let us please discuss in this in our sync meeting.

PR #1499, opened 2023-11-24
fix: change access type of uppercaseNameMap prop

nahmed22:
Change the access type of "upperCaseNameMap" property from Private to public,


voicis (approved):
- change set present
- code changes looks good

vadson71 (approved):

PR #1507, opened 2023-11-28
feat: add integration test for preview

mr25mr:
This PR 
* Add a new package `“@sap-ux-private/playwright”`
* Add simple integration test for preview. It test preview with different UI5 Versions:
    * 1.84.35
    * 1.96.22
    * 1.108.19
    * 1.111.8
    * 1.114.0
    * 1.115.0
* Add Github Actions
* Pave the ground for integration test for CPE on adaption project. Integration tests for CPE will follow in another PR.


.github/workflows/pipeline.yml
```diff
@@ -42,6 +42,30 @@ jobs:
...
+              run: pnpm run test:integration
+            - name: Upload playwright reports
+              uses: actions/upload-artifact@v3
+              if: always() && steps.check_changes.outputs.src == 'true'
+              with:
+                name: playwright-reports-os-${{ matrix.os }}-node-version-${{ matrix.node-version }}
+                path: |
+                  packages/*/playwright-report
+                  !packages/*/node_modules
+                retention-days: 15
```
devinea, line 68: @mr25mr Have you tried to use NX caching instead of detecting changes?
See test for example:
https://github.com/SAP/open-ux-tools/blob/main/nx.json#L9
https://github.com/SAP/open-ux-tools/blob/main/nx.json#L42-L52

Also with node_module caching we can probably skip `npx playwright install --with-deps`
https://github.com/SAP/open-ux-tools/blob/main/.github/workflows/pipeline.yml#L26-L34

mr25mr: @devinea I tried to use NX caching, does not seems to be straightforward to achieve same goal. We need to have playwright reports for each os with combination of node version e.g `playwright-reports-os-windows-latest-node-version-18.x` and using NX caching for this goal seemed a bit complicated.

regarding `npx playwright install --with-deps` we need to install browser dependencies due to this issue https://github.com/SAP/open-ux-tools/pull/1507/files#diff-0fc4e4b47236265ffb9c3ec977c54445ee6e60856bd910f44232abe7a91e1316R41
Here is official doc. https://playwright.dev/docs/ci-intro#setting-up-github-actions

one other constrains - [Each version of Playwright needs specific versions of browser binaries to operate. You will need to use the Playwright CLI to install these browsers](https://playwright.dev/docs/browsers#introduction). Caching them, not sure.

packages/preview-middleware/test/integration/preview/preview.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+---
+specVersion: "3.0"
+metadata:
+  name: backend-proxy-middleware
+kind: extension
+type: server-middleware
+middleware:
+  path: ../../../../../backend-proxy-middleware/dist/middleware.js
+`;
+};
```
zdravko-georgiev, line 81: Why do you add this? We want to test against the `preview-middleware`. The other middlewares should be installed.Right?

.eslintrc
```diff
@@ -103,7 +103,11 @@
                 "jsdoc/require-returns-description": "off",
                 "jsdoc/require-param-description": "off",
                 "max-nested-callbacks": ["warn", 5],
-                "sonarjs/cognitive-complexity": "off"
+                "sonarjs/cognitive-complexity": "off",
+                "jsdoc/require-param": "off",
+                "jsdoc/require-returns": "off",
+                "jsdoc/match-description": "off"
+
```
zdravko-georgiev, line 110: Why are you disabling those?

.changeset/bright-vans-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/preview-middleware': patch
+---
+
+Add integration test
```
zdravko-georgiev, line 5: I am not sure, if we actually need a changeset. We are not changing anything in the functionality of the package

packages/playwright/package.json
```diff
@@ -0,0 +1,51 @@
+{
+    "name": "@sap-ux-private/playwright",
```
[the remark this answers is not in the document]
mr25mr: Yes, currently there is no such request and changing it to public should be easy once there is such request. I also do not have any objection to make it public now, but [admin](https://github.com/orgs/SAP/teams/ux-tools-admin) should agree

packages/preview-middleware/test/integration/preview/preview.spec.ts
mr25mr: yes, adding as project dependencies mean we test agains release and adding like this actually test current main code base. There might not be much difference code wise, but testing current code base detect regression if any.

.changeset/bright-vans-add.md
mr25mr: in UI5 LA if there is any changes in any package, we include them in change set. I think we do same strategies in open-ux tools. Removing it  may cause pipeline checks fails.

packages/playwright/package.json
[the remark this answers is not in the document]
zdravko-georgiev: @SAP/ux-tools-admin Any objections to make this public?

packages/preview-middleware/test/integration/preview/preview.spec.ts
zdravko-georgiev: OK, then please remove the `backend-proxy-middleware`. Once this is submitted we can introduce a similar integration test for the `backend-proxy-middleware`.

.eslintrc
zdravko-georgiev: Ah, OK. Got it, thanks.

.changeset/bright-vans-add.md
zdravko-georgiev: I think that we can safely remove the changeset in this case. @devinea Is my assumption correct?

.github/workflows/pipeline.yml
devinea: As discussed https://github.com/SAP/open-ux-tools/pull/1517/commits/e3d0dcdae7a00e1af47a3906b9359496b7f3c66c seems like a viable approach using NX.
Playwright installed is not cached with that approach. But something like this looks nice https://github.com/microsoft/playwright/issues/7249#issuecomment-1385567519

packages/playwright/package.json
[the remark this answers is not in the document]
devinea: I am not opposed to making it public. 
But private is ok for now it we don't have concrete plans to consume back in tools-suite. It can be changed later.

.changeset/bright-vans-add.md
devinea: If the new module is private then a changeset is not needed.

devinea: On second thoughts, to write the change log properly it is best to include the changeset. 
Publishing to npm is not applicable in this case.

zdravko-georgiev: OK

.github/workflows/pipeline.yml
mr25mr: @devinea address in this commit https://github.com/SAP/open-ux-tools/pull/1507/commits/c889140a1b5509489577f01165fe1cf09e93dae3
Currently playwright is installed in mono repo only once, and when building [key for caching  ](https://github.com/SAP/open-ux-tools/pull/1507/files#diff-fe700949ba86403864aa7b6fdcab2c00c55e1047f73a6a282135a35e8a21e03bR54) we do not consider playwright version. It should be ok as long as we do not use different playwright versions. Getting playwright version to include it in key, does not work reliable on windows at least I tried a couple of time :)

donal-tobin-sap (dismissed):
Looks to be ok from a modules consumption point of view.

.eslintrc
tobiasqueck: Aehm, no. Why do you need to disable those for your tests?

devinea (dismissed):
New module setup and pipeline setup looks good. 
Changeset ✅ 
Private module ✅ 
Thanks for the updates.

.eslintrc
mr25mr: @tobiasqueck since this is only test file, we will work on those in a different PR after alignment on what rules should be switched off and which rules should not. I hope this is fine with you. Thanks.

tobiasqueck: I am sorry, but no. If you need the exception for a specific project temporarily, then put them there, but do not put them as global rules in the root folder.

zdravko-georgiev (approved):
Changes to preview-middleware look good
Changeset is OK
Review comments addressed

One thing that is missing is assigning code owners for the new package. But this can be done also separately.

tobiasqueck:
No objections from my side anymore, as soon as you have your 2nd approval, you can merge.

devinea (approved):
Reapproving

pnpm-lock.yaml
```diff
@@ -9795,6 +9881,16 @@ packages:
     transitivePeerDependencies:
       - debug
 
+  /axios@1.6.2:
```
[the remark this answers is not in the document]
mr25mr: @donal-tobin-sap  the `axios` module is not added in new module or to existing module which I add integration test to it. This change in pnpm-lock.yaml might have been due to result of merging with main branch.

Issue #1518, opened 2023-11-30
BUG - The extension points not shown as such in the Outline for ADP (the test.test app)


nikmace:
### Description
The extension points are not shown as such in the Outline for ADP.
When I start the CPE, in the Outline the extension points are not recognized (no icon shown, no context menu).

Note that the menu is shown in the RTA menu in the canvas.

### Steps to Reproduce
1. Create ADP with an application e.g. test.test.test connecting to a working system in VS Code
2. Start the editor with npm run editor
3. Extension points are rendered but there is no icon and they are not clickable

### Expected results
Extension Points are rendered correctly with icon and are clickable

### Actual results
No UI difference for Extension Points with other nodes.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [X] Mac OS
- [X] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1519, opened 2023-11-30
fix: the extension points not shown as such in the Outline for ADP

nikmace:
Fix for #1518.

- Changes cell rendering, fixes multiple tooltips not closing (when opening many at once, previous tooltips stay open).
- Fixes bug when trying to add fragment at the second extension point (when two extension points are under the same view, only the name of the first extension point shows)


zdravko-georgiev (approved):
Code changes look good
Very good test coverage
Manual testing ✅

voicis (approved):
- CPE changes looks good
- Good test coverage
- Tested locally

PR #1522, opened 2023-11-30
Add cds specific vocabularies for analytics support

UweSchwerk:
POC for supporting non-OData annotations in CDS LSP.

Still to do (by cds/analytics colleagues): 
 - add decent descriptions in the three new vocabularies (search for TODOs)
 - check feasibility of the content that has been drafted so far (for Terms: `$AppliesTo`, usage of `Edm.PropertyPath` and `Edm.NavigationPropertyPath`)
 - (_Done by now_) verify modeling of enum values as string with `@Validation.AllowedValues` - alternative is "real" OData `EnumType` (see `@Aggreagtion.default`https://github.com/SAP/open-ux-tools/pull/1522/files#r1410695634) 

Here is some background explanation and also a demo on how this looks in action:
https://github.com/SAP/open-ux-tools/assets/107176531/121edf7c-73b1-4abf-8377-36f5abc625df

After converting allowed values to odata `enumTypes` it looks as follows: 
image

for `@Aggregartion.default` (both value lists are added up):
image

Regarding conflicting vocabulary alias `Aggregation` (there is an Odata vocabulary and now a cds specific one with the same name): current workaround is that the cds terms are logically added to the OData vocabulary. This way we could restrict the code changes to the vocabulary package only. Drawback (I hope acceptable) is that hover/cc item text for `@Aggregation.default` show wrong namespace: 
image


packages/odata-vocabularies/src/resources/com.sap.cds.vocabularies.Aggregation.ts
```diff
@@ -0,0 +1,28 @@
...
+            // but cds has (https://pages.github.tools.sap/cap/docs/advanced/odata#aggregation-methods):
+        },
+        'AggregationDefaultType': {
+            '$Kind': 'EnumType',
+            'sum': 0,
+            'sum@Org.OData.Core.V1.Description': 'Description aggregation type sum (TODO)',
+            'min': 1,
+            'min@Org.OData.Core.V1.Description': 'Description aggregation type min (TODO)',
+            'max': 2,
+            'max@Org.OData.Core.V1.Description': 'Description aggregation type max (TODO)',
```
UweSchwerk, line 21: "real" OData EnumType

Issue #1523, opened 2023-11-30
BUG - The Button Save is not shown at all for UI5 versions < 1.110


Dannysht:
### Related Feature
Feature request: #1512 

### Description
The Button Save is not shown at all for UI5 versions < 1.110

### Steps to Reproduce
1. Open editor for adaptation project with ui5 version < 1.110
2. Save button not shown

### Expected results

Save button should shown

### Actual results

Save button is hidden

### Screenshots
If applicable, add screenshots to help explain the problem.

<img width="995" alt="Screenshot 2023-11-30 at 15 45 00" src="https://github.com/SAP/open-ux-tools/assets/59877556/3f1d74d1-451f-4865-91ad-342a6662268e">

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1524, opened 2023-11-30
Fix: Show save button

Dannysht:
Fix for: #1523 

- Save and exit button is the same in UI5 versions lower than 1.110


zdravko-georgiev:
@Dannysht Please delete branch if not needed.

PR #1525, opened 2023-11-30
Fix: show save button on UI5 versions lower than 1.110

Dannysht:
Fix for: #1523 

- Save and Exit button is the same on UI5 versions lower than 1.110


zdravko-georgiev (approved):
small and simple change

nikmace (approved):
The change is small and easy to understand.
The coverage is not applicable.

Issue #1516, opened 2023-11-29
TBI - Add options to the writers (considerations for CAP)


cianmSAP:
### Description (include screenshots)

There is a need for more options to be available to determine what files and configs are added to the Fiori apps.
In some cases there is a need to prevent certain files being added, or additional data is needed. See comments for details.

### Value
Reduces the need to patch or modify the app and it's files post generation.

### Architecture Elaboration

Yes

### Tasks
- [ ] Task 1

cianmSAP:
### Issue 1 - metadata.xml
**Description** : The writing of `/localService/metadata.xml` is not needed for CAP applications and can be skipped in the writing phase.

**Proposal** : Add a property to prevent this file being written. 

**Question** : Is the [call to the odata service writer](https://github.com/SAP/open-ux-tools/blob/11bb8fa9148a4162bc7ac32e2fc24b7530fcea32/packages/fiori-elements-writer/src/index.ts#L70) required at all? Or we should just prevent the file being written [here](https://github.com/SAP/open-ux-tools/blob/11bb8fa9148a4162bc7ac32e2fc24b7530fcea32/packages/odata-service-writer/src/index.ts#L120) 

### Issue 2 - typeRoots
**Description** : Currently the paths for the type roots [are hardcoded](https://github.com/SAP/open-ux-tools/blob/11bb8fa9148a4162bc7ac32e2fc24b7530fcea32/packages/ui5-application-writer/templates/optional/typescript/tsconfig.json#L19). In the case where the `node_modules` folder containing the required types is in a parent folder, the types are not found resulting in errors in .ts files.

**Proposal** : Add an additional property e.g `projectRoot` to the [App interface
](https://github.com/SAP/open-ux-tools/blob/11bb8fa9148a4162bc7ac32e2fc24b7530fcea32/packages/ui5-application-writer/src/types.ts#L13)

### Issue 3 - proxy middleware + opa tests

**Description** :  If the proxy middleware is not deisred, rather than removing it post generation, provide an option to prevent this being written to the yaml files. The ui5-test-writer would then require additional params, i.e the framework url e.g `https://ui5.sap.com` and the ui5 version. These would be added diretly to the `opaTests.qunit.html`

**Proposal** : Add an additional property e.g `addProxyMiddleware` to the [App options
[...]

PR #1533, opened 2023-12-01
feat(writers): options for proxy middleware and v4 opa tests

cianmSAP:
#1516 

Adds option to the relevant writers to not add the proxy middleware to the config.
Also updates the ui5-test-writer to optional pass in the framework url and ui5 version to be written in the opa test html file


tobiasqueck (changes requested):
Please see my comments in the corresponding ticket. I don't think that the changes in the odata-service-writer are needed, and therefore, many of the changes in the fiori-$$$-writers aren't needed either.

PR #1542, opened 2023-12-05
feat: Fragment name property in template for custom extension

broksy:
Adding optional `fragmentName` property to use as keys in manifest templates.


packages/fe-fpm-writer/templates/section/1.86/manifest.json
```diff
@@ -8,7 +8,11 @@
                             "content": {
                                 "body": {
                                     "sections": {
-                                        "<%- name %>": {
+                                        <% if (typeof fragmentName !== 'undefined') { %>
+                                            "<%- fragmentName %>": {
+                                        <% } else { %>
+                                            "<%- name %>": {
+                                        <% } %>
                                             "template": "<%- ns %>.<%- name %>",
```
815are, line 16: I suppose it should be `  "template": "<%- ns %>.<%- fragmentFile ?? name  %>",`,
but part with `"<%- name %>": {` should stay as before

packages/fe-fpm-writer/src/common/types.ts
```diff
@@ -13,6 +13,10 @@ export interface TabInfo {
  * Common properties for any custom element of the flexible programming model.
  */
 export interface CustomElement extends WriterConfig {
+    /**
+     * Name of the custom element fragment.
+     */
+    fragmentName?: string;
```
815are, line 19: should be moved to differnet interface, because not all custom extensions has fragment, for example custom action

packages/fe-fpm-writer/templates/column/1.84/manifest.json
```diff
@@ -14,7 +14,7 @@
                                                 "width": "<%- width %>",  
                                             <% } %>
                                             "position": <%- JSON.stringify(position) %>,
-                                            "template": "<%- ns %>.<%- name %>"
+                                            "template": <% if (typeof fragmentFile !== 'undefined') { %> "<%- ns %>.<%- fragmentFile %>" <% } else { %> "<%- ns %>.<%- name %>" <% } %>
```
815are, line 17: perhaps can be simplified like?:
```
"<%- ns %>.<%- typeof fragmentFile !== 'undefined' ? fragmentFile : name  %>
```

PR #1546, opened 2023-12-07
fix(writers): handle no metadata in service

cianmSAP:
#1516 

Updates odata service writer : localUri is not added to sap.app section of manifest if no metadata is passed.

Updates fiori elements writer: Does not rely on metadata being passed to write the opaTests, check for service type is CDS


packages/fiori-elements-writer/src/index.ts
```diff
@@ -141,9 +141,8 @@ async function generate<T extends {}>(basePath: string, data: FioriElementsApp<T
     extendManifestJson(fs, basePath, rootTemplatesPath, feApp);
 
     const packageJson: Package = JSON.parse(fs.read(packagePath));
-    // Add tests only if v4, for now, and we have metadata (and therefore a mock server config)
-    const addTest =
-        !!feApp.appOptions.addTests && feApp.service?.version === OdataVersion.v4 && !!feApp.service?.metadata;
+    // Add tests only if v4
+    const addTest = !!feApp.appOptions.addTests && feApp.service?.version === OdataVersion.v4;
```
devinea, line 145: Should there be a check for service uri here to replace the metadata check?

packages/fiori-elements-writer/src/types.ts
```diff
@@ -103,7 +103,8 @@ export interface FioriElementsApp<T> extends Ui5App {
     app: FioriApp;
     appOptions: Partial<AppOptions> & {
         /**
-         * Generate OPA based tests, if applicable to the specified template.
+         * Generate OPA based tests, if applicable to the specified template and OData Version is 4.
+         * The tests may not work if service metadata is not provided.
```
devinea, line 107: Not sure I'd bother with the change to the comments. This should be extended to support v2 tests also (in a separate TBI).

packages/odata-service-writer/test/unit/index.test.ts
```diff
@@ -172,6 +173,23 @@ describe('generate', () => {
             expect(fs.read(join(testDir, 'ui5.yaml'))).toContain('client: ');
         });
 
+        it('Valid OData V4 cds service', async () => {
+            const config = {
+                ...commonConfig,
```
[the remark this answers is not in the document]
cianmSAP: Thanks, updated

packages/odata-service-writer/src/index.ts
```diff
@@ -103,7 +103,7 @@ async function generate(basePath: string, service: OdataService, fs?: Editor): P
     }
 
     // Add mockserver entries
-    if (service.metadata) {
+    if (service.metadata && service.type === ServiceType.EDMX) {
```
tobiasqueck, line 106: I don't understand the change. Why do you provide the metadata if you don't want it to be written to the filesystem? The writer does nothing else with it.

tobiasqueck: It adds additional complexity to a writer that should ideally not care about where the project is being used.

devinea: @cianmSAP thanks for moving this in https://github.com/SAP/open-ux-tools/pull/1546/commits/1977f71dee94ca382a70d0dc1e3e153fadaef164

tobiasqueck (approved):
* changeset looks good
* changes are in the appropriate places
* I didn't test, just reviewed the code and it makes logical sense

PR #1548, opened 2023-12-08
feat(project-access): detect `.library` and generate `manifest.json`

johannes-kolbe:
- detects libraries without `manifest.json`, but with `.library`
- generates a `manifest.json` file

### DIFF
--- a/packages/project-access/src/project/search.ts
+++ b/packages/project-access/src/project/search.ts
@@ -15,6 +15,7 @@ import { fileExists, findBy, findFileUp, readJSON } from '../file';
 import { hasDependency } from './dependencies';
 import { getCapProjectType } from './cap';
 import { getWebappPath } from './ui5-config';
+import { default as generateLibraryManifest } from '@ui5/builder/tasks/generateLibraryManifest';
 
 /**
  * Map artifact to file that is specific to the artifact type. Some artifacts can
@@ -22,11 +23,11 @@ import { getWebappPath } from './ui5-config';
  * Further filtering for specific artifact types happens in the filter{Artifact}
  * functions.
  */
-const filterFileMap: Record<FioriArtifactTypes, string> = {
-    applications: FileName.Manifest,
-    adaptations: FileName.ManifestAppDescrVar,
-    extensions: FileName.ExtConfigJson,
-    libraries: FileName.Manifest
+const filterFileMap: Record<FioriArtifactTypes, string[]> = {
+    applications: [FileName.Manifest],
+    adaptations: [FileName.ManifestAppDescrVar],
+    extensions: [FileName.ExtConfigJson],
+    libraries: [FileName.Library, FileName.Manifest]
 };
 
 /**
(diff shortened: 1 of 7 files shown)


packages/project-access/src/project/search.ts
```diff
@@ -338,6 +339,47 @@ async function filterExtensions(pathMap: FileMapAndCache): Promise<ExtensionResu
...
+            fsBasePath: libraryPath
+        });
+        const workspace = await ResourceFactory.createWorkspace({ reader: fs, name: 'library', virBasePath });
+        try {
+            const getProject = () => {
+                return { getVersion: () => '1.0.0' };
+            };
+            await generateLibraryManifest({
+                workspace,
+                taskUtil: { getProject },
```
johannes-kolbe, line 367: @Klaus-Keller currently `taskUtil` is mocked here, not sure if `TaskUtil` instance can be easily implemented

Klaus-Keller: Should be fine. UI5 team has agreed to implement a unit test that ensures future versions will  work if only `getVersion` is implemented

packages/project-access/src/project/info.ts
```diff
@@ -203,3 +203,33 @@ export async function getProjectType(projectRoot: string): Promise<ProjectType>
...
     return capType;
 }
+
+/**
+ * Generates a manifest.json from a provided .library file path.
+ *
+ * @param dotLibraryPath path to .library file
+ * @returns virtually generated manifest.json
+ */
+export async function generateLibraryManifest(dotLibraryPath: string): Promise<Manifest> {
```
devinea, line 213: @johannes-kolbe Can you add a bit more information as to why this is useful?
The ui5 build when generate it anyway in the dist folder if it doesn't exist.

johannes-kolbe: @devinea we would like to offer the App Info Page with library details, and one aspect is that the `manifest.json` might not have been pre-generated. 
If it is generated (we don't read the dist folder) and copied into the src folder, we do take that, but this would also be a manual step, if I understand correctly

devinea: Thanks for the background. 
Was it considered that app info could .library file instead (/as well) as most of the same information is included?

RandomByte: If you are only interested in reusing the task (or even only the `manifestCreator` processor, I would expect that `@ui5/builder` and `@ui5/fs` is enough. That brings it down to 26MB.

If you need less, maybe the best approach is to just copy our task implementation and regularly check for updates in our code.

Klaus-Keller: @johannes-kolbe, while testing this I also ran into a problem with the dynamic import. What I did:

Create file `packages/project-access/src/project/check.ts` with content
```typescript
import { join } from 'path';
import { generateLibraryManifest } from './info';

async function main() {
    const manifest = await generateLibraryManifest(
        join(__dirname, '../../test/test-data/project/find-all-apps/libraries/dot-library/src/com/sap/library/.library')
    );
    console.log(manifest);
}

main();
```

run in `packages/project-access`
```
pnpm build
node dist/project/check.js
```

Error:
```
Error [ERR_REQUIRE_ESM]: require() of ES Module open-ux-tools/node_modules/.pnpm/@ui5+fs@3.0.5/node_modules/@ui5/fs/lib/adapters/FileSystem.js from open-ux-tools/packages/project-access/dist/project/info.js not supported.
Instead change the require of FileSystem.js in open-ux-tools/packages/project-access/dist/project/info.js to a dynamic import() which is available in all CommonJS modules.
    at open-ux-tools/packages/project-access/dist/project/info.js:249:76 {
  code: 'ERR_REQUIRE_ESM'
}
```

johannes-kolbe: @devinea thanks for the points - after evaluation with @Klaus-Keller we decided to remove the generation step from here. Separate PR https://github.com/SAP/open-ux-tools/pull/1710 created, which only takes the detection part.

johannes-kolbe:
Closing PR as manifest generation seems not feasible with taken approach.

PR #1551, opened 2023-12-08
feat(project-access): make `findFileUp` consumable

johannes-kolbe:
- Adds `findFileUp` to general `@sap-ux/project-access` export to be able to be consumed

### DIFF
--- a/.changeset/eight-bikes-enjoy.md
+++ b/.changeset/eight-bikes-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': patch
+---
+
+makes function `findFileUp` publicly available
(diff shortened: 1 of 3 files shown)


Klaus-Keller:
@johannes-kolbe to ensure `findFileUp()` is available on root, you can slightly adjust the test case https://github.com/SAP/open-ux-tools/blob/45bf0515ec82c9a7d6749a49d5e1ea8d6ba0c527/packages/project-access/test/file/file-search.test.ts#L3

to import `findFileUp()` from root:

```typescript
import { findFileUp, getFilePaths } from '../../src';
import { findFiles, findFilesByExtension } from '../../src/file';
```

johannes-kolbe:
Closing issue in favour of reusing other function.

Issue #1463, opened 2023-11-15
FEATURE - Remove label and icon in control property editor


nikmace:
_As an **tools developer**,
I want to **be able to integrate the control property editor without a name/logo**,
so that I **so that I can more freely use it**_

### Description

There is no need to have a name in the control property editor, its usage will be implied by the calling VSCode task or context menu entry or other link.

The browser tab will have no icon either and the document title will be: `Editing <app_name>`.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I start the control property editor
**when** it is started
**then** CPE will open and have no name and logo.

### Tasks
- [ ] Remove label and logo
- [ ] Unit Tests added
- [ ] Test Cases defined

voicis:
We still need to support both scenarios, when it is opened via VSCode and browser.

PR #1553, opened 2023-12-12
Remove label and icon in CPE, change how project scenario is set

nikmace:
Feat for #1463.

- Removes CPE title and logo for all scenarios.
- Changes how the project scenario is set.


packages/preview-middleware/src/base/flp.ts
```diff
@@ -227,14 +229,17 @@ export class FlpSandbox {
...
+                    if (scenario === 'ADAPTATION_PROJECT') {
                         templatePreviewUrl = templatePreviewUrl.replace('?', `?sap-ui-layer=${rta.layer}&`);
                     }
                     const template = readFileSync(join(__dirname, '../../templates/flp/editor.html'), 'utf-8');
                     const html = render(template, {
                         previewUrl: templatePreviewUrl,
-                        telemetry: rta.options?.telemetry ?? false
+                        telemetry: rta.options?.telemetry ?? false,
+                        appName: rta.options?.appName ?? '',
+                        scenario
```
GDamyanov, line 242: Just a suggestion. Would it be better to have default appName different than empty string?

GDamyanov (changes requested):

packages/preview-middleware/src/base/flp.ts
nikmace: The manifest will always be present, and from it we will retrieve the id (componentId) and the title, like [here](https://github.com/SAP/open-ux-tools/pull/1553/files/7d940d05be3b9273c24f399c199d778e10e20620#diff-b39fc9ea1aedc0aa1072f86f46fe70f7cf01422c26f13952e9fd1dcffca171a3R179). So I am expecting there will always be an id (title also but not really sure). If there is no id or title than it will default to empty string. In that case it will be `Editing` as the title in the browser tab, which is fine. And I don't think there will be a case like this.

packages/preview-middleware/src/base/flp.ts
```diff
@@ -196,16 +196,18 @@ export class FlpSandbox {
...
+            target: resources[componentId ?? id] ?? this.templateConfig.basePath,
             local: '.',
             intent: this.config.intent
         });
         this.addStandardRoutes();
         if (this.rta) {
             this.rta.options ??= {};
-            this.rta.options.baseId = componentId ?? manifest['sap.app'].id;
+            this.rta.options.baseId = componentId ?? id;
+            this.rta.options.appName = manifest['sap.app'].title ?? id;
```
zdravko-georgiev, line 210: This takes the title from the deployed app and from the adp prokject. As a result in the browser we see `Editing {{appTitle}}` instead of `Editing My Application`

packages/preview-middleware/templates/flp/editor.html
```diff
@@ -5,9 +5,8 @@
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <meta name="theme-color" content="#000000" />
         <meta name="description" content="Web page for rendering SAP Fiori Tools Control Property Editor" />
-        <link rel="shortcut icon" type="image/x-icon" href="./editor/favicon.ico" />
```
voicis, line 8: We still need to be able to add a favicon, perhaps as another optional start parameter with URI for the icon.

zdravko-georgiev: In which case do we still need it?

packages/preview-middleware/src/base/flp.ts
nikmace: Not sure if this is correct: [ef164f5](https://github.com/SAP/open-ux-tools/pull/1553/commits/ef164f5d741d8744ebae0aab55c6ac41e8bdf4a8) I. But now if we have the title from the manifest as {{appTitle}} we use 'My Application' instead.

packages/preview-middleware/templates/flp/editor.html
voicis: Never mind my comment, I was not aware that there was a decision to remove the icon.

packages/control-property-editor/src/App.tsx
```diff
@@ -8,7 +8,7 @@ import { PropertiesPanel, LeftPanel } from './panels';
 import { useLocalStorage } from './use-local-storage';
 import type { RootState } from './store';
 import { useAppDispatch } from './store';
-import { changePreviewScale } from './slice';
+import { changePreviewScale, setProjectScenario } from './slice';
```
tobiasqueck, line 11: what has the setProjectScenario functionality to do with this PR?

nikmace: It updates the store with the Scenario that was passed as props to the start method. The previous functionality was removed: https://github.com/SAP/open-ux-tools/pull/1553/files/ef164f5d741d8744ebae0aab55c6ac41e8bdf4a8#diff-48a8809da86b88a63a1cc54193f7774fdb7d06ff66f1282cd7e8fd31efb90e94L57

packages/preview-middleware/src/base/flp.ts
```diff
@@ -196,20 +196,23 @@ export class FlpSandbox {
...
             local: '.',
             intent: this.config.intent
         });
         this.addStandardRoutes();
         if (this.rta) {
+            const appTitle = manifest['sap.app'].title;
             this.rta.options ??= {};
-            this.rta.options.baseId = componentId ?? manifest['sap.app'].id;
+            this.rta.options.baseId = componentId ?? id;
+            this.rta.options.appName = appTitle === '{{appTitle}}' ? 'My Application' : appTitle || id;
```
zdravko-georgiev, line 211: I am not sure if this is too much overhead just to show a meaningful title. I would propose to just use the app ID. Let's discuss it today in our sync.

packages/preview-middleware/src/base/flp.ts
zdravko-georgiev: As discussed we will just show the id as appTitle

packages/preview-middleware/src/base/flp.ts
```diff
@@ -196,20 +196,23 @@ export class FlpSandbox {
...
         this.addApp(manifest, {
             componentId,
-            target: resources[componentId ?? manifest['sap.app'].id] ?? this.templateConfig.basePath,
+            target: resources[componentId ?? id] ?? this.templateConfig.basePath,
             local: '.',
             intent: this.config.intent
         });
         this.addStandardRoutes();
         if (this.rta) {
+            // const appTitle = manifest['sap.app'].title;
```
zdravko-georgiev, line 208: Please delete this comment

nikmace: Removed: https://github.com/SAP/open-ux-tools/pull/1553/commits/d883f00d3851cef9b527d9a6f829bce6714ddb34

packages/preview-middleware/src/base/flp.ts
zdravko-georgiev: As agreed we will just use the app ID

zdravko-georgiev (dismissed):
Code changes look good
Review comments addressed
Very good test coverage
Manual testing ✅

nikmace:
@tobiasqueck @GDamyanov @voicis would you mind doing a review when you have time? We have another PR waiting for this one to be merged.

voicis:
@nikmace I'll have a look.

packages/control-property-editor/src/App.tsx
```diff
@@ -38,8 +42,13 @@ export default function App(appProps: AppProps): ReactElement {
             sheet.cssRules.length
         );
     }, []);
-    const { t } = useTranslation();
-    const dispatch = useAppDispatch();
+
+    useEffect(() => {
+        if (scenario) {
+            dispatch(setProjectScenario(scenario));
```
voicis, line 48: Wouldn't it be simpler to set the scenario in the `start` function? As far as I know it should only happen at the start of the application and we have store with dispatch available there.

nikmace: We cannot set it there because it is not wrapped in a Provider and the store is not initialized. The first component that can do that is the App.tsx.
Screenshot 2024-01-11 at 15 40 10

nikmace: Moreover, the `start` function is not a function component but a method you call to start rendering the application. And by the rules of react you can only use hooks when:
- ✅ Calling them at the top level in the body of a [function component](https://react.dev/learn/your-first-component).
- ✅ Calling them at the top level in the body of a [custom Hook](https://react.dev/learn/reusing-logic-with-custom-hooks).

voicis: The point was to avoid hooks all together. This is what I had in mind, but you are right the store is not yet ready there and we would probably not get much simpler solution.
image

voicis: Never mind, somehow I had not saved the changes which was the issue. The change above actually works fine. Can we do that instead of using hooks in the react component? That seems more straight forward solution for setting something once at the start up.
image

nikmace: Changed in: https://github.com/SAP/open-ux-tools/pull/1553/commits/39c818755642cc6f06ad90b77692868cc6ee01f7

zdravko-georgiev (approved):
Code changes look good
Review comments addressed
Very good test coverage
Manual testing ✅

nikmace:
@GDamyanov @tobiasqueck can you please resolve the conversations so we can get this merged?

packages/control-property-editor/src/App.tsx
tobiasqueck: Ok but again what has it to do with "Remove label and icon in control property editor". I don't understand why it is required for removing label and icon.

nikmace: Ok, I get now what you are saying. It's not required true, but we do need this change. Do you think I should open a different PR just for the scenario change? Or add a comment to the PR's description of this addition?

nikmace: Done, thanks.

tobiasqueck (approved):
* code quality is good
* changeset is ok
* PR description/title matches changes

GDamyanov (approved):
Code changes look good
Review comments addressed
Very good test coverage
Manual testing done

PR #1556, opened 2023-12-13
Fix preview of adaption projects with decsriptor changes requiring a client-side merge

### LINKED ISSUE #1555: BUG - Client-side appdescriptor changes in adaptation projects are not applied in the preview
### Related Feature
Feature request: #1149

### Description
Changes in the `manifest.appdescr_variant` that require a client-side merger are not applied in the local preview.

### Steps to Reproduce
1. Get a project with such changes
2. Start the preview

### Expected results
Changes are applied.

### Actual results
Change are not applied

### Screenshots
n/a

OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
These changes are rare and were not tested 

### Fix
Only a flag was required to be added to the generated html, so that these changes are applied.

### Why was it missed
Very rare use case

### How can we avoid this
Such issues are ok and can be fixed later.

tobiasqueck:
Fixes #1555

I also tested the setting with regular apps and it doesn't cause problems (it is actually correct, just not required)

### DIFF
--- a/.changeset/silent-walls-whisper.md
+++ b/.changeset/silent-walls-whisper.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/preview-middleware': patch
+---
+
+FIX: appdescr changes that require client-side merging are not applied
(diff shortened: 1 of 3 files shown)


MarcelWaechter (approved):
Thanks :)

zdravko-georgiev (approved):
Small and easy to understand change
Tested
Did NOT tested manually

Issue #1557, opened 2023-12-13
BUG - loading of reuse libs fails


zdravko-georgiev:
### Description
The loading of reuse libs fails, e.g. when you want to start the application in Variant Management mode. The issue is caused by the fact that the reuse libs are defined into the `manifest.json` (so far so good) and then copied by the `preview-middleware` into the `data-sap-ui-libs` of the SAPUI5 bootstrap. The later causes the issue.

### Expected results
App with reuse libs is loaded

### Actual results
App with reuse libs is not loaded

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1558, opened 2023-12-13
fix(preview-middleware): exclude reuse libs from data-sap-ui-libs

zdravko-georgiev:
This fixes https://github.com/SAP/open-ux-tools/issues/1557

### DIFF
--- a/packages/preview-middleware/src/base/flp.ts
+++ b/packages/preview-middleware/src/base/flp.ts
@@ -22,6 +22,34 @@ const DEVELOPER_MODE_CONFIG = new Map([
 ]);
 
 /**
+ * SAPUI5 delivered namespaces from https://ui5.sap.com/#/api/sap
+ */
+const UI5_LIBS = [
+    'sap.apf',
+    'sap.base',
+    'sap.chart',
+    'sap.collaboration',
+    'sap.f',
+    'sap.fe',
+    'sap.fileviewer',
+    'sap.gantt',
+    'sap.landvisz',
+    'sap.m',
+    'sap.ndc',
+    'sap.ovp',
+    'sap.rules',
+    'sap.suite',
+    'sap.tnt',
+    'sap.ui',
+    'sap.uiext',
+    'sap.ushell',
+    'sap.uxap',
+    'sap.viz',
+    'sap.webanalytics',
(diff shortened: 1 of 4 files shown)


packages/preview-middleware/src/base/flp.ts
```diff
@@ -390,6 +418,27 @@ export class FlpSandbox {
...
+            const libNames = Object.keys(manifest['sap.ui5'].dependencies.libs);
+            return libNames
+                .filter((key) => {
+                    return UI5_LIBS.some((substring) => {
+                        return key === substring || key.startsWith(substring + '.');
+                    });
+                })
+                .join(',');
+        } else {
+            return '';
```
zdravko-georgiev, line 439: @tobiasqueck Should we add default libs for this case? E.g. `sap.ushell, sap.ui.core, sap.m`?

zdravko-georgiev: Done with https://github.com/SAP/open-ux-tools/pull/1558/commits/744f212637cc807847f14d3d6152d44fa32a1963

packages/preview-middleware/src/base/flp.ts
```diff
@@ -21,6 +21,34 @@ const DEVELOPER_MODE_CONFIG = new Map([
     ['xx-viewCache', 'false']
 ]);
 
+/**
+ * SAPUI5 delivered namespaces from https://ui5.sap.com/#/api/sap
+ */
+const UI5_LIBS = [
```
Klaus-Keller, line 27: I think it is fine to keep it here for now, but if required by any other tool we might consider moving this information to `@sap-ux/ui5-info`

Klaus-Keller (approved):
Fix to load reuse libraries looks good.
- Code follows guidelines and is easy to understand
- Test coverage is great
- Did no manual test

Approved from my side.

tobiasqueck (approved):
* changeset ok
* code does what it is supposed to do and is easy to read
* didn't test locally

packages/preview-middleware/src/base/flp.ts
zdravko-georgiev: Thanks Klaus, I will keep this in mind.

PR #1560, opened 2023-12-14
fix: correct  playwright config to avoid timeout on failure

mr25mr:
This PR fix:
* The job running on runner GitHub Actions 2 has exceeded the maximum execution time of 15 minutes.

**Explanation** 
Long `expect` timeout plus number of retries can cause exceed of maximum execution


zdravko-georgiev (approved):
Code change looks good

devinea (approved):

Issue #1562, opened 2023-12-14
BUG - `cds.root` is not known when calling `getCapModelAndServices` and can vary


IainSAP:
### Description
When compiling cds models is it important to know what the `root.cds` is as some of the model properties are relative to the root. It can be the `cwd` value which is variable (for example it changes during app generation or a refresh in BAS/VSCode when folders are added to the workspace) and unknown to the caller. The fix should set the `cds.root` to be a value known to the caller such as the project root. 

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

IainSAP:
Done

PR #1563, opened 2023-12-14
Fix for unknown `cds.root`

IainSAP:
Fix for: https://github.com/SAP/open-ux-tools/issues/1562

Updates to `@sap-ux/project-access:getCapModelAndServices`

- Sets cds.root from project root
- Adds logger to log cds root, path and version found for debugging and support
- Add tests to cover new loc

### DIFF
--- a/packages/project-access/package.json
+++ b/packages/project-access/package.json
@@ -34,6 +34,7 @@
     },
     "dependencies": {
         "@sap-ux/ui5-config": "workspace:*",
+        "@sap-ux/logger": "workspace:*",
         "findit2": "2.2.3",
         "mem-fs": "2.1.0",
         "mem-fs-editor": "9.4.0"
--- a/packages/project-access/package.json
+++ b/packages/project-access/package.json
@@ -34,12 +34,12 @@
     },
     "dependencies": {
         "@sap-ux/ui5-config": "workspace:*",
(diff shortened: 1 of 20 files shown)


devinea (commented):
Code changes lg. 
Do we need some test to cover this flow?

packages/project-access/package.json
```diff
@@ -34,6 +34,7 @@
     },
     "dependencies": {
         "@sap-ux/ui5-config": "workspace:*",
+        "@sap-ux/logger": "workspace:*",
```
Klaus-Keller, line 37: Shouldn't this be a `devDependency` as we use Logger as type only?

Klaus-Keller (commented):
@IainSAP, changeset is missing and a small comment about dependency, otherwise looks good.

IainSAP:
Wasnt supposed to be ready for review...not sure how that happened

devinea (approved):
re-approving. Thanks @IainSAP

Klaus-Keller (approved):
Thanks @IainSAP!
- Code to set `cds.root` and logging looks good
- Test coverage is great
- I've tested the solution to set `cds.root` locally

Approved from my side.

IainSAP:
Thanks...

PR #1572, opened 2024-01-02
feat: add target kinds to metadata element

vinayhospete:
Add target kinds directly to metadata element


packages/xml-odata-annotation-converter/src/parser/metadata.ts
```diff
@@ -260,6 +260,29 @@ function getTypeForNavigationProperty(context: Context, element: XMLElement): st
     return undefined;
 }
 
+/**
+ * Get OData target kinds for a metadata element.
+ *
+ * @param metadata element.
+ * @returns OData target kinds.
+ */
+function getEdmTargetKinds(element: MetadataElementProperties): TargetKind[] {
```
voicis, line 269: Since we only use `element.kind` here why not we use that as the parameter instead of the whole element? Then we can also use it easier when creating metadata elements, because we don't need the element itself, just its kind.

vinayhospete: ok

vinayhospete: done

voicis (approved):
- Change set present
- Code changes are straight forward and easy to follow
- Good test coverage

nahmed22 (approved):

nahmed22 (approved):
Change set maintained.
Good test coverage

voicis (approved):

PR #1576, opened 2024-01-09
fix: ignore sequence of changes in unit test

Klaus-Keller:
Unit test `packages/create/test/unit/tracing/trace.test.ts` -> `Modified yaml file` sometimes fails in Windows due to different sequence of changes. This happens sometimes the test runs the first time, subsequent runs of the test are successful. 

image

As shown in the screenshot, the differences are correct, but the sequence is wrong.

Seems the module [`diff`](https://www.npmjs.com/package/diff) is causing the issue as the array of differences returned by the call `diffTrimmedLines()` controls the output. I've checked the module and there have been been changes in the GitHub repository to address bugs on Windows, but no new version was released, yet. I'm also not certain that the issue we are facing would really be addressed by the fixes. 

As a workaround I've changed the unit test to test for correct changes, but ignore the sequence.

### DIFF
--- a/packages/create/test/unit/tracing/trace.test.ts
+++ b/packages/create/test/unit/tracing/trace.test.ts
@@ -109,16 +109,10 @@ nested:${ls}
 
         // Result check
         expect(loggerMock.info).toBeCalledWith(expect.stringContaining(`'${modifiedFile}' modified`));
-        expect(loggerMock.debug).toBeCalledWith(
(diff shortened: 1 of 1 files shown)


tobiasqueck (approved):
* clearly explained in the PR why this change is done
* just a change of a test so no changeset required

devinea (approved):
updates are clear.

PR #1554, opened 2023-12-12
CDS annotation parser package moved from UX Tools

vadson71:

PR #1580, opened 2024-01-11
fix(create): unit test for trace changes

Klaus-Keller:
Even after https://github.com/SAP/open-ux-tools/pull/977 we get unit test failures on windows, e.g. https://github.com/SAP/open-ux-tools/actions/runs/7473778321/job/20340697749?pr=1578

image

I was able to reproduce the issue on Windows, after removing the line ending `${lr}` which it `\r` the expect:
```typescript
expect(loggerMock.debug).toBeCalledWith(expect.stringContaining(`[31mrootProperty: 'prop on root'[39m${ls}`));
```

Went through, failed at another point though. By now I'm pretty confident, that the issue is related to different line endings on Windows and MacOS/Linux. 

### Update
Probably found the issue, with https://github.com/SAP/open-ux-tools/pull/1554 file `.gitattributes` was introduce unifying the line endings. **-> Adjusted the test case.**

### DIFF
--- a/packages/create/test/unit/tracing/trace.test.ts
+++ b/packages/create/test/unit/tracing/trace.test.ts
@@ -89,15 +89,17 @@ describe('Test traceChanges()', () => {
 
     test('Modified yaml file', async () => {
         // Mock setup
-        const ls = process.platform === 'win32' ? '\r' : '';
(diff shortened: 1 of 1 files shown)


zdravko-georgiev (approved):

devinea (approved):
thanks

Issue #1581, opened 2024-01-11
TBI - Deployment - Handle new changes for implicit security session creation for ABAP Cloud 2402


cianmSAP:
### Description (include screenshots)
This TBI will need to implement the session creation and deletion for ABAP calls for 2402 steampunk systems.

An API caller that wants to use security sessions has to pass the http header x-sap-security-session, which signals the desired action as follows:

`x-sap-security-session` with value `create` - when performing the first request
Effect: the AS ABAP system issues an HTTP Security Session cookie after a successful authentication (if the conditions of SAP Note [3201227](https://i7p.wdf.sap.corp/sap/support/notes/3201227) are met).
All subsequent calls are supposed to transmit all cookies (without filtering for particular cookie names) which have been received with the first request.

`x-sap-security-session` with value `delete` - when performing the last request
Effect: the AS ABAP system invalidates the security session cookie and releases all resources on server side after the request has been processed. So there is no need to call the ICF logoff service and one roundtrip can be avoided.

### Value
Ensure deployment continues to work after this change.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Updates in `axios-extension/src/abap/ui5-abap-repository-service.ts`

cianmSAP:
PR merged and issue is resolved

PR #1583, opened 2024-01-11
tbi(axios-ext): handle security sessions for deployment

cianmSAP:
#1581

Adds `x-sap-security-session` with value `create`, when performing the first request to the backend to check for existing app
Adds `x-sap-security-session` with value `delete`, when performing the last request i.e deployment / undeployment


cianmSAP:
Closing, https://github.com/SAP/open-ux-tools/pull/1584#event-11468146672 is the preferred approach

PR #1584, opened 2024-01-12
feat(axios-extension): use token for connecting to abap cloud

zdravko-georgiev:
https://github.com/SAP/open-ux-tools/issues/1581. The original idea was to create a security session, but after checking with the ABAP Cloud team, it became that we do not necessarily need to create a security session. Instead we can use token for connecting.


cianmSAP (dismissed):
Code updates look good
I've tested locally and deployment works as expected

cianmSAP (approved):
Same comments as above, improved code coverage ✅

Klaus-Keller (approved):

Klaus-Keller (approved):
Thanks @zdravko-georgiev, 
- change for sending bearer token looks good
- test coverage is good
- did not manually test it

Approved from my side.

zdravko-georgiev:
Thanks guys!

PR #1585, opened 2024-01-12
feat(odata-voc): add cds analytics vocabularies when requested

UweSchwerk:
- add vocabularies containing terms only used in cds LSP (scenario: use analytical engine).
- load vocabularies based new flag


UweSchwerk:
Abandoned because commits have been pushed with wrong email addres

PR #1587, opened 2024-01-12
feat(odata-voc): add cds analytics vocabularies when requested 

UweSchwerk:
- add vocabularies containing terms only used in cds LSP (scenario: use analytical engine).
- load vocabularies based on new flag


packages/odata-vocabularies/src/vocabulary-service.ts
```diff
@@ -104,10 +104,11 @@ export class VocabularyService {
     /**
      *
      * @param includeCds Flag indicating if CDS vocabulary be loaded
+     * @param includeCdsAnalytics (for includeCds=true only) Flag indicating if additional vocabularies for CDS analytics should be loaded
      * @returns Vocabulary service instance
      */
-    constructor(includeCds?: boolean) {
-        let vocabularyInformation = loadVocabulariesInformation(includeCds);
+    constructor(includeCds?: boolean, includeCdsAnalytics?: boolean) {
```
voicis, line 110: Now users can provide a combination of parameters that doesn't make sense e.g `includeCds=false` and `includeCdsAnalytics=true`. Now we silently ignore the parameter without telling the user anything and we have to check both parameters each time to make sure the correct vocabularies are loaded. 

We probably should either throw early and inform the user if wrong parameters are used or use a different kind of  parameters in a way where you can't have wrong combination of parameters.

One option with several layers of parameters
```typescript
interface Options {
  includeCds?: boolean | CdsOptions
}

interface CdsOptions {
  analytics?: boolean;
}
```

Or use union type.
```typescript
type VocabularyServiceMode = 'cds' | 'cds-analytics';
```

UweSchwerk: I simply wanted to avoid incompatible changes of the interface. Search for `new VocabularyService` gives 70+ results in `tools-suite` repo.
From the name of the parameters it should be clear that combination `newVocabularyService(false, true)` does not make sense. I will add code to throw an exception.

UweSchwerk: Done

voicis (approved):
- Code changes are straight forward and easy to follow
- Great test coverage
- Review comments are addressed

marufrasully (approved):
LGTM

* Covered new code with unit test
* address review comment

Issue #1589, opened 2024-01-15
FEATURE - Automatically detect the layer that is to be used in an adaption project


tobiasqueck:
_As a **developer**,
I want to **generate an adaptation project without being asked for the layer**,
so that I **get always the correct layer in my generated project**_

### Description
Right now, we are asking the user to pick either `vendor` or `customer_base` when generating a project, however, it is not really a choice, but a user has to pick the correct one based on whether it is an adp project created by SAP or not. 
The information whether it is an SAP app or not is available using the ATO service, so our adp prompting could fetch it and then set the layer (and remove the question).

### Technical Design
All of it should be done in the `adp-tooling` module and can be tested using the `create` module. It might be that some functionality could go into the `system-access` module.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I use the `@sap-ux/create` module to generate an adaptation project
**when** I enter a valid system url and credentials
**then** I will get an adaptation project generated with the correct layer.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #1590, opened 2024-01-15
Remove prompting for `layer` and determine it from the backend instead

tobiasqueck:
Implementation for #1589 additionally, it fetches the available apps and offers them for auto-completion. 

Best way to locally/manually test it:
```bash
pnpm build
cd packages/create
mkdir .tmp
cd .tmp
node ../dist/index.js generate adaptation-project
```


packages/adp-tooling/src/base/prompt.ts
```diff
@@ -19,67 +23,89 @@ export type PromptDefaults = {
...
+        target,
+        {
+            ignoreCertErrors: defaults.ignoreCertErrors
+        },
+        true,
+        logger
+    );
+    logger.info('Fetching system information...');
+    const ato = await provider.getAtoInfo();
+    const layer = ato.tenantType === 'SAP' ? 'VENDOR' : 'CUSTOMER_BASE';
```
zdravko-georgiev, line 59: We can use here `flexLayer.VENDOR` and `flexLayer.CUSTOMER_BASE`.

zdravko-georgiev, line 59: Maybe we can have a global const for `SAP`? It is used as string 3 times already.

tobiasqueck: all WIP ... don't review yet. The whole code is still ugly and wrongly structured. Just tried to get it running to see if it works as I assumed.

tobiasqueck: That's no cleaned up. I am not using enums but types.

tobiasqueck: Now that part of the PR is reviewable

Klaus-Keller (commented):
Thanks @tobiasqueck, small proposal, otherwise it looks good.

packages/adp-tooling/src/base/helper.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Checks if the input is a non-empty string.
+ *
+ * @param input - input to check
+ * @returns true if the input is a non-empty string
+ */
+export function isNotEmptyString(input: string | undefined): boolean {
+    return input !== undefined && input.trim().length > 0;
```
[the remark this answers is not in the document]
tobiasqueck: Not sure CoPilot and I were ok with this one :)

How about `return !!input && input.trim().length > 0;`?
I honestly don't have a preference. So, let me know which one you prefer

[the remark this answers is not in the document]
Klaus-Keller: Just thought because the function name indicates you check for a not empty **string**, so I would do exactly that. In some rare corner cases a `number`, `boolean`, `object`, or `null` could be passed into the function and it would throw an error, whereas a check for `typeof === 'string'` would just be false.

zdravko-georgiev (approved):
Code changes look good
New functionality is covered by tests
Changeset ✅ 
Manual testing ✅

PR #1592, opened 2024-01-16
feat: Replace auto-refresh with message in case of changes from under…

nahmed22:
-  Watch changes in project using file watchers
-  Push notifications to the CPE from preview middleware that files have changed via web sockets
-  Add message in CPE that something has changed in the project and reload preview when clicked


tobiasqueck (changes requested):
The concept needs to be reviewed. We already have an app reload middleware, so instead of pushing this functionality into preview middleware, let's first have a look whether the app reload middleware can be enhanced to fullfill the requirement.

vadson71:
@tobiasqueck 
Could you please review once again

tobiasqueck (changes requested):
There shouldn't be any changes in the `preview-middleware`, or am I missing something? If not, please remove them.

nahmed22:
> There shouldn't be any changes in the `preview-middleware`, or am I missing something? If not, please remove them.

updated and removed unnecessary code

packages/reload-middleware/README.md
```diff
@@ -0,0 +1,68 @@
...
+| Option                 | Type      | Default Value    | Description                                                                                                                         |
+| ---------------------- | --------- | ---------------- | ----------------------------------------------------------------------------------------------------------------------------------- |
+| `debug`                | `boolean` | `false`          | Enables debug output                                                                                                                |
+
+## Usage
+The middleware can be used without configuration. However, since the middleware intercepts a few requests that might otherwise be handled by a different middleware, it is recommended to run it before other file serving middlewares e.g. `backend-proxy-middleware` and `ui5-proxy-middleware` (and the corresponding middlewares in the `@sap/ux-ui5-tooling`).
+
+
+### Minimal Configuration
+With no configuration provided, the local FLP will be available at `/test/flp.html` and the log level is `info`.
```
tobiasqueck, line 20: That looks like copied and pasted from the `preview-middleware` readme. modify this text.

packages/reload-middleware/README.md
```diff
@@ -0,0 +1,68 @@
+#  `@sap-ux/reload-middleware`
+
+The `@sap-ux/reload-middleware` is a [Custom UI5 Server Middleware](https://sap.github.io/ui5-tooling/pages/extensibility/CustomServerMiddleware) which creates a websocket server upon the current running UI5 Server and also creates a filesystem watcher that listens for filesystem changes to the current project's flex changes files.
+
+This allows to create websocket connections to the current UI5 server and receive notifications through it about updated, created, or deleted flex change files.
```
tobiasqueck, line 5: The middleware should be covering more than just flex hanges

packages/reload-middleware/package.json
```diff
@@ -0,0 +1,60 @@
...
+        "LICENSE",
+        "dist",
+        "ui5.yaml",
+        "!dist/*.map",
+        "!dist/**/*.map"
+    ],
+    "dependencies": {
+        "ui5-middleware-websocket": "3.0.1",
+        "@sap-ux/logger": "workspace:*",
+        "@types/ws": "8.5.10"
```
voicis, line 40: Move this to `devDepenencies` as types are not needed at runtime.

packages/reload-middleware/package.json
```diff
@@ -0,0 +1,60 @@
...
+    },
+    "devDependencies": {
+        "@sap-ux-private/playwright": "0.0.2",
+        "@types/express": "4.17.21",
+        "@types/fb-watchman": "2.0.4",
+        "@types/supertest": "2.0.12",
+        "copyfiles": "2.4.1",
+        "chokidar" : "3.6.0",
```
voicis, line 48: Shouldn't this be a dependency?

packages/reload-middleware/package.json
```diff
@@ -0,0 +1,60 @@
...
+    ],
+    "dependencies": {
+        "ui5-middleware-websocket": "3.0.1",
+        "@sap-ux/logger": "workspace:*",
+        "@types/ws": "8.5.10"
+    },
+    "devDependencies": {
+        "@sap-ux-private/playwright": "0.0.2",
+        "@types/express": "4.17.21",
+        "@types/fb-watchman": "2.0.4",
```
voicis, line 45: This is no longer needed.

packages/reload-middleware/package.json
```diff
@@ -0,0 +1,60 @@
...
+        "@types/supertest": "2.0.12",
+        "copyfiles": "2.4.1",
+        "chokidar" : "3.6.0",
+        "npm-run-all": "4.1.5",
+        "supertest": "6.3.3",
+        "ws": "8.16.0"
```
voicis, line 51: Shouldn't this be a dependency?

MarcelWaechter:
> > MarcelWaechter
> 
> Hi @MarcelWaechter , We have made changes to also include listening for manifest file changes. Could you please verify?

Hi,
I gave it a quick try, I did the following:
- checked out your change in UX-Tools, did a pnpm i and pnpm build
- also pnpm i in open-ux-tools-playground
- and then run editor-v4 in this v4 project

I created a manifest change... but also after saving and clicking on navigation I didn't realize any refresh (also the changes were not applied, verified in the code)
Is there anything else I have to do to test your change?

packages/reload-middleware/package.json
nahmed22: moved

packages/reload-middleware/package.json
nahmed22: moved

packages/reload-middleware/package.json
nahmed22: moved

zdravko-georgiev (changes requested):
Can we please split this PR in 2. One for the creation of the new middleware and one for the consumption?

voicis:
> Can we please split this PR in 2. One for the creation of the new middleware and one for the consumption?

Do you mean one PR for the new middleware and the other one for the changes? We could do that, but I expect it would be more difficult to test it then.

tobiasqueck:
> > Can we please split this PR in 2. One for the creation of the new middleware and one for the consumption?
> 
> Do you mean one PR for the new middleware and the other one for the changes? We could do that, but I expect it would be more difficult to test it then.

I think the split would make sense to ensure that that middleware is also usable without CPE. The CPE integration should be a specific use case.

zdravko-georgiev:
> > Can we please split this PR in 2. One for the creation of the new middleware and one for the consumption?
> 
> Do you mean one PR for the new middleware and the other one for the changes? We could do that, but I expect it would be more difficult to test it then.

Yes, this is what I mean. It shouldn't be more difficult to test, since we already have all the necessary infrastructure in place.

tobiasqueck:
My previous test didn't work because I wrongly configured the middleware i.e. I didn't read the `README` and therefore didn't use `afterMiddleware: reload-middleware` in my `preview-middleware`. Having that fixed, the general flow works, however, I have the following new/odd behavior.

I just opened the editor (didn't do any change) and then I modified and existing xml fragment. I saw the warning that there are changes on the filesystem and I clicked on reload resulting in the following popups:
Screenshot 2024-03-13 at 17 47 56

voicis:
That is quite strange one. The issue is not reproducible if i have developer tools open, if I close them then it is reproducible again. Normally the popup would only show after saving some changes that requires reload e.g. changes in manifest.

vadson71:
On my machine if I have simultaneously an external change and unsaved UI change then the warning offers `Save and Reload`, and if I click on it then first draft save confirmation dialog opens and then this reload confirmation dialog. It seems to be some new feature in RTA.
image
image

tobiasqueck (approved):
* changeset ok
* local tests successful
* very good test coverage
* coding adheres to our coding standards

tobiasqueck:
> @tobiasqueck can you have another look. The dialog with "App Requires Reload" should not be displayed anymore and after saving manifest changes there should be an option to reload the app (same as other manual changes in project files).

@voicis works like a charm now!

zdravko-georgiev (approved):
Code change looks good and it is well structured
Very good test coverage
Did NOT tested manually

PR #1599, opened 2024-01-18
fix(project-access): detecting apps that are part of CAP

Klaus-Keller:
Function `findFioriArtifacts()` from `@sap-ux/project-access` does not detect an app correctly, if 
- embedded into CAP
- has entry `@sap/ux-ui5-tooling` in `devDependencies` of package.json and file `ui5-local.yaml`

The project root path was pointing to the application root instead of the CAP project root.

### DIFF
--- a/.changeset/seven-emus-dress.md
+++ b/.changeset/seven-emus-dress.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': patch
+---
+
+Fix for detecting apps that are part of CAP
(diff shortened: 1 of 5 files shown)


zdravko-georgiev (approved):
Simple and easy to understand change
Excellent test coverage
Did NOT tested manually

devinea (approved):
code changes seem appropriate and covered by tests 
changeset ✅

PR #1600, opened 2024-01-19
feat: i18n support

marufrasully:


marufrasully:
Will be covered in this PR https://github.com/SAP/open-ux-tools/pull/1655

somehow this PR and specially lock-file seems to be wrong. and caused unit tests to fail.

PR #1607, opened 2024-01-23
fix(fpm-writer): adjust fpm writer tests to avoid test failure on windows

815are:
I noticed following difference running on Windows:
image
vs
image

Run was on fresh clone vs old clope of `fpm-writer`.

Adjusting tests by handling both cases - detecting if new line in existing file is `\n`(1 symbol) or `\r\n`(2 symbols). Method -> `getEndOfLinesLength`

### DIFF
--- a/packages/fe-fpm-writer/test/common/index.ts
+++ b/packages/fe-fpm-writer/test/common/index.ts
@@ -1,3 +1,4 @@
+import os from 'os';
 import { rmdirSync, existsSync } from 'fs';
 import type { Editor } from 'mem-fs-editor';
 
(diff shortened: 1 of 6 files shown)


Klaus-Keller (approved):
Thank you @815are for fixing the unit test on Windows. I assume it is similar to https://github.com/SAP/open-ux-tools/pull/1580, root cause was addition of `.gitattribues` file with https://github.com/SAP/open-ux-tools/pull/1554.

Tested the branch locally on Mac, works.

Approved from my side.

devinea (approved):
Changes to tests look appropriate. Thanks

PR #1608, opened 2024-01-23
fix: CPE Icons on Changes Panel not easily visible in Light Theme (#2…

DanyalAslam:
This fix addresses the issue related to the visibility of CPE icons in the light theme. 

Navigate to the CPE. Subsequently switch the theme to Light Mode. It is observed that CPE Icons are not easily visible on light mode. 

CPE icons are now visible on light theme also, furthermore removed the inline styling from the UIIcon component and moved to the CSS class.

Screenshot 2024-01-23 at 19 13 13

### DIFF
--- a/packages/control-property-editor/src/Workarounds.scss
+++ b/packages/control-property-editor/src/Workarounds.scss
@@ -48,6 +48,22 @@
     }
 }
 
+// CPE-UIIcon-Changes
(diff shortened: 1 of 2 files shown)


marufrasully (approved):
LGTM
Easy straightforward change.

vinayhospete (approved):

Issue #1621, opened 2024-01-30
BUG - XML fragment is written to project before save button is pressed


nikmace:
### Description
XML fragment is written to the project before the save button is pressed. This behavior is unexpected, the XML fragment and the change file should be written to the project's workspace at the same time when clicking save button.

### Steps to Reproduce
1. Start an Adaptation Project.
2. Create a new fragment change from the Add Fragment dialog
3. Fragment is written to the workspace, but the save button is active

### Expected results
XML fragment is written together with the change file when clicking the save button.

### Actual results
XML fragment is created without change.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [X] Mac OS
- [X] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1625, opened 2024-01-31
fix: XML fragment is written to project before save button is pressed

nikmace:
Fix for #1621.

- Adds caching for fragment paths, and serves them after the command for addXML is pushed and executed and the fragment is requested.


nikmace:
Will not be merged. Closing this PR.

Issue #1615, opened 2024-01-25
FEATURE - Append an onpremise warning if the target deployment endpoint is onpremise


longieirl:
_As an ABAP developer
I want to see a depoyment info message that indicates if my ABAP target system is onPremise,
so that I can update the target URL to reflect the internal protocol, host and port using my onpremise configuration__

### Description

Using the following ui5-deploy.yaml configuation;
```yaml
# yaml-language-server: $schema=https://sap.github.io/ui5-tooling/schema/ui5.yaml.json

specVersion: "3.1"
metadata:
  name: ztra28111437
type: application
builder:
  resources:
    excludes:
      - /test/**
      - /localService/**
  customTasks:
    - name: deploy-to-abap
      afterTask: generateCachebusterInfo
      configuration:
        target:
          destination: MyDestination
          url: https://my-host.abap
        app:
          name: ZJLO28111449
          package: Z_I313149_PACKAGE
          transport: Y08K901166
        exclude:
          - /test/
```

The following `info` is currently shown in the console log;

```bash
info abap-deploy-task ZZ1_DEMO_AP Please copy/paste this URL in a browser for more details:
info abap-deploy-task ZZ1_DEMO_AP (Note: You will need to replace the host in the URL with the internal host, if your destination is configured using an On-Premise SAP Cloud Connector)
```

Curerntly, the SCC note is append for all destinations;
```js
if (this.isDest) {
                    this.log.info(
                        '(Note: You will need to replace the host in the URL with the internal host, if your destination is configured using an On-Premise SAP Cloud Connector)'
                    );
                }
```

The onpremise SCC message should only be shown if the destination being referenced is actually on premise and the info message should be updated to be explicit of what they need to change (exact message details to be append later).

### Technical Design
axios-extension needs to be updated to handle this flow

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have an ABAP deploy configuration and the referenced destination is onpremise
**when** I execute a successful deploy ABAP command
**then** the console output is updated to append an info note detailing what I need to change

### Tasks
- [ ] Unit Tests added
- [ ] Test Cases defined 
- [ ] Update the `deploy` command in `axios-extension` to call btp-utils to retreive a list of SAP BTP destinations, filter by the name of the referenced destination and determine if its onpremise and if the additional info should be appended

longieirl:
Confirmed, working as expected in tools-suite.

PR #1642, opened 2024-02-06
fix(axios-ext): fix message order, remove showAddInfo flag

martinjanus-sap:
#1615

Removed `showAddInfo` and the order of messages displayed as: 

```
info abap-deploy-task YY1_TESTIAN334 App available at https://my301102.s4hana.ondemand.com/sap/bc/ui5_ui5/sap/yy1_testian334
info abap-deploy-task YY1_TESTIAN334 Deployment Successful
info abap-deploy-task YY1_TESTIAN334 (Note: As the destination is configured using an On-Premise SAP Cloud Connector, you will need to replace the host in the URL above with the internal host)
```

### DIFF
--- a/packages/deploy-tooling/src/base/deploy.ts
+++ b/packages/deploy-tooling/src/base/deploy.ts
@@ -295,8 +295,7 @@ async function tryDeploy(
                 archive,
                 bsp: config.app,
                 testMode: config.test,
-                safeMode: config.safe,
-                showAddInfo: await showAdditionalInfoForOnPrem(`${config.target.destination}`)
+                safeMode: config.safe
             });
         } else {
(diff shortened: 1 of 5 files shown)


packages/deploy-tooling/src/base/deploy.ts
```diff
@@ -307,6 +306,11 @@ async function tryDeploy(
             );
         } else {
             logger.info('Deployment Successful.');
+            if (await showAdditionalInfoForOnPrem(`${config.target.destination}`)) {
```
longieirl, line 309: Just need unit tests around these changes.

cianmSAP (approved):
Changes look good
Code coverage is excellent
Changeset

longieirl (approved):
Thanks.

Issue #1637, opened 2024-02-06
FEATURE - Switch Adaptation Project's manifests editors/wizards to use Open Source writers


nikmace:
_As a **developer**,
I want to **move part of the writing functionality to OS repository**,
so that I **can further make the writing identical independent of environment**_

### Description

_Wizards should be adjusted to use the common writers of Fiori Tools to further split the code and make the writing be independent of environment._

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have an Adaptation Project yeoman generators,
**when** I reuse the writing functionality and run the editor,
**then** writing will happen correctly.

### Tasks
- [ ] Enahnce writing module for adp-tooling to allow for manifest editors to reuse this functionality
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #1643, opened 2024-02-06
feat: Switch Adaptation Project's manifests editors/wizards to use Open Source writers

nikmace:
Feat for #1637.

- Adds writer functionality for manifest change editors.


packages/adp-tooling/src/base/change-utils.ts
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * @param {string} projectPath - The root path of the project.
+ * @param {AnnotationsData} data - The data object containing information about the annotation change.
+ * @param {object} change - The annotation data change that will be written.
+ * @param {Editor} fs - The `mem-fs-editor` instance used for file operations.
+ * @returns {void}
+ */
+export function writeAnnotationChange(projectPath: string, data: AnnotationsData, change: object, fs: Editor): void {
+    try {
+        const { answers, timestamp, annotationFileName } = data;
```
testojs, line 30: Do you pass the answers from the Yeoman generators? If yes, please make sure that the writers are not bound to yeoman and can be used separately.

packages/adp-tooling/src/base/change-utils.ts
```diff
@@ -0,0 +1,202 @@
...
+        const changeFileName = `id_${timestamp}_addAnnotationsToOData.change`;
+        const changesFolderPath = path.join(projectPath, FolderTypes.WEBAPP, FolderTypes.CHANGES);
+        const manifestFolderPath = path.join(changesFolderPath, FolderTypes.MANIFEST);
+        const annotationsFolderPath = path.join(changesFolderPath, FolderTypes.ANNOTATIONS);
+
+        writeChangeToFile(`${manifestFolderPath}/${changeFileName}`, change, fs);
+
+        if (answers.targetAnnotationFileSelectOption === AnnotationFileSelectType.NewEmptyFile) {
+            fs.write(`${annotationsFolderPath}/${annotationFileName}`, '');
```
testojs, line 39: Please check if forward slashes in the paths will work for different operating systems. I think we had issues with that and it was fixed.

packages/adp-tooling/src/base/change-utils.ts
nikmace: The writers can be used outside of the yeoman environment, only that you need to provide the correct data described in the interface for that particular generator type. The method you've pointed here is a helper function for the writers and it is not exported from @sap/adp-tooling/index.ts but is used internally.

testojs: Can we find better names for these input parameters and not call them `answers` like in Yeoman? There are no questions in the writes so probably answers does not make sense.

packages/adp-tooling/src/base/change-utils.ts
nikmace: I've checked on Windows, with what is pushed till this moment and I can not reproduce the old issue and also don't understand why it was an issue in the first place and how it got to that point that it started generating some strange paths inside the change. Tested all the sub-generators and saw no problem (following the same steps described in the issue). Hopefully, the issue is obsolete.

packages/adp-tooling/src/base/change-utils.ts
nikmace: @testojs These writers can be used in CLI but again in the CLI there can be prompting so the user answers some questions. I am not sure what is a better name instead of `answers`. But this is not yeoman specific.

Maybe `responses`, `userInput`, `data` (which is too generic), or `outputs`?

packages/adp-tooling/src/base/change-utils.ts
```diff
@@ -0,0 +1,202 @@
...
+ * Constructs a generic change object based on provided parameters.
+ *
+ * @param {T} data - The base data associated with the change, including project data and timestamp.
+ * @param {object} content - The content of the change to be applied.
+ * @param {GeneratorName} generatorName - The name of the generator creating this change.
+ * @param {ChangeTypes} changeType - The type of the change.
+ * @returns An object representing the change.
+ * @template T - A type parameter extending `BaseData`.
+ */
+export function getGenericChange<T extends BaseData>(
```
tobiasqueck, line 181: Why do you use generics here? If you just make `date: BaseData` then you are good to go. I don't see any generic usage, or am I missing something?

packages/adp-tooling/src/base/change-utils.ts
```diff
@@ -0,0 +1,202 @@
...
+    content: object,
+    generatorName: GeneratorName,
+    changeType: ChangeTypes
+) {
+    const { projectData, timestamp } = data;
+    const fileName = `id_${timestamp}`;
+
+    return {
+        'fileName': fileName,
+        'namespace': posix.join(projectData.namespace, 'changes'),
```
tobiasqueck, line 192: Semantically it would be better to use `[projectData.namespace, 'changes'].join('/') because you are concatenating strings with a slash here, you are not creating a posix path (even if the end result is the same).

packages/adp-tooling/src/types.ts
```diff
@@ -138,3 +142,149 @@ export const enum HttpStatusCodes {
...
+
+export type PropertyValueType = 'boolean' | 'number' | 'string' | 'binding' | 'object';
+
+export interface AnnotationChangeAnswers {
+    targetODataSource: string;
+    targetAnnotationFileSelectOption: AnnotationFileSelectType;
+    targetAnnotationFilePath: string;
+}
+
+export interface ComponentUsagesAnswers {
```
tobiasqueck, line 239: Same as before, something like
```
interface ComponentUsage {
  if: string;
  name: string;
  ...
}
```
is more concise and carries the same informational value

packages/adp-tooling/src/types.ts
```diff
@@ -138,3 +142,149 @@ export const enum HttpStatusCodes {
...
+
+export const enum GeneratorType {
+    ADD_ANNOTATIONS_TO_ODATA = 'Add Annotations to OData',
+    ADD_COMPONENT_USAGES = 'Add Component Usages',
+    ADD_NEW_MODEL = 'Add New Model',
+    CHANGE_DATA_SOURCE = 'Change Data Source',
+    CHANGE_INBOUND = 'Change Inbound'
+}
+
+export const enum GeneratorName {
```
tobiasqueck, line 160: What is the need for the specific generator names? In the end they are only used when writing the change, correct? If yes, could we simply write `@sap-ux/adp-tooling` in all of them?

nikmace: I think that when you pass incorrect data in a particular manifest change or the change is in the incorrect format, then the support.generator will help us understand which generator has produced the wrong change. 

Also, there might be issues if you write @sap-ux/adp-tooling, maybe the generator name is taken into account server-side when calling the API endpoint to get the merged manifest.

packages/adp-tooling/src/types.ts
```diff
@@ -138,3 +142,149 @@ export const enum HttpStatusCodes {
...
+    projectData: AdpProjectData;
+    timestamp: number;
+}
+
+export interface AnnotationsData extends BaseData {
+    annotationFileName?: string;
+    isInternalUsage: boolean;
+    answers: AnnotationChangeAnswers;
+}
+export interface ComponentUsagesData extends BaseData {
```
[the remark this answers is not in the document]
nikmace: The idea was that there is BaseData that encapsulates common properties which should be passed to the change generator from all the sub-generators. Then each generator extends this base interface to include additional properties relevant to their specific context.

packages/adp-tooling/src/types.ts
```diff
@@ -138,3 +142,149 @@ export const enum HttpStatusCodes {
...
+    uri?: string;
+    type: string;
+    settings: {
+        [key: string]: unknown;
+    };
+};
+
+export type PropertyValueType = 'boolean' | 'number' | 'string' | 'binding' | 'object';
+
+export interface AnnotationChangeAnswers {
```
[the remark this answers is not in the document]
nikmace: For the **Answers** I agree, for the **target**, it is like this when prompting in yeoman. The answer is put into the variable **targetAnnotationFileSelectOption**, for example. For consistency, the interface reflects the same names as in the prompts. Therefore, to change this I would need a lot more time to go and change each prompt, and the tests, and some internal logic. Don't like the **target** prefix too.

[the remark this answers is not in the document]
tobiasqueck: >  For consistency, the interface reflects the same names as in the prompts

No, please not. With the splitting of prompting and writing instead of providing a generator, we made the decision to define the API of the writer independent of any prompts, so that we can easily reuse them wherever needed and don't define an API for a specific consumer.
So, please, when creating a writer API, do not think about the prompting, just think about the properties required for the writing. Also think about defaults.

packages/adp-tooling/src/types.ts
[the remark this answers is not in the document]
nikmace: Changed in: [3ed676b](https://github.com/SAP/open-ux-tools/pull/1643/commits/3ed676b560a71c69eac49e8258cf84f78c948ade)

packages/adp-tooling/src/types.ts
nikmace: Changed in: [3ed676b](https://github.com/SAP/open-ux-tools/pull/1643/commits/3ed676b560a71c69eac49e8258cf84f78c948ade)

packages/adp-tooling/src/types.ts
nikmace: Changed in: [3ed676b](https://github.com/SAP/open-ux-tools/pull/1643/commits/3ed676b560a71c69eac49e8258cf84f78c948ade)

packages/adp-tooling/src/types.ts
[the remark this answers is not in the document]
nikmace: Changed in: [3ed676b](https://github.com/SAP/open-ux-tools/pull/1643/commits/3ed676b560a71c69eac49e8258cf84f78c948ade)

packages/adp-tooling/src/base/change-utils.ts
nikmace: Changed in: [3ed676b](https://github.com/SAP/open-ux-tools/pull/1643/commits/3ed676b560a71c69eac49e8258cf84f78c948ade)

packages/adp-tooling/src/base/change-utils.ts
nikmace: Changed in: [3ed676b](https://github.com/SAP/open-ux-tools/pull/1643/commits/3ed676b560a71c69eac49e8258cf84f78c948ade)

packages/adp-tooling/src/base/change-utils.ts
nikmace: Added in: [83d3688](https://github.com/SAP/open-ux-tools/pull/1643/commits/83d36886cfe672437a7497660bd838323e03d8bb)

packages/adp-tooling/src/base/change-utils.ts
nikmace: Changed in: [3ed676b](https://github.com/SAP/open-ux-tools/pull/1643/commits/3ed676b560a71c69eac49e8258cf84f78c948ade)

packages/adp-tooling/src/base/change-utils.ts
tobiasqueck: Thanks, I think you need to update the JSDoc comments to reflect the change of the input parameters.

nikmace: [29624c3](https://github.com/SAP/open-ux-tools/pull/1643/commits/29624c3030aaf5a3a29179b8686642cd55790aa3)

packages/adp-tooling/src/types.ts
```diff
@@ -138,3 +142,119 @@ export const enum HttpStatusCodes {
...
+    projectData: AdpProjectData;
+    timestamp: number;
+    componentUsageID: string;
+    componentName: string;
+    componentData: string;
+    componentSettings: string;
+    isLazy: string;
+    libraryReferenceIsLazy?: string;
+    shouldAddComponentLibrary: boolean;
+    componentLibraryReference: string;
```
tobiasqueck, line 198: Shouldn't `componentLibraryReference` be optional? I am assuming if `shouldAddComponentLibrary = false` then `componentLibraryReference` is `undefined`?
If my assumption is correct, then please remove `shouldAddComponentLibrary` and just do a `undefined` check within your coding.

packages/adp-tooling/src/types.ts
```diff
@@ -138,3 +142,119 @@ export const enum HttpStatusCodes {
...
+
+export interface AnnotationsData {
+    projectData: AdpProjectData;
+    timestamp: number;
+    annotationFileName?: string;
+    isInternalUsage: boolean;
+    oDataSource: string;
+    annotationFilePath: string;
+}
+export interface ComponentUsagesData {
```
tobiasqueck, line 188: Could you please consider structuring this a bit better instead of keeping it flat? I see two general properties, then a bunch for a component and and a bunch for a possible library.

packages/adp-tooling/src/types.ts
```diff
@@ -138,3 +142,119 @@ export const enum HttpStatusCodes {
...
+export interface IWriter<T> {
+    /**
+     * Writer function.
+     *
+     * @param {T} data - Data needed for the writer function.
+     */
+    write(data: T): Promise<void>;
+}
+
+export const enum ChangeType {
```
tobiasqueck, line 157: Please add a concise description to each new type, since not all of them are self explanatory.

tobiasqueck (changes requested):
* changeset is good
* code/functionality is straight forward and clear
* test coverage is great
* APIs/types are not self-explanatory and need more comments and ideally also a bit better structuring (comments inline)

packages/adp-tooling/src/types.ts
nikmace: Changed in: [706928d](https://github.com/SAP/open-ux-tools/pull/1643/commits/706928d0e145dee24ff454510c00cb9832238bec)

packages/adp-tooling/src/types.ts
nikmace: Changed in: [706928d](https://github.com/SAP/open-ux-tools/pull/1643/commits/706928d0e145dee24ff454510c00cb9832238bec)

packages/adp-tooling/src/types.ts
nikmace: Changed in: [706928d](https://github.com/SAP/open-ux-tools/pull/1643/commits/706928d0e145dee24ff454510c00cb9832238bec)

packages/adp-tooling/src/base/change-utils.ts
```diff
@@ -0,0 +1,199 @@
...
+ * If an additional subdirectory is specified, the change file is written there.
+ *
+ * @param {string} projectPath - The root path of the project.
+ * @param {object} change - The change data to be written to the file.
+ * @param {string} fileName - The name of the file to write the change data to.
+ * @param {Editor} fs - The `mem-fs-editor` instance used for file operations.
+ * @param {string} [dir] - An optional subdirectory within the 'changes' directory where the file will be written.
+ * @returns {void}
+ */
+export function writeChangeToFolder(projectPath: string, change: object, fileName: string, fs: Editor, dir = ''): void {
```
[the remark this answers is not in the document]
nikmace: Changed in: [706928d](https://github.com/SAP/open-ux-tools/pull/1643/commits/706928d0e145dee24ff454510c00cb9832238bec)

tobiasqueck (dismissed):
* previous positive review comments still apply
* previous change requests have been applied

testojs (approved):
- one small comment that should be checked
- great coverage
- not tested manually

packages/adp-tooling/src/base/change-utils.ts
```diff
@@ -0,0 +1,220 @@
...
+    dir = ''
+): void {
+    try {
+        let targetFolderPath = path.join(projectPath, FolderTypes.WEBAPP, FolderTypes.CHANGES);
+
+        if (dir) {
+            targetFolderPath = targetFolderPath.concat(`/${dir}`);
+        }
+
+        writeChangeToFile(`${targetFolderPath}/${fileName}`, change, fs);
```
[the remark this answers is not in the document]
nikmace: I've tested on both OS and it works as expected.

nikmace:
Added a small fix because noticed the generated change file had an extra slash in the namespace field.

tobiasqueck (dismissed):
re-approved

testojs (approved):
Looks even better now. Thanks!

tobiasqueck (approved):

PR #1655, opened 2024-02-09
feat: add i18n support

marufrasully:
Introduce a new package `@sap-ux/i18n` to support i18n. This package provides low level API and utilities functions.
`@sap-ux/project-access` package is enhanced to provide high level more convenient API to deal with i18n.


packages/i18n/src/index.ts
```diff
@@ -0,0 +1,45 @@
...
+    NewI18nEntry,
+    SapLongTextType,
+    SapShortTextType,
+    SapTextType,
+    ValueNode
+} from './types';
+
+export { csvToI18nBundle, jsonToI18nBundle, propertiesToI18nEntry } from './transformer';
+
+export { initI18n, i18n } from './i18n';
```
Klaus-Keller, line 34: Why are we exporting `initI18n` and `i18n` on root index.ts of this package? Is this not for local internationalization within this package?

packages/i18n/src/index.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import { initI18n } from './i18n';
+
+import { ToolsLogger } from '@sap-ux/logger';
+
+export const logger = new ToolsLogger();
```
Klaus-Keller, line 40: Same here, should we really export `logger` on public interface of this module?

packages/i18n/src/index.ts
marufrasully: Thanks. Addressed in this commit https://github.com/SAP/open-ux-tools/pull/1655/commits/03223e5fd8b2bbe324732d82a123cd2dcec3824b

packages/i18n/src/index.ts
marufrasully: see above.

Klaus-Keller (dismissed):
Thanks @marufrasully!
- test coverage is great
- did local build and test when I've added project-access support for it

Approved from my side.

packages/i18n/src/parser/utils/position.ts
```diff
@@ -0,0 +1,68 @@
...
+ * `a𐐀b` the character offset of the character `a` is 0, the character offset of `𐐀`
+ * is 1 and the character offset of b is 3 since `𐐀` is represented using two code
+ * units in UTF-16.
+ *
+ * Positions are line end character agnostic. So you can not specify a position that
+ * denotes `\r|\n` or `\n|` where `|` represents the character offset.
+ *
+ * @description copied from: node_modules\vscode-languageserver-types\lib\umd\main.d.ts. To keep source code lightweight since it depends only on `Range`
+ */
+export interface Position {
```
voicis, line 13: We should probably just add the dependency to `vscode-languageserver-types` like in `@sap-ux/odata-annotation-core-types`.

packages/i18n/test/unit/data/properties/unicode/lexer-errors.json
```diff
@@ -0,0 +1 @@
+[]
```
voicis, line 1: This file looks redundant.

packages/i18n/test/unit/data/properties/unicode/parse-errors.json
```diff
@@ -0,0 +1 @@
+[]
```
voicis, line 1: This file looks redundant.

packages/i18n/test/unit/read/cap/bundle.test.ts
```diff
@@ -0,0 +1,84 @@
...
+            throw new Error('should-throw-error');
+        });
+        const result = getCapI18nBundle(PROJECT_ROOT, env, [
+            join(PROJECT_ROOT, 'app', 'properties-csv', 'service.cds')
+        ]);
+        return expect(result).rejects.toThrowError('should-throw-error');
+    });
+
+    test('single .properties file - mem-fs-editor', async () => {
+        const readSpy = jest.fn().mockReturnValue('Product = Service Product');
```
voicis, line 78: Shouldn't we just use `mem-fs` here instead of mocking it?

packages/i18n/test/unit/scripts/update-csv.js
```diff
@@ -0,0 +1,4 @@
+const { update } = require("../../../dist/test/unit/helper/update-test-data");
+update('csv')
```
voicis, line 2: It looks very odd to have javascript files here that are referencing dist. Perhaps we can use the same approach as in vocabularies package? https://github.com/SAP/open-ux-tools/blob/feat/1i8n-support-02/packages/odata-vocabularies/package.json#L29

packages/i18n/web.js
```diff
@@ -0,0 +1,3 @@
+'use strict';
+
+module.exports = require('./dist/src/web');
```
voicis, line 3: Why not avoid additional indirections and not just import everything from `src/types` and also call the file `types`? Basically all of the APIs require file system and currently I don't see where we would need API's for browser environment other than types or constants.

packages/project-access/src/project/access.ts
```diff
@@ -0,0 +1,293 @@
...
+    ApplicationStructure
+} from '../types';
+import { getCapI18nFolderNames, getI18nBundles, getI18nPropertiesPaths } from './i18n';
+
+import {
+    createCapI18nEntries,
+    createManifestI18nEntries,
+    createUI5I18nEntries,
+    createAnnotationI18nEntries
+} from './i18n/write';
```
voicis, line 19: This should probably be imported from `./i18n` as well.

packages/i18n/test/unit/scripts/update-csv.js
marufrasully: let's make  scope, functionality and importance of this file. 

It is used to update test data for csv parser. Sure certainly we could use `ts-node`, but this mean adding another dependency. Looking at memory zie it take roughly `1,2 MB` and importance of this file. I tend not to add that module and keep file size of this module thin.

packages/i18n/src/parser/utils/position.ts
Klaus-Keller: > We should probably just add the dependency to `vscode-languageserver-types` like in `@sap-ux/odata-annotation-core-types`.

@voicis 
you mean `devDependency`?

voicis: No actual dependency as we use the factory functions like `Position.create`

packages/i18n/test/unit/scripts/update-csv.js
voicis: This would be just a dev dependency and would not affect the size of the package when installed and we already have it as a dependency in the repository. Also then we don't need to build the test folder (which we do not normally do in other modules either).

Klaus-Keller (approved):
Re-approve

voicis (approved):
Thanks @marufrasully!

- test coverage is great
- review comments addressed
- did local build and test

Approved from my side.

PR #1657, opened 2024-02-12
fix(backend-proxy): remove Partitioned from the cookies

zdravko-georgiev:
This PR removes the `Partitioned` property if it is present in a cookie. `Partitioned` needs to be used together with `Secure`. If `Secure` is missing then the cookie is considered invalid and it is not set.

### DIFF
--- a/.changeset/nasty-trainers-sort.md
+++ b/.changeset/nasty-trainers-sort.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/backend-proxy-middleware': patch
+---
+
+Remove 'Partitioned' from cookies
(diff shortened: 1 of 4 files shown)


Klaus-Keller (approved):
Thanks @zdravko-georgiev!
- updated regular expression looks good
- test coverage is great
- did no local build and test

Approved from my side.

tobiasqueck (approved):
* changes are clear and the tests confirm that it works as intended
* changeset is clear
* I don't understand why there is a change in the lock file but the change is also ok
* also approve the failed quality gate - the complexity for a regex is ok

Issue #1552, opened 2023-12-11
BUG - FPM Writer sets wrong namespace in TypeScript based Controller Extensions


c-kobo:
### Description
An issue had been reported about not found formatter functions in controller extensions for FE apps with a TypeScript project setup:
https://answers.sap.com/questions/14015781/use-formatter-in-fiori-elements-custom-section-of.html

The reason seems to lie with the wrong generation of the namespace, as the package name is being appended to it:
https://github.com/SAP/open-ux-tools/blob/45bf0515ec82c9a7d6749a49d5e1ea8d6ba0c527/packages/fe-fpm-writer/templates/controller-extension/Controller.ts#L5

The transpiled js representation of the TS controller file gets the package name added additionally, resulting in 'not found' issues with e.g. non resolved formatters paths in XML fragments:

<img width="1192" alt="image" src="https://github.com/SAP/open-ux-tools/assets/39270906/c075d505-a406-4ecb-9fdf-7765d8d03cbe">

### Solution

The fpm writer should either define
@namespace <%- ns %>
or 
@name <%- ns %>.<%- name %>

Both options will work.

c-kobo:
Just found that the fpm writer for the Custom Page Controller sets the namespace similarly wrong:
https://github.com/SAP/open-ux-tools/blob/45bf0515ec82c9a7d6749a49d5e1ea8d6ba0c527/packages/fe-fpm-writer/templates/page/custom/1.94/ext/Controller.ts#L4

815are:
@c-kobo 

thanks for pointing to issue and solutions.
I also checked controller of custom page and it looks little bit odd for me, current version generated for ts:
```
/**
 * @namespace project1.ext.view.Aaaaaaaa.controller
 */
export default class Aaaaaaaa extends Controller {
```
js compiled version:
```
var Aaaaaaaa = Controller.extend("project1.ext.view.Aaaaaaaa.controller.Aaaaaaaa", {
```

while file is under `webapp\ext\view\Aaaaaaaa.controller.ts`

But as I tested it works ok with view.xml + also works following variant:
```
/**
 * @namespace project1.ext.view.controller
 */
export default class Aaaaaaaa extends Controller {
```
compiled as:
```
var Aaaaaaaa = Controller.extend("project1.ext.view.controller.Aaaaaaaa", {
```

815are:
fix - https://github.com/SAP/open-ux-tools/pull/1661
[...]

PR #1661, opened 2024-02-13
fix(fpm-writer): use correct namespace when generating typescript controllers of controller extension and custom page controllers

815are:
Issue #1552

# Issue
Wrong namespace was created for new controller extension in typescript project:
```
/**
 * @namespace project1.ext.controller.Test
 * @controller
 */
export default class Test extends ControllerExtension<ExtensionAPI> {
```
is compiled as:
```
var Test = ControllerExtension.extend("project1.ext.controller.Test.Test", {
```

Both proposed solutions from Christian works:
```
Solution
The fpm writer should either define
@namespace <%- ns %>
or
@name <%- ns %>.<%- name %>
```

Additionally - I noticed that custom page's controller does not look optimal as well:
```
/**
 * @namespace project1.ext.view.Aaaaaaaa.controller
 */
export default class Aaaaaaaa extends Controller {
```
compiles as:
```
var Aaaaaaaa = Controller.extend("project1.ext.view.Aaaaaaaa.controller.Aaaaaaaa", {
```

If we use generator for js controller, then js version is generated following:
```
 return PageController.extend('project1.ext.view.Aaaaaaaa', {
```

## Result/Solution

In result, applying same solution with `@namespace` approach for both:
1. controller extensions
2. custom page's controllers

____
1. controller extension:
```
/**
 * @namespace project3.ext.controller
 * @controller
 */
export default class Test extends ControllerExtension<ExtensionAPI> {
```
compiled as:
```
var Test = ControllerExtension.extend("project3.ext.controller.Test", {
```
2. custom page's controller:
```
/**
 * @namespace project3.ext.view
 */
export default class Test extends Controller {
```
compiled as:
```
var Test = Controller.extend("project3.ext.view.Test", {
```

As tested with app - both works for me

### DIFF
--- a/.changeset/brown-nails-wonder.md
+++ b/.changeset/brown-nails-wonder.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fe-fpm-writer': patch
+---
+
+fix: use correct namespace when generating typescript controllers of controller extension and custom page controllers
--- a/packages/fe-fpm-writer/templates/controller-extension/Controller.ts
+++ b/packages/fe-fpm-writer/templates/controller-extension/Controller.ts
@@ -2,7 +2,7 @@ import ControllerExtension from 'sap/ui/core/mvc/ControllerExtension';
 import ExtensionAPI from 'sap/fe/<%- typeof extension === "object" ? `templates/${extension.pageType}` : "core" -%>/ExtensionAPI';
 
 /**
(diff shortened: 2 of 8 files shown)


tobiasqueck (approved):
* straight forward change
* tests confirm correctness
* changeset meaningful

c-kobo (approved):
Looks good to me.

PR #1663, opened 2024-02-13
feat: Optional ID parameter in page templates

broksy:
Added optional `ID` property to use in page manifest templates.


schreckstefan (dismissed):
The enhancement makes sense,  is compatible, and covered by tests

815are (commented):
`id` is consumed for object page and list report, but it is not cosumed for custom page ->
https://github.com/SAP/open-ux-tools/blob/feat/pageId/packages/fe-fpm-writer/templates/page/custom/1.94/manifest.json#L5

is it expected?

packages/fe-fpm-writer/test/unit/page/__snapshots__/custom.test.ts.snap
```diff
@@ -807,6 +851,48 @@ exports[`CustomPage generateCustomPage: different versions or target folder with
...
+        },
+        Object {
+          "name": "DummyPage",
+          "pattern": "RootEntity:?query:",
+          "target": "DummyPage",
+        },
+      ],
+      "targets": Object {
+        "DummyPage": Object {
+          "options": Object {
```
schreckstefan, line 880: id is missing here. Is this an error?

broksy: Test is for 1.84 manifest - https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/templates/page/custom/1.84/manifest.json template, it doesn't use `id` property.

schreckstefan (approved):
Tested OK, creation of custom pages works as before on root and subordinate level.

815are (changes requested):
Following looks missed:
1. Object page template - https://github.com/SAP/open-ux-tools/blob/feat/pageId/packages/fe-fpm-writer/templates/page/object/manifest.json#L20
`route` in source page still has old `<%- entity %>ObjectPage` even if `id` is passed
2. Custom page template - https://github.com/SAP/open-ux-tools/blob/feat/pageId/packages/fe-fpm-writer/templates/page/custom/1.94/manifest.json#L20
https://github.com/SAP/open-ux-tools/blob/feat/pageId/packages/fe-fpm-writer/templates/page/custom/1.84/manifest.json#L19
Same as in 1st point.
3. Please add tests where you add new object page and custom page with custom 'id' and with valid source page. Current snapshots does not cover it -> https://github.com/SAP/open-ux-tools/blob/feat/pageId/packages/fe-fpm-writer/test/unit/page/__snapshots__/object.test.ts.snap#L193-L194

815are (approved):
checked:
1. code changes enables feature to provide custom id for new object, list, custom pages;
2. new tests are added with and without source page;
3. change log is present

did not run manual tests

schreckstefan (approved):

Issue #1671, opened 2024-02-19
TBI - Parse XML response from `/sap/bc/adt/cts/transportchecks` in case of errors


cianmSAP:
### Description (include screenshots)
Rather than showing the raw XML in the case of errors, parse the XML and extract the messages to be displayed clearly as errors

### Value
It gives the user a clearer indication of the error and why the transport check has failed

### Architecture Elaboration
No

### Tasks
- [ ] Update `getTransportRequestList` in `/packages/axios-extension/src/abap/adt-catalog/services/transportcheck-service.ts` to facilitate logging error

PR #1672, opened 2024-02-19
tbi(axios-ext): log error msg for transport check xml response

cianmSAP:
#1671

Rather than showing the raw XML in the case of errors, parse the XML and extract the messages to be displayed clearly as errors


longieirl (approved):
Changes lookd good.

zdravko-georgiev (approved):
Code looks good
Very good code coverage
Did NOT tested manually

PR #1674, opened 2024-02-20
feat: Controller prefix for event handler

broksy:
Added `controllerPrefix` property to custom action event handlers. Use it when resolving event handler namespace.


815are (approved):
Checked:
1. code changes are valid - new property is introduced and handled to support for `.extension` prefix for custom actions.
2. new test is added with correct expectations
3. changelog is presented

PR #1678, opened 2024-02-20
fix(preview-middleware): add express in devDependencies

zdravko-georgiev:
Adds `express` in the `devDependencies`


Klaus-Keller (approved):
Adding devDependency to express looks good, approved from my side.

longieirl (approved):

PR #1680, opened 2024-02-21
Addition of Icon for App Info Page

aislingnoone:
Addition of Icon for App info Page. Needed for new dev efforts.

Screenshot 2024-02-21 at 11 14 17

### DIFF
--- a/.changeset/curvy-fans-reflect.md
+++ b/.changeset/curvy-fans-reflect.md
@@ -1,0 +1,5 @@
+---
+"@sap-ux/ui-components": patch
+---
+
+Addition of Icon for App Info Page. Needed for new dev efforts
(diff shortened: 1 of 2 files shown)


gpeuc (approved):

Klaus-Keller (commented):
Icon looks good, I just wonder why we talk about an App Info icon and in code we name it `ProjectValidation`? Those are two completely different things.

Klaus-Keller:
> @Klaus-Keller Because we are using this set of icons in the Validation section of AppInfo page.
> 
> Or at least I think that section can be called "Validation" section 😄 Maybe we can rename the set into "Validation" icons?
> 
> Screenshot 2024-02-21 at 13 19 45 Screenshot 2024-02-21 at 13 19 56

ok, then I agree it is ok to name it project validation.

Klaus-Keller (approved):
New icon looks good

PR #1681, opened 2024-02-22
feat(project-access): expose "toReferenceUri" in package interface

UweSchwerk:
needed for calculating `using` statements during cds code generation (part of new package "fiori-annotation-api")


Klaus-Keller (approved):
Exporting `toReferenceUri` is fine, approved

voicis (changes requested):
It still isn't exposed on package level. Changes in `src/index.ts` are also needed.

UweSchwerk:
> It still isn't exposed on package level. Changes in `src/index.ts` are also needed.

My bad - yes. Corrected

voicis (dismissed):
- simple change exposing `toReferenceUri`
- review comments addressed
- tested locally

Klaus-Keller (changes requested):
Hi @UweSchwerk,

Can we also change the import in the test to reflect that it should be exported from top level here:
https://github.com/SAP/open-ux-tools/blob/938989688353d1d29140b8cf59ad8aa59d447fb2/packages/project-access/test/project/cap.test.ts#L18

Just move `toReferenceUri` it to the `import {...} from '../../src';`

UweSchwerk:
> Hi @UweSchwerk,
> 
> Can we also change the import in the test to reflect that it should be exported from top level here:
> 
> https://github.com/SAP/open-ux-tools/blob/938989688353d1d29140b8cf59ad8aa59d447fb2/packages/project-access/test/project/cap.test.ts#L18
> 
> Just move `toReferenceUri` it to the `import {...} from '../../src';`

Done

voicis (approved):

Klaus-Keller (approved):
re-approve after exposing it on root `index.ts`

PR #1686, opened 2024-02-26
FIX: preview of adp projects with controller extensions written in typescript

tobiasqueck:
Instead of trying to handle the loading of local files in the preview middleware, just redirect to root so that the other UI5 middleware can handle the loading.

### DIFF
--- a/packages/adp-tooling/test/unit/preview/adp-preview.test.ts
+++ b/packages/adp-tooling/test/unit/preview/adp-preview.test.ts
@@ -197,15 +197,14 @@ describe('AdaptationProject', () => {
         });
 
         test('/local.file', async () => {
-            const testFileContent = '~test';
+            const localPath = '/local.file';
             mockProject.byGlob.mockResolvedValueOnce([
                 {
-                    getPath: () => '/local.file',
-                    getString: () => testFileContent
+                    getPath: () => localPath
                 }
             ]);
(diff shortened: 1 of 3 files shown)


packages/adp-tooling/test/unit/preview/adp-preview.test.ts
```diff
@@ -197,15 +197,14 @@ describe('AdaptationProject', () => {
         });
 
         test('/local.file', async () => {
-            const testFileContent = '~test';
+            const localPath = '/local.file';
```
zdravko-georgiev, line 200: This is not really testing the string replace. Maybe one additional test?

tobiasqueck: The test mocks the `byGlob` so I don't think two tests improve the quality, however, I have modified the one to cover the different file extensions scenario with https://github.com/SAP/open-ux-tools/pull/1686/commits/d40c3112ecbde5120f47f7fd336150338624a71b

zdravko-georgiev: Thanks!

zdravko-georgiev (approved):
Small and easy to understand
Review comments addressed
Excellent test coverage
Did NOT tested manually

nikmace (approved):
Small change
Great coverage
Did not test manually

Issue #1689, opened 2024-02-26
TBI - Add new module `@sap-ux/ui5-application-inquirer` to support UI5 app generation prompting


IainSAP:
### Description (include screenshots)

Provides functionality to prompt end users for the inputs necessary to generate a UI5 application.

### Value
Optional - details on how this technical backlog item provides value to end-users or dev team eg. Performance, Quality, Maintainability. This will help determine the ranking in relation to feature development

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

IainSAP:
Closing as done

PR #1690, opened 2024-02-26
tbi: Adds @sap-ux/ui5-application-writer

IainSAP:
https://github.com/SAP/open-ux-tools/issues/1689


tobiasqueck (changes requested):
Please move the changes of the `cap-config` to a separate PR since you change the API. It makes it easier to review its impact.

IainSAP:
> Please move the changes of the `cap-config` to a separate PR since you change the API. It makes it easier to review its impact.

I dont think the API has changed actually. I have overloaded the existing function.

tobiasqueck:
> > Please move the changes of the `cap-config` to a separate PR since you change the API. It makes it easier to review its impact.
> 
> I dont think the API has changed actually. I have overloaded the existing function.

You are correct, the existing calls won't break. Still, you have enhanced a module that is used in multiple places, so in the spirit of atomic changes and better review-ability, extract it.

PR #1691, opened 2024-02-27
fix: Alternative approach to have a consistent save for XML Fragments

tobiasqueck:
It currently contains two changes:

- generic hook to intercept flex operations by the preview-middleware on the file system
- XML fragments are not created anymore when the add xml dialog closes but on save
- moved the ADP specific change fixing from the `preview-middleware` to `adp-tooling`

In addition to this:

- **Add Fragment** and **Add Fragment at Extension Point** modals have been enhanced and now validate the fragment name also by checking the command stack. Clicking the create button only pushes the command for the specific change to the stack. When the user clicks save the **operation is intercepted** and if the type of change is **addXML** or **addXMLAtExtensionPoint** it will create an **XML file** in the workspace.


packages/adp-tooling/src/preview/change-handler.ts
```diff
@@ -0,0 +1,73 @@
...
+import { join } from 'path';
+import type { Logger } from '@sap-ux/logger';
+
+/**
+ * A mapping object that defines how to extract change content data from changes based on their type.
+ */
+export const moduleNameContentMap: { [key: string]: (change: CommonChangeProperties) => string } = {
+    codeExt: (change) => ((change as CodeExtChange).content?.codeRef ?? '').replace('.js', ''),
+    addXML: (change) => (change as AddXMLChange).content?.fragmentPath ?? ''
+} as const;
```
GDamyanov, line 13: Are we sure that change always has content?

packages/preview-middleware-client/src/adp/utils.ts
```diff
@@ -26,3 +35,28 @@ export function createDeferred<T>(): Deferred<T> {
...
+ * Checks if the fragment name associated with a command matches the specified fragment name.
+ *
+ * @param {FlexCommand} command - The command object containing the prepared change to be examined.
+ * @param {string} fragmentName - The name of the fragment to match against the command's change.
+ * @returns {boolean} Returns true if the command's change contains a fragment path that matches
+ *                    the specified fragment name; otherwise, returns false.
+ */
+export function matchesFragmentName(command: FlexCommand, fragmentName: string): boolean {
+    const change = command.getPreparedChange().getDefinition() as unknown as FragmentChange;
+    return change.content?.fragmentPath?.includes(`${fragmentName}.fragment.xml`) || false;
```
GDamyanov, line 49: Can we cast direct to FragmentChange?

GDamyanov (changes requested):

packages/preview-middleware-client/src/adp/utils.ts
nikmace: No, we cannot because getDefinition API returns a different type which is more generic for all of the different changes. But from this one I only need content.fragmentPath which obviously does not exist on other changes. The correct way here is to cast to unknown first and then to whatever we want the interface to be.

packages/adp-tooling/src/preview/change-handler.ts
nikmace: For both the `codeExt` and `addXML` functions within `moduleNameContentMap`, if `change.content` is `undefined`, the outcome will be an empty string (`''`). The optional chaining operator (`?.`) prevents a runtime error and returns `undefined`. If `undefined` is returned then `??` returns an empty string `''`. Then when `.replace('.js', '')` is called on the empty string, effectively leaving it **unchanged**.

GDamyanov (dismissed):
Code looks good. 
Good test coverage.

nikmace:
@tobiasqueck would you mind giving a review on this PR when you have time?

nikmace:
Functionality moved to #1824.

Issue #1417, opened 2023-11-02
TBI - Migrate Fiori tools telemetry lib to open source


slin-sap:
### Description (include screenshots)

Move telemetry lib to open source and allow modules in open source to consume it.

### Value

Can be consumed by Guided Answers
Can be consumed by generators moving to open source

### Architecture Elaboration
Already discussed

### Tasks
- [ ] Task 1

PR #1698, opened 2024-02-28
tbi(telemetry): update to init telemetry settings

cianmSAP:
#1417

Update to  `ToolsSuiteTelemetryInitSettings` interface to make the param `resourceId` optional

### DIFF
--- a/packages/telemetry/src/base/config-state.ts
+++ b/packages/telemetry/src/base/config-state.ts
@@ -6,7 +6,7 @@ import packageJson from '../../package.json';
  */
 export const TelemetrySettings = {
     // Target Azure apps insight destination
-    azureInstrumentationKey: '',
+    azureInstrumentationKey: process.env.AZURE_PROD_KEY ?? '',
     // Allow user to opt out from telemetry collection
     telemetryEnabled: true,
     // Module name in telemetry/pacakge.json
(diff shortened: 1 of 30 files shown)


packages/telemetry/src/base/config-state.ts
```diff
@@ -6,7 +6,7 @@ import packageJson from '../../package.json';
  */
 export const TelemetrySettings = {
     // Target Azure apps insight destination
-    azureInstrumentationKey: '',
+    azureInstrumentationKey: process.env.AZURE_PROD_KEY ?? '',
```
cianmSAP, line 9: Not sure whether this is required? @devinea

devinea: I think this could be useful.

cianmSAP: I removed this since I've reverted the change to remove resourceId from `ToolsSuiteTelemetryInitSettings`. It could be added via env var like in the example/index.ts

devinea (approved):
Making resourceId optional lgtm.
Code changes and changeset ✅

zdravko-georgiev (approved):
Change is well structured and easy to understand
Very good test coverage
Changeset is OK
Did NOT tested manually

PR #1702, opened 2024-02-28
fix(store): export getSecureStore

Klaus-Keller:
To consume store for all Fiori tools we need `getSecureStore()` to be exported on root level


devinea (approved):
lg

Klaus-Keller:
closing this, consumers need to be adjusted.

Issue #1703, opened 2024-02-29
BUG - Support VSCode theme color styling for `UIIconButton` property 'checked'


815are:
### Steps to Reproduce
1. Create `UIIconButton` with `checked=true`, like:
```
                    <UIIconButton
                        id="btn1"
                        iconProps={{ iconName: UiIcons.Undo }}
                        checked={true}
                        title="Undo" />
```
 2. It looks following:
 
![image](https://github.com/SAP/open-ux-tools/assets/90789422/b491c758-a0f5-43c5-96b5-725585676146)

Problem - hardoced colors without VSCode theme variables.

### Expected results
![image](https://github.com/SAP/open-ux-tools/assets/90789422/d87fbea0-d196-41ac-90db-1d42f03d4c53)

### Actual results
![image](https://github.com/SAP/open-ux-tools/assets/90789422/ab27084a-675b-4a06-a364-374611c9210f)

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1705, opened 2024-02-29
fix(ui-components): Support VSCode theme color styling for UIIconButton property 'checked'

815are:
Issue - #1703

When we use `UIIconButton` with `checked=true`, like:
```
                    <UIIconButton
                        id="btn1"
                        iconProps={{ iconName: UiIcons.Undo }}
                        checked={true}
                        title="Undo" />
```
 2. It looks following:
 
image

Problem is that colors are hardcoded from fluent-ui and does not support vscode themming.

After fix:
image

### DIFF
--- a/.changeset/weak-ladybugs-sparkle.md
+++ b/.changeset/weak-ladybugs-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui-components': patch
+---
+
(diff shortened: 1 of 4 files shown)


broksy (approved):
Change is easy to understand.
Did not test it manually.

Jimmy-Joseph19 (approved):
Changes looks good.
Covered with test 
Changeset found.
Approving
Did not test locally

PR #1708, opened 2024-03-01
fix(telemetry): assigning resource id in telemetry settings

cianmSAP:
Fix for assigning resource id in telemetry settings , introduced with https://github.com/SAP/open-ux-tools/pull/1698


mikicvi-SAP (approved):
Changes look good, changeset in place.

Issue #1706, opened 2024-02-29
FEATURE - Support HTML page to run OPA tests in the `preview-middleware`


tobiasqueck:
_As a **developer**,
I want to **use the `preview-middleware` for my OPA tests**,
so that I **don't have to maintain a test specific html file in my project.**_

### Description
We have the goal to remove as many as possible of the `html` files in the `webapp/test` folder and maintain them centrally in the `preview-middleware`. With this feature, we want to remove everything generic required for running OPA tests.

### Technical Design
* enhance the init script to also execute an additional script after initialization to start the OPA tests
* enhance the middleware configuration to additional offer an OPA test page
   * already keep in mind we might add other test frameworks in the future e.g. WDI5  
* create an OPA template (or something to be injected into the existing template) 

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a OPA tests in my project but no `opaTests.qunit.html` in my project folder
**when** I execute my test script
**then** the tests will be properly executed.

### Tasks
- [x] add option for a custom init script
- [x] add options for an OPA test file exposure
- [x] Unit Tests added
- [x] Test Cases defined

tobiasqueck:
Requested functionality was contributed by @heimwege with #1726!

PR #1726, opened 2024-03-05
Feat(preview middleware) add config for qunit/opa5 and custom oninit …

tobiasqueck:
Solves https://github.com/SAP/open-ux-tools/issues/1706

Add yaml config options for

- test frameworks (OPA5 | QUNIT)
- custom onInit script to be called after the sap-ui-bootstrap (as part of data-sap-ui-oninit)

Sample:

```
    - name: preview-middleware
      afterMiddleware: compression
      configuration:
        flp:
          path: localService/index.html
        test:
          - framework: "OPA5"
            path: /test/integration/opaTests.qunit.html
            init: /test/integration/opaTests.qunit
          - framework: "Qunit"
        debug: true
```


packages/preview-middleware/src/types/index.ts
```diff
@@ -53,6 +53,11 @@ export interface FlpConfig {
     libs?: boolean;
     apps: App[];
     theme?: string;
+    test?: {
+        framework?: string;
+    };
+    fioriSandboxConfig?: boolean;
```
tobiasqueck, line 59: It is not part of the corresponding feature request. Additionally, I'd challenge that it is necessary since you can define the same with the exist `apps` part of the middleware config.
using a custom FLP sandbox config is also bound to the existing FLP sandbox implementation, it might not be there in new versions anymore.

Please remove the `fioriSandboxConfig` property. We can have a separate discussion whether to add something like it in the future.

packages/preview-middleware-client/src/flp/init.ts
```diff
@@ -232,17 +244,30 @@ export async function init({ appUrls, flex }: { appUrls?: string | null; flex?:
         await registerComponentDependencyPaths(JSON.parse(appUrls));
     }
 
+    // Load custom initialization module
+    if (customInit) {
+        await sap.ui.require([customInit]);
```
tobiasqueck, line 249: Is the promise returning version of `sap.ui.require` available in UI5 `1.71.x`? That's the oldest version we support. If not, please use a callback like in line 232

packages/preview-middleware/src/types/index.ts
```diff
@@ -53,6 +53,11 @@ export interface FlpConfig {
     libs?: boolean;
     apps: App[];
     theme?: string;
+    test?: {
```
tobiasqueck, line 56: We also need to add an optional `test/path` property since the middleware needs to expose both a simple FLP for previews as well as additional ones for testing. I would also assume that test should become an array because I need multiple different test html files if I want to test with QUnit and with Opa.

heimwege: Up to now I just created multiple entries in the yaml for testing. Might not be the best approach but it worked 🙈 

```
    - name: preview-middleware
      afterMiddleware: compression
      configuration:
        flp:
          path: localService/index.html
        debug: true
    - name: preview-middleware
      afterMiddleware: compression
      configuration:
        flp:
          path: test/integration/opaTests.qunit.html
          customInit: com/sap/cap/fe/ts/sample/test/integration/opaTests.qunit
          test:
            framework: "OPA5"
        debug: true
    - name: preview-middleware
      afterMiddleware: compression
      configuration:
        flp:
          path: test/unit/unitTests.qunit.html
          customInit: com/sap/cap/fe/ts/sample/test/unit/unitTests.qunit
          test:
            framework: "QUNIT"
        debug: true
```
How to deal with test/path (framework) dependent customInit in case everything is part of one configuration? Add it to the `test` as well?
```
    - name: preview-middleware
      afterMiddleware: compression
      configuration:
        flp:
          path: localService/index.html
          customInit: com/sap/cap/fe/ts/sample/localService/myCustomInit
          test:
            - framework: "OPA5"
              path: test/integration/opaTests.qunit.html
              customInit: com/sap/cap/fe/ts/sample/test/integration/opaTests.qunit
            - framework: "QUNIT"
              path: test/unit/unitTests.qunit.html
              customInit: com/sap/cap/fe/ts/sample/test/unit/unitTests.qunit
        debug: true
```

packages/preview-middleware-client/src/flp/init.ts
heimwege: Seems [to be supported](https://sapui5.hana.ondemand.com/1.71.64/#/api/sap.ui/methods/sap.ui.require) 😅

packages/preview-middleware/src/types/index.ts
heimwege: Fine for me 👍🏻 will remove it and create a separate PR for it (and the discussions 😸).

packages/preview-middleware/src/types/index.ts
heimwege: Hm. I'm afraid this will get quite complex in case all those paths are part of one configuration.

Given the following configuration:

```
    - name: preview-middleware
      afterMiddleware: compression
      configuration:
        flp:
          path: localService/index.html
          customInit: com/sap/cap/fe/ts/sample/localService/myCustomInit
          test:
            - framework: "OPA5"
              path: test/integration/opaTests.qunit.html
              customInit: /test/integration/opaTests.qunit
            - framework: "QUNIT"
              path: test/unit/unitTests.qunit.html
              customInit: /test/unit/unitTests.qunit
        debug: true
```

will lead to
- flp.path -> adds a resource root (based on the app id) for `com.sap.cap.fe.ts.sample` path `..`
- flp.test.path -> defines its own prefixed resource root `open.ux.preview.test` (name is subject to change) path `../..` for the customInit to work, because `com.sap.cap.fe.ts.sample` is already in use and points to the "wrong" path (alternative: costomInit must always be located next to the html file from path)

So far so good. But now it gets worse

In case you use `paths` in `.tsconfig` it could look as follows:
- `"com/sap/cap/fe/ts/sample/*": ["./webapp/*"]`

Your `customInit` than can import based on the path:
`import("com/sap/cap/fe/ts/sample/test/unit/controller/CommentsSectionControllerTest")`

This will work at designtime and crash at runtime because the resource root for `com.sap.cap.fe.ts.sample` has been overwritten to `..`.

It would work if `open.ux.preview.test` is added to `.tsconfig` as path and used in the import statement of the `customInit`:
`import("open/ux/preview/test/test/unit/controller/CommentsSectionControllerTest")`

But this still can crash at runtime in case the tests themselves (`CommentsSectionControllerTest`) or any of their dependencies (`CommentsSectionController`) use imports based on the overwritten resource root path `com.sap.cap.fe.ts.sample`, e.g. `import { Message, Model, Ui } from "com/sap/cap/fe/ts/sample/ext/utils/Constants";`

Being not able to use `paths` in `.tsconfig` for preview-middleware to work seems quite annoying. @tobiasqueck do you have any idea how to work around this? (I did not yet commit these changes but I can of course do so if it helps)

The issue does not occur with multiple preview-middleware instances for each path because this ensures the resource root for the app id is always correct.

tobiasqueck: Thansk a ton for the feedback. I mainly tested the idea with js files but will add a ts project to my testing.
Let me tinker a little with it and then come back to you.

packages/preview-middleware-client/src/flp/init.ts
heimwege: I'm asking myself if we need to `await` this `sap.ui.require` at all. In difference to line 232 this is just an additional module that should be loaded. So far nobody is interested in any return value for further processing. So we could simply go for `sap.ui.require([customInit]);`

tobiasqueck: Agree, there is no need to wait, maybe we could log something if a custom script is loaded but not really

packages/preview-middleware/src/types/index.ts
heimwege: Yeah, sorry for the lengthy comment 🙈 
I just committed something that at least forks for my test setup. Might be a good idea to continue from there with further adjustments. Needed to slightly adjust the config (test.customInit needs to use the app id to match the resource root):

```
    - name: preview-middleware
      afterMiddleware: compression
      configuration:
        flp:
          path: localService/index.html
          test:
            - framework: "OPA5"
              path: test/integration/opaTests.qunit.html
              customInit: com/sap/cap/fe/ts/sample/test/integration/opaTests.qunit
            - framework: "QUNIT"
              path: test/unit/unitTests.qunit.html
              customInit: com/sap/cap/fe/ts/sample/test/unit/unitTests.qunit
        debug: true
```

heimwege: yes, agree

heimwege: actually... doing it the other way around seems to be much easier as the express router.get needs to be instantiated without ns whereas the module.require needs a ns. So as we anyway have have the ns at hand the user just needs to provide the relative path from webapp and we can do the rest. That fits well to the relative `path` property too. What do you think?

I just tested with
```
    - name: preview-middleware
      afterMiddleware: compression
      configuration:
        flp:
          path: localService/index.html
        test:
          - framework: "OPA5"
            path: /test/integration/opaTests.qunit.html
            init: /test/integration/opaTests.qunit
          - framework: "Qunit"
        debug: true
```
and it worked pretty well.

packages/preview-middleware/src/base/flp.ts
```diff
@@ -403,6 +415,56 @@ export class FlpSandbox {
...
+                        basePath: posix.relative(posix.dirname(config.path), '/') ?? '.',
+                        initPath: `${ns}${config.init.replace('.js', '')}`
+                    };
+                    const html = render(htmlTemplate, templateConfig);
+                    res.status(200).contentType('html').send(html);
+                }
+            }) as RequestHandler);
+            // add route for the init file
+            this.logger.debug(`Add route for ${config.init}`);
+            this.router.get(config.init, (async (_req, res, next) => {
```
heimwege, line 451: @tobiasqueck In case the test config has a custom `init: /test/integration/opaTests.qunit` then a route is being created to `/test/integration/opaTests.qunit` that is never used because the file `/test/integration/opaTests.qunit.js` is being loaded from the file system. Nevertheless there's also this log message `debug preview-middleware Add route for /test/integration/opaTests.qunit` that is kind of misleading. Should we skip the whole thing in case a custom `init` is given (my preference) or enhance the custom init by `.js` so that the route can do the skip (line 456)?

tobiasqueck: I see you decided already and added
```js
if (testConfig.init !== undefined) {
       this.logger.debug(`Skip serving test init script in favor of provided script: ${testConfig.init}`);
        continue;
}
```
Good with me :)

tobiasqueck (commented):
* changeset is good
* readme update properly
* test locally with and without typescript, with custom scripts, with custom path, pattern, ... basically everything I could think of and it worked like a charm!
* proper test coverage
* code is well structured, easy to understand and well commented

Klaus-Keller (approved):
Thanks @tobiasqueck, 
- code is well structured
- test coverage is great
- did a visual review, not tested manually

Approved from my side.

heimwege (approved):

zdravko-georgiev (approved):
Change is well structured and easy to understand
Very good test coverage
Did NOT tested manually

packages/preview-middleware/templates/test/qunit.js
```diff
@@ -0,0 +1,6 @@
+sap.ui.define([
+    'sap/ui/thirdparty/qunit-2'
+], function() {
+        sap.ui.require(<%- JSON.stringify(tests) %>);
```
matz3, line 4: There's a chance to run into a race condition when QUnit is loaded and already looks for tests, but the test modules are not completely loaded, yet. It might cause that none or not all tests are run.

To prevent such scenarios, the code should be adjusted like this (untested):
```js
sap.ui.define([
    'sap/ui/thirdparty/qunit-2'
], function(QUnit) {
    QUnit.config.autostart = false;
    sap.ui.require(<%- JSON.stringify(tests) %>, function() {
        QUnit.start();
    });
});

heimwege: Fill be fixed with https://github.com/SAP/open-ux-tools/pull/1813

PR #1728, opened 2024-03-06
fix: do not trigger change property when no changes made

vinayhospete:
CPE: Do not create new change on leaving input field without making any change


marufrasully (dismissed):
LGTM

Simple code change and covered with  unit test. Did not test it locally.

packages/control-property-editor/test/unit/panels/properties/StringEditor.test.tsx
```diff
@@ -85,6 +86,27 @@ describe('StringEditor', () => {
...
+        const value = '1234';
+        const props: any = {
+            icons: [],
+            isEnabled: false,
+            isIcon: false,
+            name: 'testProperty',
+            value,
+            type: 'integer'
+        };
+        const spyChangeProperty = jest.spyOn(slice, 'changeProperty');
```
voicis, line 100: Please use the `dispatch` spy like in https://github.com/SAP/open-ux-tools/blob/50515e1c75fb1e10550a0ffeb2f718a390c82910/packages/control-property-editor/test/unit/panels/properties/Funnel.test.tsx#L20-L32 instead of spying of `slice`.

voicis (changes requested):

packages/control-property-editor/test/unit/panels/properties/StringEditor.test.tsx
vinayhospete: Done

voicis (changes requested):
Please add preview-middleware to the change set.

marufrasully (approved):
reapproving it.

PR #1732, opened 2024-03-07
feat(reload-middleware): expose the livereload port as env variable

zdravko-georgiev:
Sets the livereload port as env variable after the livereload server is started


tobiasqueck (approved):
* changeset ok
* change straight forward and required

voicis (approved):
- change set ok
- straight forward change
- good test coverage

PR #1742, opened 2024-03-12
chore(telemetry): remove unused dependency to @sap/ux-specification

Klaus-Keller:
There is a unused dependency to `@sap/ux-specification` in `@sap-ux/telemetry`. This pull request removes it.


devinea (approved):
Good catch. Thanks

tobiasqueck (approved):

PR #1743, opened 2024-03-12
fix: text input focus out do not trigger empty change

vinayhospete:
text input focus out do not trigger empty change


voicis (approved):
- Simple code change
- Change set looks good

marufrasully (approved):
LGTM

Issue #1744, opened 2024-03-12
BUG - Using inbound navigation writer to generate appropriate change does not produce correct results


nikmace:
### Description
If you create a new inbound change and when trying to update a property with the same inbound id, instead of updating the current file it will create a new change file.

### Steps to Reproduce
1. Use `generateChange` for Inbound Change creation,
2. Create change with new inbound id, then create another change with the same inbound id
3. See that two files are created instead of one being updated

### Expected results
Existing change file is updated

### Actual results
Two files are created, the change file with the same inbound id is not updated.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1745, opened 2024-03-12
fix: Using inbound navigation writer to generate appropriate change does not produce correct results

nikmace:
Fix for #1744.

- Changes the logic in `findChangeWithInboundId` searching for a change file with a specific inbound ID within a project's change directory. Now it checks all the inbound change files that are filtered and looks if the inbound ID is matched.

### DIFF
--- a/packages/adp-tooling/src/base/change-utils.ts
+++ b/packages/adp-tooling/src/base/change-utils.ts
@@ -176,6 +176,7 @@ export function findChangeWithInboundId(projectPath: string, inboundId: string):
             if (change.content?.inboundId === inboundId) {
                 changeObj = change;
                 filePath = pathToFile;
+                break;
             }
         }
 
--- a/packages/adp-tooling/src/base/change-utils.ts
+++ b/packages/adp-tooling/src/base/change-utils.ts
@@ -165,11 +165,11 @@ export function findChangeWithInboundId(projectPath: string, inboundId: string):
     }
 
     try {
-        const file: Dirent | undefined = readdirSync(pathToInboundChangeFiles, { withFileTypes: true }).find(
-            (dirent: Dirent) => dirent.isFile() && dirent.name.includes('changeInbound')
+        const files: Dirent[] = readdirSync(pathToInboundChangeFiles, { withFileTypes: true }).filter(
+            (dirent) => dirent.isFile() && dirent.name.includes('changeInbound')
         );
 
-        if (file) {
+        for (const file of files) {
             const pathToFile = path.join(pathToInboundChangeFiles, file.name);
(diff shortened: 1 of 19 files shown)


IvoSG (dismissed):
Simple and easy to understand change

zdravko-georgiev (commented):
@nikmace No need to adapt an unit test? I doubt that 😉

nikmace:
> @nikmace No need to adapt an unit test? I doubt that 😉

Checked the tests. The coverage on this method is 100%. The way they are written does not require adapting the tests.

packages/adp-tooling/src/base/change-utils.ts
```diff
@@ -165,11 +165,11 @@ export function findChangeWithInboundId(projectPath: string, inboundId: string):
     }
 
     try {
-        const file: Dirent | undefined = readdirSync(pathToInboundChangeFiles, { withFileTypes: true }).find(
-            (dirent: Dirent) => dirent.isFile() && dirent.name.includes('changeInbound')
+        const files: Dirent[] = readdirSync(pathToInboundChangeFiles, { withFileTypes: true }).filter(
+            (dirent) => dirent.isFile() && dirent.name.includes('changeInbound')
         );
```
GDamyanov, line 170: Can we use this approach for getting the needed file e.g something like this:
`const file = readdirSync(pathToInboundChangeFiles, { withFileTypes: true }).filter(
			(dirent) => dirent.isFile() && dirent.name.includes('changeInbound')).map(function(file) {
				const filePath = join(pathToInboundChangeFiles, file.name);
				return {
					changeObject: JSON.parse(readFileSync(filePath, 'utf-8')),
					filePath: filePath
				}
			}).find((element) => element.changeObject?.change?.inboundId === inboundId);`

nikmace: Can you format this please?

nikmace: I am against this approach because it will make it less readable, maintainable, and potentially less efficient.

nikmace: Added a break statement here performance: [dbb6980](https://github.com/SAP/open-ux-tools/pull/1745/commits/dbb698046df6ab7a9981aa070b00889e08ac8635)

GDamyanov (dismissed):
Code looks good.
Good test coverage.
Comments addressed.

zdravko-georgiev (approved):
Simple and straightforward change
Excellent test coverage
Did NOT tested manually

GDamyanov (approved):
Simple and easy to understand change.
Good test coverage.
Did NOT tested manually.

PR #1748, opened 2024-03-13
fix(project-access): remove setting of cds.root

zdravko-georgiev:
Removes the setting of `cds.root`, which was introduced in https://github.com/SAP/open-ux-tools/pull/1563. Setting of `cds.root` has unforeseen side effects when using `require.resolve`. Instead set the project root when calling `cds.load`.

### DIFF
--- a/packages/project-access/src/project/cap.ts
+++ b/packages/project-access/src/project/cap.ts
@@ -143,10 +143,6 @@ export async function getCapModelAndServices(
     }
 
     const cds = await loadCdsModuleFromProject(_projectRoot);
-
-    _logger?.info(`@sap-ux/project-access:getCapModelAndServices - Using 'cds.home': ${cds.home}`);
-    _logger?.info(`@sap-ux/project-access:getCapModelAndServices - Using 'cds.version': ${cds.version}`);
-
     const capProjectPaths = await getCapCustomPaths(_projectRoot);
     const modelPaths = [
         join(_projectRoot, capProjectPaths.app),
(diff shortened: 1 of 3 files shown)


packages/project-access/src/project/cap.ts
```diff
@@ -404,8 +403,7 @@ async function loadCdsModuleFromProject(capProjectPath: string): Promise<CdsFaca
     if (global) {
         global.cds = cds;
     }
-    // Ensure we use a known root path, otherwise `cwd` is used which varies between invocations.
-    cds.root = capProjectPath;
```
devinea, line 408: @IainSAP Please review.

packages/project-access/src/project/cap.ts
```diff
@@ -146,15 +146,14 @@ export async function getCapModelAndServices(
...
-    _logger?.info(`@sap-ux/project-access:getCapModelAndServices - Using 'cds.root': ${cds.root}`);
 
     const capProjectPaths = await getCapCustomPaths(_projectRoot);
     const modelPaths = [
         join(_projectRoot, capProjectPaths.app),
         join(_projectRoot, capProjectPaths.srv),
         join(_projectRoot, capProjectPaths.db)
     ];
-    const model = await cds.load(modelPaths);
+    const model = await cds.load(modelPaths, { root: _projectRoot });
```
devinea, line 156: is this supported only for recent cds versions? Do we need a check here.

zdravko-georgiev: Good question. Need to check with Christian Georgi. @chgeo could you please comment?

chgeo: The `root` property has been there for a long time now.

devinea: Thanks @chgeo

IainSAP (approved):
Thanks @zdravko-georgiev 

- So long as services paths returned from the CDS model are still relative to the root path then this change should be ok
- Did not test, but @zdravko-georgiev ran the consuming tests and reported that they still pass

packages/project-access/src/project/cap.ts
IainSAP: Should be OK as long as `cds.load(.., capProjectPath)` behaves the same and sets the root internally

PR #1753, opened 2024-03-15
Upgrading version of `sanitize-html` due to medium priority security issue

tobiasqueck:
Fixes parts of #1754


815are (dismissed):
`sanitize-html` is used only in one place - for helper function to render tooltip's content by passing html string.
Tested with storybook - scenario works ok for me - https://github.com/SAP/open-ux-tools/blob/main/packages/ui-components/stories/UITooltip.story.tsx#L22

815are (approved):

zdravko-georgiev (approved):

Issue #1666, opened 2024-02-19
BUG - generated FLP sandbox does not show application title and description


Klaus-Keller:
### Description
When previewing an application in generated Fiori Launchpad sandbox (`flp.html`), the application title and description is not displayed correctly. Instead, `{` and `}` is shown.

### Steps to Reproduce
1. Download and extract [fe_lrop_v4_odata_none-2024-02-190928.zip](https://github.com/SAP/open-ux-tools/files/14329178/fe_lrop_v4_odata_none-2024-02-190928.zip)
2. Start a terminal in extracted folder `fe_lrop_v4_odata_none-2024-02-190928`
3. Execute `npm install`
4. Execute `npm start`
5. A browser starts, change `flpSandbox` to `flp` in the URL, the URL should look like 

### Expected results
FLP sandbox is shown with the application title and description

### Actual results
Instead of title and description `{` and `}` is shown.

### Screenshots
<img width="628" alt="image" src="https://github.com/SAP/open-ux-tools/assets/66327622/f0f3c7b4-8da9-493b-8d04-02a598dcb6ad">

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
Most likely the issue is due the way application and title are generated in the flp sandbox: 
<img width="768" alt="image" src="https://github.com/SAP/open-ux-tools/assets/66327622/df6e4a13-1f38-4992-949b-f0045358f2c4">

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

heimwege:
solved with #1755

PR #1755, opened 2024-03-15
fix/1666/missing title and description on launchpad tile

heimwege:
solves https://github.com/SAP/open-ux-tools/issues/1666


packages/preview-middleware/src/base/flp.ts
```diff
@@ -475,23 +476,81 @@ export class FlpSandbox {
...
+     *
+     * @param manifest application manifest
+     * @param projectRoot absolute path to the project root
+     * @returns title of the application
+     * @private
+     */
+    private async getAppTitle(manifest: Manifest, projectRoot: string): Promise<string> {
+        const bundle = await this.getPropertiesI18nBundle(projectRoot);
+        const title = manifest['sap.app'].title;
+        const titleI18nKey = title?.substring(2, title.length - 2);
```
tobiasqueck, line 528: Wouldn't it be better to use a regex here? It would allow to detect whether it is a key or a plain text and if it is get the value

packages/preview-middleware/src/base/flp.ts
```diff
@@ -475,23 +477,81 @@ export class FlpSandbox {
...
+
+    /**
+     * Get the description of the application from the i18n properties file.
+     *
+     * @param manifest application manifest
+     * @param projectRoot absolute path to the project root
+     * @returns description of the application
+     * @private
+     */
+    private async getAppDescription(manifest: Manifest, projectRoot: string): Promise<string> {
```
tobiasqueck, line 544: `getAppTitle` and `getAppDescription` look pretty identical. Could we put this together into one function and have `manifest['sap.app'].description` or `manifest['sap.app'].title` as input

tobiasqueck (changes requested):
Two small comments, otherwise, please add a changeset (call `pnpm changeset` in the root of the project)

packages/preview-middleware/src/base/flp.ts
heimwege: Good idea 👍🏻 Done

packages/preview-middleware/src/base/flp.ts
heimwege: Not sure if it's better that way but I introduced a regex 🐱

heimwege: ...which does not make sense. Need to refactor this 🙈

heimwege:
@tobiasqueck how to handle the two `todo`s that will not be resolved with this PR? Delete them from the code or keep them as reference for later?

tobiasqueck:
> @tobiasqueck how to handle the two `todo`s that will not be resolved with this PR? Delete them from the code or keep them as reference for later?

keep them in there, please. The fix is great even without them completed but I'd like to keep them as a reference for a future improvement.

tobiasqueck (approved):
* changeset ok
* local test successful
* code is straight forward and clear
* test coverage is good
* the two sonar bugs are ok to keep a reference to a possible future improvement

packages/preview-middleware/src/base/flp.ts
```diff
@@ -481,23 +483,64 @@ export class FlpSandbox {
...
+     * Get the i18n properties bundle.
+     *
+     * @param projectRoot absolute path to the project root
+     * @returns i18n properties bundle
+     * @private
+     */
+    private async getPropertiesI18nBundle(projectRoot: string) {
+        //todo: consider relative path from manifest['sap.app'].i18n
+        //todo: consider locale
+        const i18nFilePath = join(projectRoot, 'webapp', 'i18n', 'i18n.properties');
```
Klaus-Keller, line 513: this would only work for projects that have not set a custom `webapp` path. Module `@sap-ux/project-provider` provides convenient methods for retrieving the i18n bundles and also consider custom webapp paths, see https://github.com/SAP/open-ux-tools/blob/edc3a166650d2d3e1981372cd07b3ba1811b40d8/packages/project-access/src/project/access.ts#L143

Otherwise `@sap-ux/project-provider` has also a function to get custom webapp paths:
https://github.com/SAP/open-ux-tools/blob/edc3a166650d2d3e1981372cd07b3ba1811b40d8/packages/project-access/src/project/ui5-config.ts#L14-L26

Klaus-Keller (changes requested):
Please see my comment about projects that have custom `webapp` path config

packages/preview-middleware/src/base/flp.ts
heimwege: Hi @Klaus-Keller and thanks for the review

`@sap-ux/project-provider` even deals with all variants of [JSONSchemaForSAPAPPNamespace.i18n](https://github.com/SAP/ui5-manifest/blob/main/types/manifest.d.ts#L803) 👍🏻 

I'll adjust the code.

Are there any plans to support loading the `locale` version of the `i18n.properties`? Then I could get rid of the todo here as well 😸

Klaus-Keller (approved):
Thanks @heimwege!
- code changes are well done and easy to understand
- code coverage is great
- ran a local test and did not find issues

Approved from my side

johannes-kolbe (approved):
Thanks @heimwege, lgtm
- changeset provided
- straightforward changes
- unit tests provided
- manual test ok

Issue #1763, opened 2024-03-21
FEATURE - Allow adding tiles pointing to remotely available apps in the preview-middleware


tobiasqueck:
_As a **developer**,
I want to **add existing applications to my FLP preview**,
so that I **can test app to app navigation**_

### Description
When testing my application in the FLP preview, I also want to test app-to-app navigation. Assuming I am using also the `backend-proxy-middleware` my app would be available, and I just want to add it as tile to the FLP preview with a configuration like the one below (which doesn't point to a backend but a sample app included in UI5).

```yaml
apps
  - target: /test-resources/sap/ushell/demoapps/AppNavSample
    componentId: sap.ushell.demo.AppNavSample
    intent: 
      object: TheOther
      action: preview
```

### Technical Design
Enhance the support for additional tiles to also allow adding apps not available locally but remotely.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a project with the preview-middleware and backend-proxy-middleware
**when** I add a config like the one in description
**then** I have a tile that points to my remote application.

### Tasks
- [x] Implement the feature
- [x] Unit Tests added
- [x] Test Cases defined

PR #1764, opened 2024-03-21
Allow adding tiles pointing to remotely available apps in the preview-middleware

tobiasqueck:
Implements: #1763

### DIFF
--- a/packages/preview-middleware/README.md
+++ b/packages/preview-middleware/README.md
@@ -34,14 +34,13 @@ When this middleware is used together with the `reload-middleware`, then the ord
 
 ### `flp.apps`
 Array of additional application configurations:
-| Option          | Type     | Default Value  | Description                                                                                          |
-| --------------- | -------- | -------------- | ---------------------------------------------------------------------------------------------------- |
-| `target`        | `string` |                | Target path of the additional application                                                            |
-| `local`         | `string` |  `undefined`   | Optional local path to the folder containing the application, if not provided then a `componentId` is required |
-| `componentId`   | `string` |  `undefined`   | Optional `componentId` of the remote app if no local path is provided                                |
-| `intent`        |          |                | Optional intent to be used for the application                                                       |
-| `intent.object` | `string` | `(calculated)` | Optional intent object, if it is not provided then it will be calculated based on the application id |
-| `intent.action` | `string` | `preview`      | Optional intent action                                                                               |
+| Option                   | Type     | Default Value  | Description                                                                                                   |
+| ------------------------ | -------- | -------------- | ------------------------------------------------------------------------------------------------------------- |
+| `target`                 | `string` |                | Target path of the additional application                                                                     |
+| `local` ot `componentId` | `string` |                | Either a local path to a folder containing the application or the `componentId` of a remote app is required   |
+| `intent`                 |          |                | Optional intent to be used for the application                                                                |
+| `intent.object`          | `string` | `(calculated)` | Optional intent object, if it is not provided then it will be calculated based on the application id          |
+| `intent.action`          | `string` | `preview`      | Optional intent action                                                                                        |
 
 ### `adp.target`
 | Option        | Type                           | Description                                                                                                                                     |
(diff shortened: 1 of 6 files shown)


packages/preview-middleware/README.md
```diff
@@ -37,7 +37,8 @@ Array of additional application configurations:
 | Option          | Type     | Default Value  | Description                                                                                          |
 | --------------- | -------- | -------------- | ---------------------------------------------------------------------------------------------------- |
 | `target`        | `string` |                | Target path of the additional application                                                            |
-| `local`         | `string` |                | Local path to the folder containing the application                                                  |
+| `local`         | `string` |  `undefined`   | Optional local path to the folder containing the application, if not provided then a `componentId` is required |
+| `componentId`   | `string` |  `undefined`   | Optional `componentId` of the remote app if no local path is provided                                |
```
voicis, line 41: I suppose the goal was to be backwards compatible, but the options and their descriptions are now rather confusing. Using `optional` together with `required` for two options where exactly one should be defined is also misleading and a potential usability issue.

|validity| `local` | `componentId` |  
|---| ------- | ---------------|
|invalid | undefined |  undefined |
|valid | undefined | defined |
|valid | defined |  undefined |
|valid (`componentId` is ignored)| defined | defined |

packages/preview-middleware/src/base/flp.ts
```diff
@@ -344,11 +344,24 @@ export class FlpSandbox {
...
+                        id: app.componentId,
+                        title: app.intent ? `${app.intent.object}-${app.intent.action}` : app.componentId
+                    }
+                } as Manifest;
+            }
+            if (manifest) {
+                this.addApp(manifest, app);
+                this.logger.info(`Adding additional intent: ${app.intent?.object}-${app.intent?.action}`);
+            } else {
+                this.logger.info(`Invalid application config: ${app}`);
```
voicis, line 364: I got this output:
```
info preview-middleware Invalid application config: [object Object]
```
The log could be a bit more helpful at indicating which app is causing issues. Maybe also explicitly stating that `local` or `componentId` must be defined.

tobiasqueck: Fixed with https://github.com/SAP/open-ux-tools/pull/1764/commits/e58caa2a94a017a6792b48087a6fd71f994f1712.

packages/preview-middleware/README.md
tobiasqueck: @voicis how about https://github.com/SAP/open-ux-tools/pull/1764/commits/f5a036eea12644314b8f1b0d276ca72f1d89649b? Is that better?

tobiasqueck:
> I tested it on a FE project and it worked nicely. @tobiasqueck Is it also supposed to work on ADP projects?

Yes. I haven't tested it explicitly but the logic is independent of whether it is an app project or an ADP

packages/preview-middleware/README.md
```diff
@@ -34,13 +34,13 @@ When this middleware is used together with the `reload-middleware`, then the ord
...
-| `intent`        |          |                | Optional intent to be used for the application                                                       |
-| `intent.object` | `string` | `(calculated)` | Optional intent object, if it is not provided then it will be calculated based on the application id |
-| `intent.action` | `string` | `preview`      | Optional intent action                                                                               |
+| Option                   | Type     | Default Value  | Description                                                                                                   |
+| ------------------------ | -------- | -------------- | ------------------------------------------------------------------------------------------------------------- |
+| `target`                 | `string` |                | Target path of the additional application                                                                     |
+| `local` ot `componentId` | `string` |                | Either a local path to a folder containing the application or the `componentId` of a remote app is required   |
```
voicis, line 40: This looks like a typo.

```suggestion
| `local` or `componentId` | `string` |                | Either a local path to a folder containing the application or the `componentId` of a remote app is required   |
```

packages/preview-middleware/README.md
voicis: Yes, this looks a lot better now. It is clear now from reading the documentation that one of the options must be used and together with the error message pointing out what is wrong, it should be a sufficient solution here.

voicis (approved):
- Change set present
- Code is straight forward and easy to understand
- Review comments addressed
- Good test coverage
- Tested locally

mikicvi-SAP (approved):
- Changeset in place.
- Documentation updated and makes sense.
- Code is simple and easy to follow.
- Good test coverage

PR #1765, opened 2024-03-21
export extendAdditionalMessages from inquirer common module

cianmSAP:
Export extendAdditionalMessages from inquirer common module


IainSAP (approved):
Thanks @cianmSAP 

- Moves function to allow re-use, not changed, as expected
- New tests added to cover previously not exported function, as expected

mikicvi-SAP (approved):
Changeset in place, good coverage, tests make sense, code didn't change.

PR #1771, opened 2024-03-22
fix(ui5-application-inquirer): Bug fix #1770 - use available ui5 versions only

### LINKED ISSUE #1770: BUG - `@sap-ux/ui5-application-inquirer` sets the default ui5 version choice as new entry but not yet supported by middleware
### Description

When passing a default choice UI5 version as a prompt option to `@sap-ux/ui5-application-inquirer` `prompt()` this version is added to the top of the list of UI5 versions. However, this version may not be available from CDN (`https://ui5.sap.com`) and would cause the generated app preview to fail. Until the preview middleware can be updated to directly load UI5 resources from a different URL (target system) where unreleased versions may be hosted, this behaviour should be reverted to firstly check that the version already exists in the UI5 version list (either from cdn or from fallbacks where offline). 

### Screenshots

OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

IainSAP:
Fix for #1770 

- Adds new function `getDefaultUI5VersionChoice` which will return the closest match from a known list of ui5 versions
- Updates `@sap-ux/ui5-application-inquirer` to use the new function, instead of adding anyway (system versions may not be available generally)
- Adds/updates tests to cover

### DIFF
--- a/.changeset/green-windows-yell.md
+++ b/.changeset/green-windows-yell.md
@@ -1,0 +1,6 @@
+---
+'@sap-ux/ui5-application-inquirer': minor
+'@sap-ux/inquirer-common': patch
+---
+
+When passing a default UI5 version its availability will be verified before offering as a choice
--- a/packages/inquirer-common/package.json
+++ b/packages/inquirer-common/package.json
@@ -34,11 +34,13 @@
         "fuzzy": "0.1.3",
         "i18next": "23.5.1",
         "chalk": "4.1.2",
-        "figures": "3.2.0"
+        "figures": "3.2.0",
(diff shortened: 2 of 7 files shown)


cianmSAP (approved):
Updates and coverage is good
Have not tested locally but basing of tests
Changeset
✅

devinea (approved):
Code changes and tests lg.
changeset ✅

cianmSAP (approved):
Re-approving

Issue #1735, opened 2024-03-08
BUG -  Generation of sap-fe-mockserver middleware configuration files, does not append annotations to the configuration


martinjanus-sap:
### Description

When generating a Fiori Elements application, multiple yaml files are generated. In those yaml files, the sap-fe-mockserver is configured. Unfortunately, in the configuration the annotations are configured with an empty array.

This empty array creates an issue if you run the local karma or do run the azure pipeline. The mockserver does not find the annotations, which results e.g. in a empty table. This effect cannot be seen running OPA5 tests in Business Application Studio. This makes the life from an application developer hard.

### Solution 
Please add the configuration to the annotation.xml when generating according to the following documentation.

https://github.com/SAP/open-ux-odata/tree/main/packages/ui5-middleware-fe-mockserver#usage

### Expected results
Example Correct Entry: 
```
- name: sap-fe-mockserver
  mountPath: /
  beforeMiddleware: csp
  configuration:
    annotations:
      - localPath: './webapp/localService/UI_MANAGEREPAIROBJECTS_VAN.xml'
        urlPath: '/sap/opu/odata/IWFND/CATALOGSERVICE;v=2/Annotations*'
    services:
      - urlPath: '/sap/opu/odata/sap/UI_MANAGEREPAIROBJECTS'
        metadataPath: './webapp/localService/metadata.xml'
        mockdataPath: './webapp/localService/data'
```

### Actual results
```
    - name: sap-fe-mockserver
      beforeMiddleware: csp
      configuration:
        mountPath: /
        services:
          - urlPath: /sap/opu/odata/sap/SEPMRA_PROD_MAN
            metadataPath: ./webapp/localService/metadata.xml
            mockdataPath: ./webapp/localService/data
            generateMockData: true
        annotations: []
```

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [X] Mac OS
- [X] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Issue found when testing, if a `ui5-mock.yaml` already exists, the annotations are not being appended when the command the following command is executed;
```
npx --yes @sap-ux/create@latest add mockserver-config
```

[...]

PR #1772, opened 2024-03-22
fix(mock-server-config): Updates ui5-mock.yaml with annotations

martinjanus-sap:
#1735

If ui5-mock.yaml exists, updates existing ui5-mock.yaml config with annotations. 
If ui5-mock.yaml doesn't exist but ui5.yaml exists, creates it based on ui5.yaml and updates with annotations

### DIFF
--- a/.changeset/kind-owls-act.md
+++ b/.changeset/kind-owls-act.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/mockserver-config-writer': minor
+---
+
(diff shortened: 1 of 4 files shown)


longieirl (approved):
Changes were validated locally during chat session.
Test coverage is good.
Changes are good.

unseen1980 (approved):
I haven't tested it locally, but the code looks ok, just passing annotationsConfig as a param. Code coverage is ok too. Thanks

PR #1774, opened 2024-03-25
fix: revert cds root set

815are:
Partially reverts SAP/open-ux-tools#1748

It was reported that #1748 causes follow up issues regarding `cds.root` - reason of revert.

Additionally we found reason of why there was error of failing `require.resolve` - it was not issue on `open-ux-side`.
We had call for `require.resolve` outside of `open-ux-tools` source.

### DIFF
--- a/.changeset/dry-spiders-warn.md
+++ b/.changeset/dry-spiders-warn.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': patch
+---
+
+Revert setting of cds.root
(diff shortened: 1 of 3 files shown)


IainSAP (approved):
THanks @815are 

- Reverted to previous behaviour as expected
- Test re-instated 
- Tested with consuming code, behaves as expected

devinea (approved):
Code changes are clear.
changeset ✅

Issue #1611, opened 2024-01-24
FEATURE - Enhance writer for ADP Project to be able to create proper ADP Project working in BAS and VSCODE


GDamyanov:
_As a **persona**,
I want to **have writer**,
so that I **using with config to be able to create ADP Project**_

### Description

Background: there is implementation for writer which creates ADP Project for VSCODE. This writer should be able to create ADP Project working also in BAS.

### Technical Design
The writer located in @sap-ux/adp-tooling should be enhanced to create ADP Project working in BAS

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a writer 
**when** I execute it with config
**then**  I should have ADP Project generated working in BAS and VSCODE

### Tasks
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #1775, opened 2024-03-25
feat: Switch Adaptation Project's generator/wizard to use Open Source writers

IvoSG:
Feat for https://github.com/SAP/open-ux-tools/issues/1611.

Adds writer functionality for adaptation projects.


packages/adp-tooling/test/unit/writer/__snapshots__/index.test.ts.snap
```diff
@@ -63,18 +62,17 @@ Object {
...
     \\"keywords\\": [
         \\"ui5\\",
         \\"sapui5\\",
         \\"adaptation-project\\"
     ],
-    \\"dependencies\\": {},
     \\"devDependencies\\": {
         \\"@sap/ux-ui5-tooling\\": \\"^1.11.4\\",
-        \\"@ui5/cli\\": \\"^3.9.1\\"
+        \\"@ui5/cli\\": \\"^3.6.0\\"
```
nikmace, line 75: Just a question, why did this version change? It was `^3.9.1` before. `^3.6.0` will install the latest too.

.changeset/famous-panthers-occur.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ux/adp-tooling': minor
+'@sap-ux/preview-middleware': patch
+---
+
+Adds writer functionality for generation of Adaptation Project's
```
nikmace, line 6: Spelling -> Projects

packages/adp-tooling/src/types.ts
```diff
@@ -347,3 +364,49 @@ export interface AdpProjectData {
     reference: string;
     id: string;
 }
+
+export interface AdpCustomConfig {
+    adp?: {
+        safeMode: boolean;
```
nikmace, line 370: Should we also have an `environment?: 'ABAP' | 'S4HANA' | 'CF'` under `adp`?

nikmace (changes requested):
Coverage is currently at 88.7% and could be improved a little bit.

nikmace (changes requested):

packages/adp-tooling/templates/projects/onprem/package.json
```diff
@@ -14,7 +14,7 @@
         "@sap-ux/preview-middleware": "^0.11.1",
         "@sap-ux/ui5-proxy-middleware": "^1.3.0",
         "@sap-ux/deploy-tooling": "^0.11.7"<%}%>,
-        "@ui5/cli": "^3.9.1"
+        "@ui5/cli": "^3.6.0"
```
nikmace, line 17: Same question as before, is there any reason for this version "downgrade".

packages/adp-tooling/src/base/project-utils.ts
```diff
@@ -0,0 +1,206 @@
...
+/**
+ * Writes a manifest.adpescr file within a specified folder in the project directory.
+ *
+ * @param {string} templatePath - The root path of the project template.
+ * @param {string} projectPath - The root path of the project.
+ * @param {ManifestAppdescr} data - The data to be populated in the template file.
+ * @param {Editor} fs - The `mem-fs-editor` instance used for file operations.
+ * @returns {void}
+ */
+export function writeManifestAppdecr(
```
nikmace, line 49: Spelling `writeManifestAppdecr` --> writeManifestAppde**s**cr

packages/adp-tooling/src/base/project-utils.ts
```diff
@@ -0,0 +1,206 @@
...
+ * @param {Editor} fs - The `mem-fs-editor` instance used for file operations.
+ * @param {string[]} exclude - Files to be ignored when writing.
+ * @returns {void}
+ */
+export function writeTemplateToFolder(
+    templatePath: string,
+    projectPath: string,
+    data: CfModuleData | AdpWriterConfig,
+    fs: Editor,
+    exclude: string[] | [] = []
```
nikmace, line 28: Maybe as a suggestion for more clarity, this should be `ignoredFiles: string[] ...` or `excludedFiles`

nikmace (changes requested):
GitHub actions for some reason are not running and I cannot see the result of those voters. Also, there are some merge conflicts to be resolved.

Some comments from before have not been addressed yet. And could you please write  what have you changed in my _requested changes comments_, so I can resolve those comments.

packages/adp-tooling/src/types.ts
```diff
@@ -199,6 +227,29 @@ export const enum ChangeType {
...
+    CF = 'CF'
+}
+
+/**
+ * Maps a ChangeType to the corresponding data structure needed for that type of change.
+ * This conditional type ensures type safety by linking each change type with its relevant data model.
+ *
+ * @template T - A subtype of ChangeType indicating the specific type of change.
+ */
+export type ProjectGeneratorData<T extends ProjectType> = T extends ProjectType.ON_PREM
```
nikmace, line 245: It would be good to remove the generator key word from the interface (if it is used by the create module there you are not using the generator)

packages/adp-tooling/src/types.ts
```diff
@@ -347,3 +398,50 @@ export interface AdpProjectData {
     reference: string;
     id: string;
 }
+
+export interface AdpCustomConfig {
+    adp: {
+        safeMode: boolean;
+        environment: ProjectType;
```
nikmace, line 405: TBD: the environment is now 'OnPrem' but our change editors still have checks for 'ABAP'

packages/adp-tooling/templates/projects/s4/package.json
```diff
@@ -0,0 +1,50 @@
...
+    ],
+	"repository": {
+		"type": "git",
+		"build": "ui5.js build --verbose --include-task generateCachebusterInfo"
+	},
+	"author": "",
+	"license": "ISC",
+	"ui5": {
+		"dependencies": ["@sap/ui5-builder-webide-extension", "@ui5/task-adaptation"]
+	},
```
nikmace, line 20: Looks like this file was not prettier'd and there are tabulations visible

packages/adp-tooling/templates/projects/cf/xs-security.json
```diff
@@ -0,0 +1,7 @@
+{
+	"xsappname": "<%= projectName %>",
+	"tenant-mode": "dedicated",
+	"description": "Security profile of called application",
+	"scopes": [],
+	"role-templates": []
```
nikmace, line 6: This file is not prettier'd too

packages/adp-tooling/src/base/project-utils.ts
```diff
@@ -0,0 +1,206 @@
...
+ */
+export function writeManifestAppdecr(
+    templatePath: string,
+    projectPath: string,
+    data: ManifestAppdescr,
+    fs: Editor
+): void {
+    try {
+        const appdescrTplPath = join(templatePath, 'webapp/manifest.appdescr_variant');
+        const appdescrPath = join(projectPath, 'webapp/manifest.appdescr_variant');
```
nikmace, line 57: Could you please use the enum from project access for different folders

packages/adp-tooling/src/base/project-utils.ts
```diff
@@ -0,0 +1,206 @@
...
+ * Writes a .apd/config.json file within a specified folder in the project directory.
+ *
+ * @param {string} projectPath - The root path of the project.
+ * @param {CfAdpConfig} data - The data to be populated in the template file.
+ * @param {Editor} fs - The `mem-fs-editor` instance used for file operations.
+ * @returns {void}
+ */
+export function writeAdpConfig(projectPath: string, data: CfAdpConfig, fs: Editor): void {
+    try {
+        const adpConfigPath = join(projectPath, '.adp/config.json');
```
nikmace, line 78:
```suggestion
        const adpConfigPath = join(projectPath, '.adp', 'config.json');
```

packages/create/src/cli/generate/adaptation-project.ts
```diff
@@ -78,7 +78,7 @@ async function generateAdaptationProject(
         if (!basePath) {
             basePath = join(process.cwd(), config.app.id);
         }
-        const fs = await generate(basePath, config);
+        const fs = await generate(basePath, ProjectType.ON_PREM, config);
```
nikmace, line 81: I think this config is missing some properties, when you pass the project type to the method you might see ts errors.

```suggestion
        const fs = await generate<ProjectType.ON_PREM>(basePath, ProjectType.ON_PREM, config);
```

nikmace (changes requested):

tobiasqueck (changes requested):
Sorry but that one is far away from being acceptable.

Please remove things like a `pom.xml` that only make sense for SAP internal projects and also remove all the duplicated files in the templates section.
Projects that are for an ABAP system are almost identical (no matter if on premise, private cloud our public cloud). The only difference is the additional build task and a minor change in the deploy script.
I am not an expert for CF but I am pretty sure that the generated structure is also not aligned with the latest guidance on CF projects.

My recommendation is actually to start from scratch again. Take the existing writer/templates (from `main`) and compare its results with the three different options. Then enhance the "core" that is the same for all option and add diffs if required for other targets. I'd recommend you have a look at the `fiori-elements-writer` and how it is designed.

packages/adp-tooling/test/unit/writer/index.test.ts
```diff
@@ -86,5 +86,146 @@ describe('ADP writer', () => {
...
+            const data = `ABAP_USERNAME: test_name
+                          ABAP_PASSWORD: test_pass`;
+            await generateEnv(projectDir, data, fs);
+            expect(
+                fs.dump(projectDir, (file) => file.dirname === projectDir && ['.env'].includes(file.basename))
+            ).toMatchSnapshot();
+        });
+    });
+
+    describe('generateCf', () => {
```
nikmace, line 155: This can be removed as CF will not be part of this PR?

packages/adp-tooling/src/types.ts
```diff
@@ -35,21 +35,41 @@ export interface AdpWriterConfig {
...
+        ui5Version?: string;
+        ui5EndpointUrl?: string;
     };
     package?: {
         name?: string;
         description?: string;
     };
+    flp?: {
+        flpTitle?: string;
+        flpSubtitle?: string;
```
nikmace, line 47: Since these are under `flp` maybe we can omit this prefix `flp`? So it becomes this:
```suggestion
        title?: string;
        subtitle?: string;
```

packages/adp-tooling/src/types.ts
```diff
@@ -35,21 +35,41 @@ export interface AdpWriterConfig {
     target: AbapTarget;
     ui5?: {
         minVersion?: string;
+        ui5Version?: string;
+        ui5EndpointUrl?: string;
```
nikmace, line 39: Since these are under `ui5` maybe we can omit this prefix `ui5`? So it becomes this:
```suggestion
        version?: string;
        endpointUrl?: string;
```
Or just `url`?

nikmace (changes requested):

packages/adp-tooling/test/unit/writer/index.test.ts
IvoSG: Done

packages/adp-tooling/src/types.ts
IvoSG: Done

packages/adp-tooling/src/types.ts
IvoSG: Done

packages/adp-tooling/templates/project/package.json
```diff
@@ -10,17 +10,19 @@
...
+        <%if (locals.options?.fioriTools) {%>"@sap/ux-ui5-tooling": "1"<%} else {%>"@sap-ux/backend-proxy-middleware": "^0.7.5",
         "@sap-ux/preview-middleware": "^0.11.1",
         "@sap-ux/ui5-proxy-middleware": "^1.3.0",
         "@sap-ux/deploy-tooling": "^0.11.7"<%}%>,
-        "@ui5/cli": "^3.9.1"
+        "@ui5/task-adaptation": "^1.3.0",
+        "@ui5/cli": "^3.9.2"
     },
     "scripts": {
         "build": "ui5 build --exclude-task generateFlexChangesBundle generateComponentPreload --clean-dest",
```
tobiasqueck, line 21: Could you please add `minify` to the excluded task? Otherwise, `.map` files will be generated that are rejected by LREP on your ABAP versions

packages/adp-tooling/src/writer/index.ts
```diff
@@ -42,26 +49,18 @@ export async function generate(basePath: string, config: AdpWriterConfig, fs?: E
...
-        const ui5DeployConfig = await UI5Config.newInstance(baseUi5ConfigContent);
-        enhanceUI5DeployYaml(ui5DeployConfig, fullConfig);
-        fs.write(join(basePath, 'ui5-deploy.yaml'), ui5DeployConfig.toString());
+    writeTemplateToFolder(join(tmplPath, '**/*.*'), join(basePath), fullConfig, fs, ignoredFiles);
+    if (fullConfig.appdescr) {
+        writeManifestAppdescr(tmplPath, basePath, fullConfig.appdescr, fs);
     }
+    await writeUI5DeployYaml(basePath, fullConfig, fs);
+    await writeUI5Yaml(basePath, fullConfig, fs);
+    await writeEnvFile(basePath, fullConfig, fs);
```
tobiasqueck, line 63: Does the user know that we are writing user/password to the filesystem? If not, we cannot do that. 
Btw, as far as I understand, both user/password are only used for the additional tasks for S/4HANA Cloud Public Edition scenarios, however, such systems won't allow access with user/password, therefore, I'd strongly prefer to remove this user/password/env thing.

packages/adp-tooling/src/writer/options.ts
```diff
@@ -120,3 +164,36 @@ function getOpenSourceMiddlewares(config: AdpWriterConfig): CustomMiddleware<obj
...
 }
+
+/**
+ * Get a list of required custom tasks for S4.
+ *
+ * @param config full project configuration
+ * @returns list of required tasks.
+ */
+function getAdpCloudCustomTasks(config: AdpWriterConfig & { target: AbapTarget }): CustomTask[] {
+    const user = 'env:ABAP_USERNAME';
```
tobiasqueck, line 175: As mentioned in my other comment, this shouldn't be here. In the very worst case that there is a valid scenario for this, use at least `FIORI_TOOLS_USER` and `FIORI_TOOLS_PASSWORD` because they are available if any previous Fiori tools task or middleware relied on user/password.

tobiasqueck (changes requested):

packages/adp-tooling/src/types.ts
```diff
@@ -35,21 +35,41 @@ export interface AdpWriterConfig {
...
     ui5?: {
         minVersion?: string;
+        version?: string;
+        url?: string;
     };
     package?: {
         name?: string;
         description?: string;
     };
+    flp?: {
```
tobiasqueck, line 45: That object is odd. `bspName` and `credentials` are used to access the original app while `title`, `subtitle` and `inboundId` are only being used in the i18n file. Did you miss adding a change to the appdescr if these properties are present or did I miss it in the code or is there some magic backend functionality that does something on deploy? 

What is the difference between `bspName` and `reference` in the app section?

packages/adp-tooling/src/types.ts
```diff
@@ -192,6 +213,14 @@ export const enum ChangeType {
     CHANGE_INBOUND = 'appdescr_app_changeInbound'
 }
 
+/**
+ * Enumerates the types of projects that can be made, each representing a specific kind of project structure.
+ */
+export const enum ProjectType {
```
tobiasqueck, line 219: Both are S/4 systems here, so please change it to something like `Cloud` and `OnPremise`. Even better would be if you reuse the type `OperationsType` (https://github.com/SAP/open-ux-tools/blob/main/packages/axios-extension/src/abap/types/adt-types.ts#L77) because that is exactly what you will get back from the backend as identify during the prompting phase.

packages/adp-tooling/src/types.ts
tobiasqueck: @nikmace  Why would the change editors need to know about the target environment?

packages/adp-tooling/src/types.ts
```diff
@@ -35,21 +35,41 @@ export interface AdpWriterConfig {
     target: AbapTarget;
     ui5?: {
```
tobiasqueck, line 36: Could you please use `frameworkUrl` like https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-application-writer/src/types.ts#L43, so that we keep both objects in sync.
Btw, what is the use case here? Is it for openUI5 or is there a different reason? It doesn't seem like an openUI5 use case since the target system is always an ABAP system, or did I miss something?

IvoSG: It is about the internal scenario when the snapshot version is selected and points to an internal endpoint

packages/adp-tooling/src/types.ts
IvoSG: Currently, the title, subtitle and input id are only used in the i18n file because the appdescr information comes already filtered after the prompting.
bspName is the base app repoName  and reference is the base app id

packages/adp-tooling/src/writer/options.ts
tobiasqueck: As discussed on the phone, please remove it.

packages/adp-tooling/src/writer/index.ts
tobiasqueck: As discussed, please remove it.

packages/adp-tooling/src/types.ts
tobiasqueck: As discussed on the phone, please add the change for the inbound communication to the appdescr template file with the condtion that the flp object is provided

packages/adp-tooling/src/types.ts
tobiasqueck: Thanks.

packages/adp-tooling/src/types.ts
```diff
@@ -35,21 +35,41 @@ export interface AdpWriterConfig {
...
      */
-    deploy?: Adp;
+    deploy?: Adp | BspApp;
     options?: {
         /**
          * Optional: if set to true then the generated project will be recognized by the SAP Fiori tools
          */
         fioriTools?: boolean;
     };
+    appdescr?: ManifestAppdescr;
```
tobiasqueck, line 67: As discussed on the phone, please reduce this to a change array (i.e. the content section of the appdescr) so that the caller can push additional changes but cannot affect the rest of the file.

packages/adp-tooling/templates/project/package.json
IvoSG: Done

packages/adp-tooling/src/writer/index.ts
IvoSG: Done

packages/adp-tooling/src/types.ts
IvoSG: The change for the inbound it's more complex to be in the aprdescr template and because of that it's passed with other changes in the content

packages/adp-tooling/src/types.ts
IvoSG: Done

packages/adp-tooling/src/types.ts
IvoSG: Done

packages/adp-tooling/src/types.ts
tobiasqueck: If it is too complex for ejs, implement it as a plain function and then add it to the list of changes. Most importantly move the logic here because you have a reference to it in your generated i18n.properties file

packages/adp-tooling/templates/project/webapp/manifest.appdescr_variant
```diff
@@ -7,6 +7,9 @@
   "namespace": "apps/<%- app.reference %>/appVariants/<%- app.id %>/",
   "version": "0.1.0",
   "content": [
+<%if(app.content){-%><% for(let change of app.content){-%>
+    <%- JSON.stringify(change) %><%-","%>
```
tobiasqueck, line 11: It might be nice to align the spacing of your JSON with the rest of the document e.g. JSON.stringify(change, undefined, 4)

packages/adp-tooling/src/writer/options.ts
IvoSG: done

packages/adp-tooling/templates/project/webapp/manifest.appdescr_variant
IvoSG: done

packages/adp-tooling/src/types.ts
```diff
@@ -340,3 +364,59 @@ export interface AdpProjectData {
...
+            subTitle: string;
+            signature: {
+                parameters: object | string;
+                additionalParameters: "allowed";
+            };
+        };
+    };
+}
+
+export const enum FlpConfigurationType {
```
tobiasqueck, line 420: What is this enum of one value?

IvoSG: Currently, we are supporting only ADD_NEW_TILE, but expecting to be more in the future

packages/adp-tooling/src/types.ts
```diff
@@ -52,6 +58,23 @@ export interface AdpWriterConfig {
     };
 }
 
+export interface FlpConfiguration {
+    title: string;
+    bspName: string;
```
tobiasqueck, line 63: Is the BSP name part of the FLP configuration? Or is it something else?

packages/adp-tooling/src/types.ts
```diff
@@ -52,6 +58,23 @@ export interface AdpWriterConfig {
...
+    languages: Language[];
+    inboundId: string;
```
tobiasqueck, line 65: The type indicates that `inboundId` is mandatory but looking at the code it is not.

packages/adp-tooling/src/types.ts
tobiasqueck: Could you give me a (few) examples? And will these also be called during the project generation?

packages/adp-tooling/src/types.ts
```diff
@@ -31,19 +32,24 @@ export interface AdpWriterConfig {
         reference: string;
         layer?: UI5FlexLayer;
         title?: string;
+        content?: Content[];
```
tobiasqueck, line 35: You are only using it if it is a cloud system (if I read the code correctly) which I think is a bit odd because the API does not indicate that.

packages/adp-tooling/src/types.ts
```diff
@@ -52,6 +58,23 @@ export interface AdpWriterConfig {
     };
 }
 
+export interface FlpConfiguration {
```
tobiasqueck, line 61: This type needs more explanation, please add a short comment to each property so that it is easier to understand what this config is for.

packages/adp-tooling/src/writer/index.ts
```diff
@@ -59,14 +49,15 @@ export async function generate(basePath: string, config: AdpWriterConfig, fs?: E
...
-    const fullConfig = setDefaults(config);
 
-    fs.copyTpl(join(tmplPath, '**/*.*'), join(basePath), fullConfig, undefined, {
-        globOptions: { dot: true },
-        processDestinationPath: (filePath: string) => filePath.replace(/gitignore.tmpl/g, '.gitignore')
-    });
+    const fullConfig = setDefaults(config);
 
-    await writeUi5Yaml(basePath, fullConfig, fs);
+    if (fullConfig.customConfig?.adp.environment === "C" && fullConfig.flp) {
```
tobiasqueck, line 55: Is the FLP config not valid for on prem systems?

tobiasqueck (changes requested):
I added a few inline questions.

You have a few places where you set defaults if not provided as input e.g. the inboundId. I found three places where you do it, therefore, I'd recommend you create something like a `mergeWithDefaults` function and call it right in the beginning. Additionally, you have properties that are required if another property is not provided, therefore, you should do a validation of the input right at the beginning instead of in multiple places of the code. Please check https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-elements-writer/src/index.ts#L64-L66, it has the validation and then the merging cleanly in one place at the beginning. Afterwards you can rely on a valid configuation.

.changeset/famous-panthers-occur.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sap-ux/adp-tooling': minor
+'@sap-ux/preview-middleware': patch
```
tobiasqueck, line 3: Please remove preview-middleware and axios-extension from the changeset.

packages/adp-tooling/test/unit/writer/index.test.ts
```diff
@@ -40,7 +40,7 @@ describe('ADP writer', () => {
     describe('generate', () => {
         test('minimal config', async () => {
             const projectDir = join(outputDir, 'minimal');
-            await generate(projectDir, config, fs);
+            await generate(projectDir, config);
```
tobiasqueck, line 43: What is the reason for removing the parameter `fs` here?

packages/adp-tooling/src/writer/index.ts
IvoSG: Yes

packages/adp-tooling/src/types.ts
IvoSG: Content is used by both onPrem and Cloud if it's provided

packages/adp-tooling/src/types.ts
IvoSG: As far as I know, there is expected to be a "CHANGE TILE" as well

IvoSG: Removed

packages/adp-tooling/src/types.ts
IvoSG: Done

.changeset/famous-panthers-occur.md
IvoSG: Done

packages/adp-tooling/src/writer/options.ts
```diff
@@ -153,3 +164,151 @@ function getOpenSourceMiddlewares(config: AdpWriterConfig): CustomMiddleware<obj
...
+    const inboundChangeContent = addInboundId ? getInboundChangeContentWithNewInboundID(flpConfiguration as NewInboundNavigation, app.id) : getInboundChangeContentWithExistingInboundId(flpConfiguration as ChangeInboundNavigation, app.id);
+    if (inboundChangeContent) {
+        const firstFlpChange = {
+            changeType: addInboundId ? "appdescr_app_addNewInbound" : "appdescr_app_changeInbound",
+            content: inboundChangeContent,
+            texts: {
+                "i18n": "i18n/i18n.properties"
+            }
+        };
+        const secondFlpChange = {
```
tobiasqueck, line 303: Could we have a more descriptive name than first and second change .e.g addInboundChange and removeOtherInboundsChange

packages/adp-tooling/src/types.ts
```diff
@@ -56,6 +69,32 @@ export interface AdpWriterConfig {
...
+export interface NewInboundNavigation {
+    /** Represent business entities and can bundle applications that reflect a specific scenario. */
+    semanticObject: string;
+    /** Operations which can be performed on a semantic object. */
+    action: string;
+    //** Defined instance of the semantic object (e.g. by specifying the employee ID). */
+    additionalParameters?: object;
+    title: string;
+    subTitle?: string;
+    inboundId?: string;
```
tobiasqueck, line 88: `inboundId` shouldn't exit on a new navigation, correct?

packages/adp-tooling/test/unit/writer/index.test.ts
tobiasqueck: Good point but we use the `fs` instance here to be able to write the output to the filessystem for debugging (check the `afterAll` function), thereofre, please revert your change here

packages/adp-tooling/test/unit/writer/index.test.ts
```diff
@@ -88,6 +88,70 @@ describe('ADP writer', () => {
...
+            const projectDir = join(outputDir, 'minimal');
+            await generateEnv(projectDir, '');
+            expect(
+                fs.dump(projectDir, (file) => file.dirname === projectDir && ['.env'].includes(file.basename))
+            ).toMatchSnapshot();
+        });
+
+        test('generate env file for s4hana cloud projects', async () => {
+            const projectDir = join(outputDir, 'env');
+            const data = `ABAP_USERNAME: test_name
```
tobiasqueck, line 148: Didn't we remove user/password from the code? Please also remove it from the test

IvoSG: done

packages/adp-tooling/src/writer/options.ts
IvoSG: done

packages/adp-tooling/test/unit/writer/index.test.ts
IvoSG: done

packages/adp-tooling/src/types.ts
IvoSG: Moved to the CloudApp interface

packages/adp-tooling/src/types.ts
IvoSG: Moved in internal type

packages/adp-tooling/templates/project/webapp/manifest.appdescr_variant
```diff
@@ -7,6 +7,9 @@
   "namespace": "apps/<%- app.reference %>/appVariants/<%- app.id %>/",
   "version": "0.1.0",
   "content": [
+<%if(app.content){-%><% for(let change of app.content){-%>
+    <%- JSON.stringify(change, undefined, 4) %><%-","%>
```
nikmace, line 11: Are you sure it should be 4 instead of 2? Most prettier configs have space set to 2.

nikmace (dismissed):
Code looks good
Coverage is good
Did not test manually

packages/adp-tooling/templates/project/webapp/manifest.appdescr_variant
IvoSG: yes, because it is an inner object

packages/adp-tooling/src/types.ts
```diff
@@ -175,6 +229,7 @@ export type IWriterData<T extends ChangeType> = IWriter<GeneratorData<T>>;
  *
  * @template T - The specific type of data the writer will handle, determined by the associated ChangeType.
  */
+
```
testojs, line 232: Please remove this extra empty line as it detaches the jsdoc from the interface.

testojs (dismissed):
- the code looks good
- good test coverage
- not tested manually

tobiasqueck (changes requested):
* changeset: the writer functionality was there before but you added support for cloud projects, therefore, please adjust the text
* to make use of the writer in our `create` a small enhancement is required (#1964), I'd recommend merging it into your branch and then merging it together into main so that everything stays consistent
* code is well structured and well documented
* test coverage is sufficient

Please consider the first two points, otherwise, I am happy to approve

packages/adp-tooling/src/types.ts
nikmace: Removed in: [4e27dbf](https://github.com/SAP/open-ux-tools/pull/1775/commits/4e27dbff0ba748cc271d6d99bb08a05d7657dd4a)

nikmace:
@tobiasqueck @testojs Comments are addressed, would you please do a review again?

testojs (approved):
Re-approving

tobiasqueck (approved):
All requested changes have been applied. Thank you!

PR #1786, opened 2024-03-26
Revert "fix: revert cds root set"

815are:
Reverts SAP/open-ux-tools#1774

After investigation we found that following code is executed if we change `cds.root`:
image

It is executed in following scope:
1. project path(changed value of `cds.root`) had `spec` dependency;
2. there no `spec` dependency in node_modules yet(install is in progress);
3. then there is call for `require.resolve` for `spec` with cds, which creates cached value with failure state(package.json reader cache);

Then we try to access dependency on next round(when node_modules is installed) - `resolve` throws error based on failure state

### DIFF
--- a/.changeset/cool-years-deny.md
+++ b/.changeset/cool-years-deny.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': patch
+---
+
+Remove setting of cds.root.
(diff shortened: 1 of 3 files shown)


Klaus-Keller (dismissed):
Thanks @815are for investigation. Revert to the fix looks good.

Klaus-Keller (approved):
Re-approve after changeset has been added

IainSAP (approved):
Thanks @815are 

- Removal of `cds.root` setting s as expected

PR #1787, opened 2024-03-27
feat: appdescr_variant extension watched by default

vadson71:
The appdescr_variant file extension added to the default set of watched files


voicis (approved):
- Straight forward code change
- Did not test manually

Issue #1791, opened 2024-03-27
FEATURE - Enhance preview-middleware for Adaptation Projects with Telemetry


nikmace:
_As a **developer**,
I want to **be able to record telemetry **,
so that I **can view the telemetry actions and improve the product**_

### Description

_Have telemetry available for Adaptation Projects_

### Technical Design
_What modules need to be created or enhanced?_ 
_How would the functionality be exposed and how can it be consumed?_

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have an ADP
**when** I run the editor and do some changes, change some properties, etc
**then** user actions are sent as telemetry.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #1794, opened 2024-03-28
feat: Enhance preview-middleware for Adaptation Projects with Telemetry

nikmace:
Feat for #1791.

- Adds new route for telemetry for Adaptation Projects.


packages/preview-middleware/src/base/telemetry-reporter.ts
```diff
@@ -0,0 +1,48 @@
+import { isInternalFeaturesSettingEnabled } from '@sap-ux/feature-toggle';
+import { ClientFactory, SampleRate, initTelemetrySettings } from '@sap-ux/telemetry';
+import modulePackageJson from '../../package.json';
+// import { join } from 'path';
```
voicis, line 4: Please remove commented code.

packages/preview-middleware/src/base/telemetry-reporter.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Class handling telemetry initialization and reporting.
+ */
+export class TelemetryReporter {
+    private initialized: boolean;
+    constructor(private eventName: string, private readonly logger: Logger, private layer?: string) {}
+    public async initializeTelemetry() {
+        try {
+            // const envFilePath = { path: join(__dirname, '..', '..', '.env') };
+            // dotenv.config(envFilePath);
```
voicis, line 19: Please remove commented code.

packages/preview-middleware/src/base/flp.ts
```diff
@@ -639,6 +674,7 @@ export async function initAdp(
             flp.rta.layer = layer;
             flp.rta.options = {
                 projectId: variant.id,
+                telemetry: true, // temprory until passed during intiialization
```
voicis, line 677: There should be a way for the user to disable telemetry, but now it is not possible.

packages/preview-middleware/src/base/flex.ts
```diff
@@ -90,6 +92,10 @@ export function writeChange(
         }
         const filePath = join(path, fileName + '.' + fileType);
         writeFileSync(filePath, JSON.stringify(data, null, 2));
+        const telemetryData = { category: 'Save', changeType: (data as any).changeType, layer: (data as any).layer };
```
voicis, line 95: 1. We don't need to build telemetry data if reporter isn't passed.
2. It's better to extend the parameter type definition on line 80, than using `any`.

voicis (changes requested):
Please also add changeset.

packages/preview-middleware/src/base/telemetry-reporter.ts
vinayhospete: done

packages/preview-middleware/src/base/telemetry-reporter.ts
vinayhospete: done

packages/preview-middleware/src/base/flp.ts
vinayhospete: Now using central telemetrySetting

packages/preview-middleware/src/base/flex.ts
vinayhospete: ok. Extended parameters.

packages/preview-middleware/src/base/telemetry-reporter.ts
```diff
@@ -0,0 +1,41 @@
...
+                resourceId: key
+            });
+            this.initialized = true;
+        } catch (e) {
+            this.initialized = false;
+            this.logger.error(`Could not initialize telemetry. ${e}`);
+        }
+    }
+
+    public reportTelemetry(data: any): void {
```
nikmace, line 28: Would it be good to make this method generic to accept some type depending on the use case? Because essentially data can be anything.

```suggestion
    public reportTelemetry<T extends {}>(data: T): void {
```

vinayhospete: currently we have less parameters to report , shall we use below ?
```
data: {
        category: string;
        changeType: string;
        controlType: string;
        propertyName: string;
        sapui5Version: string;
    }
```

nikmace: But then when you use more parameters or want to send data that is not on this data object, what do you do? I think we need to have more flexibility here. Or these properties will always stay the same and you can pass only those?

vinayhospete: ok  lets keep it flexible. I will change it now

packages/preview-middleware/src/base/flex.ts
```diff
@@ -76,9 +77,17 @@ export async function readChanges(project: ReaderCollection, logger: Logger): Pr
  * @returns object with success flag and optional message
  */
 export function writeChange(
-    data: object & { fileName?: string; fileType?: string },
+    data: object & {
+        fileName: string;
```
voicis, line 81: `fileName` and `fileType` were optional properties before and we're checking if they exist on line 94. We probably should not change the type definitions for those properties.

packages/preview-middleware/src/base/flex.ts
```diff
@@ -76,9 +77,17 @@ export async function readChanges(project: ReaderCollection, logger: Logger): Pr
...
+        fileType: string;
+        changeType: string;
```
voicis, line 83: We should probably be less strict with `changeType` as well and have it as optional property and later default to some `unknown` type perhaps, so we don't get `undefined` value in the telemetry data.

packages/preview-middleware/src/base/telemetry-reporter.ts
```diff
@@ -15,14 +12,10 @@ export class TelemetryReporter {
     constructor(private eventName: string, private readonly logger: Logger, private layer?: string) {}
     public async initializeTelemetry() {
         try {
-            // const envFilePath = { path: join(__dirname, '..', '..', '.env') };
-            // dotenv.config(envFilePath);
-            const internalFeature = isInternalFeaturesSettingEnabled();
             await initTelemetrySettings({
                 consumerModule: modulePackageJson,
-                internalFeature,
+                internalFeature: false,
```
voicis, line 17: Should this value be hard coded? We won't be able to change this even if it is used in an internal environment.

packages/preview-middleware/src/base/flex.ts
```diff
@@ -90,6 +99,16 @@ export function writeChange(
         }
         const filePath = join(path, fileName + '.' + fileType);
         writeFileSync(filePath, JSON.stringify(data, null, 2));
+        const telemetryData = {
+            category: 'Save',
+            changeType: data.changeType,
+            controlType: data?.selector?.type ?? '',
+            propertyName: data?.content?.property ?? '',
```
voicis, line 106: We probably need to add a test case to cover these branches to reach the required code coverage.

packages/preview-middleware/src/base/telemetry-reporter.ts
Jimmy-Joseph19: So far i have observed is for cpe and adp it was always started from cli and the method that check relies on vscodehttps://github.com/SAP/open-ux-tools/blob/dbae4d0a5d00e0e89961df1fdccfbe3559326614/packages/feature-toggle/src/featureToggle.ts#L135 and this env variable and which is not currnely used by our module so to avoid deps with FeatureToggle package i suggested that it can be always set false

packages/preview-middleware/src/base/flex.ts
Jimmy-Joseph19: done

packages/preview-middleware/src/base/telemetry-reporter.ts
Jimmy-Joseph19: I have added it back. thanks

slin-sap: @voicis  @Jimmy-Joseph19 

Internal feature toggle was not part of open source library. Therefore telemetry lib was not able to call feature toggle internally like we did in internal repo. That's why it was exposed as a configuration and let the consumer to pass in the value.

packages/preview-middleware/src/base/flex.ts
Jimmy-Joseph19: covered

.github/workflows/pipeline.yml
```diff
@@ -149,6 +149,13 @@ jobs:
                   node-version: 18.x
             - name: Install pnpm modules
               run: pnpm install
+            - name: Replace placeholder with instrumentation key
```
donal-tobin-sap, line 152: There are major changes required to the whole concept used here.

donal-tobin-sap: 1. it is embedding the internal SAP tools-suite dev telemetry key into the generated artifact - if a real telemetry key is required a new separate key/azure project is required
2. some separation of dev/test keys versus production keys are required - real consumer such as tools-suite will need to add its dev or production keys
3. the replacement cannot be based on key value - but has to ignore the value
4. If this is for test only then the test such just read the secret as an ENV var - the code to read and inject an env var is straight forward
5. the tools-suite patch for developers could be used but requires a new azure key

donal-tobin-sap (changes requested):

packages/preview-middleware/test/unit/base/flex.test.ts
```diff
@@ -92,11 +93,56 @@ describe('flex', () => {
         });
 
         test('invalid change', () => {
-            const result = writeChange({}, path, logger);
+            const result = writeChange({} as any, path, logger);
```
voicis, line 96: `any` should not be needed here anymore.

packages/telemetry/src/base/types/event-name.ts
```diff
@@ -1,5 +1,7 @@
 export enum EventName {
     Test = 'test',
+    CPE_EVENT = 'CONTROL_PROPERTY_EDITOR_EVENT',
```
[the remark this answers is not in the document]
vinayhospete: Ok, I moved it to preview-middleware package.

packages/preview-middleware/test/unit/base/flex.test.ts
vinayhospete: removed

.github/workflows/pipeline.yml
vinayhospete: As we aligned in Architecture round, telemetry will be initialized from toolsuite, open ux changes related to telemetry initialization will be removed

PR #1813, opened 2024-04-05
fix(preview-middleware): set qunit autostart to false

heimwege:
Set  qunit autostart to false as recommended by @matz3 [here](https://github.com/SAP/open-ux-tools/pull/1726/files#r1549565997)

### DIFF
--- a/packages/preview-middleware/templates/test/qunit.js
+++ b/packages/preview-middleware/templates/test/qunit.js
@@ -1,6 +1,8 @@
 sap.ui.define([
     'sap/ui/thirdparty/qunit-2'
 ], function() {
-        sap.ui.require(<%- JSON.stringify(tests) %>);
-    }
-);
+        QUnit.config.autostart = false;
+        sap.ui.require(<%- JSON.stringify(tests) %>, function() {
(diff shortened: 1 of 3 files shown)


packages/preview-middleware/templates/test/qunit.js
```diff
@@ -1,6 +1,8 @@
 sap.ui.define([
     'sap/ui/thirdparty/qunit-2'
 ], function() {
```
matz3, line 3: I'd suggest to rather access `QUnit` from the sap.ui.define import, not via global.

heimwege: oh, you are right 🙈

matz3: Looks better now 👍🏻

matz3 (approved):
LGTM

tobiasqueck (approved):
* changeset ok
* successfully tested it manually 
* simple change is straight forward and clear nothing else needed

cianmSAP (approved):
Update looks good
Snapshot updated
Changeset 
✅

Issue #1800, opened 2024-03-29
FEATURE - Add ui5 library reference inquirer module


cianmSAP:
As a developer using SAP Fiori tools, 
I want to add a reference to library present in my workspace, 
so that I am able to use reuse components in my Fiori elements or Freestyle applications

### Description

This module will provide prompts to select the Fiori project and ui5 library/components to be referenced  

### Technical Design
 Creates new module 
@sap-ux/ui5-library-reference-inquirer

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a workspace
**when** I execute prompt function from the module
**then** it will return the answers needed to update the project

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

cianmSAP:
Closing as `@sap-ux/ui5-library-reference-inquirer` is published

PR #1818, opened 2024-04-10
feat(library-reference-inquirer): adds new module @sap-ux/ui5-library-reference-inquirer

cianmSAP:
#1800

Adds new module @sap-ux/ui5-library-reference-inquirer

Updates project access to include utils for reading library configs


packages/project-access/src/library/types.ts
```diff
@@ -0,0 +1,24 @@
...
+    type: ReuseLibType;
+    uri: string;
+    dependencies: string[];
+    libRoot: string;
+}
+
+/**
+ * Choice for the reuse library
+ */
+export interface ReuseLibChoice {
```
IainSAP, line 21: Prompting related types should be in the prompting module. Also should extend  the inquirer type `extends ListChoiceOptions` or which-ever is correct

packages/project-access/src/library/helpers.ts
```diff
@@ -0,0 +1,297 @@
...
+    }
+};
+
+/**
+ * Creates the list choice options for the reuse libraries.
+ *
+ * @param libs - array of libraries found in the workspace folders.
+ * @returns list of library choices
+ */
+export const getLibraryChoices = async (libs?: LibraryResults[]): Promise<ReuseLibChoice[]> => {
```
IainSAP, line 119: This should be in the prompting module?

packages/project-access/src/library/types.ts
cianmSAP: Moved as requested

packages/project-access/src/library/helpers.ts
cianmSAP: Moved to prompting module

packages/project-access/package.json
```diff
@@ -34,12 +34,15 @@
     "dependencies": {
         "@sap-ux/i18n": "workspace:*",
         "@sap-ux/ui5-config": "workspace:*",
+        "fast-xml-parser": "4.2.7",
         "findit2": "2.2.3",
         "mem-fs": "2.1.0",
-        "mem-fs-editor": "9.4.0"
+        "mem-fs-editor": "9.4.0",
```
longieirl, line 40: this is an older version? Should we bump this https://www.npmjs.com/search?q=mem-fs-editor

cianmSAP: I guess this is in green justt because the `,` was added. I didn't update the version

sonar-project.properties
```diff
@@ -3,7 +3,7 @@ sonar.projectKey=SAP_open-ux-tools
 sonar.sources=.
 sonar.inclusions=**/*.ts, **/*.tsx
 sonar.exclusions=**/*.test.ts, **/*.test.tsx, **/*.story.tsx, **/examples/**/*, **/templates/**/*, **/test/**/*, **/jest-file-matchers/src/matchers/types.ts, **/jest-runner-puppeteer/**/*
-sonar.cpd.exclusions=**/i18n.ts, **/fiori-elements-writer/src/packageConfig.ts, **/fiori-freestyle-writer/src/packageConfig.ts, **/ui5-info/src/ui5-version-fallback.ts, **/odata-vocabularies/src/resources/*
+sonar.cpd.exclusions=**/i18n.ts, **/fiori-elements-writer/src/packageConfig.ts, **/fiori-freestyle-writer/src/packageConfig.ts, **/ui5-info/src/ui5-version-fallback.ts, **/odata-vocabularies/src/resources/*, **/ui5-library-reference-inquirer/src/prompts/helpers.ts
```
cianmSAP, line 6: This can be removed when the TBI https://github.com/SAP/open-ux-tools/issues/1819 is addressed

longieirl (approved):
I've reviewed the code only, didnt review all the new project files.
Changes look good.
Test coverage looks good.

packages/project-access/src/library/helpers.ts
```diff
@@ -0,0 +1,291 @@
...
+    }
+};
+
+/**
+ * Returns an array of the reuse libraries found in the folders.
+ *
+ * @param libs - array of libraries found in the workspace folders.
+ * @returns list of reuse library
+ */
+export const getReuseLibs = async (libs?: LibraryResults[]): Promise<ReuseLib[]> => {
```
devinea, line 110: Seems to overlap a lot with these functions https://github.com/SAP/open-ux-tools/blob/ee943db51305570cd3ecfe810d8a1010e1f10cc9/packages/project-access/src/project/search.ts#L369
Can the logic be shared?

packages/project-access/src/library/helpers.ts
```diff
@@ -0,0 +1,291 @@
...
+    const manifestDesc = manifest['sap.app']?.description;
+    if (typeof manifestDesc === 'string' && manifestDesc.startsWith('{{')) {
+        const desc = manifestDesc.replace(/(^{{)|(}}$)/g, '');
+        const i18n = manifest['sap.app']?.i18n?.toString();
+        if (i18n) {
+            return getI18nProperty(join(projectPath, i18n), desc).toString();
+        }
+    }
+
+    return (manifestDesc ?? '').toString();
```
devinea, line 259: Could we use  `getI18nAppPath` or `getI18nModelPaths` here to simplify the code?

devinea (approved):
New module setup looks good. 
Changeset ✅ 
A few comments on the project access library code.

packages/project-access/src/library/helpers.ts
cianmSAP: While there may be some overlap I'm not able to find a nice to consolidate any of the code.

I already utilise `findFioriArtifacts` (& sub function `filterLibraries`) earlier in the flow as it returns the root of the apps + libs in the workspace. 

`getReuseLibs` takes this result and goes deeper into each manifest/.library file and pulls out the relevant info for each lib or component so if it's much more involved.

packages/project-access/src/library/helpers.ts
cianmSAP: I've updated to use `getI18nPropertiesPaths` and `getPropertiesI18nBundle` from `@sap-ux/i18n`

packages/project-access/src/library/helpers.ts
devinea: OK. Thanks for looking into it.

devinea (approved):
New module set-up looks good.
Changeset 👍

longieirl (approved):
Previous comments apply, only reviewed the code and not the project configurations that were added.

PR #1823, opened 2024-04-12
fix(project-access): detect mismatch of global and project installed @sap/cds

Klaus-Keller:
To retrieve service definitions for CAP project we load module `@sap/cds` from the CAP project. As fallback, we try to load `@sap/cds` from global modules, which will be available in the common case that developer cli tools `@sap/cds-dk` is installed. This is required especially for CAP Java projects.
There is a flaw in this approach, in case the CAP project specifies a dependency to `@sap/cds` e.g. to version 7, but the node modules are not installed yet, so `npm install` was not executed.In this case the fallback to global installed `@sap/cds-dk` applies. However, the version of global installed `@sap/cds` could differ from the one specified in the project, e.g. version 6. This can lead to issues due to incompatible changes of different major versions.

This pull request addresses this issue. Internal function `loadCdsModuleFromProject()`, which is used to load `@sap/cds`, has a new optional parameter `strict`. If set, a check is conducted whether we have the situation:
-  `@sap/cds` is loaded globally
-  a dependency to `@sap/cds` exists in the `package.json` 
- major version differs

If all is true, an error is thrown. Currently, only `getCapModelAndServices()` is setting `strict` due to known incompatibilities.

### DIFF
--- a/.changeset/slimy-gifts-rescue.md
+++ b/.changeset/slimy-gifts-rescue.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': patch
+---
+
+Detect mismatching global and project installations of @sap/cds
--- a/packages/project-access/src/project/cap.ts
+++ b/packages/project-access/src/project/cap.ts
@@ -607,5 +607,5 @@ async function getCdsVersionFromPackageJson(packageJsonPath: string): Promise<st
  * @returns - major version as number
  */
 function getMajorVersion(versionString: string): number {
-    return parseInt(versionString.split('.')[0].match(/\d+/)?.[0] ?? '0', 10);
+    return parseInt(/\d+/.exec(versionString.split('.')[0])?.[0] ?? '0', 10);
 }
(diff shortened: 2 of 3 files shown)


devinea (approved):
Thanks @Klaus-Keller change looks good and should provide for a more predictable behaviour. 
Changeset ✅ 

Only nice to have suggestion might be to add more comprehensive support for CAP Java to get check the `@sap/cds-dk` version from the pom.xml

Klaus-Keller:
> Only nice to have suggestion might be to add more comprehensive support for CAP Java to get check the `@sap/cds-dk` version from the pom.xml

Yes, good point @devinea. While an inconsistency of global `@sap/cds` and CAP Java cds might be even more common, I'd like to handle the improvement for CAP Java separate.

I've fixed a [code smell](https://github.com/SAP/open-ux-tools/pull/1823/commits/f080fe094a65ef2d663758a0f4db62b71e1c4250) and set this PR to 'Ready for Review', please re-review.

devinea (approved):
Re-approving. Thanks

IainSAP (approved):
Thanks @Klaus-Keller 

- Updates as expected to support version mismatch
- Tests added

PR #1824, opened 2024-04-12
fix: Alternative approach to have a consistent save for XML Fragments

nikmace:
It currently contains two changes:

- generic hook to **intercept flex operations** by the preview-middleware on the file system
- XML fragments are not **created** anymore when the Add Fragment dialog closes but **on save**
- moved the ADP-specific change fixing from the `preview-middleware` to `adp-tooling`

In addition to this:

- **Add Fragment** and **Add Fragment at Extension Point** modals have been enhanced and now validate the fragment name also by checking the command stack. Clicking the create button only pushes the command for the specific change to the stack. When the user clicks save the **operation is intercepted** and if the type of change is **addXML** or **addXMLAtExtensionPoint** it will create an **XML file** in the workspace.


packages/adp-tooling/src/preview/adp-preview.ts
```diff
@@ -160,11 +162,42 @@ export class AdpPreview {
...
+     *
+     * @param {string} type - The type of change request.
+     * @param {CommonChangeProperties} change - An object containing properties common to all change types.
+     * @param {Editor} fs - An instance of an editor interface for file system operations.
+     * @param {Logger} logger - An instance of a logging interface for message logging.
+     * @returns {Promise<void>} A promise that resolves when the change request has been processed.
+     */
+    async onChangeRequest(
+        type: 'read' | 'write' | 'delete',
+        change: CommonChangeProperties,
```
GDamyanov, line 183: Can we extract "type" in enum?

packages/adp-tooling/test/unit/preview/change-handler.test.ts
```diff
@@ -0,0 +1,180 @@
...
+                content: {
+                    codeRef: 'coding/share.js'
+                }
+            } as unknown as CommonChangeProperties;
+
+            expect(isAddXMLChange(codeExtChange)).toBe(false);
+        });
+
+        it('should return false for change objects without a changeType', () => {
+            const unknownChange = {
```
GDamyanov, line 115: Do we need this case, since "changeType" property defined in "CommonChangeProperties" is not nullable?

GDamyanov (changes requested):

packages/adp-tooling/test/unit/preview/change-handler.test.ts
nikmace: I think that would be nice to take care here of all possible scenarios for the this particular method. The possibility that there will be no changeType is really small so I agree with you that this case is extra. But can't we leave it?

tobiasqueck (dismissed):
* changeset good
* new approach for writing xml fragment is cleaner and more consistent than the old one
* code adheres to our coding standards
* locally tested (last week) and it worked without problems

GDamyanov (approved):
Comments addressed.
Code looks good.
Manual testing done.

tobiasqueck (approved):

tobiasqueck (approved):

Issue #1815, opened 2024-04-10
BUG - @sap-ux/preview-middleware fails when serving opa5/qunit test files executed via karma


heimwege:
Hey guys,

the preview-middleware can [serve qunit and opa5 test files](https://github.com/SAP/open-ux-tools/pull/1726). This works pretty well using e.g. [ui5-test-runner](https://github.com/ArnaudBuchholz/ui5-test-runner). If you use [karma](https://karma-runner.github.io/) as test runner you get the following error:

```
[info] graph:helpers:ui5Framework: Using SAPUI5 version: 1.120.9
debug preview-middleware Config: {"flp":{"path":"/localService/index.html","apps":[]},"test":[{"framework":"Qunit"},{"framework":"OPA5"}]}
debug preview-middleware Add route for /test/unitTests.qunit.html
debug preview-middleware Add route for /test/unitTests.qunit.js
debug preview-middleware Add route for /test/opaTests.qunit.html
debug preview-middleware Add route for /test/opaTests.qunit.js
info preview-middleware Initialized for app the.id.of.the.app
debug preview-middleware Configured apps: {"app-preview":{"title":"{{appTitle}}","description":"{{appDescription}}","additionalInformation":"SAPUI5.Component=the.id.of.the.app","applicationType":"URL","url":"..","applicationDependencies":{"manifest":true}}}
05 04 2024 09:22:18.488:INFO [karma-server]: Karma v6.4.2 server started at http://localhost:9876/
05 04 2024 09:22:18.491:INFO [launcher]: Launching browsers ChromeMaximized with concurrency unlimited
05 04 2024 09:22:18.496:INFO [launcher]: Starting browser Chrome
05 04 2024 09:22:48.500:WARN [launcher]: Chrome has not captured in 30000 ms, killing.
05 04 2024 09:22:48.752:INFO [launcher]: Trying to start Chrome again (1/2).
05 04 2024 09:23:18.753:WARN [launcher]: Chrome has not captured in 30000 ms, killing.
05 04 2024 09:23:19.066:INFO [launcher]: Trying to start Chrome again (2/2).
05 04 2024 09:23:27.957:INFO [Chrome 123.0.0.0 (Windows 10)]: Connected on socket 04VO4Msgesdtr4EPAAAB with id 55656365
debug preview-middleware Serving test route: /test/unitTests.qunit.html
05 04 2024 09:23:28.026:ERROR [karma-server]: UnhandledRejection: TypeError: res.status is not a function
[...]

PR #1826, opened 2024-04-12
fix(preview-middleware-client): support connect and express api

heimwege:
Fixes https://github.com/SAP/open-ux-tools/issues/1815

### DIFF
--- a/packages/preview-middleware/src/base/flp.ts
+++ b/packages/preview-middleware/src/base/flp.ts
@@ -291,7 +291,11 @@ export class FlpSandbox {
                         scenario,
                         livereloadPort
                     });
-                    res.status(200).contentType('html').send(html);
+                    res.writeHead(200, {
+                        'Content-Type': 'text/html'
+                    });
+                    res.write(html);
+                    res.end();
                 });
                 let path = dirname(editor.path);
                 if (!path.endsWith('/')) {
@@ -305,7 +309,11 @@ export class FlpSandbox {
                     '</body>',
                     `</body>\n<!-- livereload disabled for editor </body>-->`
                 );
-                res.status(200).contentType('html').send(html);
+                res.writeHead(200, {
+                    'Content-Type': 'text/html'
+                });
+                res.write(html);
+                res.end();
             });
         }
     }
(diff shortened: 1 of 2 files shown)


packages/preview-middleware/src/base/flp.ts
```diff
@@ -411,12 +425,18 @@
                     this.logger
                 );
                 if (success) {
-                    res.status(200).send(message);
+                    res.writeHead(200);
+                    res.write(message);
```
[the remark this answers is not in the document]
heimwege: @tobiasqueck do we need something like [sanitizer](https://www.npmjs.com/package/sanitizer) to fix this issue or is this a false positive as `message` comes from `writeChange` and is actually `'FILE_CREATED ${fileName}.${fileType'`. Nevertheless I'm not sure what could happen in case of malicious file names 🙈  At least the issue must have been existing the whole time as I didn't change the general setup.

packages/preview-middleware/src/base/flp.ts
```diff
@@ -427,16 +431,33 @@ export class FlpSandbox {
...
                 }
             } catch (error) {
-                res.status(500).send(error.message);
+                this.sendResponse(res, 'text/plain', 500, error.message);
             }
         }) as RequestHandler);
     }
 
+    /**
+     * Send a response with the given content type, status and body.
```
tobiasqueck, line 445: @heimwege could you enhance the comment and mention that this function is written to work with both `express` and `connect` otherwise, someone might "optimize" it in the future.

heimwege: Done. And the integrated type check will prevent using methods that are not common 🐱

tobiasqueck (approved):
* changeset good
* changes are straightforward and well documented
* locally tested with express and didn't see any regressions
* did not test with karma

heimwege:
@tobiasqueck I tested with karma (connect) in the internal repo that led to #1815 as well as without karma (express) in my ts playground repo. Both worked, so I'm quite confident the issue is solved with this PR 😎

devinea (approved):
Code changes look appropriate. 
Changeset ✅ 
didn't test locally.

PR #1828, opened 2024-04-12
Feat/1679/update preview middleware ui5 proposal

tobiasqueck:
@ytpo-lyne the code in your PR contains multiple issues that are easiest to explain with a corresponding fix, therefore, please, have a look at the commits in this PR (please look at the commits one by one and not the PR as a whole) because each commit changes one issue at a time.

If you are ok with these changes then please merge them into your branch. Afterwards please have a look at the following open issues:

1. the new code using hungarian notation but we use camelcase notation
2. replace the remaining usage of the UrlParsing service with `URLSearchParams` as recommended in the ticket
3. define a type for the `AppState` service and then replace the usage of `any` (check how we do it for the `AppLifeCycle` service)

### DIFF
--- a/packages/preview-middleware-client/src/flp/init.ts
+++ b/packages/preview-middleware-client/src/flp/init.ts
@@ -125,35 +125,32 @@ function registerModules(
  *
  *
  */
-function resetAppState() {
-    sap.ui.require(['sap/ui/core/Core'], async function (core: Core) {
-        await core.ready();
-        const Container = sap.ui.require('sap/ushell/Container');
-        if (!Container) {
-            return;
-        }
+async function resetAppState() {
+    const Container = sap.ui.require('sap/ushell/Container');
+    if (!Container) {
+        return;
+    }
 
-        const [oURLParser, oAppState] = await Promise.all([
-            Container.getServiceAsync('URLParsing'),
-            Container.getServiceAsync('AppState')
-        ]);
-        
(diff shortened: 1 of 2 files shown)


packages/preview-middleware-client/src/flp/init.ts
```diff
@@ -240,10 +225,12 @@ export async function init({
...
     customInit?: string | null;
 }): Promise<void> {
+    const urlParams = new URLSearchParams(window.location.search);
+    const container = sap?.ushell?.Container ?? sap.ui.require('sap/ushell/Container') as typeof sap.ushell.Container;
     // Register RTA if configured
-    sap.ushell.Container.attachRendererCreatedEvent(async function () {
-        if (flex) {
-            const lifecycleService = await sap.ushell.Container.getServiceAsync<AppLifeCycle>('AppLifeCycle');
+    if (flex) {
+        container.attachRendererCreatedEvent(async function () {
```
zdravko-georgiev, line 232: @tobiasqueck Don't we need the `attachRendererCreatedEvent` also for the `resetAppState`? My understanding was that I can only use `container.getServiceAsync`, when Renderer is created.

ytpo-lyne:
Thanks, @tobiasqueck, for the proposal PR and the clear list of open issues to work on! I will merge these changes and address the open questions in the parent branch/PR.

PR #1833, opened 2024-04-16
Fiori Elements Writer - Add support for V4 table types and add TreeTable type

devinea:
Fiori Elements Writer - Add support for V4 table types and add TreeTable type

### DIFF
--- a/packages/fiori-elements-writer/src/types.ts
+++ b/packages/fiori-elements-writer/src/types.ts
@@ -46,12 +46,12 @@ export enum TableSelectionMode {
     MULTI = 'Multi',
     SINGLE = 'Single'
 }
-export interface BaseFloorPlanSettings {
+export interface TableSettings {
     tableType?: TableType;
     qualifier?: string;
     hierarchyQualifier?: string;
 }
-export interface LROPSettings extends BaseFloorPlanSettings {
+export interface LROPSettings extends TableSettings {
     entityConfig: EntityConfig;
 }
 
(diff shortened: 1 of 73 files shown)


packages/fiori-elements-writer/src/types.ts
```diff
@@ -31,12 +31,15 @@ export interface EntityConfig {
         EntitySet: string; // Defines the entity set for object page navigation
         Name: string; // Defines the entity name for object page navigation
     };
+    tableType?: TableType;
```
IainSAP, line 34: Would it be better to add these to the settings type for each supporting template. IM not sure they belong in `EntityConfig` ?

packages/fiori-elements-writer/test/unit/defaults.test.ts
```diff
@@ -40,11 +40,11 @@ describe('Defaults', () => {
               "autoHide": undefined,
               "entityConfig": Object {
                 "mainEntityName": "",
+                "tableType": "AnalyticalTable",
               },
               "multiSelect": undefined,
               "qualifier": undefined,
```
IainSAP, line 46: Strange that `qualifier` is still specified under the template setting root rather than the entity config (as per type, although Im not sure thats correct either)

devinea: Good catch. Thanks.

packages/fiori-elements-writer/src/types.ts
devinea: Addressed in https://github.com/SAP/open-ux-tools/pull/1833/commits/6afb1265e55563eb8ca01296620f6baa02f7a11d

IainSAP (approved):
Thanks @devinea 

- New types make sense
- Tests updated as required, confirming behaviour

zdravko-georgiev (approved):
Change is simple and easy to understand
Excellent test coverage
Did NOT tested manually

PR #1834, opened 2024-04-16
fix(ui5-ref-lib): fix additionalMessages and update readme

cianmSAP:
#1800

fix additionalMessages and update readme


IainSAP (approved):
Thx @cianmSAP 

- Updated as described

PR #1838, opened 2024-04-16
fix(project-access): add error code for cds mismatch

Klaus-Keller:
Add `error.code` with value `CDS_VERSION_MISMATCH` to error that is thrown, if a fallback to global `@sap/cds` results in a different major version then the project has. See also https://github.com/SAP/open-ux-tools/pull/1823.


IainSAP:
@Klaus-Keller Can we also add an error code for the error :
```
[Error: Error while reading CAP service metadata. Path: '/Users/i057546/git/open-ux-tools/packages/odata-service-inquirer/test/unit/prompts/cap-project/fixtures/bookshop', service uri: 'odata/v4/incident/', error: 'Error: Could not load cds module. Attempt to load module @sap/cds from project threw error 'Error: Module '@sap/cds' not installed in project '/Users/i057546/git/open-ux-tools/packages/odata-service-inquirer/test/unit/prompts/cap-project/fixtures/bookshop'.
    Error: Path to module not found.', attempt to load module @sap/cds from @sap/cds-dk threw error 'Error: Module path not found''}]
```

Where an install has not bee done on the project?

IainSAP (approved):
Thanks @Klaus-Keller 

- Approving as needed by new module `@sap-ux/odata-servivce-inquirer`
- Further error codes can be added as required
- Test as expected

packages/project-access/src/project/cap.ts
```diff
@@ -408,9 +408,11 @@ async function loadCdsModuleFromProject(capProjectPath: string, strict: boolean
         if (typeof cdsDependencyVersion === 'string') {
             const globalCdsVersion = cds.version;
             if (getMajorVersion(cdsDependencyVersion) !== getMajorVersion(globalCdsVersion)) {
-                throw Error(
+                const error = new Error(
                     `The @sap/cds major version (${cdsDependencyVersion}) specified in your CAP project is different to the @sap/cds version you have installed globally (${globalCdsVersion}). Please run 'npm install' on your CAP project to ensure that the correct CDS version is loaded.`
-                );
+                ) as Error & { code: string };
+                error.code = 'CDS_VERSION_MISMATCH';
```
IainSAP, line 414: Should these error code strings be exported as consts for import and re-use in other modules?

Klaus-Keller: We could add the error codes to JSDoc.

zdravko-georgiev (approved):
Code straightforward and easy to understand
Excellent test coverage
Did NOT tested manually

kranthie-sap (approved):
* Verified that changeset config exists
* Changes to `cap.ts` and corresponding unit test look good. It is odd that the `Error` type used doesn't have the `code` property exposed. The workaround to augment the type looks good. More information about the issue has been discussed along with various solutions are in this thread - https://stackoverflow.com/questions/51523509/in-typescript-how-do-you-make-a-distinction-between-node-and-vanilla-javascript

PR #1837, opened 2024-04-16
fix: i18n for CAPJava

marufrasully:
Get i18n bundles for CAPJava.

PR #1839, opened 2024-04-16
fix: i18n for CAPJava

marufrasully:
Get i18n bundles for CAPJava.


marufrasully:
covered in this PR https://github.com/SAP/open-ux-tools/pull/1837

PR #1840, opened 2024-04-17
chore(project-access): expose lib types

cianmSAP:
Expose the types for libraries


packages/project-access/src/index.ts
```diff
@@ -1,5 +1,5 @@
 export { FileName, DirName } from './constants';
-export { getFilePaths } from './file';
+export { getFilePaths, readJSON } from './file';
```
longieirl, line 2: Wondering if we should just export everything from here? As its providing all the `memFs` parameters as well.

cianmSAP: Is there a reason these functions have not been exposed yet @Klaus-Keller ?

devinea (dismissed):
Code changes lg. 
changeset 👍

Klaus-Keller (changes requested):
Please hold on with this PR, there are ongoing discussions about exporting low-level functionality such as  `readJSON()` through `@sap-ux/project-access`.

If no `mem-fs-editor` is required,  function `readJSON()` can be implemented by these two lines of code:

```typescript
import { promises as fs } from 'fs';

const parsed = JSON.parse(await fs.readFile(<PATH_TO_FILE>, { encoding: 'utf8' })) as <TYPE>;
```

That way no dependency to `@sap-ux/project-access` is required.

The risks I see when exporting such low-level functionality:
- Modules add a dependency to `@sap-ux/project-access` solely to use exported low-level functions, which is an unnecessary dependency because the alternative to use `promises` from `fs` is node built-in and does not need any additional dependency.
- Once published, we can’t make any incompatible changes without increasing the major version, e.g. if we do changes to the `mem-fs-editor,` which is also implemented into these functions. Ergo, we eliminate our flexibility to do internal changes for the sake of having one line of code reusable for others.

packages/project-access/src/index.ts
Klaus-Keller: -> https://github.com/SAP/open-ux-tools/pull/1840#pullrequestreview-2007513845

longieirl: Thank you for explaination, all makes sense.

cianmSAP:
> Please hold on with this PR, there are ongoing discussions about exporting low-level functionality such as `readJSON()` through `@sap-ux/project-access`.
> 
> If no `mem-fs-editor` is required, function `readJSON()` can be implemented by these two lines of code:
> 
> ```ts
> import { promises as fs } from 'fs';
> 
> const parsed = JSON.parse(await fs.readFile(<PATH_TO_FILE>, { encoding: 'utf8' })) as <TYPE>;
> ```
> 
> That way no dependency to `@sap-ux/project-access` is required.
> 
> The risks I see when exporting such low-level functionality:
> 
> * Modules add a dependency to `@sap-ux/project-access` solely to use exported low-level functions, which is an unnecessary dependency because the alternative to use `promises` from `fs` is node built-in and does not need any additional dependency.
> * Once published, we can’t make any incompatible changes without increasing the major version, e.g. if we do changes to the `mem-fs-editor,` which is also implemented into these functions. Ergo, we eliminate our flexibility to do internal changes for the sake of having one line of code reusable for others.

Thanks for the explanation. I've reverted the change to expose the functions from `./file`

This PR is now just to expose the types for libraries

devinea (approved):

PR #1842, opened 2024-04-17
Fix(preview-middleware): Add NavTargetResolution

### LINKED ISSUE #1841: BUG - preview middleware served launchpad not able to navigate to (fiori elements) application in UI5 122.x
Hey hey 👋🏻 

In UI5 >=1.122.x the launchpad served by the preview middleware is no longer able to open a Fiori Elements app.

Details see internal incident `DINC0116030`

Seems like `NavTargetResolution` is now required 🤷🏻 

### Expected results

App opens on click of tile

### Actual results

Error message:

```
App could not be opened because the SAP UI5 component of the application could not be loaded.
Failed to load UI5 component for navigation intent "#Samples-display"
```

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

heimwege:
fixes #1841 

I successfully tested a fiori elements app with 1.120.11 and 1.122.2, someone needs to test non-fe apps in 1.122.x and < 1.122.x before merge to ensure they are fine as well.


zdravko-georgiev (changes requested):
As discussed this issue seems to be UI5 related. Therefore a fix might be provided by UI5 team. Please don't merge yet.

Issue #1843, opened 2024-04-17
BUG - ui-components. UISections issue after resize window resize


815are:
### Steps to Reproduce
1. Open storybook;
2. Open `Splitter/Horizontal Sections`;
3. Resize splitter to left edge;
4. Do window resize;
5. Click on resize splitter and do not resize;

Problem - left pane becomes invisible:
![image](https://github.com/SAP/open-ux-tools/assets/90789422/8c89cafb-5328-4443-8363-da5b53f2cb3b)

### Expected results
Sections are visible and do not disappear

### Actual results
Left section disappears

### Screenshots
![image](https://github.com/SAP/open-ux-tools/assets/90789422/8c89cafb-5328-4443-8363-da5b53f2cb3b)

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1844, opened 2024-04-17
fix(ui-components): update cached section sizes

815are:
Issue #1843

There is scenario which involves following actions:
1. splitter resize;
2. window resize;
3. splitter resize without mouse move(mouse up+mouse down without mouse move);

1. Open storybook;
2. Open `Splitter/Horizontal Sections`;
3. Resize splitter to left edge;
4. Do window resize;
5. Click on resize splitter and do not resize;

Problem - left pane becomes invisible:
image

Problem was that `this.resizeSections` was not refreshed after window resize.


vadson71 (approved):
Code changes look good. Didn't test them. Covered by tests. Approved

Jimmy-Joseph19 (approved):
Code change looks good.
Covered with unit test
Changeset found.
Did not test locally.

broksy (approved):
Understandable change and fully covered with unit test.
Did not test locally.

Issue #1850, opened 2024-04-19
FEATURE - Add fiori generator shared module


cianmSAP:
_As a developer,
I want to have a central module for functions used by the fiori generator,
so that I reuse these functions in my development_

### Description
With future additions of the Fiori generator and it's sub-generators, there is a need for a central module to contain the functions which can be shared across the modules

### Technical Design

`@sap-ux/fiori-generator-shared`

### Tasks
- [ ] Add function for reading cap custom paths
- [ ] Unit Tests added
- [ ] Test Cases defined

cianmSAP:
Closing as module is published

PR #1851, opened 2024-04-19
feat(fiori-gen): adds new module @sap-ux/fiori-generator-shared

cianmSAP:
#1850

Adds new module @sap-ux/fiori-generator-shared


packages/fiori-generator-shared/src/cap/index.ts
```diff
@@ -0,0 +1,77 @@
...
+        db: 'db/',
+        srv: 'srv/'
+    };
+
+    const cdsrcPath = join(`${capProjectPath}/.cdsrc.json`);
+    const packageJsonPath = join(`${capProjectPath}/package.json`);
+
+    try {
+        for (const folder of Object.keys(capPaths)) {
+            if (!existsSync(join(`${capProjectPath}/${folder}`))) {
```
devinea, line 24: Don't expect we should check for folder existence here. Just continue merging configs in order.

devinea (commented):
New module setup looks good. 
Changeset ✅ 

one small comment.

devinea (approved):
Thanks for the updates.

packages/fiori-generator-shared/src/cap/index.ts
```diff
@@ -0,0 +1,75 @@
+import { existsSync, readFileSync } from 'fs';
```
IainSAP, line 1: Can we use fs promises instead of sync versions?

devinea: This function is to be sync only, right?
maybe it should be called `getCapFolderPathsSync`

packages/fiori-generator-shared/test/cap/index.test.ts
```diff
@@ -3,6 +3,7 @@ import { getCapFolderPaths } from '../../src/cap';
 import { join } from 'path';
 
 describe('getCapFolderPaths', () => {
+    jest.setTimeout(3000000);
```
IainSAP, line 6: Should be removed?

cianmSAP: Sorry merged to early by mistake

devinea (approved):
Re-approving

IainSAP (commented):
Thanks @cianmSAP 

- Cov is good

IainSAP (approved):

PR #1853, opened 2024-04-22
fix(axios-extension): rename SCENARIO to FIORI_TOOLS_SCENARIO

zdravko-georgiev:
Renames `SCENARIO` to `FIORI_TOOLS_SCENARIO` in order to be consistent to other env variables expected by the middlewares.


tobiasqueck (approved):
* changeset ok
* no objections against the new name

Klaus-Keller (approved):
Renaming for the sake of consistency is great. Approved.

Issue #1767, opened 2024-03-22
[Security] `follow-redirects` Information Exposure


tobiasqueck:
### Description
`follow-redirects` is an open source, drop-in replacement for Node's http and https modules that automatically follows redirects. In affected versions follow-redirects only clears authorization header during cross-domain redirect, but keep the proxy-authentication header which contains credentials too. This vulnerability may lead to credentials leak, but has been addressed in version 1.15.6. Users are advised to upgrade. There are no known workarounds for this vulnerability.

More information at https://security.snyk.io/vuln/SNYK-JS-FOLLOWREDIRECTS-6444610

### Tasks
- [ ] user overrides to remove the nested dependencies to the vulnerable version (#1766)
- [ ] Remove  "nx-cloud>axios@<1.6.8": ">=1.6.8" as soon as possible
- [ ] Remove  "@sap/bas-sdk>axios@<1.6.8": ">=1.6.8"as soon as possible
- [ ] Remove  "http-proxy>follow-redirects@<1.15.6": ">=1.15.6"as soon as possible
- [ ] Remove "socks-proxy-agent>socks@<2.8.1": ">=2.8.1"as soon as possible
- [ ] Remove  "ip@<2.0.1": ">=2.0.1" as soon as possible

PR #1860, opened 2024-04-24
fix(security): remove ip override

donal-tobin-sap:
- upgrade storybook
- in order to be able to remove the ip overrride

Issue: https://github.com/SAP/open-ux-tools/issues/1767


tobiasqueck (changes requested):
* no changeset but it is ok since the change does not affect delivered code
* please also update https://github.com/SAP/open-ux-tools/blob/fix/1767/upgrade-storybook/docs/version-overrides.md so that it matches the actual state

Klaus-Keller (approved):
Update and removal of version overrides look good, thanks @donal-tobin-sap

PR #1872, opened 2024-04-25
chore(tools-offline): update list of project dependencies

Klaus-Keller:
Update file `docs/tools-offline-instructions/info/projectModules.json` which is used in scripts to make SAP Fiori tools offline available.

### DIFF
--- a/docs/tools-offline-instructions/info/projectModules.json
+++ b/docs/tools-offline-instructions/info/projectModules.json
@@ -3,7 +3,10 @@
         "7.14.7"
     ],
     "@sap-ux/eslint-plugin-fiori-tools": [
-        "^0.3.1"
+        "^0.4.0"
+    ],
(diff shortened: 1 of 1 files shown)


docs/tools-offline-instructions/info/projectModules.json
```diff
@@ -81,6 +87,9 @@
         "^3.0.2",
         "3.0.2"
     ],
+    "sqlite3": [
```
longieirl, line 90: Why is this required?

Klaus-Keller: There is a script running to collect all dependencies from generated projects. Not sure if I should actively filter out anything. I PM you some details.

longieirl: Thanks for the offline explanation.

longieirl (approved):
Comments addressed, changes look good.

PR #1894, opened 2024-05-03
fix: isS4Cloud

tobiasqueck:
This PR fixes an incorrect implementation of the `isS4Cloud` function in the `axios-extension`. The function was incorrectly returning false if the client is configured for development (embedded Steampunk) and not extensibility.
While fixing it, I also removed a variable that was only exposed for testing and fixed the corresponding tests.

### DIFF
--- a/.changeset/violet-pianos-own.md
+++ b/.changeset/violet-pianos-own.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/axios-extension': patch
+---
+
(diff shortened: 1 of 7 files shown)


docirl (approved):
Changes look good and as expected
Tests updated and coverage is good
Not tested manually

longieirl (approved):
Changes look good.
Code appended with appropiate comments.
Test coverage looks good.

Issue #1897, opened 2024-05-04
BUG - Link to Guided Answer isn't clickable as the output is encoded


gregorwolf:
### Related Feature

Feature request: #1805

### Description

With the PR #1805 a link to the Guided Answer [How to validate ADT API services used for deployment configuration?](https://ga.support.sap.com/dtp/viewer/index.html#/tree/3046/actions/45995:45996:50742:46000:57266). But when an error occurs during deployment that provides a link to this Guided Answer it isn't helpful as the link can't be clicked do due it's encoding.

### Steps to Reproduce

1. Setup a project to be deployed to an ABAP System where the OData service /UI5/ABAP_REPOSITORY_SRV is not yet active
2. run `npm run deploy`
3. Confirm to deploy

### Expected results

```
error abap-deploy-task ZEXCEL_DU Change logging level to debug your issue
error abap-deploy-task ZEXCEL_DU        (see examples https://github.com/SAP/open-ux-tools/tree/main/packages/deploy-tooling#configuration-with-logging-enabled)
error builder:custom deploy-to-abap Request failed with status code 403. Guided Answers troubleshooting guide available at https://ga.support.sap.com/dtp/viewer/index.html#/tree/3046/actions/45995:45996:50742:46000:57266
error builder:custom deploy-to-abap Request failed with status code 403
```

### Actual results

```
error abap-deploy-task ZEXCEL_DU Change logging level to debug your issue
error abap-deploy-task ZEXCEL_DU        (see examples https://github.com/SAP/open-ux-tools/tree/main/packages/deploy-tooling#configuration-with-logging-enabled)
error builder:custom deploy-to-abap Request failed with status code 403. Guided Answers troubleshooting guide available at https:&#x2F;&#x2F;ga.support.sap.com&#x2F;dtp&#x2F;viewer&#x2F;index.html#&#x2F;tree&#x2F;3046&#x2F;actions&#x2F;57266
error builder:custom deploy-to-abap Request failed with status code 403
```

### Screenshots

![image](https://github.com/SAP/open-ux-tools/assets/2639535/973484a7-ae3f-42a0-a327-18919c56aa18)

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [X] Mac OS
- [ ] Windows
- [ ] Other

tobiasqueck:
@ianquigley-sap

tobiasqueck:
[...]

PR #1898, opened 2024-05-04
feat(axios-extension): Add PATH to debug output

gregorwolf:
Hello Open UX Tools,

during the Fiori Training I've delivered last week I came across the issue that I ran the deploy to an ABAP Backend but it failed with with an error. Setting `log: 4` did only provide this additional information:

```
Retrieving application ZEXCEL_DU, AxiosError: Request failed with status code 403
```

Only because I've developed the backend service using ABAP in Eclipse I've noticed the following error in the Feed Reader:

image

with this details:

image

After activating the service in transaction /IWFND/MAINT_SERVICE the issue is gone. But I think this should be made easier by adding information which PATH is called in the ABAP backend to the debug output. It might be even better if the PATH is already shown in the error.

Best Regards
Gregor


longieirl:
Typically this is the GA page we provide when a HTTP 403 is returend;

https://ga.support.sap.com/dtp/viewer/index.html#/tree/3046/actions/45995:45996:50742:46000

Once the prerequisites are met, it resolves most problems.

longieirl (approved):
Code changes look good.
Not sure how this will help other ABAP developers? In the sense, they will know to enable the appropiate transaction.

gregorwolf:
As filed in #1897 this Guided Answer isn't clickable. Also the client app should provide detail information at least in debug mode to help the user directly.

devinea:
@gregorwolf Thanks for your contribution. 
I've copied your commits onto a branch on https://github.com/SAP/open-ux-tools so that the build pipeline runs fully.
https://github.com/SAP/open-ux-tools/pull/1915

Issue #1901, opened 2024-05-07
TBI - Create a separate Adaptation template file writer for Migration


korotkovao:
### Description (include screenshots)
Create a separate Adaptation template writer for Migration

### Value
Maintainability. This could be extended for migration in the future without having any impact on the existing Adaptation files generation.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

PR #1903, opened 2024-05-07
feat(adp-tooling): add migration adaptation template writer

korotkovao:
TBI - Create a separate Adaptation file writer for Migration #1901


packages/adp-tooling/test/unit/writer/index.test.ts
```diff
@@ -86,5 +87,10 @@ describe('ADP writer', () => {
                 )
             ).toMatchSnapshot();
         });
+        test('migrate minimal config', async () => {
```
tobiasqueck, line 90: Could you please put this into a new describe block. We usually have a describe for the whole file and then nested describes for each function we test, and in there the specific test cases.

For the test itself, could you please add a fixture containing a simple "old" project and then run the migrate command?

packages/adp-tooling/src/writer/index.ts
```diff
@@ -42,26 +59,41 @@ export async function generate(basePath: string, config: AdpWriterConfig, fs?: E
...
+/**
+ * Writes the adp-project template to the mem-fs-editor instance during migration.
+ *
+ * @param basePath - the base path
+ * @param config - the writer configuration
+ * @param fs - the memfs editor instance
+ * @returns the updated memfs editor instance
+ */
+
+export async function migrateAdp(basePath: string, config: AdpWriterConfig, fs?: Editor): Promise<Editor> {
```
tobiasqueck, line 83: What is the difference between `migrateAdp` and `generate`? I understand what the different use cases are but I don't see the difference in the code.
Shouldn't you handle the possibility of an existing `ui5.yaml` or `package.json`?

tobiasqueck (changes requested):

packages/adp-tooling/src/writer/index.ts
korotkovao: The idea is to extend it in the near future, exactly for what you said
>Shouldn't you handle the possibility of an existing ui5.yaml or package.json?

packages/adp-tooling/src/writer/index.ts
```diff
@@ -42,26 +59,41 @@ export async function generate(basePath: string, config: AdpWriterConfig, fs?: E
...
+    if (!fs) {
+        fs = create(createStorage());
     }
 
+    const fullConfig = setDefaults(config);
+
+    // copy all files except app discriptor
+    fs.copyTpl(join(tmplPath, '**/*.*'), join(basePath), fullConfig, undefined, {
+        globOptions: { dot: true, ignore: ['manifest.appdescr_variant'] },
```
korotkovao, line 92: @tobiasqueck 
Thank you for the review. I need to exclude the `manifest.appdescr_variant` file, we don't want to call `generate` as is, as `generate` function will create that file. We are creating a `migrate` function to handle migration logic and only write the files that are needed.
The current implementation is to not handle existing ui5.yaml and package.json files, but to generate the new ones. We had a discussion on this with @devinea and agreed we can do it later if needed.

tobiasqueck: aah... there is the difference :) I missed that when looking at the code. Then it is fine

packages/adp-tooling/test/unit/writer/index.test.ts
tobiasqueck: Please remove the `pom.xml` from your example. That is only required internally, so it doesn't need to be here for a test.

tobiasqueck: Shouldn't your migrate function also delete the `xs-app.json` and the `.che` folder?

korotkovao: Thanks again, I have updated the code.

korotkovao: > Shouldn't your migrate function also delete the `xs-app.json` and the `.che` folder?

I think we should keep that separate along with other migration logic. Here we would only generate the needed adp template files.

devinea: Code has been updated to remove `.che` and `neo-app.json`. 👍

devinea (dismissed):
Code changes look as expected to support migration for webIDE created ADP projects. 
Covered by tests. 
changeset ✅

packages/adp-tooling/test/unit/writer/index.test.ts
```diff
@@ -87,4 +90,18 @@ describe('ADP writer', () => {
             ).toMatchSnapshot();
         });
     });
+
+    describe('migrate', () => {
```
tobiasqueck, line 94: The test is fully correct, however, I'd like to slightly simplify it so that it is easier to understand and enhance it in the future.

I would suggest to remove all changes from the root describe, and just add the more concise 

```typescript
    describe('migrate', () => {
        const migrateInputDir = join(__dirname, '../../fixtures/webide-adaptation-project');

        test('migrate minimal config', async () => {
            const projectDir = join(outputDir, 'webide-adaptation-project');
            fs.copy(migrateInputDir, projectDir, { globOptions: { dot: true } });

            await migrate(projectDir, config, fs);
            expect(fs.dump(projectDir)).toMatchSnapshot();
            expect(fs.exists(join(projectDir, '.che/project.json'))).toBeFalsy();
            expect(fs.exists(join(projectDir, 'neo-app.json'))).toBeFalsy();
        });
    });
```
Reasoning:
* everything that is only relevant for migration is in the migration describe block
* structure and names (e.g. `projectDir`) is exactly like it is done in generate
* the expect calls before the actual test are removed because "we trust" our test setup
* remove the two checks of existing files because these two are checked with the snapshot (note: I kept the check for the deleted files because they don't show up in the snapshot since they are added and deleted on the same editor instance)

tobiasqueck (changes requested):
minor change request for the test structure but the rest looks great

packages/adp-tooling/test/unit/writer/index.test.ts
korotkovao: Thank you for the suggestion and the thorough explanation, appreciate it. I have added the changes.

tobiasqueck (approved):
* changeset ok
* code clear and easy to understand
* proper test coverage

devinea (approved):
Re-approving

korotkovao:
Thank you @tobiasqueck @devinea

Issue #1881, opened 2024-05-02
TBI - (odata-service-writer): remove synchronizationMode for manifest.json template ui5 versions 1.110


martinjanus-sap:
### Description (include screenshots)
The v4 model colleagues deprecated the synchronizationMode. 
OData V4 apps should have `synchronizationMode` removed from the manifest.json starting from minUi5version `1.110.x`

```
               {
                "dataSource": "mainService",
                "settings": {
                    "synchronizationMode": "None",
                    "operationMode": "Server",
                    "autoExpandSelect": true,
                    "earlyRequests": true }
                }
```

Add change in the writer to [templates/extend/manifest.json](https://github.com/SAP/open-ux-tools/blob/9728d4433e0690b572efec85c1bb3b2e7a23e2b2/packages/odata-service-writer/templates/extend/manifest.json#L39) to conditionally write synchronizationMode based on the minUI5Version

### Value
- Maintaining manifest.json 

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Conditionally write synchronizationMode based on minUI5Version
- [ ] Relevant tests

martinjanus-sap:
Validated internally on staging as part of 28025

PR #1910, opened 2024-05-09
feat(odata-service-writer) synchronizationMode conditionally removed

martinjanus-sap:
#1881

- synchronizationMode is conditionally removed based on `minUI5Version` from manifest. 
- As it is deprecated from version `1.110.0`, it is no longer added for newer versions but still supported for versions `1.110.0` and older.


packages/odata-service-writer/src/updates.ts
```diff
@@ -17,7 +18,7 @@ export function updateManifest(basePath: string, service: OdataService, fs: Edit
     // manifest.json
     const manifestPath = join(basePath, 'webapp', 'manifest.json');
     // Get component app id
-    const manifest = fs.readJSON(manifestPath);
+    const manifest = fs.readJSON(manifestPath) as any as Manifest;
```
IainSAP, line 21:
```suggestion
    const manifest = fs.readJSON(manifestPath) as unknown as Manifest;
```

IainSAP (approved):
Thanks @martinjanus-sap 

- OK as an initial approach to versioning service updates but may need a more declarative approach if more is added
- Good test cov
- Not tested locally

Issue #1911, opened 2024-05-09
BUG - @sap-ux/odata-service-inquirer doesnt ensure prompt texts are loaded


IainSAP:
### Description
When calling the API `getPrompts()` the prompt texts may not be loaded unless there is some additional delay/waits on the consumer side. For example calling the function and immediately testing the prompt texts:

```
const { prompts: odataServicePrompts, answers: odataServiceAnswers } = await getPrompts();
const capAndFilePrompts = odataServicePrompts.filter((prompt) => prompt.name !== promptNames.datasourceType);
```
Results in prompts without texts:

```
...
{
      "guiOptions": {
          "breadcrumb": true,
          "mandatory": true,
        },
        "guiType": "file-browser",
        "message": "prompts.metadataFile.message",
        "name": "metadataFilePath",
        "type": "input",
        "validate": [Function],
        "when": [Function],
      },
...
```

We should ensure the texts are loaded prior to returning.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1912, opened 2024-05-09
fix(odata-service-inquirer): Ensure prompt texts are loaded

IainSAP:
https://github.com/SAP/open-ux-tools/issues/1911

- Wait for load of i18n texts when calling `getPrompts` otherwise we may not have texts
- Add tests to cover
- Add tests for `prompt-helper`
- Update readme example


devinea (approved):
code changes look good. 
Covered by tests
changeset ✅

longieirl (approved):
Changes look good.
Changeset present.
Explaination of why the change is required.
Good test coverage.

Issue #1793, opened 2024-03-27
TBI - To separate the mock server logic from odata-service-writer


longieirl:
### Description (include screenshots)

There  are two ways to generate a ui5-mock.yaml configuration;

1. using odata-service-writer 
2. using mock-server-writer

The following references, are using `odata-service-writer` to create the ui5-mock.yaml configuration;

https://github.com/SAP/open-ux-tools/blob/434025c7516669dc29b076275d41c5fa723275f9/packages/fiori-freestyle-writer/src/index.ts#L6
https://github.com/SAP/open-ux-tools/blob/434025c7516669dc29b076275d41c5fa723275f9/packages/fiori-elements-writer/src/index.ts#L7
https://github.com/SAP/open-ux-tools/blob/434025c7516669dc29b076275d41c5fa723275f9/packages/ui5-config/src/ui5config.ts#L284

These instances rely on the`@sap-ux/ui5-config` to return the `sap-fe-mockserver` configuration and update what ever path is being given to it i.e. 

```
    public addMockServerMiddleware(path?: string, annotationsConfig?: MockserverConfig['annotations']): this {
        this.document.appendTo({
            path: 'server.customMiddleware',
            value: getMockServerMiddlewareConfig(path, annotationsConfig)
        });
        return this;
    }
```

This flow does not align with the `mock-server-writer` which also implements the same logic. Additionaly, the method `addMockServerMiddleware` is only ever run against a ui5-mock.yaml configuration and never a ui5.yaml configuration.

### Value
Clear separate of concern.

### Architecture Elaboration
Does it requires architecture elaboration? Yes

### Tasks
TO BE DEFINED

longieirl:
Confirmed, working as expected when consumed in tools-suite.

PR #1913, opened 2024-05-09
fix(odata-service-writer): fs param mockserver logic

martinjanus-sap:
Fix for #1793 
- added `fs` for `await getWebappPath(basePath, fs)` to support memfs

### DIFF
--- a/packages/odata-service-writer/src/index.ts
+++ b/packages/odata-service-writer/src/index.ts
@@ -108,7 +108,7 @@ async function generate(basePath: string, service: OdataService, fs?: Editor): P
     if (service.metadata) {
         // copy existing `ui5.yaml` as starting point for ui5-mock.yaml
         if (paths.ui5Yaml && ui5Config) {
-            const webappPath = await getWebappPath(basePath);
+            const webappPath = await getWebappPath(basePath, fs);
             const config = {
                 webappPath: webappPath,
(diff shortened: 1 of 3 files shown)


devinea (dismissed):
Thanks @martinjanus-sap

devinea (dismissed):
Re-approving

packages/odata-service-writer/src/index.ts
```diff
@@ -108,7 +108,7 @@ async function generate(basePath: string, service: OdataService, fs?: Editor): P
     if (service.metadata) {
         // copy existing `ui5.yaml` as starting point for ui5-mock.yaml
         if (paths.ui5Yaml && ui5Config) {
-            const webappPath = await getWebappPath(basePath);
+            const webappPath = await getWebappPath(basePath, fs);
```
longieirl, line 111: I'm wondering if its worth adding a spy to an existing test to ensure the fs parameter is passed to `getWebappPath` i.e.
```
import * as projectAccess from '@sap-ux/project-access'
const getWebappPathMock = jest.spyOn(projectAccess, 'getWebappPathMock');
expect(getWebappPathMock).toHaveBeenCalledWith(expect.anything(), fs);
```
Or even if you can count the number of parameters passed is two.

longieirl (approved):
Re-approving, tests updated as requested.

PR #1915, opened 2024-05-09
feat(axios-extension): Add PATH to debug output

devinea:
Copied from #1898 by @gregorwolf 

Hello Open UX Tools,

during the Fiori Training I've delivered last week I came across the issue that I ran the deploy to an ABAP Backend but it failed with with an error. Setting `log: 4` did only provide this additional information:

```
Retrieving application ZEXCEL_DU, AxiosError: Request failed with status code 403
```

Only because I've developed the backend service using ABAP in Eclipse I've noticed the following error in the Feed Reader:

image

with this details:

image

After activating the service in transaction /IWFND/MAINT_SERVICE the issue is gone. But I think this should be made easier by adding information which PATH is called in the ABAP backend to the debug output. It might be even better if the PATH is already shown in the error.

Best Regards
Gregor


longieirl (approved):
Changes are good.
Changeset present.

IainSAP (approved):
Minor log update.

devinea:
Thanks @gregorwolf

PR #1917, opened 2024-05-10
fix: small changes to simplify consumption

Klaus-Keller:
- Add const `SchemaName` that contain Fiori Tools File System (ftfs) schema.
- Export `ODataVersion` as alias to `NonNullable<ManifestNamespace.Setting['odataVersion']>`
- Export `CustomItem` from `ui5-config`

### DIFF
--- a/.changeset/cold-coats-do.md
+++ b/.changeset/cold-coats-do.md
@@ -1,0 +1,6 @@
+---
+'@sap-ux/project-access': minor
+'@sap-ux/ui5-config': patch
+---
+
+Improvements for consumption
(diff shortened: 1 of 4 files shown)


zdravko-georgiev (approved):
Change is small and easy to understand
Only types updates

johannes-kolbe (approved):
- Types update only, code understandable
- Changeset provided

tobiasqueck (approved):
* 'minor' seems a bit much for that change but I am ok
* types and consts make sense
* nothin to test, just consts and types

Issue #1916, opened 2024-05-10
FEATURE - Reducing eslint warnings in preview-middleware-client


GDamyanov:
_As a **developer**,
When I  **execute pnpm lint script on preview-middleware-client package**
I want to **have little number of eslint warnings**

### Description

Currently lint report generate 209 warnings on "preview-middleware-client". So the purpose of this issue is to reduce them as much as possible. 

### Technical Design
"preview-middleware-client" module needs to be enhanced

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a cloned open-ux-tools repo
**when** I execute pnpm lint on preview-middleware-client module
**then** I should get reduced number of eslint warnings in the report.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

heimwege:
Closing this one as we are constantly working on reducing the number of lint issues and this one is almost 8 months old.

PR #1918, opened 2024-05-13
feat(preview-middleware-client): Reduce eslint warnings in preview-middleware-client

GDamyanov:
Feat for #1916 

- Reduce "preview-middleware-client" package eslint warnings


IvoSG (dismissed):
Easy to understand changes
Code looks good

zdravko-georgiev (dismissed):
Please define the missing types instead of typecasting everything

packages/preview-middleware-client/src/adp/controllers/AddFragment.controller.ts
```diff
@@ -134,9 +136,9 @@ export default class AddFragment extends BaseDialog {
         const source = event.getSource<Button>();
         source.setEnabled(false);
 
-        const fragmentName = this.model.getProperty('/newFragmentName');
-        const index = parseInt(this.model.getProperty('/selectedIndex'));
-        const targetAggregation = this.model.getProperty('/selectedAggregation/value');
+        const fragmentName = this.model.getProperty('/newFragmentName') as string;
```
tobiasqueck, line 139: I am not a big fan of the `as string` everywhere. Could we define it more centrally? Something like https://github.com/SAP/open-ux-tools/compare/feat/1916/reduce-eslint-warnings...feat/1916/reduce-eslint-warnings-proposal?expand=1 ?

Btw, is `selectedIndex` really a string? I think it is a number at least that is what you set in `setProperty`

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
```diff
@@ -243,8 +243,8 @@ export default class ControllerExtension extends BaseDialog {
             const data = await getExistingController(controllerName);
             return data;
         } catch (e) {
-            MessageToast.show(e.message, { duration: 5000 });
-            throw new Error(e.message);
+            MessageToast.show((e as Error).message, { duration: 5000 });
```
tobiasqueck, line 246: Technically this is wrong. In JavaScript I can throw everything as an error, so I could have something like `throw "Everything is bad";` that would result in an error in your catch clause.
So, I think you need to introduce a typeguard and then use it here.

```typescript
function isErrorWithMessage(e: unknown): e is Error {
  return (e as Error).message !== undefined;
}

// ...

try {
...
} catch (e) {
  if (isErrorWithMessage(e)) {
     MessageToast.show(e.message, { duration: 5000 });
     throw e.message;
  } else {
    throw e;
  }
}
```

Since you have the same catch-pattern all over the place, you could consider putting the whole catch block into a reusable function like `showErrorToast`

testojs: Another approach would be to introduce a function that accepts `e` and always returns an `Error` object - if `e` is an instance of `Error` just return `e`, otherwise - stringify `e` and return a new `Error` with the stringified version of `e`. And than just read the `message` ...

packages/preview-middleware-client/src/adp/controllers/AddFragment.controller.ts
tobiasqueck: We could also do it purely with typings i.e. we could enforce that the correct type is set by removing the generic `setProperty` function and defining the allowed values per key e.g. 
```typescript
type AddFragmentModel = Omit<JSONModel, 'setProperty'> & {
    getProperty(sPath: '/newFragmentName'): string;
    getProperty(sPath: '/selectedIndex'): string;
    getProperty(sPath: '/selectedAggregation/value'): string;

    setProperty(sPath: '/indexHandlingFlag', value: boolean): void;
};
```

packages/preview-middleware-client/src/adp/api-handler.ts
```diff
@@ -86,7 +87,7 @@ export async function request<T>(endpoint: ApiEndpoints, method: RequestMethod,
                 return response.json() as T;
         }
     } catch (e) {
-        throw new Error(e.message);
+        throw new Error((e as Error)?.message);
```
tobiasqueck, line 90: Why are we doing this?

packages/preview-middleware-client/src/adp/controllers/BaseDialog.controller.ts
```diff
@@ -61,7 +62,7 @@ export default abstract class BaseDialog extends Controller {
         const beginBtn = this.dialog.getBeginButton();
 
         const fragmentName: string = input.getValue();
-        const fragmentList: { fragmentName: string }[] = this.model.getProperty('/fragmentList');
+        const fragmentList: Fragments = this.model.getProperty('/fragmentList') as Fragments;
```
tobiasqueck, line 65: please also get rid of this one as you did it for the others

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
```diff
@@ -243,8 +252,9 @@ export default class ControllerExtension extends BaseDialog {
             const data = await getExistingController(controllerName);
             return data;
         } catch (e) {
-            MessageToast.show(e.message, { duration: 5000 });
-            throw new Error(e.message);
+            const error = getError(e);
```
tobiasqueck, line 255: Why don't you use `handleError` here?

packages/preview-middleware-client/src/adp/controllers/AddFragment.controller.ts
```diff
@@ -216,8 +225,9 @@ export default class AddFragment extends BaseDialog {
 
             this.model.setProperty('/fragmentList', fragments);
         } catch (e) {
-            MessageToast.show(e.message);
-            throw new Error(e.message);
+            const error = getError(e);
```
tobiasqueck, line 228: why don't you use `handleErrror` here

tobiasqueck (changes requested):
Please consistently use the type for the model as well as the `handleError` function that you introduced

packages/preview-middleware-client/src/adp/controllers/AddFragment.controller.ts
GDamyanov: Fixed in [39505b0](https://github.com/SAP/open-ux-tools/pull/1918/commits/39505b01b7d9cb4ccdc43b1ab21387e06780dc7a)

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
GDamyanov: Fixed in [39505b0](https://github.com/SAP/open-ux-tools/pull/1918/commits/39505b01b7d9cb4ccdc43b1ab21387e06780dc7a)

packages/preview-middleware-client/src/adp/api-handler.ts
GDamyanov: Fixed in [39505b0](https://github.com/SAP/open-ux-tools/pull/1918/commits/39505b01b7d9cb4ccdc43b1ab21387e06780dc7a)

packages/preview-middleware-client/src/adp/controllers/BaseDialog.controller.ts
GDamyanov: Fixed in [39505b0](https://github.com/SAP/open-ux-tools/pull/1918/commits/39505b01b7d9cb4ccdc43b1ab21387e06780dc7a)

tobiasqueck (approved):
* changeset ok
* code quality improved (could be even better but it clearly improved)

testojs (approved):
* comments addressed
* code improved
* not tested manually

tobiasqueck (approved):

Issue #1920, opened 2024-05-13
BUG - issues in service generation in @sap-ux/axios-extension


docirl:
### Related Feature
Feature: https://github.com/SAP/open-ux-tools/issues/1893

### Description
1. Business object retrieval uses incorrect param in request. Param should be `releaseState` instead of `releaseStatus`
2. Export Business Object type and UI service generator class for use by consumers

### Steps to Reproduce
1. Use business object retrieval api
2. Results are not filtered by param `releaseState`

### Expected results
Business Object results are correctly filtered

### Actual results
Business objects are returned not filtered, and too many results are returned

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

docirl:
Fix is now merged.

PR #1921, opened 2024-05-13
fix(axios-extension): use correct param for business object request, add exports

docirl:
https://github.com/SAP/open-ux-tools/issues/1920

### DIFF
--- a/packages/axios-extension/src/abap/adt-catalog/services/businessobjects-service.ts
+++ b/packages/axios-extension/src/abap/adt-catalog/services/businessobjects-service.ts
@@ -28,7 +28,7 @@ export class BusinessObjectsService extends AdtService {
      * @param maxResults - The maximum number of business objects to be returned.
      * @returns A list of business objects.
      */
-    public async getBusinessObjects(maxResults?: number): Promise<BusinessObject[]> {
+    public async getBusinessObjects(maxResults = 10000): Promise<BusinessObject[]> {
         const config = {
             headers: {
                 Accept: 'application/xml'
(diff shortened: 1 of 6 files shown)


packages/axios-extension/src/abap/adt-catalog/services/businessobjects-service.ts
```diff
@@ -38,7 +38,7 @@ export class BusinessObjectsService extends AdtService {
                 query: `*`,
                 maxResults: maxResults || 10000,
                 objectType: 'BDEF',
-                releaseStatus: 'USE_IN_CLOUD_DEVELOPMENT'
+                releaseState: 'USE_IN_CLOUD_DEVELOPMENT'
```
slin-sap, line 41: Not related to this line. But the method signature. 

Default value is preferred over optional parameter. E.g. you can write

`getBusinessObjects(maxResults = 10000)`

instead of 

`getBusinessObjects(maxResults?: number)`

Hence, you also don't need

`maxResults: maxResults || 10000,`

You can use:

```
 params: {
                operation: 'quickSearch',
                query: `*`,
                maxResult,
                ...
}
```

slin-sap (changes requested):

packages/axios-extension/src/abap/adt-catalog/services/businessobjects-service.ts
docirl: Thanks, I'll make the change

slin-sap (approved):
- changeset is added
- minor change that looks good to me
- unit test has been updated accordingly

tobiasqueck (approved):
* changeset ok
* good test coverage

PR #1925, opened 2024-05-14
feat: Add Adaptation Project's Change Data Source generator prompting

mmilko01:
https://github.com/SAP/open-ux-tools/issues/2002

- Add prompting relevant to Change Data Source generator of Adaptation Project
- Enhance axios-extension with functionality to fetch application manifests
- Add CLI command for executing Change Data Source for Adaptation Projects


packages/create/src/cli/change-data-source/change-data-source.ts
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Add a new sub-command to change the data source of an adaptation project to the given command.
+ *
+ * @param {Command} cmd - The command to add the change-data-source sub-command to.
+ */
+export function addChangeDataSourceCommand(cmd: Command): void {
+    cmd.command('change-data-source')
+        .argument('<path>', 'path to the Adaptation Project folder')
+        .requiredOption('--url [url]', 'url pointing to the target system containing the original app')
+        .requiredOption('--username [username]', 'username for the target system containing the original app')
```
tobiasqueck, line 27: Please don't ask allow providing the password as CLI parameter. Most proper shells store these inputs in a plain history file.

packages/adp-tooling/src/base/prompt.ts
```diff
@@ -16,6 +16,8 @@ export type PromptDefaults = {
     ft?: boolean;
     package?: string;
     transport?: string;
+    username?: string;
```
tobiasqueck, line 19: Please reuse https://github.com/SAP/open-ux-tools/blob/main/packages/adp-tooling/src/base/prompt.ts#L115 for connecting to a system and authenticating. We need more than just basic auth support.

packages/adp-tooling/src/prompt/changes/prompts/change-data-source.ts
```diff
@@ -0,0 +1,298 @@
+import type { InputQuestion } from '@sap-ux/inquirer-common';
+import type { Answers, ListQuestion, Question, NumberQuestion } from 'inquirer';
+import type { AdpChangeDataSourceAnswers, AdpChangeDataSourceQuestions } from '../../../types';
+import { t } from '../../../i18n';
+import { validateURI, validateODataServices } from '../../validators';
+
+const ODATA_V2_PREFIX_ABAP = '/sap/opu/odata/';
```
tobiasqueck, line 7: I don't understand what these prefixes do. Especially on an ABAP system I can define an alias for any of my ICF endpoints.

packages/adp-tooling/src/prompt/changes/prompts/change-data-source.ts
```diff
@@ -0,0 +1,298 @@
...
+
+/**
+ * Gets the questions for changing the data source.
+ *
+ * @param {AdpChangeDataSourceQuestions} data - the data for the questions
+ * @returns {Question<AdpChangeDataSourceAnswers>[]} - the questions
+ */
+export function getQuestions(data: AdpChangeDataSourceQuestions): Question<AdpChangeDataSourceAnswers>[] {
+    return [
+        {
```
tobiasqueck, line 114: Why are we asking whether we are in safe mode or not?

packages/adp-tooling/src/prompt/changes/prompts/change-data-source.ts
```diff
@@ -0,0 +1,298 @@
...
+            guiOptions: {
+                type: 'label',
+                applyDefaultWhenDirty: true
+            }
+        } as InputQuestion<AdpChangeDataSourceAnswers>,
+        {
+            type: 'input',
+            name: 'oDataURIV2Info',
+            message: () => {
+                const versionOne = 'V2';
```
tobiasqueck, line 172: We have a constant/type for odata version (I think in project-access) please reuse it.

packages/create/package.json
```diff
@@ -40,13 +40,18 @@
         "@sap-ux/logger": "workspace:*",
         "@sap-ux/mockserver-config-writer": "workspace:*",
         "@sap-ux/project-access": "workspace:*",
+        "@sap-ux/ui5-config": "0.22.3",
         "chalk": "4.1.2",
         "commander": "9.4.0",
         "diff": "5.1.0",
+        "inquirer": "8.2.6",
```
tobiasqueck, line 47: The `inquirer` dependency is a no-go here. The create module shall not depend on it.
Additionally, I don't think you need mem-fs and mem-fs-editor

packages/create/src/cli/change-data-source/change-data-source.ts
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * @param {string} basePath - The path to the adaptation project.
+ * @param {PromptDefaults} defaults - The default values for the prompts.
+ */
+async function changeDataSource(basePath: string, defaults: PromptDefaults): Promise<void> {
+    if (defaults.url && defaults.username && defaults.password && defaults.reference) {
+        const url = new URL(defaults.url);
+        const provider = createForAbap({
+            baseURL: url.origin,
+            ignoreCertErrors: true,
```
tobiasqueck, line 46: no, sorry, you need to ask the user or allow to provide it as CLI parameter, but you cannot disable SSL validation by default.

packages/create/src/cli/change-data-source/change-data-source.ts
```diff
@@ -0,0 +1,106 @@
...
+import { readFileSync } from 'fs';
+import { join } from 'path';
+import { UI5Config } from '@sap-ux/ui5-config';
+
+/**
+ * Add a new sub-command to change the data source of an adaptation project to the given command.
+ *
+ * @param {Command} cmd - The command to add the change-data-source sub-command to.
+ */
+export function addChangeDataSourceCommand(cmd: Command): void {
```
tobiasqueck, line 23: Check the other commands, we usually also have a mode that will not write the change but just shows the diffs so people can see what would be happening before they do it.

tobiasqueck (changes requested):
I know it is only in draft state but here are couple of things you should look at.

packages/axios-extension/src/abap/app-index-service.ts
```diff
@@ -34,4 +35,34 @@ export abstract class AppIndexService extends Axios implements Service {
...
+     * @returns {Promise<string>} - URL to the app manifest.
+     */
+    public async getManifestUrl(appId: string): Promise<string> {
+        let result = '';
+        const params = {
+            'id': appId
+        };
+        try {
+            const response = await this.get('/ui5_app_info_json', { params });
+            const responseValues: any = Object.values(JSON.parse(response.data));
```
tobiasqueck, line 52: No any please.
Why do we check for `manifestUrl` or `manifest`? Is a different property used in different versions?

packages/create/README.md
```diff
@@ -24,6 +24,15 @@ Calling `sap-ux add` allows adding a feature to a project.
 ## remove
 Calling `sap-ux remove` allows removing a feature to a project.
 
+## adp-add
```
tobiasqueck, line 27: I would prefer adding it to `add` and not have the additional `adp-add`.

tobiasqueck: And then for the full command `add data-source-change`.

packages/create/src/cli/change-data-source/change-data-source.ts
```diff
@@ -0,0 +1,152 @@
...
+            if (config.environment === 'CF') {
+                throw new Error('Changing data source is not supported for CF projects.');
+            }
+        }
+
+        const variant = JSON.parse(readFileSync(join(basePath, 'webapp', 'manifest.appdescr_variant'), 'utf-8'));
+        const ui5Config = await UI5Config.newInstance(readFileSync(join(basePath, 'ui5.yaml'), 'utf-8'));
+        const { destination, url, client } =
+            ui5Config.findCustomMiddleware<{ backend: Array<{ destination?: string; url?: string; client?: string }> }>(
+                'fiori-tools-proxy'
```
tobiasqueck, line 59: You should use the preview middleware instead because thats the middleware that is used during a preview to merge the appdescr with the manifest

packages/create/src/cli/change-data-source/change-data-source.ts
```diff
@@ -0,0 +1,152 @@
...
+        let target;
+        if (destination) {
+            target = { destination };
+        } else if (url) {
+            target = { url, client };
+        } else {
+            throw new Error('No system configuration found in ui5.yaml');
+        }
+
+        const provider = await createAbapServiceProvider(
```
tobiasqueck, line 71: Please use the system-access module here to handle all the different authentication mechanisms and fallbacks.

mmilko01: Could you please give a bit more details about your idea here? 
As far as we understand from you comment, we shouldn't do any check whether we have destination or system URL in the ui5.yaml, but directly pass whatever we have to createAbapServiceProvider and let it do the checks for us. Is our understanding correct here?

packages/adp-tooling/src/prompt/changes/prompts/change-data-source.ts
mmilko01: Fixed with commit: fcb2eb3c603db4f005757221894faf5bf36e714e

packages/adp-tooling/src/prompt/changes/prompts/change-data-source.ts
mmilko01: The prompt about safe mode was of type label and it was used to only show a message.
It was removed along the other prompts of type label with commit: fcb2eb3c603db4f005757221894faf5bf36e714e

packages/adp-tooling/src/prompt/changes/prompts/change-data-source.ts
mmilko01: Fixed with commit : fcb2eb3c603db4f005757221894faf5bf36e714e

packages/create/package.json
mmilko01: Fixed with commit: 496998c4003cd968bb4cf6346e9a2dd6bfd2bcd2

packages/create/src/cli/change-data-source/change-data-source.ts
mmilko01: Fixed with commit: f52e8525f3561ddc294690f963bd287bdb42d811

packages/adp-tooling/src/base/prompt.ts
mmilko01: Fix with: b71f9c20e861416fd5f2fb9fc05dfd336700dafb

packages/axios-extension/src/abap/app-index-service.ts
mmilko01: We check for both due to historic reasons, as the name of the property varies in the manifest.json of some applications.
Fixed with: 0b3bbd403dd4a46cc48f14ac48822e9a55d4b136

packages/create/src/cli/change-data-source/change-data-source.ts
mmilko01: Fixed with: b335f48373d6f448c5f9ef6201e220ed30bbca88

packages/create/README.md
mmilko01: Fixed with: 8fbf2b395b9db4e830c8f261b6eefcac83f56f4a

mmilko01: I didn't notice the first time that you prefered shuffling the words in the command. Could you give me some more details on the convention in such cases, as I couldn't notice any similar command already existing?
Also, I will update the readme.

tobiasqueck: Sure, the idea was to have commands consisting of two parts (1) activity and (2) object. As activities, we have so far `add`, `remove` and `generate` (`eject` likely coming soon). The object should describe what is being modified. In your case you add a new flex change that changes the data source of an app, so I would say a "datasource change" therefore, the suggested command `add data-source-change`.
I would expect that we soon also get more things like `*-change` objects for other flex changes

packages/create/src/cli/change-data-source/change-data-source.ts
mmilko01: @tobiasqueck Just as a kind reminder, could you please take a look at my previous comment?

tobiasqueck: Sorry, I missed your question. Just checked you code again, you are doing 
```typescript
import { createAbapServiceProvider } from '@sap-ux/system-access';
// ...
const provider = await createAbapServiceProvider(
        target,
        {
            ignoreCertErrors: defaults.ignoreCertErrors
        },
        true,
        logger
    );
```
which is exactly what I want. As result, you will get a valid connection (BAS or VSCode) based on either destinations or stored system configs, or the user is being prompted.

Did you change it already, or did I misread your code when reviewing?

packages/create/README.md
mmilko01: The other manifest change wizards are as follows:
- add-annotations-to-odata 
- add-component-usages
- add-new-model
- change-inbound

For example, in the change file resulting from add-annotations-to-odata wizard, the ChangeType property will be appdescr_app_addAnnotationsToOData. How do you think we could fit to the logic you described?

packages/create/src/cli/change-data-source/change-data-source.ts
mmilko01: As far as I can see the code was there when you did the review. If I understand you correctly, there is not need for a change here?

packages/adp-tooling/src/prompt/changes/prompts/change-data-source.ts
tobiasqueck: Looks much better however, there is still one usage at https://github.com/SAP/open-ux-tools/blob/fcb2eb3c603db4f005757221894faf5bf36e714e/packages/adp-tooling/src/prompts/change-data-source/index.ts#L70.

packages/create/README.md
tobiasqueck: Hmm ... good question, I'd go for descriptive concise names since the user is supposed to enter them in the cli.

First shot:
* `add data-source-change`
* `add new-model-change` (btw, when is the difference between that one and the data source?)
* `add inbound-nav-change`
* `add component-usage-change`
* `add local-annotations` (I removed change here because that one does more than just adding a change)

Quite frankly, they are not concise or intuitive, maybe we should "play" with the verb and introduce "change" and have something like:
* `change data-source`
* `add new-model`
* `change inbound-navigation`
* `add component-usage`
* `add local-annotations`

packages/adp-tooling/src/prompt/changes/prompts/change-data-source.ts
mmilko01: This also no longer exists in the latest version of the code.
Fixed with: ecdfc9efae33793c05a54fe53f96a3921f703efd

packages/adp-tooling/src/types.ts
```diff
@@ -345,9 +334,32 @@ export interface AdpProjectData {
...
     adp: {
         safeMode: boolean;
         environment: OperationsType;
     };
-}
\ No newline at end of file
+}
+
+export interface PreviewConfigAdp {
+    adp?: {
```
tobiasqueck, line 358: That type is already defined at https://github.com/SAP/open-ux-tools/blob/main/packages/adp-tooling/src/types.ts#L20.

Btw, it also contains a flag for ignoring cert errors - maybe you could also read that because if it is set to true here (by the user) you can use it and don't have to prompt the users.

packages/adp-tooling/src/writer/changes/writers/data-source-writer.ts
```diff
@@ -56,28 +56,32 @@ export class DataSourceWriter implements IWriter<DataSourceData> {
...
             this.projectPath,
             change,
-            `id_${data.timestamp}_changeDataSource.change`,
+            `id_${timestamp}_changeDataSource.change`,
             this.fs,
             DirName.Manifest
         );
 
-        const shouldAddAnnotation = service.annotationUri && service.annotationUri.length > 0;
+        const shouldAddAnnotation = annotationId && annotationUri;
```
tobiasqueck, line 75: I'd skip this line and put the expression directly in the if-block. Otherwise, people (like me) wonder where else in the code you use it.

packages/axios-extension/src/abap/app-index-service.ts
```diff
@@ -34,4 +35,29 @@ export abstract class AppIndexService extends Axios implements Service {
...
         return JSON.parse(response.data).results as AppIndex;
     }
+
+    /**
+     * Gets the URL to the app manifest with the specified id.
+     *
+     * @param {string} appId - The id of the app.
+     * @returns {Promise<string>} - URL to the app manifest.
+     */
+    public async getManifestUrl(appId: string): Promise<string> {
```
tobiasqueck, line 45: I understand that this is fully sufficient for your use case but the call returns so many more important information that I am sure we need in other places as well, therefore, could you please enhance that function to be something like `getAppInfo` and return an object of type
```typescript
interface AppInfo {
   name: string;
   url: string;
   manifestUrl: string // or manifest, whatever is the newer name of the property and you take care of the mapping here;
  asyncHints: Array<...>;
}
```

packages/axios-extension/src/abap/lrep-service.ts
```diff
@@ -256,6 +256,28 @@ export class LayeredRepositoryService extends Axios implements Service {
         }
     }
 
+    /**
```
tobiasqueck, line 259: I don't think that one is a good idea. You are already in possession of the full url and replace parts in here because you know that the url will be going to this service. That is quite some implicit knowledge.

I'd recommend to remove this function and where you currently call it, just call `provider.get(manifestUrl)`

tobiasqueck (changes requested):

packages/create/src/cli/add/change-data-source.ts
```diff
@@ -0,0 +1,146 @@
...
+            logger.error(`Authentication failed. Please check your credentials. Login attempts left: ${loginAttempts}`);
+            await changeDataSource(basePath, defaults, simulate);
+            return;
+        }
+        logger.debug(error);
+    }
+}
+
+/**
+ *
```
nikmace, line 88: Missing JSDoc description and returns

packages/create/src/cli/add/change-data-source.ts
```diff
@@ -0,0 +1,146 @@
...
+ * @param basePath
+ * @param defaults
+ * @param logger
+ * @param variant
+ * @returns
+ */
+async function getManifest(basePath: string, defaults: PromptDefaults, logger: ToolsLogger, variant: any) {
```
nikmace, line 95: `any` should be changed to actual type

packages/create/src/cli/add/change-data-source.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    const appIndexService = provider.getAppIndex();
+    const manifestUrl = await appIndexService.getManifestUrl(variant.reference);
+    const lrepService = provider.getLayeredRepository();
+    const manifest = await lrepService.getManifest(manifestUrl);
+    return manifest;
+}
+
+/**
+ *
```
nikmace, line 126: Missing JSDoc description and returns, also add return type

packages/create/src/cli/add/change-data-source.ts
```diff
@@ -0,0 +1,146 @@
...
+ * @param basePath
+ * @returns
+ */
+function getVariant(basePath: string) {
+    return JSON.parse(readFileSync(join(basePath, 'webapp', 'manifest.appdescr_variant'), 'utf-8'));
+}
+
+/**
+ *
```
nikmace, line 135: Missing JSDoc description and returns, also add return type void

packages/create/test/unit/cli/add/change-data-source.test.ts
```diff
@@ -0,0 +1,240 @@
...
+                getLayeredRepository: jest.fn().mockReturnValue({
+                    getManifest: abapServicesMock.getManifest
+                })
+            };
+        }
+    };
+});
+
+describe('adp-add/change-data-source', () => {
+    // Mocks
```
nikmace, line 49: All those comments are unneeded below, just leave empty space

packages/create/test/unit/common/prompts.test.ts
```diff
@@ -0,0 +1,68 @@
...
+        message: 'List of strings',
+        choices: [
+            { name: 'first', value: 'a' },
+            { name: 'second', value: 'b' }
+        ],
+        default: 'b',
+        validate: (value: string) => value === 'a'
+    } as ListQuestion
+};
+//(choice) => ({ title: choice.name, value: choice.value })
```
nikmace, line 25: Is this comment still needed?

packages/create/test/unit/common/prompts.test.ts
```diff
@@ -0,0 +1,68 @@
...
+                ],
+                validate: expect.any(Function),
+                initial: expect.any(Function)
+            });
+            expect((result.initial as Function)()).toBe(1);
+        });
+    });
+
+    describe('promptYUIQuestions', () => {
+        test.todo('TBD');
```
nikmace, line 66: Missing test suite?

nikmace (changes requested):
Also, please follow the conventional commits, here is a good [cheatsheet](https://kapeli.com/cheat_sheets/Conventional_Commits.docset/Contents/Resources/Documents/index). You are **misusing** the chore commit.

.changeset/slow-ears-argue.md
```diff
@@ -0,0 +1,8 @@
+---
+"@sap-ux/adp-tooling": patch
+"@sap-ux/axios-extension": patch
+"@sap-ux/create": patch
+"@sap-ux/types": patch
```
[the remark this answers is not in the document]
mmilko01: Fixed with: c1c0fa562d5d8a5a5696579883d075f19d054c04

packages/create/test/unit/common/prompts.test.ts
mmilko01: @tobiasqueck Could you please check?

packages/create/test/unit/common/prompts.test.ts
mmilko01: @tobiasqueck Could you please check?

packages/adp-tooling/src/base/change-utils.ts
```diff
@@ -200,16 +201,41 @@ export function getGenericChange(
...
+
+/**
+ * Constructs a generic change object based on provided parameters.
+ *
+ * @param {DescriptorVariant} variant the app descriptor variant
+ * @param {number} timestamp - The timestamp
+ * @param {object} content - The content of the change to be applied
+ * @param {ChangeType} changeType - The type of the change
+ * @returns An object representing the change
+ */
```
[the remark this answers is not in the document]
mmilko01: Fixed with: f9e4c62b1bcdcf6bb755e72a032ac252fc9a2907 and 399ed354dfcc942423ecbf0f9f19ea973445e833

packages/adp-tooling/src/prompts/change-data-source/index.ts
```diff
@@ -0,0 +1,58 @@
...
+import { isNotEmptyString } from '../../base/helper';
+
+/**
+ * Gets the prompts for changing the data source.
+ *
+ * @param {Record<string, ManifestNamespace.DataSource>} datasources data sources from the manifest
+ * @returns {YUIQuestion<ChangeDataSourceAnswers>[]} the questions/prompts
+ */
+export function getPrompts(
+    datasources: Record<string, ManifestNamespace.DataSource>
```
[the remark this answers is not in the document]
mmilko01: Fixed with: 01ea9ff8d5b8b50622d2cb9c299a9560cded6c15

packages/adp-tooling/src/prompts/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Gets the data sources that are of type OData.
+ *
+ * @param {{ [k: string]: ManifestNamespace.DataSource } | undefined} dataSources - The data sources from the manifest.
+ * @returns {DataSource[]} - The data sources that are of type OData.
+ */
+export function getTargetDataSources(
+    dataSources: { [k: string]: ManifestNamespace.DataSource } | undefined
+): DataSource[] {
```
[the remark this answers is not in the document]
mmilko01: Fixed with: a9dd8dcb190649d0f55839d9927b4e79372edf99

packages/adp-tooling/src/prompts/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+    if (!dataSources) {
+        return [];
+    }
+    return Object.entries(dataSources)
+        .filter((dS) => dS[1]?.type === 'OData')
+        .map((dS) => {
```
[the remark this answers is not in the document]
mmilko01: Fixed with: a9dd8dcb190649d0f55839d9927b4e79372edf99

packages/adp-tooling/src/prompts/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+                uri: dS[1]?.uri,
+                annotations: dS[1]?.settings?.annotations ?? []
+            };
+        });
+}
+
+/**
+ * Gets the data sources that are of type ODataAnnotation.
+ *
+ * @param {{ [k: string]: ManifestNamespace.DataSource } | undefined} dataSources - The data sources from the manifest.
```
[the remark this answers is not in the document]
mmilko01: Fixed with: a9dd8dcb190649d0f55839d9927b4e79372edf99

packages/adp-tooling/src/prompts/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+ */
+export function getTargetODataAnnotations(
+    dataSources: { [k: string]: ManifestNamespace.DataSource } | undefined
+): KeyValueString {
+    const annotations: { [key: string]: string } = {};
+    if (!dataSources) {
+        return annotations;
+    }
+    return Object.entries(dataSources)
+        .filter((dS) => dS[1]?.type === 'ODataAnnotation')
```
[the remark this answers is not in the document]
mmilko01: Fixed with: a9dd8dcb190649d0f55839d9927b4e79372edf99

packages/create/src/cli/add/change-data-source.ts
mmilko01: Fixed with: cdd8157df89ff219fe223b367ae69befd7b86d13

packages/create/src/cli/add/change-data-source.ts
mmilko01: Fixed with: cdd8157df89ff219fe223b367ae69befd7b86d13

packages/create/src/cli/add/change-data-source.ts
mmilko01: Fixed with: 01ea9ff8d5b8b50622d2cb9c299a9560cded6c15

packages/create/src/cli/add/change-data-source.ts
mmilko01: Fixed with: 01ea9ff8d5b8b50622d2cb9c299a9560cded6c15

packages/create/test/unit/cli/add/change-data-source.test.ts
mmilko01: It seems that similar comments are used in some of the tests (e.g., https://github.com/SAP/open-ux-tools/blob/567c715072a4e0d81163bf8c7cd1fe000f222011/packages/create/test/unit/cli/generate/adaptation-project.test.ts#L32). Regardless, the comments are removed from our tests with: 01ea9ff8d5b8b50622d2cb9c299a9560cded6c15

packages/adp-tooling/src/prompts/utils.ts
```diff
@@ -0,0 +1,96 @@
+import type { ManifestNamespace } from '@sap-ux/project-access';
```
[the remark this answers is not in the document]
mmilko01: I think that the name is fine when we consider that it is under `prompts`.  But, it is under consideration to move the code from `utils `to a more common place as suggested by Tobias.

packages/adp-tooling/package.json
```diff
@@ -36,29 +36,34 @@
...
         "@types/ejs": "3.1.2",
         "@types/express": "4.17.21",
+        "@types/inquirer": "8.2.6",
         "@types/mem-fs": "1.1.2",
         "@types/mem-fs-editor": "7.0.1",
         "@types/prompts": "2.4.4",
         "@types/supertest": "2.0.12",
-        "@types/adm-zip": "0.5.5",
-        "express": "4.19.2",
+        "@ui5/manifest": "1.61.0",
```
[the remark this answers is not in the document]
mmilko01: Fixed with: d274783f43ea1b051ece4671bd503b5b3707cadb

packages/adp-tooling/src/prompts/utils.ts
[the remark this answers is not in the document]
mmilko01: It seems that we cannot reuse any functionality, but I moved `filterDataSourcesByType` to @sap-ux/project-access. 
Fixed with: 000067194c8c372136c6c802763822fdba9f0295

packages/adp-tooling/src/index.ts
```diff
@@ -3,3 +3,5 @@ export * from './preview/adp-preview';
 export { generate, migrate } from './writer';
 export { generateChange } from './writer/editors';
 export { promptGeneratorInput, PromptDefaults } from './base/prompt';
+export * from './prompts/utils';
```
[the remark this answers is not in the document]
mmilko01: Fixed with: 000067194c8c372136c6c802763822fdba9f0295

packages/adp-tooling/package.json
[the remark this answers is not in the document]
mmilko01: Fixed with: d274783f43ea1b051ece4671bd503b5b3707cadb

packages/adp-tooling/src/base/change-utils.ts
```diff
@@ -200,16 +201,41 @@ export function getGenericChange(
...
+export function getChange(
```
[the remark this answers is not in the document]
mmilko01: Currently, `getGenericChange` expects too many parameters that are not used afterwards. The idea here is to refactor this code and remove` getGenericChange`. Currently, `getChange` is basically overloading `getGenericChange`. The current state of the code is just temporary until we move all the prompts to this repository.

packages/adp-tooling/src/writer/changes/writers/data-source-writer.ts
mmilko01: Fixed with: 9f6d864a51cb9db3fbfbc3f8e54db1a0822695b5

packages/axios-extension/src/abap/lrep-service.ts
mmilko01: I think that using directly `provider.get(manifestUrl)` would lead to a need of handling the errors in the CLI. There is also a possibility that we would need this method for the other generators. 
The replace is no longer needed with: a83e51f8dfe9d80500d2884300109b86c866e406

packages/axios-extension/src/abap/app-index-service.ts
mmilko01: Good point, the implementation is now enhanced as per your recommendation with: 3d7fc54d7441290ff13f8020f49c734e930b5c02

packages/create/src/cli/change/change-data-source.ts
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Get the manifest of the base application.
+ *
+ * @param {string} basePath - The path to the adaptation project.
+ * @param {ToolsLogger} logger - The logger.
+ * @param {DescriptorVariant} variant - The app descriptor variant.
+ * @returns {Promise<Manifest>} The manifest.
+ */
+async function getManifest(basePath: string, logger: ToolsLogger, variant: DescriptorVariant): Promise<Manifest> {
```
nikmace, line 82: This method will be reused in other "generators", so it should be moved to a some common file?

nikmace (changes requested):
Please resolve **all** of the issues in **Sonar** and also _increase the coverage_ a bit, because it is barely making 81% and some functions are untested.

Also, **do not resolve the review comments** yourself, only the reviewer should resolve his own comments if he agrees with your statement.

packages/create/src/cli/change/change-data-source.ts
mmilko01: - The Sonar issues and code coverage are currently coming from a functionality that @tobiasqueck provided for us. This was originally part of its own PR (https://github.com/SAP/open-ux-tools/pull/1940) . This point needs to be discussed with Tobias to confirm whether those changes would be part of this PR or the other.
- Regarding the rosolving of the comments, I followed the documentation here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/about-pull-request-reviews#resolving-conversations. It suggests that owner of the PR resolves the comments, making it easier to find conversations that still need to be address. We could discuss if following this process is not prefered.
- Currently, there is a common method `getManifest` in `@sap-ux/axios-extension`, which can be reused in other generator. The logic in the `getManifest` method you point to is more specific to this generator. But if it is the case that this logic can be reused when other generators are migrated, it can be refactored then.

nikmace: Every organization, and repository has its own set of rules, for open-ux-tools it is the following: The reviewer resolves his comments which is logical. This is like a conversation, I ask something and you answer. But who should say whether the answer is sufficient, acceptable, or correct? I think the reviewer since he is the one who asked. If you decide to resolve the comment yourself that means first the comment is hidden now, and when I go to check your answer I can't find it and I have to go through all the code to check whether it was resolved. Second, if you replied you automatically think this is a correct answer, but this is a conversation, and I may disagree with your answer. I hope that clears it out. :)

If you received approval but I for example forgot to resolve some of my comments, then you can resolve them yourself.

mmilko01: Yes, this made it clear :). @testojs made me aware that this is mentioned in [open-ux-tools/CONTRIBUTING.md at 59410cffdc3de151c0c35aa445a586c56e62f321 · SAP/open-ux-tools (github.com)](https://github.com/SAP/open-ux-tools/blob/59410cffdc3de151c0c35aa445a586c56e62f321/CONTRIBUTING.md#how-to-contribute). I will have it in mind for future comments. Btw, you could expand all resolved comments in a PR by Alt + Click / Option + Click on one of them. Hopefully, this will make it easier to find my answers in this PR

nikmace: For no leaving this comment open TBD with Tobias, since I do not understand why those Sonar and tests do not need to be written for the functionality he provided to you.

packages/adp-tooling/src/base/change-utils.ts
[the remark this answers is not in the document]
tobiasqueck: Sorry why do you want to push temporary code? Please do it properly right away. Having two functions with different names but the same description is just not acceptable.
Clearly describe their difference or write the code as you intent it in the future.

packages/axios-extension/src/abap/lrep-service.ts
tobiasqueck: Please revert a83e51f. The input parameter for the `getLayeredRepository` is there to define an alias if the service is available on a different url, but you misuse it for a hack.  If you call `getLayeredRepository('/')` then the returned service will not work with any of the other offered functions.

packages/create/src/cli/change/change-data-source.ts
tobiasqueck: @nikmace there is nothing to understand, the tests are just missing - need to do them. I was working on the functionality in a separate branch ensuring that the approach works for two different ongoing implementations. Both work with it now, so next will be the tests, and then it is either a separate PR or part of this one (currently I prefer separate.

packages/adp-tooling/src/base/change-utils.ts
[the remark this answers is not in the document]
mmilko01: Thanks for the hint!
Fixed with: f6529d211f163f6b7c357314c67e9058f95fd961

packages/axios-extension/src/abap/lrep-service.ts
mmilko01: Fixed with: c0d4b4265d63513c9a5639844b3d45ab8f1567a4

packages/create/src/cli/change/change-data-source.ts
tobiasqueck: @nikmace the quality of that code is now as required

nikmace (approved):
Code looks good
Test coverage is good
Did not test manually

testojs:
@tobiasqueck Can you please review it again?
Thanks!

packages/create/src/cli/change/change-data-source.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/**
+ * Check if the project is a CF project.
+ *
+ * @param {string} basePath - The path to the adaptation project.
+ * @throws {Error} If the project is a CF project.
+ */
+function checkEnvironment(basePath: string): void {
+    const configJsonPath = join(basePath, '.adp', 'config.json');
```
tobiasqueck, line 93: I would also check if the project actually is an adp and not an app - I think `project-access` has a function to check if it is adp

packages/create/package.json
```diff
@@ -37,12 +37,17 @@
         "@sap-ux/app-config-writer": "workspace:*",
         "@sap-ux/cap-config-writer": "workspace:*",
         "@sap-ux/cards-editor-config-writer": "workspace:*",
+        "@sap-ux/inquirer-common": "workspace:*",
         "@sap-ux/logger": "workspace:*",
         "@sap-ux/mockserver-config-writer": "workspace:*",
         "@sap-ux/project-access": "workspace:*",
+        "@sap-ux/system-access": "workspace:*",
+        "@sap-ux/ui5-config": "0.22.3",
```
tobiasqueck, line 45: please use `workspace:*` also for ui5-config

packages/create/README.md
```diff
@@ -21,6 +21,15 @@ npx sap-ux
 ## add
 Calling `sap-ux add` allows adding a feature to a project.
 
+## change
+Calling `sap-ux change` allows changing a feature of a project.
+
+### data-source
+Calling `sap-ux change data-source` allows replacing the OData Source of the base application in an adaptation project.  
+```sh
+sap-ux change data-source /path/to/adaptation-project
```
tobiasqueck, line 30: Maybe you could mention that the path is optional and that the cwd is used if not provided

tobiasqueck (changes requested):
* changeset ok
* code is easy to read and understandable
* a few minor change requests

I will do manual testing while you can address my review comment.

packages/create/src/cli/change/change-data-source.ts
```diff
@@ -0,0 +1,100 @@
...
+async function changeDataSource(basePath: string, defaults: PromptDefaults, simulate: boolean): Promise<void> {
+    const logger = getLogger();
+    try {
+        if (!basePath) {
+            basePath = process.cwd();
+        }
+        checkEnvironment(basePath);
+
+        const variant = getVariant(basePath);
+        const ui5Conf = await UI5Config.newInstance(readFileSync(join(basePath, 'ui5.yaml'), 'utf-8'));
```
tobiasqueck, line 41: Please add the UI5 yaml file name as optional CLI parameter - if not provided use `ui5.yaml` because the name of the file could be anything.

packages/adp-tooling/src/prompts/change-data-source/index.ts
```diff
@@ -0,0 +1,58 @@
...
+            },
+            validate: isNotEmptyString,
+            when: !!dataSourceIds.length,
+            store: false
+        } as InputQuestion<ChangeDataSourceAnswers>,
+        {
+            type: 'number',
+            name: 'maxAge',
+            message: t('prompts.maxAgeLabel'),
+            when: (answers: ChangeDataSourceAnswers) => answers.uri !== ''
```
tobiasqueck, line 47: Why do we need the `maxAge` question? I haven't seen that setting in any UI5 app (ever). Is this something that might have been useful in the past but is obsolete now?

packages/adp-tooling/src/prompts/change-data-source/index.ts
```diff
@@ -0,0 +1,58 @@
...
+        } as NumberQuestion<ChangeDataSourceAnswers>,
+        {
+            type: 'input',
+            name: 'annotationUri',
+            message: t('prompts.oDataAnnotationSourceURILabel'),
```
tobiasqueck, line 52: Could we please indicate in the label that it is not required to be entered.

packages/create/src/cli/change/change-data-source.ts
tobiasqueck: I'd recommend `-c --config` to be consistent with the ui5 tooling

packages/create/src/cli/change/change-data-source.ts
testojs: I checked the whole `open-ux-tools` repo but I cannot find such a function.

packages/create/package.json
testojs: Changed in cd1b20bbdddd1c595d925b5276ce529ca7ab8fa5

packages/create/README.md
testojs: Mentioned with cd1b20bbdddd1c595d925b5276ce529ca7ab8fa5

packages/create/src/cli/change/change-data-source.ts
tobiasqueck: @Klaus-Keller did I make that up? I thought we have a function in the `project-access` to identify ADPs?

packages/adp-tooling/src/prompts/change-data-source/index.ts
testojs: In Yeoman UI's `guiOptions` all mandatory questions have `mandatory` flag set to `true`. This will render mandatory asterisk in the UI which is sufficient for our use-case.
The annotation URI and max age are optional so they are `mandatory: false` by default.
Maybe the generic prompt converter can add this info to the label in the simple prompt lib case.

packages/adp-tooling/src/prompts/change-data-source/index.ts
testojs: Not my expertise but it is described in the documentation as optional:
https://help.sap.com/docs/bas/developing-sap-fiori-app-in-sap-business-application-studio/replacing-odata-service?q=maxAge

packages/create/src/cli/change/change-data-source.ts
```diff
@@ -0,0 +1,100 @@
...
+        const adp = ui5Conf.findCustomMiddleware<{ adp: AdpPreviewConfig }>('fiori-tools-preview')?.configuration?.adp;
```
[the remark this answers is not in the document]
testojs: What do you mean? What is the issue here?

packages/create/src/cli/change/change-data-source.ts
testojs: It could not be anything. We are generating this configuration file during project creation and later we expect it to be `ui5.yaml`.

tobiasqueck: While it is the ui5.yaml during generation, every developer can change it at any time, the name of the config file is configurable for every command relying on the ui5 CLI. I, for example, usually have multiple ui5 files when testing against different systems.

packages/create/src/cli/change/change-data-source.ts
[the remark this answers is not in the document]
tobiasqueck: If you generate the project with the open source middleware used directly and not wrapped in the `@sap/ux-ui5-tooling` (which is a valid usecase) then it doesn't work because your logic above cannot find it.

packages/adp-tooling/src/prompts/change-data-source/index.ts
tobiasqueck: Good idea, I will contribute it.

packages/adp-tooling/src/prompts/change-data-source/index.ts
tobiasqueck: I'd still remove it because if an app developer never uses it, it is unlikely to me that someone would set it in an adaptation project, so the question is more confusing then helpful, however, I leave that choice to you.

packages/adp-tooling/src/prompts/change-data-source/index.ts
tobiasqueck: https://github.com/SAP/open-ux-tools/pull/1925/commits/770da73d49d8f47186debf42d5c0287307f5cdc0

packages/create/src/cli/change/change-data-source.ts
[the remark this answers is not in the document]
mmilko01: Fixed with: fd93e76b7055e9d142e2d54416fe7dbf03fa2f1b

packages/create/src/cli/change/change-data-source.ts
mmilko01: Fixed with: 110e617d071c60c49534231288f4d092dd1391b3

packages/adp-tooling/src/prompts/change-data-source/index.ts
mmilko01: I discussed with our PO and we agreed that we would like to keep this questions as optional, to be in sync with the documentation. If this is okay with you?

packages/create/src/cli/change/change-data-source.ts
mmilko01: I think I found the function you mentioned in `project-access`. I tested locally and it seems to be working as expected. Could you please take a look at: 4a3eb7207c181990d27a88c33639ca8d43c98d95

tobiasqueck: Looks good!

tobiasqueck (approved):
* changeset ok
* code quality is good
* all requested changes from the previews review have been applied
* local tests successful

tobiasqueck (approved):

nikmace (approved):
Code looks good
Test coverage is good
Did not test manually

Issue #1919, opened 2024-05-13
BUG - Fragment files are not created together with an addXML change in onChangeRequest handler


nikmace:
### Description
When you have created an addXML change with a fragment with some name in the same session and then deleted this fragment. The mem-fs caches this path and does not create the fragment file with the same name (even though it is deleted from the computer's fs).

### Steps to Reproduce
1. Start an ADP project
2. Add a new Fragment, and delete it locally.
3. Try to create a fragment with the same name again.
4. Only the change file is created

### Expected results
XML Fragment with the same name is created along with the change

### Actual results
Only the change is created without any fragment

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #1926, opened 2024-05-14
fix: Fragment files are not created together with an addXML change in onChangeRequest handler

nikmace:
Fix for #1919.

- Removes the check for the fragment that already exists in mem fs. Check is done on the UI so the user cannot create an XML fragment with the same name.


GDamyanov (approved):
- Simple change.
- Code looks good.
- Good test coverage
- Tested manually

zdravko-georgiev (approved):
simple and straightforward change
did NOT tested manually

Issue #971, opened 2023-03-31
FEATURE - Fiori preview middleware


tobiasqueck:
As a **developer**,
I want to **preview my Fiori applications and adaptation projects locally in an FLP sandbox without having to maintain test files locally**,
so that I **don't have to worry about changes of these files in the future.**

### Description
The functionality available in the `@sap/ux-ui5-tooling` preview middleware (https://www.npmjs.com/package/@sap/ux-ui5-tooling#4-preview) is to be published to open source. It then needs to be extended not only be usable for variant creation but the simple preview with a sandbox FLP. As result, it is not necessary anymore to generate the `test/flpSandbox.html`, consequently, we can easily update it (through the middleware) with default them or other best practices.
Additionally, the published preview middleware needs to handle adaptation projects, so that a developer used to Fiori app development gets the same experienced when developing a Fiori adaptation project.

#### axios-extension
* enhance the lrep service to allow generating a merged `manifest.json` based on an app descriptor variant file

#### ui5-proxy-middleware
* remove the requirement to have a `manifest.json` in the project to determine the ui5 version

#### fiori-preview-middleware (new module)
* **preview**
  * define ejs template for the flp sandbox
  * enable flexibility to the hosted path
  * use project manifest for configurations
* **adp**
  * on startup use the axios-extension to fetch the merge manifest
  * generate the preview flp with the information from the merge manifest
  * add proxy to load the required files from the reference application from the backend

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a Fiori application project with a ui5.yaml configured with the new middleware (and no html files for testing)
**when** I execute `ui5 serve`
**then** a sandboxed FLP opens with my local app running in it.

**Given** I have a Fiori adaptation project with a ui5.yaml with the new middleware properly configured for adaptation projects (and no html files for testing)
**when** I execute `ui5 serve`
**then** a sandboxed FLP opens with my local adaptation project merged with the reference app running in it.

**Given** I need to add project specific code to my FLP sandbox
**when** I execute a command to eject the `flpSandbox.html`
**then** an `flpSandbox.html` file is added to my project and I can modify it for my needs.

### Notes
Since the requested feature is quite complex and combines existing functionality from different modules, it needs to be  sequentially implemented and released.

### Tasks
- [x] Extract only the simple preview functionality from `@sap/ux-ui5-tooling` and make it better configurable (v0.1) - #1093
[...]

PR #1930, opened 2024-05-15
feat(preview-middleware): Introduce check for existence of HTML files

ytpo-lyne:
Feat for #971 

- Introduce check for the existence of the HTML files


packages/preview-middleware/src/ui5/middleware.ts
```diff
@@ -47,11 +47,16 @@ async function createRouter(
...
     } else {
-        const manifest = await resources.rootProject.byPath('/manifest.json');
-        if (manifest) {
-            await flp.init(JSON.parse(await manifest.getString()));
+        const flpPath = config.flp.path ?? '/test/flpSandbox.html';
+        if (await resources.rootProject.byPath(flpPath)) {
+            logger.info('File to render preview found on file system.');
         } else {
-            throw new Error('No manifest.json found.');
+            const manifest = await resources.rootProject.byPath('/manifest.json');
```
zdravko-georgiev, line 54: I just noticed, this is wrong, we need to adapt the function https://github.com/SAP/open-ux-tools/blob/f79c7a1ead5577b50fe35f8fd64ee64d9c1e60cc/packages/preview-middleware/src/base/flp.ts#L346

zdravko-georgiev (changes requested):

packages/preview-middleware/src/ui5/middleware.ts
tobiasqueck: Yeah, that change needs to be reverted because it won't work for adaptation projects. The manifest needs to be provided during init.

tobiasqueck: Actually, I am confused what this change does..

tobiasqueck: Sorry, 3rd comment (deleted the other 2), but it should be later in the routes, when the file is loaded. There should already be code that handles the situation and returns the local file if present.

zdravko-georgiev: thanks @tobiasqueck will do.

packages/preview-middleware/test/fixtures/flpExists/package.json
```diff
@@ -0,0 +1,16 @@
+{
```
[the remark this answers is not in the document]
zdravko-georgiev: Done with https://github.com/SAP/open-ux-tools/pull/1930/commits/4b81e4e8431ed3e97b313dc6ee5d2bba5d6e3795

packages/preview-middleware/src/ui5/middleware.ts
zdravko-georgiev: https://github.com/SAP/open-ux-tools/pull/1930/commits/9cd77089776dfb45a2138bc98d3866a4ef5c6fd1

packages/preview-middleware/src/base/flp.ts
```diff
@@ -79,7 +79,7 @@ const DEFAULT_THEME = 'sap_horizon';
 /**
  * Default path for mounting the local FLP.
  */
-const DEFAULT_PATH = '/test/flp.html';
+const DEFAULT_PATH = '/test/flpSandbox.html';
```
tobiasqueck, line 82: Ehm ... no. It might be the default when wrapped by the `@sap/ux-ui5-tooling` but the default in the open-ux-tools has been `/test/flp.html` from the beginning, so changing this is quite a braking change (for no reason).

tobiasqueck (changes requested):
Sorry, but you introduced a breaking change that is not related to the actual problem you are trying to solve.

packages/preview-middleware/src/base/flp.ts
zdravko-georgiev: OK, will revert the change 👍

zdravko-georgiev: Done [ee99bad](https://github.com/SAP/open-ux-tools/pull/1930/commits/ee99bada8a7b30cd34a047ea9e91d66cbc661f10)

tobiasqueck (approved):
Looks good now

zdravko-georgiev (approved):
Simple and easy to understand
Excellent test coverage
Manual testing ✅

PR #1931, opened 2024-05-15
fix: CPE does not start UI Adaptation for ui5 versions lower than 1.120 for Adaptation Projects

### LINKED ISSUE #1929: BUG - CPE does not start UI Adaptation for ui5 versions lower than 1.120 for Adaptation Projects
### Related Feature
Feature request: #1874

### Description
CPE breaks for Adaptation Projects with the ui5 version lower than 1.120.

This import is causing the error in `packages/preview-middleware-client/src/adp/init.ts`: 

```ts
import UI5ElementRegistry from 'sap/ui/core/ElementRegistry';
```

### Steps to Reproduce
1. Start and ADP with ui5 version 1.108.31 for example
2. Open dev tools and look at the console
3. See the error (screenshot below)

### Expected results
UI Adaptation is started by RTA

### Actual results
RTA does not start.

### Screenshots

Screenshot 2024-05-15 at 13 48 05

Screenshot 2024-05-15 at 13 49 30

OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

GDamyanov:
fixes #1929 

I successfully tested with UI5 Versions lower and bigger than 1.120

### DIFF
--- a/packages/preview-middleware-client/src/adp/init.ts
+++ b/packages/preview-middleware-client/src/adp/init.ts
@@ -1,7 +1,5 @@
 import log from 'sap/base/Log';
 import type RuntimeAuthoring from 'sap/ui/rta/RuntimeAuthoring';
-import UI5ElementRegistry from 'sap/ui/core/ElementRegistry';
-
 import init from '../cpe/init';
 import { initDialogs } from './init-dialogs';
 import {
@@ -46,7 +44,7 @@ export default async function (rta: RuntimeAuthoring) {
         }
     );
 
-    const syncViewsIds = getAllSyncViewsIds();
+    const syncViewsIds = await getAllSyncViewsIds(minor);
     initDialogs(rta, syncViewsIds);
 
     if (minor > 77) {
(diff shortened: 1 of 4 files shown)


nikmace (commented):
Changes look good
Test coverage good
Did not test manually

packages/preview-middleware-client/src/adp/init.ts
```diff
@@ -46,7 +44,7 @@ export default async function (rta: RuntimeAuthoring) {
         }
     );
 
-    const syncViewsIds = getAllSyncViewsIds();
+    const syncViewsIds = await getAllSyncViewsIds(minor);
```
voicis, line 47: This seems to be executed before the app is completely loaded which means that maybe some of the views are not loaded. But this is not related to this particular fix.

voicis (approved):
- Code changes look good
- Test coverage is good
- Tested locally with UI5 `1.20.14` and `1.123.1`

nikmace (approved):
Changes look good
Test coverage good
Did not test manually

PR #1936, opened 2024-05-16
fix(odata-service-inquirer): misspelled i18n key

IainSAP:
- Fix misspelled 18n key


longieirl (approved):
You should be pair programming with AI!
Smallest change yet!

devinea (approved):

PR #1937, opened 2024-05-17
fix: fetch maintained UI5 version for UI integration

marufrasully:
This PR fetch maintained UI5 versions and run UI integration test agains those versions and removed hardcode UI5 versions.


tobiasqueck (approved):
* just tests, no changeset needed
* tests are clear and easy to understand
* I like that we removed the hardcoded tests and test against all maintained version 

Just one question: did the tests work offline before? If yes, then I think you need a fallback in case you cannot connect to ui5.sap.com, if not, then I am ok as it is.

marufrasully:
> offline before?
@tobiasqueck thanks for reviewing. To your question on `offline before`, no tests do not work in offline mode.

zdravko-georgiev (approved):
Code looks good
Local testing ✅

PR #1939, opened 2024-05-17
fix: rta not starting for cpe for ui5 v1.71.x

Jimmy-Joseph19:
Fix for CPE Fails to load rta for ui5 version 1.71.x due to the following import statements
[import workspaceConnector from '../flp/WorkspaceConnector';](https://github.com/SAP/open-ux-tools/blob/544b23d4896ba4751d44b886baebe13d4ef9dda3/packages/preview-middleware-client/src/cpe/init.ts#L18)


packages/preview-middleware-client/src/cpe/connector-service.ts
```diff
@@ -0,0 +1,33 @@
...
+ * A Class of WorkspaceConnectorService
+ */
+export class WorkspaceConnectorService {
+    /**
+     * Initializes connector service.
+     *
+     * @param sendAction action sender function
+     */
+    public async init(sendAction: ActionSenderFunction): Promise<void> {
+        const version = sap.ui.version;
```
voicis, line 14: We should avoid using globals. Please use `sap/ui/VersionInfo` instead here.

voicis (changes requested):

packages/preview-middleware-client/src/cpe/connector-service.ts
tobiasqueck: Could you run `npx @ui5/linter --details` on the code? It should find such issues.

Jimmy-Joseph19: Thanks for this command its very helpful. Probably we need an issue. there are currently 29 issues in the package preview-middleware-client. 
The one introduced with this PR is been resolved.

voicis (dismissed):
- Code changes looks good
- Changeset present
- Good test coverage
- Review comments addressed

.changeset/heavy-windows-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux-private/preview-middleware-client': patch
```
zdravko-georgiev, line 2: when a change is made in `preview-middleware-client`, then also a changeset for the `preview-middleware` needs to be added as well 😉 Please add one

Jimmy-Joseph19: Thanks. Updated :)

voicis (approved):

zdravko-georgiev (approved):
Code is well structuted and easy to understand
Chageset is OK
Excellent test coverage

.changeset/heavy-windows-dance.md
zdravko-georgiev: thx

Issue #1942, opened 2024-05-20
TBI - (fiori-elements-writer): Remove dependency on 'sap.fe.templates' for V4


martinjanus-sap:
### Description (include screenshots)
Feedback from our V4 Fiori elements colleagues about changes to the V4 generated projects (regardless of UI5 version):

- The UI5 dependencies should depend on sap.fe.core and not sap.fe.templates in the manifest.json
- Remove the manifest.json dependency on sap.ushell, it shouldn't be needed.
- The generated flpSandbox.htmlshould reference sap.fe.core and not sap.fe.templates

### Value
Align with best practice recommendations from V4 Fiori elements.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

quiggs1001:
FYI @tobiasqueck.  Might be some middleware updates required as part of this for the `flpSandbox.html` update.

tobiasqueck:
@nlunets could you please check this one. I am aware that you are the source for this but I don't think the requirements above are correct, I think it should be more like
* `sap.fe.templates` should still be used for LROP
* `sap.fe.core` should be used instead of `sap.fe.templates` if a custom page project is generated
* ushell should not be in the manifest (that is correct)
* the test/flp.html should contain templates (and core) to have the best preview expierence - keep in mind the flp.html is only used for local preview, not productively

nlunets:
Yes @tobiasqueck you are correct

cfg74:
@tobiasqueck @nlunets @ianquigley-sap a few additions:

1. propose to keep any changes to the preview html files out-of-scope here as their generation is soon be obsolete as we can create them on the fly in the preview middleware, and changes will be done there instead
2. this BLI has a dependency to an FPM writer enhancement to also maintain the lib section when adding pages, e.g. adding `sap.fe.templates` when creating a floorplan page like the list report, and adding `sap.fe.core` for FPM custom pages.

Would this be something you can take together with the BLI as both have to be delivered in sync or shall I create a separate BLI for it?

PR #1948, opened 2024-05-22
feat(fiori-elements-writer): remove `sap.fe.templates` dependency v4 fpm

martinjanus-sap:
#1942 

- UI5 dependences for FPM v4 removed dependency on `sap.fe.templates`, now depends on `sap.fe.core`
  - updated validate in fpm-writer for `sap.fe.core` 
  - updated to load `sap.fe.templates` in local-ui5.yaml if it's V4 FPM app and doesn't already contain `sap.fe.templates` ,  as its needed to load local preview. 
- Removed dependency on `sap.ushell` for V4 manifest.json.
  - updated to preload `sap.m`, `sap.ui.core` and `sap.ushell` in local-ui5.yaml, needed to load local preview
- flp pre-loads `sap.m`, `sap.ui.core` and `sap.ushell`


packages/fiori-elements-writer/src/data/templateAttributes.ts
```diff
@@ -32,7 +32,7 @@ const commonUi5Libs: FrameworkLibs = {
         'sap.ui.generic.app',
         'sap.suite.ui.generic.template'
     ],
-    [OdataVersion.v4]: ['sap.m', 'sap.ushell', 'sap.fe.templates']
+    [OdataVersion.v4]: ['sap.m', 'sap.fe.templates']
```
devinea, line 35: Removing `sap.ushell` breaks `npm run start-local`
It seems we need to ensure `sap.ushell` is not added to the manifest.json libs. 
However I thought it was needed for cross-app navigation?

But I think we need to still add `sap.ushell` in the ui5-local.yaml and in the `flpSandbox.html` `data-sap-ui-libs` parameter.
@tobiasqueck Any thoughts?

packages/fe-fpm-writer/src/common/validate.ts
```diff
@@ -35,9 +35,10 @@ export function validateBasePath(basePath: string, fs?: Editor): boolean {
         throw new Error(`Invalid project folder. Cannot find required file ${manifestPath}`);
     } else {
         const manifest = fs.readJSON(manifestPath) as any;
-        if ((manifest['sap.ui5']?.dependencies?.libs?.['sap.fe.templates'] !== undefined) === false) {
+        const ui5Dependencies = manifest['sap.ui5']?.dependencies?.libs;
+        if (!(ui5Dependencies && (ui5Dependencies['sap.fe.core'] || ui5Dependencies['sap.fe.templates']))) {
```
tobiasqueck, line 39: `sap.fe.macros` works as well

packages/fiori-elements-writer/src/data/templateAttributes.ts
tobiasqueck: `sap.ushell` should not be needed. If you run your app in FLP (local or productive) then it should be there. If you don't run it in FLP then the shell related functionality is disabled in `sap.fe`.
So, I would also not think it is required in the `flpSandbox.html` but if you have an example project that fails please share it with me. I am curious.

packages/preview-middleware/src/base/flp.ts
```diff
@@ -271,11 +271,14 @@ export class FlpSandbox {
             : '@sap-ux/preview-middleware';
         const config = { ...this.templateConfig };
         /* sap.ui.rta needs to be added to the list of preload libs for variants management and adaptation projects */
-        if (!config.ui5.libs.includes('sap.ui.rta')) {
-            const libs = config.ui5.libs.split(',');
-            libs.push('sap.ui.rta');
-            config.ui5.libs = libs.join(',');
-        }
+        const preloadUI5Lib = ['sap.ui.rta', 'sap.fe.core', 'sap.fe.templates'];
```
tobiasqueck, line 274: we should only do that if the app is a v4 app.

packages/preview-middleware/test/unit/base/__snapshots__/flp.test.ts.snap
```diff
@@ -489,7 +489,7 @@ exports[`FlpSandbox router editor with config 1`] = `
     <!-- Bootstrap the UI5 core library. 'data-sap-ui-frameOptions=\\"allow\\"'' is a NON-SECURE setting for test environments -->
     <script id=\\"sap-ui-bootstrap\\"
         src=\\"../resources/sap-ui-core.js\\"
-        data-sap-ui-libs=\\"sap.m,sap.ui.core,sap.ushell,sap.f,sap.ui.comp,sap.ui.generic.app,sap.suite.ui.generic.template,sap.ui.rta\\"
+        data-sap-ui-libs=\\"sap.m,sap.ui.core,sap.ushell,sap.f,sap.ui.comp,sap.ui.generic.app,sap.suite.ui.generic.template,sap.ui.rta,sap.fe.core,sap.fe.templates\\"
```
tobiasqueck, line 492: oh my ... that list is bad. I think there is was an issue even before your change. The v2 libs should only be there in case of v2 and the v4 libs in case of v4.

packages/fiori-elements-writer/src/data/templateAttributes.ts
devinea: @tobiasqueck  Try 
```
cd packages/fiori-elements-writer/
UX_DEBUG=true UX_DEBUG_FULL=true  npm test
cd test/test-output/lropTemplates/lrop_v4_addtests
npm i && npm run start-local
```
page does not load at http://localhost:8080/test/flpSandbox.html?sap-ui-xx-viewCache=false#lrop_v4_addtests-tile

adding `    - name: sap.ushell` allows the app to load. 
Screenshot 2024-05-22 at 19 54 21

if I recall adding `sap.ushell` to `data-sap-ui-libs` in flpSandbox.html was for performance change rather than waiting for it to be resolved as a dependency later, go ahead a fetch it early as it will be needed.

packages/preview-middleware/src/base/flp.ts
tobiasqueck: I reverted the change. It was the wrong spot to add a lib. Here we are adding additional libs required for the flexibility editor. Getting the preload libs depending on the apps happens in `getUI5Libs` in the same file.

packages/preview-middleware/test/unit/base/__snapshots__/flp.test.ts.snap
tobiasqueck: This looks good again after reverting the change in the `flp.ts`. This test result is correct, however, it might make sense to modify the test to be more realistic.

tobiasqueck: Fixed with 82aa6ccd58d35ede39c5fb2e786f978eee3665a7

packages/fiori-elements-writer/src/data/templateAttributes.ts
tobiasqueck: @devinea I stand corrected! The flp bootstrap does not load the `sap.ushell` lib as I assumed but only creates a global object `sap.ushell`, so, we always need to add `sap.ushell` to the preload libs.
I have corrected it in the `preview-middleware` with 82aa6ccd58d35ede39c5fb2e786f978eee3665a7. Something similar should be done for the `flpSandbox.html` template as well.

packages/fe-fpm-writer/src/common/validate.ts
martinjanus-sap: tests are failing with `sap.fe.macros` for fpm, any other suggestions?

devinea: fpm apps seem to need these added to the ui5-local.yaml. 
``
    - name: sap.fe.macros
    - name: sap.fe.templates
```
test app fails to load from `/packages/fiori-elements-writer/test/test-output/fpmTemplates/with-js`

Comment (martinjanus-sap) [reply packages/fe-fpm-writer/src/common/validate.ts:39]: Suggestions to add them specifically for FPM? Currently preloading `sap.ushell` for ui5-local.yaml, extending this to include `sap.fe.templates` and `sap.fe.macros` could mean unnecessary code for other apps ui5-local.yaml ?

```
    // preload ui5-local.yaml with 'sap.ushell'
    const ui5Libs = ui5App.ui5.ui5Libs as string[];
    if (!ui5Libs.includes('sap.ushell')) {
        ui5Libs.push('sap.ushell');
    }
 ```
 
Looks to me that this is the point where the yaml file is written, and nothing to distinguish apps based on ui5-local.yaml ? Maybe can read manifest to determine template type, if fpm preload `templates` and `macros`?

packages/fiori-elements-writer/src/data/templateAttributes.ts
```diff
@@ -88,7 +88,7 @@ const templateLibs: TemplateLibs = {
         },
         [TemplateType.FlexibleProgrammingModel]: {
             baseComponent: appComponentLibFioriElements,
-            ui5Libs: [...commonUi5Libs[OdataVersion.v4], 'sap.fe.templates']
+            ui5Libs: [...commonUi5Libs[OdataVersion.v4].filter((lib) => lib !== 'sap.fe.templates'), 'sap.fe.core']
```
IainSAP, line 91: A little confusing. Can we instead remove `sap.fe.templates` from `commonUi5Libs`, if it is no longer "common".

devinea (dismissed):
Code changes and snapshot changes look as expected.
changeset ✅ 
I haven't tested latest changes locally.

packages/ui5-application-writer/src/types.ts
```diff
@@ -79,6 +79,11 @@ export interface AppOptions {
      * Excludes the index.html from the template and does not add the `start-noflp` script in package.json
      */
     generateIndex?: boolean;
+
+    /**
+     * Enables specific ui5 libs for FPM apps
+     */
+    loadFpmUi5Libs?: boolean;
```
tobiasqueck, line 86: Sorry, no. You are in a generic UI5 writer and you add a property that is Fiori elements specific. Please change this to e.g. allow defining additional libs but don't put FE/FPM specific logic in the generic ui5 writer

tobiasqueck (changes requested):

packages/ui5-config/src/ui5config.ts
```diff
@@ -168,6 +176,17 @@ export class UI5Config {
         return this;
     }
 
+    public updateUI5Libs(): void {
```
tobiasqueck, line 179: Sorry, but no, now you have FE specific code in an even more generic module.
Please add something like an `addLibs(...libs: string[])` function and then call it in the FE specific code with 'sap.fe.templates` as parameter.

packages/ui5-config/src/ui5config.ts
```diff
@@ -154,6 +154,14 @@ export class UI5Config {
         ui5Libraries: string[],
         ui5Theme = 'sap_fiori_3'
     ): UI5Config {
+        // Libraries to always load
+        const defaultLibraries = ['sap.m', 'sap.ushell', 'sap.ui.core'];
```
tobiasqueck, line 158: Not ideal to have that here. It makes sense for the Fiori tools use cases because we always have an flp preview and need `sap.ushell` but this is a generic module to modify ui5.yaml files and not all consumers might want these libs in there. So, please remove that here, and add the libs to the already existing `ui5Libraries` parameter when you call the function

.changeset/flat-seahorses-collect.md
```diff
@@ -0,0 +1,10 @@
+---
```
tobiasqueck, line 1: I don't think these are all minor:
* freestyle - shouldn't be mentioned at all, it is just patch due to a change in ui5-config that will be automatically generated during the release
* ui5-app - it is just a path, projects that worked before still work, just a little bit better performance
* elements - minor is correct here
* fpm - just a patch in an incorrect validate function
* preview-middleware - just a patch that the consumer won't even notice

tobiasqueck (changes requested):

packages/ui5-config/src/ui5config.ts
tobiasqueck: Thanks for the fix! One suggestion though: consider using `add` instead of `update` otherwise, a consumer could assume a replacement of libs.

tobiasqueck (approved):
Thanks, all my requested changes have been applied.
I have not tested it locally but the snapshots looks as expected.

packages/ui5-application-writer/src/index.ts
```diff
@@ -47,15 +47,19 @@ async function generate(basePath: string, ui5AppConfig: Ui5App, fs?: Editor): Pr
...
 
+    // preload ui5-local.yaml
+    const ui5DefaultLibs = ['sap.m', 'sap.ushell', 'sap.ui.core'];
+    const ui5Libs = ui5App.ui5.ui5Libs as string[];
+    ui5DefaultLibs.forEach((lib) => {
+        if (!ui5Libs.includes(lib)) {
+            ui5Libs.push(lib);
+        }
+    });
+
```
devinea, line 58: `sap.ushell` should added in by the FE and FF writers as they add they `flpSandbox.html` files which needs `sap.ushell`

`sap.m` &&  `sap.ui.core` can be added in ui5-application-writer by default. 
Which seems to be already covered here https://github.com/SAP/open-ux-tools/blob/ebb77d926a837e56741e8ff378a7a8d0466e7f7e/packages/ui5-application-writer/src/data/defaults.ts#L84

packages/fiori-freestyle-writer/src/index.ts
```diff
@@ -92,6 +104,12 @@ async function generate<T>(basePath: string, data: FreestyleApp<T>, fs?: Editor)
         fs.write(ui5LocalConfigPath, ui5LocalConfig.toString());
     }
 
+    // Entend ui5-local.yaml with additional UI5 lib
```
devinea, line 107: typo `Entend`

devinea (commented):
Code changes and snapshot changes as expected. Thanks
Please add a changset and fix small typo.

packages/fiori-freestyle-writer/src/index.ts
mikicvi-SAP: Addressed in [89f5a8a](https://github.com/SAP/open-ux-tools/pull/1948/commits/89f5a8a9cfc05c49dff465ee079a7226c8a0b1a4), thanks!

packages/ui5-application-writer/src/index.ts
mikicvi-SAP: Addressed in [a43bc3c](https://github.com/SAP/open-ux-tools/pull/1948/commits/a43bc3cd3d90fc3f9565fae57ca54c95b007f80b) and [74e857c](https://github.com/SAP/open-ux-tools/pull/1948/commits/74e857ca2c3c9c1be08866ed40ddb12f8b409a73).
Thanks @devinea

devinea (approved):
Thanks @martinjanus-sap @mikicvi-SAP 
Code changes and test are cleanly seperated. 
changeset ✅

tobiasqueck (approved):
No problems with the changes since my last review

Issue #1949, opened 2024-05-22
TBI - To support the exclusion of annotations from manifest json for cap projects


kjose90:
### Description 
 Utilize the application's service type to determine the handling of annotations:

- For EDMX projects, append annotations as a data source to the manifest json and generate an annotation XML file.
- For CDS projects, refrain from adding annotations to the manifest json, instead write annotations into an annotation.CDS file.

Enhanced the `OdataService` to accommodate both `EdmxAnnotationsInfo` and `CdsAnnotationsInfo` types for annotations support.

`EdmxAnnotationsInfo`: Represents information related to EDMX annotations.

- `name`?: string: Optional name for the annotations.
- `technicalName`: string: Technical name of the annotations.
- `xml`: string: XML content of the annotations.

`CdsAnnotationsInfo`: Represents information related to CDS annotations.

- `cdsFileContents`: string: The contents to be written into the annotation cds file.
- `projectPath`: string: The path to the CAP project.
- `appPath`: string: The relative path to the app folder within the CAP project.
- `projectName`: string: The name of the project.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [] `fiori-elements-writer`:
   Update fiori-elements-writer to not append local annotation names to FE service if the service type is CDS.
- [] Move logic to write annotation cds to `odata-service-writer`
   Add logic for writing annotation CDS files to the odata-service-writer module. If the service type is CDS and annotations are provided, write the annotations into the CDS file.
- [] Enhance the `OdataService` annotation feature to accommodate both EDMX and CAP interfaces.

kjose90:
Closing this this issue. Done.

PR #1950, opened 2024-05-22
add support for exclude annotations as part of app gen

kjose90:
#1949

- Added logic in `fiori-elements-writer` to not add local annotations when `excludeAnnotations` is true
- Added option `excludeAnnotations` to `FioriElementsApp` type in `fiori-elements-writer`
- Added a test to `odata-service-writer` which generates project when no service annotations are provided


tobiasqueck:
@kjose90 if understand it correctly then this is for CAP, so I was wondering if it makes more sense to add an option of the annotation format instead of excluding annotations because you would want an `annotation.cds` file in CAP, correct?

kjose90:
Closing this PR as a revised plan is going to be implemented. The decision was made to write `annotation.cds` for CAP applications.

PR #1959, opened 2024-05-23
fix: destination with basic auth issue

### LINKED ISSUE #1951: BUG - preview-middleware fails with 401 in BAS if the destination requires basic authentication
### Related Feature

### Description
The preview middleware is using the `system-access` module to connect to the backend. It works without flaw in VSCode but has an issue in BAS if the destination requires basic authentication. The error is detect (401) but instead of prompting the user like in VSCode it is not handled.

### Steps to Reproduce
1. Generate an adaptation project in BAS
2. run `npm i`
3. run `npm start`

### Expected results
The preview opens in a new tab.

### Actual results
```
Error Message:
Request failed with status code 401
```

### Screenshots

OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

tobiasqueck:
Fixes #1951 by doing a test request during initialization and if it fails with a 401, prompt the user for credentials.

### DIFF
--- a/.changeset/slow-lizards-shake.md
+++ b/.changeset/slow-lizards-shake.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/system-access': patch
+---
+
+FIX: handle 401 in BAS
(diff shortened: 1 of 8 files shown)


korotkovao (approved):
Thank you @tobiasqueck 
Looks good to me 👍

PR #1969, opened 2024-05-27
fix(axios-extension): cleanup deploy logging

longieirl:
The current deployment payload log statment is logging the entire arhive zip in base64 and makes reviewing logs every difficult. For exampe, using BAS terminal, the log statements before this payload debug statement cant be seen as the terminal log is truncated.


zdravko-georgiev (approved):
Simple and easy to understand

cianmSAP (approved):
Code changes make sense based on description
Changeset ✅

PR #1974, opened 2024-05-28
Fix "Uncaught TypeError: Assignment to constant variable." in changes_loader.ts

devinea:
Fix "Uncaught TypeError: Assignment to constant variable." in changes_loader.ts

### DIFF
--- a/.changeset/thick-geese-live.md
+++ b/.changeset/thick-geese-live.md
@@ -1,0 +1,5 @@
+---
+"@sap-ux/fiori-elements-writer": patch
+---
+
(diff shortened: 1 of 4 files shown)


IainSAP (dismissed):
Thx @devinea 

- Minor bug fix, updated as expected

Klaus-Keller (approved):
Thanks @devinea, fix looks good. Did just a visual review but the change is small and the description explains the fix. 

Approved.

IainSAP (approved):
Re-approving

PR #1947, opened 2024-05-22
feature(odata-service-inquirer): Return cds info with cds services

IainSAP:
https://github.com/SAP/open-ux-tools/issues/1934

- Adds and exports type `CdsInfo` to `@sap/project-access`
- Exports type `ServiceInfo` from `@sap/project-access`
- Returns `CdsInfo`, cds version information, from `getCapModelAndServices`
- Returns `CdsInfo` from `@sap-ux/odata-service-inquirer` when a cap service is prompted and selected

PR #1902, opened 2024-05-07
tbi(@sap-ux/cap-config-writer): Add additional cap configuration writing functionalities

kjose90:
Add additional cap configuration writing functionalities 

#1908

PR #1979, opened 2024-05-29
Use cds info from project access to update package json in cap writers 

kjose90:
#1902

This PR consumes #1947

### DIFF
--- a/packages/cap-config-writer/src/cap-writer/package-json.ts
+++ b/packages/cap-config-writer/src/cap-writer/package-json.ts
@@ -1,7 +1,7 @@
 import { type Package, getCapCustomPaths } from '@sap-ux/project-access';
 import type { Editor } from 'mem-fs-editor';
 import path, { join } from 'path';
-import { CdsUi5PluginInfo, CapServiceInfo } from '../cap-config/types'
+import { CdsUi5PluginInfo, CapServiceCdsInfo } from '../cap-config/types'
 import { enableCdsUi5Plugin, checkCdsUi5PluginEnabled, satisfiesMinCdsVersion } from '../cap-config';
 import type { Logger } from '@sap-ux/logger';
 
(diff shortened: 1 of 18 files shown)


packages/cap-config-writer/src/cap-writer/package-json.ts
```diff
@@ -0,0 +1,130 @@
...
+    //log?: Logger
+): Promise<void> {
+    const packageJson = (fs.readJSON(packageJsonPath) ?? {}) as Package;
+    const hasNPMworkspaces = await checkCdsUi5PluginEnabled(packageJsonPath, fs);
+    // Determine whether to add cds watch scripts for the app based on the availability of minimum CDS version information.
+    // If 'cdsUi5PluginInfo' contains version information and it satisfies the minimum required CDS version,
+    // or if 'cdsUi5PluginInfo' is not available and the version specified in 'package.json' satisfies the minimum required version,
+    // then set 'addScripts' to true. Otherwise, set it to false.
+    const addScripts = cdsUi5PluginInfo?.hasMinCdsVersion ? cdsUi5PluginInfo.hasMinCdsVersion : satisfiesMinCdsVersion(packageJson)
+    if(addScripts) {
```
kjose90, line 82: Here we add cds watch scripts only in cdsUi5PluginInfo has minimum cds version available, this is set during the prompt phase of the app generation. Question for head less app generation - according to the logic here I simply read from the package.json, is this consistent ??? Is there a better way to cater for this watch script addition for headless app?

kjose90:
Please refer the updated logic for updating CDS watch scripts in `cap-config-writer/src/cap-config-writer/package-json.ts.`

For the generation of headless applications, where there's no prompt for `cdsUi5Plugin`, what's the optimal approach to retrieve the cds version from package.json?

We have a function `loadModuleFromProject`  from [project-access](https://github.com/SAP/open-ux-tools/blob/53eb4a6b5255b64d740512d917d90889d041af0e/packages/project-access/src/project/cap.ts#L386) which will require pre-installing node modules. - i am not sure if this is the right way, if it could have effects elsewhere, and also considering the overhead associated with exporting this function.

In this PR I am currently re-reading this information from  package json using `satisfiesMinCdsVersion`  [here](https://github.com/SAP/open-ux-tools/blob/53eb4a6b5255b64d740512d917d90889d041af0e/packages/cap-config-writer/src/cap-config/package-json.ts#L73)

PR #1985, opened 2024-05-30
Feat(axios extension): Load service annotations without service catalog request

IainSAP:
https://github.com/SAP/open-ux-tools/issues/1992

- Allows annotation retrieval for a specific service without loading all services from the catalog (which adds an extra network call and cant be used with mock server)
- Adds patching of certs for sap systems


packages/axios-extension/src/abap/catalog/v2-catalog-service.ts
```diff
@@ -138,8 +145,11 @@ export class V2CatalogService extends CatalogService {
      * @param filter.path filter by path
      * @returns service annotations
      */
-    protected async getServiceAnnotations({ id, title, path }: FilterOptions): Promise<ODataServiceV2Info[]> {
-        if (!this.entitySet) {
+    protected async getServiceAnnotations(
+        { id, title, path }: FilterOptions,
+        preloadCatalog = true
```
tobiasqueck, line 150: Getting the correct entity set is not about preloading, it is about fining the correct services depending on the version of the ABAP backend.

tobiasqueck: Looking at it again, I see your point. It is required for fetching the services, however, for fetching the annotation of a specific service you can always use `ServiceCollection`, so there is no need to add this preload parameter but you can remove the  `this.determineEntitySet` call from and always use `/ServiceCollection` as the base path.

IainSAP: Done thanks

tobiasqueck (changes requested):
* changeset missing
* could you please add a comment to the global-root cert file where you got it from (just one a clear reference to the publicly available file so that nobody gets confused and assumes there is something secret

devinea (approved):
Code changes and tests look good.
Changeset ✅

cianmSAP (approved):
Changes are clear 
Coverage above 80
Changeset ✅

PR #1993, opened 2024-05-31
fix: missing RTA toolbar

voicis:
Fix regression after #1945 where RTA toolbar is not visible in variant management mode.


vadson71 (approved):
Changes look reasonable. Didn't test them

zdravko-georgiev (approved):
Change is simple and easy to understand
Did NOT tested manually

PR #1994, opened 2024-05-31
To support the exclusion of annotations from manifest json for cap projects

kjose90:
#1949

- `fiori-elements-writer`:
Update fiori-elements-writer to not append local annotation names to FE service if the service type is CDS.
- Move logic to write annotation cds to `odata-service-writer`
Add logic for writing annotation CDS files to the odata-service-writer module. If the service type is CDS and annotations are provided, write the annotations into the CDS file.- 
- Enhance the `OdataService` annotation feature to accommodate both EDMX and CAP interfaces.


packages/fiori-elements-writer/src/data/defaults.ts
```diff
@@ -92,7 +92,9 @@ export function setAppDefaults<T>(feApp: FioriElementsApp<T>): FioriElementsApp<
         ui5Libs: getUi5Libs(feApp.template.type, feApp.service.version)?.concat(feApp.ui5?.ui5Libs ?? [])
     };
 
-    if (!feApp.service.localAnnotationsName) {
+    // Assign a default annotation name if the service type is EDMX and no local annotation name is provided
+    const serviceType = feApp.service.type ? feApp.service.type : ServiceType.EDMX;
```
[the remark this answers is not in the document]
kjose90: Thanks John

devinea (dismissed):
Code changes appear to coverage the necessary changes to support CAP projects service and annotation writing. 
changeset ✅

packages/odata-service-writer/src/data/annotations.ts
```diff
@@ -13,15 +13,17 @@ import type { NamespaceAlias, OdataService } from '../types';
 export function getAnnotationNamespaces({ metadata, annotations }: Partial<OdataService>): NamespaceAlias[] {
     // Enhance service with annotations namespaces
     const schemaNamespaces = metadata ? getNamespaces(metadata) : [];
-
-    if (annotations?.xml) {
+    const edmxAnnotations = annotations as EdmxAnnotationsInfo;
```
IainSAP, line 16: Should we not test for the property of `EdmxAnnotationsInfo` before casting to that type?

```suggestion
    if (edmxAnnotations?.xml) {
        const edmxAnnotations = annotations as EdmxAnnotationsInfo;
```

devinea (approved):
Latest commits to skip adding proxy middleware backend config is appropriate for CAP.
Re-approving.

longieirl (approved):
Changes look good.
Test coverage is good.
Changeset looks good.

kjose90:
Thanks for reviewing everyone

packages/odata-service-writer/src/data/annotations.ts
IainSAP: Or better still change the parameter type to be `EdmxAnnotationInfo` if that is what should be passed to this function.

PR #1996, opened 2024-05-31
test: update snapshots

voicis:
#1993 did not properly update snapshots after changes and PR build did not catch that.

No changeset needed as this only affects tests and the other PR already has the required changeset.


IainSAP (approved):
Thx @voicis

devinea (approved):

Klaus-Keller (approved):
Snapshot updated look good, thanks @voicis

PR #1998, opened 2024-06-03
WIP: experimenting with Karma and the preview-middleware

tobiasqueck:
@heimwege fyi

### DIFF
--- a/packages/preview-middleware/src/base/flp.ts
+++ b/packages/preview-middleware/src/base/flp.ts
@@ -506,6 +506,7 @@ export class FlpSandbox {
         this.router.get(config.path, (async (_req, res) => {
             this.logger.debug(`Serving test route: ${config.path}`);
             const templateConfig = {
+                basePath: this.templateConfig.basePath,
                 initPath: config.init
             };
             const html = render(testsuite, templateConfig);
--- a/packages/preview-middleware/templates/test/testsuite.qunit.html
+++ b/packages/preview-middleware/templates/test/testsuite.qunit.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
     <title>QUnit TestSuite</title>
-    <script src="/resources/sap/ui/qunit/qunit-redirect.js"></script>
+    <script src="<%- basePath %>/resources/sap/ui/qunit/qunit-redirect.js"></script>
     <script src="<%- locals.initPath %>" data-sap-ui-testsuite></script>
 </head>
 <body>
--- a/packages/ui5-proxy-middleware/src/ui5/middleware.ts
+++ b/packages/ui5-proxy-middleware/src/ui5/middleware.ts
@@ -1,4 +1,5 @@
 import type { RequestHandler, NextFunction, Request, Response } from 'express';
+import { Router as createRouter } from 'express';
 import type { Options } from 'http-proxy-middleware';
 import { ToolsLogger, UI5ToolingTransport } from '@sap-ux/logger';
 import type { ProxyConfig } from '../base';
@@ -38,15 +39,11 @@ function createProxyOptions(logger: ToolsLogger, config: UI5ProxyConfig): Option
  * @returns handler function
  */
 function createRequestHandler(routes: { route: string; handler: RequestHandler }[]): RequestHandler {
-    return (req, res, next): void => {
-        for (const route of routes) {
-            if (req.path.startsWith(route.route)) {
-                route.handler(req, res, next);
-                return;
-            }
-        }
-        next();
-    };
+    const router = createRouter();
+    for (const route of routes) {
+        router.get(`${route.route}*`, route.handler);
+    }
+    return router;
 }
 
 /**


packages/preview-middleware/templates/test/qunit.html
```diff
@@ -12,7 +12,7 @@
 		}'
 		data-sap-ui-animation="false"
 		data-sap-ui-compatVersion="edge"
-		data-sap-ui-async="true"
+		data-sap-ui-async="false"
```
heimwege, line 15: Is this change related to proxying the ui5 sources? Never had issues with that one when using the sources served by tooling via npmjs

tobiasqueck: I cannot explain it to you but I currently assume that the proxy is faster and thats why something is loaded to early. I am not accepting `false` here, I am looking for a better option. Also started to pull in  @petermuessig but it isn't clear yet why things are happening.

tobiasqueck: the initial qunit lib needs to be loaded synchronously, so I moved it into an html script tag. With that I can keep true here.

packages/preview-middleware/templates/test/qunit.js
```diff
@@ -1,8 +1,4 @@
-sap.ui.define([
-    'sap/ui/thirdparty/qunit-2'
-], function(QUnit) {
-        QUnit.config.autostart = false;
-        sap.ui.require(<%- JSON.stringify(tests) %>, function() {
-            QUnit.start();
-    });
-});
\ No newline at end of file
+QUnit.config.autostart = false;
```
heimwege, line 1: [UI5 recommendation](https://github.com/SAP/open-ux-tools/pull/1813#discussion_r1553301524) was not to go for the global QUnit. Is this change really needed?

packages/preview-middleware/templates/test/qunit.html
```diff
@@ -12,10 +12,12 @@
 		}'
 		data-sap-ui-animation="false"
 		data-sap-ui-compatVersion="edge"
-		data-sap-ui-async="false"
-		data-sap-ui-oninit="module:<%- initPath %>">
+		data-sap-ui-async="true">
 	</script>
 
+	<script src="<%- basePath %>/resources/sap/ui/thirdparty/qunit-2.js"></script>
```
heimwege, line 18: I'm afraid this setup results in a race condition between the script initializing global `Qunit` and the `data-sap-ui-oninit` module accessing it (`QUnit.config.autostart = false;`)

packages/preview-middleware/templates/test/qunit.js
tobiasqueck: @petermuessig I need your help here :) I also preferred the other approach but it wasn't working in karma for me.

heimwege: The issue seems to be that qunit starts despite `QUnit.config.autostart = false;` (I don't know why 🤷🏻)
Also this issue seems to only occur when proxying the ui5 resources and not when serving them via tooling from npmjs.

According to [Object preconfig chapter](https://qunitjs.com/api/config/) in the Qunit documentation (no deep link available, sorry)

_You can create a QUnit global variable with a config property, before QUnit itself is loaded. Any properties on this QUnit.config placeholder object will be carried over and applied to the real QUnit.config object once it exists._

So whereas `QUnit.config.autostart = false;` did not work as preconfig in a `script` tag of the html file (again, I don't know why 🤷🏻), the following did actually work when put even before the ui5 bootstrap script tag:
`window.QUnit = { config: {autostart: false}};`

Weirdly also `window.QUnit.config.autostart = false` did not work 🤷🏻

With this setup `npm run test` works in the playground but karma is quitting with an error message

```
Chrome Headless 126.0.6478.115 (Windows 10) ERROR
  Uncaught TypeError: d.done is not a function
  at <path_to_playground>/node_modules/karma-ui5/dist/browser-bundle.js:15:458

  TypeError: d.done is not a function
      at d (<path_to_playground>/node_modules/karma-ui5/dist/browser-bundle.js:2:95253)
      at l.useIframe.g._frame.onload (<path_to_playground>/node_modules/karma-ui5/dist/browser-bundle.js:2:96262)
``` 

Also `QUnit.config.autostart = false;` seems to be a fair approach at least according to the [qunit documentation](https://qunitjs.com/api/config/autostart/)

heimwege: UPDATE: `karma-ui5` mistakes the `window.Qunit` preconfig for the real `QUnit`. That's why karma now quits with an error message. `d.done` actually is `Qunit.done` which is indeed not a function in the precondition object.
So the question to me is, if that's actually a bug in `ui5-karma` 🤔

tobiasqueck: @heimwege please check https://github.com/SAP/open-ux-tools/pull/1998/commits/596ba41e7304637d5f1cd458f87437426f4d6d94 . It tried setting the window object before loading QUnit based on your feedback and that seems to work for me with karma as well as when using the ui5-tooling

packages/preview-middleware/templates/test/qunit.html
tobiasqueck: It's gone again with https://github.com/SAP/open-ux-tools/commit/596ba41e7304637d5f1cd458f87437426f4d6d94

tobiasqueck:
@matz3 & @petermuessig your guidance is required here

packages/preview-middleware/templates/test/qunit.js
heimwege: So the actual issue seems to be caused by the `data-sap-ui-oninit`. Using the required `QUnit` works perfectly fine as long as the `qunit.js` is not loaded via `data-sap-ui-oninit` but using a separate script tag in the `qunit.html`. So the changes in the .js file seem to be not needed.

tobiasqueck: Just to be sure, I get it correct, we can remove
```
window.QUnit ??= {};
QUnit.config ??= {};
QUnit.config.autostart = false;
```
again? 

The key change was that we move the qunit.js file into a script tag instead of an onInit script?

heimwege: At least that's what worked for me 🤷🏻 I would of course still be interested in the opinions of @petermuessig and @matz3.

tobiasqueck: I have updated the `qunit.js` template in https://github.com/SAP/open-ux-tools/pull/1998/commits/1cf3e0605dd65a1d59b71e26deea91ffddac80ba one more time based on the feedback by @petermuessig.
@heimwege could you please do yet another test run in your repo? For me it works now with `karma`, `ui5-test-runner` and `ui5-tooling`.

packages/preview-middleware/templates/test/qunit.js
```diff
@@ -1,8 +1,28 @@
...
+    }
+});
+
+sap.ui.require([
+    "sap/ui/thirdparty/qunit-2",
+    "sap/ui/qunit/qunit-junit",
+    "sap/ui/qunit/qunit-coverage"
+], function (QUnit) {
+    'use strict';
+    QUnit.config.autostart = false;
```
petermuessig, line 24: After understanding the timing issue with the async loader used for testing, this code here will not be safe that QUnit doesn't start automatically. When requiring QUnit with `sap.ui.require` or `sap.ui.define` it will always automatically start unless the `QUnit.config.autostart` is set to false before QUnit is loaded.

Two solution options:
1. Use the script tags again
2. Set autostart to false before loading QUnit

```js
sap.ui.loader.config({
    shim: {
        "sap/ui/qunit/qunit-junit": {
            deps: ["sap/ui/thirdparty/qunit"]
        },
        "sap/ui/qunit/qunit-coverage": {
            deps: ["sap/ui/thirdparty/qunit"]
        },
        "sap/ui/thirdparty/sinon-qunit": {
            deps: ["sap/ui/thirdparty/qunit", "sap/ui/thirdparty/sinon"]
        },
        "sap/ui/qunit/sinon-qunit-bridge": {
            deps: ["sap/ui/thirdparty/qunit", "sap/ui/thirdparty/sinon-4"]
        }        
    }
});

window.QUnit ??= {};
window.QUnit.config ??= {};
window.QUnit.config.autostart ??= false;

sap.ui.define([
    "sap/ui/thirdparty/qunit",
    "sap/ui/qunit/qunit-junit",
    "sap/ui/qunit/qunit-coverage"
], function () {
```

heimwege: Solution 2 (using QUnit preconfig) crashes ui5-test-runner. I'm in contact with @ArnaudBuchholz about that.

tobiasqueck: Thanks!

petermuessig (approved):
Approved with the remark that there could be race conditions and the QUnit tests start before the autostart was set to false. But in case of fast tests, we may ignore that for now and as this is only a temporary solution I'm somehow Ok with it.

PR #2001, opened 2024-06-03
Generic inquirer to prompts adapter 

tobiasqueck:
Prework for #1925


testojs (approved):
* Code looks good
* Test coverage is OK
* Tested with #1925 - works good

IainSAP (changes requested):
Thanks @tobiasqueck 

- Code is clear
- Tests cover new functionality
- Did not manually integrate and test
- `list` validators are not executed but should be covered by inquirer modules covering their execution for CLI usage by alternative prompts.

IainSAP (approved):

PR #2004, opened 2024-06-04
fix: Generated custom action path

broksy:
After some manual testing, I found out that custom actions under custom section are not generated in correct position:
page.settings.body.sections.CustomSection.actions.CustomAction

is not correct path, it should be (similarly like for header/footer actions under content property):
page.settings.**content**.body.sections.CustomSection.actions.CustomAction

### DIFF
--- a/.changeset/red-emus-play.md
+++ b/.changeset/red-emus-play.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fe-fpm-writer': patch
+---
+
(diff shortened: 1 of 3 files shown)


devinea (approved):
I haven't tested locally.
changeset ✅

815are (approved):
checked:
1. code fixes position of custom action in custom section;
2. test snapshot is updated acccording to change
3. changelog persists

PR #2010, opened 2024-06-05
feat: show fe building block controls instead of mdc controls

marufrasully:


packages/control-property-editor-common/src/api.ts
```diff
@@ -106,6 +110,7 @@ export interface OutlineNode {
     children: OutlineNode[];
     icon?: string;
     hasDefaultContent?: boolean;
+    fe?: boolean;
```
voicis, line 113: The UI should be as simple as possible and it shouldn't contain UI5 specific information.

The node filtering should happen in the `preview-middleware-client` when the outline nodes from UI5 are converted.

marufrasully: Thanks. Good point.

PR #2011, opened 2024-06-05
feat: Update of JSON files with previous spacing

broksy:
Adds `updatePackageJSON` and `updateManifestJSON` functions to `project-access` that updates JSON files while keeping the previous indentation/spacing.


packages/project-access/src/file/file-access.ts
```diff
@@ -1,5 +1,7 @@
 import { promises as fs } from 'fs';
 import type { Editor } from 'mem-fs-editor';
+import { Manifest, Package } from '../types';
```
Klaus-Keller, line 3:
```suggestion
import type { Manifest, Package } from '../types';
```

packages/project-access/src/file/file-access.ts
```diff
@@ -66,3 +68,49 @@ export async function fileExists(path: string, memFs?: Editor): Promise<boolean>
...
     }
 }
+
+/**
+ * Updates package.json file asynchronously by keeping the previous indentation.
+ *
+ * @param path - path to file
+ * @param packageJson - updated package.json file content
+ */
+export async function updatePackageJSON(path:string, packageJson: Package): Promise<void> {
```
Klaus-Keller, line 78:
```suggestion
export async function updatePackageJSON(path: string, packageJson: Package): Promise<void> {
```

packages/project-access/src/file/file-access.ts
```diff
@@ -66,3 +68,49 @@ export async function fileExists(path: string, memFs?: Editor): Promise<boolean>
...
+ */
+export async function updateManifestJSON(path: string, manifest: Manifest): Promise<void> {
+    await updateJSON(path, manifest);
+}
+
+/**
+ * Updates JSON file asynchronously by keeping the indentation from previous content with new content for given path.
+ *
+ * @param path - path to file
+ * @param manifest - updated JSON file content
```
Klaus-Keller, line 96:
```suggestion
 * @param content - updated JSON file content
```

packages/project-access/src/file/file-info.ts
```diff
@@ -0,0 +1,79 @@
+import { JSONFileInfo } from "../types";
```
Klaus-Keller, line 1:
```suggestion
import type { JSONFileInfo } from '../types';
```

packages/project-access/src/types/info/index.ts
```diff
@@ -91,3 +91,12 @@ export type AppType =
     | 'SAPUI5 Extension'
     | 'Fiori Reuse'
     | 'Fiori Adaptation';
+
+export interface JSONFileInfo {
+    // Spaces size
+    size?: number;
+    // Use '\t' symbol instead of space
+    useTabSymbol?: boolean;
+    // End of file
```
Klaus-Keller, line 100: Could we simplify this to one property `space: number | string` which is the same as described in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#space?

```suggestion
    // Space as defined in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#space
    sapce: number | string;
```

packages/project-access/src/file/file-info.ts
```diff
@@ -0,0 +1,79 @@
...
+    // 'tabInfo' exists - it was passed as custom configuration or calculated from target file
+    if (tabInfo.useTabSymbol) {
+        // Tab symbol should be used as tab
+        space = CHAR_TAB.repeat(tabInfo.size ?? 1);
+    } else {
+        // Spaces should be used as tab
+        space = tabInfo.size;
+    }
+    return space;
+}
```
Klaus-Keller, line 24: If we simplify the structure to have `space` property, this function would be obsolete.

packages/project-access/test/file/file-access.test.ts
```diff
@@ -90,4 +90,30 @@ describe('fileAccess', () => {
             expect(exists).toBe(true);
         });
     });
+
+    describe('updatePackageJSON', () => {
+        test('Should update package.json using previous indentation', async () => {
```
Klaus-Keller, line 95: Can we please add some tests for files with different indenting and different characters like spaces or `\t`?

Klaus-Keller (changes requested):
Hi @broksy, 
Please see my comments. There also some minor linting issues.

packages/project-access/src/file/file-access.ts
```diff
@@ -66,3 +68,39 @@ export async function fileExists(path: string, memFs?: Editor): Promise<boolean>
...
+
+/**
+ * Updates JSON file asynchronously by keeping the indentation from previous content with new content for given path.
+ *
+ * @param path - path to file
+ * @param content - updated JSON file content
+ */
+async function updateJSON(path: string, content: object): Promise<void> {
+    // read old contents and indentation of the JSON file
+    const oldContentText = await readFile(path);
```
Klaus-Keller, line 100: If we add optional parameter `memFs?: Editor`, this line would then become 
```suggestion
    const oldContentText = await readFile(path, memFs);
```

packages/project-access/src/file/file-access.ts
```diff
@@ -66,3 +68,39 @@ export async function fileExists(path: string, memFs?: Editor): Promise<boolean>
...
+    const oldContentJson = parseJsonError(oldContentText);
+    const indent = Symbol.for('indent');
+    // prepare new JSON file content with previous indentation
+    const result = JSON.stringify(content, null, oldContentJson[indent]) + '\n';
+    await writeFile(path, result);
```
Klaus-Keller, line 105: If we add optional parameter `memFs?: Editor`, this line would then become 
```suggestion
    await writeFile(path, result, memFs);
```

packages/project-access/package.json
```diff
@@ -46,6 +46,7 @@
         "@types/mem-fs-editor": "7.0.1",
         "@types/semver": "7.5.2",
         "@ui5/manifest": "1.61.0",
+        "json-parse-even-better-errors": "3.0.2",
```
Klaus-Keller, line 49: Shouldn't this be declared in `dependencies`?
Additionally, `@types/json-parse-better-errors` can be added in `devDependencies`

Klaus-Keller (changes requested):
@broksy, thanks a lot! See some minor comments below.

packages/project-access/src/project/access.ts
```diff
@@ -153,6 +156,26 @@ class ApplicationAccessImp implements ApplicationAccess {
...
     }
 
+    /**
+     * Updates package.json file asynchronously by keeping the previous indentation.
+     *
+     * @param packageJson - updated package.json file content
+     * @param memFs - optional mem-fs-editor instance
+     */
+    async updatePackageJSON(packageJson: Package, memFs?: Editor): Promise<void> {
+        await updatePackageJSON(join(this.app.appRoot, 'package.json'), packageJson, memFs);
```
Klaus-Keller, line 166: Could we add a contant to `packages/project-access/src/constants.ts`

packages/project-access/package.json
```diff
@@ -36,11 +36,13 @@
         "@sap-ux/ui5-config": "workspace:*",
         "fast-xml-parser": "4.2.7",
         "findit2": "2.2.3",
+        "json-parse-even-better-errors": "3.0.2",
         "mem-fs": "2.1.0",
         "mem-fs-editor": "9.4.0",
         "semver": "7.5.4"
     },
     "devDependencies": {
+        "@types/json-parse-better-errors": "1.0.3",
```
Klaus-Keller, line 45: Sorry @broksy, seems this does not contain correct types, I've misread the package name, we are using json-parse-**even**-better-errors, so we can remove this and keep your declaration file.
```suggestion
```

Klaus-Keller (changes requested):
Hi @broksy, two comments, rest looks good.

Klaus-Keller (approved):
Thanks @broksy!
- changest exists
- code coverage is great

Approved from my side.

packages/project-access/src/file/file-access.ts
```diff
@@ -66,3 +68,42 @@ export async function fileExists(path: string, memFs?: Editor): Promise<boolean>
...
+ * @param content - updated JSON file content
+ * @param memFs - optional mem-fs-editor instance
+ */
+async function updateJSON(path: string, content: object, memFs?: Editor): Promise<void> {
+    // read old contents and indentation of the JSON file
+    const oldContentText = await readFile(path, memFs);
+    const oldContentJson = parseJsonError(oldContentText);
+    const indent = Symbol.for('indent');
+    // prepare new JSON file content with previous indentation
+    const result = JSON.stringify(content, null, oldContentJson[indent]) + '\n';
```
815are, line 107: @Klaus-Keller 
should we consider `\r\n` line endings?
I think example in https://github.com/npm/json-parse-even-better-errors contains it

815are (commented):
Checked:
1. code changes simple to read and understand;
2. tests are added
3. changelog presents

Did not test manually, but noticed small question about line endings

815are (approved):

Issue #2007, opened 2024-06-04
FEATURE - Add deployment writer module to support adding Cloud Foundry deployment configuration


longieirl:
As a developer using SAP Fiori tools, I want to add Cloud Foundry (CF) deployment configurations to my app, so that I am able to deploy my app to an CF target system.

### Technical Design
New module : 
- @sap-ux/deployment-writer

### Acceptance Criteria

**Given** I have a Fiori project with no CF deployment configuration
**when** I invoke this writer with the correct data
**then** the CF deployment configuration will be added to my Fiori project

**Given** I have a Fiori project with an existing mta configuration i.e. standalone or managed
**when** I invoke this writer with the correct data
**then** the CF deployment configuration will be added to my Fiori project

### Tasks
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #2013, opened 2024-06-06
fix(deploy-config-writer): add new deploy-config-writer

longieirl:

- Introduce a new deploy-config-writer, will be used for both ABAP and CF deployment flows.
- The initial change is to export the `MtaConfig` class, this will provide a wrapper for the `@sap/mta-lib` to edit `mta.yaml` files used in the CF generation flow.
- Followup PR's will export the ABAP and CF `generate` functionality to be consumed by the respective generators


IainSAP:
@longieirl I looked but cannot find the `generate` entry point and associated param types which are the minimum requirement for a writer module.

longieirl:
Closing for now and will re-align based on feedback.

Issue #2014, opened 2024-06-06
BUG - Add new model writer has misspelled property that breaks the validation of the change


nikmace:
### Related Feature
Feature request: _#1643_

### Description
In the content of the addNewModel.change the property under settings should be odataVersion instead of version.

### Steps to Reproduce
1. Use the New Model writer and generate a change
2. See incorrect change content version property

### Expected results
Content has odataVersion under settings

### Actual results
Content has version

### Screenshots
If applicable, add screenshots to help explain the problem.

<img width="522" alt="Screenshot 2024-06-06 at 15 42 34" src="https://github.com/SAP/open-ux-tools/assets/39880364/707dc67a-d09b-4f4f-8341-376c35946d37">

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2016, opened 2024-06-06
fix: Add new model writer has misspelled property that breaks the validation of the change

nikmace:
Fix for #2014.

- Changes property name to odataVersion.


.changeset/mean-beds-notice.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ux/adp-tooling': patch
+'@sap-ux/preview-middleware': patch
+---
```
GDamyanov, line 4: Why we need patch update on '@sap-ux/preview-middleware' ?

GDamyanov (changes requested):

.changeset/mean-beds-notice.md
nikmace: Because @sap-ux/preview-middleware is consumes @sap-ux/adp-tooling, otherwise the version in @sap-ux/preview-middleware will not be updated.

GDamyanov (approved):
Small and simple change.
Code looks good.
Good test coverage.
Did not tested manually.

voicis (approved):
- Changeset present
- Simple code change
- Good test coverage
- Did not test manually

PR #2023, opened 2024-06-07
Chore - update tsconfig target ES version to ES2022 aligning with NodeJS 18+

devinea:
Update ES target to ES2022.
- output Javascript is smaller and easier to debug and read. 
- No longer need to support older runtime envs.


IainSAP (approved):
Great!

Klaus-Keller (approved):
Thanks @devinea, switch to `ES2022` looks good, approved from my side.

packages/axios-extension/test/abap/abap-service-provider.test.ts
```diff
@@ -144,7 +144,7 @@ describe('AbapServiceProvider', () => {
             const catalog = provider.catalog(ODataVersion.v2);
             expect(catalog).toBeDefined();
             expect(catalog.defaults.baseURL).toBe(`${server}${V2CatalogService.PATH}`);
-            expect(provider.catalog(ODataVersion.v2)).toEqual(catalog);
+            expect(provider.catalog(ODataVersion.v2).services).toEqual(catalog.services);
```
tobiasqueck, line 147: curious: why

devinea: With the original code we get a failure as below. 
I assume we are really only interested in the services returned from the catalog call.
Screenshot 2024-06-24 at 07 05 05

IainSAP (approved):
Thx @devinea 

- es version update only, changes as expected

packages/cds-odata-annotation-converter/tsconfig.json
```diff
@@ -1,33 +1,34 @@
...
-    "rootDir": "src",
-    "lib": ["ES2021"]
-  },
+    "extends": "../../tsconfig.json",
+    "compilerOptions": {
+        "incremental": true,
+        "outDir": "dist",
+        "rootDir": "src",
+        "lib": ["ES2021"],
+        "target": "ES2021"
```
voicis, line 8: This one still has target set to `ES2021`. Was it intentional?

packages/cds-annotation-parser/tsconfig.json
```diff
@@ -4,6 +4,7 @@
         "outDir": "dist",
         "rootDir": "src",
         "lib": ["ES2021"],
+        "target": "ES2021",
```
voicis, line 7: This one still has target set to `ES2021`. Was it intentional?

packages/cds-odata-annotation-converter/tsconfig.json
devinea: Screenshot 2024-07-05 at 15 40 31
Tests fail when updated in annotation packages. 
All seem to show this line as https://github.com/SAP/open-ux-tools/blob/main/packages/cds-annotation-parser/src/parser/parser.ts#L345

voicis: ok, it looks like there is some issue with chevrotain with that target. We can look at it separately.

voicis (approved):
Control Property Editor changes looks good and tested it locally with node 18.

Issue #2031, opened 2024-06-14
TBI - Exclude writing of specific files for CAP applications


kjose90:
### Description (include screenshots)
When the application is a CAP, ensure the following files are not added:
- `.gitignore`
- `ui5-local.yaml`
- `./webapp/index.html`

### Tasks
- [ ] Add a flag in `ui5-application-writer` to check if the application is CAP. If true, do not copy `.gitignore` and `ui5-local.yaml `from templates. Also, exclude the `start-local `script in the package.json.
- [ ] Do not write `ui5-local.yaml `if the service is of CDS type in `odata-service-writer`.
- [ ] Set a flag in `fiori-elements-writer` (to be passed to sub modules) to indicate that the application is CAP. If true, exclude the `start-local `script in the package.json tasks.
- [ ] Set a flag in `fiori-freestyle-writer` to indicate that the application is CAP. If true, exclude the` start-local `script in the package.json tasks.
- [ ] Update relevant snapshots in all modules to reflect changes for CAP applications.

kjose90:
Closing this this issue. Done.

PR #2032, opened 2024-06-14
Exclude writing of specific files for CAP applications 

kjose90:
#2031


kjose90:
This PR is an initial rough trial to identify all the files that need modification to prevent writing `.gitignore`, `ui5-local.yaml`, and `webapp/index.html `when the application is a CAP application. The flag name used in this case can be changed to something generic. 

Also not sure how to deal with not writing `ui5-local.yaml` Flexible Programming Model [here](https://github.com/SAP/open-ux-tools/blob/073996cbfdf955b4747f77728146e2f7ed2f46c2/packages/fiori-elements-writer/src/index.ts#L149)

kjose90:
Introduced `isCapProject`  to `AppOptions` interface under `ui5-application-writer` to provide a unified and consistent way to handle CAP project-specific configurations. By using this flag:
- we can exclude `ui5-local.yaml` and `.gitignore` files from CAP projects.
- we can conditionally exclude scripts `start-local` in package defaults based on whether the project is a CAP project or not.

Alternative suggestion would  be to introduce a flag like `writeUi5Local` specifically for controlling the writing of the `ui5-local.yaml` file. However, this approach would necessitate a separate flag for handling the `.gitignore` file exclusion or maybe we can leave `.gitignore` file for cap projects ( after all its a git ignore file)

below is a description of files which will need to be changed where ui5-local yaml files are being written (the code for now just uses a flag `isCapProject`)

`packages/ui5-application-writer/src/types.ts`
- Added `isCapProject`a boolean type to `AppOptions` interface to exclude files `ui5-local.yaml` and `.gitignore` from the template for CAP projects if set to true.

`packages/ui5-application-writer/src/data/defaults.ts`
- If `isCapProject` is true, do not add `start-local` scripts to package defaults.

`packages/fiori-elements-writer/src/index.ts`
- Use the same `isCapProject` from `AppOptions` interface
    - to determine whether to write the `ui5-local.yaml` if cap project.
    - Flexible Programming Model: Use `isCapProject` to decide on the inclusion of the `ui5-local.yaml` file.
    - Pass `isCapProject` to the package json config to add `start-local` script to package JSON scripts only for non-CAP applications only.
    packages/fiori-freestyle-writer/src/index.ts

`packages/fiori-freestyle-writer/src/index.ts`
- Maintain the behavior of using the existing `CapType` to not write to `ui5-local.yaml `files for CAP projects.

kjose90:
closing this PR as this is outdated

PR #2033, opened 2024-06-14
Reuse `system-access` in `backend-proxy-middleware`

tobiasqueck:
- [ ] enhance `system-access` with refreshToken handling as it is in the `backend-proxy-middleware`
- [ ] replace `backend-proxy-middleware` code that accesses a system with `system-access`
- [ ] enhance and refactor the unit tests

### DIFF
--- a/packages/backend-proxy-middleware/package.json
+++ b/packages/backend-proxy-middleware/package.json
@@ -30,7 +30,7 @@
         "!dist/**/*.map"
     ],
     "dependencies": {
-        "@sap-ux/axios-extension": "workspace:*",
+        "@sap-ux/system-access": "workspace:*",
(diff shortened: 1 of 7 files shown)


heimwege:
@tobiasqueck do you want to follow-up on this one? Still seems like a reasonable adjustment to me.

tobiasqueck:
> @tobiasqueck do you want to follow-up on this one? Still seems like a reasonable adjustment to me.

I would be happy if someone takes it from here, I won't have any time over the next weeks.

IainSAP:
@heimwege I dont think this is needed anymore. There will be no more support for UAA authentication see:  https://github.com/SAP/open-ux-tools/pull/3693

PR #2038, opened 2024-06-17
Moving xml js from dev dependency to real dependency in package json

kjose90:
#2041

Moving xml js from dev dependency to real dependency in package json for cap config writer


Klaus-Keller (approved):
Fix for dependency to `xml-js` looks good.

cianmSAP (approved):
Update to package.json and lock file looks good

PR #2044, opened 2024-06-18
Fix/1611/wrong config

tobiasqueck:
Issue: the ADP template created a wrong configuration of the `fiori-tools-proxy`, this PR is fixing it as well as a few linting issues.

### DIFF
--- a/packages/adp-tooling/src/writer/options.ts
+++ b/packages/adp-tooling/src/writer/options.ts
@@ -1,4 +1,10 @@
-import type { CustomMiddleware, UI5Config, CustomTask, AbapTarget } from '@sap-ux/ui5-config';
+import type {
+    CustomMiddleware,
+    UI5Config,
+    CustomTask,
+    AbapTarget,
+    FioriToolsProxyConfigBackend
+} from '@sap-ux/ui5-config';
 import type {
     CustomConfig,
     AdpWriterConfig,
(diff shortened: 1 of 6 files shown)


packages/adp-tooling/test/unit/writer/__snapshots__/index.test.ts.snap
```diff
@@ -80,18 +80,18 @@ server:
             url: http://sap.example
           ignoreCertErrors: false
     - name: fiori-tools-proxy
-      afterMiddleware: fiori-tools-preview
+      afterMiddleware: compression
       configuration:
-        ignoreCertErrors: false
+        ignoreCertError: false # If set to true, certificate errors will be ignored. E.g. self-signed certificates will be accepted
```
nikmace, line 85: When did this comment get added: `# If set to true, certificate errors will be ignored. E.g. self-signed certificates will be accepted`? I think there were no comments in the ui5.yaml previously

nikmace (changes requested):

packages/adp-tooling/test/unit/writer/__snapshots__/index.test.ts.snap
tobiasqueck: That comment was always generated for a Fiori app project but not for adp. By reusing the existing app code, it is now added here as well.

packages/adp-tooling/src/writer/options.ts
```diff
@@ -98,36 +103,27 @@ function getFioriToolsMiddlwares(config: AdpWriterConfig): CustomMiddleware<unkn
...
+    ]);
+    ui5Config.addFioriToolsProxydMiddleware({
+        ui5: {
+            url: config?.ui5?.frameworkUrl,
+            version: config?.ui5?.minVersion ?? '' //default to latest if version is not set
+        },
+        backend: [backendConfig as FioriToolsProxyConfigBackend]
+    });
+
+    return;
```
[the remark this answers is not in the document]
tobiasqueck: That was a copy and paste issue. Fixed with 4824b76a523d81c51b798748155212b9e6a624d1

nikmace (approved):
Code looks good
Coverage is good
Did not test manually

Klaus-Keller (approved):
- code fix looks good
- changeset exists
- test coverage is good
- did not run a local test, just visual review

Approved from my side

Issue #2049, opened 2024-06-18
TBI - Convert ux-store `AuthenticationType` to string literal type as ux-store is importing keytar


IainSAP:
### Description (include screenshots)
Imports only requiring the `AuthenticationType` from `@sap-ux/store` are also importing its dependencies (e.g keytar) since this is an enum. We should replace with a real type  (string literal) and move to dev dependencies (@sap-ux/ui5-config)

https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-config/src/types/middlewares.ts#L1

### Value

Prevents issues with bundling when keytar is included.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

IainSAP:
Done

PR #2052, opened 2024-06-19
Replace `@sap-ux/store` enum `AuthenticationType` with type

IainSAP:
https://github.com/SAP/open-ux-tools/issues/2049

- Replace enum with const def and string literal type to prevent adding deps to consume the enum (as this leads to pulling in keytar also)


cianmSAP (approved):
Thanks @IainSAP 
Updates look good

Changeset ✅

devinea (approved):
Makes sense to change this. 
Minor patch should cover the change in 0 major version.

Issue #2053, opened 2024-06-19
TBI - Update cds-plugin-ui5 version


korotkovao:
### Description (include screenshots)
Update cds-plugin-ui5 version

### Value
There is a fix introduced in the latest cds-plugin-ui5: "^0.9" for the CAP Kyma build compatibility. 

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

PR #2054, opened 2024-06-19
feat(cap-config-writer): update cds plugin ui5

korotkovao:
TBI - Update cds-plugin-ui5 version #2053

There is a fix introduced in the latest cds-plugin-ui5: "^0.9" for the CAP Kyma build compatibility.


packages/cap-config-writer/test/fixture/cap-valid-cds-plugin-ui/package.json
```diff
@@ -1,9 +1,9 @@
 {
     "dependencies": {
-        "@sap/cds": "^6.8.2"
+        "@sap/cds": "^7.9.2"
```
longieirl, line 3: This is a major version bump in CDS, should the cap-config-writer also be a major bump?

devinea (commented):
Doesn't this need to be updated?
https://github.com/SAP/open-ux-tools/blob/e39dc02d9561f974d11ad7fcb424e708e7cc0d7f/packages/cap-config-writer/src/cap-config/package-json.ts#L6

packages/cap-config-writer/test/fixture/cap-valid-cds-plugin-ui/package.json
korotkovao: You are right, i will stick with ^6.8.2
cds-plugin-ui5: "0.9.3" is what i really need

korotkovao:
> Doesn't this need to be updated?
> 
> https://github.com/SAP/open-ux-tools/blob/e39dc02d9561f974d11ad7fcb424e708e7cc0d7f/packages/cap-config-writer/src/cap-config/package-json.ts#L6

Yes, thank you!

packages/cap-config-writer/test/unit/cap-config/index.test.ts
```diff
@@ -17,7 +17,7 @@ describe('Test enableCdsUi5Plugin()', () => {
             },
             'workspaces': ['app/*'],
             'devDependencies': {
-                'cds-plugin-ui5': '^0.6.13'
+                'cds-plugin-ui5': '^0.9.3'
```
devinea, line 20:
```suggestion
                'cds-plugin-ui5': '^0.1.0'
```
Don't this should be updated. The point of this test was that it would be updated by the writer.

devinea (changes requested):

devinea (approved):
👍

longieirl (approved):
Changes look good.
Thanks for making changes.

korotkovao:
Thank you All

PR #2058, opened 2024-06-20
TBI support CAP JAVA loading CDS model from maven jar

devinea:
Support CAP JAVA loading CDS model from maven jar.
More details:
https://cap.cloud.sap/docs/java/building-plugins#reference-the-new-cds-model-in-an-existing-cap-java-project


devinea:
May not be needed.

Issue #2060, opened 2024-06-21
BUG - `odata-service-inquirer` is returning the supported odata service url datasource as unimplemented


IainSAP:
### Description

Using `odata-service-inquirer` to prompt, when selecting `Connect to an OData Service Url` the prompt shows an unimplemented message, however it is implemented.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

IainSAP:
Fixed

PR #2061, opened 2024-06-21
Remove unimplemented message for implemented datasource

IainSAP:
https://github.com/SAP/open-ux-tools/issues/2060

- Remove not-yet-implemented message for now inplemented `odata service url` datasource
- Update test to cover

### DIFF
--- a/packages/odata-service-inquirer/src/prompts/prompts.ts
+++ b/packages/odata-service-inquirer/src/prompts/prompts.ts
@@ -53,7 +53,6 @@ function getDatasourceTypeQuestion(options?: DatasourceTypePromptOptions): YUIQu
                 [
                     DatasourceType.businessHub,
                     DatasourceType.none,
-                    DatasourceType.odataServiceUrl,
                     DatasourceType.projectSpecificDestination,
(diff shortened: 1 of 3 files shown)


packages/odata-service-inquirer/src/prompts/prompts.ts
```diff
@@ -53,7 +53,6 @@ function getDatasourceTypeQuestion(options?: DatasourceTypePromptOptions): YUIQu
                 [
                     DatasourceType.businessHub,
```
cianmSAP, line 54: Doesn't have to be in this PR but `businessHub` can eventually go

cianmSAP (approved):
Changes are clear
Tests to cover
Changeset ✅

longieirl (approved):
Changes look good.
Tests to cover change.

packages/odata-service-inquirer/src/prompts/prompts.ts
IainSAP: Thx @cianmSAP , yes, I was waiting for confirmation before I remove it completely.

PR #2065, opened 2024-06-24
fix: correct target for inline composition

Jimmy-Joseph19:
PR contains fix for Compositon targets. 
Existing Result: The code was not correctly using the alias of annotation file.
Actual Result: Fully qualifed name is converted to aliasQuliafied name when writing to file.


packages/fiori-annotation-api/src/cds/adapter.ts
```diff
@@ -428,8 +428,24 @@ export class CDSAnnotationServiceAdapter implements AnnotationServiceAdapter, Ch
         const parsedName = parseIdentifier(pathBase);
         const fullyQualifiedPath =
             toFullyQualifiedName(aliasInfo.aliasMap, aliasInfo.currentFileNamespace, parsedName) ?? '';
+        const fQParsedName = parseIdentifier(fullyQualifiedPath);
```
voicis, line 431: It seems that we shouldn't use the parsing logic used for OData names and paths here, because it doesn't have the same semantics and we can end up in having to add workarounds for case that does not fit OData logic.

We can use `aliasMap` here and simply the logic a little bit. We do not need the fully qualified name of the path base, we just need to look for its namespace in the alias map and if we find it just replace it with an alias (assuming the aliased value is passed already in the change as it normally should be).

```suggestion
        const metadataElement = this.metadataService.getMetadataElement(fullyQualifiedPath);
        let originalPathBase = metadataElement?.originalName ?? pathBase;
        if (parsedName.namespaceOrAlias !== undefined) {
            const namespace = aliasInfo.aliasMap[parsedName.namespaceOrAlias];
            if (namespace) {
                originalPathBase = originalPathBase.replace(namespace, parsedName.namespaceOrAlias);
            }
        }
        change.target.name = [originalPathBase, ...pathSegments].join('/');
```

Jimmy-Joseph19: Thanks adapted.

voicis (approved):
Thanks @Jimmy-Joseph19!
- Review comments addressed
- Simple code change
- Good test coverage

voicis (approved):

nahmed22 (approved):

PR #2066, opened 2024-06-24
feat: add local module cache to project-access

Klaus-Keller:
Add feature to locally cache node modules. First consumer of this is specification to cache local versions of `@sap/ux-specification`.

The cache for node modules is located at `~/.fioritools/module-cache`. A folder structure including scope, name, and version will be created for each cached module. For instance, module `@scope/module@1.2.3` will installed in `~/.fioritools/module-cache/@scope/module/1.2.3` which will result in following folder structure:

```
~
└── .fioritools
    └── module-cache
        └── @scope
            └── module
                └── 1.2.3
                    ├── node_modules
                    │   └── @scope
                    │       └── module
                    │            ...
                    ├── package-lock.json
                    └── package.json
```

New public exported functions:

```typescript
/**
 * Loads and return specification from project or cache.
 * 1. if package.json contains devDependency to specification, attempts to load from project.
 * 2. if not in package.json of project, attempts to load from cache.
 *
 * @param root - root path of the project/app
 * @param [options] - optional options
 * @param [options.logger] - logger instance
 * @returns - specification instance
 */
export async function getSpecification<T>(root: string, options?: { logger?: Logger }): Promise<T> 
```
 and 

```typescript
/**
 * Refreshes the specification dist-tags cache. Also cleans specification modules in cache that are not required anymore.
 *
 * @param [options] - optional options, like logger
 * @param [options.logger] - logger instance
 */
export async function refreshSpecificationDistTags(options?: { logger?: Logger }): Promise<void>
```


packages/project-access/src/project/specification.ts
```diff
@@ -0,0 +1,127 @@
...
+    const logger = options?.logger;
+    try {
+        const distTagsString = await execNpmCommand(['view', '@sap/ux-specification', 'dist-tags', '--json'], {
+            logger
+        });
+        const distTags = JSON.parse(distTagsString);
+        await writeFile(specificationDistTagPath, JSON.stringify(distTags, null, 4));
+        const uniqueVersions = new Set(Object.values(distTags));
+
+        // Check if we have cached versions that are not required anymore
```
devinea, line 77: Nice feature t o cache the dist tags too.

devinea (dismissed):
Nice feature. Clearly implemented and tested. 
Maybe it could be a separate module by itself, but fine for now.
changeset ✅

packages/project-access/src/project/specification.ts
Klaus-Keller: Currently specific for `dist-tags` in specification as we use it there intensively, but can also be generalized if required.

packages/project-access/test/command/npm-command.test.ts
```diff
@@ -0,0 +1,182 @@
...
+    afterAll(() => {
+        Object.defineProperty(process, 'platform', {
+            value: originalPlatform
+        });
+    });
+
+    test('Install node module with logger', async () => {
+        // Mock setup
+        Object.defineProperty(process, 'platform', {
+            value: 'darwin'
```
johannes-kolbe, line 106: Shouldn't this be `win` for windows tests, or be removed as defined in `beforeAll`?
```suggestion
            value: 'win'
```

packages/project-access/test/command/npm-command.test.ts
```diff
@@ -0,0 +1,182 @@
...
+        jest.resetAllMocks();
+    });
+
+    afterAll(() => {
+        Object.defineProperty(process, 'platform', {
+            value: originalPlatform
+        });
+    });
+
+    test('Install node module with logger', async () => {
```
johannes-kolbe, line 103: Seems like with no options

packages/project-access/test/command/npm-command.test.ts
Klaus-Keller: good catch, I've switched later to a complete block with mocking `process.platform`, was probably a leftover. Fixed with https://github.com/SAP/open-ux-tools/pull/2066/commits/a8122363522a7e855668cc42c5e4cb3ccedda2c0

packages/project-access/test/command/npm-command.test.ts
Klaus-Keller: good catch, I've switched later to a complete block with mocking `process.platform`, was probably a leftover. Fixed with https://github.com/SAP/open-ux-tools/pull/2066/commits/a8122363522a7e855668cc42c5e4cb3ccedda2c0

packages/project-access/src/project/specification.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Refreshes the specification dist-tags cache. Also cleans specification modules in cache that are not required anymore.
+ *
+ * @param [options] - optional options, like logger
+ * @param [options.logger] - logger instance
+ */
+export async function refreshSpecificationDistTags(options?: { logger?: Logger }): Promise<void> {
+    const logger = options?.logger;
+    try {
+        const distTagsString = await execNpmCommand(['view', '@sap/ux-specification', 'dist-tags', '--json'], {
```
johannes-kolbe, line 70: @Klaus-Keller This seems not to work for me manually, if I delete the `dist-tags.json` I get an error on the next load: 
```
Error executing npm command 'npm view @sap/ux-specification dist-tags --json': Error: spawn npm ENOENT
Error refreshing specification dist-tags: Error: spawn npm ENOENT
Failed to load specification: Error: ENOENT: no such file or directory, open '.../.fioritools/specification-dist-tags.json'
```

johannes-kolbe: Retested with bundled `.vsix` - problems do not occur anymore and behaviour works as expected.

johannes-kolbe (approved):
LGTM
- code is easy to understand and structured
- tests and great coverage provided
- Manual test ok (consumption via tools-suite with VSIX)
  - check, change, reload of modules

PR #2069, opened 2024-06-24
Fiori AI: Axios Extensions for RAP Service Generation

UweSchwerk:
Add RAP service generation
- validate request (executed to validate user input to ensure consistent service generation)
- preview request ( returns a list of artifacts, which will be generated in abap with user input.)
- generate  request (triggers the generation of all artifacts of UI Service. After objects are generated, the activation or publishing of the service is done automatically)


packages/axios-extension/src/abap/adt-catalog/services/publish-service.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * Publish the service binding.
+     *
+     * @param type - The type of the service binding.
+     * @param bindingName - The name of the service binding.
+     * @returns The response status message.
+     */
+    public async publish(type: string, bindingName: string): Promise<PublishResponse> {
```
docirl: Publish is not needed with latest generation adt api's, not sure if this block is needed

docirl (approved):
There is similarity to the UI Service Gen service in axios-ext, but the flows are different (using a different generator type `published-x-ui-service`). There might be some ways we can align the code for reuse between these generator service types, I can look into this separately.

Overall code looks good
Comment about publish service not needed
Tests added to cover
Not tested manually

packages/axios-extension/src/abap/adt-catalog/services/publish-service.ts
vinayhospete: yes . we can remove this block.

vinayhospete: removed

devinea (approved):
code changes are clear and covered by tests. 
changeset ✅ 

@docirl @vinayhospete - please follow-up on comment https://github.com/SAP/open-ux-tools/pull/2069#pullrequestreview-2925093030 in another PR.

PR #2071, opened 2024-06-24
fix(locate-reuse-lib): corrected extraction of component name

paraggarg89:
"sap.ui5"/"componentUsages" contains object map having "name" property where component name is specified. 
e.g. 

componentUsages: {
            "componentUsage1": {
                "name": "sap.nw.core.applogs.lib.reuse.smarttemplate",
                "lazy": true
            }, 
            "componentUsage2": {
                "name": "sap.s4.cfnd.bpf.lib.reuse.component.smarttemplate",
                "lazy": true
            }
}

So instead of key of componentUsages object, logic should extract component name from "name" property inside the object.


devinea:
Thanks @paraggarg89 
Moving changes to https://github.com/SAP/open-ux-tools/pull/2077 so that the build can run sonar checks

Issue #969, opened 2023-03-29
TBI - refactor `locate-reuse-lib.js` template


devinea:
### Description (include screenshots)
What will this TBI achieve, and how will it be measured as done. Is it documentation, increase of test coverage, refactoring

Refactor `locate-reuse-lib.js` template
- Replace deprecated UI5 API's with newer ui5 API's
- Use Async loading for perforance
- Address lint issues

### Value
Optional - details on how this technical backlog item provides value to end-users or dev team eg. Performance, Quality, Maintainability. This will help determine the ranking in relation to feature development

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [x] Replace deprecated UI5 API's with newer ui5 API's
- [x] Use Async loading for performance
- [x] Address lint issues
- [x] update snapshots in tests
- [ ] enable strict mode on `locate-reuse-libs.js`  https://github.com/SAP/open-ux-tools/pull/966#issuecomment-1497916420
- [ ] Remove the rules below
/*eslint-disable fiori-custom/sap-browser-api-warning, fiori-custom/sap-no-dom-access*/

heimwege:
hey guys 👋🏻 

just stumbled across this issue.
`ui5linter` currently complains about the following 26 findings of the current version of `locate-reuse-lib.js` in main:

```
C:\SAPDevelop\git\grl-template-ui\webapp\test\locate-reuse-libs.js
   45:4  error Access of global variable '$' ($.ajax)
  113:23 error Import of deprecated module 'sap/base/util/UriParameters'
  121:13 error Access of global variable '$' ($.ajax)
  162:1  error Access of global variable 'sap' (sap.registerComponentDependencyPaths)
  171:3  error Access of global variable 'sap' (sap.ui.getCore)
  171:10 error Call to deprecated function 'getCore' (sap.ui.getCore)
  171:20 error Call to deprecated function 'attachInit' of class 'Core'
  172:18 error Access of global variable 'sap' (sap.ui.getCore)
  172:25 error Call to deprecated function 'getCore' (sap.ui.getCore)
  172:35 error Call to deprecated function 'getConfiguration' of class 'Core'
  172:54 error Call to deprecated function 'getLanguage' of class 'Configuration'
  184:5  error Access of global variable 'sap' (sap.ui.getCore)
  184:12 error Call to deprecated function 'getCore' (sap.ui.getCore)
  184:22 error Call to deprecated function 'attachInit' of class 'Core'
[...]

PR #2077, opened 2024-06-25
fix(locate-reuse-lib): corrected extraction of component name

devinea:
Changes from https://github.com/SAP/open-ux-tools/pull/2071. Thanks @paraggarg89

"sap.ui5"/"componentUsages" contains object map having "name" property where component name is specified. 
e.g. 
```JSON
{
...
 componentUsages: {
            "componentUsage1": {
                "name": "sap.nw.core.applogs.lib.reuse.smarttemplate",
                "lazy": true
            }, 
            "componentUsage2": {
                "name": "sap.s4.cfnd.bpf.lib.reuse.component.smarttemplate",
                "lazy": true
            }
 }
...
}
```
So instead of key of componentUsages object, logic should extract component name from "name" property inside the object.

- call createRender using async
https://github.com/SAP/open-ux-tools/issues/969#issuecomment-2211065454


tobiasqueck (changes requested):
Please make sure that the same also works in the preview middleware. I think it should be here: https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/src/flp/init.ts#L93

devinea:
> Please make sure that the same also works in the preview middleware. I think it should be here: https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/src/flp/init.ts#L93

@tobiasqueck  Done. Please re-review. 
Also updated createRender to be async like in init.ts

devinea:
@tobiasqueck Please re-review.

tobiasqueck (approved):
* reviewed only the `preview-middleware-client`
* functionality is properly tested
* changeset ok
* did not test locally

kjose90 (approved):
Code looks good and clear
Changeset ✅

PR #2078, opened 2024-06-25
fix: updating null values

Jimmy-Joseph19:
When updating the values of property values to null. Causes exception and removed the unwanted textNode creation for null cases


packages/fiori-annotation-api/src/avt/to-internal.ts
```diff
@@ -257,10 +257,10 @@ export function convertApplyToInternal(aliasInfo: AliasInformation, apply: Apply
  */
 export function convertPrimitiveValueToInternal(
     type: string,
-    value: string | number | boolean,
+    value: string | number | boolean | undefined,
     aliasInfo: AliasInformation
 ): string {
-    const text = value.toString();
+    const text = String(value);
```
voicis, line 263: We probably should not return `'undefined'` as a value in case the value does not exist.
```suggestion
    const text = value ? value.toString() : '';
```

packages/fiori-annotation-api/src/change-converter.ts
```diff
@@ -582,23 +590,26 @@ export class ChangeConverter {
...
                     uri: file.uri,
                     pointer: pointer + `/attributes/${type}`,
                     newAttributeName: content.value.type,
                     newAttributeValue: newValue
-                };
-                this.annotationFileChanges.push(internalChange);
+                });
             } else if (node.name === valueType) {
                 // element notation
+                let childContent: ElementChild[] = [];
```
voicis, line 602: This can be a constant.

packages/fiori-annotation-api/src/change-converter.ts
```diff
@@ -582,23 +590,26 @@ export class ChangeConverter {
...
+                    childContent.push(createTextNode(newValue));
+                }
                 const internalChange: ReplaceElement = {
                     type: REPLACE_ELEMENT,
                     uri: file.uri,
                     pointer: pointer,
                     newElement: createElementNode({
                         name: content.value.type,
-                        content: [createTextNode(newValue)]
+                        content: [...childContent]
```
voicis, line 612: We do not need to make a copy of the array, we can just use it here.

voicis (changes requested):
Overall changes looks good, but please fix the sonar issue.

packages/fiori-annotation-api/src/change-converter.ts
Jimmy-Joseph19: done

packages/fiori-annotation-api/src/change-converter.ts
Jimmy-Joseph19: adapted

voicis (approved):
Thanks @Jimmy-Joseph19!
- Code looks good
- Good test coverage
- Review comments addressed

nahmed22 (approved):

Issue #2080, opened 2024-06-27
BUG - Duplicated change object in manifest.appdescr_variant template for ADP Project


GDamyanov:
### Description
Duplicated change with type "appdescr_ui5_addNewModelEnhanceWith" occurs in the manifest.appdescr_variant file of Adaptation Project generated with the writer from @sap-ux/adp-tooling in case that writer config contains "appdescr_ui5_addNewModelEnhanceWith" change.

### Steps to Reproduce
1. Generate adp project with the writer from  @sap-ux/adp-tooling
2. Provide change with type "appdescr_ui5_addNewModelEnhanceWith" in config.app.content of the writer
3. Open the manifest.appdescr_variant of the created adp project

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2081, opened 2024-06-27
fix: Handle duplicated change in manifest.appdescr_variant template for Adp Project 

GDamyanov:
fix for: #2080 

- handle duplicated change in manifest.appdescr_variant template for Adp Project


nikmace (approved):
Change looks straightforward
Did not test manually

tobiasqueck (changes requested):
TL;DR: I don't think your implementation does exactly what you intend, additionally, what you are trying to support here, should actually be fixed in the consumer and not here. Either fix the consumer, or do it generically.

In more detail:
* by default we are adding a change of the title which refers to the modified i18n model
* so we require it in the template an i18n model change as well
* if a consumer provides additional changes then, I agree, we shouldn't add duplicates, 
     * however you only check for the change type but not if it is actually the same model
     * or if someone adds a duplicate setTitle change
     * or if someone provides already duplicates with the content property

Please take the following content array for testing and modify your code so that you don't have duplicates
```
[
{ 
      "changeType": "appdescr_ui5_addNewModelEnhanceWith",
      "content": { 
              "modelId": "i18n|sap.suite.ui.generic.template.ObjectPage|ABC" },
              "texts": { "i18n": "i18n/ObjectPage/ABC/i18n.properties" }
},
{
      "changeType": "appdescr_app_setTitle",
      "content": {},
      "texts": {
        "i18n": "i18n/i18n.properties"
      }
}.
{ 
      "changeType": "appdescr_ui5_addNewModelEnhanceWith",
      "content": { 
              "modelId": "i18n|sap.suite.ui.generic.template.ObjectPage|ABC" },
              "texts": { "i18n": "i18n/ObjectPage/ABC/i18n.properties" }
}
]

PR #2083, opened 2024-06-27
feat: Add Adaptation Projects' Add Annotations To OData generator prompting

mmilko01:
https://github.com/SAP/open-ux-tools/issues/2119

- Add prompting relevant to Add Annotations to Odata generator of Adaptation Project
- Add CLI command for executing Add Annotations to Odata for Adaptation Projects


packages/inquirer-common/src/types.ts
```diff
@@ -53,6 +53,10 @@ export interface GuiOptions {
      * Indicate state in the left hand navigation panel in YUI
      */
     breadcrumb?: boolean | string;
+    /**
+     * Indicates the type of the prompt
+     */
+    type?: string;
```
IainSAP, line 59: Can you please add this as a restricted template literal type?

IainSAP (changes requested):

packages/adp-tooling/src/prompts/add-annotations-to-odata/index.ts
```diff
@@ -0,0 +1,77 @@
...
+                hint: t('prompts.fileSelectOptionTooltip')
+            },
+            when: (answers: AddAnnotationsAnswers) => answers.id !== ''
+        } as ListQuestion<AddAnnotationsAnswers>,
+        {
+            type: 'input',
+            name: 'filePath',
+            message: t('prompts.filePathLabel'),
+            guiOptions: {
+                type: 'file-browser',
```
IainSAP, line 55: Please use https://github.com/SAP/open-ux-tools/blob/main/packages/inquirer-common/src/types.ts#L66

IainSAP (changes requested):

packages/adp-tooling/src/prompts/add-annotations-to-odata/index.ts
mmilko01: Fixed with: abc2d5cec3f6012f060e64690c499662cc2cda70

packages/adp-tooling/src/base/helper.ts
```diff
@@ -17,3 +24,109 @@ export function isNotEmptyString(input: string | undefined): boolean {
...
+}
+
+/**
+ * Check if a file already exists in a directory.
+ *
+ * @param {string} filePath - The path to the file.
+ * @param {string} checkDirectory - The directory to check.
+ * @returns {boolean} true if the file exists in the directory, false otherwise
+ */
+export function checkDuplicateFile(filePath: string, checkDirectory: string): boolean {
```
tobiasqueck, line 82: What is the difference between `checkDuplicateFile` and `checkFileExists` ?

packages/create/README.md
```diff
@@ -26,6 +26,11 @@ Calling `sap-ux add html` will add html files for local preview and testing to t
 ```sh
 sap-ux change add html ui5-test.yaml
 ```
+### annotations-to-odata
```
tobiasqueck, line 29: Please rename it to just `annotations`. `sap-ux add annotations` is totally fine. OData is a give in the context of Fiori apps.

tobiasqueck (changes requested):
Review: added two inline comments
Local test: creating a new annotation file results in a completely empty file, could you please instead create already a skeleton like https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-writer/templates/add/annotation.xml. That at least has references to the main vocabularies that a developer usually uses.

packages/inquirer-common/src/types.ts
mmilko01: Fixed with: e192056901a826ecd83af9c45188ea1c68231240

mmilko01:
> Review: added two inline comments Local test: creating a new annotation file results in a completely empty file, could you please instead create already a skeleton like https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-writer/templates/add/annotation.xml. That at least has references to the main vocabularies that a developer usually uses. (The part of the template that is dynamic would require you to connect to the backend and to get the namespace. You don't need to do it here, just don't add it)

@hristotsolev could you please comment here? The current implementation creates an empty `xml` file, do you see any problem in providing a skeleton instead as suggested?

mmilko01:
> Review: added two inline comments Local test: creating a new annotation file results in a completely empty file, could you please instead create already a skeleton like https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-writer/templates/add/annotation.xml. That at least has references to the main vocabularies that a developer usually uses. (The part of the template that is dynamic would require you to connect to the backend and to get the namespace. You don't need to do it here, just don't add it)

@hristotsolev 

> Review: added two inline comments Local test: creating a new annotation file results in a completely empty file, could you please instead create already a skeleton like https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-writer/templates/add/annotation.xml. That at least has references to the main vocabularies that a developer usually uses. (The part of the template that is dynamic would require you to connect to the backend and to get the namespace. You don't need to do it here, just don't add it)

@tobiasqueck we discussed with Hristo and pointed out that the reason we are creating an empty file is that we expect the user to copy/paste the content inside. As far as we understand there is an Annotations modeler tool, that can be used to generate the content of this file. Overall, we are ok with adding the skeleton you proposed, but we are not sure if it will be helpful to the user. Also, if we decide to go with the skeleton you propose, can I reuse your example (without the dynamic part) as I don't completely understand the content and I'm not sure if it is a `dummy` content.

packages/create/README.md
mmilko01: Renamed with: 5a50c963a83e1bb04fb9fabce3b9dec051eb587a

packages/adp-tooling/src/base/helper.ts
mmilko01: After one more look It really seems they are the same :) I refactored the code with: 60ba31705b0cfeb8c22274312dc0cdcbe3de5ae6

packages/adp-tooling/src/base/helper.ts
```diff
@@ -1,19 +1,97 @@
...
 /**
- * Checks if the input is a valid SAP client.
+ * Check if the file exists.
+ *
+ * @param {string} filePath - The path to the file.
+ * @returns {boolean} true if the file exists, false otherwise
+ */
+export function checkFileExists(filePath: string): boolean {
+    return existsSync(filePath);
+}
```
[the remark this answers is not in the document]
mmilko01: Fixed with: 662d64916ee1ccaf365a9e5a41e9a46279350881

tobiasqueck:
>  As far as we understand there is an Annotations modeler tool, that can be used to generate the content of this file. Overall, we are ok with adding the skeleton you proposed, but we are not sure if it will be helpful to the user. Also, if we decide to go with the skeleton you propose, can I reuse your example (without the dynamic part) as I don't completely understand the content and I'm not sure if it is a dummy content.

The annotation model won't be of much help in an empty file. The skeleton I asked you to add, contains the references to the vocabularies required for creating UI annotations, the annotation modeler uses these.
Yes, you should reuse the one I referred to without the dynamic part.
I'd strongly recommend that you also try it out, add an annotation file and then use the Fiori tools annotation modeler extension to e.g. change the table columns shown in a list report.

mmilko01:
> The annotation model won't be of much help in an empty file. The skeleton I asked you to add, contains the references to the vocabularies required for creating UI annotations, the annotation modeler uses these. Yes, you should reuse the one I referred to without the dynamic part. I'd strongly recommend that you also try it out, add an annotation file and then use the Fiori tools annotation modeler extension to e.g. change the table columns shown in a list report.

Added with: 22b9dedeb7891523383ccecfc9637e6a4a04559c

tobiasqueck (approved):
* changeset ok
* code is well structured and understandable
* tested locally after (https://github.com/SAP/open-ux-tools/pull/2083/commits/22b9dedeb7891523383ccecfc9637e6a4a04559c) but not the final PR and it worked as expected
* test coverage good

IainSAP (approved):

testojs (approved):
* Code looks good
* Good test coverage
* Not tested manually - @nnikovSAP will do extensive manual testing

Issue #2087, opened 2024-06-27
BUG - error in parsing ui service generation response


docirl:
### Related Feature
Feature request: https://github.com/SAP/open-ux-tools/issues/2063

### Description
Parsing of ui service generation response is throwing an error. Response is incorrectly expected to be json but is xml, and it attempted to be json parsed.
Service is still generated/published in the background.
Fix is to not json parse, but use existing `parseResponse` function.

### Steps to Reproduce
1. call generate for a ui service bo
2. see error thrown after service is generated

### Expected results
calling the api to generate should return a valid response

### Actual results
response is throwing an error.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

docirl:
Fixed with #2088

PR #2088, opened 2024-06-27
fix(axios-extension): parsing of service gen response

docirl:
https://github.com/SAP/open-ux-tools/issues/2087

Fix parsing of ui service generation response, use `parseResponse` instead of json.parse

### DIFF
--- a/packages/axios-extension/src/abap/adt-catalog/generators/ui-service-generator.ts
+++ b/packages/axios-extension/src/abap/adt-catalog/generators/ui-service-generator.ts
@@ -124,6 +124,6 @@ export class UiServiceGenerator extends AdtService {
                 corrNr: transport
             }
         });
-        return JSON.parse(response.data);
+        return this.parseResponse(response.data);
     }
 }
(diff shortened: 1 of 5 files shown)


packages/axios-extension/src/abap/adt-catalog/generators/ui-service-generator.ts
```diff
@@ -124,6 +124,6 @@ export class UiServiceGenerator extends AdtService {
                 corrNr: transport
             }
         });
-        return JSON.parse(response.data);
+        return this.parseResponse(response.data);
```
Klaus-Keller, line 127: @docirl, in line 120 above we set `Accept` to `+json` and `+xml`. With this change we handle only xml, no json anymore. Are we sure the response is always xml?

docirl: Yes the response is json, and response.data is an xml string.

devinea (approved):
Code changes and tests ✅ 
changeset ✅

Klaus-Keller (approved):
Thanks @docirl,
- fix is easy to understand 
- test coverage is great
- did no local test, just visual review

Approved from my side.

docirl:
thanks @devinea @Klaus-Keller

PR #2091, opened 2024-06-28
fix return null value for expression type null

Jimmy-Joseph19:
New value is set to empty when set to null leading to compilation errors
https://github.com/SAP/open-ux-tools/blob/e8ac45dcce29fe9ce05178d8f7125edf8a7024a9/packages/fiori-annotation-api/src/change-converter.ts#L584

**Actual Result** `ScaleFactor: `
**ExpectedResult**: `ScaleFactor:  null`

### DIFF
--- a/.changeset/two-lions-explode.md
+++ b/.changeset/two-lions-explode.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fiori-annotation-api': patch
+---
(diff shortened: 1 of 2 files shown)


voicis (approved):
- Straight forward change
- Good code coverage
- Did not test locally

Issue #2089, opened 2024-06-28
FEATURE - Using preview-middleware with cds-launchpad-plugin


marianfoo:
_As a **Developer**,
I want to use the `preview-middleware` to show the UI5 apps in a launchpad,
so that I can see all UI5 Apps in a automatically created FLP._
### Description

In my CAP projects, i use [`cap-launchpad-plugin`](https://github.com/geert-janklaps/cds-launchpad-plugin) to create a FLP with all UI5 Applications i defined in the CAP Project, without create manually a FLP html page.
A similar approach like [`preview-middleware`](https://github.com/SAP/open-ux-tools/tree/main/packages/preview-middleware).

### Technical Design
In the [`cap-launchpad-plugin`](https://github.com/geert-janklaps/cds-launchpad-plugin) we also have [templates](https://github.com/geert-janklaps/cds-launchpad-plugin/tree/main/templates) which will be filled the the necessary data at runtime (see code [here](https://github.com/geert-janklaps/cds-launchpad-plugin/blob/55e5fb938c1e969007f5c2ccf5fc4f1827dcc3f6/src/index.ts#L93-L109)).
Ideally i would like to replace this with the `preview-middleware`.
But when i look at the code, as this module is intended to be used a a ui5 tooling middleware, it is deeply connected to `@ui5/server`, but in CAP uses it own express server.
I also can find a API  like _here are the all the options, return the complete html file_. I think close to it is [`generateSandboxForEditor`](https://github.com/SAP/open-ux-tools/blob/e8ac45dcce29fe9ce05178d8f7125edf8a7024a9/packages/preview-middleware/src/base/flp.ts#L272), but still missing configuration like all the apps data.

### Acceptance Criteria

**Given** I have a API to create a FLP HTML Pages
**when** I execute the `cds-launchpad-plugin`
**then** a FLP Page will automatically created.

### Notes

I don't know if this feature request is even in the scope of this module.
I absolutely understand if the deep integration to the ui5 tooling should not be removed because there is a lot to refactor.
Theoretically the cap-launchpad-plugin could use the ui5 tooling, but I don't know how to integrate it with CAP Server.

Even if the feature request is not implemented, an architecture help on how to proceed would be very nice!
Generally the goal is to use your open source packages.

Cheers!

tobiasqueck:
[...]

Issue #1332, opened 2023-09-25
FEATURE - add eject functionality to preview middleware


Klaus-Keller:
As a **developer**,
I want to **adjust the HTML file that is used to preview an app**,
so that I **can do local changes to manipulate the app preview**.

### Description
UI5 middleware `@sap-ux/preview-middleware` generates the HTML and JavaScript files on the fly to preview the application.  While this is perfectly fine for most cases, there are situations in which it is helpful to have a 'materialized' HTML that can be edited. The idea is to have an 'eject' function that writes the HTML ~~and JavaScript~~ files to the local disc. If those files exist, they should be used instead of generating files.

### Technical Design
- Implementation of eject is in `@sap-ux/preview-middleware`, files are written to virtual file system `mem-fs`
- Module `@sap-ux/create` consumes eject from middleware, files are written here 

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a SAP Fiori project with middleware configuration
**when** I execute `@sap-ux/create add preview-html` 
**then** the HTML for previewing the SAP Fiori app will be written to the file system.

### Tasks
- [ ] Eject functionality
- [ ] Unit Tests added
- [ ] Test Cases defined

Klaus-Keller:
We've discussed to eject HTML files only. JavaScript files will be still virtually served. In order to overwrite or extend JavaScript content, the HTML file can be ejected and the scripts can be changed to point to a local file. The local file needs to be manually created.

PR #2093, opened 2024-06-28
Enhancing the `generatePreviewFiles` to support all configurations and CAP

tobiasqueck:
Adds support for #2089 

Improvements for #1332:
* improved README
* more meaningful function names
* more flexible internal functions
* handles the previously ignored `apps` config

### DIFF
--- a/packages/create/src/cli/add/html.ts
+++ b/packages/create/src/cli/add/html.ts
@@ -13,7 +13,7 @@ import type { AdpPreviewConfig } from '@sap-ux/adp-tooling';
  *
  * @param cmd - commander command for adding navigation inbounds config command
  */
-export function addLocalHtmlFiles(cmd: Command): void {
+export function addAddHtmlFilesCmd(cmd: Command): void {
     cmd.command('html [path]')
         .option('-c, --config <string>', 'Path to project configuration file in YAML format', 'ui5.yaml')
         .option('-s, --simulate', 'simulate only do not write config; sets also --verbose')
--- a/.changeset/twelve-walls-teach.md
+++ b/.changeset/twelve-walls-teach.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/preview-middleware': minor
+---
+
+Add support for using the preview in CAP projects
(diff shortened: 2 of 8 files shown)


tobiasqueck:
@heimwege this one might be interesting for your setup as well.

packages/create/src/cli/add/html.ts
```diff
@@ -13,7 +13,7 @@ import type { AdpPreviewConfig } from '@sap-ux/adp-tooling';
  *
  * @param cmd - commander command for adding navigation inbounds config command
  */
-export function addLocalHtmlFiles(cmd: Command): void {
+export function addAddHtmlFilesCmd(cmd: Command): void {
```
heimwege, line 16: nitpick: I had to read it twice to understand that this is no typo 😸  But not sure if shuffling the words to `addCmdAddHtmlFiles` would be any better.

schreckstefan (approved):
Thanks for improving the app handling. Looks good now, when I check the unit tests and snapshots

packages/create/src/cli/add/html.ts
tobiasqueck: We should have a consistent naming for all of them:
```
    // create-fiori add mockserver-config
    addAddMockserverConfigCommand(addCommands);
    // create-fiori add smartlinks-config
    addAddSmartLinksConfigCommand(addCommands);
    // create-fiori add cds-plugin-ui5
    addAddCdsPluginUi5Command(addCommands);
    // create-fiori add inbound-navigation-config
    addInboundNavigationConfigCommand(addCommands);
    // create-fiori add cards-editor
    addCardsEditorConfigCommand(addCommands);
    // create-fiori add html
    addAddHtmlFilesCmd(addCommands);
```

Is your proposal `addCmd<object>`?

heimwege: Oh there exist more of those. I wasn't aware of that 🙈 So actually yes that would have been my proposal, but I also don't want you to invest too much effort because of my grammatical nitpicking. So it's also fine for me as is.

tobiasqueck: I think I would prefer not changing it in this one but have a separate PR later and fix it for all names.

heimwege (approved):
lgtm
- very good test coverage
- changeset ok
- did NOT (yet) test locally

Issue #2094, opened 2024-06-28
FEATURE - Add Abap On Premise prompting support to `@sap-ux/odata-service-inquirer`


IainSAP:
### Description

Add prompts to connect to Abap on premise systems and select a service

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a ...
**when** I execute ...
**then** ... will happen.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

IainSAP:
Done

PR #2095, opened 2024-06-28
Fix/2060/remove unimplmented message

IainSAP:
#2094 

- Adds Abap on prem prompts

**TODO:**
- Add tests
- Reduce prompt getter function complexity
- Wire up new system type prompt (currently not returned)


IainSAP:
Wrong branch

PR #2097, opened 2024-06-29
Fix/preview middleware/test racing condition

heimwege:
Use [Qunit preconfig](https://qunitjs.com/api/config/#preconfiguration) to avoid racing condition in qunit tests. Now that preconfigs [no longer crash the ui5-test-runner](https://github.com/ArnaudBuchholz/ui5-test-runner/commit/84c0dce0d36a082b4558e6072bd49aaf57311450) (fixed in version 5.3.0) we should go for it.


tobiasqueck (approved):
* changeset ok
* locally tested the code

packages/preview-middleware/templates/test/qunit.js
```diff
@@ -11,18 +11,21 @@ sap.ui.loader.config({
         },
         "sap/ui/qunit/sinon-qunit-bridge": {
             deps: ["sap/ui/thirdparty/qunit", "sap/ui/thirdparty/sinon-4"]
```
petermuessig, line 13: Why is this the only shim using `qunit` instead of `qunit-2`?

packages/preview-middleware/templates/test/qunit.js
```diff
@@ -11,18 +11,21 @@ sap.ui.loader.config({
...
-        }        
+        }
     }
 });
 
+window.QUnit ??= {};
```
petermuessig, line 18: BTW: wouldn't it look nicer this way:

```js
window.QUnit = Object.assign({}, { config: { autostart: false } }, window.QUnit);
```

petermuessig (commented):
One question and one suggestion

packages/preview-middleware/templates/test/qunit.js
heimwege: Oh I think you are right. I'll fix it as part of this PR as well 👍🏻

packages/preview-middleware/templates/test/qunit.js
heimwege: Good idea but what's the need for source1 = window.QUnit? Isn't this the same as

```js
window.QUnit = Object.assign({}, { config: { autostart: false } });
```

🤔

heimwege: I think I understood your intention. It's to ensure that is does not override QUnit in case QUnit already exists (so this would not be a preconfig). Right?

petermuessig (approved):
Looks brilliant... 😄

tobiasqueck (approved):

PR #2099, opened 2024-07-01
fix:  Add cap index html template handling logic 

kjose90:

- Modified `ui5-application-writer` and updated template for `templates/core/webapp/index.html` file to dynamically display the script source based on the `projectType`.
For CAP projects, it will use` ui5.frameworkUrl `and `ui5.version` to construct fully qualified url.
For EDMX projects, it will use the relative path.

- Added unit tests to `ui5-application-writer` to support for `projectType` CAP, validating that the framework URL and version are correctly included for CAP projects.

- Updated snapshots in `fiori-elements-writer`:
Updated snapshots in` lrop-test.ts` to support fully qualified url for lrop projects `projectType` is CAP.
Snapshots updates for testcase: 
`lrop_v4_addtests_cds`
`lrop_v4_addtests_cds_typescript`

- Updated snapshots in `fiori-freestyle-writer` to support fully qualified url for lrop projects where `projectType` is CAP.
Snapshots updates for testcase: 
`basic_cap`
`basic_cap_typescript`

### DIFF
--- a/packages/fiori-freestyle-writer/test/__snapshots__/basic.test.ts.snap
+++ b/packages/fiori-freestyle-writer/test/__snapshots__/basic.test.ts.snap
@@ -17,17 +17,10 @@ Object {
   \\"dependencies\\": {},
   \\"devDependencies\\": {
     \\"@ui5/cli\\": \\"^3.0.0\\",
-    \\"@sap/ux-ui5-tooling\\": \\"1\\",
-    \\"@sap-ux/ui5-middleware-fe-mockserver\\": \\"2\\"
+    \\"@sap/ux-ui5-tooling\\": \\"1\\"
   },
   \\"scripts\\": {
-    \\"start\\": \\"fiori run --open \\\\\\"test/flpSandbox.html?sap-ui-xx-viewCache=false#nods1-tile\\\\\\"\\",
-    \\"start-local\\": \\"fiori run --config ./ui5-local.yaml --open \\\\\\"test/flpSandbox.html?sap-ui-xx-viewCache=false#nods1-tile\\\\\\"\\",
-    \\"build\\": \\"ui5 build --config=ui5.yaml --clean-dest --dest dist\\",
-    \\"deploy\\": \\"fiori verify\\",
-    \\"deploy-config\\": \\"fiori add deploy-config\\",
-    \\"start-noflp\\": \\"fiori run --open \\\\\\"index.html?sap-ui-xx-viewCache=false\\\\\\"\\",
-    \\"start-mock\\": \\"fiori run --config ./ui5-mock.yaml --open \\\\\\"test/flpSandbox.html?sap-ui-xx-viewCache=false#nods1-tile\\\\\\"\\"
+    \\"deploy-config\\": \\"npx -p @sap/ux-ui5-tooling fiori add deploy-config cf\\"
   }
 }
 ",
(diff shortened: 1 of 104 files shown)


packages/ui5-application-writer/src/index.ts
```diff
@@ -39,7 +39,18 @@ async function generate(basePath: string, ui5AppConfig: Ui5App, fs?: Editor): Pr
         ignore.push('**/gitignore.tmpl');
     }
 
-    fs.copyTpl(join(tmplPath, 'core', '**/*.*'), join(basePath), ui5App, undefined, {
+    // Determine the framework URL and version based on the project type, if project is not EDMX  include ui5App.ui5.frameworkUrl & ui5App.ui5.version
+    const frameworkUrl = !isEdmxProjectType && ui5App.ui5.frameworkUrl ? ui5App.ui5.frameworkUrl + '/' : undefined;
+    const version = !isEdmxProjectType && ui5App.ui5.version ? ui5App.ui5.version + '/' : undefined;
```
tobiasqueck, line 44: I think you should ignore the version if there is no framework url because something like `/1.172.1/resources/...` wouldn't work anywhere.

tobiasqueck (changes requested):
I also think it is not a good idea to change the version to version + `/` because now it is not a version anymore. What happens if someone else uses `ui5.version` in a different file in the template folder.
Additionally, setting the framework url to undefined if you don't have CAP is also a bad idea. It still is the valid framework url, it should just not be used in the `index.html` template (for CAP).

Please rethink the approach and consider these kind of side-effects when changing the template config.

packages/ui5-application-writer/test/index.test.ts
```diff
@@ -104,6 +104,42 @@ describe('UI5 templates', () => {
...
+    });
+
+    it('Check webapp/index.html templates are generated correctly for EDMX application', async () => {
+        const projectDir = join(outputDir, 'testapp-cap');
+        ui5AppConfig.app.projectType = 'EDMXBackend';
+        const fs = await generate(projectDir, { ...ui5AppConfig, ui5: { version: '1.96.1' } });
+        const indexHtmlPath = join(projectDir, 'webapp/index.html');
+        // Check if webapp/index.html exists
+        expect(fs.exists(indexHtmlPath)).toBe(true);
+        // Check if the index.html contains the correct UI5 framework URL with version included
```
devinea, line 138:
```suggestion
        // Check if the index.html contains the correct UI5 framework URL without version included
```
Typo?

kjose90:
> I also think it is not a good idea to change the version to version + `/` because now it is not a version anymore. What happens if someone else uses `ui5.version` in a different file in the template folder. Additionally, setting the framework url to undefined if you don't have CAP is also a bad idea. It still is the valid framework url, it should just not be used in the `index.html` template (for CAP).
> 
> Please rethink the approach and consider these kind of side-effects when changing the template config.

@tobiasqueck Thanks for the feedback. Does this [approach](https://github.com/SAP/open-ux-tools/pull/2099/files#diff-1bff8e3813a4755fbee9efa4f5c4d3cbb9cf7c0b8bbf32e83399a394320e1060R43) look okay?

tobiasqueck:
> @tobiasqueck Thanks for the feedback. Does this [approach](https://github.com/SAP/open-ux-tools/pull/2099/files#diff-1bff8e3813a4755fbee9efa4f5c4d3cbb9cf7c0b8bbf32e83399a394320e1060R43) look okay?

Yes. (you could optimize the code by combining the if and the else block (since they do the same) but otherwise, my concerns are addressed.

packages/fiori-freestyle-writer/test/__snapshots__/basic.test.ts.snap
```diff
@@ -164,7 +164,7 @@ title=App Title",
     </style>
     <script
         id=\\"sap-ui-bootstrap\\"
-        src=\\"resources/sap-ui-core.js\\"
+        src=\\"https://ui5.sap.com/1.78.11/resources/sap-ui-core.js\\"
```
longieirl, line 167: I'm wondering how this will affect projects deployed to CF? If its hardcode, then they'll have to change the files in source control and redeploy.
This is why the destination is used in CF deployment i.e. 

https://github.com/SAP-samples/fiori-tools-samples/blob/main/cap/cap-fiori-mta/app/feproject-ui/xs-app.json#L15

https://github.com/SAP-samples/fiori-tools-samples/blob/main/cap/cap-fiori-mta/mta.yaml#L100

The advantage of this is, the developer can login to their subaccount, select the service instance and change the UI5 URL to point any version thus not requiring a rebuild and redeploy.

packages/fiori-freestyle-writer/test/__snapshots__/basic.test.ts.snap
```diff
@@ -862,7 +862,7 @@ title=App Title",
...
```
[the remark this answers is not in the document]
kjose90: > @kitzkan Could you explain why this has changed please...getting harder to see what test is associated with which snapshot entry in the file. Im guessing this is a CAP test?

Yes, this is a CAP test located in `packages/fiori-freestyle-writer/test/__snapshots__/basic.test.ts.snap`, specifically for configurations named `basic_cap` and `basic_cap_typescript`. Please refer to the [configurations](https://github.com/SAP/open-ux-tools/blob/d6af42e0aaaa1855ce25eb9829c4c5069c9c0974/packages/fiori-freestyle-writer/test/basic.test.ts#L170) for more details.

packages/ui5-application-writer/src/index.ts
```diff
@@ -39,11 +39,24 @@ async function generate(basePath: string, ui5AppConfig: Ui5App, fs?: Editor): Pr
         ignore.push('**/gitignore.tmpl');
     }
 
-    fs.copyTpl(join(tmplPath, 'core', '**/*.*'), 
-        join(basePath), ui5App, undefined,  {
-            globOptions: { dot: true, ignore },
-            processDestinationPath: (filePath: string) => filePath.replace(/gitignore.tmpl/g, '.gitignore')
-        });
+    // Determine the UI5 resource URL based on project type and UI5 framework details
+    let ui5ResourceUrl: string;
```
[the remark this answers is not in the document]
kjose90: > Is it worth placing the resource url construction code into a separate function?

I moved construction code to [here](https://github.com/SAP/open-ux-tools/pull/2099/files#diff-8651a8bccdacbc50e0fc8b935e886f22d73e6ad4a7b3f02da31c28b740bafa7dR136). Let me know if thats okay?

packages/fiori-freestyle-writer/test/__snapshots__/basic.test.ts.snap
devinea: This is how CAP has been configured in Fiori tools to support loading during local preview as no proxy was provided by `cds-serve`. This has changed e.g. for typescript configurations using https://www.npmjs.com/package/cds-plugin-ui5. But a general change to this behaviour is out of scope of the current TBI as it has wider implications to be considered in a separate TBI.

devinea (approved):
Changes look good and snapshot updates as expected. 
Please add a changset

longieirl (approved):
Changes look good.
No changeset required for merging into feature branch.
Tests look good.

PR #2100, opened 2024-07-01
Fix/preview middleware/path issue

heimwege:
Let's also include windows users 🐱

### DIFF
--- a/packages/preview-middleware/src/base/flp.ts
+++ b/packages/preview-middleware/src/base/flp.ts
@@ -263,7 +263,7 @@ export class FlpSandbox {
         for (const app of this.config.apps) {
             let manifest: Manifest | undefined;
             if (app.local) {
-                manifest = JSON.parse(readFileSync(join(app.local, 'webapp/manifest.json'), 'utf-8'));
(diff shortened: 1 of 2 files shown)


packages/preview-middleware/src/base/flp.ts
```diff
@@ -263,7 +263,7 @@ export class FlpSandbox {
         for (const app of this.config.apps) {
             let manifest: Manifest | undefined;
             if (app.local) {
-                manifest = JSON.parse(readFileSync(join(app.local, 'webapp/manifest.json'), 'utf-8'));
+                manifest = JSON.parse(readFileSync(join(app.local, 'webapp', 'manifest.json'), 'utf-8'));
```
tobiasqueck, line 266: What? That should also work in Windows. `join` should be taking care of it. I am using this all over the place.

heimwege: Arg. Nevermind. The transpiled js sources tricked me and I slipped in the line 😞
My bad. Thou shalt not do too many things in parallel

PR #2090, opened 2024-06-28
refactor: Exclude writing of ui5 local yaml and gitignore files if projectType is CAP

kjose90:
#2031

Added new option: `projectType` to App interface to determine if the project is `EDMXBackend`, `CAPJava`, or `CAPNodejs`.

- In `ui5-application-writer`, the `projectType` flag is used to exclude the `ui5-local.yaml` and `.gitignore files` from the templates for CAPJava and CAPNodejs projects.
- In `fiori-freestyle-writer`, the `ui5-local.yaml ` file is written only if the projectType is not CAPJava or CAPNodejs.

Updated Tests:

- Added tests to `ui5-application-writer` to ensure that `ui5-local.yaml `and `.gitignore` files are excluded for CAPNodejs projects.
- Updated `basic.test.ts` in `fiori-freestyle-writer` to verify that `ui5-local.yaml `and `.gitignore` files are not generated for CAPNodejs projects.
- Updated `lrop.test.ts` in `fiori-elements-writer` to verify that `ui5-local.yaml `and `.gitignore` files are not generated for CAPNodejs projects.

PR #2102, opened 2024-07-01
fix: Exclude scripts from package json for cap project

kjose90:

- Added a check to the `ui5-application-writer`, `odata-service-writer`, `fiori-freestyle-writer`, and `fiori-elements-writer `modules to conditionally modify the scripts section in the `package.json` file. 
- For CAP projects, excluded the following scripts: `start`, `start-local`, `build`, `start-noflp`, `start-mock`, `int-test`, and `deploy`. 
- Only the `deploy-config` script is included in the `package.json` for CAP projects.
- Updated unit tests

### DIFF
--- a/packages/odata-service-writer/test/index.test.ts
+++ b/packages/odata-service-writer/test/index.test.ts
@@ -54,8 +54,7 @@ describe('ODataService templates', () => {
         const ui5Yaml = (await UI5Config.newInstance('')).addFioriToolsProxydMiddleware({ ui5: {} }).toString();
         fs.write(join(testDir, 'ui5.yaml'), ui5Yaml);
         fs.write(join(testDir, 'ui5-local.yaml'), '');
-        const packageJson = serviceType === ServiceType.CDS ? {} : { ui5: { dependencies: [] } };
-        fs.writeJSON(join(testDir, 'package.json'), packageJson);
+        fs.writeJSON(join(testDir, 'package.json'), {});
         fs.write(
             join(testDir, 'webapp/manifest.json'),
             JSON.stringify({
--- a/packages/odata-service-writer/test/index.test.ts
+++ b/packages/odata-service-writer/test/index.test.ts
@@ -49,12 +49,13 @@ describe('ODataService templates', () => {
      * @param name testDir name
      * @returns testDir Path
      */
-    async function createTestDir(name: string): Promise<string> {
+    async function createTestDir(name: string, serviceType?: ServiceType): Promise<string> {
         const testDir = join(outputDir, name);
         const ui5Yaml = (await UI5Config.newInstance('')).addFioriToolsProxydMiddleware({ ui5: {} }).toString();
         fs.write(join(testDir, 'ui5.yaml'), ui5Yaml);
         fs.write(join(testDir, 'ui5-local.yaml'), '');
-        fs.writeJSON(join(testDir, 'package.json'), { ui5: { dependencies: [] } });
+        const packageJson = serviceType === ServiceType.CDS ? {} : { ui5: { dependencies: [] } };
+        fs.writeJSON(join(testDir, 'package.json'), packageJson);
         fs.write(
             join(testDir, 'webapp/manifest.json'),
             JSON.stringify({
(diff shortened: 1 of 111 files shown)


kjose90:
@devinea 
[comment 1](https://github.com/SAP/open-ux-tools/pull/2090#discussion_r1665426608) & [comment 2](https://github.com/SAP/open-ux-tools/pull/2090#discussion_r1665461251) should be resolved here

packages/odata-service-writer/test/index.test.ts
```diff
@@ -49,12 +49,13 @@ describe('ODataService templates', () => {
...
      */
-    async function createTestDir(name: string): Promise<string> {
+    async function createTestDir(name: string, serviceType?: ServiceType): Promise<string> {
         const testDir = join(outputDir, name);
         const ui5Yaml = (await UI5Config.newInstance('')).addFioriToolsProxydMiddleware({ ui5: {} }).toString();
         fs.write(join(testDir, 'ui5.yaml'), ui5Yaml);
         fs.write(join(testDir, 'ui5-local.yaml'), '');
-        fs.writeJSON(join(testDir, 'package.json'), { ui5: { dependencies: [] } });
+        const packageJson = serviceType === ServiceType.CDS ? {} : { ui5: { dependencies: [] } };
+        fs.writeJSON(join(testDir, 'package.json'), packageJson);
```
devinea, line 58:
```suggestion
```
I don't think we need to do this update at all. `ui5: { dependencies: [] }` was needed for @ui5/cli V1 and V2. It is no longer needed in V3 which we currently write to the project.

packages/odata-service-writer/test/index.test.ts
```diff
@@ -70,6 +71,17 @@ describe('ODataService templates', () => {
         const testDir = await createTestDir('odata-service-v2');
         await generate(testDir, validServiceConfig as OdataService, fs);
         expect(fs.dump(testDir)).toMatchSnapshot();
+        const packagePath = join(testDir, 'package.json');
+        expect(fs.readJSON(packagePath)).toEqual({
+            ui5: {
+              dependencies: [ '@sap-ux/ui5-middleware-fe-mockserver', '@sap/ux-ui5-tooling' ]
+            },
```
devinea, line 78:
```suggestion
```

devinea (commented):
Code changes look like they cover the requirement well. 
Couple of suggestions I noticed could be cleaned up while in that area of code.

packages/odata-service-writer/test/index.test.ts
kjose90: resolved [here](https://github.com/SAP/open-ux-tools/pull/2102/files#diff-d7cf042154f276c73f8188e4c58f5032491c0cd869b9dc815fd8d293dce0442dR57)

longieirl (approved):
Interesting that is is passing in tools-suite as I would have expected the CAP flows to have caught this!
Changes are required and test coverage is good.

Here is a sample app with the deploy config added;

https://github.com/SAP-samples/fiori-tools-samples/blob/main/cap/cap-fiori-mta/app/feproject-ui/package.json

packages/odata-service-writer/test/index.test.ts
kjose90: I have removed this dependency.

cianmSAP (changes requested):
Changeset required

Otherwise looks good

cianmSAP (approved):
My apologies, changeset not required in this case

Updates look good, 
tests to cover

packages/ui5-application-writer/src/options.ts
```diff
@@ -111,8 +111,13 @@ export async function applyOptionalFeatures(
     tmplPath: string,
     ui5Configs: UI5Config[]
 ) {
+    const isEdmxProjectType = ui5App.app.projectType === 'EDMXBackend';
     if (ui5App.appOptions) {
         for (const [key, value] of Object.entries(ui5App.appOptions)) {
+            if(key === 'sapux' && !isEdmxProjectType) {
```
IainSAP, line 117: Could we instead remove the option before calling this code instead of embedding the business logic directly here please? This will make the code cleaner.

PR #2103, opened 2024-07-02
fix: update access modifier

nahmed22:
expose vocabulary service to use outside of the package.


voicis (approved):
Thanks @nahmed22!
- Simple change
- Changeset present

vinayhospete (approved):

PR #2104, opened 2024-07-02
chore upgrade nx

devinea:
- Apply nx module upgrades
- run nx migration
- fix nx.json config for scripts that use nx (e.g. lint and lint:fix) to ensure task is run on sub modules.
- Apply lint fixes


Klaus-Keller (approved):
Thanks @devinea!
- update of nx and config looks good
- fix for linting issues make sense
- not 100% sure if we need a change set, but I think probably not

Approved from my side.

Issue #2111, opened 2024-07-03
BUG - Templates Not Included in files Array of fiori-generator-shared Package json


kjose90:
### Related Feature
#2019

### Description
Templates Not Included in files Array of `fiori-generator-shared` Package. Currently, the `fiori-generator-shared` package is missing the inclusion of templates in its files array within the package.json. This omission prevents the templates from being included when the package is published or used as a dependency, which affects the usability of the package.

### Expected results
The files array in package.json should include the templates directory to ensure that all necessary templates are included when the package is installed.

### Actual results
The templates directory is missing from the files array, causing templates to be excluded from installation.

kjose90:
Fix provided in https://github.com/SAP/open-ux-tools/pull/2113
Read me templates available in published npm version.

PR #2113, opened 2024-07-03
Fix: Include Templates in 'files' Array of fiori-generator-shared Package.json

kjose90:
#2111

Added missing templates folder to the files array in the `package.json` of `fiori-generator-shared` to be included in installation of package.


devinea (approved):
Update lg. 
changeset ✅

longieirl (approved):
Changes look good.

PR #2115, opened 2024-07-03
feat(project-access): support freestyle CAP

PetrStribrny:
Add freestyle CAP app support.


nahmed22 (dismissed):

Klaus-Keller (changes requested):
@PetrStribrny, the change set is missing. This should be a minor change.

PetrStribrny:
> @PetrStribrny, the change set is missing. This should be a minor change.

@Klaus-Keller changeset added, please check if `patch` ic correct type. Thanks.

.changeset/short-planes-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/project-access': patch
```
Klaus-Keller, line 2: @PetrStribrny as mentioned, I think this is minor as the result of call `getProject()` as more data. 
```suggestion
'@sap-ux/project-access': minor
```

PetrStribrny: > @PetrStribrny as mentioned, I think this is minor as the result of call `getProject()` as more data.

@Klaus-Keller done

Klaus-Keller (approved):
Thanks @PetrStribrny,
- code changes to include freestyle applications look good
- changeset is correct
- test coverage is good
- did local test, no issues found

Approved from my side.

nahmed22 (approved):
Looks good to me, 

Good test coverage and changeset maintained.

PR #2116, opened 2024-07-04
fix: skip adding comma for elements that are getting deleted

Jimmy-Joseph19:
Currently comma is inserted for elements that are marked for deletion which leads to an additional trailing comma and compilation errors.


packages/fiori-annotation-api/src/cds/writer.ts
```diff
@@ -1416,3 +1399,26 @@ function processNode(content: ContainerContentBlock[], item: Comment | AstNode):
...
+) {
+    return !!changes.find((change) => {
+        if (!change.type.startsWith('delete')) {
+            return false;
+        }
+        const astNodes = getAstNodesFromPointer(document, change.pointer);
+        const toBeDeletedNodes = astNodes[astNodes.length - 1];
+        const node = content[insertAfterIndex];
+        if (node.type === 'element' && node?.element === toBeDeletedNodes) {
+            if (toBeDeletedNodes.type === 'annotation' && content.length === 1) {
```
Jimmy-Joseph19, line 1416: This condition is for following case
https://github.com/SAP/open-ux-tools/blob/47c1b25e04ff418a09ab9bbc5ea378d41edbf4b2/packages/fiori-annotation-api/test/unit/cds/writer.test.ts#L611

Jimmy-Joseph19: Here the target is coverted to annotationGroup after deletion

packages/fiori-annotation-api/src/cds/writer.ts
```diff
@@ -963,54 +990,10 @@ function printChange(parent: ContainerNode | undefined) {
...
-            return {
-                position: anchor,
-                commaPosition: copyPosition(previousElement.element.range.end)
-            };
-        }
-        return { position: anchor };
-    }
-    return undefined;
+function indexInserts<T extends ElementInserts>(change: T, containerSize: number): [number | undefined, T] {
+    const index =
```
voicis, line 994: We probably can clean up this function even further. Only the number is currently used and change is passed through which isn't necessary. 

We can also consider removing the function all together, but maybe having a utility function for resolving index would be good to have.

packages/fiori-annotation-api/src/cds/writer.ts
voicis: That is compound annotation. Just not to mix it up with the annotation group AST node which groups multiple annotations from the same vocabulary.

packages/fiori-annotation-api/test/unit/cds/writer.test.ts
```diff
@@ -772,6 +772,115 @@ annotate S.E with {
...
+        {
+            $Type : 'UI.DataField',
+            Value : test,
+            Criticality : #Critical
+        },
+    ];`;
+                await testWriter(
+                    fixture,
+                    [
+                        {
```
voicis, line 788: I would encourage using factory functions for creating changes. It makes tests a bit more concise and avoids having to repeat the type constants.

packages/fiori-annotation-api/test/unit/cds/writer.test.ts
```diff
@@ -772,6 +772,115 @@ annotate S.E with {
...
+                        },
+                        {
+                            type: 'delete-record-property',
+                            pointer: '/targets/0/assignments/0/value/properties/0'
+                        }
+                    ],
+                    `
+    Service S { entity E {}; };
+    annotate S.E with @UI.DataPoint  : {
+
```
voicis, line 879: It looks like we have some extra whitespace here. Is it difficult to remove it?

packages/fiori-annotation-api/src/cds/writer.ts
Jimmy-Joseph19: Changed.

packages/fiori-annotation-api/test/unit/cds/writer.test.ts
Jimmy-Joseph19: Changed.

packages/fiori-annotation-api/test/unit/cds/writer.test.ts
Jimmy-Joseph19: Its not an easy fix if i call the parent range to delete can cause overlapping edits. Will address in another PR

voicis (approved):
- Code changes looks good
- Review comments addressed
- Good test coverage

nahmed22 (approved):
test coverage is good, changeset is maintained.

Issue #2121, opened 2024-07-04
BUG - Preview Middleware - Integration tests fail for older UI5 versions below 1.100.0


devinea:
### Related Feature
Feature request: _issue number_

### Description
A clear description of what the bug is.

### Steps to Reproduce
```
pnpm install
pnpm run build
cd packages/preview-middleware
pnpm run test:integration:run --debug
```

### Expected results
tests pass

### Actual results
tests fail. 
On CI tests timeout with no clear failure reason.

<img width="1655" alt="Screenshot 2024-07-04 at 15 47 11" src="https://github.com/SAP/open-ux-tools/assets/10102054/d329ef7c-9b4d-4e9d-b321-0f7e97f3e348">

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

devinea:
Fixes merged. Thanks.

PR #2124, opened 2024-07-05
fix: Missing Scenario API in lower SAPUI5 versions

mmilko01:
https://github.com/SAP/open-ux-tools/issues/2121

- Consume scenario from control property editor common

### DIFF
--- a/packages/preview-middleware-client/src/flp/init.ts
+++ b/packages/preview-middleware-client/src/flp/init.ts
@@ -1,11 +1,13 @@
 import Log from 'sap/base/Log';
 import type AppLifeCycle from 'sap/ushell/services/AppLifeCycle';
 import type { InitRtaScript, RTAPlugin, StartAdaptation } from 'sap/ui/rta/api/startAdaptation';
+import scenarios, { type Scenario } from 'sap/ui/fl/Scenario';
 import type { FlexSettings, RTAOptions } from 'sap/ui/rta/RuntimeAuthoring';
 import IconPool from 'sap/ui/core/IconPool';
 import ResourceBundle from 'sap/base/i18n/ResourceBundle';
 import AppState from 'sap/ushell/services/AppState';
-import type Localization from 'sap/base/i18n/Localization';
+import { getManifestAppdescr } from '../adp/api-handler';
+
 /**
  * SAPUI5 delivered namespaces from https://ui5.sap.com/#/api/sap
  */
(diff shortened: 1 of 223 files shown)


nikmace (dismissed):
Simple change
Changeset present
Tests adjusted
Did not test manually

packages/preview-middleware-client/src/flp/init.ts
```diff
@@ -1,7 +1,7 @@
 import Log from 'sap/base/Log';
 import type AppLifeCycle from 'sap/ushell/services/AppLifeCycle';
 import type { InitRtaScript, RTAPlugin, StartAdaptation } from 'sap/ui/rta/api/startAdaptation';
-import scenarios, { type Scenario } from 'sap/ui/fl/Scenario';
+import { scenario as scenarios, type Scenario } from '@sap-ux-private/control-property-editor-common';
```
tobiasqueck, line 4: That looks odd. Why do we have this singular vs. plurar change?

nikmace: This is used in the method that accepts the scenario as an argument:

https://github.com/SAP/open-ux-tools/blob/8493f7fc3ff72629d8f2c55b0dceaf90bed74329/packages/preview-middleware-client/src/flp/init.ts#L198

What would be a better name here?

mmilko01: In the code after that I need to declare a `scenario` variable and there was a clash of the names. As the import here is actually a structure constant, containing all available scenarios, to me it made sense to be plural.

voicis (commented):
@mmilko01 Can you please also add `since` JSDoc comment to the type definitions in `sap.ui.fl.d.ts`? So it is clearer in to future that this is not available for all UI5 versions.

mmilko01:
> @mmilko01 Can you please also add `since` JSDoc comment to the type definitions in `sap.ui.fl.d.ts`? So it is clearer in to future that this is not available for all UI5 versions.

Added with: d89ed6a42aac7bca00f416aa0ec57aa42012c7d7
Please take a look.

devinea (commented):
@mmilko01 Thanks for the update.
Could you also remove skip code I added in the integration test
https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/test/integration/preview/preview.spec.ts#L119-L120

devinea (approved):
Thanks for fixing the integration tests and removing skip for older ui5 versions. 
changeset ✅

packages/preview-middleware-client/src/flp/init.ts
tobiasqueck: It is a constant, so it should actually be `SCENARIO`. Could you please fix it at the source and update all its references.

@voicis or would you disagree?

mmilko01: @tobiasqueck I agree with your proposal, but do you think it is better to this refactoring in a different PR?

nikmace (dismissed):
Change looks good
Tests adjusted
Did not test manually

packages/preview-middleware-client/src/flp/init.ts
tobiasqueck: I am ok with a different one but then do the different one first, so that we don't add that odd usage here.

mmilko01: I think it would be better to fix it here. 
Fixed with: 8447f1c3c9a83860ddf0726fe5124896444d03f4

nikmace (approved):
Re-approving

packages/preview-middleware-client/src/flp/init.ts
tobiasqueck: Looks good to me, now, I would like to see @voicis approving this PR.

voicis (approved):
- Review comments addressed
- Straight forward change to constants
- Did not test locally

Issue #2074, opened 2024-06-25
FEATURE - `@sap-ux/axios-extension` return service type property when calling `listServices`


IainSAP:
### Description

Services may define a property `serviceType` the indicates for what purpose the service is intended. For example,

<img width="1200" alt="image" src="https://github.com/SAP/open-ux-tools/assets/46536134/0095c388-5017-4c39-91bc-003e7c2f098f">

This can be used to indicate to developers if a service is suitable as the service of a UI5 app. In some cases an additional network call may be required to determine this value using Hub services. This can be done as a subsequent call for a specific service and should be added as follows:

Call a different endpoint in the catalog service, i.e ServiceTypeForHUBServices entity with a KEY of service identifier like
/sap/opu/odata/IWFND/CATALOGSERVICE;v=2/ServiceTypeForHUBServices('ZTEST_SYSALIAS_RB_SRV_0001')
which would return the correct service type of the service. The payload will not be as big as previous end points. This is just intended to get the correct service type.

```
export const SERVICE_TYPE = {
    UI: 'UI',
    NOT_CLASSIFIED: 'Not Classified',
    WEB_API: 'WEB_API',
    NOT_DETERMINED: 'Not Determined'
};
```

### Technical Design
As described.

### Acceptance Criteria

Service Type is available in V2 services. An additional function is exported from `catalog` with the signature `getServiceType(service.path)` which can be used to determine the service type if it is not available for v2 services and service type set in the service is of value `Not Determined`

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

IainSAP:
Done

PR #2126, opened 2024-07-05
feat(axios-ext): service type for catalog requests

cianmSAP:
#2074

- Adds `serviceType` to to V2/V4 Catalog service info results
- Adds method `getServiceType` to the V2 catalog class


packages/axios-extension/src/abap/catalog/v2-catalog-service.ts
```diff
@@ -219,4 +221,25 @@ export class V2CatalogService extends CatalogService {
...
         return annotations;
     }
+
+    /**
+     * Calls endpoint `ServiceTypeForHUBServices` to determine the service type.
+     *
+     * @param path service path
+     * @returns service type
+     */
+    public async getServiceType(path: string): Promise<ServiceType> {
```
longieirl, line 231: Should this be returning `Promise<ServiceType | undefined>` as `serviceType` is undefined by default in the method?

cianmSAP: Thanks @longieirl I've updated

IainSAP (approved):
Thanks @cianmSAP 

- New API added as expected
- Tests added to cover
- Have not manually tested (just yet...)

PR #2132, opened 2024-07-08
fix: Add logic to determine resource URLs  for flpSandbox based on project type 

kjose90:

- Improve Resource URL Determination for UShell and UI5 Bootstrap Based on Project Type in `templates/common/add/webapp/test/flpSandbox.html` for writers
- Implemented `getBootstrapResourceUrls` function in `fiori-generator-shared/src/common.ts` to determine resource URLs for UShell bootstrap and UI5 bootstrap.
- Updated snapshots in `fiori-freestyle-writer` and `fiori-elements-writer` to reflect the new URL determination logic - where relative URLs are used for Edmx projects and fully qualified URLs are used for CAP projects


packages/fiori-elements-writer/src/index.ts
```diff
@@ -82,11 +83,19 @@ async function generate<T extends {}>(basePath: string, data: FioriElementsApp<T
     await addOdataService(basePath, feApp.service, fs);
 
     const coercedUI5Version = semVer.coerce(feApp.ui5?.version)!;
+    // Determine if the changes preview should be enabled based on the project type and UI5 version
```
IainSAP, line 86: Might be worth moving the templateOptions logic into a new function to keep the code clean.

kjose90: > Might be worth moving the templateOptions logic into a new function to keep the code clean and have a single condition for project type so its clear which options are for which project type

I moved this logic [here](https://github.com/SAP/open-ux-tools/pull/2132/files#diff-ff4edbeac8e42e9c7ce005e9ba864236b8e5909f154fde419beb25fa1098fa06R156)

longieirl (approved):
Based on the offline conversations around this topic, I've a better understanding of how this impacts CAP and Fiori UI projects.
Changes look good.
Test coverage is good.
This will introduce changes into tools-suite, as expected.

IainSAP (commented):
Small comment

packages/fiori-freestyle-writer/src/index.ts
```diff
@@ -26,16 +27,31 @@ async function generate<T>(basePath: string, data: FreestyleApp<T>, fs?: Editor)
     await initI18n();
     // Clone rather than modifying callers refs
     const ffApp = cloneDeep(data);
+    // Determine if the project type is 'EDMXBackend'.
+    const isEdmxProjectType = ffApp.app.projectType === 'EDMXBackend';
```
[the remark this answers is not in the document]
kjose90: > Usually we would define variable immediately before use.

Yes thanks for that. I moved it closer to the use.

IainSAP (approved):
Thx @kjose90 

- Updated as expected
- Tests added to cover

Issue #2145, opened 2024-07-11
BUG - Incorrect definition of fiori-tools-proxy in ui5.yaml of Adaptation Projects


nikmace:
### Description
fiori-tools-proxy afterMiddleware should have fiori-tools-preview middleware. Otherwise, it does not work properly and produces correct changes with the reference to the app variant. Currently, the changes done by CPE have the reference pointing to the base app ID instead of the app variant name.

### Steps to Reproduce
1. Generate an ADP project with the Adaptation Project writer
2. Open CPE and make propertyChanges for example
3. Open the saved change file and see incorrect reference

### Expected results
Reference points to app variant name

### Actual results
Reference in the change points to base app id

### Screenshots

Correct definition of ui5.yaml proxy middleware should be:
<img width="538" alt="Screenshot 2024-07-11 at 13 59 31" src="https://github.com/SAP/open-ux-tools/assets/39880364/3967a8ac-0e11-4215-8c13-1cdd84fe5a78">

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

tobiasqueck:
> Otherwise, it does not work properly and produces correct changes with the reference to the app variant. Currently, the changes done by CPE have the reference pointing to the base app ID instead of the app variant name

I'd like to understand how the order of the middlewares affect that value.

PR #2146, opened 2024-07-11
fix: generate correct ui5.yaml for adp

GDamyanov:
Fix for #2145 

- Configure custom middlewares in ui5.yaml in correct way

### DIFF
--- a/packages/adp-tooling/src/writer/options.ts
+++ b/packages/adp-tooling/src/writer/options.ts
@@ -105,14 +105,16 @@ function addFioriToolsMiddlwares(ui5Config: UI5Config, config: AdpWriterConfig) 
             }
         }
     ]);
-    ui5Config.addFioriToolsProxydMiddleware({
-        ui5: {
-            url: config?.ui5?.frameworkUrl,
-            version: config?.ui5?.minVersion ?? '' //default to latest if version is not set
+    ui5Config.addFioriToolsProxydMiddleware(
+        {
+            ui5: {
+                url: config?.ui5?.frameworkUrl,
+                version: config?.ui5?.minVersion ?? '' //default to latest if version is not set
+            },
+            backend: [backendConfig as FioriToolsProxyConfigBackend]
         },
-        backend: [backendConfig as FioriToolsProxyConfigBackend],
-        afterMiddleware: 'fiori-tools-preview'
-    });
+        'fiori-tools-preview'
+    );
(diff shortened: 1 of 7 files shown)


packages/ui5-config/src/types/middlewares.ts
```diff
@@ -38,6 +38,7 @@ export interface FioriToolsProxyConfig {
     backend?: FioriToolsProxyConfigBackend[];
     ui5?: Partial<FioriToolsProxyConfigUI5>;
     ignoreCertError?: boolean;
+    afterMiddleware?: string;
```
tobiasqueck, line 41: Please remove this here again. The `afterMiddleware` property is not part of the proxy config, it is a separate property for every middleware

packages/adp-tooling/src/writer/options.ts
```diff
@@ -110,7 +110,8 @@ function addFioriToolsMiddlwares(ui5Config: UI5Config, config: AdpWriterConfig)
             url: config?.ui5?.frameworkUrl,
             version: config?.ui5?.minVersion ?? '' //default to latest if version is not set
         },
-        backend: [backendConfig as FioriToolsProxyConfigBackend]
+        backend: [backendConfig as FioriToolsProxyConfigBackend],
+        afterMiddleware: 'fiori-tools-preview'
```
tobiasqueck, line 114: I don't object that change but can you explain why this is a problem? I am afraid we are not understanding the root cause here.

tobiasqueck (changes requested):

packages/ui5-config/src/types/middlewares.ts
GDamyanov: Fixed in [9310a2a](https://github.com/SAP/open-ux-tools/pull/2146/commits/9310a2aec0854ff120effa183bb33a63b15e221f)

tobiasqueck (approved):
* changeset ok
* issue correctly fixed (tested it locally)
* test snapshots correctly updated
* readme update explaining the requirement

packages/adp-tooling/src/writer/options.ts
nikmace: Readme updated in: [619ae1e](https://github.com/SAP/open-ux-tools/pull/2146/commits/619ae1e4d64848291997270a4134034d24b97707)

PR #2147, opened 2024-07-11
Refactor: Move sap-ux Logic Based on projectType to Feature Implementation

kjose90:

Moved adding of `sap-ux` logic based on `projectType` to feature implementation instead of containing the logic in optional feature logic under `ui5-application-writer`
Addressing [comment](https://github.com/SAP/open-ux-tools/pull/2102#discussion_r1673596529)

### DIFF
--- a/packages/ui5-application-writer/src/options.ts
+++ b/packages/ui5-application-writer/src/options.ts
@@ -6,12 +6,13 @@ import { getFilePaths } from '@sap-ux/project-access';
 import type { UI5Config } from '@sap-ux/ui5-config';
 import { ui5NPMSupport, ui5TSSupport } from './data/ui5Libs';
 import { mergeObjects, UI5_DEFAULT } from '@sap-ux/ui5-config';
+import type { ProjectType } from '@sap-ux/project-access';
 
 /**
  * Input required to enable optional features.
  */
 export interface FeatureInput {
-    ui5App: { app: { id: string; baseComponent?: string } };
+    ui5App: { app: { id: string; baseComponent?: string; projectType?: ProjectType } };
     fs: Editor;
     basePath: string;
     tmplPath: string;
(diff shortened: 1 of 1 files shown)


IainSAP (approved):
Thx @kjose90 

- Small update, already covered by tests

packages/ui5-application-writer/src/options.ts
```diff
@@ -56,7 +57,11 @@ const factories: { [key: string]: (input: FeatureInput) => Promise<void> } = {
     codeAssist: async (input: FeatureInput) => await copyTemplates('codeAssist', input),
     eslint: async (input: FeatureInput) => await copyTemplates('eslint', input),
     loadReuseLibs: async (input: FeatureInput) => await copyTemplates('loadReuseLibs', input),
-    sapux: async (input: FeatureInput) => await copyTemplates('sapux', input),
+    sapux: async (input: FeatureInput) => {
+        if (input.ui5App.app.projectType === 'EDMXBackend') {
```
longieirl, line 61: Do you need to apply some optional chaining check here for undefined? Its a deeply nested object, will these values always exist?

kjose90: > Do you need to apply some optional chaining check here for undefined? Its a deeply nested object, will these values always exist?

`ui5App` and `app` are mandatory properties of [FeatureInput](https://github.com/SAP/open-ux-tools/pull/2147/files#diff-8651a8bccdacbc50e0fc8b935e886f22d73e6ad4a7b3f02da31c28b740bafa7dR15). `ProjectType` within `app` is optional.  If `projectType` is undefined, the condition `input.ui5App.app.projectType === 'EDMXBackend'` will safely evaluate to false know? so do we need optional chaining here? Also `projectType` is mandatory for `Ui5App`, so `projectType` will always be provided at this point. What do you think ?

longieirl (approved):
Changes look good.
Test coverage is good.

kjose90:
Thanks for the review.

PR #2148, opened 2024-07-11
Performance/project access/cache global cds module promise

ManuelPastorini:
To load global installed cds module, we use shell command to identify the path to load from. This is quite expensive, and gets even worse if called multiple times in parallel. So far, only after retrieving the path, we cached it to avoid further requests - now, we already cache the promise, thus we can avoid sending the same shell command again, while the first request is still running.


Klaus-Keller (changes requested):
Hi @ManuelPastorini,
please add a changeset by running 
```
pnpm changeset
```
in root of this repository. I propose a patch update.

ManuelPastorini:
Hi Klaus,

I added a change-set now. I hope, I did it correctly ;)

packages/project-access/src/project/cap.ts
```diff
@@ -557,29 +557,38 @@ async function readPackageNameForFolder(baseUri: string, relativeUri: string): P
...
+                    resolve(loadModuleFromProject<CdsFacade>(versions.home, '@sap/cds'));
+                } else {
+                    reject(
+                        new Error(
+                            'Can not find global installation of module @sap/cds, which should be part of @sap/cds-dk'
+                        )
+                    );
+                }
+            }, reject);
+        });
```
Klaus-Keller, line 583: Would the async/await notation not be more readable, what do you think?

```suggestion
            const versions = await getCdsVersionInfo();
            if (versions.home) {
                resolve(loadModuleFromProject<CdsFacade>(versions.home, '@sap/cds'));
            } else {
                reject(
                    new Error(
                        'Can not find global installation of module @sap/cds, which should be part of @sap/cds-dk'
                    )
                );
            }
```

Klaus-Keller (approved):
Thanks @ManuelPastorini,
- code changes to prepare for parallel processing of `cds --version` look good
- changeset exists
- test coverage is great
- did a local test, could not found issues

Approved from my side,

johannes-kolbe (approved):
- Code looks good, 
- Cache approach sounds good
- Did not test manually

PR #2149, opened 2024-07-11
fix: output specification version in cache

johannes-kolbe:
# Issue

#2066

- Outputs the specification `version` instead of `dist-tag` when loading from cache

### DIFF
--- a/packages/project-access/src/project/specification.ts
+++ b/packages/project-access/src/project/specification.ts
@@ -59,10 +59,10 @@ async function hasSpecificationDevDependency(root: string): Promise<boolean> {
 async function getSpecificationModule<T>(root: string, options?: { logger?: Logger }): Promise<T> {
     const logger = options?.logger;
     let specification: T;
-    const distTag = await getProjectDistTag(root, { logger });
+    const version = await getSpecificationVersion(root, { logger });
     try {
-        specification = await getSpecificationByDistTag<T>(distTag, { logger });
-        logger?.debug(`Specification loaded from cache using dist-tag '${distTag}'`);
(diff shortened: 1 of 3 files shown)


packages/project-access/src/project/specification.ts
```diff
@@ -187,15 +186,15 @@ async function getSpecificationVersion(root: string, options?: { logger?: Logger
  */
 export async function getSpecificationPath(root: string, options?: { logger?: Logger }): Promise<string> {
     const logger = options?.logger;
-    const moduleName = '@sap/ux-specification';
+    const moduleName = join('@sap', 'ux-specification');
```
Klaus-Keller, line 189: Not sure about this, the module name is `@sap/ux-specification` no matter which operating system. The fact we use this to store it under path `join(scope, name)` is purely internal. The function `getModulePath()` expects the name.
To avoid confusion I'd use `join()` later in this function when we do `slice`

johannes-kolbe: @Klaus-Keller done with https://github.com/SAP/open-ux-tools/pull/2149/commits/0ce1cf3a0a54e478d3625d9d852889d082ef68fe, would that work?

Klaus-Keller (approved):
Thanks @johannes-kolbe,
- fix for windows and logging of version instead of dist-tag looks good
- changeset exists
- test coverage is great
- did a local test

Approved from my side

815are (approved):
checked:
1. code changes easy to understand - windows fix with `join` and `logger.debug` update
2. tests adjusted

did not test manually

Issue #2151, opened 2024-07-12
BUG - Remove safeMode from `adp-tooling` types and change order of ui5.yaml custom configuration


nikmace:
### Description
Remove safeMode from `adp-tooling` types `AdpProjectData` interface because it is no longer needed.

Change the order of the ui5.yaml custom configuration to have a clearer structure of the config.

### Steps to Reproduce
1. Generate a project with an ADP writer from adp-tooling
2. Open ui5.yaml and see customConfiguration at the bottom

### Expected results
SafeMode is not required in `AdpProjectData` interface.
Config structure is adjusted to have customConfiguration at the top

### Actual results
SafeMode is required
CustomConfiguration is at the bottom of the ui5.yaml

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2152, opened 2024-07-12
fix: Remove safeMode from `adp-tooling` types and change order of ui5.yaml custom configuration

nikmace:
Fix for #2151.

- Removes safeMode from `AdpProjectData` interface because it is no longer needed.
- Changes the order of the blocks in ui5.yaml where customConfiguration is at the top for clearer structure.


tobiasqueck (approved):
* changeset ok
* changes do what they are supposed to do

GDamyanov (approved):
- Code looks good.
- Excellent test coverage.
- Tested manually.

Issue #2155, opened 2024-07-12
TBI - Removal of private: true Flag from Package.json Templates


kjose90:
Remove the `private: true` flag from the package.json templates for EDMX and CAP applications from: 

- `ui5-application-writer` template package.
- `ui5-library-writer template` template package.
- `ui5-library-writer template` template for  typescript package.

kjose90:
Closing this this issue. Done.

PR #2156, opened 2024-07-12
fix: Remove private true flag from  package json templates

kjose90:
#2155

Removed `private: true flag` from the package.json templates for EDMX and CAP applications from:

- `ui5-application-writer` package json template.
- `ui5-library-writer `template package json template.
- `ui5-library-writer` template typescript package json template.

Updated snapshots  under `fiori-elements-writer`, `fiori-freestyle-writer`, `ui5-application-writer` and `ui5-library-writer`


devinea (approved):
Removal of the private flag makes sense as we should not assume what people want to do with these modules. 
Snapshot updates.
patchset ✅

.changeset/famous-tigers-cough.md
```diff
@@ -0,0 +1,8 @@
+---
+'@sap-ux/fiori-freestyle-writer': minor
+'@sap-ux/ui5-application-writer': minor
+'@sap-ux/fiori-elements-writer': minor
+'@sap-ux/ui5-library-writer': minor
```
devinea, line 5:
```suggestion
'@sap-ux/ui5-application-writer': minor
'@sap-ux/ui5-library-writer': minor
```
I think we can reduce the list here as FE and FF writers are just snapshot updates.

kjose90: > I think we can reduce the list here as FE and FF writers are just snapshot updates.

Yes correct - Thanks

devinea (approved):
Thanks for the updates. Re-approving.

PR #2158, opened 2024-07-12
fix(project-access): export findCapProjectRoot

longieirl:

- required by new modules but doing this early so we can do some local testing


longieirl:
@Klaus-Keller please review.

devinea (approved):
Export makes sense. 
Changset ✅

longieirl:
Thanks @devinea this has already been previously agreed with Klaus.

kjose90 (approved):
Looks good.
Changeset ✅

Issue #1679, opened 2024-02-21
TBI - Prepare preview middleware for UI5 2.0


tobiasqueck:
### Description (include screenshots)
Internal testing of the preview middleware with a preliminary version of UI5 2.0 has shown issues in the code. This ticket is to track (and solve) these issues.

### Value
Getting the preview middleware ready for UI5 2.0.

### Architecture Elaboration
All fixes are expected in the `preview-middleware-client` module

### Tasks
- [ ] `sap.ui.getCore().getConfiguration().getLanguage()` needs to be replaced, with `sap.ui.core.Configuration.getLanguage`
- [ ] `sap.ushell.Container` needs to be replaced with `sap.ui.require(“sap/ushell/Container”)`
- [ ] `sap.ushell.Container.createRenderer()` is also deprecated - however there is no production ready solution yet
- [ ] `sap.base.util.UriParameters` needs to be replaced with `sap/base/util/UriParameters`

codeworrior:
Small corrections to avoid misunderstandings:
* [`sap.ui.getCore().getConfiguration().getLanguage()`](https://ui5.sap.com/#/api/sap.ui.core.Configuration%23methods/sap.ui.core.Configuration.getLanguage) needs to be replaced with [`import("sap/base/i18n/Localization").getLanguage`](https://ui5.sap.com/#/api/module:sap/base/i18n/Localization%23methods/sap/base/i18n/Localization.getLanguage)
* [sap/base/util/UriParameters.fromQuery(...)](https://ui5.sap.com/#/api/module:sap/base/util/UriParameters%23methods/sap/base/util/UriParameters.fromQuery) needs to be replaced with the native [new URLSearchParams(...)](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)

zdravko-georgiev:
For FLP Sandbox and UI5 2.0 we've discussed the following options:

1. We explicitly expect the user to tell us
    - We add a new config property to the preview-middleware of type boolean allowing to switch to the new sandbox
    - We add a url parameter with the same effect
    - We add a env variable with the same effect
1. We add a handler for both urls to the preview-middleware and if either of them returns a 404, we just redirect to the other. Tricky part here: we don’t know if the preview middleware is executed before or after the ui5-proxy middleware (or any other that might be serving the sources)
1. We create our own bootstrap script that checks the UI5 version and then depending on the result pick one or the other

The decision is to go for option 3.

PR #2159, opened 2024-07-12
feat: preview 2.0

schreckstefan:
Issue: https://github.com/SAP/open-ux-tools/issues/1679

Allows two options for the target sandbox, depending on the UI5 major version the one or other gets chosen.
This `bootstrapPath` gets defaulted now from the manifest.json., at the templating point in time.
In cases where this does not fit at the client side, e.g. in case of run configs having a different version, the proxy error  gets caught and we try to load the different resource.

Besides, the sandbox2.js obviously does not have a function `createRenderer`, but only `createRendererInternal`.

For testing (with https://github.tools.sap/ux/open-ux-tools-playground), I had to set the right target url https://sapui5untested2.int.sap.eu2.hana.ondemand.com
a) at the yaml file of the test app:

>  - name: preview-middleware
>       afterMiddleware: compression
>       configuration:
>         ui5:
>           - path: /resources
>             url: https://sapui5untested2.int.sap.eu2.hana.ondemand.com
>           - path: /test-resources
>             url: https://sapui5untested2.int.sap.eu2.hana.ondemand.com

b) ui5-proxy-middleware did not react on similar settings, I had to set the corresponding env variable for it:
> export FIORI_TOOLS_UI5_URI=https://sapui5untested2.int.sap.eu2.hana.ondemand.com

The PR also replaces sap.ui.version, which gets invalid with UI5 2.0, by `await VersionInfo.load()`


packages/ui5-proxy-middleware/src/base/utils.ts
```diff
@@ -15,11 +15,16 @@ import { t } from '../i18n';
...
  *
  * @param proxyRes - proxy response object
+ * @param req
  * @param etag - ETag for the cached sources, normally the UI5 version
  */
-export const proxyResponseHandler = (proxyRes: IncomingMessage, etag: string): void => {
+export const proxyResponseHandler = (proxyRes: IncomingMessage, req: Request, etag: string): void => {
     proxyRes.headers['Etag'] = etag;
     proxyRes.headers['cache-control'] = 'no-cache';
+    if (proxyRes.statusCode === 404 && req.path.includes('/resources/sap/ushell/bootstrap/sandbox2.js')) {
```
heimwege, line 24: Not sure if this is relevant here but please ensure that this also works when preview middleware is used with karma (uses connect API instead of express). See [karma-ui5 documentation](https://github.com/SAP/karma-ui5?tab=readme-ov-file#notes-on-ui5-tooling-custom-server-middleware) (and #1968 and #1826)

schreckstefan: @heimwege 
Hello Dominik, it does not look like this coding enhancement contradicts to the statements and PR's concerning karma, but if you can test it it would be helpful. Or let me do about the right test environment that I can use.

heimwege: Sure, I'll drop you a mail as the repo is internal

packages/preview-middleware-client/src/flp/initConnectors.ts
```diff
@@ -6,14 +8,16 @@
  * This setup allows for flexibility in using different connectors based on the UI5 version.
  *
  * @example
- * intiConnectors(); // Simply call the function without any arguments.
+ * initConnectors(); // Simply call the function without any arguments.
  * @returns {void}
```
heimwege, line 12: According to the changes JSDoc needs to be adjusted to returning `Promise<void>`

heimwege: The `@returns` was part of the JSDoc because it describes the return value of the `initConnectors` function as such. So it's not part of the `@example` that's correct but it still should be part of the JSDoc of `initConnectors` because it describes the return value of the function (see [JSDOC example](https://jsdoc.app/tags-example)). The only thing that was wrong was the type. Although `initConnectors` has been adjusted in this branch to return a promise it was still `void` according to JSDoc.
At least that how it is done e.g. in the [init.ts](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/src/flp/init.ts#L154) of the middleware client. So for consistency reasons I would stick to it.

Technically the TypeScript type checker can [infer](https://www.typescriptlang.org/docs/handbook/2/functions.html#inference) the return type anyway. So `@returns` as well as `: Promise<void>` could be omitted.

schreckstefan:
Work in progress: dynamic loading of sap/ushell/bootstrap/sandbox now works, but the ushell container does not get initialized, an then runs into an exception
image

PR #2160, opened 2024-07-12
Performace/project access/parallel filter applications

ManuelPastorini:
Execute filtering of applications according to manifest parallel instead of sequentiell.

This PR should only be merged after https://github.com/SAP/open-ux-tools/pull/2148. There's no technical dependency and no functional problem, but merging this one would lead to worse performance because of the issue solved with that one. 

Similar change would make sense for all places with code like 
`for (const entry of list){
   ...
  await doSomething(entry)
  ...
}
` - but always needs to come along with corresponding measurement. (If measurement results in worse performance for parallel execution, this does not mean not to parallelize, but points to an issue somewhere in `doSomething(...)` that needs to be analyzed.)

Esp. here the methods `filterAdaptations`, `filterExtensions`, and `filterLibraries` are build similar and should be reworked.


Klaus-Keller (approved):
Thanks @ManuelPastorini, 
- Code to filter for applications in parallel looks good
- Change set exists
- Test coverage is great
- Did a local test, could not find issues

Approved from my side.

AlinaGovoruhina (approved):
Change to parallel filter of apps by manifest looks good from my side.

Issue #2165, opened 2024-07-16
TBI - (ui5-info): add ability to mark the latest maintained UI5 version as default


mikicvi-SAP:
### Description (include screenshots)
- As seen recently in [UI5 versions](https://ui5.sap.com/version.json?sap-ui-config-patches=true&sap-ui-config-showall=true) if the latest version (greatest number) is marked as `Out of maintenance`, it is picked as the default version isui5-application-inquirer.
- Internal ticket number: 30529

### Value
- Additional guard against external errors coming from published ui5 versions

### Architecture Elaboration
Does it requires architecture elaboration?

mikicvi-SAP:
Closing as implemented/merged.

PR #2166, opened 2024-07-16
tbi(ui5-info): includeMaintained and includeDefault combination finds the next maintained version

mikicvi-SAP:
Ticket: #2165


packages/ui5-info/src/types.ts
```diff
@@ -39,6 +39,10 @@ export interface UI5VersionFilterOptions {
      * Includes only versions filtered by latest patch.
      */
     onlyLatestPatchVersion?: boolean;
+    /**
+     * Marks the latest maintained version as default
```
IainSAP, line 43:
```suggestion
     * Marks the latest maintained version as default in the case where the default is otherwise unmaintained
```

packages/ui5-info/src/ui5-version-info.ts
```diff
@@ -357,7 +357,14 @@ export async function getUI5Versions(filterOptions?: UI5VersionFilterOptions): P
...
         };
-        if (filterOptions?.includeDefault && defaultUI5Version === ui5) {
+        if (
+            filterOptions?.includeDefault &&
+            defaultUI5Version === ui5 &&
+            !filterOptions.markLatestMaintainedAsDefault
+        ) {
+            ui5Version.default = true;
+        }
+        if (filterOptions?.markLatestMaintainedAsDefault && latestUI5Version === ui5) {
```
IainSAP, line 367: `default` should only ever be included where `includeDefault` is specified.

mikicvi-SAP: Addressed in [8346988](https://github.com/SAP/open-ux-tools/pull/2166/commits/83469882db35845d335cfb76671fedaa209b415a), thank you @IainSAP.

packages/ui5-info/test/testdata/official-response-latest-out-of-maintenance.json
```diff
@@ -0,0 +1,31 @@
+{
+    "latest": {
+        "version": "1.125.1",
+        "support": "Maintenance",
```
IainSAP, line 4: Shouldnt this be `Out of Maintenance` i.e. 1.126 to test the specific case?

packages/ui5-info/src/ui5-version-info.ts
```diff
@@ -343,33 +343,30 @@ export async function getUI5Versions(filterOptions?: UI5VersionFilterOptions): P
...
+        : undefined;
 
-    let ui5VersionsOverview: UI5VersionOverview[] | undefined;
-    if (filterOptions?.includeMaintained === true) {
-        ui5VersionsOverview = (await retrieveUI5VersionsCache(
-            ui5VersionsType.overview,
-            filterOptions?.useCache
-        )) as UI5VersionOverview[];
-    }
+    const isMaintained = (ui5: string) =>
```
IainSAP, line 352: Can refactor to a new function please?

packages/ui5-info/src/ui5-version-info.ts
```diff
@@ -343,33 +343,30 @@ export async function getUI5Versions(filterOptions?: UI5VersionFilterOptions): P
         );
         filteredUI5Versions = defaultUi5Versions.slice();
     }
+
     const defaultUI5Version = filteredUI5Versions[0];
+    const ui5VersionsOverview = filterOptions?.includeMaintained
```
IainSAP, line 348: Not sure we need a ternary operator here?

packages/ui5-info/src/ui5-version-info.ts
```diff
@@ -343,33 +343,30 @@ export async function getUI5Versions(filterOptions?: UI5VersionFilterOptions): P
...
-            ui5Version.maintained = ui5VersionsOverview.some((v): boolean | undefined =>
-                v !== undefined
-                    ? `${major(v.version)}.${minor(v.version)}` === `${major(ui5)}.${minor(ui5)}` &&
-                      v.support === 'Maintenance'
-                    : undefined
-            );
-        }
-        return ui5Version;
-    });
+    const finalDefaultUI5Version =
```
IainSAP, line 360: Could we not use a ternary, makes the code hard to read in this instance?

packages/ui5-info/test/testdata/official-response-latest-out-of-maintenance.json
mikicvi-SAP: Nope, I'm testing that 1.126 is the higher version but out of maintenance, i.e replicating incident from last week.

IainSAP: OK but what will happen if "latest" is `out of maintenance` ?

devinea: That shouldn't happen normally right?

packages/ui5-info/src/ui5-version-info.ts
mikicvi-SAP: Addressed, with some additional comments, thanks!

packages/ui5-info/src/ui5-version-info.ts
mikicvi-SAP: Addressed with additional type and comment added fro clarity. Thank you for suggestions.

packages/ui5-info/test/testdata/official-response-latest-out-of-maintenance.json
mikicvi-SAP: Test added in [03a5023](https://github.com/SAP/open-ux-tools/pull/2166/commits/03a5023ee5d196a664b9f599e5c49ae8a9dc8314).

IainSAP (approved):

Klaus-Keller (commented):
Hi @mikicvi-SAP, 

Just yesterday we got a request to NOT use 
https://github.com/SAP/open-ux-tools/blob/08957786a0a1a18eafcd49c157cce1c0e849ca13/packages/ui5-info/src/constants.ts#L23
anymore, all information should be provided through
https://github.com/SAP/open-ux-tools/blob/08957786a0a1a18eafcd49c157cce1c0e849ca13/packages/ui5-info/src/constants.ts#L22

I had a brief look at the code, seems we use only property `support` from 
https://ui5.sap.com/versionoverview.json, which is also included in 
https://ui5.sap.com/version.json?sap-ui-config-patches=true&sap-ui-config-showall=true

Can we move away from `versionoverview.json`?

ianquigley-sap:
@Klaus-Keller We would rather to the update you suggest in a separate PR.   From discussions with the AppGen team, this might not be straightforward to not use `versionoverview.json`.

You ok if we keep this for the original issue and have a follow up to replace 'versionoverview.json'?

Klaus-Keller:
> @Klaus-Keller We would rather to the update you suggest in a separate PR. From discussions with the AppGen team, this might not be straightforward to not use `versionoverview.json`.
> 
> You ok if we keep this for the original issue and have a follow up to replace 'versionoverview.json'?

Can we check whether it is straight forward or not? I did a brief check and could not find any other property than `support` to be used from `versionoverview.json`.

mikicvi-SAP:
> > @Klaus-Keller We would rather to the update you suggest in a separate PR. From discussions with the AppGen team, this might not be straightforward to not use `versionoverview.json`.
> > You ok if we keep this for the original issue and have a follow up to replace 'versionoverview.json'?
> 
> Can we check whether it is straight forward or not? I did a brief check and could not find any other property than `support` to be used from `versionoverview.json`.

I spent some time with this on Friday, and this is no trivial task by any means. It would involve completely changing how parser works and potential caching changes too.
I believe that there is very high chance of introducing a regression, and this suggestions should be explored further for potential issues that could stem from it.

Klaus-Keller:
@IainSAP @mikicvi-SAP, fine to remove `versionoverview.json` separate, I've created https://github.com/SAP/open-ux-tools/issues/2186 for this. Would be great if it could be accomplished right after this PR has been merged.

Klaus-Keller (approved):
My concerns with usage of `versionoverview.json` are addressed seperately.
- code changes look good
- changeset exists
- test coverage is good
- did not do a manual test, just visual review

Approved from my side.

Issue #2162, opened 2024-07-15
FEATURE - Move prompting code of Adaptation Project generator


nikmace:
_As a **developer**,
I want to **create an Adaptation Project**,
so that I **can adapt base applications**_

### Description

_Prompting code of Adaptation Project generator should be moved to OSR so only the skeleton of the generator is left and the logic is split in prompting and generate methods._

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I consume `adp-tooling` package
**when** I can use the prompting for ADP generator
**then** to create a new ADP locally.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

nikmace:
The issue is absolete, this is a new issue tracking the progress of the generator code migration

PR #2169, opened 2024-07-16
feat: Add prompting code of Adaptation Project generator

nikmace:
Feat for #2162.

- Moves prompting for Adaptation Project generator.
- Enhances CLI prompting.


IainSAP:
Package.json versions should only be changed by the github action (automated release process) rather than manually as appears to have been done in this PR.

packages/adp-tooling/src/base/node-generator-resolver.ts
```diff
@@ -0,0 +1,28 @@
+import { resolve } from 'path';
+
+/**
+ * Attempts to resolve the path to a specific node module generator from the NODE_PATH environment variable.
+ *
+ * @returns {string | undefined} The resolved path to the generator module if found, or undefined if not found.
+ */
+export function resolveNodeModuleGenerator(): string | undefined {
```
tobiasqueck, line 8: What is this function doing and where do we use it?

packages/adp-tooling/src/client/abap-provider/index.ts
```diff
@@ -0,0 +1,124 @@
...
+        }
+        return this.provider;
+    }
+
+    /**
+     * Retrieves wheter the configured ABAP service provider is connected to ABAP system.
+     *
+     * @returns {boolean} - if provider is connected to ABAP system.
+     */
+    public getIsConnected(): boolean {
```
tobiasqueck, line 47: why this wrapper? wouldn't be a getter/setter a nicer option if it is just about preventing it to be changed from the outside? If you really need this function then please rename it to `isConnected` and the variable to `connected`.

packages/adp-tooling/src/common/app-type.ts
```diff
@@ -0,0 +1,34 @@
...
+export function getApplicationType(manifest: Manifest): ApplicationType {
+    if (Object.keys(manifest).length > 0) {
+        const appInfo = manifest['sap.app'];
+        const isSmartTemplate = !!manifest['sap.ui.generic.app'];
+        const hasSmartTemplateId = appInfo?.sourceTemplate?.id?.toLowerCase() === 'ui5template.smarttemplate';
+
+        if (manifest['sap.ovp']) {
+            return ApplicationType.FIORI_ELEMENTS_OVP;
+        } else if (hasSmartTemplateId || isSmartTemplate) {
+            return ApplicationType.FIORI_ELEMENTS;
```
tobiasqueck, line 27: You are not catching Fiori elements for OData v4 here.

packages/adp-tooling/src/common/app-type.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @param {Manifest} manifest - The manifest configuration object of the application.
+ * @returns {ApplicationType} The type of the application as defined by the ApplicationType enum.
+ *
+ * The classification is done based on the presence and values of specific keys in the manifest:
+ * - 'sap.ovp' indicates a Fiori Elements Overview Page (OVP).
+ * - 'sap.ui.generic.app' or 'sap.app' with a specific sourceTemplate id indicates a Fiori Elements application.
+ * - If none of these conditions are met, the function defaults to categorizing the application as Free Style.
+ * - If the manifest is empty, it returns None.
+ */
+export function getApplicationType(manifest: Manifest): ApplicationType {
```
tobiasqueck, line 18: Why do we actually need the application type? I thought that was something we removed from the templates.

packages/adp-tooling/src/common/endpoints-manager.ts
```diff
@@ -0,0 +1,181 @@
...
+    private isExtensionInstalled: boolean;
+
+    /**
+     * Creates an instance of EndpointsManager.
+     *
+     * @param {ToolsLogger} logger - The logger.
+     */
+    private constructor(private logger: ToolsLogger) {
+        this.endpoints = [];
+        this.isExtensionInstalled = isExtensionInstalledVsCode('sapse.sap-ux-application-modeler-extension');
```
tobiasqueck, line 69: the extension sapse.sap-ux-application-modeler-extension is not a pre-requisite here. The extension allows modifying the stored systems, however, the systems itself are stored in the OS secure storage and can be added/manipulated with other tools as well, therefore, please remove that check.

packages/adp-tooling/src/base/node-generator-resolver.ts
nikmace: This is used in the prompt for extension project when the application is not supported by adaptation project. This is used in BAS to resolve the generator. It was the only option that worked with this.composeWith() of the yeoman generator. We get the path back to the generator (if found) from resolveNodeModuleGenerator. If no path is found we show a message that Extensibility Generator is not installed in the dev space and the user cannot continue.

tobiasqueck: Could you enhance the jsdoc comment please was something similar to what you just wrote here.

packages/adp-tooling/src/common/app-type.ts
nikmace: We check the type because we need to see if the Adaptation Project supports it. These are supported:
FIORI_ELEMENTS = 'FioriElements',
FIORI_ELEMENTS_OVP = 'FioriElementsOVP',
FREE_STYLE = 'FreeStyle'
If it is something else we return an error stating that: `... Adaptation project doesn't support the selected application.`

packages/adp-tooling/src/common/endpoints-manager.ts
nikmace: Then we need to check whether OS secure storage exists? Otherwise, if it doesn't, then how can we understand what prompts to show. If it exists then we need a list prompt with all the systems. If it doesn't exist then we show input prompts for system, client, username and password.

packages/adp-tooling/src/base/node-generator-resolver.ts
nikmace: Changed in: [6d395b6](https://github.com/SAP/open-ux-tools/pull/2169/commits/6d395b63c1e16105d873c3bce51cc8646da5c0f1)

packages/adp-tooling/src/common/app-type.ts
tobiasqueck: You return one of these three if `Object.keys(manifest).length > 0`, so technically, you validations just checks if the manifest is neither `undefined` nor `{}`.

packages/adp-tooling/src/common/endpoints-manager.ts
tobiasqueck: Just call the store module and ask it for a list of systems. If at least one is returned you are good to go would be an option. The better option is that to reuse the functionality that is currently being implemented in the `odata-service-inquirer` because that one always offers a `new system` choice that then triggers the system creation prompts.

packages/adp-tooling/src/client/abap-provider/index.ts
GDamyanov: Fixed in [d4795dc](https://github.com/SAP/open-ux-tools/pull/2169/commits/d4795dc11696b6184a7ec64bb12286b9bc077f06)

packages/adp-tooling/src/common/endpoints-manager.ts
GDamyanov: Fixed in [ad544c4](https://github.com/SAP/open-ux-tools/pull/2169/commits/ad544c432e18945f92cb07099b155a4474eec626)

packages/adp-tooling/src/common/app-type.ts
nikmace: We removed this check for a valid application type here https://github.com/SAP/open-ux-tools/pull/2169/commits/3bb3bc6a249590729b79ba69812d876dd8bcb47d#diff-034445208bf7ca40b734423c5cf343d6d0a4761a4b55cee844776b5458a1c9a0L86-L102.

But since this is used in the writing part we did not remove the method. It is used in a method that builds the descriptor content: `createDescriptorChangeForResourceModel`.

packages/adp-tooling/src/common/app-type.ts
nikmace: We don't need to differentiate between fiori elements v2 or v4, we don't have such logic which relies on that.

Issue #2170, opened 2024-07-17
BUG - `module-loader` does not install specification dependency in `.fioritools` folder when node_modules exists on `C:\Users\{User}` level


815are:
### Steps to Reproduce
1. install `@sap/ux-specification` in `C:\Users\{USER}` folder(replace with your user or adjust to OS) by running
```
npm install @sap/ux-specification
```
2. delete folder `C:\Users\{USER}\.fioritools\module-cache`(replace with your user or adjust to OS) by running
3. simulate `module-loader` to install `@sap/ux-specification` in  `C:\Users\{USER}\.fioritools` folder

Problem `npm install @sap/ux-specification` does not install files in `C:\Users\{USER}\.fioritools\module-cache\@sap\ux-specification\{version}`, but it installs dependencies  `C:\Users\{USER}\node_modules` 

### Expected results
specification dependencies installed in `C:\Users\{USER}\.fioritools\module-cache\@sap\ux-specification\{version}`

### Actual results
specification dependencies installed in `C:\Users\{USER}\node_modules` 

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2171, opened 2024-07-17
fix: module-loader does not install specification dependency in `.fioritools` folder when node_modules exists on C:\Users\{User} level

815are:
Issues happens in Windows and MACOS

Scenario from #2170
1. install `@sap/ux-specification` in `C:\Users\{USER}` folder(replace with your user or adjust to OS) by running
```
npm install @sap/ux-specification
```
2. delete folder `C:\Users\{USER}\.fioritools\module-cache`(replace with your user or adjust to OS) by running
3. simulate `module-loader` to install `@sap/ux-specification` in  `C:\Users\{USER}\.fioritools` folder

In 3rd step code runs:
```
npm install @sap/ux-specification@1.120.14
```
and with such setup - it does not install dependencies in `C:\Users\{USER}\.fioritools\module-cache\@sap\ux-specification\{version}`

Proposal fixes:
1. use --prefix during install -  - https://stackoverflow.com/questions/14469515/how-to-npm-install-to-a-specified-directory, like:
```
npm install --prefix C:\Users\{USER}\.fioritools\module-cache\@sap\ux-specification\1.120.14 @sap/ux-specification@1.120.14
```
2. create empty package.json if module does not exist

In this PR I am using 2nd approach as it seems most simplest to go

### DIFF
--- a/.changeset/good-waves-boil.md
+++ b/.changeset/good-waves-boil.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': patch
+---
+
+fix: module-loader fails to install specification dependecy in '.fioritools' folder when user has 'node_modules' in user home folder
--- a/packages/project-access/src/project/module-loader.ts
+++ b/packages/project-access/src/project/module-loader.ts
@@ -1,5 +1,5 @@
 import { existsSync } from 'fs';
-import { mkdir, rm } from 'fs/promises';
+import { mkdir, rm, writeFile } from 'fs/promises';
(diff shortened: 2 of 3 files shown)


Klaus-Keller (approved):
Thanks @815are,
- fix to ensure correct node_modules dir is used looks good
- changeset exists
- code coverage is great

Approved from my side.

broksy (approved):
- Understandable fix and changeset
- Good test coverage

Approved from my side.

PR #2183, opened 2024-07-19
Fix(preview middleware) preview client url

heimwege:
Using an absolute preview client url does not work in combination with [cds-plugin-ui5](https://www.npmjs.com/package/cds-plugin-ui5).

Sample:
- URL without cds-plugin-ui5: http://localhost:8080/localService/one/more/path/index.html#Shell-home (core resource url http://localhost:8080/resources/sap-ui-core.js)
- URL with cds-plugin-ui5: http://localhost:4004/com.sap.cap.fe.ts.sample/localService/one/more/path/index.html#Shell-home (core resource url http://localhost:4004/com.sap.cap.fe.ts.sample/resources/sap-ui-core.js)

This is fixed in this PR.


packages/preview-middleware/src/base/config.ts
```diff
@@ -261,16 +262,17 @@ export function createFlpTemplateConfig(config: FlpConfig, manifest: Partial<Man
...
+        basePath: basePath,
         apps: {},
         init: config.init ? ns + config.init : undefined,
         ui5: {
             libs: getUI5Libs(manifest),
             theme: ui5Theme,
             flex,
             resources: {
-                [PREVIEW_URL.client.ns]: PREVIEW_URL.client.url
+                [PREVIEW_URL.client.ns]: PREVIEW_URL.client.getUrl(basePath)
```
heimwege, line 275: An alternative would be to keep PREVIEW_URL as is and add the basePath here
`[PREVIEW_URL.client.ns]: posix.join(basePath, PREVIEW_URL.client.url)`
Not sure which one is better 🤷🏻

heimwege:
Now I'm being blamed from the Inclusive Language Checker for the main branch name of OpenUI5. That's not fair 😿

packages/preview-middleware/src/base/config.ts
tobiasqueck: It's a matter of taste, go for what you prefer.

heimwege: OK then let's leave it as it currently is. This hopefully reduces the risk that someone else just uses the path als URL without basePath

tobiasqueck:
@heimwege while you were fixing the occurrences of the absolute url, someone added another one in `main` at https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/templates/flp/sandbox.html#L43. Could you please also handle it?

heimwege:
> @heimwege while you were fixing the occurrences of the absolute url, someone added another one in `main` at https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/templates/flp/sandbox.html#L43. Could you please also handle it?

I adjusted that one but still it does not run with cds-plugin-ui because there seem to be other locations as well: [bootstrap.js](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/src/flp/bootstrap.js#L12) 🙈

packages/preview-middleware-client/src/flp/bootstrap.js
```diff
@@ -9,14 +9,18 @@
  * @param fnCallback {Function} The callback function to be executed after the bootstrap is loaded.
  */
 async function ushellBootstrap(fnCallback) {
-    let src = '/test-resources/sap/ushell/bootstrap/sandbox.js';
+    const currentPath = location.pathname.substring(0, location.pathname.lastIndexOf('/'));
```
heimwege, line 12: Not sure if that's the best way to do this but at least it seems to work with cds-plugin-ui5 as well as without it (unit tests are still green 🐱). Basic idea taken from [testsuite.qunit.js template](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/templates/test/testsuite.qunit.js#L3)

tobiasqueck: Not beautiful but I don't have a smart suggestion, so I am ok with it.

tobiasqueck (approved):
* changeset ok
* important additional info in the readme
* didn't locally test but checked the changes of the snapshots and they meet expectations

tobiasqueck (approved):
reviewed changes since last approval: all as expected

voicis (approved):
- changeset present
- code adheres to the coding standards
- good test coverage
- didn't test locally

Issue #2137, opened 2024-07-10
FEATURE -  Add Adaptation Project's Add Components Usage generator prompting


mmilko01:
As a developer of an Adaptation Project,
I want to add a component usage to a base application.

### Description

After creating an Adaptation Project for a give base application, we can utilize a new feature to add a component usage to a base application.

This feature provides a CLI command sap-ux add component-usage that starts a prompting guiding us in the process adding a component usage.

### Technical Design

* enhance the `adp-tooling`
  * add prompting relevant to add component usage functionality in an Adaptation Project
* enhance the `create`
  * add CLI command `sap-ux add component-usage <path>`

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have an Adaptation Project that needs to add component usage to a base application
**when** I execute `sap-ux add component-usage <path>`
**then** prompting for specifying the new component file is started and after finished results in new change files in `changes` directory of the Adaptation Project 

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #2184, opened 2024-07-19
feat: Add Adaptation Project's Component Usages prompting

mmilko01:
Feat for https://github.com/SAP/open-ux-tools/issues/2137.

- Adds Component Usages prompts to be used for prompting independent of environment.
- Add CLI command to prompt and generate a change.

### DIFF
--- a/packages/adp-tooling/src/prompts/add-component-usages/index.ts
+++ b/packages/adp-tooling/src/prompts/add-component-usages/index.ts
@@ -31,88 +31,88 @@ export function getPrompts(basePath: string, layer: UI5FlexLayer): YUIQuestion<A
         {
             type: 'input',
             name: `id`,
-            message: t('prompts.componentUsageIDLabel'),
+            message: t('prompts.component.usageIdLabel'),
             validate: (value: string) =>
                 validateContentDuplication(
                     value,
                     'componentUsages',
                     componentUsageChangeFiles,
                     isCustomer,
-                    t('prompts.componentUsageIDLabel'),
-                    t('prompts.componentUsage')
+                    t('prompts.component.usageIdLabel'),
+                    t('prompts.component.usage')
                 ),
             default: isCustomer ? 'customer.' : '',
             store: false,
             guiOptions: {
                 mandatory: true,
-                hint: t('prompts.componentUsageIDTooltip')
+                hint: t('prompts.component.usageIdTooltip')
             }
         } as InputQuestion<AddComponentUsageAnswers>,
         {
             type: 'input',
             name: 'name',
-            message: t('prompts.componentNameLabel'),
-            validate: (value: string) => validateSpecialChars(value, t('prompts.componentNameLabel')),
+            message: t('prompts.component.nameLabel'),
+            validate: (value: string) => validateSpecialChars(value, t('prompts.component.nameLabel')),
             store: false,
             guiOptions: {
                 mandatory: true,
-                hint: t('prompts.componentNameTooltip')
+                hint: t('prompts.component.nameTooltip')
             }
         } as InputQuestion<AddComponentUsageAnswers>,
         {
             type: 'list',
             name: 'isLazy',
-            message: t('prompts.componentIsLazyLabel'),
+            message: t('prompts.component.isLazyLabel'),
             choices: isLazyDropDownOptions,
             default: isLazyDropDownOptions[1].value,
             store: false,
             guiOptions: {
                 mandatory: true,
-                hint: t('prompts.componentIsLazyTooltip')
+                hint: t('prompts.component.isLazyTooltip')
             }
         } as ListQuestion<AddComponentUsageAnswers>,
         {
             type: 'editor',
             name: `settings`,
-            message: t('prompts.componentSettingsLabel'),
+            message: t('prompts.component.settingsLabel'),
             validate: (value: string) => {
-                return validateJSON(value, t('prompts.componentSettingsLabel'));
+                return validateJSON(value, t('prompts.component.settingsLabel'));
             },
             store: false,
             guiOptions: {
                 mandatory: false,
-                hint: t('componentSettingsTooltip')
+                hint: t('component.settingsTooltip')
             }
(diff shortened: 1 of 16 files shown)


packages/adp-tooling/test/unit/prompts/add-component-usages/indest.test.ts
```diff
@@ -0,0 +1,314 @@
+import * as i18n from '../../../../src/i18n';
```
tobiasqueck, line 1: Seems like a type in the file name.

packages/adp-tooling/src/types.ts
```diff
@@ -307,26 +307,27 @@ export const enum AnnotationFileSelectType {
...
-        /** Reference to the component's library. */
-        reference: string;
-        /** Optional flag indicating if the library reference is lazy. */
-        referenceIsLazy?: string;
-    };
+    variant: DescriptorVariant;
+    answers: AddComponentUsageAnswers;
+}
+
+export interface AddComponentUsageAnswers {
```
tobiasqueck, line 314: You changed a clean structured type into something flat because it is the outcome of the prompting. Please keep the clean structure you had before, and map your flat prompting output to the structure.
Keep in mind that we separate writer and prompting on purpose with clean APIs so that they can be easily changed/enhanced in the future without e.g. having the writer rely on the name of prompts.

packages/create/src/cli/add/component-usages.ts
```diff
@@ -0,0 +1,52 @@
...
+import { generateChange, getVariant, ChangeType, getPromptsForAddComponentUsages } from '@sap-ux/adp-tooling';
+import { getLogger, traceChanges } from '../../tracing';
+import { validateAdpProject } from '../../validation/validation';
+import { promptYUIQuestions } from '../../common';
+
+/**
+ * Add a new sub-command to change the data source of an adaptation project to the given command.
+ *
+ * @param {Command} cmd - The command to add the change data-source sub-command to.
+ */
```
testojs, line 11: This is the command for `add component usages`.

packages/adp-tooling/src/types.ts
mmilko01: As far as I can see we followed the same approach of flattening of the structure in all other generators. Like in the `change-data-source`
https://github.com/SAP/open-ux-tools/blob/cf5dabaab0eb3ce30431b1b04adf566151d141da/packages/adp-tooling/src/types.ts#L391 
or most recently in the `new-model`
https://github.com/SAP/open-ux-tools/blob/34836760f91921b6b721f0b53c7643ffdb8a6250/packages/adp-tooling/src/types.ts#L337

This was convenient to avoid big chunks of mapping in the CLI and the Yeoman UI. If we change the approach now, we should refactor the other generators as well. What do you think?

packages/adp-tooling/test/unit/prompts/add-component-usages/indest.test.ts
mmilko01: Fixed in: bdd8262301bee934e0dcfd5def4dd5d18267b793

packages/adp-tooling/src/prompts/add-component-usages/index.ts
```diff
@@ -0,0 +1,234 @@
...
+            }
+        } as ListQuestion<AddComponentUsageAnswers>,
+        {
+            type: 'editor',
+            name: `settings`,
+            message: t('prompts.component.settingsLabel'),
+            validate: validatePromptJSON,
+            store: false,
+            guiOptions: {
+                mandatory: false,
```
testojs, line 184: Redundant.

packages/adp-tooling/src/prompts/add-component-usages/index.ts
```diff
@@ -0,0 +1,234 @@
...
+            }
+        } as EditorQuestion<AddComponentUsageAnswers>,
+        {
+            type: 'editor',
+            name: `data`,
+            message: t('prompts.component.dataLabel'),
+            validate: validatePromptJSON,
+            store: false,
+            guiOptions: {
+                mandatory: false,
```
testojs, line 195: Redundant. Default.

testojs (changes requested):

packages/create/src/cli/add/component-usages.ts
mmilko01: Fixed with: 4ea2aa9c0137314014ddeeee2e8e1df80149357b

packages/adp-tooling/src/prompts/add-component-usages/index.ts
mmilko01: Fixed with: 7a2cf5e70cb2858aaebe90b1999c68cc169ce66a

testojs (approved):
* code looks good
* comments addressed
* perfect test coverage
* not tested manually

packages/adp-tooling/src/types.ts
tobiasqueck: omg ... yes, I cleared missed that in other PRs. The input for the writer needs to be well defined and structured. It is on purpose not a flat list of answers from prompts. It is totally ok, that in the create a transformation between flat prompt results and proper structure happens.
The goal of separating both prompts and writer is that they can be used separately when needed. If you do it like it is done now, each change of a prompt will change the API of a write which is unacceptable. 
Please keep this guideline in mind for the prompting module in this PR and the other ones we have to fix later.

tobiasqueck: As agreed on the phone, I will approve it for now, and you will look into https://github.com/SAP/open-ux-tools/issues/2216 afterwards

tobiasqueck (commented):
* changeset good
* code is easy to understand - discussed issue will be followed up in a different PR
* great test coverage
* problems with a local test: I entered `{"x":"g"}` as settings but it tells me it is invalid json

tobiasqueck (approved):
Just learned that I need to enter it without the wrapping `{}` e.g. `"x":"g", "t":6` . While I think it should be different, it needs to stay the same for backward compatibility.

PR #2187, opened 2024-07-23
redirect to new sandbox if the old one is not available

schreckstefan:


schreckstefan:
@tobiasqueck changes taken over to https://github.com/SAP/open-ux-tools/pull/2159, thus closing this one

PR #2188, opened 2024-07-23
fix: xml formatting when moving element

nahmed22:
Fix  readability issue with formatting in `annotation.xml` after section is droped/moved to group section


packages/fiori-annotation-api/src/xml/writer.ts
```diff
@@ -268,6 +268,40 @@ function handleXmlElementAttributeInserts(element: XMLElement, attributeInserts:
     return edits;
 }
 
+function getTextFragmentIndentation(text: string): number {
+    const lines = text.split('\n');
+    const linesWithTextIndices = lines.map((line, idx) => (line.length > 0 ? idx : -1));
+    const indentTextIndex = linesWithTextIndices.find((idx) => idx > -1);
+    if (indentTextIndex == undefined) {
```
vadson71, line 275: === are better than ==

packages/fiori-annotation-api/src/xml/writer.ts
```diff
@@ -268,6 +268,40 @@ function handleXmlElementAttributeInserts(element: XMLElement, attributeInserts:
...
+        return 0;
+    }
+    const indentLineText = lines[indentTextIndex];
+    return Math.ceil(indentLineText.length - indentLineText.trimStart().length) / 4;
```
vadson71, line 279: ceil should be calculated after division by 4
Would be nice to have a test case for it

vadson71 (changes requested):

packages/fiori-annotation-api/src/xml/writer.ts
nahmed22: done

packages/fiori-annotation-api/src/xml/writer.ts
nahmed22: done

vadson71 (approved):
Changes look good, Comments addressed. Tests coverage is good

Issue #2186, opened 2024-07-22
TBI - change @sap-ux/ui5-info to use version.json


Klaus-Keller:
### Description
The usage of https://ui5.sap.com/versionoverview.json is discouraged as the format is not stable and might change any time. Instead we should use https://ui5.sap.com/version.json?sap-ui-config-patches=true&sap-ui-config-showall=true. 

If we encounter any missing information, we can reach out to the UI5 team to add it to `version.json?sap-ui-config-patches=true&sap-ui-config-showall=true`.

### Value
Usage of stable API to avoid future issues.

### Architecture Elaboration
`versionoverview.json` is defined in constant here
https://github.com/SAP/open-ux-tools/blob/08957786a0a1a18eafcd49c157cce1c0e849ca13/packages/ui5-info/src/constants.ts#L23

and should be replaced by 
https://github.com/SAP/open-ux-tools/blob/08957786a0a1a18eafcd49c157cce1c0e849ca13/packages/ui5-info/src/constants.ts#L22

Seems we use only property `support` from `versionoverview.json`, which is also included in `version.json?sap-ui-config-patches=true&sap-ui-config-showall=true`

### Tasks
- [x] Remove usage of versionoverview.json

PR #2189, opened 2024-07-24
tbi(ui5-info): move away from versionoverview

mikicvi-SAP:
Issue: #2186 

---
- Moves away from using `versionoverview.json` for metadata per ui5 version
- Adjusts unit test data, removes redundant test data


packages/ui5-info/test/__snapshots__/ui5-version-info.test.ts.snap
mikicvi-SAP, line 1: Snapshots are now aligned with the test data from [official-response.json](https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-info/test/testdata/official-response.json)

packages/ui5-info/src/types.ts
```diff
@@ -46,20 +46,16 @@ export interface UI5Version {
      * UI5 semantic version
      */
     version: string;
+    support?: string;
+    lts?: string;
```
Klaus-Keller, line 50: Hi @mikicvi-SAP, do we need to change the type that is used as return type of public interface function `getUI5Versions()`?

Perhaps we can have a type that extends `UI5Version` by `support` and `lts` and is used internally. Or was this intentional to extend the return type?

Klaus-Keller (commented):
See my comment

packages/ui5-info/src/types.ts
mikicvi-SAP: The change was intentional, as it made the most sense and it reads well. 
Of course, if you think that `UI5Version` type should be extended by another type, that can be done instead.

Klaus-Keller: Hi @mikicvi-SAP,

Consumers call `getUI5Versions()` can control the presence of properties `default` and `maintained` via options:
https://github.com/SAP/open-ux-tools/blob/13d5ee59f6ab2b7c05a74b86d78729fc1d13dad7/packages/ui5-info/src/types.ts#L30-L37

Is it also possible for a consumer to fill the newly added properties `support` and `lts`?

If not I would say we keep the public interface for return value of `getUI5Versions()` as is and create a internal interface based on it for internal handling like cache storage.

mikicvi-SAP: Hi @Klaus-Keller, 

Yes, the consumer can specify the properties, calling `getUI5Versions()` will return only the version. The support is returned when the `includeMaintained` property is passed in.

As far as `lts` goes, it has been in the type before, but it hasn't been used anywhere. 
It can be added, in a similar fashion, we would have to add a new parameter `includeLts?:boolean`.

I have kept it in as I did not want to remove any properties, I don't see any usages in ui5-info so it is redundant as of now.

packages/ui5-info/src/ui5-version-info.ts
```diff
@@ -165,10 +167,10 @@ async function parseUI5VersionsOverview(): Promise<UI5VersionOverview[]> {
...
 const retrieveUI5VersionsCache = async (
-    type: ui5VersionsType.official | ui5VersionsType.snapshot | ui5VersionsType.overview,
+    type: ui5VersionsType.official | ui5VersionsType.snapshot | ui5VersionsType.support,
     useCache = false,
     snapshotUrl?: string
-): Promise<string[] | UI5VersionOverview[]> => {
+): Promise<string[] | UI5Version[]> => {
     if (!useCache) {
         switch (type) {
             case ui5VersionsType.official:
```
devinea, line 176:
```suggestion
            case ui5VersionsType.official:
            case ui5VersionsType.support:
```
Shouldn't we just merge official and support as they are both making the same API call. 
Also, I think in some flows require both data and should now be able to get the data using a single HTTP call instead of the 2 previously.

packages/ui5-info/test/__snapshots__/ui5-version-info.test.ts.snap
```diff
@@ -3113,67 +3113,67 @@ exports[`getUI5Versions filterOptions: returns support info - includeMaintained
...
     "version": "1.102.0",
   },
   {
-    "maintained": true,
+    "maintained": false,
     "version": "1.101.1",
   },
   {
-    "maintained": true,
+    "maintained": false,
```
devinea, line 3176: All these diffs are a bit alarming. 
But based on the test data https://github.com/SAP/open-ux-tools/blob/d773961f479d876872099aa27b6ee324f7480438/packages/ui5-info/test/testdata/official-response.json 
it seems it is OK.

devinea (commented):
A few comments

packages/ui5-info/src/types.ts
```diff
@@ -50,16 +50,15 @@ export interface UI5Version {
...
+export interface UI5VersionSupport extends UI5Version {
+    support?: string;
     lts?: string;
 }
 
 export interface UI5VersionsResponse {
     [key: string]: {
         patches?: string[];
-    } & UI5VersionOverview;
+    } & UI5Version;
```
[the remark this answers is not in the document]
mikicvi-SAP: Correct, I missed that one. Thanks @devinea

packages/ui5-info/src/ui5-version-info.ts
mikicvi-SAP: Agreed. 
Initially, I tried to take the least invasive approach to reduce the risk of regressions, but now that we're making a call to the same API, it makes sense to consolidate.
Implemented in [2062785](https://github.com/SAP/open-ux-tools/pull/2189/commits/20627858a7f146eb6dabc79f8d01c4dad53546c5).
Thanks @devinea

devinea (dismissed):
Thanks for the updates @mikicvi-SAP 
Changes and tests look as expected. 
changeset ✅

devinea (approved):
Re-approving

longieirl (approved):
Changes look good.
Confirmed, there is no deprecated changes required.
Test coverage is good.

PR #2175, opened 2024-07-18
fix: getMinUI5VersionAsArray

schreckstefan:
Issue: https://github.com/SAP/open-ux-tools/issues/2173

Introduces the semver valid check at getMinUI5VersionAsArray; only valid version get returned.
But it can be controlled by a new parameter `noValidation` if the check takes place.

Also includes an upgrade of @ui5/manifest to 1.66.0 at project-access and ui5-application-writer. 
Some adjustment were needed afterwards at components working with minUI5Version, due to the new array option for it.

PR #2190, opened 2024-07-24
fix(fe-fpm-writer): add `@sap-ux/project-access` to dependencies

johannes-kolbe:
With #2175, functions (not only types any longer) from `@sap-ux/project-access` were added to fe-fpm-writer: https://github.com/SAP/open-ux-tools/pull/2175/files#diff-128f273a33ffbb49ddbdff27138d11ef5be6d7deaedcb60e65b52aba00f7932bR6

During consumption of `@sap-ux/fe-fpm-writer` that dependency is now required.

Adding to `dependencies` to add the required module.

### DIFF
--- a/.changeset/warm-shirts-warn.md
+++ b/.changeset/warm-shirts-warn.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fe-fpm-writer': patch
+---
+
+fix: add `@sap-ux/project-access` dependency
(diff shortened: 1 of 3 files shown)


815are (approved):
checked:
1. transition from `devDependency` to `dependency` looks valid based on https://github.com/SAP/open-ux-tools/pull/2175
previously only types were used from `@sap-ux/project-access`, but now functions `getMinUI5VersionAsArray, getMinimumUI5Version` are used in source code

schreckstefan (approved):

PR #2192, opened 2024-07-25
Preview 2.0 (2nd try)

schreckstefan:


packages/ui5-proxy-middleware/src/base/proxy.ts
```diff
@@ -28,7 +28,7 @@ export const ui5Proxy = (config: ProxyConfig, options?: Options, filter?: Filter
     });
     const today = new Date();
     const etag = `W/"${config.version || 'ui5-latest-' + today.getDate() + today.getMonth() + today.getFullYear()}"`;
-    const ui5Ver = config.version ? `/${config.version}` : '';
+    const ui5Ver = config.version && !config.version.startsWith('2') ? `/${config.version}` : '';
```
tobiasqueck, line 31: That line is odd.

heimwege (commented):
Just some smaller finding

packages/ui5-proxy-middleware/src/base/proxy.ts
schreckstefan: @tobiasqueck 
without this, the redirect adds /2.0.0/ on the url when trying to access thre /resources, and this request then fails.
Guess that UI5 will introduce version-specific subfolders on their final destination, but at the moment (= temporary test instances) they do not exist.

packages/preview-middleware-client/src/flp/bootstrap.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Calculates the script content for accessing the right sap/ushell/bootstrap sandbox.
+ */
+(window as any)['sap-ui-config'] = {
+    'open-ux-bootTask': ushellBootstrap
+};
+
+function ushellBootstrap(fnCallback: any) {
```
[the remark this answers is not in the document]
schreckstefan: done

packages/preview-middleware-client/src/flp/init.ts
```diff
@@ -304,15 +305,23 @@ export async function init({
     const resourceBundle = await loadI18nResourceBundle(scenario as Scenario);
     setI18nTitle(resourceBundle);
     registerSAPFonts();
-    const renderer = await container.createRenderer(undefined, true);
+    const major = version ? parseInt(version.split('.')[0], 10) : 2;
+    const renderer =
+        major < 2
+            ? await container.createRenderer(undefined, true)
+            : await (container as any).createRendererInternal(undefined, true);
```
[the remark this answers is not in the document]
schreckstefan: Good idea, done

packages/preview-middleware-client/src/flp/init.ts
```diff
@@ -304,15 +305,23 @@ export async function init({
...
-        customInit: bootstrapConfig.getAttribute('data-open-ux-preview-customInit')
-    }).catch(() => Log.error('Sandbox initialization failed.'));
+    try {
+        init({
+            appUrls: bootstrapConfig.getAttribute('data-open-ux-preview-libs-manifests'),
+            flex: bootstrapConfig.getAttribute('data-open-ux-preview-flex-settings'),
+            customInit: bootstrapConfig.getAttribute('data-open-ux-preview-customInit')
+        });
+    } catch (e) {
+        Log.error('Sandbox initialization failed: ' + e.message);
```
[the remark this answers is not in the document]
schreckstefan: using getError now

packages/preview-middleware-client/src/flp/init.ts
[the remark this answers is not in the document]
heimwege: You can remove the `any` here as it is now fully typed by `InternalContainer`.

```suggestion
            : await container.createRendererInternal(undefined, true);
```

packages/ui5-proxy-middleware/src/base/proxy.ts
tobiasqueck: That is a yaml config, if you add a version there, then it is expected to be added to the url. If you don't then it isn't. So, the yaml config needs to just keep it empty, and must not add a 2.0.0 if the server doesn't offer the specific versioning.
That's nothing you need to "fix" in your code, it is a configuration by the user/consumer.

packages/preview-middleware-client/src/flp/bootstrap.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Calculates the script content for accessing the right sap/ushell/bootstrap sandbox.
+ */
+(window as any)['sap-ui-config'] = {
+    'xx-bootTask': ushellBootstrap
+};
+
+function ushellBootstrap(fnCallback: Function) {
+    fetch('/resources/sap-ui-version.json')
```
tobiasqueck, line 9: Please use async/await. It makes the code more readable.
Alternatively (and maybe cleaner), use `sap.ui.require` and get `sap/ui/VersionInfo` instead of getting the json file

packages/preview-middleware-client/src/flp/bootstrap.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Calculates the script content for accessing the right sap/ushell/bootstrap sandbox.
+ */
+(window as any)['sap-ui-config'] = {
```
tobiasqueck, line 4: please remove any

packages/preview-middleware-client/src/flp/bootstrap.ts
schreckstefan: tried this, but it leads to many other eslint problems, fur t the asyn call

packages/preview-middleware-client/src/flp/bootstrap.ts
```diff
@@ -0,0 +1,34 @@
+import { getError } from '../cpe/error-utils';
```
schreckstefan, line 1: have to remove this again, it leads to "sap is not defined"

schreckstefan: next problem: the import from from 'sap/open/ux/preview/global' led to an export / ESM error at run-time. 
Replaced by a local definition.

schreckstefan: Both things solved, but now facing
image

packages/preview-middleware-client/src/flp/bootstrap.ts
schreckstefan: done

packages/preview-middleware-client/src/flp/bootstrap.ts
schreckstefan: done: had to move the call of initConnectors

packages/preview-middleware-client/src/flp/bootstrap.ts
heimwege: Hm. It seems to work [here](https://github.com/SAP/open-ux-tools/blob/988727a9a226755df3d90f27b71369597676cd01/packages/preview-middleware-client/src/flp/init.ts#L276) pretty well 🤔 

```ts
import VersionInfo from 'sap/ui/VersionInfo';
//...
export async function ushellBootstrap(fnCallback: () => void)
    const { version } = (await VersionInfo.load()) as { version: string };
    const major = Number(version?.charAt(0)) || 2;
//...
```

or is `'xx-bootTask'` not allowed to return a promise?

packages/preview-middleware-client/types/global.d.ts
```diff
@@ -0,0 +1,7 @@
+declare module 'sap/open/ux/preview/global' {
```
heimwege, line 1: [Bootstrap.ts](https://github.com/SAP/open-ux-tools/pull/2192/files#diff-8d76ca9a519d1d51f8dcbd0428e926d6ce2e33cfc4120c47390fbd9d2062bc52) has its own interface definition for `Window`. In my opinion we should either use a global one or a local one but not both in parallel.

packages/preview-middleware-client/src/flp/bootstrap.ts
tobiasqueck: The problem with the import like that at the top of the page would be executed too early. It can only be executed within the script because only then UI5 is loaded.

packages/preview-middleware-client/types/global.d.ts
tobiasqueck: Absolutely, one is enough. And I also don't understand the wrapping 'declare' at the top, it is not necessary. I'd keep it global (and not locally in the bootsrap.ts) because it is globally valid in our context. We might even enhance it further with other global configs in UI5 (or QUnit).

packages/preview-middleware-client/src/flp/bootstrap.ts
heimwege: Ok, understood. But not giving up yet 😄 How about using a dynamic import then?

```ts
//...
export async function ushellBootstrap(fnCallback: () => void)
    const versionInfo = (await import('sap/ui/VersionInfo')).default;
    const { version } = await versionInfo.load() as { version: string };
    const major = Number(version?.charAt(0)) || 2;
//...
```

packages/preview-middleware-client/types/global.d.ts
tobiasqueck: @heimwege please check again

packages/preview-middleware-client/src/flp/bootstrap.ts
tobiasqueck: I needed to switch the file to js so that the transpiler doesn't make it UI5 conform. That at least makes the script finally work properly. I will now check your proposal above which looks good to me

packages/preview-middleware-client/src/flp/init.ts
```diff
@@ -317,6 +324,9 @@ export async function init({
         await registerComponentDependencyPaths(JSON.parse(appUrls), urlParams);
     }
 
+    // Load rta connector
```
tobiasqueck, line 327: @schreckstefan why did you add this here?

tobiasqueck: @testojs please check if has any negative effects on ADP

packages/preview-middleware-client/src/flp/bootstrap.ts
heimwege: Fine for me. I don't have anything working right now. It was just an idea. Did not try 🙈 Will push some eslint fixes

packages/preview-middleware-client/src/flp/init.ts
```diff
@@ -7,6 +7,9 @@ import IconPool from 'sap/ui/core/IconPool';
 import ResourceBundle from 'sap/base/i18n/ResourceBundle';
 import AppState from 'sap/ushell/services/AppState';
 import { getManifestAppdescr } from '../adp/api-handler';
+import VersionInfo from 'sap/ui/VersionInfo';
+import { getError } from '../cpe/error-utils';
+import initConnectors from './initConnectors';
```
heimwege, line 12: That line causes ui5 tooling served via npmjs (so without proxy) to no longer work because `LrepConnector.js` is not being served via UI5 tooling / npmjs

image

image

see also https://www.npmjs.com/package/@openui5/sap.ui.fl?activeTab=code

I don't yet understand why this is not an issue when serving the UI5 sources via proxy because
https://sapui5.hana.ondemand.com/1.124.0/resources/sap/ui/fl/LrepConnector.js also does not exist 🤷🏻 
(whereas https://sapui5.hana.ondemand.com/1.124.0/resources/sap/ui/fl/FakeLrepConnector.js does)

What would work is https://sapui5.hana.ondemand.com/1.124.0/resources/sap/ui/fl/initial/_internal/connectors/LrepConnector.js

I don't see any of those request when running via proxy

tobiasqueck: Wouldn't a a trycatch around the execution of `initConnectors` "solve" the problem?

packages/preview-middleware-client/src/flp/init.ts
testojs: > @testojs please check if has any negative effects on ADP

Should not be an issue.

packages/preview-middleware-client/src/flp/init.ts
heimwege: Ok, seems like when served via npmjs the version is the app version from package.json❗

image

loading that directly from CDN it contains the SAPUI5 Distribution version (which is ok as it comes directly from UI5)

image

So the `version` property we use is simply the wrong one. We should use the version of the sap.fe.core library

image

image

That should fix it

tobiasqueck: Wow, that is in interesting finding. It is actually a bigger problem then just the 2.0 enablement since it already causes a wrong behavior in other scenarios (that were just not as dramatic as now).

So, if I understood correctly, the correct solution would be to not look at the top level version but look at the libraries, find `sap.ui.core` and then take the version from there. That would work for both, a CDN ui5 version as well as something running locally with the modules from npmjs. @heimwege did I get it correctly?

@cfg74 should it get fix in this PR or is later ok? It will break the preview-middleware usage for runs with the `ui5-local.yaml` in Fiori tools projects.

heimwege: yep that's also my understanding 👍🏻

heimwege: There are quite some locations where the version is being used like this. Maybe having this in a central place would be a good idea

cfg74: @tobiasqueck breaking things is bad, why would it break the preview from downloaded sources?

heimwege: @cfg74 because the UI5 version handling is currently (and most likely has ever been) wrong. As of now it was assumed that the top level version from `sap-ui-version.json` represents the UI5 version but it represents the application version (from the application manifest `sap.app.applicationVersion.version`). Only when loading that file [directly from UI5](https://sapui5.hana.ondemand.com/resources/sap-ui-version.json) (e.g. via proxy and thus without the context of an actual application) the top level version equals the UI5 version.

In my case the application version is 1.0.0 and then tooling assumes this is my UI5 version and strange things happen 🙈

heimwege: But I'm working on fixing this and as son as @tobiasqueck agrees (because its contains changes in a lot of files and I'm not sure if all of this should be part of this PR) I can push my changes. Locally it seems to work pretty fine with the adjustments

cfg74: @heimwege @tobiasqueck clarified with Tobias, that _without_ the fix it will break local preview, hence we have to do the fix to avoid a regression. Can one of you take that please?

heimwege: I just pushed my changes. Hopefully that will fix it 🤞🏻
At least a test on my local machine was successfully with and without proxy

packages/preview-middleware-client/src/cpe/connector-service.ts
```diff
@@ -12,7 +13,7 @@ export class WorkspaceConnectorService {
      * @param sendAction action sender function
      */
     public async init(sendAction: ActionSenderFunction): Promise<void> {
-        const { version } = (await VersionInfo.load()) as { version: string };
+        const version = (await VersionInfo.load() as VersionInformation)?.libraries?.find((lib) => lib.name === 'sap.ui.core')?.version ?? '';
```
heimwege, line 16: `as string` or  `?? ''` which one is to be preferred? 🤔

tobiasqueck: as string is wrong because it could be undefined, so ?? '' is the better option

heimwege:
Cross link to https://github.com/SAP/openui5/issues/4105 for a better type description for `sap-ui-version.json`. Once it is done [this one](https://github.com/SAP/open-ux-tools/blob/26ebcaef2d2d240990da9fbd1d36c64b71933fbc/packages/preview-middleware-client/types/global.d.ts#L7) can be deleted.

packages/preview-middleware-client/src/flp/init.ts
tobiasqueck: I have tested it locally and it works like a charm!

tobiasqueck (approved):
* changeset ok
* successfully tested with CDN 1.x/2.0 and local 1.x

packages/preview-middleware-client/src/cpe/connector-service.ts
heimwege: If `undefined` is a real option there's quite some more to be adjusted. See [here](https://github.com/SAP/open-ux-tools/blob/feat/1679/preview20/packages/preview-middleware-client/src/adp/init.ts#L21) or [here](https://github.com/SAP/open-ux-tools/blob/feat/1679/preview20/packages/preview-middleware-client/src/cpe/connector-service.ts#L17) or [here](https://github.com/SAP/open-ux-tools/blob/feat/1679/preview20/packages/preview-middleware-client/src/cpe/outline/nodes.ts#L92) or ... all of these ignore a potentially `undefined` `version` property.

Good thing is the type checker finds all of them now by adjusting [SingleVersionInfo](https://github.com/SAP/open-ux-tools/blob/feat/1679/preview20/packages/preview-middleware-client/types/global.d.ts#L7) to be `| undefined` which is anyway a better type according to [sap/ui/VersionInfo.load](https://github.com/SAP/openui5/blob/master/src/sap.ui.core/src/sap/ui/VersionInfo.js#L40).

Errors  Files
     2  src/adp/init.ts:21
     1  src/cpe/connector-service.ts:17
     1  src/cpe/outline/nodes.ts:92
     1  src/flp/init.ts:281
     1  src/flp/WorkspaceConnector.ts:74

Klaus-Keller (approved):
Thanks,
- code to address UI5 2.x differences looks good
- changeset exists
- code coverage is good
- did a local test and it works

packages/preview-middleware-client/src/cpe/connector-service.ts
heimwege: I created a new issue for that one to be done in an other PR at a later point in time: https://github.com/SAP/open-ux-tools/issues/2207

Issue #2193, opened 2024-07-25
BUG - `@sap-ux/odata-service-inquirer` multiple issues with new system -> Abap-on-prem prompting


IainSAP:
### Description

Using the 'Abap On Premise' new system prompt, when authorization is required the password prompt incorrectly validates as true when a 401 or 403 is returned. 

A number of other related issues are also present:

- changing sap-client does not re-authenticate
- new system name is allowing an empty string which breaks storage
- new system names gets overwritten when changing client or system url

### Expected results
The `password` prompt validator returns an appropriate error message.

### Actual results
The `password` prompt validator returns true when a 401/403 is returned from the connection validator

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2196, opened 2024-07-26
Fixes for multiple abap-on-prem new system prompts

IainSAP:
https://github.com/SAP/open-ux-tools/issues/2193

Fixes for :

- failed basic auth allows user to continue
- changing sap-client does not re-authenticate
- new system name is allowing an empty string which breaks storage
- new system names gets overwritten when changing client or system url

### DIFF
--- a/packages/odata-service-inquirer/src/prompts/connectionValidator.ts
+++ b/packages/odata-service-inquirer/src/prompts/connectionValidator.ts
@@ -25,7 +25,7 @@ interface Validity {
     urlFormat?: boolean;
     // True if the url is reachable
     reachable?: boolean;
-    // True if the url requires authentication (note even once authenticated, this will remain true)
+    // True if the url requires authentication, i.e. returns a 401/403  (note even once authenticated, this will remain true)
     authRequired?: boolean;
     // True if the url is authenticated and accessible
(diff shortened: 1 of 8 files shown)


packages/odata-service-inquirer/src/prompts/connectionValidator.ts
```diff
@@ -387,11 +441,18 @@ export class ConnectionValidator {
                 isSystem,
                 odataVersion
             });
-
+            LoggerHelper.logger.debug(`ConnectionValidator.validateUrl() - status: ${status}; url: ${url}`);
+            // Since an exception was not thrown, this is a valid url
+            this.validity.urlFormat = true;
+            this._validatedUrl = url;
```
longieirl, line 447: Do we need to ensure that even if a http 200 is returned, it's not a saml HTML response.

IainSAP: I think that should be dealt with by a lower level module, i.e. axios extension?

longieirl (approved):
Changes look good.
Test coverage is good.
Changeset is present and correct.

korotkovao (approved):
Looks good to me, thank you @IainSAP

Issue #2200, opened 2024-07-30
TBI - Update CAP serviceInfo API handling


korotkovao:
### Description (include screenshots)
We should update to use endpoints of kind: 'odata' from each service instead of existing urlPath

### Value
Support updated API available in @sap/cds: 7.8.0

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

PR #2201, opened 2024-07-30
feat(project-access): 29050 update cap service info api handling 

korotkovao:
## #2200

Support updated API available in @sap/cds: 7.8.0
Update to use endpoints of kind: 'odata' from each service instead of existing urlPath


packages/project-access/src/project/cap.ts
```diff
@@ -156,7 +156,7 @@ export async function getCapModelAndServices(
         services = services.map((value) => {
             return {
                 name: value.name,
-                urlPath: uniformUrl(value.urlPath),
+                urlPath: uniformUrl(value.endpoints?.find((val) => val.kind === 'odata')?.path ?? value.urlPath),
```
kjose90, line 159: Would be nice to improve readability 👍 

```suggestion
const { endpoints, urlPath } = value;
const odataEndpoint = endpoints?.find((endpoint) => endpoint.kind === 'odata');
const endpointPath = odataEndpoint?.path ?? urlPath;
return { ... urlPath: uniformUrl(endpointPath) ...};

```

korotkovao: Thank you, i have added the change as suggested.

longieirl (approved):
Changes look good.
Test coverage is good.
Changeset is good.

kjose90 (approved):
Changes looks good and clear
Changeset & tests ✅ 
Thanks.

packages/project-access/test/project/cap.test.ts
```diff
@@ -78,7 +78,184 @@ describe('Test getCapModelAndServices()', () => {
...
+                        },
+                        {
+                            'name': 'Backslash',
+                            'endpoints': [
+                                {
+                                    'path': '\\odata\\service\\with\\backslash/',
+                                    'kind': 'odata'
+                                }
+                            ]
+                        },
```
devinea, line 114: Path is a web URI so I don't this backslash is valid.

devinea (approved):
Code changes are clear. 
changeset ✅

packages/project-access/test/project/cap.test.ts
korotkovao: Thank you, it's a good point. I might leave this test there for now.

PR #2203, opened 2024-07-31
fix(security): update fast-xml-parser

donal-tobin-sap:
Issue: tools-suite/issues/30768
Issue: https://github.com/SAP/open-ux-tools/security/dependabot/110

### DIFF
--- a/.changeset/three-colts-count.md
+++ b/.changeset/three-colts-count.md
@@ -1,0 +1,39 @@
+---
+'@sap-ux/odata-service-inquirer': patch
+'@sap-ux/odata-service-writer': patch
+'@sap-ux/axios-extension': patch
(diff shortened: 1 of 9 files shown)


devinea (commented):
Looks like these resolutions can be removed as `@sap/wing-service-explorer` is no longer used. 
```
            "@sap/wing-service-explorer>axios@<1.6.8": ">=1.6.8",
            "@sap/wing-service-explorer>fast-xml-parser@4.2.4": ">=4.4.1",
```

hitesh-parmar (approved):
- changeset looks good
- update of fast-xml-parser also looks correct
- version override doc updated

devinea (approved):
Thanks for the updates

Issue #2173, opened 2024-07-18
BUG - getMinUI5VersionAsArray should only return valid versions


schreckstefan:
### Related Feature
Feature request: https://github.com/SAP/open-ux-tools/issues/1988

### Description
There is an inconsistency between the functions  `getMinimumUI5Version` and `getMinUI5VersionAsArray`:
- `getMinimumUI5Version` does a semver `valid` check and only returns a valid version
- `getMinUI5VersionAsArray` does not check validity and returns all versions as they are.

The request now is to transfer the `valid` check to `getMinUI5VersionAsArray` , so that only valid versions are returned by both functions, for consistency reasons. 
The consequence is, that placeholder notations like `${sap.ui5.dist.version}` will not be considered by both functions.
`getMinimumUI5Version` will return undefined, `getMinUI5VersionAsArray` an empty array.
Consumers that want to work with the manifest entry as is still can use function `getMinUI5VersionFromManifest`, but consumers 
of `getMinUI5VersionAsArray` don't need to apply validity checks on their own.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [X] Mac OS
- [X] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

schreckstefan:
Implemented by https://github.com/SAP/open-ux-tools/pull/2175

PR #2206, opened 2024-07-31
fix(#2173): consider variables in minUI5Version

hitesh-parmar:
Follow up to issue: #2173 

semver valid check at getMinUI5VersionAsArray was introduced with PR https://github.com/SAP/open-ux-tools/pull/2175 but this causes problems with applications using `variables` as minUI5Version. This PR fixes that.

Before this PR

Application Info Page
image

After this PR
Application Info Page
image


packages/project-access/test/project/minUI5Version.test.ts
```diff
@@ -9,6 +9,7 @@ describe('getters for minUI5Version', () => {
     describe('getMinUI5VersionFromManifest', () => {
         const testVersions: Array<TestCase> = [
             { minUI5Version: undefined, expectedResult: undefined },
+            { minUI5Version: '${sap.ui5.dist.veresion}', expectedResult: '${sap.ui5.dist.veresion}' },
```
devinea, line 12:
```suggestion
            { minUI5Version: '${sap.ui5.dist.version}', expectedResult: '${sap.ui5.dist.version}' },
```
Typo? Same below.

hitesh-parmar: Thanks for the review @devinea . Typo corrected with next commit

devinea (approved):
Changeset lg.

AlinaGovoruhina (approved):
App info page is updated, code is covered by tests

815are (approved):
checked:
1. change is straightforward and simple to understand 
2. test is updated with additional cases
3. changeset persists

Klaus-Keller:
@hitesh-parmar, can I veto this change. To see also invalid versions like `${sap.ui5.dist.version}` consumers can call the function `getMinUI5VersionAsArray()` with optional parameter `noValidation = true`.

hitesh-parmar:
Thanks for the review @Klaus-Keller. I will revert this change as discussed.
The reason for this solution is that I consider `${sap.ui5.dist.version}` as a valid version in the `manifest.json`. 

With the previous change valid versions are only those that can be loaded from CDN https://ui5.sap.com/

Issue #2207, opened 2024-07-31
BUG - handling of undefined response from sap/ui/VersionInfo.load


heimwege:
### Description
As described in https://github.com/SAP/open-ux-tools/pull/2192#discussion_r1698292305 the handling of `undefined` response from `sap/ui/VersionInfo.load` should be fixed.

1. Add `| undefined` at the end of [SingleVersionInfo](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/types/global.d.ts#L16) type
2. Fix errors reported by `tsc --noEmit`

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

heimwege:
Fixed in https://github.com/SAP/open-ux-tools/pull/2208

PR #2208, opened 2024-08-01
Fix/2207/undefined ui5 version

heimwege:
Fix for https://github.com/SAP/open-ux-tools/issues/2207

Please note: This PR introduces a central folder `preview-middleware-client/src/utils` where central re-use functions (like retrieving the ui5 version) can be placed. Things like e.g. [cpe/error-utils.ts](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/src/cpe/error-utils.ts) or [adp/ui5-version-utils.ts](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/src/adp/ui5-version-utils.ts) could be relocated to this folder at a later point in time as they are not solely used in cpe/apd.


packages/preview-middleware-client/src/utils/utils.ts
```diff
@@ -0,0 +1,13 @@
+import type {SingleVersionInfo} from '../../types/global';
+import VersionInfo from 'sap/ui/VersionInfo';
+
+/**
+ * Retrieve the UI5 version from sap.ui.core library
+ */
+export async function getUi5Version() {
+    const version = (await VersionInfo.load({library:'sap.ui.core'}) as SingleVersionInfo)?.version;
+    if (!version) {
+        throw new Error('Could not get UI5 version');
```
tobiasqueck, line 10: I am not sure we should break the preview when having trouble to read the version. Shouldn't we give a default value back and just log the error in the console? Something like 1.121.0 or so that would lead to the usage of the most common functionality.

heimwege: Fine for me as well. I was just not sure what a good default would be 🐱

tobiasqueck (approved):
* changeset ok
* code changes make it cleaner and less error prone
* existing unit tests prove that it did not introduce a regression

nikmace (commented):
Changes look good
Did not test manually

A couple of things:
- It seems like the utils.ts is not prettiered.
- There are currently no tests for the new directory `utils`.
- `utils/utils.ts` is too generic, everything can go inside and become a mess. Maybe it should be something like `utils/version.ts` (or something else for the name of the file), and an `index.ts` file that exports everything inside the `utils` folder?

nikmace (approved):
Review comments addressed
Tests added
Did not test manually

packages/preview-middleware-client/src/utils/version.ts
```diff
@@ -0,0 +1,15 @@
...
+import Log from 'sap/base/Log';
+
+/**
+ * Retrieve the UI5 version from sap.ui.core library
+ */
+export async function getUi5Version() {
+    let version = ((await VersionInfo.load({ library: 'sap.ui.core' })) as SingleVersionInfo)?.version;
+    if (!version) {
+        Log.error('Could not get UI5 version of application. Using 1.121.0 as fallback.');
+        version = '1.121.0';
```
voicis, line 12: Why do we set this particular version as fallback? We need to support versions from `1.71` and there are quite a few things that will not work if we assume it is `1.121`. If we can not determine the correct version then I suspect it will be quite likely that the following code will not work correctly, so I'm not convinced that it is worth continuing in such case.

packages/preview-middleware-client/src/utils/version.ts
```diff
@@ -0,0 +1,15 @@
+import type { SingleVersionInfo } from '../../types/global';
+import VersionInfo from 'sap/ui/VersionInfo';
+import Log from 'sap/base/Log';
+
+/**
+ * Retrieve the UI5 version from sap.ui.core library
+ */
+export async function getUi5Version() {
```
voicis, line 8: Could we return parsed version already? In many cases we need to know the minor version and then each time we parse it. Since we now have a common function, it would be convenient to have the parsing in the same place as well.

It could look something like this.
````typescript
interface UI5Version {
   full: string;
   major: number;
   minor: number;
   patch: number;
}

File: packages/preview-middleware-client/src/utils/version.ts (hunk 2 above)
Comment (heimwege) [reply packages/preview-middleware-client/src/utils/version.ts:12]: @tobiasqueck suggested to use that version as fallback instead of throwing an error. See [resolved discussion above](https://github.com/SAP/open-ux-tools/pull/2208#discussion_r1699871477)

File: packages/preview-middleware-client/src/utils/version.ts (hunk 3 above)
Comment (heimwege) [reply packages/preview-middleware-client/src/utils/version.ts:8]: I totally agree but I would not start the actual refactoring in this PR. This PR just wants to fix the undefined UI5 version handling bug.
I saw a parsing function in [adp/ui5-version-utils.ts](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/src/adp/ui5-version-utils.ts) and there are quite some other functions that can be relocated to general utils as well. I just think it's not the scope of this PR to do it.

File: packages/preview-middleware-client/src/adp/init.ts [hunk 4]
```diff
@@ -8,16 +8,14 @@ import {
...
     enableTelemetry
 } from '@sap-ux-private/control-property-editor-common';
-
 import { ActionHandler } from '../cpe/types';
-import VersionInfo from 'sap/ui/VersionInfo';
 import { getUI5VersionValidationMessage } from './ui5-version-utils';
 import UI5Element from 'sap/ui/dt/Element';
 import { getError } from '../cpe/error-utils';
-import type {SingleVersionInfo} from '../../types/global';
+import { getUi5Version } from '../utils';
```
voicis, line 15: This will result in runtime crash.

UI5 module loader can't handle implicit index files, so we need to explicitly add it in the module name or avoid the indirect import entirely.

```suggestion
import { getUi5Version } from '../utils/version';
```

voicis (changes requested):

packages/preview-middleware-client/src/utils/version.ts
voicis: ok, lets keep it likes this for now. But I think we should have another look at this later, perhaps when trying to unify the version handling logic. Regular preview can perhaps function without knowing exact version, but then it shouldn't use this API and use some other heuristics to check if the required APIs are there from UI5. 

But for adaptation editor it is more critical and without knowing UI5 version there I don't think it makes sense to continue. It is too likely that either the editor will crash anyways in a less graceful way or the editor will not work correctly.

packages/preview-middleware-client/src/utils/version.ts
voicis: ok, lets not get side tracked here. It seems that more changes would be required there to do this properly, lets have a separate PR for that.

packages/preview-middleware-client/src/adp/init.ts
heimwege: So basically you say what @nikmace suggested will not work, right? Ok then let's revert it 👍🏻

voicis (approved):
Thanks @heimwege!
- Review comments addressed
- Good code coverage
- Tested locally

nikmace (approved):
Re-approving, thanks

PR #2210, opened 2024-08-01
fix(odata-service-inquirer): fix adding backend system to prompt state

cianmSAP:
Fixes issue where the backend system is not added to the Prompt state when the default system name is used


longieirl (approved):
Changes look good.
Appropiate tests to cover changes
Changeset is good.

mikicvi-SAP (approved):
Changeset ✅
Code change is straight forward and easy to follow.
Test to cover the changes

PR #2215, opened 2024-08-02
Fix cap java jar loading

devinea:
- cds root is not always correctly set based on the option passed in. 
- Reload the cds env based on the CAP root. 
- help with loading cds dependencies from jar files for example.


packages/project-access/src/project/cap.ts
```diff
@@ -425,7 +426,9 @@ async function loadCdsModuleFromProject(capProjectPath: string, strict: boolean
     if (global) {
         global.cds = cds;
     }
-
+    // correct cds.env for current project root. Especially needed CAP Java projects loading cds dependency from jar file
+    const cdsEnv: any = cds.env.for('cds', capProjectPath);
+    cds.env = cdsEnv;
```
Klaus-Keller, line 431: small proposal to avoid `any`
```suggestion
    cds.env = cds.env.for('cds', capProjectPath) as typeof cds.env;
```

Klaus-Keller (commented):
Hi @devinea, thanks for the fix, small proposal from my side to avoid `any`.
Otherwise the fix is fine with me

packages/project-access/src/project/cap.ts
devinea: merged. Thanks

heimwege (dismissed):
excellent test coverage
change is straightforward
changeset is ok
did NOT test manually

packages/project-access/test/project/cap.test.ts
```diff
@@ -24,6 +24,17 @@ import type { Logger } from '@sap-ux/logger';
...
+    'for': jest.fn().mockImplementation(() => ({
+        'for': jest.fn().mockImplementation(() => ({
+            'for': jest.fn().mockImplementation(() => ({
+                folders: { app: 'MY_APP', db: 'MY_DB', srv: 'MY_SRV' }
+            })),
+            folders: { app: 'MY_APP', db: 'MY_DB', srv: 'MY_SRV' }
+        })),
+        folders: { app: 'MY_APP', db: 'MY_DB', srv: 'MY_SRV' }
+    }))
+};
```
Klaus-Keller, line 37: Hi @devinea, 
sorry, missed this in my first review. This 3 levels of `for` recursion is required because this is the max amount is it called? 

Here is a proposal that works independent of the number of calls:
```suggestion
const jestEnvForMock = jest.fn().mockImplementation(() => ({
    'for': jestEnvForMock,
    folders: {
        app: 'MY_APP',
        db: 'MY_DB',
        srv: 'MY_SRV'
    }
}));
const jestMockEnv = {
    'for': jestEnvForMock
};
```

Klaus-Keller (approved):
Thanks @devinea,
- code change to fix issue with @sap/cds looks good
- test coverage is great
- did a local test and could not find any issues

Approved from my side

heimwege (approved):

devinea:
Thanks @Klaus-Keller @heimwege

Issue #2216, opened 2024-08-02
TBI - Clean separation between prompting and writing in adp-tooling


tobiasqueck:
### Description (include screenshots)
The goal of the separation between the prompting and the writer modules is to be able to use and evolve them separately. Especially when creating complex prompting modules with multiple conditional branches that need to run in the YeomanUI the prompting output will be large flat structures. Such an output can map to one or more writers that need only a subset of the answers. 
To solve that we decided that the general flow is `prompt` -> `transform` -> `write`. The prompting results in usually a flat type named `xyzAnswers` and the writer requires a well structured `xyzConfig`. The mapping between both would happen in the consumer of both prompting and writing. 

The goal of this task is to rework the types and interfaces of https://github.com/SAP/open-ux-tools/tree/main/packages/adp-tooling/src/prompts and their corresponding writers.

### Value
Clean separation of concerns allows better reusability, testability and maintainability. 

### Architecture Elaboration
n/a

### Notes
#2184 was only approved with the agreement that this TBI will be implemented right after.

### Misc
@testojs @mmilko01 

### Tasks
- [ ] add-annotations-to-odata
- [ ] add-new-model
- [ ] change-data-source
- [ ] change-inbound
- [ ] add-component-usage

PR #2218, opened 2024-08-02
refactor: Clean separation between prompting and writing in adp-tooling

mmilko01:
https://github.com/SAP/open-ux-tools/issues/2216


tobiasqueck (dismissed):
* changeset ok
* refactored code nicely separates prompting from writing
* locally tested some of the touched functionality without problems

packages/adp-tooling/src/base/change-utils.ts
```diff
@@ -39,15 +39,15 @@ export function writeAnnotationChange(
     fs: Editor
 ): void {
     try {
-        const { fileName, answers } = data;
+        const { annotation } = data;
```
testojs, line 42: The `data` parameter of this function is of type `AnnotationsData`. This means that the function requires  also`variant` to be provided, but it is never used. 
Here we see that only the `annotation` object is used from the `data` parameter.

testojs (changes requested):

packages/adp-tooling/src/base/change-utils.ts
tobiasqueck: Changing the signature to:
```
export function writeAnnotationChange(
    projectPath: string,
    timestamp: number,
    { annotation }: AnnotationsData,
    change: ManifestChangeProperties,
    fs: Editor
): void {
```
should work (I think).

mmilko01: I tried to change the type and use `Pick` to get only the `annotation` part, but as the properties are nested, it became quite ugly, so I followed Tobias' advice :) 

Fixed with: eb8529ef6a8aa2dc6f1d959c41e796f5e2d56d85

testojs: But with this solution, the function still expects the whole object, including `variant`, to be provided.

testojs (approved):
* code looks good
* comments addressed
* changeset OK
* test coverage is good

packages/adp-tooling/src/base/change-utils.ts
tobiasqueck: Very nice.

tobiasqueck (approved):
reapproving

packages/adp-tooling/src/base/change-utils.ts
testojs: Great, thanks!

Issue #2220, opened 2024-08-05
TBI - Improve RTA command and flex change types


voicis:
### Description (include screenshots)
Currently the types we use for interacting with RTA commands and FlexObjects in FlexCommands does not match to the real objects in runtime, which can lead to wrong assumptions made by developer and resulting in runtime crashes like #2209.

We should update the type definitions to better reflect objects that are used in runtime.

### Value
Reduced chance of improper handling of runtime objects.

### Tasks
- [ ] Task 1

PR #2221, opened 2024-08-05
fix: add xml fragment

### LINKED ISSUE #2209: BUG - Add XML Fragment does not work if there is an unsaved "hideControl" change
### Description
A clear description of what the bug is.

### Steps to Reproduce
1. Open `Adaptation Editor` in a LROP app
2. Go to object page
3. Select one of the header facets
4. Open context menu and click on `Remove` button 
5. Try to add an XML fragment to any of the controls

### Expected results
It is possible to enter fragment name and click on `Create` button.

### Actual results
`Create` button is disabled and there is an error in console:
```
utils.js:39 Uncaught TypeError: command.getPreparedChange is not a function
    at matchesFragmentName (utils.js:39:28)
    at BaseDialog.controller.js:68:18
    at Array.some (<anonymous>)
    at _constructor._checkForExistingChange [as checkForExistingChange] (BaseDialog.controller.js:63:26)
    at _constructor._onFragmentNameInputChange [as onFragmentNameInputChange] (BaseDialog.controller.js:47:33)
    at r.fireEvent (EventProvider-dbg.js:240:38)
    at m.fireEvent (Element-dbg.js:798:44)
    at c.fireLiveChange (ManagedObjectMetadata-dbg.js:826:49)
    at U.oninput (Input-dbg.js:1822:8)
    at m._handleEvent (Element-dbg.js:361:10)
```

### Screenshots
image

OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

voicis:
Fixes #2209.

This issue is a symptom of a larger issue where we do not have correct types for RTA commands and flex objects. I started updating the types, but it turned out it would require more effort and we should not block this fix by that, so I made a separate TBI for that #2220.

### DIFF
--- a/.changeset/great-buttons-look.md
+++ b/.changeset/great-buttons-look.md
@@ -1,0 +1,6 @@
+---
+'@sap-ux-private/preview-middleware-client': patch
+'@sap-ux/preview-middleware': patch
+---
+
+Fixed Add XML Fragment dialog not working if there is an unsaved "hideControl" change
--- a/packages/preview-middleware-client/src/adp/controllers/BaseDialog.controller.ts
+++ b/packages/preview-middleware-client/src/adp/controllers/BaseDialog.controller.ts
@@ -126,7 +126,7 @@ export default abstract class BaseDialog<T extends BaseDialogModel = BaseDialogM
         const allCommands = this.rta.getCommandStack().getCommands();
 
         return allCommands.some((command: FlexCommand) => {
-            if (command?.getProperty('name') === 'composite') {
+            if (typeof command.getCommands === 'function') {
                 const addXmlCommand = command
                     .getCommands()
                     .find((c: FlexCommand) => c?.getProperty('name') === 'addXMLAtExtensionPoint');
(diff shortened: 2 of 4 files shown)


nikmace (commented):
I fail to understand what is the reason why this happens. Is it because the `hideControl` command does not have `getPreparedChange`? Or is it because the dialog expects a `composite` command and some `composite` command don't have `getCommands`?

The change looks simple, there are missing tests for that though. Will they be added later when the real solution to the bigger problem is implemented?

voicis:
The main issue was that the composite command check was not working correctly. There can be composite commands with a different name as is the case with `hideControl`.  The other change is to make sure that it is a flex command before calling `getPreparedChange`, because it would not exist on different changes.

As for tests I guess I can technically add them now, but such cases should be handled by the typescript compiler if we have correct types.

nikmace (approved):
Change looks good
Coverage is excellent
Did not test manually

heimwege (approved):
changes are straightforward
changeset is fine
extraordinary test coverage
did NOT test manually

PR #2230, opened 2024-08-07
feat(store): 2224 generator to fallback to application modeler keytar

korotkovao:
Fallback to loading keytar from the Application Modeler extension node_modules folder (e.g. <HOMEDIR>\.vscode\extensions\sapse.sap-ux-application-modeler-extension-1.14.2\node_modules\keytar) if it cannot be loaded.


packages/store/src/secure-store/index.ts
```diff
@@ -12,6 +15,23 @@ function getKeytar(log: Logger): typeof Keytar | undefined {
...
+        const keytarGlobVscode = join(
+            homedir(),
+            '.vscode/extensions/sapse.sap-ux-application-modeler-extension-**/node_modules/keytar/package.json'
+        );
+        const keytarGlobVscodeInsiders = join(
+            homedir(),
+            '.vscode-insiders/extensions/sapse.sap-ux-application-modeler-extension-**/node_modules/keytar/package.json'
+        );
+        const files = globSync([keytarGlobVscode, keytarGlobVscodeInsiders]);
+        console.log('files: \n' + JSON.stringify(files));
```
devinea, line 28: remove `console.log` or replace with `log.info`

packages/store/src/secure-store/index.ts
```diff
@@ -12,6 +15,23 @@ function getKeytar(log: Logger): typeof Keytar | undefined {
...
+        const dirnames: string[] = [];
+        files.forEach((filePath: string) => {
+            dirnames.push(dirname(filePath));
+        });
+        console.log('keytarDirectories: \n' + JSON.stringify(dirnames));
```
devinea, line 33: same

packages/store/src/secure-store/index.ts
```diff
@@ -12,6 +15,23 @@ function getKeytar(log: Logger): typeof Keytar | undefined {
         log.warn(errorString(err));
         log.warn(`Could not "require('keytar')". Trying VSCode's copy`);
```
devinea, line 16: This message should be below the new code.

packages/store/src/secure-store/index.ts
korotkovao: Thank you, applied ✅

packages/store/src/secure-store/index.ts
korotkovao: Thank you, applied ✅

packages/store/src/secure-store/index.ts
korotkovao: Thank you, applied ✅

devinea (commented):
Thanks for the updates. Code changes lg. 
Unit tests failing

packages/store/src/secure-store/index.ts
```diff
@@ -4,14 +4,44 @@ import { isAppStudio, errorString } from '../utils';
...
     } catch (err) {
+        try {
+            const files = globSync([keytarGlobVscode, keytarGlobVscodeInsiders]);
+            log.info('files: \n' + JSON.stringify(files));
+            const appModDirs: string[] = [];
+            files.forEach((filePath: string) => {
+                appModDirs.push(dirname(filePath));
+            });
+            log.info('keytarDirectories: \n' + JSON.stringify(appModDirs));
+            if (appModDirs.length >= 0) {
```
Klaus-Keller, line 32: If `appModDirs.length` is `0` `appModDirs[0]` is undefined.

Klaus-Keller (changes requested):
Are we sure we want to create this indirect dependency from `@sap-ux/store` to Visual Studio Code extension Application Modeler? Where is this consumed from?

**Proposal**: in module `@sap-ux/store` we allow to pass the path to a place where keytar module with binaries can be found. The consumer then needs to perform the search, but we would avoid having a node module searching for VSCode extension on the file system.

korotkovao:
Successfully tested locally by deleting the `node modules` `keytar`

packages/store/src/secure-store/index.ts
```diff
@@ -4,14 +4,59 @@ import { isAppStudio, errorString } from '../utils';
 import { DummyStore } from './dummy-store';
 import { KeytarStore } from './keytar-store';
 import type { SecureStore } from './types';
+import { join, dirname } from 'path';
+import { homedir } from 'os';
+import { default as fs } from 'fs';
+declare function __non_webpack_require__(m: string): any;
```
IainSAP, line 10: Is this correct?

IainSAP (commented):
Seems there is some odd looking webpack related coding here.

packages/store/src/secure-store/index.ts
korotkovao: Thanks for the review.
There was a bug without the webpack coding snd it's needed.

packages/store/src/secure-store/index.ts
```diff
@@ -4,14 +4,59 @@ import { isAppStudio, errorString } from '../utils';
...
+        // keytar is not installed or is removed from the fs by virus scanner.
+        try {
+            const AppMKeytarDirs = getKeytarPaths(false).concat(getKeytarPaths(true));
+            log.info('keytarDirectories: \n' + JSON.stringify(AppMKeytarDirs));
+            if (AppMKeytarDirs.length > 0) {
+                // try to load keytar from the first directory found
+                const keytarDir = AppMKeytarDirs[0];
+                log.info('Try to load keytar from :' + JSON.stringify(keytarDir));
+                // Support bundling
+                if (typeof __non_webpack_require__ === 'function') {
```
IainSAP, line 46: Wont this always be true since the function is declared above

packages/store/src/secure-store/index.ts
IainSAP: Can you please add a comment explaining why this is needed for others, thx/

packages/store/src/secure-store/index.ts
```diff
@@ -4,14 +4,59 @@ import { isAppStudio, errorString } from '../utils';
...
+import { default as fs } from 'fs';
+declare function __non_webpack_require__(m: string): any;
+
+function getKeytarPaths(insiders: boolean): string[] {
+    const vscodeRootPath = insiders ? '.vscode-insiders' : '.vscode';
+    const vscodeExtensionsPath = join(homedir(), vscodeRootPath, 'extensions/');
+    const AppMFoldersVscode =
+        fs
+            .readdirSync(vscodeExtensionsPath)
+            .filter((fn) => fn.startsWith('sapse.sap-ux-application-modeler-extension')) ?? [];
```
IainSAP, line 18: Should we not allow the consumer to pass the fallback extension rather than hard-coding?

IainSAP (commented):
See comments

packages/store/src/secure-store/index.ts
devinea: This is runtime type rather than typescript. So no, it is only true in bundle.

packages/store/src/secure-store/index.ts
devinea: We have various consumers e.g. Fiori generator, Adp generator and ui5 tooling modules that need this fallback in certain cases, hence the reason to hardcode it centrally. AppM hosts the SAP System view and should always be installed in Fiori tools setup.

packages/store/src/secure-store/index.ts
devinea: Comment added in f5398a5ad0f273602d9e27797454db4537804680

devinea (commented):
Tested in tools-suite on mac and windows and fallback works as expected.

packages/store/src/secure-store/index.ts
IainSAP: Agreed however, this is an internal tools suite extension id and as such may be subject to change. Would it not be more robust to allow the consumer to pass the extension id as the consumer will be able to determine that at runtime.

packages/store/src/secure-store/index.ts
IainSAP: I dont follow sorry, its defined here https://github.com/SAP/open-ux-tools/pull/2230/files#diff-b235f66bb5ed958a1d510e9bff09af9bc65446b6fe34d31815fcc09c818dd7aeR11 at the top of this file.

IainSAP: Thx

IainSAP (approved):
Thanks @korotkovao @devinea 

- Code is clear
- UT added to cover
- Comment about hardcoded tools suite extension id (it could be parameterised), but since this is a short term fix, OK for now.
- Not tested

longieirl (approved):
Changes look good.
All issues addressed.
No issues with code analysis.
Havent tested localy.

devinea (approved):
Changes lg. 
Tested locally on macOS and Windows in tools-suite

ullasholla:
It seems that @Klaus-Keller's questions weren't fully addressed:

> Are we sure we want to create this indirect dependency from @sap-ux/store to the Visual Studio Code extension, Application Modeler? Where is this being consumed from?
>
> Proposal: In the module @sap-ux/store, we could allow passing the path to a location where the keytar module with binaries can be found. The consumer would then be responsible for the search, avoiding the need for this low-level module @sap-ux/store to search for the VSCode extension on the file system.

Introducing this dependency couples the low-level library to the presence of the Application Modeler extension. Could you clarify where this is consumed? The pull request description is quite sparse and doesn't explain the motivation behind this change.

Additionally, this change seems to introduce a fallback to the Application Modeler's copy before attempting to use VSCode's copy. There may be valid reasons for this, but without further explanation, it's difficult to understand the rationale.

PR #2234, opened 2024-08-09
fix: enhance uilink and uncontextual menu components

Jimmy-Joseph19:
Contains enhancements for contextual menu and uilink


nahmed22 (approved):
Changes looks good. 
Changeset maintained. 
Tesst coverage is good.

PR #2236, opened 2024-08-12
Refactor usage of ui5 version

heimwege:
- move `adp/ui5-version-utils` to `utils/version`
- directly parse ui5 version in `utils/version -> getUi5Version`
- centralize checks for minimum ui5 version in `utils/version`
- future-proof ui5 minor version checks by adding `1` as major version


heimwege:
@voicis is this what you had in mind w.r.t. the ui5 version handling?
(There still need to be some adjustments because e.g. just checking if minor version is >72 is not hta future proof with ui5 2.x ahead)

voicis:
Yes, this is what I had in mind.

I also noticed that some of the checks now might have potential issues with the UI5 2.x. We should look into how to future proof the version checks as well.

heimwege:
@voicis done. Feel free to review 😸

packages/preview-middleware-client/src/flp/bootstrap.js
```diff
@@ -13,9 +11,9 @@ async function ushellBootstrap(fnCallback) {
     try {
         const response = await fetch('/resources/sap-ui-version.json');
         const json = await response.json();
-        const version = json?.libraries?.find((lib) => lib.name === 'sap.ui.core')?.version;
-        const major = version ? parseInt(version.split('.')[0], 10) : 2;
-        if (major >= 2) {
+        const version = json?.libraries?.find((lib) => lib.name === 'sap.ui.core')?.version ?? '1.121.0';
+        const [ majorUi5Version ] = version.replace(/snapshot-untested|snapshot-|snapshot/, '').split('.').map(Number);
```
voicis, line 15: I don't think we should limit what kind of suffixes the version can have, because if UI5 decides to add a new one it will break this logic. Currently we do not use the minor version, so corrupting that part would not cause an issue right now. 

Also `parseInt` seems to be better suited here, because it will only take the first numeric characters and ignore everything after the first non numeric character. Which means we do not need to do replacement then either.

packages/preview-middleware-client/src/utils/version.ts
```diff
@@ -1,15 +1,75 @@
...
+ * @returns Ui5VersionInfo
  */
 export async function getUi5Version() {
     let version = ((await VersionInfo.load({ library: 'sap.ui.core' })) as SingleVersionInfo)?.version;
     if (!version) {
         Log.error('Could not get UI5 version of application. Using 1.121.0 as fallback.');
         version = '1.121.0';
     }
-    return version;
+    const [major, minor] = version.replace(/snapshot-untested|snapshot-|snapshot/, '').split('.').map(Number);
```
voicis, line 35: Same as the comment above regarding replacement and converting to number.

packages/preview-middleware-client/src/flp/bootstrap.js
heimwege: I basically copied that suffix handling from the adp/ui5-version-utils. But you are right. I adjusted it.

packages/preview-middleware-client/src/utils/version.ts
heimwege: done

packages/preview-middleware-client/src/adp/init.ts
```diff
@@ -44,21 +45,20 @@ export default async function (rta: RuntimeAuthoring) {
...
     );
 
-    const syncViewsIds = await getAllSyncViewsIds(minor);
-    initDialogs(rta, syncViewsIds, minor);
+    const ui5VersionInfo = await getUi5Version();
+    const syncViewsIds = await getAllSyncViewsIds(ui5VersionInfo);
+    initDialogs(rta, syncViewsIds, ui5VersionInfo);
 
-    if (minor > 77) {
+    if (!isLowerThanMinimalUi5Version(ui5VersionInfo, { majorUi5Version: 1, minorUi5Version: 77 })) {
```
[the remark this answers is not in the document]
heimwege: fixed

voicis (dismissed):
- code adheres to coding standards
- review comments addressed
- good code coverage
- did not test locally

heimwege:
@voicis I needed to solve some merge conflicts from #2183 and adjust tests. Please re-review.

packages/preview-middleware-client/src/utils/version.ts
```diff
@@ -50,7 +50,7 @@ export async function getUi5Version() {
 export function isLowerThanMinimalUi5Version(
     ui5VersionInfo: Ui5VersionInfo,
     minUi5VersionInfo: Ui5VersionInfo = minVersionInfo
-): boolean {
+) {
```
voicis, line 53: Why the return type was removed from here?

packages/preview-middleware-client/src/utils/version.ts
```diff
@@ -70,6 +70,6 @@ export function isLowerThanMinimalUi5Version(
  * @param ui5VersionInfo to be mentioned in the message
  * @returns string with validation message.
  */
-export function getUI5VersionValidationMessage(ui5VersionInfo: Ui5VersionInfo): string {
+export function getUI5VersionValidationMessage(ui5VersionInfo: Ui5VersionInfo) {
```
voicis, line 73: Why the return type was removed from here?

heimwege: Because TypeScript infers the return type anyway. I thought this is a general TS feature and not IDE specific. At least in my IDE `:string` is superfluous. If that assumption proves wrong I can of course add it again

packages/preview-middleware-client/src/utils/version.ts
heimwege: Because TypeScript infers the return type anyway. I thought this is a general TS feature and not IDE specific. At least in my IDE `:boolean` is superfluous. If that assumption proves wrong I can of course add it again

packages/preview-middleware-client/src/utils/version.ts
tobiasqueck: We don't have a clear guideline for that in this repo (I think). Having the type there has the advantage that when you change the code and accidentally forgot to return something in an branch then your IDE will show an error directly here, otherwise, it would show errors where it is consumed e.g. if an undefined check is missing.

I prefer having return value explicitly typed because it helps people that don't know the code that well but, again, it is not a rule in this repo (I think).

voicis: The reason for bringing this up is that there is an [ESLint rule](https://typescript-eslint.io/rules/explicit-function-return-type/) for this and somewhat recently in one of the repositories complaining about this. It looks like it wasn't this one or it is no longer the case here, because I don't see the warnings anymore.

heimwege: Ok, then I'll add it again 👍🏻

packages/preview-middleware-client/src/utils/version.ts
heimwege: Added missing return type

voicis (approved):
Review comments addressed and changes since the last approval looks good.

heimwege:
I'm sorry that I dismissed your review again @voicis 🙈 But @tobiasqueck did not like the longish Ui5VersionInfo type property names.

tobiasqueck (approved):
* changeset ok
* refactoring clearly improves the code quality
* thanks for proactively getting the code ready for UI5 2.x
* did not manually test, just reviewed

voicis (approved):

packages/preview-middleware-client/src/utils/version.ts
tobiasqueck: #2252 to have guidance in the future

PR #2240, opened 2024-08-14
Fix(fpm-writer): change building block required library to sap.fe.core

AlinaGovoruhina:
Issue #2241 

When adding a building block validate minimum required `sap.fe.core` library in manifest.json, not `sap.fe.templates`.

### DIFF
--- a/.changeset/selfish-forks-relax.md
+++ b/.changeset/selfish-forks-relax.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fe-fpm-writer': minor
+---
+
+Update fpm-writer to validate sap.fe.core as minimum library required for building block creation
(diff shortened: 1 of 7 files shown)


815are (commented):
I think we should rename test `'validate sap.fe.templates manifest dependency'`?
https://github.com/SAP/open-ux-tools/blob/fix/building-block-required-library/packages/fe-fpm-writer/test/unit/building-block.test.ts#L75

AlinaGovoruhina:
> I think we should rename test `'validate sap.fe.templates manifest dependency'`? https://github.com/SAP/open-ux-tools/blob/fix/building-block-required-library/packages/fe-fpm-writer/test/unit/building-block.test.ts#L75

Thanks, updated that

815are (approved):
checked:
1. code easy to understand and does not provide regression
2. test project adjusted to fulfil new expectation
3. changeset persists

devinea (approved):
Changes look good. 
I assume there is no need to accept `sap.fe.templates` as fallback if `sap.fe.core` is missing.

PR #2242, opened 2024-08-14
test: simplify virtual fs logic

voicis:
CDS layering test started failing after dce8838a2dcd93fc9922ade4e2a8118f9d950f9a. It was caused due to some changes in `node_modules`. However, since we're now using `mem-fs-editor` in the tests we can remove the logic that pre-fills file caches and just use the `Editor` object.

Adding changeset from #2230, since it failed to publish.

### DIFF
--- a/.changeset/tasty-plants-kick.md
+++ b/.changeset/tasty-plants-kick.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/store': minor
+---
+
+Load keytar from application modeler extension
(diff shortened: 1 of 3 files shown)


marufrasully (approved):
@voicis Thanks. 
Simple and straightforward code change.

.changeset/tasty-plants-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/store': minor
+---
```
devinea, line 3: Thanks for adding this to fix failed release

devinea (approved):

vadson71 (approved):
Changes look reasonable

nahmed22 (approved):
LG, changeset maintained

Issue #2238, opened 2024-08-13
BUG - I18n entries are written directly below app root folder, reference in `manifest.json` does not work


johannes-kolbe:
### Description
1. Currently the `project-access` implementation of [`createUI5I18nEntriesBase`](https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/project/i18n/write.ts#L42) uses a default path `i18n/i18n.properties` directly below a project's root to write newly created `i18n.properties` files. 
This leads to issues during preview and generation of new i18n files, since the references (same default path) are in `manifest.json`, which usually is within a `webapp` folder, and thus cannot correctly resolve the path.

```
app root level
| i18n (newly created)
| | - i18n.properties
| webapp
| | manifest.json
| | - i18n: `i18n/i18n.properties`
```

2. In coordination with @marufrasully: not an issue, default path stays as `i18n/i18n.properties`
    ~~Another issue is with the `defaultPath` itself. It is not uncommon to have separate folders for e.g. `app`, `model` and `annotation` i18n properties. Thus the option to define a path for creation should be available for consumers such as [`createUI5I18nEntries`](https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/project/i18n/write.ts#L112) and [`createAnnotationI18nEntries`](https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/project/i18n/write.ts#L146).~~

### Steps to Reproduce
1. Create `i18n` entries with [`createUI5I18nEntries`](https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/project/i18n/write.ts#L112) or [`createAnnotationI18nEntries`](https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/project/i18n/write.ts#L146).
2. See that folder is created directly below root level.
4. Start preview of app
5. See that i18n entries are not resolved

### Expected results
- `i18n.properties` file is written in context of the `manifest.json`
- ~~path to `i18n.properties` file can be adjusted via caller~~

### Actual results
- `.i18n.properties` file is written below root level, no default access via `manifest.json`
- ~~no way to define path to `i18n` folder~~

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

[...]

PR #2244, opened 2024-08-15
fix(project-access): create `i18n` under `webapp` path

johannes-kolbe:
#2238

Fixes issue with creation of new `i18n` entries on root level of app:
- Uses `manifestPath` as root path for i18n folder

**Removed:**
~~Provides optional folder path for i18n creation: Default for annotation creation is `annotation`. E.g. `app/webapp/i18n/annotation/i18n.properties`~~


packages/project-access/src/project/i18n/write.ts
```diff
@@ -45,9 +46,10 @@ async function createUI5I18nEntriesBase(
     i18nPropertiesPaths: I18nPropertiesPaths,
     newEntries: NewI18nEntry[],
     modelKey: string,
-    fs?: Editor
+    fs?: Editor,
+    i18nFolder = ''
 ): Promise<boolean> {
-    const defaultPath = 'i18n/i18n.properties';
+    const defaultPath = join('i18n', i18nFolder, 'i18n.properties');
```
voicis, line 52: This is also used as the value which is added to the manifest and we should not have windows delimiters there. I suspect this would cause issues when trying to load the resources at the runtime.

johannes-kolbe: Removed after call with @marufrasully

marufrasully (approved):
Thanks @johannes-kolbe LGTM
* Changes are simple and straightforward. 
* Secured with unit tests

voicis (approved):
- Changeset present
- Code adheres to the coding standards
- Good test coverage
- Did not test locally

devinea (approved):
changes look good. 
Changset ✅

PR #2246, opened 2024-08-16
export getFioriToolsDirectory, FioriToolsSettings from store

devinea:
allow getFioriToolsDirectory, FioriToolsSettings to be used by other modules in place of common utils in tools suite


mikicvi-SAP (approved):
Changeset ✅
Small change - export looks good

longieirl (approved):
Changes look good.

Issue #2250, opened 2024-08-19
BUG - fpm-writer. Building block Chart - create 'contextPath' when 'BuildingBlockMetaPath' object is received for property 'metaPath'


815are:
### Description
We noticed that following example does not workjfor Chart:
```
<macros:Chart
    id="Chart"
    metaPath="/SessionMetrics/@com.sap.vocabularies.UI.v1.Chart#chartMacro1"
/>
```
but works following:
```
<macros:Chart
    id="Chart"
    metaPath="@com.sap.vocabularies.UI.v1.Chart#chartMacro1"
    contextPath="/SessionMetrics"
/>
```

Documentation -> https://help.sap.com/docs/ABAP_PLATFORM_NEW/468a97775123488ab3345a0c48cadd8f/52d065ab8bb740c58c834a0c985e3b9e.html

As we tested - it is only relevant to charts

### Steps to Reproduce
1. Pass following object to `generateBuildingBlock`:
```
{
    "buildingBlockData": {
        "buildingBlockType": "chart",
        "metaPath": {
            "bindingContextType": "absolute",
            "entitySet": "GT7Service.SessionMetrics",
            "qualifier": "@com.sap.vocabularies.UI.v1.Chart#chartMacro1"
        }
    },
}
```
2. Result would be:
```
<macros:Chart
    id="Chart"
    metaPath="/SessionMetrics/@com.sap.vocabularies.UI.v1.Chart#chartMacro1"
/>
```
which does not work in runtime

### Expected results
```
<macros:Chart
    id="Chart"
    metaPath="/SessionMetrics/@com.sap.vocabularies.UI.v1.Chart#chartMacro1"
/>
```

### Actual results
```
<macros:Chart
    id="Chart"
    metaPath="/SessionMetrics/@com.sap.vocabularies.UI.v1.Chart#chartMacro1"
/>
```

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

PR #2251, opened 2024-08-19
fix(fpm-writer): Building block Chart - create 'contextPath' when 'BuildingBlockMetaPath' object is received for property 'metaPath'

815are:
Issue #2250

We noticed that following example does not work for Chart:
```
<macros:Chart
    id="Chart"
    metaPath="/SessionMetrics/@com.sap.vocabularies.UI.v1.Chart#chartMacro1"
/>
```
but works following:
```
<macros:Chart
    id="Chart"
    metaPath="@com.sap.vocabularies.UI.v1.Chart#chartMacro1"
    contextPath="/SessionMetrics"
/>
```

As solution - specifically for Chart, we convert `BuildingBlockMetaPath` into `metaPath` and `contextPath` attributes instead of single `metaPath` attribute.


AlinaGovoruhina (approved):
Changes to use contextPath for chart building block creation is well done, check for absolute or relative path is preserved, unit test coverage good.

devinea (approved):
Code changes are clear
Changeset 👍

PR #2254, opened 2024-08-20
Show binding context question for filterbar building block

815are:
As requested - show question for binding context selection for filter bar same as it is for table and chart:
image


AlinaGovoruhina (approved):
Binding context type question added to Filter bar building block, unit tests updated accordingly

devinea (approved):
code changes and snapshots look ok.
changeset ✅

PR #2256, opened 2024-08-20
fix(ui-prompting): remove left margin when in multi-columns view

AlinaGovoruhina:
Remove left margin from .prompt-entries class elements


815are (approved):
checked:
1. style update - checked with storybook(examples/prompting-ui)
2. changelog presented
3. no test as only style update

broksy (approved):
Changeset is clear and matches respective changes.
Approved from my side.

PR #2257, opened 2024-08-20
fix: regression with 'UIDefaultButton' - expand icon is applied even there no submenu

815are:
regression after - https://github.com/SAP/open-ux-tools/pull/2234
we missed that by applying
```
menuIconProps: {
                    iconName: UiIcons.ArrowDown
                }
```
then icon would be applied for button even if there no contextual submenu enabled.
image

As fix - applying default expand icon only when submenu items are passed for button.


voicis (approved):
Expand icon no longer is displayed by default for UIDefaultButton.

Jimmy-Joseph19 (approved):
Changes looks good.
Changeset found.
Tested with storybook.

Issue #2225, opened 2024-08-06
BUG - @sap-ux/ui5-proxy-middleware directLoad not working


zinserjan:
### Related Feature
Feature request: -

### Description
- `directLoad: true` does not work and fails with Error: `EISDIR: illegal operation on a directory, read`

### Steps to Reproduce
1. use `directLoad: true` in @sap-ux/ui5-proxy-middleware
2. Open /index.html or /test/flpSandbox.html
3. See error

### Expected results
- no Error
- URLs are replaced

### Actual results
- Error: `EISDIR: illegal operation on a directory, read`

#### Version/Components/Environment	
Add any other context about the problem here

OS:                      
- [ ] Mac OS
- [x] Windows
- [ ] Other

Node.js v20.16.0

Packages:
- @ui5/cli v4.0.3
- @sap-ux/ui5-proxy-middleware v1.4.3

### Problem
- `req.baseUrl` is here always an empty string https://github.com/SAP/open-ux-tools/blob/c0f92980c658ad2fc9f28de75b3e778d5fa36559/packages/ui5-proxy-middleware/src/base/utils.ts#L275-L280
- `htmlFilePath` will be just the webapp folder instead of the path to the requested HTML file
- that's why the further logic doesn't work and instead of replacing the URL in `injectUI5Url` it tries to read the webapp directory

### Fix
- just use `reg.url` or` reg.originalUrl`

tobiasqueck:
@cfg74 could you please plan to get this one fixed.

PR #2258, opened 2024-08-20
fix(ui5-proxy-middleware): request url for directLoad

ytpo-lyne:
Fix for: #2225
`directLoad: true` does not work and fails with `Error: EISDIR: illegal operation on a directory, read`. 

Changes:
- replaced `req.baseUrl` object with `req.url`
- added check for realpath to fix code scanning complain


packages/ui5-proxy-middleware/src/base/utils.ts
```diff
@@ -272,13 +272,13 @@ export const injectScripts = async (
...
-        const htmlFileName = getHtmlFile(req.baseUrl);
+        const htmlFileName = getHtmlFile(req.url);
         const yamlFileName = getYamlFile(args);
         const ui5YamlPath = join(projectRoot, yamlFileName);
         const webAppFolder = await getWebAppFolderFromYaml(ui5YamlPath);
-        const htmlFilePath = join(projectRoot, webAppFolder, htmlFileName);
-        const html = injectUI5Url(htmlFilePath, ui5Configs);
-
+        let htmlFilePath = join(projectRoot, webAppFolder, htmlFileName);
+        htmlFilePath = realpathSync(resolve(projectRoot, htmlFilePath));
```
tobiasqueck, line 280: could you please make htmlFilePath a const and put this in one line. It is odd to calculate it and the use the same variable again for a 2nd calculation. So, please const and one line or two lines but with two different variables.

tobiasqueck: btw, are you sure that works? You add the`projectRoot in the join and then again in the resolve. To be honest, the whole thing looks a little odd. Could you describe what the problem is and how you solve it?

ytpo-lyne: Done db1f62760fa73d569722c3bdb376383a43a867bc

packages/ui5-proxy-middleware/test/base/utils.test.ts
```diff
@@ -465,8 +468,10 @@ describe('utils', () => {
         });
 
         test('HTML is modified and response is sent', async () => {
+            const projectRoot = process.cwd();
```
tobiasqueck, line 471: The test is not very helpful since you mock the file access. Please add a folder with a sample project structure under test/fixtures and then use that for testing instead of mocking fs.

packages/ui5-proxy-middleware/src/base/utils.ts
ytpo-lyne: The problem was detected by the code scanning check: https://github.com/SAP/open-ux-tools/security/code-scanning/175 Followed the proposed solution to validate the file path

packages/ui5-proxy-middleware/test/base/utils.test.ts
ytpo-lyne: Done with c0bdb3cbd8e4eeb655a2256e312f16db7ccefa37

ytpo-lyne:
Replaced by https://github.com/SAP/open-ux-tools/pull/2319

PR #2259, opened 2024-08-20
fix(cpe): copy to clipboard

voicis:
Fixed copy to clipboard caching old values.

Also fixed sonar issues in the file.


vadson71 (approved):
Changes look good, addressed in unit tests

Jimmy-Joseph19 (approved):
Changes looks good.
Changeset found.
Covered with unit test.
Approved.

PR #2260, opened 2024-08-21
fix(fe-fpm-writer): Building block Prompts. Use EntitySets instead of EntityTypes for displaying choices and writing metaPath or contextPath

815are:
We noticed that our generated prompts generates building blocks with paths like
```
<macros:Table id="Table" metaPath="/BookingType/@com.sap.vocabularies.UI.v1.LineItem#tableMacro" />
```
which does not work in fe runtime - it expects `EntitySets` instead of `EntityTypes`

Adjusted code to use show `EntitySets` in choices and also in result `EntitySets` is used when building block is generated for xml file
 image


packages/fe-fpm-writer/test/unit/building-block/prompts/utils/service.test.ts
```diff
@@ -51,32 +51,29 @@ describe('utils - service', () => {
...
+                                fullyQualifiedName: 'IncidentService.EntityContainer/IncidentProcessTimeline'
                             }
                         ]
                     }
                 })
             } as unknown as FioriAnnotationService);
-            const entityTypes = await getEntityTypes(capProject, capAppFolder);
+            const entityTypes = await getEntitySets(capProject, capAppFolder);
             expect(entityTypes.length).toEqual(3);
             expect(entityTypes.map((entityType) => entityType.fullyQualifiedName)).toEqual([
```
AlinaGovoruhina, line 73:
```suggestion
            const entitySets = await getEntitySets(capProject, capAppFolder);
            expect(entitySets.length).toEqual(3);
            expect(entitySets.map((entitySet) => entitySet.fullyQualifiedName)).toEqual([
```

815are: thanks -> 936865b4666fda99ef6696267fd82530473786ea

AlinaGovoruhina (approved):
EntityType prompt changed to entitySet: in question choices and metaPath, tests updated accordingly.

devinea (approved):
changeset ✅ 
I didn't test locally.

PR #2266, opened 2024-08-21
fix(security): update axios overrides

donal-tobin-sap:
- remove all overrides as no longer required
- update modules to get axios updates

Issue: tools-suite/issues/30924
Issue: https://github.com/SAP/open-ux-tools/security/dependabot/118


tobiasqueck (approved):

cianmSAP (approved):
Updates look good

Issue #2270, opened 2024-08-22
BUG - `axios-extension` and `store` are logging at too high log levels polluting CLI prompting


IainSAP:
### Related Feature
Feature request: _issue number_

### Description

The log levels being used are too high, and should be debug/trace for low level modules. Axios extension does not allow the log level or logger to be configured as a consumer, so any CLI consumers are getting uncontrolled log statements confusing users. In the case of `store` the shown error logs are not errors as the end user has declined access and so there should be no logs as there are no credentials to parse, how could that be an error!

<img width="751" alt="image" src="https://github.com/user-attachments/assets/3e19e4ae-58ef-4d82-9818-7058f207aa1b">

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Related issue https://github.com/SAP/open-ux-tools/issues/3193

PR #2271, opened 2024-08-22
fix(axios-extension): Change log levels to debug so they dont pollute CLIs.

IainSAP:
https://github.com/SAP/open-ux-tools/issues/2270

Fix for CLI pollution (axios-extension):

image

- Change info to debug as info is not appropriate for a low level networking module that is consumed by CLI prompts, and there is no way to change the default level or disable.

### DIFF
--- a/packages/axios-extension/src/abap/message.ts
+++ b/packages/axios-extension/src/abap/message.ts
@@ -51,7 +51,7 @@ function logLevel(severity: string, msg: string, log: Logger, error = false): vo
     if (severity) {
         severity = severity.toLowerCase();
         if (severity === 'success') {
-            log.info(msg);
+            log.debug(msg);
         } else {
             if (severity === 'warning') {
                 severity = 'warn';
@@ -59,7 +59,7 @@ function logLevel(severity: string, msg: string, log: Logger, error = false): vo
             log[severity](msg);
         }
     } else {
-        error ? log.error(msg) : log.info(msg);
+        error ? log.error(msg) : log.debug(msg);
     }
 }
 
@@ -85,7 +85,7 @@ export function prettyPrintMessage({
 }): void {
     try {
         const jsonMsg = JSON.parse(msg) as SuccessMessage;
-        log.info(jsonMsg.message);
+        log.debug(jsonMsg.message);
         logFullURL({ host, path: jsonMsg['longtext_url'], log, isDest });
         if (jsonMsg.details) {
             jsonMsg.details.forEach((entry) => {
@@ -120,13 +120,13 @@ function logFullURL({
         const base = new URL(host).origin; // We only care for the origin value
         // Add this instruction to the user because of this bug in VS Code: https://github.com/microsoft/vscode/issues/144898
         // It undoes the encoding of special characters in the URL sent to the browser
-        log.info('Please copy/paste this URL in a browser for more details:');
+        log.debug('Please copy/paste this URL in a browser for more details:');
         if (isDest) {
-            log.info(
+            log.debug(
                 '(Note: You will need to replace the host in the URL with the internal host, if your destination is configured using an On-Premise SAP Cloud Connector)'
             );
         }
-        log.info(new URL(path, base).href);
+        log.debug(new URL(path, base).href);
     }
 }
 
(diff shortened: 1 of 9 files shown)


packages/axios-extension/src/auth/index.ts
```diff
@@ -97,7 +97,7 @@ export function getReentranceTicketAuthInterceptor({
         if (apiUrl && apiUrl != provider.defaults.baseURL) {
             // Reentrance tickets work with API hostnames. If the original URL was not one, this will replace it
             // with the API hostname returned
-            provider.log.warn(
+            provider.log.debug(
```
IainSAP, line 100: Should not be a warning, the end user cannot do anything about this!

packages/axios-extension/src/abap/message.ts
```diff
@@ -120,13 +120,13 @@ function logFullURL({
         const base = new URL(host).origin; // We only care for the origin value
         // Add this instruction to the user because of this bug in VS Code: https://github.com/microsoft/vscode/issues/144898
         // It undoes the encoding of special characters in the URL sent to the browser
-        log.info('Please copy/paste this URL in a browser for more details:');
+        log.debug('Please copy/paste this URL in a browser for more details:');
```
IainSAP, line 123: Should not be end user messaging in a low level module. The consumer needs to deal with this directly based on a returned message.

IainSAP:
The correct approach is to support log configuration (log level, or a new logger instance) so consumers can set the log level to off or whatever is appropriate for their usage. But for now we need to quickly hide them.

cianmSAP (approved):
Thanks @IainSAP 

Changes look good
Tests updated
Changeset ✅

cianmSAP (approved):
Re-approving

packages/axios-extension/src/abap/message.ts
devinea: Makes sense. But better to keep this as info for now until that is addressed?

packages/axios-extension/src/abap/message.ts
```diff
@@ -120,13 +120,13 @@ function logFullURL({
...
         if (isDest) {
-            log.info(
+            log.debug(
```
devinea, line 125: same

packages/axios-extension/src/abap/message.ts
```diff
@@ -120,13 +120,13 @@ function logFullURL({
...
                 '(Note: You will need to replace the host in the URL with the internal host, if your destination is configured using an On-Premise SAP Cloud Connector)'
             );
         }
-        log.info(new URL(path, base).href);
+        log.debug(new URL(path, base).href);
```
devinea, line 129: same

packages/axios-extension/src/abap/ui5-abap-repository-service.ts
```diff
@@ -207,7 +207,7 @@ export class Ui5AbapRepositoryService extends ODataService {
                 const query = this.defaults.params?.['sap-client']
                     ? '?sap-client=' + this.defaults.params['sap-client']
                     : '';
-                this.log.info(`App available at ${frontendUrl}${path}${query}`);
+                this.log.debug(`App available at ${frontendUrl}${path}${query}`);
```
devinea, line 210: same

packages/axios-extension/src/abap/message.ts
```diff
@@ -85,7 +85,7 @@ export function prettyPrintMessage({
 }): void {
     try {
         const jsonMsg = JSON.parse(msg) as SuccessMessage;
-        log.info(jsonMsg.message);
+        log.debug(jsonMsg.message);
```
devinea, line 88: This is the message from the server to show to the user? Should still be info then.

packages/axios-extension/src/abap/ui5-abap-repository-service.ts
```diff
@@ -381,7 +381,7 @@ export class Ui5AbapRepositoryService extends ODataService {
                 // We've nothing to return as we dont want to show the exception to the user!
                 return Promise.resolve(undefined);
             } else {
-                this.log.info(`${appName} found on target system: ${isExisting}`);
+                this.log.debug(`${appName} found on target system: ${isExisting}`);
```
devinea, line 384: same

packages/axios-extension/src/abap/message.ts
IainSAP: End users do not use this module directly and so all messages should be returned for display in the appropriate way depending on the UI environment. Im curious how an end user would ever see these messages from YUI or VSIX?

Issue #2268, opened 2024-08-22
BUG - URI validation not executed for Replace OData prompting


mmilko01:
### Description
OData URI prompt doesn't validate the input for valid URI

### Steps to Reproduce
1. Create an Adaptation Project
2. Start `Replace OData Service` generator
3. Input invalid URI for `OData Source URI` or `OData Annotation Data Source URI` (e.g., `/sap /test`)
4. The value is allowed

### Expected results
The value is rejected

### Actual results
The value is allowed

### Screenshots
![image](https://github.com/user-attachments/assets/5615eeec-6e5d-4bfe-9e3b-c324f03f6d38)

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2273, opened 2024-08-22
fix: URI validation not executed for Replace OData prompting

mmilko01:
Fix for https://github.com/SAP/open-ux-tools/issues/2268.

- Adds URI validations for Replace OData prompting.


nikmace (approved):
Code looks good
Test coverage is excellent
Did not test manually

packages/adp-tooling/src/prompts/change-data-source/index.ts
```diff
@@ -3,7 +3,45 @@ import type { ManifestNamespace } from '@sap-ux/project-access';
...
+    return true;
+}
+
+/**
+ * Validates the Annotation URI prompt.
+ *
+ * @param value The value to validate.
+ * @returns {boolean | string} True if the value is a valid URI or empty, or an error message if not a valid URI.
+ */
+function validatePromptAnnotationURI(value: string): boolean | string {
```
longieirl, line 33: Would it make sense to try to join these two methods as the only difference is if `validationResult` is a string or a boolean?

mmilko01: I see your point, but I didn't see a nice way to merge to the functions and not make the code uglier. I would either have to pass a parameter to know for which prompt the new functions is executing or have some logic inline in the prompt to parse the result. In both cases the I wouldn't have the nice looking passing of the validator functions directly in the prompts as it currently is. If you have an idea that I don't currently see, please suggest :)

longieirl: Happy with explanation to cover the existing implementation. Thanks.

longieirl (approved):
Changes look good.
Comments addressed.
Tests updated.
Changeset present.

PR #2275, opened 2024-08-22
FEAT - Move Guided Help code into a new module.

devinea:
FEAT - Move Guided Help code into a new module to make easier to consume in modules outside the generators
replacing common-utils


donal-tobin-sap (approved):
Looks fine to me for module changes

Issue #2128, opened 2024-07-05
FEATURE - ABAP deploy configuration inquirer 


cianmSAP:
_As a developer,
I want to prompt a user for abap deploy configuration,
so that I updated and create the files as required for ABAP deployment_

### Description
The new inquirer module will provide the user with the prompts to gather the data for creating ABAP deployment configuration.

### Technical Design
`@sap-ux/abap-deploy-config-inquirer`
Consumed as part of a yeoman generator

### Acceptance Criteria
**Given** I have a yeoman generator (or similar inquirer module)
**when** I execute getPrompts
**then** it will return the prompt

### Notes
`@sap-ux/create` needs to be enhanced also to incorporate the new functionality 

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

docirl:
Please include the following change to deploy config prompting for abap for package/transport prompts:

Context
- the existing logic for 'validating' the transport request prompt (manual transport input) checked if the package provided started with $ L l T t and if so assumed the package was 'local' therefore not requiring a transport to be provided. The validation would return true for transport request manual input prompt. This is not correct, and as in the ui service generator, the package should be checked in the backend system
- the transport request ADT api (for 'select existing transport from list') when called with a package that is 'local' will return a warning text to state the package is local and no transport request is required, instead of returning a list of transports. We can use this response to show/hide the transport prompts. I asked ADT colleagues if other api's or methods were available to check the local status, but this is the same basic method they use also.

Summary of changes
- dont use static check of package name to validate transport request prompt
- check if package provided is 'local' from the backend. Use [ transport check adt api](https://github.com/SAP/open-ux-tools/blob/64f9c513bccb4b5580ed2cb7391b94d5157cd745/packages/axios-extension/src/abap/adt-catalog/services/transportcheck-service.ts#L38) as used already when retrieving the list of transport requests
- if package is local, hide transport request prompts

If anything above is unclear let me know.

cianmSAP:
Closing as inquirer is published

PR #2276, opened 2024-08-23
fix(abap-deploy-config): add autocomplete option and tidy add return types

cianmSAP:
#2128

- Removes use of feature toggle and adds `useAutocomplete` as an option
- Add return type to functions to fix lint warnings


longieirl (dismissed):
Changes look good.
Good test coverage.

longieirl (approved):
re-approving.

Issue #2241, opened 2024-08-14
BUG - Update fpm-writer to support minimum library required for building block guides i.e. sap.fe.core


hitesh-parmar:
### Related Feature
Feature request: _issue number_

### Description

Currently, when trying to add a macro (executing a building block guide) using Fiori tools Guided Development extension, an error is thrown in case `sap.fe.templates` is missing in the manifest.json . This error is incorrect and comes from `fe-fpm-writer`.
The minimum requirement as of today for supporting macros is `sap.fe.core` library.

Application generator is going to stop generating `sap.fe.templates` as a library  and hence we also need to fix this in `fpm-writer` so as new apps do not throw an error.

### Steps to Reproduce

1. Open any FE V4 project
1. Open Guided Development extension
1. Open manifest.json and remove `sap.fe.templates` from libs section
1. Open the _Add a table building block_
1. When you click on Insert snippet an error will be thrown complaining about missing dependency to `sap.fe.templates` which is wrong

### Expected results

An error is thrown if `sap.fe.core` is missing in the libs section of manifest.json

### Actual results

An error is thrown if `sap.fe.templates` is missing in the libs section of manifest.json

### Screenshots

![image](https://github.com/user-attachments/assets/2a6bdbf0-f660-4c7e-a040-e587b269e186)

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2284, opened 2024-08-27
fix(fe-fpm-writer): validate either "sap.fe.core" or "sap.fe.templates" library when creating a building block

AlinaGovoruhina:
Follow up PR for https://github.com/SAP/open-ux-tools/issues/2241

Throw error when both  "sap.fe.core" and "sap.fe.templates" libraries are missing when creating a building block.
Do not throw error if at least one library is present.
Adding missing libraries support will be added as a follow up.


packages/fe-fpm-writer/src/common/validate.ts
```diff
@@ -58,3 +59,29 @@ export function validateBasePath(basePath: string, fs?: Editor, dependencies = [
...
 }
+
+/**
+ * Validates base path, checks and adds missing libraries.
+ *
+ * @param {string} basePath - the base path
+ * @param {Editor} fs - the memfs editor instance
+ * @param {boolean} addMissingLibraries - indicates whether to add missing libraries or throw an error
+ */
+export function checkRequiredLibraries(basePath: string, fs: Editor, addMissingLibraries = false): void {
```
ManuelPastorini, line 70: "check" should just check and return a result
"add" should add
and maybe there could be a method "ensureNeededDependencies" or similar, that first calls check, and based on the result calls add.

815are (commented):
As we checked with @AlinaGovoruhina then it simpler to change existing method `validateBasePath` than writing new method. We just need to change how `validateBasePath` checks dependency - instead of checking if one of passed dependencies are missing, we would need check if all passed dependencies are missing -> `OR` vs `AND`.
As we checked - in all current cases only single dependency is passed from outside and there was no case when multiple dependencies are passed for `validateBasePath`.
As conclusion, proposal - we just modify method `validateBasePath` instead of creating new method

packages/fe-fpm-writer/test/unit/building-block.test.ts
```diff
@@ -72,10 +72,8 @@ describe('Building Blocks', () => {
         ).toThrowError(/Invalid view path/);
     });
 
-    test('validate sap.fe.core manifest dependency', async () => {
+    test('validate sap.fe.templates and sap.fe.core manifest dependencies are missing', async () => {
         const basePath = join(testAppPath, 'validate-manifest-dep');
-        fs.write(join(basePath, manifestFilePath), JSON.stringify(testManifestContent));
-
         // Test generator without sap.fe.core as dependency in manifest.json
         fs.write(join(basePath, manifestFilePath), JSON.stringify({ ...testManifestContent, 'sap.ui5': {} }));
```
815are, line 78: can we please add additional test case when we have `sap.fe.templates` and `sap.fe.core` is missing? 
Like:
```
"sap.ui5": {
        "dependencies": {
            "libs": {
                "sap.fe.templates": {}
            }
        }
    }
```

Other tests for building block uses setup:
```
"sap.ui5": {
        "dependencies": {
            "libs": {
                "sap.fe.core": {}
            }
        }
    }
```

packages/fe-fpm-writer/src/common/validate.ts
hitesh-parmar: Reviewed by Andis https://github.com/SAP/open-ux-tools/pull/2284#pullrequestreview-2265845394

815are (approved):
checked:
1. code changes are straightforward. Was not sure if we can change `validateBasePath` method logic - logic regarding how it validates multiple dependencies, but as we checked how method is called- we did not find any existing place where multiple dependencies are passed to `validateBasePath`. In result we are updating `validateBasePath` instead of creating new function. Method checks if at least one of passed dependency exists, then it considers as valid.
2. New tests are added;
3. Changelog persists

devinea (approved):

PR #2288, opened 2024-08-27
Fix/2285/odata service inq use v4 if v2 not auth

IainSAP:
Fix for #2287 

- Adds fallback to v4 of v2 catalog request fails with 404, 401, 403
- Adds test to cover


longieirl (approved):
Changes look good.
Tests cover change.
Changeset present.

devinea (approved):
Great. Thanks @IainSAP 
Code looks good and tests.
Changeset ✅

PR #2295, opened 2024-08-28
fix(guided-answers-helper) change HELP_TREE to const instead of enum and simplify structure

devinea:
fix(guided-answers-helper) change HELP_TREE to const instead of enum and simplify structure


cianmSAP (approved):
Updates look good
Changeset ✅

longieirl (approved):
Changes look good.
Cleanup looks good.

PR #2296, opened 2024-08-28
fix: ensure that the controller template include an example call to super

nlunets:
The current template don't include any example of calling back super.onInit() or PageController.prototype.onInit.apply(this, arguments).

This is mandatory for application otherwise the controllers are not properly instantiated :)

### DIFF
--- a/.changeset/weak-peaches-battle.md
+++ b/.changeset/weak-peaches-battle.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fe-fpm-writer': patch
+---
+
+fix: ensure that the controller template include an example call to super.onInit()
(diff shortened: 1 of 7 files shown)


tobiasqueck (approved):
* changeset ok
* only comments added to generated code, so nothing to locally validate

tobiasqueck:
> @tobiasqueck Should examples be tagged with `@example` as per https://jsdoc.app/tags-example?

It is not required but it also doesn't hurt, so please use it where it makes sense.

tobiasqueck:
> > @tobiasqueck Should examples be tagged with `@example` as per https://jsdoc.app/tags-example?
> 
> It is not required but it also doesn't hurt, so please use it where it makes sense.

Ok, that was a generic response. For this one here, I wouldn't because it is code that people just un-comment when they want to get started.

devinea (approved):
changes useful guide. 
changeset ✅

PR #2298, opened 2024-08-28
fix(i18n): CAP scenario. `getCapI18nFolder` -> handle absolute path instead of relative path to resolve i18n folder

815are:
We noticed that method `getCapI18nFolder` expects that second parameter `path` should be relative path to cds file, while other methods and also `ApplicationStructure` expects/contains absolute paths, then proposal is to adjust function `getCapI18nFolder` to same expectation.

### DIFF
--- a/.changeset/dull-pillows-grab.md
+++ b/.changeset/dull-pillows-grab.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': minor
+---
+
+Method `createCapI18nEntries` - handle absolute path to cds file instead of relative path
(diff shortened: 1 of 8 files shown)


voicis (approved):
- Straight forward change
- Good test coverage
- Changeset present

voicis (approved):

Klaus-Keller (approved):
Thanks @815are!
- code change to use absolute path is clear and easy to understand
- changeset exists
- test coverage is great

Approved from my side.

PR #2299, opened 2024-08-28
fix(ui5-application-writer) backward support for older ui5 versions

### LINKED ISSUE #2297: BUG - locate-reuse-lib.js does not work with older versions of UI5
### Related Feature

### Description
A clear description of what the bug is.

### Steps to Reproduce

1. generate a FE project with ui5 version 1.71.x 
2. preview

### Expected results
App loads

### Actual results

app fails to load. 
Screenshot 2024-08-28 at 19 39 44

This seems to because this code is not backward compatible:
https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-application-writer/templates/optional/loadReuseLibs/webapp/test/locate-reuse-libs.js#L231-L233

Update to be compatible with UI5 V2
```JS
    sap.ushell.Container.createRenderer(true).then(function (component) {
        component.placeAt("content");
    });
```
This older version works with UI5 V1 versions
```JS
    sap.ushell.Container.createRenderer().placeAt(\\"content\\");
```
### Screenshots

OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

devinea:
fix #2297

### DIFF
--- a/.changeset/friendly-trees-admire.md
+++ b/.changeset/friendly-trees-admire.md
@@ -1,0 +1,7 @@
+---
+"@sap-ux/fiori-elements-writer": patch
+"@sap-ux/fiori-freestyle-writer": patch
+"@sap-ux/ui5-application-writer": patch
+---
+
+ui5-application-writer - fix backward support for older ui5 versions in locate-reuse-libs.js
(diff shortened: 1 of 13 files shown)


IainSAP (approved):
Thanks @devinea 

- Snapshots updated as expected
- Not tested locally

longieirl (approved):
Changes look good.
Snapshots covering change.
Changeset present.

devinea:
Thanks all

PR #2303, opened 2024-08-29
fix: adjustment of the color of chevron on selection

emrecevik06:
There was a problem with the color of the chevron when selecting in the light theme. When selecting, the chevron was still dark and did not use the color of the text (white).


packages/control-property-editor/src/panels/outline/Tree.tsx
```diff
@@ -413,7 +413,7 @@ export const Tree = (): ReactElement => {
                     onContextMenu={(e) => isExtensionPoint && handleOpenTooltip(e, tooltipId)}>
                     {groupHeaderProps?.group?.count !== 0 && (
                         <Icon
-                            className={`${chevronTransform}`}
+                            className={`${chevronTransform} chevron`}
```
Jimmy-Joseph19, line 416:
```suggestion
                            className={`${chevronTransform} active-foreground`}
```

Jimmy-Joseph19 (approved):
Changes looks good.
Changeset found
Approving

packages/control-property-editor/src/panels/outline/Tree.tsx
```diff
@@ -413,7 +413,7 @@ export const Tree = (): ReactElement => {
                     onContextMenu={(e) => isExtensionPoint && handleOpenTooltip(e, tooltipId)}>
                     {groupHeaderProps?.group?.count !== 0 && (
                         <Icon
-                            className={`${chevronTransform}`}
+                            className={`${chevronTransform} active-foreground`}
```
voicis, line 416: I don't think we should add new class for this. We already have a class for selected row that we can use `app-panel-selected-bg` in combination with the row class `tree-row`. All icons in the tree should be handled generically and we shouldn't need to fix  this for each of them, but with current solution extension icons when selected still has wrong colour.

image

voicis (changes requested):

voicis (approved):
- Straight forward style change
- Review comments addressed
- Changeset present
- Tested locally

Jimmy-Joseph19 (approved):
Feedback Changes looks good
Approving

PR #2305, opened 2024-08-29
fix(ui-components): UIContextualMenu.getUIContextualMenuItemStyles - make `props` param option

815are:
During updating `ui-components` to latest version, we noticed that method `getUIContextualMenuItemStyles` expects first param `props` to be required. 
image

It was introduced during -> https://github.com/SAP/open-ux-tools/pull/2234

but according to code - method can be adjusted to make `props` as optional.

### DIFF
--- a/.changeset/stale-falcons-cry.md
+++ b/.changeset/stale-falcons-cry.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui-components': patch
+---
+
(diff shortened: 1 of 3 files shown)


Jimmy-Joseph19 (approved):
Changes looks good.
Test coverage good
Changeset found
Approving

broksy (approved):
Changeset reflects to the changes made. Approved.

PR #2306, opened 2024-08-30
fix(ui-components): UIContextualMenu.getUIcontextualMenuCalloutStyles- change `props` param to `styles` param and make it as optional

815are:
During updating `ui-components` to latest version, we noticed that method `getUIcontextualMenuCalloutStyles` expects first param `props` to be required. 
image

It was introduced during -> https://github.com/SAP/open-ux-tools/pull/2234

but according to code - method can be adjusted to make `props` as optional. 
Additionally - as only `styles` are used from `props`, then changed param from `props` to `styles` to allow pass styles without required properties like `items`

### DIFF
--- a/.changeset/moody-frogs-battle.md
+++ b/.changeset/moody-frogs-battle.md
@@ -2,4 +2,4 @@
 '@sap-ux/ui-components': patch
 ---
 
-UIContextualMenu.getUIcontextualMenuCalloutStyles - make `props` param optional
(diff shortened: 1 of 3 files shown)


broksy (approved):
Changeset reflects to the changes made. Changes are covered by tests. Approved from my side.

Jimmy-Joseph19 (approved):
Changes ✅ 
Test coverage ✅ 
Changeset found
Approving

PR #2312, opened 2024-09-02
feat(fe-fpm-writer): apply context path to a building block based on min ui5 version

AlinaGovoruhina:
Apply contextPath for minimum UI5 versions equal or below 1.96 ( + always for chart building block).
Check for minimum UI5 version when applying the building block and when getting the code snippet content.

### DIFF
--- a/packages/fe-fpm-writer/src/building-block/index.ts
+++ b/packages/fe-fpm-writer/src/building-block/index.ts
@@ -10,7 +10,7 @@ import format from 'xml-formatter';
 import { getErrorMessage, validateBasePath } from '../common/validate';
 import { getTemplatePath } from '../templates';
 import { CodeSnippetLanguage, type FilePathProps, type CodeSnippet } from '../prompts/types';
-import { coerce, lte } from 'semver';
+import { coerce, lt } from 'semver';
 import type { Manifest } from '@sap-ux/project-access';
 import { getMinimumUI5Version } from '@sap-ux/project-access';
 
@@ -192,7 +192,7 @@ function getTemplateContent<T extends BuildingBlock>(
         const minUI5Version = manifest ? coerce(getMinimumUI5Version(manifest)) : undefined;
         const applyContextPath =
             buildingBlockData.buildingBlockType === BuildingBlockType.Chart ||
-            !!(minUI5Version && lte(minUI5Version, '1.96.0'));
+            !!(minUI5Version && lt(minUI5Version, '1.97.0'));
         // Convert object based metapath to string
         const metadataPath = getMetaPath(applyContextPath, buildingBlockData.metaPath, usePlaceholders);
         buildingBlockData = { ...buildingBlockData, metaPath: metadataPath.metaPath };
--- a/packages/fe-fpm-writer/src/building-block/index.ts
+++ b/packages/fe-fpm-writer/src/building-block/index.ts
@@ -30,7 +30,7 @@ interface MetadataPath {
  *
  * @param {string} basePath the base path
  * @param {Editor} fs the memfs editor instance
- * @returns {Manifest} the manifest content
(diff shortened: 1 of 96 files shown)


packages/fe-fpm-writer/test/unit/prompts/prompts.test.ts
```diff
@@ -246,6 +247,20 @@ describe('Prompts', () => {
             expect(result.viewOrFragmentPath.filePathProps?.fileName).toBe('Main.view.xml');
         });
 
+        test.each(types)('Type "%s", get code snippet, min ui5Version = 1.95.0', async (type: PromptsType) => {
```
815are, line 250: could you please add test to version like `1.96.34` or there is already such test?

packages/fe-fpm-writer/src/building-block/index.ts
```diff
@@ -156,25 +172,30 @@ function getMetaPath(
...
 ): string {
     const templateFolderName = buildingBlockData.buildingBlockType;
     const templateFilePath = getTemplatePath(`/building-block/${templateFolderName}/View.xml`);
     if (typeof buildingBlockData.metaPath === 'object' || buildingBlockData.metaPath === undefined) {
+        // Special handling for chart - while runtime does not support approach without contextPath
+        // or for equal or below UI5 v1.96.0 contextPath is applied
+        const minUI5Version = manifest ? coerce(getMinimumUI5Version(manifest)) : undefined;
+        const applyContextPath =
+            buildingBlockData.buildingBlockType === BuildingBlockType.Chart ||
+            !minUI5Version ||
```
815are, line 195: I think if `minUI5Version` is not resolved we consider it as latest, example of existing check -> https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/common/validate.ts#L16

packages/fe-fpm-writer/test/unit/prompts/prompts.test.ts
AlinaGovoruhina: Added cases 1.96.0 - context path applied;
1.96.25 - no context path for filter or table building blocks

packages/fe-fpm-writer/src/building-block/index.ts
AlinaGovoruhina: Fixed: if not defined, contextPath is NOT applied to the building block

packages/fe-fpm-writer/test/unit/prompts/__snapshots__/prompts.test.ts.snap
```diff
@@ -1211,6 +1230,24 @@ exports[`Prompts getCodeSnippet Type "filter-bar", get code snippet, min ui5Vers
 />"
 `;
 
+exports[`Prompts getCodeSnippet Type "filter-bar", get code snippet, min ui5Version = 1.96.25 1`] = `
+"<macros:FilterBar
+    id=\\"id\\"
+    metaPath=\\"/entity/qualifier\\"
```
815are, line 1236: @AlinaGovoruhina  I think it is bug, on version `1.96.25` it should use `contextPath`. @hitesh-parmar  please correct if I am wrong

hitesh-parmar: Yes, this is correct. For all version 1.96.x there should be a contextPath

AlinaGovoruhina: This is fixed.

815are (approved):
checked:
1. code is simple to understand and version check conditions fulfil expectation;
2. new tests are added
3. changeset presented

devinea (approved):
code changes and tests look good. 
changeset ✅

PR #2321, opened 2024-09-04
fix: quick action panel displaying fix for v 1.71

vadson71:
Bug fix. Quick Action panel was not displayed in case of UI5 v1.71


packages/preview-middleware-client/src/adp/quick-actions/fe-v2/registry.ts
```diff
@@ -58,12 +55,4 @@ export default class FEV2QuickActionRegistry extends QuickActionDefinitionRegist
         }
         return definitionGroups;
     }
-
-    protected getComponentContainerFromPage(page: Control): Control | undefined {
-        if (page instanceof XMLView) {
```
voicis, line 63: This is FE V2 specific logic, so this should stay here. We just need to have a version specific branch here for the case when there is no XML view under NavContainer.

voicis (changes requested):

packages/preview-middleware-client/src/adp/quick-actions/fe-v2/registry.ts
vadson71: refactored

voicis (approved):
- Straight forward change
- Review comments addressed
- Tested locally

vinayhospete (approved):

Issue #2294, opened 2024-08-28
FEATURE - Generate SAP ABAP annotations in CAP CDS


voicis:
_As a **developer**,
I want to **generate SAP ABAP annotations**,
so that I **create an ABAP service with basic UI annotations**_

### Description
Add an option to generate SAP ABAP annotations instead of OData annotations from annotation generator as an experimental feature.

### Technical Design
_What modules need to be created or enhanced?_ 
Extend `@sap-ux/fiori-annotation-api` so it can insert SAP ABAP annotations for `LineItem` and `FieldGroup`. (only insert change needs to be supported)
_How would the functionality be exposed and how can it be consumed?_
A `writeSapAnnotations` parameter needs to be passed when creating a new instance of `FioriAnnotationService`.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have fresh CAP project with a UI5 application
**when** I execute generate annotations function with`writeSapAnnotations` parameter
**then** basic SAP annotations for UI will be generated in applications CDS file.

### Tasks
- [x] Implement OData -> SAP ABAP annotation conversion
- [x] Add `writeSapAnnotations` parameter to the `FioriAnnotationService`
- [x] Call conversion in CDS adapter before passing the `insert` change to the writer
- [x] Error handling (currently only limited scope needs to be supported and if the API is called outside the bounds of this scope we'll throw an error)
- [x] Unit Tests added

PR #2322, opened 2024-09-04
feat(annotation-generator): generate SAP annotations

voicis:
#2294 

Add option to generate SAP annotations in CAP CDS projects.


vinayhospete (approved):

packages/fiori-annotation-api/src/cds/adapter.ts
```diff
@@ -281,31 +286,36 @@ export class CDSAnnotationServiceAdapter implements AnnotationServiceAdapter, Ch
...
+                }
+                if (!writer) {
+                    //writable cds document (augment)
+                    const textDocument = TextDocument.create(change.uri, 'cds', 0, cachedFile);
+                    writer = new CDSWriter(
+                        this.facade,
+                        this.vocabularyService,
+                        document.ast,
+                        document.comments,
+                        document.tokens,
```
Jimmy-Joseph19, line 308: probably can pass document itself

Jimmy-Joseph19 (approved):
Changes looks good.
Covered all scenarios with unit test.
Changeset found.
Approving

packages/fiori-annotation-api/src/cds/adapter.ts
voicis: Yes, but that would be refactoring of already existing code. I'd prefer to keep that out of this PR and make a separate change for it.

PR #2325, opened 2024-09-04
fix(abap-deploy-config-inquirer): bug found with showing manual package prompt

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/2128

Two issues found
1. bug found with manual package being shown if package choice is a list
- tests updated to confirm flow

2. Issue with abap state not being updated with the target URL so transport request is returning incorrect error as `createAbapServiceProvider` was throwing the exception `'Unable to handle the configuration in the current environment.'` since the URL was empty.
- tests to cover state change

### DIFF
--- a/packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
+++ b/packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
@@ -206,13 +206,13 @@ export function defaultOrShowManualPackageQuestion(
     useAutocomplete = false
 ): boolean {
     return (
-        (!isCli || packageInputChoice === PackageInputChoices.EnterManualChoice || !useAutocomplete) &&
+        ((!isCli && !useAutocomplete) || packageInputChoice === PackageInputChoices.EnterManualChoice) &&
         defaultOrShowPackageQuestion()
     );
 }
 
 /**
- * Determines if the search (autcomplete) package input prompt can be shown based on backend availability.
+ * Determines if the search (autocomplete) package input prompt can be shown based on backend availability.
  *
  * @param isCli - is in CLI
  * @param packageInputChoice - package input choice from previous answers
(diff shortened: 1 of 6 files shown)


tobiasqueck (dismissed):
* changeset ok
* change makes sense and fixes the bug
* local test using the `create` module works as before

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
```diff
@@ -225,6 +225,9 @@ export function defaultOrShowSearchPackageQuestion(
     useAutocomplete = false
 ): boolean {
     // Only show the autocomplete prompt when the autocomplete prompt is supported; CLI or YUI specific version
```
IainSAP, line 227: "YUI specific version" is no longer relevant as its been supported since 1.10.something.

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
```diff
@@ -225,6 +225,9 @@ export function defaultOrShowSearchPackageQuestion(
...
+    if (!useAutocomplete) {
```
IainSAP, line 228: This is a change in behaviour, did we not support CLI usage without autocomplete previously,. The following logic seems redundant now:
```
return (
        (isCli || useAutocomplete) &&
```

IainSAP (commented):
Some comments, not clear what the behaviour should be now?

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
longieirl: Updated.

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
longieirl: Updated based on conversation.

longieirl:
@IainSAP can you please review again, comments addressed and a 2nd issue found when testing.

IainSAP (approved):
thx @longieirl 

- comments addressed
- tests added to cover

tobiasqueck (approved):

PR #2330, opened 2024-09-05
fix(cpe): quick action titles

voicis:
Use correct case in quick action titles.

Also added feature toggle to enable quick actions. By default they will be disabled.

### DIFF
--- a/packages/preview-middleware-client/src/adp/quick-actions/load.ts
+++ b/packages/preview-middleware-client/src/adp/quick-actions/load.ts
@@ -8,6 +8,9 @@ import type { ApplicationType } from '../../utils/application';
  * @returns Quick Action registries.
  */
 export async function loadDefinitions(appType: ApplicationType): Promise<QuickActionDefinitionRegistry<string>[]> {
+    if (localStorage.getItem('com.sap.ux.control-property-editor.features.quick-actions') !== 'true') {
+        return [];
+    }
     if (appType === 'fe-v2') {
         const FEV2QuickActionRegistry = (await import('open/ux/preview/client/adp/quick-actions/fe-v2/registry'))
             .default;
--- a/packages/preview-middleware-client/src/adp/quick-actions/load.ts
+++ b/packages/preview-middleware-client/src/adp/quick-actions/load.ts
@@ -8,6 +8,7 @@ import type { ApplicationType } from '../../utils/application';
  * @returns Quick Action registries.
  */
 export async function loadDefinitions(appType: ApplicationType): Promise<QuickActionDefinitionRegistry<string>[]> {
+    // eslint-disable-next-line fiori-custom/sap-no-localstorage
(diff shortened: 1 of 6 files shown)


nahmed22 (approved):
no major change, titles improved. 
change set maintained. 
tests adapted.

packages/preview-middleware-client/src/adp/quick-actions/load.ts
```diff
@@ -8,6 +8,10 @@ import type { ApplicationType } from '../../utils/application';
  * @returns Quick Action registries.
  */
 export async function loadDefinitions(appType: ApplicationType): Promise<QuickActionDefinitionRegistry<string>[]> {
+    // eslint-disable-next-line fiori-custom/sap-no-localstorage
+    if (localStorage.getItem('com.sap.ux.control-property-editor.features.quick-actions') !== 'true') {
```
Jimmy-Joseph19, line 12: One suggestion would be using cookie instead of local storage so we can avoid feature toggle for different port numbers.

packages/preview-middleware-client/src/messagebundle.properties
```diff
@@ -1,17 +1,17 @@
-QUICK_ACTION_ADD_PAGE_CONTROLLER=Add controller to page
-QUICK_ACTION_SHOW_PAGE_CONTROLLER=Show page controller
+QUICK_ACTION_ADD_PAGE_CONTROLLER=Add Controller to Page
+QUICK_ACTION_SHOW_PAGE_CONTROLLER=Show Page Controller
 QUICK_ACTION_OP_ADD_HEADER_FIELD=Add Header Field
 
-V2_QUICK_ACTION_CHANGE_TABLE_COLUMNS=Change table columns
+V2_QUICK_ACTION_CHANGE_TABLE_COLUMNS=Change Table Columns
```
Jimmy-Joseph19, line 5: It was earlier feedback from @MN-public to maintain it as sentence case. Did you discuss?

voicis: Yes, we should use title case here. https://experience.sap.com/fiori-design-web/ui-text-guidelines-for-sap-fiori/#case

packages/preview-middleware-client/src/adp/quick-actions/load.ts
voicis: Cookies will also be sent to the server on every request, which we do not need in this case. Also having origin specific feature toggles seems to be more flexible - if you want to enable it only for a specific origin.

Jimmy-Joseph19 (approved):
Changes looks fine.
Tested locally and works as expected.
Changeset found.
Test Coverage excellent.

Approving.

Klaus-Keller (approved):
Thanks @voicis
- code changes are easy to understand
- changeset exists
- test coverage is great

Approved from my side.

PR #2331, opened 2024-09-06
Change themes 03

emrecevik06:
Update variable colors and themes


packages/control-property-editor/src/components/ThemeSelectorCallout.tsx
```diff
@@ -17,7 +17,7 @@ export type ThemeName = 'dark' | 'light' | 'high contrast';
 export function ThemeSelectorCallout(): ReactElement {
     const { t } = useTranslation();
     const [isCalloutVisible, setValue] = useState(false);
-    const theme = localStorage.getItem('theme') ?? 'dark';
+    const theme = localStorage.getItem('theme') ?? 'dark modern';
```
voicis, line 20: Please use `useLocalStorage` hook instead, which correctly prefixes the keys.

We can't also just change the theme names, because if there is an existing theme stored in local storage, then we will end up with no theme variables loaded, because a theme with that name no longer exists. Switching to the prefixed key, should avoid this issue.

voicis (changes requested):

packages/control-property-editor/src/use-theme.ts
```diff
@@ -0,0 +1,19 @@
+import type React from 'react';
+import { useEffect } from 'react';
+import { useLocalStorage } from './use-local-storage';
+
+export type ThemeName = 'dark modern' | 'light modern' | 'high contrast black';
+/**
+ * Use local storage.
```
voicis, line 7: JSDoc seems to be misleading here. It looks like it was copied from `useLocalStorage` function.

voicis (approved):

voicis (approved):
- Code changes looks good
- Good test coverage
- Tested locally

nikmace (approved):
Code looks good
Good test coverage
Also previewed locally

There is no linked issue, and I don't completely understand the high-level purpose of the change. Is it to make the colors more vibrant? When I previewed locally I noticed that for the dark mode, the panels got darker.

emrecevik06:
> There is no linked issue, and I don't completely understand the high-level purpose of the change. Is it to make the colors more vibrant? When I previewed locally I noticed that for the dark mode, the panels got darker.

It was a recurring task to keep the variables in sync with VS Code. VS Code had some changes to their variables and this PR changes the variables accordingly.

PR #2333, opened 2024-09-06
fix: Dont update launch.json if app is not in workspace but launch config exists

kjose90:
Prevent updates to launch.json File if found when generating launch config outside of workspace


devinea (approved):
Thanks @kjose90 
Changset ✅

longieirl (approved):
Changes are fixing a very specific bug.
Tests to cover change.
Changeset present.

PR #2338, opened 2024-09-09
Remove odata service inquirer dependency from launch config

kjose90:
[#31230 ](https://github.wdf.sap.corp/ux-engineering/tools-suite/issues/31230)

### DIFF
--- a/packages/launch-config/src/types/types.ts
+++ b/packages/launch-config/src/types/types.ts
@@ -1,6 +1,5 @@
 import type { ODataVersion } from '@sap-ux/project-access';
 import type { FioriToolsProxyConfigBackend } from '@sap-ux/ui5-config';
-import type { OdataVersion, DatasourceType } from '@sap-ux/odata-service-inquirer';
 
 export enum Arguments {
     FrameworkVersion = '--framework-version',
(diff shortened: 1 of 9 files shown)


packages/launch-config/src/types/types.ts
```diff
@@ -60,6 +59,17 @@ export interface LaunchConfigInfo {
     filePath: string;
 }
 
+export enum DatasourceType {
+    capProject = 'capProject',
+    odataServiceUrl = 'odataServiceUrl',
+    metadataFile = 'metadataFile'
+}
+
+export enum OdataVersion {
```
IainSAP, line 68: Please use a type from a common module. Maybe project access?

packages/launch-config/src/types/types.ts
```diff
@@ -60,6 +59,17 @@ export interface LaunchConfigInfo {
     filePath: string;
 }
 
+export enum DatasourceType {
```
IainSAP, line 62: Please consider renaming this to something specific to launch config as it could be confused with the odata-service-inquirer datasource type which is a different concept. I'd suggest something like: 
```
type startScriptType = 'cds' | 'mockServer'
```

I cant find any reference to 'odataServiceUrl' so I guess its not needed?

IainSAP (changes requested):
Please see comments.

kjose90:
already done

PR #2339, opened 2024-09-09
fix(fiori-annotation-api): write annotations outside cds project

voicis:
#2294

Support writing SAP annotations in a file that is not part of the CDS project.


vinayhospete (approved):

Jimmy-Joseph19 (approved):
Changes looks good. 
Covered with unit test.
Changeset found
Approving

PR #2344, opened 2024-09-10
2128/fix package selection

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/2128

Two issues found
- mandatory field not required
- translation misconfigured


longieirl:
Wrong branch

Issue #2346, opened 2024-09-10
BUG - `odata-service-inquirer` : Default system name does not include sap-client value


IainSAP:
### Description

Client number is missing from system name:

<img width="838" alt="image" src="https://github.com/user-attachments/assets/50c4f1fc-cc3e-438a-bc28-cd34b87220fb">

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

IainSAP:
Done

PR #2349, opened 2024-09-10
Fix/2346/odata service inquirer client prompt issues

IainSAP:
Fix for #2346 

Various client prompt related issues

-  Re-request services if client changes
-  Adds client to system name prompt default value
-  Maintain validated client in ConnectionValidator 
-  Dont reset validity when creating system (we need this state)


longieirl (approved):
Changes look good.
Tests in place to cover change.

kjose90 (approved):
changes looks good.
changeset ✅

PR #2353, opened 2024-09-11
feat: building blocks -> additional property to `BuildingBlockMetaPath` - to always generate absolute paths with options to turn relative

815are:
We noticed that we need different result for 2 different extensions:
1. 1st extension - we have page context and we should allow to generate relative paths;
2. 2nd extension - we do not have page context and in result we always need to generate absolute path;

As solution - adding addition property to `BuildingBlockMetaPath` -> `alwaysAbsolutePath`, where default value is `true`, but for first scenario we can pass `false`

Updated example with option to toggle new property:
1. `alwaysAbsolutePath=true`:
image

2. `alwaysAbsolutePath=false`:
image


AlinaGovoruhina (approved):
Code is understandable, has unit tests, prompting-ui example updated to verify the new parameter

devinea (approved):
Code changes are clear and tested.
Changeset ✅

PR #2355, opened 2024-09-11
test(launch-config): fix failing tests on windows

815are:
fix failing tests on windows:
image


kjose90 (approved):
Thanks @ 815are for the fix.

broksy (approved):
Fix looks good as previously the build was failing locally on Windows.

Issue #2354, opened 2024-09-11
Fix font family issues to include -apple-system fonts


gpeuc:
On Control Property Editor, because it's browser based and cannot rely on the parent system of VS Code, it seems we have copy/pasted a bit wrong font stack.

At the moment `--vscode-font-family` resolves as hard coded `"Segoe WPC", "Segoe UI", sans-serif`, which is Windows stack. This happened because developers who lifted CSS from VS Code to be used in standalone environment of browser did this on Windows machine. However, on Mac `--vscode-font-family` resolves as: `-apple-system, BlinkMacSystemFont, sans-serif`

Therefore, if I am on Mac, and I am looking at CPE, I will get the Windows font stack loaded, which I do not want. 

The solution is to add `-apple-system, BlinkMacSystemFont` to the beginning of the stack in CPE CSS.

So, it should be something like this:

`--vscode-font-family: -apple-system, BlinkMacSystemFont, "Segoe WPC", "Segoe UI", sans-serif;`

This way Mac will pick up first block, and Windows will ignore it and go to the next item which successfully resolves (probably Segoe).

@815are

815are:
PR -> https://github.com/SAP/open-ux-tools/pull/2356

815are:
@gpeuc PR is merged, can we close issue?

gpeuc:
I have even managed to check it on my own machine using the magic of pulling this branch. I'm a big boy now. Closing.

PR #2356, opened 2024-09-11
fix(control-property-editor): Update the CSS variable --vscode-font-family with a MacOS-specific font-family to match the VSCode UI when the control property editor is running on MacOS.

815are:
Issue -> #2354

Changes:
1. `control-property-editor` - Update the CSS variable --vscode-font-family with a MacOS-specific font-family to match the VSCode UI when the control property editor is running on MacOS.
2. `ui-components` - update styles which is used in dev mode(when `pnpm storybook` launched) same like in 1st point.


voicis (approved):
- Straightforward change.
- Tested locally

vadson71 (approved):
Straightforward changes

Issue #2361, opened 2024-09-12
BUG - Deploy scripts for Cloud Adp Project are not working


GDamyanov:
### Description
Deploy scripts generated by the writer for Adp Project from @sap-ux/adp-tooling are not working for Cloud Adp Projects

### Steps to Reproduce
1. Generate Adp Project from Cloud System.
2. Open terminal and navigate to project directory.
3. Execute npm run deploy.

### Expected results
The project should be deployed successfully.

### Actual results
Errors occurs in the console and the process is terminated.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2362, opened 2024-09-12
fix: Update Adp writer deploy scripts

GDamyanov:
fix for #2361 
Update deployment related scripts generated from the writer in @sap-ux/adp-tooling for Adp Projects


packages/adp-tooling/src/writer/options.ts
```diff
@@ -178,21 +180,28 @@ function addOpenSourceMiddlewares(ui5Config: UI5Config, config: AdpWriterConfig)
...
+        appName: config?.app?.bspName,
+        languages: config?.app?.languages?.map((language: Language) => {
+            return {
+                sap: language.sap,
+                i18n: language.i18n
+            };
+        })
+    };
+
+    if (!isAppStudio()) {
```
tobiasqueck, line 195: Could you always add this please, so that projects are ready for usage everywhere.
Even if you run in AppStudio, you have the url and auth type. The url is part of the destination configuration and the authtype is always `reentranceTicket` since we only add the task for ABAP cloud systems.

packages/adp-tooling/templates/project/package.json
```diff
@@ -15,14 +15,15 @@
...
+        "@ui5/cli": "^3.9.2"<%if (locals.deploy) {%>,
+        "rimraf": "^5.0.5"<%}%>
     },
     "scripts": {
         "build": "ui5 build --exclude-task generateFlexChangesBundle generateComponentPreload minify --clean-dest",
         "start": "<%= locals.options?.fioriTools ? 'fiori run' : 'ui5 serve' %> --open /test/flp.html#app-preview",
         "start-editor": "<%= locals.options?.fioriTools ? 'fiori run' : 'ui5 serve' %> --open /test/adaptation-editor.html"<%if (locals.deploy) {%>,
-        "deploy": "ui5 build --config ui5-deploy.yaml --exclude-task <%= locals.options?.fioriTools ? 'deploy-to-abap' : 'abap-deploy-task' %> generateFlexChangesBundle generateComponentPreload --clean-dest && <%= locals.options?.fioriTools ? 'fiori deploy' : 'deploy' %> --config ui5-deploy.yaml",
-        "undeploy": "<%= locals.options?.fioriTools ? 'fiori undeploy' : 'undeploy' %> --config ui5-deploy.yaml --lrep \"apps/<%= app.reference %>/appVariants/<%= app.id %>/\"",
+        "deploy": "npm run build && fiori deploy --config ui5-deploy.yaml && rimraf archive.zip",
```
tobiasqueck, line 25: The deploy process should not generate an `archive.zip` if not explicitly requested, therefore, I am confused that you require `rimraf` here.
If it does generate an `archive.zip` then there is a regression that needs an investigation.

tobiasqueck: You also need to add the `IF` again because you only have the `fiori deploy` command if the `Fiori tools` are added, otherwise, you need to use `ui5 build`

tobiasqueck (changes requested):

packages/adp-tooling/templates/project/package.json
longieirl: @tobiasqueck in SAP Fiori tools, the package.json deploy script is appended as follows;
```
"deploy": "npm run build && fiori deploy --config ui5-deploy.yaml && rimraf archive.zip",
```
During the deploy task, there is a zip generated on the fly. However, a user can specify a path to an already generated zip which will obviously bypass this zip generation task.

https://github.com/SAP/open-ux-tools/blob/main/packages/deploy-tooling/src/cli/archive.ts#L82

packages/adp-tooling/src/writer/options.ts
GDamyanov: For BAS we don't need this configuration, because we are using destinations.

packages/adp-tooling/templates/project/package.json
GDamyanov: I checked in @sap-ux/abap-deploy-config-writer and this module is always  generating the same scripts  like in this PR

tobiasqueck: Well ... I still don't understand the usage of `rimraf` because as @longieirl said, the zip is generated on the fly, so nothing needs to be deleted.
The deploy config generated is only generating things for the Fiori tools, while the adp writer already supports both, Fiori tools and a "pure" open source version, so instead of introducing a regression in the adp-writer, this needs to be improved in the deploy-config writer, and then, maybe reused here instead of copying an implementation

packages/adp-tooling/src/writer/options.ts
tobiasqueck: A project generated in BAS and then checked into e.g. github, could be checked out in VSCode and it should also work there as well. Just because it is generated in BAS, does not necessarily mean it only used in BAS.

packages/adp-tooling/templates/project/package.json
longieirl: @tobiasqueck the archive remains if there is any exception thrown during the process! I think this is more of a cleanup task.

packages/adp-tooling/src/writer/options.ts
GDamyanov: Fixed in [a3c9caa](https://github.com/SAP/open-ux-tools/pull/2362/commits/a3c9caada35cc2c1ce209f4e86e7b03a04f73ba5)

packages/adp-tooling/templates/project/package.json
GDamyanov: Fixed in [a3c9caa](https://github.com/SAP/open-ux-tools/pull/2362/commits/a3c9caada35cc2c1ce209f4e86e7b03a04f73ba5)

packages/adp-tooling/test/unit/writer/__snapshots__/index.test.ts.snap
```diff
@@ -17,14 +17,15 @@ Object {
...
+        \\"@ui5/cli\\": \\"^3.9.2\\",
+        \\"rimraf\\": \\"^5.0.5\\"
     },
     \\"scripts\\": {
         \\"build\\": \\"ui5 build --exclude-task generateFlexChangesBundle generateComponentPreload minify --clean-dest\\",
         \\"start\\": \\"fiori run --open /test/flp.html#app-preview\\",
         \\"start-editor\\": \\"fiori run --open /test/adaptation-editor.html\\",
-        \\"deploy\\": \\"ui5 build --config ui5-deploy.yaml --exclude-task deploy-to-abap generateFlexChangesBundle generateComponentPreload --clean-dest && fiori deploy --config ui5-deploy.yaml\\",
-        \\"undeploy\\": \\"fiori undeploy --config ui5-deploy.yaml --lrep \\\\\\"apps/the.original.app/appVariants/my.test.app/\\\\\\"\\",
+        \\"deploy\\": \\"npm run build && fiori deploy --config ui5-deploy.yaml && rimraf archive.zip\\",
```
longieirl, line 27: `.gitignore` should also exclude this from source control which I think from looking at the snapshots, is appended i.e. `*.zip`

longieirl (approved):
From my understanding of ADP, changes look good.
Changeset is good
Test coverage is good.
Requires another review from a developer with more ADP experience.

packages/adp-tooling/templates/project/package.json
```diff
@@ -15,14 +15,15 @@
...
+        "@ui5/cli": "^3.9.2"<%if (locals.deploy) {%>,
+        "rimraf": "^5.0.5"<%}%>
     },
     "scripts": {
         "build": "ui5 build --exclude-task generateFlexChangesBundle generateComponentPreload minify --clean-dest",
         "start": "<%= locals.options?.fioriTools ? 'fiori run' : 'ui5 serve' %> --open /test/flp.html#app-preview",
         "start-editor": "<%= locals.options?.fioriTools ? 'fiori run' : 'ui5 serve' %> --open /test/adaptation-editor.html"<%if (locals.deploy) {%>,
-        "deploy": "ui5 build --config ui5-deploy.yaml --exclude-task <%= locals.options?.fioriTools ? 'deploy-to-abap' : 'abap-deploy-task' %> generateFlexChangesBundle generateComponentPreload --clean-dest && <%= locals.options?.fioriTools ? 'fiori deploy' : 'deploy' %> --config ui5-deploy.yaml",
-        "undeploy": "<%= locals.options?.fioriTools ? 'fiori undeploy' : 'undeploy' %> --config ui5-deploy.yaml --lrep \"apps/<%= app.reference %>/appVariants/<%= app.id %>/\"",
+        "deploy": "npm run build && <%= locals.options?.fioriTools ? 'fiori deploy' : 'deploy' %> --config ui5-deploy.yaml && rimraf archive.zip",
```
tobiasqueck, line 25: This is quite a change. Before your change, the the build configuration from the `ui5-deploy.yaml` was used for building the sources, now it is using the `ui5.yaml`. Are you sure about that?

So, maybe something like `npm run build -- --config ui5-deploy.yaml && <%= locals.options?.fioriTools ? 'fiori deploy' : 'deploy' %> --config ui5-deploy.yaml && rimraf archive.zip` is the better option.

GDamyanov: Well we don't need ui5-deploy.yaml for build, we need it only for deployment.

tobiasqueck: Actually, you do. The idea of having a separate `ui5.yaml` and a `ui5-deploy.yaml` is that you can have separate build configurations e.g. to filter test/local files for the deploy (productive) build. So, having the clean separation here makes sense and should not be removed even if the generated (and not modified) files are the same.

packages/adp-tooling/templates/project/package.json
tobiasqueck: Not sure I understand what was fixed with https://github.com/SAP/open-ux-tools/pull/2362/commits/a3c9caada35cc2c1ce209f4e86e7b03a04f73ba5. I am expecting that we add `--clean-dest` again and remove `rimraf`. I think it is a good idea to keep the zip if something went wrong with the deployment, if all goes well, no zipfile is created if I understood @longieirl correctly.

packages/adp-tooling/src/writer/options.ts
tobiasqueck: Maybe I am misreading the code (and the snapshots) but as far as I can see the url is still not being added when a project is generated in BAS. Please ensure that if you generate in BAS, the project also works in VSCode.

packages/adp-tooling/templates/project/package.json
GDamyanov: I tested your proposal adding the ui5-deploy.yaml for build and deploy is failing there is some issue in the zipped archive of the project which is catched by the backend. Our deployment for cloud project is not working at the moment and we have very high ticket for it. Do you think we could do a sync to discuss and agree on a solution to expedite the resolution of the problem?

packages/adp-tooling/src/writer/options.ts
GDamyanov: Fixed in [5467cec](https://github.com/SAP/open-ux-tools/pull/2362/commits/5467cec391857a90519c5c396b8eab7edc06088d)

packages/adp-tooling/src/writer/options.ts
```diff
@@ -178,21 +182,35 @@ function addOpenSourceMiddlewares(ui5Config: UI5Config, config: AdpWriterConfig)
...
+    let target: CloudCustomTaskConfigTarget;
+    if (config.target.destination) {
+        target = {
+            destination: config.target.destination,
+            url: config.target?.url ?? vscodeUrl
+        };
+    } else {
+        target = {
+            url: config.target.url,
+            authenticationType: config.target.authenticationType,
```
longieirl, line 194: I would add the authentication to the destination config as well. As local preview will read this value when spinning up a local proxy.

GDamyanov: For BAS we don't need authentication type.

IvoSG (approved):
Code looks good
Test coverage is good
Not tested manually

packages/adp-tooling/src/writer/options.ts
```diff
@@ -14,9 +14,13 @@ import type {
     Content,
     CloudApp,
     ChangeInboundNavigation,
-    InternalInboundNavigation
+    InternalInboundNavigation,
+    CloudCustomTaskConfig,
+    CloudCustomTaskConfigTarget
 } from '../types';
 
+const vscodeUrl = 'https://REQUIRED_FOR_VSCODE.example';
```
nikmace, line 22: If this is a const, it should be uppercase letters, something like: `VSCODE_URL`. Otherwise, I don't see any reason for this extraction into a variable.

packages/adp-tooling/src/types.ts
```diff
@@ -533,6 +533,24 @@ export interface CustomConfig {
...
+    url: string | undefined;
+    authenticationType: string | undefined;
+    ignoreCertErrors: boolean;
+}
+
+export interface OnPremTarget {
+    destination?: string;
+}
+
+export type CloudCustomTaskConfigTarget = OnPremTarget | CloudTarget;
```
nikmace, line 546: I am not sure why we need these if we already have interfaces in https://github.com/SAP/open-ux-tools/blob/main/packages/system-access/src/types.ts?

packages/adp-tooling/src/writer/options.ts
```diff
@@ -178,21 +182,35 @@ function addOpenSourceMiddlewares(ui5Config: UI5Config, config: AdpWriterConfig)
  * @returns list of required tasks.
  */
 function getAdpCloudCustomTasks(config: AdpWriterConfig & { target: AbapTarget } & { app: CloudApp }): CustomTask[] {
+    let target: CloudCustomTaskConfigTarget;
+    if (config.target.destination) {
```
nikmace, line 186: Maybe it's good to have "?." check just in case?
```suggestion
    if (config?.target?.destination) {
```

nikmace (changes requested):

packages/adp-tooling/src/writer/options.ts
GDamyanov: Fixed in [216a14a](https://github.com/SAP/open-ux-tools/pull/2362/commits/216a14ab037ab88ab871d6d3620f7bfd88aa0e5f)

packages/adp-tooling/src/writer/options.ts
GDamyanov: Fixed in [857a668](https://github.com/SAP/open-ux-tools/pull/2362/commits/857a6683440c925497339881d418afaa488ef86c)

nikmace (approved):
Code looks good
Coverage is good
Did not test manually

packages/adp-tooling/src/writer/options.ts
longieirl: Sorry, I resolved this based on an offline conversation.

Can you confirm if you think the authenticationType needs to be appended? I think if you are porting apps from BAS to VSC then the authenticationType can play a role.

longieirl (approved):
Changes look good.
Changeset is appropiate.
Tests to cover change.

tobiasqueck (approved):

PR #2363, opened 2024-09-12
fix(preview-middleware): change fragment index

### LINKED ISSUE #2360: BUG - Error when choosing the non-default index in the Add Fragment dialog
### Description
When adding a new fragment with the index value other than default, error message is displayed in the pop-up.

### Steps to Reproduce
1. Open the Adaptation Editor.
2. Select any control e.g table
3. Open context menu and select `Add Fragment`
4. In the dialog, choose Index value that is not suggested by default
5. Enter fragment name and click Create

### Expected results
Pop-up requesting save is displayed

### Actual results
Pop-up informing of the error is displayed.
```
Could not get command for 'addXML'. "3" is of type string, expected int for property "index" of ManagedObject sap.ui.rta.command.AddXML#__xml1
```

Error:
```
Log-dbg.js:499 2024-09-12 00:42:55.998900 EventProvider.fireEvent: Event Listener for event 'press' failed during execution. - Error: Could not get command for 'addXML'. "3" is of type string, expected int for property "index" of ManagedObject sap.ui.rta.command.AddXML#__xml1  
 Error: Could not get command for 'addXML'. "3" is of type string, expected int for property "index" of ManagedObject sap.ui.rta.command.AddXML#__xml1
    at m.validateProperty (ManagedObject-dbg.js:1641:11)
    at m.setProperty (ManagedObject-dbg.js:1498:17)
    at c.setIndex (ManagedObjectMetadata-dbg.js:231:42)
    at m.applySettings (ManagedObject-dbg.js:1342:21)
    at ManagedObject-dbg.js:537:11
    at c.constructor (ManagedObject-dbg.js:515:4)
    at c [as constructor] (Metadata-dbg.js:504:19)
    at c [as constructor] (Metadata-dbg.js:504:19)
    at new c (Metadata-dbg.js:504:19)
    at CommandFactory-dbg.js:463:15
```

OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

voicis:
Fixes #2360


vadson71 (approved):
Simple and straightforward changes

Issue #2365, opened 2024-09-13
TBI - cap-config-writer: remove dependency -> dev dependency on `@sap-ux/odata-service-writer`


IainSAP:
### Description (include screenshots)

Move the dep -> dev dep

### Value
Optional - details on how this technical backlog item provides value to end-users or dev team eg. Performance, Quality, Maintainability. This will help determine the ranking in relation to feature development

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

IainSAP:
Done

PR #2366, opened 2024-09-13
tbi(cap-config-writer): Move type only dependency -> dev dependency

IainSAP:
https://github.com/SAP/open-ux-tools/issues/2365

Moves dep only types are consumed


longieirl (approved):
Changes look good.

kjose90 (approved):
Thanks @IainSAP 👍

PR #2348, opened 2024-09-10
feat: During building block creation, add the 'sap.fe.macros' library to the 'manifest.json' if it is not already listed

815are:
All changes related to creation of building blocks like `Chart`, `Table`, `Filterbar`.
Changes:
1. If 'sap.fe.templates' or 'sap.fe.core' are not listed in `manifest.json`, then do not throw error about missing dependency;
2. If 'sap.fe.macros' is not listed in `manifest.json`, then automatically add it to 'manifest.json' during building block creation;

PR #2367, opened 2024-09-13
feat(fe-fpm-writer): reuse 'async getManifest' function and make API methods async

815are:
follow up after -> https://github.com/SAP/open-ux-tools/pull/2348
where we started to use function `getWebappPath` from `@sap-ux/project-access`.
Such change required to make API methods `async`.

In this PR - I continued with all other API calls ->
1. reusing `getManifest` or `getManifestPath` function;
2. making involved API methods `async`

Addition change:
1. `@sap-ux/fiori-elements-writer` -> call 'generateCustomPage' from 'fe-fpm-writer' as async function

### DIFF
--- a/.changeset/lucky-candles-promise.md
+++ b/.changeset/lucky-candles-promise.md
@@ -1,6 +1,0 @@
----
-'@sap-ux/fe-fpm-writer': patch
----
-
-- During building block creation, add the 'sap.fe.macros' library to the 'manifest.json' if it is not already listed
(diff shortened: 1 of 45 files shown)


devinea (approved):
Thanks @815are 
Impact was bigger than expected. But changes look good.
Changeset ✅

AlinaGovoruhina (approved):
Changes look good, unit tests updated

PR #2368, opened 2024-09-13
fix: missing unique react key

815are:
fix to avoid react warning about unique react key:
image

scenario:
1. open storybook
2. open -> http://localhost:64403/?path=/story/tables-uitable--read-only-table

Problem - see warning in console


Klaus-Keller (dismissed):
Thanks @815are,
- changes to add unique key for react control looks good
- changeset exists
- test exists

approved from my side

voicis (approved):
- changes looks good
- changeset exists
- missing key errors no longer appear in CPE tests

Klaus-Keller (approved):
Thanks @815are!
- adding unique id for items looks good
- change set exists
- test coverage is good
- did not test manually, just visual review

Approved from my side.

Klaus-Keller (approved):
re-approving

PR #2376, opened 2024-09-17
fix: right chevron color for high contrast black theme

emrecevik06:
There was an issue with the chevron color for the high contrast black theme after this [PR](https://github.com/SAP/open-ux-tools/pull/2331). Previously, there was also an issue with the chevron colors, which I fixed in this [PR](https://github.com/SAP/open-ux-tools/pull/2303), but those changes are no longer needed. After the theme update, new color variables were introduced.


vinayhospete (approved):

voicis (approved):
- Simple and straight forward change.
- Tested locally.

PR #2377, opened 2024-09-17
fix: theme picker hover and border for dropdowns

emrecevik06:
There is now a border for the dropdowns and also a background color when hovering over the themes in the theme picker.


vinayhospete (approved):

packages/control-property-editor/src/components/ThemeSelectorCallout.scss
```diff
@@ -46,6 +46,16 @@
...
 }
 
+.theme-child:hover {
+    background-color: var(--vscode-toolbar-hoverBackground, var(--vscode-menubar-selectionBackground));
+    outline: 1px dashed var(--vscode-contrastActiveBorder);
+}
+
+.selected:hover {
+    border-radius: 1px;
+    background-color: var(--vscode-progressBar-background);
```
voicis, line 56: This is duplicating styles from line 44-47. Is this to avoid styles being overwritten by selector `.theme-child:hover` styles? If that is the case then we can use [not pseudo class](https://developer.mozilla.org/en-US/docs/Web/CSS/:not), to avoid altering `selected` styles even when hover is active.

.changeset/pretty-gorillas-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/control-property-editor': patch
+---
+
+Added hover over theme picker and border for dropdowns
```
voicis, line 5: Changes here are for `Callout` control, dropdown is a different control.

packages/control-property-editor/src/components/ThemeSelectorCallout.scss
emrecevik06: Yes, I need to avoid overwriting `.theme-child:hover` to prevent the selected theme from having a different background color when hovering. If I didn't, the selected theme would show the background color meant for hovering, rather than the selected background color. I will try to do it with using not pseudo class. Thanks!

voicis (approved):
- changes looks good
- changeset present
- tested locally

vinayhospete (approved):

PR #2380, opened 2024-09-17
fix(service-inq): determine sap client from cookies - sap-usercontext=sap-client

cianmSAP:
For BTP systems the sap-client can be found in the cookies e.g `sap-usercontext=sap-client=100`

### DIFF
--- a/packages/odata-service-inquirer/src/prompts/connectionValidator.ts
+++ b/packages/odata-service-inquirer/src/prompts/connectionValidator.ts
@@ -259,7 +259,11 @@ export class ConnectionValidator {
                 // Full service URL
                 await this.createOdataServiceConnection(axiosConfig, url.pathname);
             }
-            this._validatedClient = url.searchParams.get(SAP_CLIENT_KEY) ?? undefined;
+            this._validatedClient =
+                url.searchParams.get(SAP_CLIENT_KEY) ??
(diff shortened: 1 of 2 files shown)


packages/odata-service-inquirer/src/prompts/connectionValidator.ts
```diff
@@ -259,7 +259,11 @@ export class ConnectionValidator {
                 // Full service URL
                 await this.createOdataServiceConnection(axiosConfig, url.pathname);
             }
-            this._validatedClient = url.searchParams.get(SAP_CLIENT_KEY) ?? undefined;
+            this._validatedClient =
+                url.searchParams.get(SAP_CLIENT_KEY) ??
+                this.getSapClient(this._serviceProvider?.cookies?.toString()) ??
```
IainSAP, line 264: The only supported client setting is through the search params. Therefore this code should be unnecessary.

IainSAP:
You can access the client from the service key directly. The point of the _validatedClient is to indicate to prompts that the client has been validated by 1) An input validator and 2) By connecting. I cannot see any reason to do this.

IainSAP (changes requested):
Client should only be set by prompts, not sure what the implication of setting default clients as the validated client will be. We need to discuss an alternative.

PR #2381, opened 2024-09-17
feat: Enhance Quick action list to add custom actions

nahmed22:
Add quick-actions to add table and page action


nahmed22:
absolete: https://github.com/SAP/open-ux-tools/pull/2408

PR #2383, opened 2024-09-18
fix(cpe): manifest change notification

voicis:
- When manifest changes are created then show notification that reload is required
- Change `Save` button to `Save and Reload` if the changes requires reload


nikmace (approved):
Changed look good
Coverage is good
Did **not** test manually

Jimmy-Joseph19 (approved):

Jimmy-Joseph19 (approved):
Changes looks good.
Approvind.
Did not test locally

PR #2384, opened 2024-09-18
feat(ui-prompting): list/dropdown question - automatically select the option when only one item is available in a dropdown list

815are:
It was requested/proposed during tests, that if combobox has single option, then we should auto select such single option.

changes:
1. updated logic in `Select.tsx` for single selection combobox;
2. updated `README.md` - there was `ui-components` mentioned instead of `ui-prompting`;
3. storybook is updated with additional example to test combobox with single option;


broksy (approved):
I didn't test locally, but changes are clear and reflects to the changeset and are covered by tests.

Jimmy-Joseph19 (approved):
Changes looks good.
Changeset found.
Covered with unit test 
Have not tested manually.

PR #2387, opened 2024-09-18
fix(abap-delploy-config): verify destination input

cianmSAP:
Verify the destination prompt has input rather than automatically returning true


longieirl (approved):
Re-approving

kjose90 (approved):
Changes looks good ✅

PR #2389, opened 2024-09-19
fix(cpe): application mode after reload

### LINKED ISSUE #2341: BUG - Adaptation Editor displays wrong mode after reload
### Description
In Adaptation Editor incorrect mode is displayed after reloading preview.

### Steps to Reproduce
1. Open `Adaptation Editor`
2. Add any fragment
3. Save changes
4. Change to `Navigation` mode
5. Click on `reload` button in the changes panel 

### Expected results
Either editor stays in `Navigation` mode or the UI is updated to show that editor is in `Adaptation` mode.

### Actual results
After reload editor is in `Adaptation` mode, but in the toolbar you can see that `Navigation` mode is enabled.

OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

voicis:
Fixes #2341 and varius usability fixes

- Disabled QuickActions in navigation mode because some of them are only supported in adaptation mode (styling changes are implemented in #2394)
- Do not show nested quick action title if there is only one child
- Increased max width for quick actions to fill the available space in the panel


Jimmy-Joseph19 (dismissed):
Changes looks good.
Test coverage is good.
Changeset found.
Haven't tested manually
Approving

packages/preview-middleware-client/types/sap.ui.rta.d.ts
```diff
@@ -240,7 +241,8 @@ declare module 'sap/ui/rta/RuntimeAuthoring' {
         } & Component;
         stop: (bSkipSave, bSkipRestart) => Promise<void>;
         attachStop: (handler: (event: Event) => void) => void;
-        setMode: (sNewMode: string) => void;
+        getMode: () => RtaMode;
```
nikmace, line 244: Is this method available for all ui5 versions or only higher ones? If it's not lower versions will throw error

nikmace (changes requested):

packages/preview-middleware-client/types/sap.ui.rta.d.ts
nikmace: Checked that it works in lower versions including 1.71.70

nikmace (approved):
Code looks good
Coverage is good
Did not test manually

Jimmy-Joseph19 (approved):

PR #2394, opened 2024-09-20
fix(ui-components): Applied missing styles for the disabled state of UILink.

815are:
It was reported that when UIlink is disabled using property `disabled`, then there no any visual difference that link is disabled.
Changes:
1. Apply styles for link when property 'disabled' is 'true';
2. Update storybook with option to disable links in example

image

image

### DIFF
--- a/.changeset/rude-walls-clap.md
+++ b/.changeset/rude-walls-clap.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui-components': patch
+---
+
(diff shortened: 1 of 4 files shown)


vadson71 (approved):
Changes are simple and straightforward. Tested, included in the story

Klaus-Keller (approved):
Thanks @815are, 
- changes for style of disabled control looks good
- changeset exists
- test coverage is great
- did just visual review, no manual testing

Approved from my side

PR #2395, opened 2024-09-20
fix(cpe): quick action improvements

voicis:
Various quick action usability fixes #https://github.wdf.sap.corp/ux-engineering/tools-suite/issues/31326:

- Customised  `Add Fragment` dialog for quick actions (only required fields are visible and dedicated title)
- For fragments with templates save and reload button is enabled
- Refactored communication with CPE to be more accessible (only changed the places that affect this PR to limit the number of changes, the rest should be handled in a separate PR)
- Added translatable texts for `Add Fragment` dialog


vadson71 (dismissed):
Changes look good, test coverage is also good. 
New functionality tested manually locally.

packages/preview-middleware-client/src/cpe/connector-service.ts
```diff
@@ -35,13 +35,26 @@ export class WorkspaceConnectorService {
...
 
-    private onChangeSaved(fileName: string, kind: 'delete' | 'create', changeType?: string) {
+    private onChangeSaved(fileName: string, kind: 'delete' | 'create', change: unknown = {}) {
+        const { changeType, content } = change as {
+            changeType?: string;
+            content?: {
+                templateName?: string;
+                fragmentPath?: string;
+            };
+        };
```
nikmace, line 45: Would it make sense to have less arguments for this method since fileName already exists in the change and you can take this parameter by restructuring the change as you are already doing?

packages/preview-middleware-client/src/cpe/communication-service.ts
```diff
@@ -0,0 +1,41 @@
...
+        this.sendAction = sendAction;
+    }
+
+    /**
+     * Creates a listener to receive actions from the CPE.
+     *
+     * @param handler - Action handler.
+     */
+    static subscribe(handler: ActionHandler): void {
+        // console.log(handler)
```
nikmace, line 38: A console log comment was left here.

nikmace (changes requested):
Just some minor comments

packages/preview-middleware-client/src/cpe/connector-service.ts
vadson71: The fileName callback parameter is defined in the Storage interface 
image
This callback is called for any file, not obligatory for flex change file. Thus, fileName must be provided

packages/preview-middleware-client/src/cpe/communication-service.ts
vadson71: removed

nikmace (approved):
Code looks good
Coverage is good
Did not test manually

PR #2396, opened 2024-09-20
Uitable a11y2

tom-sap:
follow up from https://github.com/SAP/open-ux-tools/pull/2391
(git merge went wrong)

The `:focus` style should have been added to the column header cells (not to the header row).


815are (approved):
checked:
1. style update - did not notice issue
2. changeset is presented

devinea (approved):

PR #2402, opened 2024-09-23
Add skipLiveScripts flag to `DebugConfig` options 

kjose90:
This update removes the `ProjectDataSourceType` from `DebugConfig` and introduces the use of a new flag, `skipLiveScripts`. Now, live scripts are only generated when `skipLiveScripts` is set to false.


IainSAP (approved):
Thanks @kjose90 

- Improvement simplifies API and decouples from datasource concept
- Updates as expected

PR #2403, opened 2024-09-24
fix: remove controlid popup and add tooltip for control id/-type values

emrecevik06:
Remove the control id pop up and add a tooltip for control id/-type values.


packages/control-property-editor/src/panels/properties/HeaderField.tsx
```diff
@@ -81,8 +81,10 @@ export function HeaderField(headerFieldProps: Readonly<HeaderFieldProps>): React
             </UITooltip>
             <TextField
                 id={label}
+                data-testid={label}
                 value={value}
                 readOnly={true}
+                title={value}
```
Jimmy-Joseph19, line 87: Is this information not required anymore

> 'The unique identifier within a page, either configured or automatically generated.'

emrecevik06: This information is related to the removed popup. We decided to remove the pop up, because it had hardcoded values and was not really helpful. Instead of the popup, we decided to add some tooltips for the values, because some of them are truncated.

emrecevik06:
Please do not review yet

emrecevik06:
obsolete. Handled in this PR: https://github.com/SAP/open-ux-tools/pull/2460

Issue #2406, opened 2024-09-24
FEATURE - UI5 library generator


cianmSAP:
_As a developer,
I want to use a yeoman generator module,
so that I can be prompted for ui5 library config and generate the library_

### Description
Prompting `@sap-ux/ui5-library-inquirer` and writing `@sap-ux/ui5-library-writer` are published modules in open-ux-tools, this module will encompass both into the ui5 library generator 

### Technical Design
`@sap-ux/fiori-generator-shared`: `getDefaultTargetFolder` required to determine the target folder
`@sap-ux/ui5-library-inquirer`: update `UI5LibraryAnswers` interface 
`@sap-ux/nodejs-utils`: update `CommandRunner` to support additional `opts` to install in specific directories

Create the new module `@sap-ux/ui5-library-generator`

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

cianmSAP:
Closing as module @sap-ux/ui5-library-sub-generatro is published

PR #2407, opened 2024-09-24
feat(ui5-lib-gen): adds new module @sap-ux/ui5-library-generator

cianmSAP:
#2406 

Adds new module @sap-ux/ui5-library-generator

`@sap-ux/fiori-generator-shared`: adds `getDefaultTargetFolder` to determine the target folder
`@sap-ux/ui5-library-inquirer`: updates `UI5LibraryAnswers` interface so properties are not optional
`@sap-ux/nodejs-utils`: updates `CommandRunner` to support additional opts to install in specific directories


packages/ui5-library-generator/README.md
```diff
@@ -0,0 +1,18 @@
+# @sap-ux/generator-ui5-library
+
```
IainSAP, line 2: Can we please add more info about what this generator actually does and maybe an example.

packages/ui5-library-generator/src/app/index.ts
```diff
@@ -0,0 +1,118 @@
...
+     *
+     * @param args - arguments passed to the generator
+     * @param opts - options passed to the generator
+     */
+    constructor(args: string | string[], opts: Generator.GeneratorOptions) {
+        super(args, opts);
+
+        this.appWizard = AppWizard.create(opts);
+        this.vscode = opts.vscode;
+        ReuseLibGenLogger.logger = new ToolsLogger({ logPrefix: '@sap-ux/generator-ui5-library' });
```
IainSAP, line 44: This is no longer integrated to YUI logging. How will this work now when integrated back into YUI, it needs to allow the YUI logger to also be used and only create a new logger when one is not configured via YUI settings.

packages/ui5-library-generator/src/app/types.ts
```diff
@@ -0,0 +1,16 @@
+import type { Prompts, AppWizard } from '@sap-devx/yeoman-ui-types';
+import type { UI5LibraryAnswers } from '@sap-ux/ui5-library-inquirer';
+
+export interface LibAnswers {
```
IainSAP, line 4: We re-defining an existing interface `UI5LibraryAnswers` is the definition that should be used?

IainSAP (changes requested):
Please continue to  support integration with YUI logging as is currently supported.

packages/ui5-library-generator/src/utils/constants.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const generatorTitle = 'SAP Fiori Reusable Library';
+export const prompts = [
+    {
+        name: 'Generate UI5 Library',
+        description: 'Create a Reusable Library for use in SAP Fiori applications'
+    }
+];
+
+export const initialAnswers = {
```
IainSAP, line 11: Strange that we are setting everything to empty strings. Usually we do not do this in generators. Not sure its necessary. Undefined should suffice as you are otherwise setting values that have not been input.

cianmSAP: This is not for setting defaults in the inquirer. 

The issue I'm facing is, the `UI5LibraryAnswers` found in `packages/ui5-library-inquirer/src/types.ts` all had optional properties. 

The `ui5-library-writer` uses `UI5LibConfig` which has required properties e.g `libraryName` & `namespace`.

So when calling the writer with the `UI5LibraryAnswers` I get the typescript error about properties cannot be undefined. 

Is there a better approach? The only alternative I can see is to pass it something like 
```
 const ui5Lib = {
            libraryName: this.answers.libraryName || defaultLibName
.....
```

packages/ui5-library-generator/src/app/types.ts
cianmSAP: This was added by mistake and not used. I'll remove

packages/ui5-library-generator/src/utils/constants.ts
IainSAP: I see. I think answers interfaces may all be undefined, setting the interface to mandatory wont make any difference to the runtime as prompts may not execute or may be skipped so you are correct to assign default values when calling the `generate` if they are required. I would have defaults for ui5libConfig, not call it answers, as only inquirers can generate answers, and apply it when creating the `ui5LibConfig` object so its clear whats happening?

cianmSAP: Ok thanks

packages/ui5-library-generator/README.md
cianmSAP: Done

packages/ui5-library-generator/src/app/index.ts
cianmSAP: Added logWrapper in fiori gen shared, as I expect it will be needed going forward

IainSAP (approved):
Thanks @cianmSAP 

- Previous comments addressed
- Not tested locally

packages/ui5-library-generator/package.json
```diff
@@ -0,0 +1,59 @@
+{
+    "name": "@sap-ux/ui5-library-sub-generator",
```
devinea, line 2: The folder should be aligned with the name of the module e.g. `ui5-library-sub-generator`

cianmSAP: Thanks I've updated

.changeset/old-hornets-tie.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sap-ux/fiori-generator-shared': minor
+'@sap-ux/nodejs-utils': minor
+'@sap-ux/generator-ui5-library': patch
```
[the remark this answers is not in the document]
cianmSAP: Updated, cheers

devinea (dismissed):
Thanks @cianmSAP 
New module setup looks good. 
Changes to existing looks good. 
Changeset ✅

devinea (approved):

longieirl (approved):
Changeset present
All comments addressed.
Good test coverage.
Will be consumed and validated in tools-suite.

PR #2411, opened 2024-09-26
fix: add border for read only input fields

emrecevik06:
Add border for control id/-type read only fields


packages/control-property-editor/src/panels/properties/HeaderField.tsx
```diff
@@ -88,7 +88,14 @@ export function HeaderField(headerFieldProps: Readonly<HeaderFieldProps>): React
                     field: {
                         color: 'var(--vscode-input-foreground)',
                         fontSize: defaultFontSize,
-                        backgroundColor: 'var(--vscode-sideBar-background)'
+                        backgroundColor: 'var(--vscode-sideBar-background)',
+                        border: '1px solid var(--vscode-input-border)',
+                        selectors: {
+                            ':hover': {
+                                borderColor: 'var(--vscode-focusBorder)'
```
Jimmy-Joseph19, line 95: The changes looks good.
do we need hover effect for readonly elements?

Jimmy-Joseph19 (approved):
Changes looks good.
Chageset found

PR #2413, opened 2024-09-26
Change separator color for panels and toolbar

emrecevik06:
Change separator color for left and right panel and also for the toolbar


vinayhospete (approved):

Jimmy-Joseph19 (approved):
Changes looks good.
Tested manually.
Changeset found.

PR #2415, opened 2024-09-27
feat(ui-prompting): rename ui prompting examples package from `@sap-ux/prompting-ui` to `@sap-ux-private/ui-prompting-examples`

815are:
it was noticed that we have two packages with similar namings:
1. `@sap-ux/ui-prompting` - public package for ui prompting https://github.com/SAP/open-ux-tools/blob/main/packages/ui-prompting/package.json
2.  `@sap-ux/prompting-ui` - private package containing examples/playground using `@sap-ux/fe-fpm-writer` and `@sap-ux/ui-prompting`

Problem - names `@sap-ux/ui-prompting` and `@sap-ux/prompting-ui` similar and hard to detect which one is public and which is private with examples.

Change:
1. rename folder `examples/prompting-ui` ->`examples/ui-prompting-examples`;
2. rename package name from `@sap-ux/prompting-ui` to `@sap-ux-private/ui-prompting-examples`


Klaus-Keller (approved):
Thanks @815are!
- renaming and private scope looks good
- changeset exists
- I've did a local test with storyboard, did not see any issues

Approved from my side.

broksy (approved):
Code changes corresponds to the changeset.
Did not test locally, but approved from my side.

PR #2417, opened 2024-09-27
fixing start mock

kjose90:
- Extend `DebugOptions` with `targetMockHtmlFile`, which should be populated when the project includes `flpSandboxMockServer.html`. This flag instructs the Start Mock command to open `flpSandboxMockServer.html` instead of `flpSandbox.html`.
- If both `targetMockHtmlFile` and `migratorMockIntent` are provided, use `migratorMockIntent` as part of the arguments to be executed instead of `flpAppId`.
- Updated tests accordingly.


packages/launch-config/src/debug-config/config.ts
```diff
@@ -8,6 +8,64 @@ const testFlpSandboxHtml = 'test/flpSandbox.html';
...
+ */
+export function getParamsForMockHtml(
+    targetMockHtmlFile: string | undefined,
+    migratorMockIntentWithHash?: string,
+    flpAppIdWithHash?: string
+): string {
+    if (targetMockHtmlFile === testFlpSandboxMockServerHtml && migratorMockIntentWithHash) {
+        return migratorMockIntentWithHash;
+    }
+    return flpAppIdWithHash ?? '';
```
devinea, line 43:
```suggestion
`    return migratorMockIntentWithHash ?? flpAppIdWithHash ;`
```
Not sure that we need to strictly check value to `targetMockHtmlFile` . Can't we accept what the user passes in?
Also, could we simplify the function to this.

kjose90: Good point 👍 
Thanks ✅

devinea (approved):
Thanks for the updates. 
Code changes look good.
changeset ✅

unseen1980 (approved):

PR #2418, opened 2024-09-29
Feat(preview-middleware) make rta path configuration more resilient

heimwege:
- allow rta.editors.path that does not start with `/`
- enhance readme with information on how to launch RTA


nikmace (approved):
Code looks good
Readme updated
Coverage is good
Did not test manually

Issue #2421, opened 2024-09-30
BUG - No validation on FLP input fields


longieirl:
### Related Feature
Feature request: _issue number_

### Description
During the FLP wizard page, enter `SRAP_###` as the semantic object name;

![Screenshot 2024-09-30 at 09 30 15](https://github.com/user-attachments/assets/aedfdfd7-a43b-4194-bf75-9008a5c4e70e)

The manifest.json is updated as follows;
```json
    "crossNavigation": {
      "inbounds": {
        "SRAP100_###-display": {
          "semanticObject": "SRAP100_###",
          "action": "display",
          "title": "{{flpTitle}}",
          "signature": {
            "parameters": {},
            "additionalParameters": "allowed"
          }
        }
      }
    }
  },
```

The manifest.json schema https://github.com/SAP/ui5-manifest/blob/main/schema.json defines the semanticObject as `"pattern": "^[\\w\\*]{0,30}$"` which is not being validated for.

Also, the semantic object name can contain asterisks i.e. `Test_*123` but the unique ID for the inbound which joins the semantic object name and action, cannot contain asterisks.

### Steps to Reproduce
1. Create any project using any template
2. Select Yes to Add FLP Config
3. Enter semantic object name with hash keys
4. Project is built

### Expected results

FLP wizard should block invalid semantic object

### Actual results

Allow any character to be inputted.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [x] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Interested to get feedback on this bug, Semantic Object Name and Action both support asterisks i.e. Test_123#* but when these are joined to create the inbonud unique ID, asterisks is not supported; 

![Screenshot 2024-09-30 at 11 50 38](https://github.com/user-attachments/assets/9a5abd80-0038-4596-b3cf-143f3cd2a1e9)

My approach, would be to restrict Semantic Object Name and Action to only support alphanumeric and underscore characters.

longieirl:
Consumed and validated in tools-suite.

PR #2422, opened 2024-09-30
fix(app-config-writer): export validation function, apply extra validation to input field

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/2421

- extra validation applied to semantic object and action fields
- tests updated
- validation method exported, required in tools-suite to remove duplication of code
- asterisks will not be supported in the semantic object and action fields as these are joined to create the unique inbound ID which doesn't support asterisks


packages/app-config-writer/src/prompt/navigation-config.ts
```diff
@@ -44,9 +44,10 @@ export async function promptInboundNavigationConfig(
  * @param input the text input to validate
  * @param inputName the name of the input as seen by the user
  * @param maxLength optional, the maximum length of text to allow
+ * @param applyRegex optional, not all fields need to be validated with regex
  * @returns true, if all validation checks pass or a message explaining the validation failure
  */
-function validateText(input: string, inputName: string, maxLength = 0): boolean | string {
+export function validateText(input: string, inputName: string, maxLength = 0, applyRegex = true): boolean | string {
```
IainSAP, line 50: Fairly obscure pararm `applyRegex`. What about `disallowSpecialCharacters` ?

longieirl: I've applied a change that takes a list of allowed characters which is appended to the alphanumeric regex so that any other character is handled.

IainSAP (approved):
Thanks @longieirl 

- Comments addressed
- Not tested locally

cianmSAP (approved):
Updates are clear
Tests to cover
Changeset ✅

PR #2424, opened 2024-09-30
fix: ignore developerMode from rta editor config in case of no adp project

heimwege:
Currently the log claims property `developerMode` of `rta.editors` is being set to false in case of rta without adp project but actually the route for this editor is not added at all.

This is fixed in this PR.


tobiasqueck (approved):
* changeset ok
* clean and easy to understand fix
* tests adapted for the change

nikmace (approved):
Code looks good and simple to understand
Coverage is good
Did not test manually

PR #2425, opened 2024-09-30
fix(env-check): add html5 timeout

longieirl:
- need more insight into the HTML5.Timeout property for a BAS destination


packages/environment-check/src/checks/destination.ts
```diff
@@ -38,6 +38,14 @@ export async function checkBASDestination(
         });
     }
 
+    const html5Timeout = !!destination['HTML5.Timeout'];
+    if (!html5Timeout) {
+        logger.push({
+            severity: Severity.Error,
```
cianmSAP, line 44: I'm wondering should this be considered as an error. I don't recall the reason for `DynamicDestination` getting logged here either tbh

cianmSAP: But from an end user persepctive, if they haven't set the `html5Timeout` and are using the default. Should it be logged with `🚫` ?

packages/environment-check/src/translations/env-check.i18n.json
```diff
@@ -63,6 +63,7 @@
         "getTransportRequestsError": "Error checking ability to retrieve available Transport Requests",
         "invalidPkgJson": "Found invalid package.json. Error : {{- error}}",
         "missingDynamicDestProperty": "Additional property 'HTML5.DynamicDestination = true' missing for destination {{- destination}}",
+        "missingTimeoutDestProperty": "Additional property 'HTML5.Timeout = 60000' missing for destination {{- destination}}",
```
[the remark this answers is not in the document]
longieirl: 60000 is the minimium value in most of the SAP documentation that I've seen.

packages/environment-check/src/checks/destination.ts
longieirl: I think it should be a required field for every destination to be honest! There have been many issues with this field not being set which leads to ABAP deployment issues and catalog request issues.

longieirl: Sorry, unless I'm not understanding the code, this will only be shown if the timeout property is missing?

cianmSAP: Yes, it will only be if it is missing. That's fine if you think this property warrants it

longieirl: @cianmSAP I can set this to a warning, if you think it makes more sense?

Klaus-Keller: With this addition we check for existence of `HTML5.Timeout`. Could there be a case where it is set, but the timeout is too short?

longieirl: Yes, there was a customer yesterday where the timeout was set to `6000` instead of `60000`.

packages/environment-check/src/translations/env-check.i18n.json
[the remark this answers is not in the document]
longieirl: Done

packages/environment-check/src/checks/destination.ts
longieirl: Warning message updated to suggest that there is a minimum value required.

cianmSAP (approved):
Thanks @longieirl , looks good from my side
- Comments addressed
- Excellent test coverage
- Have not tested locally but snapshots updates are clear

packages/environment-check/src/types.ts
```diff
@@ -99,6 +99,7 @@ export interface EndpointResults {
     isSapUi5Repo?: boolean; // SAPUI5 repository service for deployment available
     isTransportRequests?: boolean; // Ability to retrieve available Transport Requests
     HTML5DynamicDestination?: boolean;
+    HTML5TimeoutDestination?: boolean;
```
Klaus-Keller, line 102: @longieirl why is it named `HTML5TimeoutDestination` and not `HTML5Timeout`

longieirl: Updated now, thanks.

Klaus-Keller (approved):
Thanks @longieirl,

- adding check for HTML5.Timeout to environment-check looks good
- changeset is present
- test coverage is great

Approved from my side.

PR #2429, opened 2024-10-01
fix: CPE loading changes from backend and not from workspace

mmilko01:


packages/axios-extension/src/abap/lrep-service.ts
```diff
@@ -166,13 +166,20 @@ export class LayeredRepositoryService extends Axios implements Service {
...
      * @returns a promise with an object containing merged app descriptors with their id as keys.
      */
     public async mergeAppDescriptorVariant(
-        appDescriptorVariant: Buffer
+        appDescriptorVariant: Buffer,
+        workspacePath?: string
     ): Promise<{ [key: string]: MergedAppDescriptor }> {
+        let path = '/appdescr_variant_preview/';
+        if (workspacePath) {
+            path += `?workspacePath=${workspacePath}`;
```
tobiasqueck, line 178: Technically this is not part of the path but it is a query parameter. While your approach works, it gets complicated if we start adding more optional parameters in the future, therefore, I'd always recommend handling parameters as parameters and not as concatenated strings, so please use the `params` property of the request config.

packages/preview-middleware-client/src/flp/bootstrap.js
```diff
@@ -15,10 +15,24 @@ async function ushellBootstrap(fnCallback) {
...
         const json = await response.json();
         const version = json?.libraries?.find((lib) => lib.name === 'sap.ui.core')?.version ?? '1.121.0';
-        const majorUi5Version = parseInt(version.split('.')[0], 10);
+        const [major, minor] = version.split('.');
+        const majorUi5Version = parseInt(major, 10);
+        const minorUi5Version = parseInt(minor, 10);
         if (majorUi5Version >= 2) {
             src = `${basePath}/resources/sap/ushell/bootstrap/sandbox2.js`;
         }
+        if (majorUi5Version === 1 && minorUi5Version < 72) {
```
tobiasqueck, line 24: Please add a concise comment here explaining why you are doing what you are doing.

packages/preview-middleware/src/base/config.ts
```diff
@@ -204,22 +205,34 @@ function getFlexSettings(): TemplateConfig['ui5']['flex'] {
...
+    const appName = `${app.intent?.object}-${app.intent?.action}`;
     templateConfig.ui5.resources[id] = app.target;
-    templateConfig.apps[`${app.intent?.object}-${app.intent?.action}`] = {
+    templateConfig.apps[appName] = {
         title: (await getI18nTextFromProperty(app.local, manifest['sap.app']?.title, logger)) ?? id,
         description: (await getI18nTextFromProperty(app.local, manifest['sap.app']?.description, logger)) ?? '',
         additionalInformation: `SAPUI5.Component=${app.componentId ?? id}`,
         applicationType: 'URL',
-        url: app.target,
-        applicationDependencies: { manifest: true }
```
tobiasqueck, line 221: Did you test this with a Fiori application project? Isnt `applicationDependencies: { manifest: true }`required for apps?

packages/axios-extension/src/abap/lrep-service.ts
mmilko01: The fix was a bit difficult here because passing the query parameter via the `params` automatically encodes the value which causes the request to fail with `Error 400 - Invalid workspace path`.  Another issue was the setting the `params` for the request overwrites the default params from the instance (where we have `saml2=disabled`).
I think with the following commits this should be fixed: cd4deddace55670fa261dc00d30444d9215d8b49 and 130d7a2cd8aecdd0df9ef124965390db316295f5

packages/preview-middleware-client/src/flp/bootstrap.js
mmilko01: Added in: e31fc3ee49d0c18c72c8e7d1a9a250e465318b47

packages/preview-middleware/src/base/config.ts
mmilko01: I did some tests with Fiori application projects and I found no issues with the changes. I also compared the resulting `flpSandbox.html` without my changes and I also don't see the `applicationDependencies: { manifest: true }`.  But if you think the risk of a regression is too big, I can add it back.

image

nikmace (changes requested):
Test snapshots do not reflect the changes I think, I only see {\\"manifest\\":true}}, but in the case of ADP there should be the descriptor, or at least its mock.

Please also add a short description of the changes in this PR.

packages/axios-extension/src/abap/lrep-service.ts
```diff
@@ -166,13 +166,24 @@ export class LayeredRepositoryService extends Axios implements Service {
...
+        const params = new URLSearchParams(this.defaults?.params);
+
+        if (workspacePath) {
+            params.append('workspacePath', workspacePath);
+        }
+
         try {
-            const response = await this.put('/appdescr_variant_preview/', appDescriptorVariant, {
+            const response = await this.put(path, appDescriptorVariant, {
+                paramsSerializer: (params) => decodeURIComponent(params.toString()),
```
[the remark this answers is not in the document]
mmilko01: It seems that the default serialization is encoding the params, which causes the request to fail with `Error 400 - Invalid workspace path`. This is the reason I used the paramSerializer

packages/preview-middleware-client/src/flp/bootstrap.js
```diff
@@ -15,10 +15,30 @@ async function ushellBootstrap(fnCallback) {
...
+                    ? obj[target]
+                    : Object.values(obj).reduce((acc, val) => {
+                          if (acc !== undefined) return acc;
+                          if (typeof val === 'object') return getNestedProperty(val, target);
+                      }, undefined);
+            
+                const asyncHints = getNestedProperty(window['sap-ushell-config'], 'asyncHints');
+                if (asyncHints && asyncHints.requests) {
+                    asyncHints.requests = [];
+                }
```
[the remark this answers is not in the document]
mmilko01: As discussed, I changed the algorithm used for setting the nested property.
Fixed in: 57805ceabc6e1e09a4fb028c115d0721ad21c328

nikmace (approved):
Code looks good
Comments addressed, test for adding descriptor added
Excellent coverage
Did not test manually

packages/axios-extension/src/abap/lrep-service.ts
tobiasqueck: Wow, interesting findings. I'd expect that the defaults are always added automatically.

packages/preview-middleware/src/base/config.ts
tobiasqueck: Thanks for verifying my concerns. All good with me.

tobiasqueck (dismissed):
review feedback applied as requested.

packages/preview-middleware-client/src/flp/bootstrap.js
```diff
@@ -15,10 +15,41 @@ async function ushellBootstrap(fnCallback) {
...
+
+                    for (const key in current) {
+                        if (typeof current[key] === 'object' && current[key] !== null) {
+                            stack.push(current[key]);
+                        }
+                    }
+                }
+            };
+            setNestedProperty(window['sap-ushell-config']['applications']);
+        }
```
heimwege, line 52: Can we please have an own function for this and put the comment about what it does and why it's needed in the respective JSDoc? Including `setNestedProperty` into the `ushellBootstrap` function disrupts the readability of the code.

```suggestion
        if (majorUi5Version === 1 && minorUi5Version < 72) {
            removeAsyncHintsRequests();
        }
```

```js
/**
 * For UI5 version 1.71 and below, we need to remove the asyncHints.requests
 * to load the changes in an Adaptation project.
 * This logic needs to be executed here to have a reliable check for
 * UI5 version and remove the asyncHints.requests before the sandbox is loaded.
 * The sandbox shell modifies the `window['sap-ushell-config']`.
 */
function removeAsyncHintsRequests() {
                const obj = window['sap-ushell-config']['applications'];

                if (!obj || typeof obj !== 'object') return;

                const stack = [obj];

                while (stack.length > 0) {
                    const current = stack.pop();

                    if (current.asyncHints) {
                        if (current.asyncHints.requests) {
                            current.asyncHints.requests = [];
                        }
                        return;
                    }

                    for (const key in current) {
                        if (typeof current[key] === 'object' && current[key] !== null) {
                            stack.push(current[key]);
                        }
                    }
                }
}
```

mmilko01: Fixed with: 819600b123158267c9a7d86f6e5a259223b10b8c

nikmace (approved):
Re-approving

PR #2430, opened 2024-10-01
Fix/async loading of sap.ushell.container

heimwege:
w.r.t. UI5 2.x all libs should be loaded async

### DIFF
--- a/.changeset/great-pugs-deliver.md
+++ b/.changeset/great-pugs-deliver.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux-private/preview-middleware-client': patch
+---
+
+fix: no longer load sap.ushell.Container syncronously
--- a/packages/preview-middleware-client/src/flp/init.ts
+++ b/packages/preview-middleware-client/src/flp/init.ts
@@ -267,7 +267,7 @@ export async function init({
     customInit?: string | null;
 }): Promise<void> {
     const urlParams = new URLSearchParams(window.location.search);
-    const container = sap?.ushell?.Container ?? sap.ui.require('sap/ushell/Container');
+    const container = sap?.ushell?.Container ?? await import('sap/ushell/Container');
     let scenario: string = '';
     const ui5VersionInfo = await getUi5Version();
     // Register RTA if configured


heimwege:
@tobiasqueck as discussed yesterday. Do we need some tests for the require branch? Currently this seems to not have been covered by tests

heimwege:
> > @tobiasqueck as discussed yesterday. Do we need some tests for the require branch? Currently this seems to not have been covered by tests
> 
> Well, the answer should always be `yes`, however, I wouldn't know how to properly test it.
> 
> I am still confused that it breaks now but worked so far with UI5 2.0. I checked #1679 it was suggested to do it that way. @petermuessig can you shed some light here?

@tobiasqueck don't get this wrong. The sync/async is not the root cause for the current issues with UI5 2.x!
The actual issue is that after requiring sap.ushell.Container the oConfig property of the container is empty and the container crashes when e.g. trying to createRendererInternal or getServiceAsync. I need to report that internally (because sap.ushell.Container is internal) but to do so I need the async loading. At least that's my current understanding

heimwege:
@petermuessig the `await import('sap/ushell/Container')` is transpiled to `await __ui5_require_async('sap/ushell/Container')` and

```js
function __ui5_require_async(path) {
        return new Promise(function (resolve, reject) {
            sap.ui.require([path], function (module) {
                if (!(module && module.__esModule)) {
                    module = module === null || !(typeof module === 'object' && path.endsWith('/library')) ? { default: module } : module;
                    Object.defineProperty(module, '__esModule', { value: true });
                }
                resolve(module);
            }, function (err) {
                reject(err);
            });
        });
    }
```

which returns a container that can only be used as `container.default.doSomething()` instead of `container.doSomething()`🙈 
So is the whole idea of using import here to async load the container valid at all?

petermuessig:
This is the correct behavior because the static import syntax on top of the file implicitly does this default mapping! See:

```ts
import MessageBox from "sap/m/MessageBox";
MessageBox.show(`Hello World!`);

const MessageBoxAsync = (await import("sap/m/MessageBox")).default;
MessageBoxAsync.show(`Hello World!`);
```

For the static import you could also map it like this:

```ts
import { default as MessageBox } from "sap/m/MessageBox";
MessageBox.show(`Hello World!`);
```

For dynamic imports you could also map it like this:

```ts
const { default: MessageBoxAsync } = (await import("sap/m/MessageBox"));
MessageBoxAsync.show(`Hello World!`);
```

This is the default behavior of the JS language!

heimwege:
Thanks a lot for the clarification @petermuessig 🙏🏼

tobiasqueck (commented):
* new line of code looks ugly BUT except wrapping it into a separate functions (which seems like an overhead) I don't have a better idea
* thanks for removing that other obsolete/forgotten line
* as discussed, I don't know how to properly test that with a unit tests, so it is ok, not to have one
* changeset ok

tobiasqueck (approved):

PR #2431, opened 2024-10-01
Feat/make rta url params optional

heimwege:
This PR makes the usage of url params for the rta editor endpoints optional. In case they are missing a re-route will happen and the parameters will be added 🥳


heimwege:
@ytpo-lyne this PR should help us for https://github.com/SAP/open-ux-tools/tree/feat/create-variantsConfig because the url parameter for the rta editor endpoints are no longer needed with this PR.
@tobiasqueck or do we need to keep them because of some downward compatibility? But if this PR is merged first we should hopefully be fine 🤔

heimwege:
@tobiasqueck I think we do not need to insist on this logic working with the connect API used by the karma test runner as rta editor is not subject to karma tests. Would you agree?

packages/preview-middleware/test/unit/base/flp.test.ts
```diff
@@ -336,7 +341,7 @@ describe('FlpSandbox', () => {
 
         test('rta with developerMode=true and plugin', async () => {
             await server.get('/with/plugin.html').expect(200);
-            const response = await server.get('/with/plugin.html.inner.html').expect(200);
+            const response = await server.get('/with/plugin.html.inner.html').expect(302);
```
tobiasqueck, line 344: We control exactly what is called as `inner.html` therefore, we should be able to also add the url parameters if required, thus, I would not expect that there is a redirect necessary for the inner html.
Is the test case just wrong i.e. should we test here with parameters or is there an unnecessary redirect now?

tobiasqueck (changes requested):
See inline comment - the redirects for tests with `developerMode=true` are not expected.

packages/preview-middleware/test/unit/base/flp.test.ts
heimwege: Are the URL parameters needed at all in case of `inner.html`?

tobiasqueck: Thanks for adjusting the test case

tobiasqueck (dismissed):
* changeset ok
* code change is simple BUT very useful since it allows us to simplify the generated call scripts
* didn't test locally but reviewing the jest tests gives me enough confidence

tobiasqueck (dismissed):

tobiasqueck (approved):

ytpo-lyne (approved):
- changeset present
- code changes are straight forward
- code coverage is excellent 

PR #2432, opened 2024-10-02
fix: create new annotationFile

Jimmy-Joseph19:
For discussing creation of new annotation file for CPE


packages/adp-tooling/src/base/abap.ts
```diff
@@ -33,10 +33,16 @@ export async function getManifest(appId: string, adpConfig: AdpPreviewConfig, lo
 
         // retrieve localmetadata from base project
         const dataSourceName = Object.keys(manifest?.['sap.app']?.dataSources)?.[0];
-        const metadataUrl = `${appInfo.url}/${manifest?.['sap.app']?.dataSources?.[dataSourceName].settings.localUri}`;
+        const metadataPath = manifest?.['sap.app']?.dataSources?.[dataSourceName];
+        let metadataUrl = '';
+        if (metadataPath?.settings?.localUri) {//v2
+            metadataUrl = `${appInfo.url}/${metadataPath.settings.localUri}`;
+        } else if (metadataPath?.uri) { //v4
+            metadataUrl = `${appInfo.url}/${metadataPath.uri}$metadata`;
```
Jimmy-Joseph19, line 41: I was able to get metadata for all v2 apps. V4 app tested in one and the request was failing. But the same request passes in CPE when baseUrl is localhost.

PR #2439, opened 2024-10-07
fix: Sass warning about deprecate usage of '/' for division outside of calc

815are:
Fix for Sass warning about deprecate usage of '/' for division outside of calc:

image


broksy (approved):
Changeset reflects to the code changes.
Approved from my side.

PR #2443, opened 2024-10-07
fix:(ui-prompting) Do not show the translation input if the translation props are not provided

815are:
Do not show the translation input if the translation props are not provided - if there no I18N bundle and no translation event handler


broksy (approved):
Changeset reflects to the changes made. Good test coverage. Approved from my side.

Jimmy-Joseph19 (approved):
Changes looks good.
Unit test coverage good
Changeset found

Issue #2172, opened 2024-07-17
FEATURE - command to remove the local flpSandbox.html and use the preview-middleware instead


tobiasqueck:
_As a **developer**,
I want to **get rid of the flpSandbox.html file in my webapp/test folder and use the on-the-fly generated one from the preview-middleware instead**,
so that I **I don't have to worry about keeping it up-to-date**_

### Description
Currently (but soon to be changed), a generated Fiori application project contains a file at `webapp/test/flpSandbox.html` that is used to locally preview the application but is not used productively. This file has evolved over time - especially  to get it ready for UI5 2.0 - therefore, as an owner of an older project, I need to keep it update to date. In the future, this file will not be generated into new projects anymore, instead it is generated on-the-fly by the preview middleware.
This feature request is to allow remove the file and update the configurations of an existing project, so that this file isn't required anymore.

### Technical Design
Enhance the `@sap-ux/create` module with a new command `sap-ux delete flp` which then
1. deletes `webapp/test/flpSandbox.html` and `webapp/test/locate-reuse-libs.js`
2. reads from the scripts in the `package.json` what configuration should be used 
3. adds/updates the configuration of the preview-middleware
4. updates the scrips in the `package.json` if required

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have an existing Fiori application project with the deprecated middleware config and an flpSandbox file ...
**when** I execute `sap-ux delete flp`
**then** then the `ui5*.yaml` files, and the `package.json` will change so that they work with the local flpSandbox, and the obsolete files are removed.

### Notes
Assuming a project has following scripts in the`package.json`
```json
{
   "start": "fiori run --open \"test/flpSandbox.html?sap-ui-xx-viewCache=false#v2rop-tile\"",
    "start-local": "fiori run --config ./ui5-local.yaml --open \"test/flpSandbox.html?sap-ui-xx-viewCache=false#v2rop-tile\"",
    "build": "ui5 build --config=ui5.yaml --clean-dest --dest dist",
    "start-mock": "fiori run --config ./ui5-mock.yaml --open \"test/flpSandbox.html?sap-ui-xx-viewCache=false#v2rop-tile\"",
    "deploy": "fiori verify",
    "deploy-config": "fiori add deploy-config",
    "start-noflp": "fiori run --open \"index.html?sap-ui-xx-viewCache=false\"",
[...]

PR #2446, opened 2024-10-08
Feat/create command for variants configuration

ytpo-lyne:
PR introduces create command to add Configuration for Variants Creation.

Prerequisite for:
https://github.com/SAP/open-ux-tools/issues/2172

Also the readme.md for the create package has been completely revised


packages/app-config-writer/src/types/variantsConfig.ts
```diff
@@ -0,0 +1,16 @@
+import type { MiddlewareConfig as PreviewConfig } from '@sap-ux/preview-middleware';
+
+export type FioriToolsDeprecatedPreviewConfig = {
```
heimwege, line 3: @tobiasqueck There already exists a type that represents the deprecated config preview config in [ui5-config/src/types/index.ts](https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-config/src/types/index.ts#L49). That still seems to be used to generate outdated configs (see [here](https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-config/src/ui5config.ts#L238)). Do we need to address this in this PR as well? Or is this part of some generator adjustments that will come later because it seems to be used in app writer?

packages/app-config-writer/src/variants-config/utils.ts
```diff
@@ -0,0 +1,139 @@
...
+ * @param basePath - path to project root, where package.json and ui5.yaml is located
+ * @param query - query to create fragment
+ * @returns - review url parameters
+ */
+export async function getRTAUrl(basePath: string, query: string): Promise<string | undefined> {
+    let previewMiddleware: CustomMiddleware<PreviewConfigOptions> | undefined;
+    try {
+        previewMiddleware = await getPreviewMiddleware(undefined, basePath);
+    } catch (error) {
+        //todo: what to do in case there is no ui5.yaml file? try FileName.Ui5MockYaml or FileName.Ui5LocalYaml as fallback?
```
heimwege, line 123: @tobiasqueck you thoughts about potential fallbacks for RTA editor path?

packages/app-config-writer/src/types/variantsConfig.ts
heimwege: Should this type (and previewConfigOptions) be moved to preview middleware?

packages/app-config-writer/src/variants-config/utils.ts
tobiasqueck: While the `ui5.yaml` is not required, it is in every project we generate, so if you are in a project where it doesn't exist, then either something is wrong or the owner knows exactly what they do, therefore, I am ok if our helper tooling doesn't work and they are forced to manually add the script.
So, maybe, just stop here, print an error explaining that we need the `ui5.yaml` to continue or they can add the script manually as described at https://github.com/SAP/open-ux-tools/tree/main/packages/preview-middleware#runtime-adaptation-support

packages/app-config-writer/src/types/variantsConfig.ts
tobiasqueck: I think the conflict/duplication between the types in `ui5-config` and the corresponding middlewares is a separate bigger problem. I don't have a great idea to solve it but I think it needs to be solved just not here. So, here I am ok with having duplicated definitions of the type.

packages/app-config-writer/src/variants-config/utils.ts
heimwege: Even better: we can use ui5.yaml as default but expert users can provide an other yaml file (path) directly via CLI option 🤓

ytpo-lyne: Done with 0810182939708d4afbfa5d85b1f9441bb1d4793d

packages/app-config-writer/package.json
```diff
@@ -29,6 +29,7 @@
         "!dist/**/*.map"
     ],
     "dependencies": {
+        "@sap-ux/preview-middleware": "workspace:*",
```
tobiasqueck, line 32: This should be a `devDep` right? You only need some types.

packages/app-config-writer/src/variants-config/utils.ts
```diff
@@ -0,0 +1,145 @@
...
+ * Gets the preview middleware form the yamlConfig or provided path.
+ * The middleware can either be named 'fiori-tools-preview' or 'preview-middleware'.
+ *
+ * @param yamlConfig - the yaml configuration to use; if not provided, the file will be read with the provided basePath and filename
+ * @param basePath - path to project root, where ui5.yaml is located
+ * @param filename - name of the ui5 yaml file to read from basePath; default is 'ui5.yaml'
+ * @returns preview middleware configuration if found<br>
+ * Rejects if neither yamlConfig nor basePath is provided or if the file can't be read
+ */
+export async function getPreviewMiddleware(
```
tobiasqueck, line 19: If a function is marked as `async` then you don't need to wrap errors into a `Promise.reject`, you can just throw the error.
With that in mind, I think this function can be simplified.

tobiasqueck (changes requested):
* changeset correct
* implementation is very nicely structured and well document, therefore easy to understand
* excellent test coverage, and easy to understand tests, thanks to the fixtures
* two issues that need attention (separate comments)

Once the issues are fixed, I am happy to approve.

packages/app-config-writer/package.json
heimwege: adjusted in d7e5e5c1bed316c6774e4ed3d2389ae03b8f05e2 and c43b05a9b346c5d4498b14828d17d8bfa81d99c3

packages/app-config-writer/src/variants-config/utils.ts
heimwege: fixed in bff79e66c647e99f33bfef00a58bc2d82eba2999

tobiasqueck (approved):
requested changes applied

Klaus-Keller (commented):
Hi @ytpo-lyne, 
I've tested the command `add variants-config` and picked accidentally a project, that has already script `start-variants-management`.

As result an error message is shown:
```
Error while executing add variants-config: Script 'start-variants-management' cannot be written to package.json. Script already exists.
```
Is this intended? 
For add `mockserver-config` we did a different approach, if the config already exists, we check whether it is up to date and adjust or skip it. This way, you can use the `create` module also to update parts of your project to the latest setting, which was also a significant reason for the `create` module. The alternative to get the latest and greatest project setup is to generate a new project, which is not ideal.

Could we add the check for up-to-date script `start-variants-management` here as well instead of stopping with an error message?

heimwege:
> Hi @ytpo-lyne, 
> I've tested the command `add variants-config` and picked accidentally a project, that has already script `start-variants-management`.
> 
> As result an error message is shown:
> ```
> Error while executing add variants-config: Script 'start-variants-management' cannot be written to package.json. Script already exists.
> ```
> Is this intended? 
> For add `mockserver-config` we did a different approach, if the config already exists, we check whether it is up to date and adjust or skip it. This way, you can use the `create` module also to update parts of your project to the latest setting, which was also a significant reason for the `create` module. The alternative to get the latest and greatest project setup is to generate a new project, which is not ideal.
> 
> Could we add the check for up-to-date script `start-variants-management` here as well instead of stopping with an error message?
> 

@cfg74 does the remark from Klaus change your initial decision about how to handle an existing start-variants-management script?

heimwege:
@Klaus-Keller discussed with @cfg74. Would it be ok for you if we adjust the behavior in case of an existing script in [converter](https://github.com/SAP/open-ux-tools/tree/feat/2172/convert-local-preview-files) follow-up branch? There will be anyway adjustments in the create variants config for the converter but this PR is a prerequisite to be able to continue working on the converter so we want to merge is asap.

Klaus-Keller (approved):
* code to add variants configuration looks good
* changeset exists
* test coverage looks good
* manual test brought up the issues mentioned in https://github.com/SAP/open-ux-tools/pull/2446#pullrequestreview-2389625617, but these will be addressed with a follow up PR

Approved from my side.

PR #2455, opened 2024-10-11
feat: Support multiple services for ui5 yaml files and manifest.json

broksy:
Allows to add multiple services in YAML file configurations and Manifest (and multiple annotations per service).
Previous logic only rewrote whole mockserver config. This is still possible, but through an optional parameter.
During mockserver config generation `localService/data` folder (mockdataPath) is also generated if it does not exist in project.


Klaus-Keller (changes requested):
Hi @broksy,

I've tested adding a mockserver and found following issues:

In interactive mode there is no text for overwrite, it just gives the technical information `questions.overwrite › (y/N)`:
```
✔ questions.pathToMock › mainService: /sap/opu/odata4/dmo/sb_travel_mdsk_o4/srvd/dmo/sd_travel_mdsk/0001/
? questions.overwrite › (y/N)
```
I can see a maintained [text](https://github.com/SAP/open-ux-tools/pull/2455/files#diff-5098140c4b2b86b24ddb0b9170b27ae702f527e2bd39c5e142ca35577999197aR4) "Overwrite existing services", but can't see it in the question when running on command line.

After mockserver config has been added to a project, the empty file`webapp/localService/data/keep` is stored in that project. Is this intentionally? I don't think we should add the empty `keep` file. We can create the folder empty, but in my opinion it is then for the developer to decide, what to do with the empty folder.

There is an issue with the unit tests in `@sap-ux/mockserver-config-writer`
image

Klaus-Keller:
> Hi @broksy,
> 
> I've tested adding a mockserver and found following issues:
> 
> In interactive mode there is no text for overwrite, it just gives the technical information `questions.overwrite › (y/N)`:
> 
> ```
> ✔ questions.pathToMock › mainService: /sap/opu/odata4/dmo/sb_travel_mdsk_o4/srvd/dmo/sd_travel_mdsk/0001/
> ? questions.overwrite › (y/N)
> ```
> 
> I can see a maintained [text](https://github.com/SAP/open-ux-tools/pull/2455/files#diff-5098140c4b2b86b24ddb0b9170b27ae702f527e2bd39c5e142ca35577999197aR4) "Overwrite existing services", but can't see it in the question when running on command line.

I found the issue with i18n, there was a conflict due to global initialization of i18n in `@sap-ux/cap-config-writer` and `@sap-ux/mockserver-config-writer`. I switched now to an own instance of i18n for `@sap-ux/mockserver-config-writer`, see commit https://github.com/SAP/open-ux-tools/pull/2455/commits/5162ea34229ba8e6fc86b8fb1a4ea812a7f8ee7b.

packages/odata-service-writer/src/types.ts
```diff
@@ -71,7 +71,7 @@ export interface OdataService {
     /**
      * Annotations can either be EDMX annotations or CDS annotations.
      */
-    annotations?: EdmxAnnotationsInfo | CdsAnnotationsInfo;
+    annotations?: EdmxAnnotationsInfo[] | CdsAnnotationsInfo[];
```
IainSAP, line 74: This is a breaking type change. `@sap/fiori-generator` will need to be updated to accomodate, or better if you continue to support backward compatible single entries

broksy: Ok, I changed it back and kept array logic.

packages/fiori-elements-writer/test/__snapshots__/alp.test.ts.snap
```diff
@@ -2338,6 +2338,10 @@ xmlns:edmx=\\"http://docs.oasis-open.org/odata/ns/edmx\\">
 </edmx:Edmx>",
     "state": "modified",
   },
+  "webapp/localService/data/keep": Object {
+    "contents": "",
+    "state": "modified",
+  },
```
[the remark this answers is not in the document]
broksy: `mem-fs` does not provide an option to create empty directory.
Current solution is to use empty files and clean them after the commit (from client side).

PR #2457, opened 2024-10-11
fix(project-access): check CAP root for srv with app folder target

korotkovao:
Prevent the user from generating Fiori project into a CAP project folder, warn with the following message: `The selected project folder is part of an existing CAP project.  This application must therefore use the local CAP project as the data source`
Including for `CAPproject/app` folder as a target


korotkovao:
Validated the fix manually and it works for me with different CAP project destinations 

image

PR #2449, opened 2024-10-09
feat(abap-deploy-config): update abap deploy config prompt options

cianmSAP:
#2442 

Updates `AbapDeployConfigPromptOptions` to have prompt specific types

PR #2462, opened 2024-10-14
fix(abap-deploy-config): fix backend target choices check

cianmSAP:
Introduced in https://github.com/SAP/open-ux-tools/pull/2449 .. when no backend target exists an error is thrown trying to replace trailing slash


longieirl (approved):
Changes look good.
Test coverage is good.
Changese is good.

kjose90 (approved):
LGTM 
changeset and tests ✅

Issue #2140, opened 2024-07-10
BUG - Values in the UI.Dropdown control is not center aligned once selected


hitesh-parmar:
### Related Feature
Feature request: _issue number_

### Description

I noticed that the selected value in the UI.Dropdown control is not center aligned which is easy visible when the control is surrounded by other controls.

### Steps to Reproduce

1. Open Guided Development for any project
2. Notice that values in the Group By and View By control are not center aligned

The same is also true for the Page Map toolbar as the same control is used there.

You can also check the issue by opening the storybook `Start UI Components StoryBook`

### Expected results

The value selected in the dropdown is center-aligned.

### Actual results

The value selected in the dropdown is not center-aligned.

### Screenshots
If applicable, add screenshots to help explain the problem.

![image](https://github.com/SAP/open-ux-tools/assets/24454497/4cf76cd8-1041-4e9b-9dab-4dadb233b26c)

![image](https://github.com/SAP/open-ux-tools/assets/24454497/d7d374ea-0157-4dbe-bd3b-051825a92ccf)

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2464, opened 2024-10-15
fix(ui-components): #2140 - fix dropdown vertical align input field

eouin:
BUG - Values in the UI.Dropdown control is not centered aligned (vertically)

issue #2140 

- [x] remove the border top and bottom from the total height to get the line-height which should the full height - 2px

### DIFF
--- a/.changeset/fresh-hotels-tap.md
+++ b/.changeset/fresh-hotels-tap.md
@@ -1,0 +1,5 @@
+---
+"@sap-ux/ui-components": patch
+---
+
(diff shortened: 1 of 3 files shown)


815are (approved):
checked:
1. code change - style fix is applied to consider border when line-height is set
2. changeset persists
3. tested with storybook

Klaus-Keller (approved):
Thanks @eouin,
- fix for style issue looks good
- changeset exists
- did just a visual review, no manual test

approved from my side.

PR #2465, opened 2024-10-15
fix: revert major version update

emrecevik06:
Revert major version update


kranthie-sap (approved):
Changes to the package.json and the changeset look good. This change should revert the major version update and do a patch update of control property editor and dependent packages.

Jimmy-Joseph19 (approved):
Thanks @emrecevik06  and @kranthie-sap

PR #2466, opened 2024-10-15
fix(ui-components): UIComboBox, UIDropdown: Allow extending 'calloutProps' instead of overwriting.

815are:
We noticed problem with `UIComboBox`, `UIDropdown` - if we pass callout props to them, then all `calloutProperties` are overwritten instead of merging it with properties defined within `UIComboBox`, `UIDropdown` components

### DIFF
--- a/.changeset/witty-carrots-give.md
+++ b/.changeset/witty-carrots-give.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui-components': patch
+---
+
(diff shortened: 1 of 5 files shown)


Jimmy-Joseph19 (approved):
Changes for merging props looks good.
Covered scenario with unit test.
Changeset found.
I haven't tested locally.

broksy (approved):
Did not test locally, but code looks good and changeset reflects to the changes made.
Approved from my side.

PR #2468, opened 2024-10-16
Fix: Zero State of Changes Panel

emrecevik06:
Add a zero state for changes panel. 
Before:
image

After: 
image


vinayhospete (dismissed):

packages/control-property-editor/src/i18n/i18n.json
```diff
@@ -73,13 +73,14 @@
   "SHOW_CHANGES": " the app preview to show those changes.",
   "CONFIRM_CHANGE_SUMMARY_DELETE_SUBTEXT": "Are you sure you want to delete the change for this property? This action cannot be undone.",
   "CONFIRM_OTHER_CHANGE_DELETE_SUBTEXT": "Are you sure you want to delete this {{name}} change? This action cannot be undone.",
-  "NO_CONTROL_CHANGES_FOUND": "No Control Changes Found",
+  "NO_CHANGES_FOUND": "No historic changes",
+  "MODIFY_CHANGES": "This application was not modified yet",
```
voicis, line 77: Text keys should be a bit more descriptive and it seems a bit misleading as it seems it is related to making some changes not the description of the `NO_CHANGES_FOUND`. Since we have two related texts now instead perhaps we can have something like this `NO_CHANGES_FOUND_TITLE` and `NO_CHANGES_FOUND_DESCRIPTION`.

packages/control-property-editor/src/panels/changes/ChangesPanel.module.scss
```diff
@@ -8,18 +8,29 @@
...
     text-align: center;
-    padding-top: 20px;
-    margin-left: 35px;
     color: var(--vscode-foreground);
 }
 
-.infoIcon {
-    margin-left: 15px;
-    margin-top: 5px;
+.modifyChanges {
```
voicis, line 25: Same as above, the class name here is also a bit confusing and misleading.

voicis (changes requested):
Top padding is missing. It should be similar to the no control changes with some.
image

voicis (approved):
- changeset present
- code changes looks good
- review comments addressed
- good test coverage
- tested locally

vinayhospete (approved):

PR #2479, opened 2024-10-18
fix(launch-config): type error for empty json

ytpo-lyne:
Fix for type error when launch.json file is empty.


heimwege (approved):
changeset ok
excellent test coverage
fix is straight forward

Klaus-Keller (approved):
Thanks @ytpo-lyne!
- fix for empty launch config looks good
- changeset exists
- test coverage is great

Approved from my side

PR #2480, opened 2024-10-21
fix: updateVocabularies

gernotkuhr:
- update odata and sap-ux vocabularies


vinayhospete (approved):

marufrasully (approved):
Thanks @gernotkuhr 
* changes is straight forward
* did not test it locally

PR #2484, opened 2024-10-22
tbi(ui5-app-inq): update ui5 default version logic

cianmSAP:
Updates default handling for ui5 version

- If default version is a snapshot, the logic is unchanged, i.e. the closest published version will be returned as the default
- If default version is not found in the version list returned from ui5-info, it will be added to the top of the list (outside of the maintained, out of maintenance seperators)

### DIFF
--- a/packages/inquirer-common/src/prompts/utility.ts
+++ b/packages/inquirer-common/src/prompts/utility.ts
@@ -1,7 +1,7 @@
 import { getUi5Themes, type UI5Theme, type UI5Version } from '@sap-ux/ui5-info';
 import * as fuzzy from 'fuzzy';
 import type { ListChoiceOptions } from 'inquirer';
-import { coerce, eq, lte } from 'semver';
+import { coerce, eq, lte, type SemVer } from 'semver';
 import { t } from '../i18n';
 import type { UI5VersionChoice } from '../types';
 import { Separator } from './separator';
@@ -44,7 +44,7 @@ export function searchChoices(searchVal: string, searchList: ListChoiceOptions[]
  *
  * @param versions ui5Versions
  * @param includeSeparators Include a separator to visually identify groupings, if false then grouping info is included in each entry as additional name text
- * @param defaultChoice optional, provides an additional version choice entry that is added as the first entry in the version choices and sets as the default
+ * @param defaultChoice optional, provides an additional version choice entry that is added as the first entry in the version choices (if it does not already exist in the version list) and sets as the default
  * @returns Array of ui5 version choices and separators if applicable, grouped by maintenance state
  */
 export function ui5VersionsGrouped(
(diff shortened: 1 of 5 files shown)


packages/inquirer-common/src/prompts/utility.ts
```diff
@@ -95,27 +95,60 @@ export function ui5VersionsGrouped(
...
+ * @param defaultChoice - optional default choice to be used
+ * @returns The default UI5 version choice or the latest provided version
  */
 export function getDefaultUI5VersionChoice(
     ui5Versions: UI5Version[],
     defaultChoice?: UI5VersionChoice
 ): UI5VersionChoice | undefined {
+    let useClosestVersion = false;
     if (defaultChoice) {
+        if (defaultChoice.value.endsWith('SNAPSHOT')) {
```
IainSAP, line 146: Why caps?

packages/inquirer-common/src/prompts/utility.ts
```diff
@@ -95,27 +95,60 @@ export function ui5VersionsGrouped(
...
 
+/**
+ * Will either return the UI5 version that is the closest to the default choice version or simply the version that is passed to it.
+ *
+ * @param ui5Versions - UI5 versions ordered by version descending latest first
+ * @param useClosestVersion - whether to use the closest available version to the default choice
+ * @param defaultChoiceVersion - default version chosen
+ * @returns - ui5 version to be used as default
+ */
+function getUI5Version(
```
IainSAP, line 115: Ambiguous name. 
```suggestion
function findDefaultUI5Version
```

packages/inquirer-common/src/prompts/utility.ts
```diff
@@ -95,27 +95,60 @@ export function ui5VersionsGrouped(
 
     const versionChoices = [...maintChoices, ...notMaintChoices];
     if (defaultChoice) {
-        versionChoices.unshift(defaultChoice);
+        // adds or moves default choice to the top of the list
```
IainSAP, line 98: Is this comment accurate, we dont appear to be moving anything but unshifting the found version.

packages/inquirer-common/src/prompts/utility.ts
cianmSAP: I've only seen upper case, but thanks I've updated

packages/inquirer-common/src/prompts/utility.ts
cianmSAP: Updated

packages/inquirer-common/src/prompts/utility.ts
cianmSAP: Sorry the comment was written for a previous implementation but wasn't edited. 
Yes relying on the default method to work

IainSAP (approved):
Thanks @cianmSAP 

- Comments addressed
- Tests added to cover
- Did not test locally

kjose90 (approved):
LGTM ✅ 
changes are clear 
changeset included
test coverage is good

devinea (approved):

PR #2485, opened 2024-10-23
Fix: npm i after executing a create command fails on Windows

SRKPendyala:
We've fixed now create command for Windows OS


tobiasqueck:
@SRKPendyala the fix looks good, however, could we think about, completely removing the implementation and call https://github.com/SAP/open-ux-tools/blob/16041f41724e07fee6e366c751d95399c910d5e1/packages/project-access/src/command/npm-command.ts#L13 instead? 
It looks like the same issues was considered they already.

@Klaus-Keller opinions?

cfg74:
Please also check with node 18 and node 20.

Klaus-Keller:
> @SRKPendyala the fix looks good, however, could we think about, completely removing the implementation and call
> 
> https://github.com/SAP/open-ux-tools/blob/16041f41724e07fee6e366c751d95399c910d5e1/packages/project-access/src/command/npm-command.ts#L13
> 
> instead?
> It looks like the same issues was considered they already.
> @Klaus-Keller opinions?

Yes, I agree. Module `@sap-ux/create` has already dependency to `@sap-ux/project-access`, so I think we should export `execNpmCommand()` and consume it here. There are two relevant options for Windows: `windowsVerbatimArguments` and `shell`, both are already considered in the existing implementation. 

Please consume it from:
https://github.com/SAP/open-ux-tools/blob/f9ffbe0fad90f441cf74e2393f408c73908bc94e/packages/project-access/src/command/npm-command.ts#L13

SRKPendyala:
> @SRKPendyala the fix looks good, however, could we think about, completely removing the implementation and call
> 
> https://github.com/SAP/open-ux-tools/blob/16041f41724e07fee6e366c751d95399c910d5e1/packages/project-access/src/command/npm-command.ts#L13
> 
> instead?
> It looks like the same issues was considered they already.
> @Klaus-Keller opinions?

Done

SRKPendyala:
> Please also check with node 18 and node 20.

Done

Klaus-Keller:
Changes look good now, but the pipeline failed. I'm wondering whether this has to do with the fork this changes are on? Try to re-run the pipeline.

Klaus-Keller:
No, did not help to re-run. @SRKPendyala, can you check from your end why the pipeline fails to build?

SRKPendyala:
> No, did not help to re-run. @SRKPendyala, can you check from your end why the pipeline fails to build?

@Klaus-Keller :  Created below pull request to fix this build issues - https://github.com/SAP/open-ux-tools/pull/2508

tobiasqueck:
@Klaus-Keller @SRKPendyala 

> > No, did not help to re-run. @SRKPendyala, can you check from your end why the pipeline fails to build?
> 
> @Klaus-Keller : Like discussed, created a branch fix/create instead of forking the repo. Created below pull request - #2508

Closing this one

PR #2486, opened 2024-10-23
chore: add eslint warning for unsafe assignments

heimwege:
There are quite some unsafe assignments in open-ux-tools and it might help avoiding issues if we get at least warned about them as type any prevents the type checker from actually doing its work.


devinea (approved):
Adding new eslint rule is useful addition.

ytpo-lyne (approved):
Helpful enhancement to avoid unsafe assignments. Approving.

Issue #2472, opened 2024-10-17
FEATURE - UI5 reference library generator


cianmSAP:
_As a developer,
I want to use a yeoman generator module,
so that I can be prompted to select a ui5 library to be referenced in a  generate the library_

### Description
Prompting `@sap-ux/ui5-library-reference-inquirer` and writing `@sap-ux/ui5-library-reference-writer` are published modules in open-ux-tools, this module will encompass both into the ui5 library sub generator 

### Technical Design
Additions required to `@sap-ux/fiori-generator-shared`
This module will be consumed as a subgenerator i.e it will not be usable as a general generator by yeoman (this is done by not prefixing the module with `generator-`

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #2489, opened 2024-10-23
feat(fiori-gen): add telemetry helper class and associated functions

cianmSAP:
#2472

Adds some common utilities to `@sap-ux/fiori-generator-shared` in prerparation for the reference library generator


packages/fiori-generator-shared/src/environment.ts
```diff
@@ -10,3 +13,16 @@ export function isCli(): boolean {
         return false;
     }
 }
+
+/**
+ * Determine if the current prompting environment is cli or a hosted extension (app studio or vscode).
+ *
+ * @returns the platform name and technical name
+ */
+export function getHostEnvironment(): { name: string; technical: string } {
```
IainSAP, line 22: Doesnt this already exist elsewhere in open source?

cianmSAP: It's in odata service inquirer. But I attempted to make updates to this before and you didn't want contributions in the module yet.

packages/fiori-generator-shared/src/installedCheck.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Check for an installed extension, optionally specifying a minimum version.
+ * Note, this does not check for activation state of specified extension.
+ *
+ * @param vscode - vscode instance
+ * @param extensionId - the id of the extension to find
+ * @param minVersion - the minimum version of the specified extension, lower versions will not be returned. Must be a valid SemVer string.
+ * @returns true if the extension is installed and the version is >= minVersion (if provided), false otherwise
+ */
+export function isExtensionInstalled(vscode: any, extensionId: string, minVersion?: string): boolean {
```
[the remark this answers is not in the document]
cianmSAP: It will be need in the reference libary generator, I added it while I was in here. And I can also make the updates when consuming back into tool-suite

packages/fiori-generator-shared/src/translations/fiori-generator-shared.i18n.json
```diff
@@ -8,5 +8,8 @@
     },
     "debug": {
         "loggingConfigured": "Logging has been configured at log level: {{logLevel}}"
+    },
+    "telemetry": {
+        "unknownOs": "Unknown"
```
longieirl, line 13: Is this correct? Should it be
```
"unknownOs": "UnknownOs"

File: packages/fiori-generator-shared/src/environment.ts (hunk 1 above)
Comment (IainSAP) [reply packages/fiori-generator-shared/src/environment.ts:22]: This is ok since its not impacting the code thats under development directly, its a utility. Please go ahead and update, thx @cianmSAP

File: packages/fiori-generator-shared/src/translations/fiori-generator-shared.i18n.json (hunk 3 above)
Comment (cianmSAP) [reply packages/fiori-generator-shared/src/translations/fiori-generator-shared.i18n.json:13]: No this is the existing telemetry property used, check `UNKNOWN_OS` in internal repo. So I assume this needs to remain the same for consistency

File: packages/odata-service-inquirer/src/utils/index.ts [hunk 4]
```diff
@@ -13,19 +12,6 @@ import type { ListChoiceOptions } from 'inquirer';
 
 const osVersionName = osName();
 
-/**
- * Determine if the current prompting environment is cli or a hosted extension (app studio or vscode).
- *
- * @returns the platform name and technical name
- */
-export function getHostEnvironment(): { name: string; technical: string } {
```
IainSAP, line 21: Since this returns different results that the one that has been added to `fiori-generator-shared` I dont think we can replace it?

packages/fiori-generator-shared/src/environment.ts
IainSAP: So the function that has been added is not doing the same thing and will return different results.

packages/odata-service-inquirer/src/utils/index.ts
cianmSAP: Reverted the changes

packages/fiori-generator-shared/src/telemetry/telemetryHelper.ts
```diff
@@ -0,0 +1,103 @@
...
+} from '@sap-ux/telemetry';
+import type { TelemetryData } from './types';
+import { getHostEnvironment } from '../environment';
+import osName from 'os-name';
+import { t } from '../i18n';
+
+/**
+ * Helper class for intialising and preparing event data for telemetry.
+ */
+export abstract class TelemetryHelper {
```
IainSAP, line 15: Nothing specific to Fiori generators in here, would this be better situated in `@sap-ux/telemetry` for re-use?

packages/fiori-generator-shared/src/telemetry/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+        eventName: generationEvent,
+        properties: telemetryData,
+        measurements: generationTime ? { GenerationTime: generationTime } : {}
+    };
+}
+
+/**
+ * Sends the telemetry event to the telemetry client.
+ *
+ * @param generationEvent - the event name to be reported
```
IainSAP, line 30: This could be any event, as its defined as a stirng?

cianmSAP: Fair point, I've updated to be `telemetryEventName`

packages/fiori-generator-shared/src/telemetry/telemetryHelper.ts
cianmSAP: We chatted about this on a previous call, I was of the understanding that because this is just a helper class which is used/shared across the sub generators, it was the appropriate place for it.

packages/fiori-generator-shared/src/telemetry/telemetryHelper.ts
```diff
@@ -0,0 +1,103 @@
...
+                Platform: getHostEnvironment().technical,
+                OperatingSystem: osVersionName
+            };
+        }
+
+        if (filterDups) {
+            const newTelemData = { ...this._telemetryData, ...additionalData };
+            if (
+                Math.abs(this._previousEventTimestamp - currentTimestamp) < 1000 &&
+                JSON.stringify(newTelemData) === JSON.stringify(this._telemetryData)
```
longieirl, line 71: Can the data set change? If soo, then this logic will return false.

cianmSAP: The aim is to filter duplicate events, so we want it to return false if the data is different

IainSAP (approved):
Thanks @cianmSAP 

- Comments address
- Test cover new code

PR #2494, opened 2024-10-24
fix(inquirer): adding i18n bundle of common module in consumers

cianmSAP:
https://github.com/SAP/open-ux-tools/issues/2496

Exposes function to add the resource bundle for the `@sap-ux/inquirer-common` module
This allows consuming modules to call this after initialising i18next and ensuring all translations are loaded at the time of prompting

### DIFF
--- a/packages/ui5-application-inquirer/src/i18n.ts
+++ b/packages/ui5-application-inquirer/src/i18n.ts
@@ -1,6 +1,7 @@
 import type { TOptions } from 'i18next';
 import i18next from 'i18next';
 import translations from './translations/ui5-application-inquirer.i18n.json';
+import { addi18nResourceBundle as addInquirerCommoni18nResourceBundle } from '@sap-ux/inquirer-common';
 
 const ui5AppInquirerNamespace = 'ui5-application-inquirer';
 export const defaultProjectNumber = 1;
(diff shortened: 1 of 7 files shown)


packages/ui5-application-inquirer/src/i18n.ts
```diff
@@ -1,25 +1,26 @@
...
+                defaultProjectNumber
             }
-        },
-        () => i18next.addResourceBundle('en', ui5AppInquirerNamespace, translations)
-    );
+        }
+    });
+    i18next.addResourceBundle('en', ui5AppInquirerNamespace, translations);
+    // add the inquirer common i18n resource bundle to ensure all translations are available
+    addInquirerCommoni18nResourceBundle();
```
longieirl, line 23: Do we need to add tests to ensure that this method is executed?

cianmSAP: Tests added

longieirl (approved):
Thanks @cianmSAP 
Changes look good.
Havent tested locally.
Tests cover change.

IainSAP (approved):
Thanks @cianmSAP 

- As discussed a follow-up TBI is required to remove the i18n loading weirdness (consuming modules have to load dependant bundles)
- Tests updated as expected
- Did not test locally

PR #2507, opened 2024-10-28
fix: add aria-label for UIFlexibleTable Add Row Button

emrecevik06:
Add aria-label for UIFlexibleTable `Add Row` Button


vadson71 (approved):
Changes look good.
Covered in tests.
Didn't test locally

PR #2509, opened 2024-10-29
fix(cpe): undo composite change

### LINKED ISSUE #2495: BUG - CPE, incorrect active/inactive status of an unsaved change when there is an unsaved composite change in the stack
### Description
In the CPE unsaved changes in the change stack may have active and inactive status after undo/redo operations.
If there is unsaved composite change in the stack then status of other consequent changes is displayed incorrectly.

### Steps to Reproduce
1. Open an ADP V4 project in the Adaptation Editor
2. In adaptation mode right click on the LR table filter bar and choose 'Configuration'
3. On RTA configuration dialog change value of two checkboxes and apply
4. In the change stack observe two unsaved changes
5. Right click on the table filter bar again and choose Add: Fragment. On the dialog enter fragment name and click Create. Observe corresponding third change in the stack
6. Press Undo. Observe no visual changes in the stack. Press Undo again. Observe all three changes are marked as disabled.
7. Press Redo. Observe all three changes are marked as enabled. Press Redo again, observe no visual changes.
8. Press save and reload the app preview.
9. In the saved changes stack you can delete all three changes one by one only, despite the first two were produced as composite change and partial deletion of composite changes set may lead to errors in common case

### Expected results
Undo/redo should change changes status sequentially for each independent change
Changes which are members of one composite change should be deleted with one go.

### Actual results
Composite change affects status displaying of consequent changes
Members of one composite change can be deleted one by one only

OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

voicis:
Fixes #2495 

`getAllExecutedCommands` returns flattened array of executed changes, but `getCommands` is not flattened, so the number of inactive changes was incorrectly calculated.

`isActive` property was not correctly handled in `UnkownChange` and `ControlChange` components.


nikmace (approved):
Code looks good
Coverage is good
Tested manually, reload button appears

PR #2513, opened 2024-10-31
fix: Create Custom Table Column quick action bug fix

vadson71:
Fixed bug. Click on quick action doesn't work with Analytical/Grid/Tree tables in v2

### DIFF
--- a/.changeset/young-coins-marry.md
+++ b/.changeset/young-coins-marry.md
@@ -3,4 +3,4 @@
 '@sap-ux/preview-middleware': patch
 ---
 
-Add Custom Table Column quick action bug fix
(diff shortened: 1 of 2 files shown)


.changeset/young-coins-marry.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ux-private/preview-middleware-client': patch
+'@sap-ux/preview-middleware': patch
+---
+
+Add Custom Table Column quick action bug fix
```
voicis, line 6: This is very vague and it is not clear what issue was actually fixed.

vadson71: Improved

nikmace (approved):
Change is simple
Changeset looks good
Did not test manually

Issue #2514, opened 2024-10-31
FEATURE - Intercept console error messages from the iframe in the React app


nikmace:
_As a **developer**,
I want to **be able to see important messages from the browser's console in the Info Center**,
so that I **have the errors consolidated in one place and be user friendly**_

### Description

_What is the high level purpose and value of the feature?  The description should ensure that anyone reading understands what it is doing (at a high level) and most importantly why._

We need to intercept the console calls from the iframe in order to show them in the Info Center, which is being implemented in this PR: #2511.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have started the Editor
**when** I the application loads
**then** I will see all the filtered errors in the Info Center

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #2515, opened 2024-10-31
feat: Intercept console error messages from the iframe in the React app

nikmace:
Feat for #2514.

- Add console interceptor for iframe, still logs messages in the iframe.
- Adds filtration for messages.


nikmace:
This is no longer needed, will close the PR.

PR #2516, opened 2024-10-31
fix(ui5-info): fix fallback logic for retrieving maintained versions

cianmSAP:
With changes in https://github.com/SAP/open-ux-tools/pull/2189 to move from using `versionoverview` the try/catch + fallback for offline or errors was removed. 

Updates in this PR : 
- `supportedUi5VersionFallbacks` now also returns property `support` as it is useful for the consumer (for labelling and grouping versions)
-  when the option `includeMaintained` is passed, the call to `retrieveUI5VersionsCache` is wrapped in a try/catch and will fallback to `supportedUi5VersionFallbacks`

### DIFF
--- a/.changeset/fuzzy-toes-talk.md
+++ b/.changeset/fuzzy-toes-talk.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui5-info': patch
+---
+
+fix fallback logic for retrieving maintained versions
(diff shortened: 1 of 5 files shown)


IainSAP (approved):
Thx @cianmSAP 

- Updates look appropriate (did not test locally)
- Tests updated to cover

devinea (approved):
code changes and tests look good. 
changeset ✅

Klaus-Keller (approved):
Thanks @cianmSAP!
* fix to fallback to static list of UI5 versions looks good
* changeset valid
* test coverage is good
* performed a manual test without internet connection and did not find any issues

Approved from my side.

Issue #2473, opened 2024-10-17
BUG - Store access is outputting errors which should be warnings


korotkovao:
### Related Feature
Feature request: _issue number_

### Description
Using `yo @sap/fiori` if the user does not allow access to the keystore for existing systems, errors are shown. In addition axios extension is incorrectly logging low level outputs as info polluting the CLI.

### Expected results
When the user denies it is not an error

### Screenshots
![image](https://github.com/user-attachments/assets/9403fed1-ae56-40ea-a263-c9c33af457c6)

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2517, opened 2024-10-31
fix(store): no error message for denied keychain access

korotkovao:
[BUG - Store access is outputting errors which should be warnings](https://github.com/SAP/open-ux-tools/issues/2473#top)
#2473


packages/store/src/secure-store/keytar-store.ts
```diff
@@ -53,7 +53,9 @@ export class KeytarStore implements SecureStore {
         try {
             return (await this.keytar.findCredentials(service)).reduce((result, entry) => {
                 try {
-                    result[String(entry.account)] = JSON.parse(entry.password);
+                    if (entry.account) {
```
IainSAP, line 56: Could you please add a comment about when a user denies permission to access the keystore that the `entry.account` is undefined (if that is whats happening?), thx

korotkovao: Thank you, added the comment.

devinea (approved):
Code changes are clear
changeset ✅

Klaus-Keller (approved):
Thanks @korotkovao!
* code change to avoid error when key is not present look good
* changeset exists
* test coverage is great

Approved from my side

IainSAP (approved):
thx @korotkovao 

- Comments addressed
- Tests added to cover

korotkovao:
Thank you @devinea @Klaus-Keller @IainSAP

Issue #2535, opened 2024-11-06
BUG - Higher layer changes broke CPE


GDamyanov:
### Related Feature
Feature request: _issue number_

### Description
Currently, in CPE if the base application contains higher level changes or there are existing higher layer changes then the current instance of RTA is destroyed and all components that are coming from RTA (outline tree, properties pane) are not visible.

### Steps to Reproduce
1. Create ADP project in VENDOR layer. 
2. Start CPE.
3. Create some change.
4. Update layer on the created change to some higher one for example CUSTOMER_BASE.
5. Refresh CPE

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2536, opened 2024-11-06
fix: Handle higher layer changes

GDamyanov:
fix for #2535 

- Handle higher layer changes in CPE


packages/preview-middleware-client/src/flp/init.ts
```diff
@@ -351,3 +357,28 @@ if (bootstrapConfig) {
...
+ * When RTA detects higher layer changes an error with Reload triggered text is thrown, the RTA instance is destroyed and the application is reloaded.
+ * For UI5 version lower then 1.84.0 RTA is showing a popup with notification text about the detection of higher layer changes.
+ *
+ * @param error the error thrown when there are higher layer changes when starting UI Adaptation.
+ * @param ui5VersionInfo ui5 version info
+ */
+export async function handleHigherLayerChanges(error: unknown, ui5VersionInfo: Ui5VersionInfo): Promise<void> {
+    const err = getError(error);
+    if (err.message.includes('Reload triggered')) {
+        if (ui5VersionInfo && ui5VersionInfo.minor >= 84 && ui5VersionInfo.major >= 1) {
```
nikmace, line 372: Would you be able to use this method to stay consistent:

```suggestion
        if (!isLowerThanMinimalUi5Version(ui5VersionInfo, { major: 1, minor: 84 })) {
```

packages/preview-middleware-client/src/flp/init.ts
```diff
@@ -351,3 +357,28 @@ if (bootstrapConfig) {
...
+            const bundle = await getTextBundle();
+            const action  = showMessage({
```
nikmace, line 374: Could you please prettier this file to get rid of this double space?

packages/preview-middleware-client/test/unit/flp/init.test.ts
```diff
@@ -13,6 +13,28 @@ import { fetchMock, sapMock } from 'mock/window';
...
+    }
+});
+
+Object.defineProperty(window, 'location', {
+    value: {
+        ...window.location,
+        reload: jest.fn()
+    },
+    writable: true
+});
```
nikmace, line 36: When you define a property like this, you need to reset it to the original value at the end of the test in order not to break the other chain of tests that uses the same code.

I would suggest to move this inside the new test case and also clear this at the end of the expect statement.

Example:

```ts
const reloadSpy = jest.fn();
const location = window.location;
Object.defineProperty(window, 'location', {
     value: {
         reload: reloadSpy
     }
);

...

expect(...)
Object.defineProperty(window, 'location', {
     value: location
});
```

nikmace (changes requested):

packages/preview-middleware-client/test/unit/flp/init.test.ts
GDamyanov: Fixed in [c2c0266](https://github.com/SAP/open-ux-tools/pull/2536/commits/c2c0266b492ca697c56288d62a0cd94ec0933034)

packages/preview-middleware-client/src/flp/init.ts
GDamyanov: Fixed in [c2c0266](https://github.com/SAP/open-ux-tools/pull/2536/commits/c2c0266b492ca697c56288d62a0cd94ec0933034)

packages/preview-middleware-client/src/flp/init.ts
GDamyanov: Fixed in [6d6f900](https://github.com/SAP/open-ux-tools/pull/2536/commits/6d6f900315c8f0446b6829c935ec360bf8f506c2)

packages/preview-middleware-client/test/unit/flp/init.test.ts
voicis: This issue doesn't seem to be resolved yet. It probably shouldn't be set in global scope, but just before/within tests where it is actually needed. Also cleanup should be in the `after*` blocks, because the test might fail and the code at the end of the test it self might never run and leave the environment in an inconsistent state.

packages/preview-middleware-client/test/unit/flp/init.test.ts
```diff
@@ -298,5 +320,56 @@ describe('flp/init', () => {
...
+                });
+                expect(reloadSpy).toHaveBeenCalled();
+                Object.defineProperty(window, 'location', {
+                    value: location
+                });
+            }, 5 * 1000);
+        });
     });
-});
+});
```
voicis, line 375: It looks like the file is not correctly formatted either, please do so.

packages/preview-middleware-client/test/unit/flp/init.test.ts
```diff
@@ -298,5 +320,56 @@ describe('flp/init', () => {
...
+            const mockService = { attachAppLoaded: jest.fn() };
+            sapMock.ushell.Container.getServiceAsync.mockResolvedValueOnce(mockService);
+
+            await rendererCb();
+
+            const loadedCb = mockService.attachAppLoaded.mock.calls[0][0] as (event: unknown) => Promise<void>;
+            await loadedCb({ getParameter: () => { } });
+
+            setTimeout(() => {
+                expect(sendActionSpy).toHaveBeenCalled();
```
voicis, line 359: Is this actually checked? It happens after the test block has run, so I would expect Jest to just ignore them.

Why do we need `setTimeout`? We already have an async function so we should be able to test the asynchronous code in here properly.

packages/preview-middleware-client/src/flp/init.ts
```diff
@@ -351,3 +357,28 @@ if (bootstrapConfig) {
         Log.error('Sandbox initialization failed: ' + error.message);
     });
 }
+
+/**
+ * Handle higher layer changes when starting UI Adaptation.
+ * When RTA detects higher layer changes an error with Reload triggered text is thrown, the RTA instance is destroyed and the application is reloaded.
+ * For UI5 version lower then 1.84.0 RTA is showing a popup with notification text about the detection of higher layer changes.
```
heimwege, line 364: typo: `lower than`

packages/preview-middleware-client/src/messagebundle.properties
```diff
@@ -37,3 +37,5 @@ ADP_REUSE_COMPONENTS_MESSAGE = Reuse components are detected for some views in t
 CPE_CHANGES_VISIBLE_AFTER_SAVE_AND_RELOAD_MESSAGE = Note: The change will be visible after save and reload. 
 
 TABLE_ROWS_NEEDED_TO_CREATE_CUSTOM_COLUMN=At least one table row is required to create new custom column. Make sure the table data is loaded and try again.
+
+HIGHER_LAYER_CHANGES_INFO_MESSAGE=The application was reloaded because of changes in a higher layer.
```
heimwege, line 41: Question: Can this text really be understood by a user? I have to admit I don't know what "higher layer" actually means.

heimwege (commented):
one minor finding and a question

packages/preview-middleware-client/test/unit/flp/init.test.ts
GDamyanov: Fixed in [652a18e](https://github.com/SAP/open-ux-tools/pull/2536/commits/652a18e63fd64cf9f64e2a3b65d78254075b0fbb)

packages/preview-middleware-client/test/unit/flp/init.test.ts
GDamyanov: Fixed in [652a18e](https://github.com/SAP/open-ux-tools/pull/2536/commits/652a18e63fd64cf9f64e2a3b65d78254075b0fbb)

packages/preview-middleware-client/src/flp/init.ts
GDamyanov: Fixed in [652a18e](https://github.com/SAP/open-ux-tools/pull/2536/commits/652a18e63fd64cf9f64e2a3b65d78254075b0fbb)

packages/preview-middleware-client/src/messagebundle.properties
GDamyanov: Fixed in [a584e48](https://github.com/SAP/open-ux-tools/pull/2536/commits/a584e48923d502d5f4c5e021fd806ef74adfac63)

heimwege (approved):
- Review comments addressed
- test coverage is good
- changeset ok
- did NOT test locally

nikmace (approved):
Code looks good
Coverage is alright
Did not test manually

voicis (approved):
- Changes looks good
- Review comments addressed
- Good test coverage
- Did not test

PR #2538, opened 2024-11-07
fix: table action when variantmanagement is not enabled for v4

nahmed22:
in V4 application, enable table action, even if the `variant managemen`t is not available.


Jimmy-Joseph19 (approved):
Changes looks good.
Changeset found
Haven't tested locally.

packages/preview-middleware-client/src/adp/quick-actions/fe-v4/table-quick-action-base.ts
```diff
@@ -24,17 +24,20 @@ export abstract class TableQuickActionDefinitionBase {
     constructor(
         public readonly type: string,
         protected readonly defaultTextKey: string,
-        protected readonly context: QuickActionContext
+        protected readonly context: QuickActionContext,
+        protected readonly skipVariantManagement?: boolean
```
vadson71, line 28: Maybe better to name this parameter like isSkipVariantManagementCheck or isIgnoreVariantManagement
because it is boolean flag, its name should not sound like command

nahmed22: adapted `isSkipVariantManagementCheck `

vadson71 (approved):
Changes look good.
Test coverage is good
Comments addressed
Didn't test locally

Jimmy-Joseph19 (approved):

PR #2546, opened 2024-11-08
fix(cf-deploy-config-writer): cleanup based on local testing when using headless

longieirl:

- updated snapshots to reflect latest changes
- mbt should only be added if it the `package.json` is updated scripts that are dependent on it, it was being added regardless before
- tested locally against headless flow in deployment-generator
- cleanup readme


cianmSAP (dismissed):
Upates look good
I've not tested locally
Tests to cover

kjose90 (approved):
@longieirl code changes are good and clear
havent tested locally
tests are good ✅

cianmSAP (approved):
Re-approving

PR #2548, opened 2024-11-08
fix(store): warn instead of error

korotkovao:
## #2473


packages/store/src/secure-store/keytar-store.ts
```diff
@@ -58,7 +58,7 @@ export class KeytarStore implements SecureStore {
                         result[String(entry.account)] = JSON.parse(entry.password);
```
cianmSAP, line 58: I debugged this code an checking that `entry. account` is defined should be sufficient. 
I think this needs more investigation

korotkovao:
This is not needed, as PR https://github.com/SAP/open-ux-tools/pull/2517 achieved no `error`s logged for denied keychain access. Closing this PR.

PR #2566, opened 2024-11-14
chore: only run component build from preview middleware

voicis:
After 8906662ef7005c75b32069bad9ac820b0b853e66 build is failing because references from `@sap-ux-private/control-property-editor-common` could not be resolved. We do not actually need to run type checking from `preview-middleware` as we only need to copy the artefacts over, so we can run only component build and relay on type checking being done by `preview-middleware-client` build.

Included changeset from the failed release build.


devinea (approved):

voicis:
After restarting the action again, it worked.

PR #2568, opened 2024-11-14
fix(ui-components): UITranslationInput. Screen readers read the resolved translation tooltip/title of UITranslationInput as part of the screen reader speech

815are:
accessibility team reported issue, that additional information is read by screen reader:
image

It is about part with  `Value: .... Translation: ...`.

This text is coming from `title` attribute - we resolve by purpose within `UITranslation` input to show it in tooltip.
As solution - move `title` attribute from input to wrapper element. In result tooltip appears as before, but screen reader ignores it.

image

from screen reader
image


vadson71 (approved):
Changes look good.
Good test coverage
Didn't test locally

iha89 (approved):
Changes look good to me with a proper test coverage.
Tested with mac VoiceOver

Jimmy-Joseph19 (approved):
Changes looks good.
Test coverage good
Changeset found.
Didn't test locally

Issue #2518, opened 2024-10-31
Evaluate legacy method of setting NX CI Access Token


tobiashofmann:
Seems that the NX CI access token is included in nx.json: https://github.com/SAP/open-ux-tools/blob/main/nx.json

"accessToken": "ZjFiYzA4OTYtYjFlNS00NTI2LTk4OTktNWZiYTAyMGE0ZGRmfHJlYWQtd3JpdGU=",

This is considered legacy by NX: https://nx.dev/ci/recipes/security/access-tokens

"We do not recommend that you commit an access token to your repository but older versions of Nx do support this"

Why not use secrets and put the token there? Seems that the token used in a read/write token: 
f1bc0896-b1e5-4526-9899-5fba020a4ddf|read-write

The NX documentation cited above states additional recommendation for a read/write token. 

In case the token is still used: why not follow the NX recommendations?
If not used: why not remove it?

devinea:
Thanks. Investigating.

devinea:
@tobiashofmann 
That was a little trickier than expected due to some side effects on build performance and breaking changes across NX versions. 
Finally updated in https://github.com/SAP/open-ux-tools/pull/2583, https://github.com/SAP/open-ux-tools/pull/2519 and   c21227ec9e9c492aa5b6d3a86aa4ebbf76d8df44

Thanks for flagging it.

PR #2583, opened 2024-11-15
Chore - nx upgrade and migration

devinea:
https://github.com/SAP/open-ux-tools/issues/2518
- Upgrade nx and run mx migrate on config. 
- Add macos worker in prep for part 2 to write nx cache.


donal-tobin-sap (approved):
Changes look good.
Follow on changes discussed.

PR #2584, opened 2024-11-18
fix/v4 quick action config prop issue

Jimmy-Joseph19:
This PR fixes

1. First issue is that when a user changes a manifest property in the Property Panel, this change is not reflected in the Quick Action List toggle's state (whether it's on or off). Conversely, if a user tries to toggle a property in the Quick Action List, this change does not update the property settings in the Property Pane
2. Second Issue is after save of manifest changes the app gets reloaded and the selection service is reinitialized where it loses state of the selected control before refresh. When user tries to change the property in properties pane of the selected control the undo/redo operation cease to work


marufrasully (dismissed):
@Jimmy-Joseph19 thanks.
* changes looks good to me
* great test coverage
* did not test it locally

packages/preview-middleware-client/src/cpe/rta-service.ts
```diff
@@ -57,6 +58,9 @@ export class RtaService {
             // eslint-disable-next-line fiori-custom/sap-no-location-reload
             location.reload();
         });
+        this.rta.attachStart(() => {
+            sendAction(appLoaded());
```
voicis, line 62: I suppose now this was the last thing to load, but I'm not convinced that this will always be reliable. We also need other services to have finished being initialised. We probably should just return promise when initialising this service which is fulfilled when `start` event is fired.

Jimmy-Joseph19: adapted

voicis (approved):
- Changes looks good
- Review comments addressed
- Good test coverage
- Tested locally

Issue #2555, opened 2024-11-12
BUG - apps generated with S/4HANA Cloud BAS destinations not supported for use in VSCode


cianmSAP:
### Related Feature
Feature request: https://github.com/SAP/open-ux-tools/issues/1965

### Description
The url in the yaml contains `-api` which breaks the preview. 
Deploy yaml missed the authType property i.e. `authenticationType: reentranceTicket # SAML support for vscode`

### Steps to Reproduce
1. Go to BAS
2. Generate an app using S4 Hana Cloud system 

### Expected results
The url does not contain `-api` in hostname &
deploy yaml contains `authenticationType: reentranceTicket # SAML support for vscode`

### Actual results
URL contains `-api` in hostname
deploy yaml does not contain `authenticationType`

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2586, opened 2024-11-18
feat(btp-utils): adds options for list destination api

cianmSAP:
#2555 

Adds options `ListDestinationOpts` for btp-utils API `listDestinations`
- `stripS4HCApiHosts` removes `-api` from host name


longieirl (approved):
Changes look good.
Test coverage is good.
Applied to the relevant modules.

IainSAP (approved):
Thanks @cianmSAP 

- Tested use in BAS, works as expected
- Updates to consuming modules as appropriate
- Tests to cover

Issue #2590, opened 2024-11-18
BUG - `@sap-ux/odata-service-inquirer`: Intermittently  i18n texts are not displayed (loaded)


IainSAP:
### Description
Sometimes, the labels for shared error messages are not loaded and only the key is shown:

<img width="1082" alt="image" src="https://github.com/user-attachments/assets/cd111fc9-59f3-44ef-9724-22a385b70057">

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

IainSAP:
Reopening for another module

Issue #2498, opened 2024-10-24
TBI - Investigate i18next loading for modules


cianmSAP:
### Description (include screenshots)
There have been issues found in the prompting modules loading the i18next bundles of the sub modules. 
A temp fix has been added to add the resource bundle of the sub module directly in the consuming module, after i18next has been initialised. See https://github.com/SAP/open-ux-tools/pull/2494

While this approach works it potentially requires consumers of the 'sub-module' to know to add the bundle.
In addition, the consuming module may end up having to add multiple bundles after initialising which could become hard to manage. 

### Value
It will improve the usability of the modules, i.e the i18next resource bundle will not have to be explicitly added but should be automatically preloaded when the module is consumed.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

PR #2592, opened 2024-11-19
fix(odata-service-inquirer): i18n bundles not loading

IainSAP:
Fix for : #2590


cianmSAP (approved):
Fix looks good 
Existing TBI to investigate this issue https://github.com/SAP/open-ux-tools/issues/2498

Issue #2567, opened 2024-11-14
TBI - Check integration of UI5 Test Starter into preview middleware


heimwege:
### Description (include screenshots)
The UI5 [Test Starter](https://sapui5.hana.ondemand.com/#/topic/032be2cb2e1d4115af20862673bedcdb) has been officially rolled out today in the [UI5ersLive](https://www.youtube.com/playlist?list=PLHUs_FUbq4dX8lUBM21QKGC_9CwCe6mBC). We should check if Preview Middleware should be adjusted to make use of it.

### Value
Latest best practices from SAP being used

### Architecture Elaboration
Does it requires architecture elaboration? yes

### Notes
@tobiasqueck 

### Tasks
- [ ] adjust [templates](https://github.com/SAP/open-ux-tools/tree/main/packages/preview-middleware/templates/test)

heimwege:
Look like currently there would be no real benefit for the user if we integrate the test starter concept. Hence prio is low.

Issue #2476, opened 2024-10-18
BUG - UI5linter sap ui5 bootstrap parameter changes not reflected in the preview sandbox.html template


heimwege:
### Description
There seem to be some changes to the ui5 bootstrap parameters so the preview [sandbox template](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/templates/flp/sandbox.html) needs to be adjusted.

- `data-sap-ui-compatVersion` -> `data-sap-ui-compat-version`
- `data-sap-ui-resourceroots` -> `data-sap-ui-resource-roots`
- `data-sap-ui-oninit` -> `data-sap-ui-oninit`

sample from other project:
![image](https://github.com/user-attachments/assets/290b5c8c-a193-47bf-9ba1-a026f886b762)

Not sure about the downward compatibility. But outdated notation might not be available in UI5 2.x

### Steps to Reproduce
1. eject html sandbox template to file system
2. run ui5linter
3. see findings

### Expected results
No ui5Linter issues

### Actual results
ui5Linter issues

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [X] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2598, opened 2024-11-20
Use ui5 version specific flp sandbox template

heimwege:
Use specific sandbox templates per UI5 major version instead of trying to dynamically adjust the flp sandbox bootstrap.

Prerequisite for UI5 version dependent integration of UI5 Test Starter (https://github.com/SAP/open-ux-tools/issues/2567)

fixes:
- race condition when dynamically adjusting the `src` property of `sap-ushell-bootstrap` that can lead to the ushell services being not properly initialized
- parts of https://github.com/SAP/open-ux-tools/issues/2476


packages/preview-middleware/src/base/flp.ts
```diff
@@ -273,20 +273,77 @@ export class FlpSandbox {
...
             // inform the user if a html file exists on the filesystem
             const file = await this.project.byPath(this.config.path);
             if (file) {
                 this.logger.info(`HTML file returned at ${this.config.path} is loaded from the file system.`);
                 next();
             } else {
-                const template = readFileSync(join(__dirname, '../../templates/flp/sandbox.html'), 'utf-8');
-                const html = render(template, this.templateConfig);
+                //ui5-patched-router comes with cds-plugin-ui5
+                const ui5Version = await this.getUi5Version(req.headers.host, req['ui5-patched-router']?.baseUrl ?? '');
```
heimwege, line 288: @petermuessig can I make use of the `req['ui5-patched-router'].baseUrl` property introduced by [cs-plugin-ui5](https://www.npmjs.com/package/cds-plugin-ui5)? Or is this something private/subject to change that I should rather not rely on?

packages/preview-middleware/templates/flp/sandbox2.html
```diff
@@ -0,0 +1,89 @@
...
+    <!-- Bootstrap the UI5 core library. 'data-sap-ui-frameOptions="allow"' is a NON-SECURE setting for test environments -->
+    <script id="sap-ui-bootstrap"
+        src="<%- basePath %>/resources/sap-ui-core.js"
+        data-sap-ui-libs="<%- ui5.libs %>"
+        data-sap-ui-async="true"
+        data-sap-ui-preload="async"
+        data-sap-ui-theme="<%- ui5.theme %>"
+        data-sap-ui-compat-version="edge"
+        data-sap-ui-language="en"
+        data-sap-ui-bindingSyntax="complex"
```
tobiasqueck, line 57: What about `bindingSyntax`, `frameOptions`, `componentPreload` and `flexibilityServices`? Will they stay camelCase with 2.0 or also change?

@petermuessig

tobiasqueck:
I asked my AI assistant friend and they think you did a good job:

> The provided PR seems to align with the description provided and the referenced issues. The changes focus on using specific sandbox templates per UI5 major version instead of dynamically adjusting the flp sandbox bootstrap. This addresses the reported issues related to a race condition and the UI5 linter parameters not being reflected in the preview sandbox.html template.
> Upon reviewing the diff, it appears that the changes are appropriate and address the stated issues by updating import statements, removing deprecated code, and refactoring configuration files. Therefore, it seems suitable for merging into the codebase.

packages/preview-middleware/templates/flp/sandbox2.html
heimwege: Valid point. At least `frame-options` and `flexibility-services` are listed in the [documentation](https://sapui5.hana.ondemand.com/sdk/#/topic/91f2d03b6f4d1014b6dd926db0e91070). Not sure about `bindingSyntax` and `componentPreload` though. I'll eject the file and see what the linter has to say about it.

heimwege: Hm. Maybe `data-sap-ui-bindingSyntax` and `data-sap-ui-preload` are no longer valid at all? 🤔 
image

No information on `data-sap-ui-xx-componentPreload` from the linter 🤷🏻

heimwege: There seem to be some more parameters (found in the test snaps):
- `data-sap-ui-xx-designMode`
- `data-sap-ui-xx-suppressDeactivationOfControllerCode`
- `data-sap-ui-xx-viewCache`

nikmace (changes requested):

packages/preview-middleware/src/base/flp.ts
heimwege: Checked with Peter, can be used.

packages/preview-middleware/test/unit/base/flp.test.ts
```diff
@@ -304,6 +308,16 @@ describe('FlpSandbox', () => {
             server = await supertest(app);
         });
 
+        test('test/flp.html UI5 2.x', async () => {
+            global.fetch = jest.fn(() =>
+                Promise.resolve({
+                    json: () => Promise.resolve({ libraries: [{ name: 'sap.ui.core', version: '2.0.0' }] })
+                })
+            ) as jest.Mock;
```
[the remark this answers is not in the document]
heimwege: Oh, I wasn't aware about that one. Nice 👍🏻
Adjusted in https://github.com/SAP/open-ux-tools/pull/2598/commits/38a4b20675514adebd043bc4589941bb4e656fb8 (didn't want to adjust the whole .tsconfig so I did not introduce a specific `path` for the import)

packages/preview-middleware/templates/flp/sandbox2.html
heimwege: @matz3 can you by any chance comment on the UI5 bootstrap parameter names?

matz3: `sap-ui-compat-version="edge"` should be set. It also sets the `bindingSyntax` to `complex` internally, so no need to provide that parameter additionally.

I don't know the status about `sap-ui-xx-componentPreload` / `sap-ui-xx-viewCache`. Better double-check this with the core framework team. That's also the reason why UI5 linter currently doesn't show a finding.

`sap-ui-xx-designMode` and `sap-ui-xx-suppressDeactivationOfControllerCode` are restricted parameters that must only be used by the listed consumers (e.g. see https://github.com/SAP/openui5/blob/master/src/sap.ui.core/src/sap/ui/base/DesignTime.js)

heimwege: `data-sap-ui-preload` seems to be replaced by `data-sap-ui-async="true"` since UI5 quite some time (see [documentation of UI5 1.71.57](https://ui5.sap.com/1.71.57/#/topic/91f2d03b6f4d1014b6dd926db0e91070)

heimwege: `sap-ui-xx-designMode` and `sap-ui-xx-suppressDeactivationOfControllerCode` are allowed for RTA so this should be fine, just need to check the naming if its still `camelCase` or if `sap-ui-xx-` switched to `kebap-case` as well.

@codeworrior can you comment on the naming convention of the `sap-ui-xx-` properties in UI5 2.x?

codeworrior: As HTML5 does not support case-sensitive attribute names (DOM APIs return attribute names all lower case), the new sap/base/config enabled kebab-case everywhere. The old names (all lower case, created from camelCase in HTML source) are still understood, but I would recommend to switch to the kebab-case, if the UI5 version can be assumed >= 1.120.

mmilko01 (approved):
Focused my review on ADP scenario, changes look good
Changeset OK
Good test coverage
Tested manually with SAPUI5 1.71 and >1.71

heimwege:
@mmilko01 I adjusted the changeset and thus dismissed your review 🙈 Could you please re-approve? Thanks

packages/preview-middleware/src/base/flp.ts
```diff
@@ -155,14 +150,39 @@ export class FlpSandbox {
...
     }
 
+    /**
+     * Get the configuration for the developer mode.
+     *
+     * @param ui5MajorVersion - the major version of UI5
+     * @returns the configuration for the developer mode
+     * @private
+     */
+    private getDeveloperModeConfig(ui5MajorVersion: number): Map<string, string> {
```
tobiasqueck, line 160: I think this function would look simpler structured like
```
if (ui5MajorVersion < 2) {
   return new Map([ ... ]);
} else {
  return new Map([ ... ]);
}
```

tobiasqueck (approved):
* changeset ok
* fixes the racing condition issue and in general makes the switch 2.0/1.x choice more robust
* code is clear -- added a suggestion for improvement but it is not critical, I am approving also without it
* successfully tested it locally

packages/preview-middleware/src/base/flp.ts
heimwege: ok, will do. Had it like this before and switch to this approach after consulting AI but I agree that having it explicitly written will increase the readability.

heimwege: done in https://github.com/SAP/open-ux-tools/pull/2598/commits/c8d6e8569dd5da189ca1c518e1c75e3136a97dca

tobiasqueck (approved):
thanks for applying my feedback

mmilko01 (approved):
Re-approving

Issue #2605, opened 2024-11-21
FEATURE - Enhance extracted FLP prompts to support ADP scenario


nikmace:
_As a **developer**,
I want to **be able to configure Fiori Launchpad config in Adaptation Project scenario**,
so that I **can see my ADP application in the Fiori launchpad after the project has been deployed**_

### Description

_What is the high-level purpose and value of the feature?  The description should ensure that anyone reading understands what it is doing (at a high level) and most importantly why._

Currently, FT offers a different FLP configuration than the ADP. I want to have the same experience when using Adaptation Project - the ability to use the same tool to create FLP configurations for adaptation projects.

I want to have the following changes in prompting, when this tool is working for ADP:
- the S/4 flp configuration question in the beginning should NOT be shown
- another question under the visible 4 textboxes should be shown - the existing one for Parameters in our current ADP generation process for S/4 HANA Cloud.

Add new inbound. 
Case 1. manifest.appdescr_variant HAS NO FLP Configuration:
  a. If the base application has inbounds, then show the inbounds list input with all available inbounds plus we show the 4 prompts as Fiori do (Semantic Object, Action, Title, Subtitle).

  b. Base application has no inbounds, then we show the 4 prompts as Fiori do (Semantic Object, Action, Title, Subtitle)  and the parameters question. After that we create a "dummy" inbound which is "<app.id>.InboundID".

In YUI at the end of the 5 questions a new question should be shown with the following details (in YUI):
  - type: Radio Button with Yes and No answers
  - text: "Do you want to add or edit another inbound?" with tooltip text: "If you select Yes and press Finish the current one will be saved and you will have the opportunity to add another one"
  - When I press Finish if I have selected Yes on the radio button, the "Add Inbound" change should be saved and the tool should be relaunched again. If I have selected NO, just the change should be saved.

### Technical Design
_What modules need to be created or enhanced?_ 
_How would the functionality be exposed and how can it be consumed?_

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a ADP
**when** I execute `add navigation-config` command
**then** an FLP configuration is added to my project (manifest.appdescr_variant).

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #2610, opened 2024-11-21
feat: Enhance extracted FLP prompts to support ADP scenario

nikmace:
Feat for #2605.

- Add Adaptation Project specific prompts, which can be hidden with the help of **promptOptions**.
- Enhance the `add inbound-navigation` to work for both Fiori and Adaptation projects.
- Changes the `ManifestService` in `adp-tooling` to accept the provider as a parameter.


packages/axios-extension/src/abap/abap-service-provider.ts
```diff
@@ -94,12 +94,7 @@ export class AbapServiceProvider extends ServiceProvider {
         if (this.atoSettings === undefined) {
             await this.getAtoInfo();
         }
-        return (
-            this.atoSettings.tenantType === TenantType.Customer &&
-            this.atoSettings.operationsType === 'C' &&
-            this.atoSettings.developmentPrefix !== '' &&
-            this.atoSettings.developmentPackage !== ''
-        );
+        return this.atoSettings.tenantType === TenantType.Customer && this.atoSettings.operationsType === 'C';
```
cianmSAP, line 97: Could you explain the reason for this?

My understanding is steampunk systems could satisfy `this.atoSettings.tenantType === TenantType.Customer && this.atoSettings.operationsType === 'C'` but have empty strings for `developmentPrefix` and `developmentPackage`

nikmace: I was testing with a system that made the method validate to false because the `atoSettings` had **developmentPrefix** and **developmentPackage** set to empty strings. Although, the system is cloud. 

I've asked @tobiasqueck about this and he proposed that I remove those two conditions. And that **operationsType** and **tenantType** would be sufficient.

cc @testojs @mmilko01

cianmSAP: When working on something related previously, I was provided the ATO settings for a steampunk system which contained `<operationsType="C" tenantType="CUSTOMER">`

So I'm not conviced this is sufficient to determine it is an S/4HANA cloud system as per the JSDOC.

However, the current implementation assumes an `S/4HANA cloud system` must have a `developmentPrefix` & `developmentPackage`, and as per your investigation, it's not always the case either. 

Maybe this function needs to be renamed to `isCloud` ? Are ATO settings the correct way to go to determine this or can we look at something else. @tobiasqueck

packages/inquirer-common/src/types.ts
```diff
@@ -54,6 +54,23 @@ export interface GuiOptions {
      * Indicate state in the left hand navigation panel in YUI
      */
     breadcrumb?: boolean | string;
+    /**
+     * Indicates how the question will be displayed in YUI
+     */
+    type?: 'label' | 'folder-browser';
```
cianmSAP, line 60: We have a property `guiType` for `folder-browser` on the prompt itself
See https://github.com/SAP/open-ux-tools/blob/2020de9f511a74236ff46f1fd25580f1f5232052/packages/ui5-application-inquirer/src/prompts/prompts.ts#L336

packages/axios-extension/src/abap/abap-service-provider.ts
nikmace: Changes to the `isS4Cloud` method reverted in [ab3ff4a](https://github.com/SAP/open-ux-tools/pull/2610/commits/ab3ff4ab37d8e6f17fec5607d0a2e62e675f343a)

packages/inquirer-common/src/types.ts
cianmSAP: Understood that this is added due to existing inquirers using `folder-browser`

cianmSAP (approved):
Updates to the FLP inquirer and inquirer common look good to me 
Tests covering the changes
Changeset added
✅

heimwege (approved):
- excellent test coverage
- changes exists
- did NOT test locally

packages/adp-tooling/src/base/helper.ts
```diff
@@ -13,6 +17,57 @@ export function getVariant(basePath: string): DescriptorVariant {
...
+}
+
+/**
+ * Determines whether the project at the given base path is an Adaptation Project (ADP).
+ *
+ * @param {string} basePath - The base path to the project.
+ * @returns {Promise<boolean>} A promise that resolves to true if the project is an ADP project, or false otherwise.
+ * @throws {Error} If the project type cannot be determined.
+ */
+export async function isAdpProject(basePath: string): Promise<boolean> {
```
mmilko01, line 38: Can we use https://github.com/SAP/open-ux-tools/blob/c59e479be60306b5e7807ba7f6ca46f45fad8c62/packages/project-access/src/project/info.ts#L139 instead?

nikmace: Changed in: [c295d36](https://github.com/SAP/open-ux-tools/pull/2610/commits/c295d3650878c552b5c692db7d878e7d02e84341)

mmilko01 (dismissed):
- Changeset OK
- Excellent test coverage
- Comments resolved
- Partly tested manually

cianmSAP (approved):
Re-approving

mmilko01 (approved):
Re-approving

cianmSAP (approved):

heimwege (approved):
re-approve

Issue #2616, opened 2024-11-22
BUG - `@sap/ux/odata-service-inquirer`: Partial destinations service path validation errors


IainSAP:
### Description

Users are allowed to continue with invalid service paths. 

![image](https://github.com/user-attachments/assets/fa4c7fc8-d8a5-4b7c-90f1-5ee838de57fa)

![image](https://github.com/user-attachments/assets/acd42839-47b1-4de3-8cd3-92bf1355d13e)

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

IainSAP:
The "fix" incorrectly sets the `validatedUrl` only when auth is not required or authenticated instead of just reachable. This means the basic auth credential prompts will never be shon for full/partial urls.

PR #2617, opened 2024-11-22
fix(`@sap-ux/odata-service-inquirer`): Invalid partial dest service paths are allowed

IainSAP:
Fix for #2616 

- Adds more path validations
- Ensure service validation prompt is not shown until a valid path is input
- Adds more tests


packages/odata-service-inquirer/src/prompts/connectionValidator.ts
```diff
@@ -590,16 +590,23 @@ export class ConnectionValidator {
...
         this._destination = destination;
         // No need to apply sap-client as this happens automatically (from destination config) when going through the BAS proxy
         const status = await this.checkUrl(new URL(destUrl), undefined, undefined, {
             odataVersion: requiredOdataVersion
         });
-        this._validatedUrl = destUrl;
+
         const validationResult = this.getValidationResultFromStatusCode(status);
 
+        if (this.validity.reachable && (!this.validity.authRequired || this.validity.authenticated)) {
```
longieirl, line 606: Does this handle the scearnio where for a SAMLAssertion destination, a HTTP 401 shouldn't prompt the user for their credentials?

IainSAP: Not sure I follow the question. There is no prompting here. That would be a different part of the code . This is to ensure we dont add the url as validated if it is not reachable or needs auth.

longieirl: Thanks @IainSAP I will create a separate issue.

longieirl (approved):
Changes look good.
Issues addressed.
Tests are good.

devinea (approved):
Changes look good. 
Changeset ✅

cianmSAP (approved):
Fix looks good
Tests added
Patch changeset

PR #2620, opened 2024-11-25
fix(cf-deploy-config-writer): cleanup tests

longieirl:
Test is failing
https://cloud.nx.app/runs/CKCeGkvcCe

No changeset required.


packages/cf-deploy-config-writer/test/unit/index-app.test.ts
```diff
@@ -71,56 +71,53 @@ describe('CF Writer App', () => {
         jest.resetAllMocks();
     });
 
-    describe('Generate HTML5 App Config', () => {
```
cianmSAP, line 74: Just curious was this describe block causing the hooks not to be ran as expected?

cianmSAP (approved):
Test updates look good
No changeset needed

packages/cf-deploy-config-writer/test/unit/index-app.test.ts
longieirl: I dont! I just know from past experience around timeouts, that nested describes don't act nicely! 

This might not be the root cause but I cant think of anything else.

devinea (approved):
Thanks for taking a look. @longieirl

Issue #2625, opened 2024-11-26
BUG - odata-service-inquirer: Should use refresh token when connecting to abap systems if available


IainSAP:
### Description

When a `BackendSystem` is selected by the user form the systems prompt, if the system connection configuration  has a refresh token then it should be used instead of generating a new token and reauthenticating.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2626, opened 2024-11-26
fix for refresh token not being used when reconnecting

IainSAP:
Fix for : #2625

- Adds passing of refresh token for backend systems that have it defined
- Adds tests to cover


longieirl (approved):
Changes look good.
Appropiate test changes.

PR #2627, opened 2024-11-26
fix: update quick action title for semantic date range

nahmed22:
update `title` for semantic date range in filter bar.


vadson71 (approved):

voicis (approved):
Simple text change

PR #2632, opened 2024-11-27
fix: remove feature toggle check for semantic date range quick action

nahmed22:
Remove feature toggle check for semantic date range quick action


voicis (approved):

vadson71 (approved):
Changes look reasonable

PR #2633, opened 2024-11-27
fix: adapt ai contributing reference

marufrasully:
Adapting AI contributing reference.


donal-tobin-sap (approved):
Change looks good to me.

PR #2634, opened 2024-11-27
fix(cf-deploy-config-writer): update destination policy and remove excess module

longieirl:

- the `existing_destinations_policy` should default to `update`
- the app content module shouldn't be added when creating a base config as this is only required when a HTML5 is being added
- `addDestinationService` is not required and is not referenced anywhere so is being removed


packages/cf-deploy-config-writer/src/cf-writer/base-config.ts
```diff
@@ -53,7 +53,7 @@ async function addRoutingConfig(config: CFBaseConfig, fs: Editor): Promise<void>
     if (config.routerType === RouterModuleType.Standard) {
         await addStandaloneRouter(config, mtaConfigInstance, fs);
     } else {
-        await mtaConfigInstance.addRoutingModules(true);
+        await mtaConfigInstance.addRoutingModules(true, false);
```
kjose90, line 56: Thanks @longieirl 
PR is good and changes are clear.
A minor comment - would be nice to pass params as options here...
`options: { isManagedApp?: boolean; doCleanUp?: boolean } `

kjose90 (dismissed):
Thanks @longieirl 
Approved ✅

packages/cf-deploy-config-writer/src/mta-config/mta.ts
```diff
@@ -515,14 +516,17 @@ export class MtaConfig {
      * Append and/or cleanup the destination resource if missing in mta.yaml.
      *
      * @param {boolean} isManagedApp - if true, append managed approuter configuration
+     * @param {boolean} doCleanUp - if true, will ensure any missing modules | resources are appended
```
IainSAP, line 519: `doCleanUp` is a little ambiguous ? What about `addMissingModules`?

longieirl: Done

cianmSAP (approved):
Changes are clear
Changeset added
Not tested

kjose90 (approved):
Reapproving - Thanks!

PR #2661, opened 2024-12-04
fix(cf-deploy-config-writer): add runinband

longieirl:
- to allow tests run in sequence, there is no reason for a timeout and the timeout value is already set to 10000
- there is no time difference either in running the tests with this flag


cianmSAP (dismissed):
Update for test script looks good
No changeset needed

kjose90 (dismissed):
Thanks John - no change set required - script looks good 👍

cianmSAP (approved):
Re-approving

kjose90 (approved):
re aproving

PR #2662, opened 2024-12-04
fix: hide quick action if the application has old manifest structure

nahmed22:
hide the manifest change actions from the quick action list, if the manifest of the underlying app has array structure.

### DIFF
--- a/packages/preview-middleware-client/src/adp/quick-actions/fe-v2/lr-enable-semantic-date-range-filter-bar.ts
+++ b/packages/preview-middleware-client/src/adp/quick-actions/fe-v2/lr-enable-semantic-date-range-filter-bar.ts
@@ -4,7 +4,7 @@ import { QuickActionContext, SimpleQuickActionDefinition } from '../../../cpe/qu
 import { pageHasControlId } from '../../../cpe/quick-actions/utils';
 import { getControlById, isA } from '../../../utils/core';
 import { SimpleQuickActionDefinitionBase } from '../simple-quick-action-base';
-import { areManifestChangesSupported, prepareManifestChange } from './utils';
+import { areManifestChangesNotSupported, prepareManifestChange } from './utils';
 import SmartFilterBar from 'sap/ui/comp/smartfilterbar/SmartFilterBar';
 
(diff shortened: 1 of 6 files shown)


packages/preview-middleware-client/src/adp/quick-actions/fe-v2/lr-enable-semantic-date-range-filter-bar.ts
```diff
@@ -25,7 +25,8 @@ export class ToggleSemanticDateRangeFilterBar
 
     async initialize(): Promise<void> {
         const isUI5VersionNotSupported = await areManifestChangesSupported();
-        if (isUI5VersionNotSupported) {
+        const pagesStructureInManifest = this.context.manifest['sap.ui.generic.app'].pages;
+        if (isUI5VersionNotSupported || Array.isArray(pagesStructureInManifest)) {
```
voicis, line 29: This check should be part of `areManifestChangesSupported` function, so we have one function that we can use to check if manifest changes can be used.

voicis (changes requested):

packages/preview-middleware-client/src/adp/quick-actions/fe-v2/lr-enable-semantic-date-range-filter-bar.ts
nahmed22: done

voicis: We shouldn't use `not` in function names. In this particular case we need the inverse value, but it is quite likely that we would also need to check if the manifest changes are supported and we would need to use double negation, which is just looks weird and is more confusing `!areManifestChangesNotSupported()`. `!areManifestChangesSupported()` is a lot easier to read and understand.

voicis (approved):
- Code changes looks good
- Review comments addressed
- Good test coverage
- Did not test locally

vadson71 (approved):
Changes look good
Coverage is good

PR #2663, opened 2024-12-04
fix: Modify existing services

broksy:
After introduction of multiple services for `odata-service-writer` (https://github.com/SAP/open-ux-tools/pull/2499), existing services for previously generated applications should be checked and adapted if necessary.

Previously whenever service was generated, service files were always generated in localService folder. Currently multiple services are supported so each service should have its own folder.

For example:
>       "mainService": {
>         "uri": "/sap/opu/odata4/dmo/sb_travel_mdsk_o4/srvd/dmo/sd_travel_mdsk/0001/",
>         "type": "OData",
>         "settings": {
>           "annotations": [
>             "annotation"
>           ],
>           "localUri": "localService/metadata.xml",
>           "odataVersion": "4.0"
>         }
>       }

`localUri` of the existing OData dataSource should be modified to `localService/mainService/metadata.xml` and respective service files moved to service folder, like:
`project/webapp/localService/metadata.xml` -> `project/webapp/localService/mainService/metadata.xml`
`project/webapp/localService/service-annotation.xml` -> `project/webapp/localService/mainService/service-annotation.xml`.

**Summary of changes:**
Added one step before manifest.json template is generated for service:
`await updateExistingServices(webappPath, manifest, fs);` which goes through existing services, modifies `localUri` and moves related files to service folder.

### DIFF
--- a/.changeset/pretty-ghosts-cough.md
+++ b/.changeset/pretty-ghosts-cough.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/odata-service-writer': patch
+---
+
+Update existing services in a way that is required for multiple services support - service files in their respective folders.
--- a/packages/odata-service-writer/src/index.ts
+++ b/packages/odata-service-writer/src/index.ts
@@ -255,7 +255,7 @@ async function generate(basePath: string, service: OdataService, fs?: Editor): P
     // Prepare template folder for manifest and xml updates
     const templateRoot = join(__dirname, '../templates');
     // Update manifest.json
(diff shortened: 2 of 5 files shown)


devinea (approved):
Code changes are clear. 
Patchset ✅  (I think patch is ok, but could be minor due to function becoming async?)

Klaus-Keller (approved):
Thanks @broksy!
* code changes are easy to understand
* changeset exists
* test coverage is great
* did no local test, just visual review

Approved from my side

PR #2667, opened 2024-12-05
fix: we now properly generate a sap-ui-version.json for ui5 mapping

nlunets:
Previously this file was missing and in some scenario it is required.


tobiasqueck (approved):
* changeset ok
* fix makes sense
* code is clear

PR #2671, opened 2024-12-06
Adds conditional changeset file validation

IainSAP:
- Uses https://github.com/andstor/file-existence-action to test for changeset files existing before running the cset status
- Tested with no cset, empty cset, cset with changes, multiple csets and invalid csets


Klaus-Keller (approved):
Thanks @IainSAP, adding check for changeset before validating looks good.

voicis (approved):

PR #2504, opened 2024-10-28
Feat/2172/convert local preview files

heimwege:
https://github.com/SAP/open-ux-tools/issues/2172

**app-config-writer/convert** + **create**
- introduce command `convert` with parameter `preview-config` that converts an app to virtual preview files

**app-config-writer/variants-config**
- delete RTA parameters (depending on #2431)
- various bug fixes and adjustments for the converter

**project-access**
- add function `getAllUi5YamlFileNames` that reads all yaml files in the given directory
- enhance `readUi5Yaml` to optionally validate the schema of the yaml

**ui5-config**
- enhance the `build` run script to download the UI5 yaml schema to `dist` to also support offline schema validations
- add function for schema validation of a ui5-config instance
- (optionally) validate the schema when creating a `newInstance`

**preview-middleware**
- add `sap-ui-xx-condense-changes` to RTA re-route
- add migration chapter to readme

PR #2675, opened 2024-12-06
feat (create) add option to convert test runners to preview-config command

heimwege:
follow up on https://github.com/SAP/open-ux-tools/pull/2504

- Add option to convert test runners to preview-config command
- some refactoring/fixes

cross link to https://github.com/SAP/open-ux-tools/issues/2172


packages/app-config-writer/src/preview-config/ui5-yaml.ts
```diff
@@ -102,14 +99,32 @@ async function isUi5YamlAlreadyConverted(
...
+ *
+ * @param path - the path
+ * @param configuration - the preview configuration
+ * @returns indicator if the path is a test path
+ */
+export function isTestPath(path: string | undefined, configuration?: PreviewConfig): boolean {
+    if (!path) {
+        return false;
+    }
+    //todo: how to ensure we don't mistake other test scripts for preview scripts?
```
heimwege, line 123: @tobiasqueck: do you have a better idea to identify test (runner) files than eiter filename contains `.qunit.html` or `path` is being used as `test.path`  in yaml config of preview-middleware?

packages/app-config-writer/src/preview-config/package-json.ts
```diff
@@ -49,8 +49,9 @@
     intent: FlpConfig['intent'] | undefined;
 } {
     //extract the URL from the 'open' command of the script
-    let url = / (?:--open|-o|--o) ([^"]?\S*)/.exec(script)?.[1] ?? undefined;
-    url = url?.startsWith('"') ? url.slice(1) : url;
+    let url = / (?:--open|-o|--o) (\S*)/.exec(script)?.[1] ?? undefined;
+    //delete double or single quotes from the URL
+    url = url?.replace(/['"]/, '');
```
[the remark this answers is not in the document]
heimwege: This is not really needed as we only need the first part of the URL to add the path to the yaml but It also doesn't harm (hopefully) so lets do it.

.changeset/dirty-jeans-fly.md
Klaus-Keller, line 1: @heimwege, there are also changes for `@sap-ux/reload-middleware` but it is not in the change set. Is this intentional?

packages/app-config-writer/src/preview-config/index.ts
```diff
@@ -14,11 +14,17 @@ import { type ToolsLogger } from '@sap-ux/logger';
...
  * @param basePath - base path to be used for the conversion
+ * @param convertTests - indicator if test suite and test runner should be included in the conversion (default: false)
  * @param logger logger to report info to the user
  * @param fs - file system reference
  * @returns file system reference
  */
-export async function convertToVirtualPreview(basePath: string, logger?: ToolsLogger, fs?: Editor): Promise<Editor> {
+export async function convertToVirtualPreview(
+    basePath: string,
+    convertTests: boolean = false,
```
Klaus-Keller, line 24: This is an incompatible change, luckily the module has still a `0.x.y` version. However, considering the growing number of arguments and to avoid incompatible changes in future, do you want to convert the additional parameters into an options object, like:

```typescript
export async function convertToVirtualPreview(
    basePath: string,
    options: { convertTests: boolean; logger?: ToolsLogger; fs?: Editor }
): Promise<Editor>
```

packages/app-config-writer/src/preview-config/prerequisites.ts
```diff
@@ -4,6 +4,29 @@ import type { Editor } from 'mem-fs-editor';
...
+ */
+function isLowerThanMinimalVersion(
+    packageJson: Package,
+    dependencyName: string,
+    minVersionInfo: string,
+    mandatory: boolean = true
+): boolean {
+    const versionInfo = packageJson?.devDependencies?.[dependencyName] ?? packageJson?.dependencies?.[dependencyName];
+    if (!versionInfo) {
+        return mandatory;
```
Klaus-Keller, line 26: Can you explain this from a consumer perspective? If I `mandatory` is `true` and the module is not found in dependencies of devDependencies, function `isLowerThanMinimalVersion()` returns true?

packages/app-config-writer/src/preview-config/preview-files.ts
```diff
@@ -2,12 +2,13 @@ import { basename, join } from 'path';
...
 import type { Editor } from 'mem-fs-editor';
 import type { ToolsLogger } from '@sap-ux/logger';
+import { TEST_CONFIG_DEFAULTS } from './ui5-yaml';
 
 const renameMessage = (filename: string): string =>
     `Renamed '${filename}' to '${filename.slice(
         0,
         -5
-    )}_old.html'. This file is no longer needed for the preview functionality. If you have not modified this file, you can delete it. If you have modified this file, move the modified content to a custom init script for the preview middleware. For more information, see https://www.npmjs.com/package/preview-middleware#migration.`;
+    )}_old.html'. This file is no longer needed for the preview functionality. If you have not modified this file, you can delete it. If you have modified this file, move the modified content to a custom init script for the preview middleware. For more information, see https://github.com/SAP/open-ux-tools/tree/main/packages/preview-middleware#migration.`;
```
Klaus-Keller, line 11: Just out of curiosity, why the change from npmjs.com to GitHub?

.changeset/dirty-jeans-fly.md
heimwege: I assume you are you referring to preview-middleware, right? That's a valid point. I'll add it

Klaus-Keller: yes, typo

packages/app-config-writer/src/preview-config/prerequisites.ts
heimwege: The question was what to do in case a (dev)dependency is mandatory but has not been found. I thought it might be best to behave as if the version was too low. This means the user will get [an error message](https://github.com/SAP/open-ux-tools/blob/3888918983c348ee8ab8aa5bc583d3e28b1bf8e7/packages/app-config-writer/src/preview-config/prerequisites.ts#L63) with the name of the missing package and the lowest version needed.

e.g. `UI5 CLI version 3.0.0 or higher is required to convert the preview to virtual files. For more information, see https://sap.github.io/ui5-tooling/v3/updates/migrate-v3.`

An alternative would be to throw an exception. But as we would most likely use exactly the same error message I didn't see a real benefit. Would you prefer the exception?

packages/app-config-writer/src/preview-config/preview-files.ts
heimwege: The initial link was wrong and need to be adjusted. But then I noticed the deep link didn't work in npmjs readme whereas it worked in github. I always ended up at the very top of the readme in npmjs instead of the migration section. But as I just re-tested it now it seems to work again 🤷🏻
So if you prefer npmjs I can adjust the link as both seem to work now.

.changeset/dirty-jeans-fly.md
heimwege: added in https://github.com/SAP/open-ux-tools/pull/2675/commits/83f3e82a66f6e75fc12eb47b94062c4b4d4ebdd2

packages/app-config-writer/src/preview-config/index.ts
heimwege: Oh you're right 🙈 I adjusted the code according to your suggestion in https://github.com/SAP/open-ux-tools/pull/2675/commits/83f3e82a66f6e75fc12eb47b94062c4b4d4ebdd2

Klaus-Keller (approved):
Thanks @heimwege,
- code to convert test runner looks good
- changeset exists
- test coverage is good

Approved from my side

ytpo-lyne (approved):
- changeset present
- code changes look good
- nice code coverage

Approving

Issue #2679, opened 2024-12-09
BUG - `axios-extension` : Paging does not work for v4 service list API


IainSAP:
### Description

The parsing of the next link url is incorrect.  The property `response.data` is not an object but a string but is referenced to extract the next link as if it was expected to be an object:

https://github.com/SAP/open-ux-tools/blob/main/packages/axios-extension/src/abap/catalog/v4-catalog-service.ts#L88

This means only the first page of service (groups) is ever returned.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2680, opened 2024-12-09
fix(axios-extension): listing v4 odata services only returns first page

IainSAP:
Fixes :https://github.com/SAP/open-ux-tools/issues/2679

- Odata control info was not being returned by the `parseODataResponse` function and so `@odata.nextlink` was not available for paging request construction
- Adds optional inclusion of odata control info (`@odata` annotations) to get `@odata.nextlink` control info
- Fixes tests which were returning false positives
- Adds support for `URLSearchParams` when passing axios config to `get` url since axios supports params as `object` or `URLSearchParams`

### DIFF
--- a/.changeset/giant-garlics-tan.md
+++ b/.changeset/giant-garlics-tan.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/axios-extension': patch
+---
+
+Fix v4 odata services not paged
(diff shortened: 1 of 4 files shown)


longieirl (approved):
Change only affects v4 requests soo its isolated.
Tests look good.
Changeset is appropriate.

cianmSAP (approved):
Updates look good
Have no tested
Tests to cover
✅

PR #2684, opened 2024-12-10
TBI - Prevent CRLF injection in logs

devinea:
Use `encodeURI` to santize strings before logging to avoid CRLF injection


heimwege (approved):
- changes are straightforward
- changeset ok
- did NOT test manually

PR #2685, opened 2024-12-10
fix: running test multiple times

nlunets:
Currently in certain case the pathMapping function is lost when running the test multiple time.
That's bad and we need to fix it


voicis (approved):

Klaus-Keller (approved):
Thanks @nlunets,
* fix to cache UI5 version looks good
* changeset exists
* test coverage is great

Approved from my side

PR #2686, opened 2024-12-10
fix: Manifest template for services

broksy:
1. Improved manifest.json template.
JSON parsing error was thrown for case with `annotations: []` and defined `localAnnotationsName`.

2. Updated `localAnnotationsName` type to also allow to use local annotations defined in array.

### DIFF
--- a/packages/odata-service-writer/templates/extend/manifest.json
+++ b/packages/odata-service-writer/templates/extend/manifest.json
@@ -10,14 +10,20 @@
                         <% if (annotations.length) { %>
                             <% annotations.forEach(function(annotation, index) { %>
                                 <% if (annotation.technicalName) { %>
-                                    "<%= annotation.name %>"<% if (index < annotations.length - 1 || locals.localAnnotationsName) { %>,<% } %>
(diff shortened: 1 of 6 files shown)


packages/odata-service-writer/templates/extend/manifest.json
```diff
@@ -10,13 +10,12 @@
                         <% if (annotations.length) { %>
                             <% annotations.forEach(function(annotation, index) { %>
                                 <% if (annotation.technicalName) { %>
-                                    "<%= annotation.name %>"<% if (index < annotations.length - 1) { %>,<% } %>
+                                    "<%= annotation.name %>"<% if (index < annotations.length - 1 || locals.localAnnotationsName) { %>,<% } %>
```
devinea, line 13: `localAnnotationsName` was fine for app generation initially. 
But for update cases it could be limiting as it is possible to support multiple local annotation files in the manifest. In fact the Annotation File manage provides the ability to create them.
Could you address that?

broksy: `localAnnotationsName` type should be changed then from `string` to `string | string[]` for backward compatibility.

devinea (approved):
Code changes are clear. Thanks for the update. 
I haven't tested locally. 
Changeset ✅

815are (approved):
checked:
1. code easy to understand - additional support for array of names is added
2. new tests are added 
3. changeset presented

PR #2690, opened 2024-12-11
Cleanup cf tests

longieirl:
- ensure there is no full mocking of modules, only isolated methods should be mocked


IainSAP (approved):
Thanks @longieirl 

- More targeted mocking reduces side-effects

devinea (approved):
Test updates look good.
Do we need these changeset files?

PR #2691, opened 2024-12-11
Feat(preview-middleware-client): adjust loading UI5 version to be more precise

heimwege:
Fetching the `sap.ui.core` lib version does not exactly represent the SAPUI5 version (e.g. see https://ui5.sap.com/1.96.38/patchinfo.html). The patch version may differ. This PR adds support for requesting custom lib versions and uses `sap.ui.core` as fallback.


nlunets (approved):
The timeout increase on the jest part looks ok

packages/preview-middleware-client/package.json
```diff
@@ -26,9 +26,12 @@
     "engines": {
         "node": ">=18.x"
     },
-    "dependencies": {},
+    "dependencies": {
+        "semver": "7.6.3"
```
voicis, line 30: What particular issue did you want to solve with this dependency? So far we don't have any external dependencies and I don't think we should change that without a good reason. What I can see that it replace two if statements, does not look like good enough reason for adding this, but maybe there is something else I'm not aware of.

packages/preview-middleware-client/src/utils/version.ts
```diff
@@ -23,12 +24,36 @@ const minVersionInfo = {
...
+ * @param lib - (optional) specific library name to get the version from, e.g. 'sap.m'
  * @returns Ui5VersionInfo
  */
-export async function getUi5Version(): Promise<Ui5VersionInfo> {
-    let version = ((await VersionInfo.load({ library: 'sap.ui.core' })) as SingleVersionInfo)?.version;
+export async function getUi5Version(lib?: string): Promise<Ui5VersionInfo> {
+    let version: string | undefined;
+    //fallback 1: sap.ui.server.abap because its always released together with SAPUI5 and thus has the same version
+    //fallback 2: sap.ui.core because its always present in a SAPUI5 application
+    const libraries = [lib, 'sap.ui.server.abap', 'sap.ui.core'].filter(Boolean);
```
voicis, line 50: I'm not sure if we need to complicate this logic that much, especially when it still can't guarantee that it will always be correct and that this is needed for only a very small subset of the cases (it should be only relevant for patch versions). 

I think we just have to accept the fact that there could be different library patch versions and do the appropriate checks - look which library version is needed for the particular case and check that. Using `sap.ui.core` as default library for version check still makes sense because majority of the cases thats exactly what we're interested.

packages/preview-middleware-client/src/utils/version.ts
```diff
@@ -46,42 +71,40 @@ export async function getUi5Version(): Promise<Ui5VersionInfo> {
...
-    return false;
+    checkVersionInfo(ui5VersionInfo);
+    checkVersionInfo(minUi5VersionInfo);
+    const ui5VersionString = `${ui5VersionInfo.major}.${ui5VersionInfo.minor}.${ui5VersionInfo.patch ?? 0}`;
+    const minUi5VersionString = `${minUi5VersionInfo.major}.${minUi5VersionInfo.minor}.${minUi5VersionInfo.patch ?? 0}`;
+    return lt(ui5VersionString, minUi5VersionString);
 }
 
 /**
  * Checks if the given version is equal to the specified version.
```
voicis, line 90: We should mention it here that patch versions for a specific library might not match the framework/core version and what steps should be taken to ensure that version check produces the desired result.

packages/preview-middleware-client/package.json
heimwege: Fair point. Current code version from main completely ignores the patch version so I thought using semver lt is easier than to compare all those version parts in various if statements.
Nevertheless semver might be a bit too much as we don't expect to get any ranges from the UI5 version fetch. I just recently discovered it and now maybe want to use it too often 😄 Don't know. Maybe you are right and it's not worth introducing the first dependency here.

packages/preview-middleware-client/src/utils/version.ts
heimwege: ok agree. Will adjust to lib parameter or sap.ui.core

voicis (approved):
- Code changes looks good
- Review comments addressed
- Good test coverage
- Did not test locally

Klaus-Keller (approved):
Thanks @heimwege,
* additional checks for UI5 version look good
* change set exists
* test coverage is great

Approved from my side.

Issue #2681, opened 2024-12-10
FEATURE - Ensuring UI5 2.x compliance for cards-editor-middleware Plugin


UBSusmitha:
### Description

https://github.tools.sap/ux/sap.cards.ap/issues/861

### Technical Design
https://github.tools.sap/ux/sap.cards.ap/issues/861

### Acceptance Criteria
https://github.tools.sap/ux/sap.cards.ap/issues/861

UBSusmitha:
Testing steps:
Go to open-ux-tools -> packages -> cards-editor-middleware
pnpm build
npm link
 
Go to any app in sap.cards.ap:
npm link @sap-ux/cards-editor-middleware
pnpm start-cards-generator

Right click when the app is launched after opening generator preview >> View Page Source >> Check for the below line:
```
<script src="[./Init.js](http://localhost:8081/test/Init.js)"></script>
```

If this is there then node module is linked correctly.
Add some group properties >> save >> check if manifest and i18n files are being saved >> reopen the dialog to see if nothing has changed


https://github.com/user-attachments/assets/9ecf6678-7d6b-4264-8e21-99b24abb39ec

UBSusmitha:
Commit merged:
https://github.com/SAP/open-ux-tools/pull/2613

PR #2613, opened 2024-11-22
feat: Ensuring UI5 2.x compliance for cards-editor-middleware Plugin

UBSusmitha:
https://github.com/SAP/open-ux-tools/issues/2681

PR #2695, opened 2024-12-12
fix: move Init.js functionality inline in flpSandbox.html

UBSusmitha:
https://github.com/SAP/open-ux-tools/pull/2613

Commit merged for the task https://github.com/SAP/open-ux-tools/issues/2681 had a wrong path for oninit in flpSandbox.html
```
data-sap-ui-oninit="module:Init"
```

It was working in local apps since the ApiRoutes has previewGeneratorSandbox = '/test/flpGeneratorSandbox.html' as the end point. 

```
export const enum ApiRoutes {
    previewGeneratorSandbox = '/test/flpGeneratorSandbox.html',
}
```
In local apps have webapp/test/Init.js and the merged commit above had 
```
<script src="./Init.js"></script>
```
This PR has the below changes, which were not a part of https://github.com/SAP/open-ux-tools/pull/2613:
=> https://github.com/SAP/open-ux-tools/blob/b5b280b78e67d0541394e84b4fd43698d6cecf5c/packages/cards-editor-middleware/templates/flpSandbox.html#L106  loading core since sap-data-oninit callback is not used.
=> https://github.com/SAP/open-ux-tools/blob/b5b280b78e67d0541394e84b4fd43698d6cecf5c/packages/cards-editor-middleware/templates/flpSandbox.html#L85


chaudhary1-rohit:
Can you please add some description for this change?

chaudhary1-rohit (approved):
- Change looks okay
- local testing done

heimwege (approved):
Having all in one giant html file is of course not the best solution but as this is only interim before its being integrated in the preview middleware it's fine for me.

- changeset ok
- did NOT test locally

PR #2697, opened 2024-12-12
fix(ui-components): Added logic for aria-invalid to UIComboBox depending on errorMessage

Thomas-Neuhauser:


packages/ui-components/src/components/UIComboBox/UIComboBox.tsx
```diff
@@ -165,6 +180,18 @@ export class UIComboBox extends React.Component<UIComboBoxProps, UIComboBoxState
...
 
+    /**
+     * This is a workaround for FluentUI not handling aria-invalid correctly for a ComboBox.
+     * For the time being we add logic here to set aria-invalid based on the presence of
+     * an errorMessage.
+     */
+    private updateAriaInvalid(): void {
+        const htmlInputField = this.comboboxDomRef.current?.getElementsByClassName('ms-ComboBox-Input').item(0);
+        if (htmlInputField) {
+            htmlInputField.ariaInvalid = this.props.errorMessage !== undefined ? 'true' : 'false';
```
815are, line 191: instead of manipulating directly with DOM, can we use props like here? -> https://github.com/SAP/open-ux-tools/blob/9d0e41b1a450d10b512904e4acc98451193f75e1/packages/ui-components/src/components/UIComboBox/UIComboBox.tsx#L705-L723

815are: like?:
```
private getAutofillProps(): IAutofillProps {
        const { highlight, readOnly, disabled, e } = this.props;
        const autofill: IAutofillProps = {};
        // Handle search highligh
        if (highlight) {
            autofill.onKeyDownCapture = this.onKeyDown;
        }
        const tabIndex = 'tabIndex' in this.props ? this.props.tabIndex : undefined;
        // Handle readOnly property
        if (readOnly && !disabled) {
            autofill.readOnly = readOnly;
            autofill.tabIndex = tabIndex;
            // Adjust aria attributes for readonly
            autofill['aria-disabled'] = undefined;
            autofill['aria-readonly'] = true;
        } else if (disabled) {
            autofill.disabled = undefined;
            autofill.readOnly = true;
            autofill.tabIndex = tabIndex;
        }
        if (this.props.errorMessage !== undefined) {
           autofill['aria-invalid'] = true;
        }
        return autofill;
    }
```

Thomas-Neuhauser: @815are perfect, thanks, exactly what I needed!

packages/ui-components/src/components/UIComboBox/UIComboBox.tsx
```diff
@@ -693,6 +693,10 @@ export class UIComboBox extends React.Component<UIComboBoxProps, UIComboBoxState
             autofill.readOnly = true;
             autofill.tabIndex = tabIndex;
         }
+        // This is a workaround for FluentUI not handling aria-invalid correctly for a ComboBox.
+        // For the time being we add logic here to set aria-invalid based on the presence of
+        // an errorMessage.
+        autofill['aria-invalid'] = this.props.errorMessage !== undefined;
```
815are, line 699: @Thomas-Neuhauser 

I just checked if we pass empty string, then visually we do not see error message, but this code will add `aria-invalid` as `true`.
image

Also checked `TextField` - if empty `errorMessage` is passed, then `aria-invalid`is `false`
image


Does it makes sense in this case to add `aria-invalid` if empty error message is passed? I think if empty `errorMessage` is passed, then we should set `aria-invalid` as `false` like it is for UITextInput

815are (changes requested):

packages/ui-components/src/components/UIComboBox/UIComboBox.tsx
Thomas-Neuhauser: Hello Andis, ok, the empty string case is a good one, I will extend the logic to cover that.
Setting aria-invalid = false is correct from my point of view, since this replicates the behavior other FluentUI controls already show.
You can check https://developer.microsoft.com/en-us/fluentui#/controls/web/dropdown and scroll down a little, there's an example where you can switch an error message on and off and if you check the render output, there's always an aria-invalid attribute either true or false.
I assume that screenreaders react to this change and I wanted to replicate this behavior.
I'll add a check for an empty string.

815are: @Thomas-Neuhauser 

my 'complain' was only about empty `errorMessage`, otherwise looks ok

packages/ui-components/src/components/UIComboBox/UIComboBox.tsx
```diff
@@ -693,6 +693,10 @@ export class UIComboBox extends React.Component<UIComboBoxProps, UIComboBoxState
             autofill.readOnly = true;
             autofill.tabIndex = tabIndex;
         }
+        // This is a workaround for FluentUI not handling aria-invalid correctly for a ComboBox.
+        // For the time being we add logic here to set aria-invalid based on the presence of
+        // an errorMessage.
+        autofill['aria-invalid'] = this.props.errorMessage !== undefined && this.props.errorMessage !== '';
```
815are, line 699: probably can be simplified as `!!this.props.errorMessage`, but seems minor

815are (approved):
checked:
1. code changes is clear to understand - `aria-invalid` attribute is set `true` if `errorMessage` is passed through props.
2. changeset exists
3. addition tests are added;

small proposal to simply check, but I think minor

AlinaGovoruhina (approved):
Aria-invalid set if error message present, unit tests and story added, changeset exists.

Issue #2699, opened 2024-12-12
TBI - Expose Cloud Foundry instance types


longieirl:
### Description (include screenshots)

Expose Cloud Foundry instance types. These instance types are used when leveraging cf-tools https://www.npmjs.com/package/@sap/cf-tools for example to filter CF instances by type;

https://github.com/SAP/cloud-foundry-tools-api/blob/edc848eaeb45ac4405b02e8f4f7a25e5047e43f0/src/cfServicesUtil.ts#L13

Sample code;
```
const filteredInstances = [
        ServiceType.ABAP,
        ServiceType.ABAP_TRIAL,
        ServiceType.ABAP_CANARY,
        ServiceType.ABAP_OEM,
        ServiceType.ABAP_OEM_CANARY,
        ServiceType.ABAP_HAAS,
      ];
      serviceInstances = await cfTools.apiGetServicesInstancesFilteredByType(
        filteredInstances
      );
```

Sample list;
```
export enum ServiceType {
    /**
     * ABAP instance type
     */
    ABAP = 'abap',
    /**
     * ABAP trial instance type
     */
    ABAP_TRIAL = 'abap-trial',
    /**
     * ABAP CANARY trial instance type
     */
    ABAP_CANARY = 'abap-canary',
    /**
     * ABAP OEM instance type
     */
    ABAP_OEM = 'abap-oem',
    /**
     * ABAP OEM CANARY instance type
     */
    ABAP_OEM_CANARY = 'abap-oem-canary',
    /**
     * ABAP HAAS instance type
     */
    ABAP_HAAS = 'abap-haas',
    /**
     * Destination service instance type
     */
    DESTINATION = 'destination',
    /**
     * ABAP Staging service instance type
     */
    ABAP_STAGING = 'abap-staging',
    /**
     * ABAP Internal Staging service instance type
     */
    ABAP_INTERNAL_STAGING = 'abap-internal-staging'
}
```

### Tasks
- [ ] Expose new CF instance list via `btp-utils` as this will be consumed by other modules

longieirl:
Consumed as part of latest odata-inqquirer changes.

PR #2700, opened 2024-12-12
fix(btp-utils): add cf instance types

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/2699


longieirl:
@IainSAP wondering if `enum` should be changed to `const, to allow for better performance and more concise code.

IainSAP (approved):
- Moving code, nothing to test
- Cset as expected

cianmSAP (approved):
Change are clear and simple
Changeset added

PR #2703, opened 2024-12-13
feat: Expose credentials prompts for reuse in ADP

mmilko01:


packages/odata-service-inquirer/src/prompts/datasources/sap-system/credentials/questions.ts
```diff
@@ -76,7 +78,8 @@ export function getCredentialsPrompts<T extends Answers>(
                     return false;
                 }
                 // We may have a previously selected system
-                const selectedSystem = answers?.[promptNames.systemSelection] as SystemSelectionAnswerType;
+                const selectedSystem =
+                    (answers?.[promptNames.systemSelection] as SystemSelectionAnswerType) ?? systemSelection;
```
IainSAP, line 82: This should never be possible. The credentials prompts are only for use with the system prompt.

PR #2707, opened 2024-12-16
fix: make module loading in project access more robust

Klaus-Keller:
With the current implementation of module caching in `@sap-ux/project-access`, there is a possibility we end up with an empty (`{}`) `package.json` in `~/.fioritools/module-cache/**/package.json`. This can happen, if for some reason the npm install process is not successful (network issue, registry configuration, etc.). If we end up with an empty package.json, the next time the caching algorithm does not retry to load the module, as the `package.json` already exists.
The empty `package.json` was introduced with https://github.com/SAP/open-ux-tools/pull/2171 to overcome [BUG - module-loader does not install specification dependency in .fioritools folder when node_modules exists on C:\Users\{User} level #2170](https://github.com/SAP/open-ux-tools/issues/2170)

With the change on this pull request, the install process is  changed to `npm install --prefix ...`, which does not require an empty `package.json` to be present and still solve the issue with having modules installed in `node_module` of a parent folder.


heimwege (approved):
Nice improvement 👍🏼

- change is straightforward
- changeset ok
- excellent test coverage
- did NOT test locally

815are (approved):
checked:
1. code change is easy to understand
2. existing test adjusted to new logic

did not test manually

devinea (approved):
Code changes look good. 
Changeset 👍

Issue #2708, opened 2024-12-16
BUG - the CF ABAP instances list is not returning the correct service and label


longieirl:
### Related Feature
Feature request: _issue number_

### Description
The following cf-inquirer`answers` object is missing the `serviceName` when the CF ABAP choices are returned;

https://github.com/SAP/open-ux-tools/blob/eedd6d1f85f7fb7521b9fa625559b7f952838102/packages/cf-deploy-config-inquirer/src/prompts/app-router-prompts.ts#L159

Response;
```JSON
 {
  "inputDestinationRoot": "/Users/I313149/Documents/tools-suite/packages/deployment-generator",
  "prefix": "testme2",
  "mtaDescription": "",
  "mtaVersion": "0.0.1",
  "routerType": "standard",
  "addConnectivityService": true,
  "useABAPServiceBinding": true,
  "selectedABAPService": {
    "label": "abap-05"
  }
}
```

The `selectedABAPService` is not returning the `service` as defined by the type;

```
    abapServiceProvider?: {
        label?: string;
        service?: String;
    };
```

### Steps to Reproduce
1. Load CF Approuter Configuration
2. Select Standalone as router type 
3. Configured advanced options  Yes
4. Select ABAP Env for direct service binding

### Expected results

Select the service and generate standalone approuter project

### Actual results

cf-writing phase is failing since the serviceName is not being passed.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Validated in tools-suite.

PR #2709, opened 2024-12-16
708/return cf abap service name

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/2708

- add async and return correct label and service for the selected CF ABAP instance
- only return the expected type of `label:string` and `service:string` rather than the entire service object which `getCFAbapInstanceChoices` returns by default.
- add extra logging in cf-writer to make it easier to debug
- cleanup logging which is not formatting the mtaPath correctly


devinea (approved):
Code changes are clear
changeset ✅

IainSAP (approved):
Thanks @longieirl 

- Tests added to cover
- Not tested locally

PR #2718, opened 2024-12-18
fix: added tooltip for funnel and copy icon

emrecevik06:
Added tooltip for funnel and copy icon.
image
image
image


vadson71 (approved):
Changes are simple and straight forward
Didn't test locally

voicis (approved):
- straight forward changes
- changeset present
- tested locally

PR #2722, opened 2024-12-18
fix(preview-middleware-client): dialog and manifest change quick actions

voicis:
- fixed an issue where trying to open multiple dialogs would prevent from opening any other dialogs in the same session.
- fixed an issue where unsaved manifest changes were not correctly reflected in quick actions for FE v2 applications.
- removed unused `ui5Version` field from controllers.
- extracted dialog handling logic into a separate `DialogFactory` class.
- added validator pattern to check if quick action should be enabled. (e.g dialog quick actions should be only enabled if there is not already a dialog open)
- added table filtering quick action tests back that were incorrectly removed in #2632


Jimmy-Joseph19 (approved):
CPE package changes looks good.
Test coverage good.
Changeset found.
Have not tested locally

packages/preview-middleware-client/src/adp/quick-actions/fe-v2/lr-enable-semantic-date-range-filter-bar.ts
```diff
@@ -33,8 +33,19 @@ export class ToggleSemanticDateRangeFilterBar
             const isActionApplicable = pageHasControlId(this.context.view, control.controlId);
             const modifiedControl = getControlById<FilterBar>(control.controlId);
             if (isActionApplicable && modifiedControl) {
-                this.isUseDateRangeTypeEnabled = modifiedControl.getProperty('useDateRangeType');
                 this.control = modifiedControl;
+
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                const id: string = this.control.getProperty('persistencyKey') ?? this.control.getId();
```
heimwege, line 39: If you typecast this via `as string` you can get rid of the eslint disablement comment

packages/preview-middleware-client/src/adp/quick-actions/fe-v2/lr-enable-semantic-date-range-filter-bar.ts
```diff
@@ -33,8 +33,19 @@ export class ToggleSemanticDateRangeFilterBar
...
+                if (typeof id !== 'string') {
```
heimwege, line 40: But if you still check the type of `id` afterwards, typecasting it to `string` seems to be wrong 🤔 Shouldn't the type rather be `unknown` then? Would also solve the eslint `any` issue (I guess)

packages/preview-middleware-client/src/adp/quick-actions/fe-v2/lr-enable-table-filtering.ts
```diff
@@ -21,15 +24,40 @@ export class EnableTableFilteringQuickAction
     implements SimpleQuickActionDefinition
 {
     constructor(context: QuickActionContext) {
-        super(ENABLE_TABLE_FILTERING, CONTROL_TYPES, 'QUICK_ACTION_ENABLE_TABLE_FILTERING', context);
+        super(ENABLE_TABLE_FILTERING, CONTROL_TYPES, 'QUICK_ACTION_ENABLE_TABLE_FILTERING', context, [
+            {
+                run: () => {
+                    if (this.control) {
+                        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                        const id: string = this.control.getProperty('persistencyKey') ?? this.control.getId();
```
heimwege, line 32: Same comment w.r.t `string` vs. `unknown`

packages/preview-middleware-client/src/adp/dialog-factory.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+type Controller = AddFragment | AddTableColumnFragments | ControllerExtension | ExtensionPoint;
+
+export const OPEN_DIALOG_STATUS_CHANGED = 'OPEN_DIALOG_STATUS_CHANGED';
+
+export class DialogFactory {
+    private static readonly eventTarget = new EventTarget();
+    private static isDialogOpen = false;
+    /**
```
heimwege, line 29: Shouldn't the jsdoc be above the property?

packages/preview-middleware-client/src/adp/quick-actions/enablement-validator.ts
```diff
@@ -0,0 +1,18 @@
+export interface EnablementValidatorSuccess {
```
heimwege, line 1: As this file only contains types, shouldn't the filename be `...-types` or the location of the file a `types` folder?

heimwege (commented):
Some nit-pick comments

packages/preview-middleware-client/src/adp/dialog-factory.ts
voicis: It is for the public property `canOpenDialog`, so it is correct.

packages/preview-middleware-client/src/adp/quick-actions/enablement-validator.ts
voicis: There were also some functions here earlier, but they were later not needed and now only the types were left. I suppose we could more or rename, but I'm not sure how much value it brings. In general we do not split the types from the implementation here.

packages/preview-middleware-client/src/adp/quick-actions/fe-v2/lr-enable-semantic-date-range-filter-bar.ts
voicis: Yes, just type casting to `string` is not really safe. There have been quite a few cases that the assumption turns out not to be correct, so I prefer to do the explicit checks. I'm not too happy about the comment to ignore the ESLint rule, I'll see if we can somehow get rid of it.

voicis: Fixed by https://github.com/SAP/open-ux-tools/pull/2722/commits/c6df94041ce1edd6d03041b431a3ede48f77d67d

packages/preview-middleware-client/src/adp/quick-actions/fe-v2/lr-enable-table-filtering.ts
voicis: Fixed by https://github.com/SAP/open-ux-tools/pull/2722/commits/c6df94041ce1edd6d03041b431a3ede48f77d67d

heimwege (dismissed):
- changeset exists
- good test coverage
- did NOT test locally

IvoSG (approved):
- The refactoring (e.g., DialogFactory and Validator pattern) is clean and improves maintainability.
- Bug fixes address the reported issues effectively, especially around dialog handling and unsaved manifest changes.
- Test coverage is comprehensive, and the reintroduction of table-filtering quick action tests ensures regression prevention.
- Did not tested locally.

vadson71 (dismissed):
Changes look good.
Good test coverage.
Didn't test locally

Jimmy-Joseph19 (dismissed):
Reapproving CPE Changes after addressing comments

IvoSG (approved):
Reapproving after addressing comments

Jimmy-Joseph19 (approved):
Reapproving after merge with master

Issue #2723, opened 2024-12-18
BUG - `odata-service-inquirer`: Previous connection state is retained when selecting `cfAbapEnvService` option


IainSAP:
### Description

Toggling between Cloud Foundry ABAP environment on SAP Business Technology Platform and existing SAP BTP systems causes duplication of service list fields.

<img width="1722" alt="Image" src="https://github.com/user-attachments/assets/34900078-b109-4467-94ea-405ad105c3d0" />

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2729, opened 2024-12-19
fix(`@sap-ux/odata-service-inquirer`): Reset connection when selecting new system type

IainSAP:
Fix for https://github.com/SAP/open-ux-tools/issues/2723

- Reset connection on system type change
- Return true to defer validation to later prompts (otherwise the user is blocked)


cianmSAP (approved):
Changes are clear
Tests to cover
Patch changeset ✅

devinea (approved):
code changes are clear and tested. 
changeset ✅

Issue #2731, opened 2025-01-02
FEATURE - Detecting and showing OData failures in Info Center


nikmace:
_As a **developer**,
I want to **see when the OData has failed**,
so that I **better verify the OData connectivity right away when starting the editor**_

### Description

_We need to display the OData status in the Info Center and alert the user when an error occurs._

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have an adaptation project
**when** I run the editor and there is a problem with OData connectivity
**then** I will see an error in the Info Center.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #2732, opened 2025-01-02
feat: Detecting and showing OData failures in Info Center

nikmace:
Feat for #2731.
- Adds a new server route to check all data sources in the merged manifest.
- The result is then handled in the UI to be displayed in the Info Center.


nikmace:
Changeset will be put later, because this PR depends on another PR.

Issue #2734, opened 2025-01-06
TBI - Allow custom ABAP instances to be filtered during the CF discovery flow


longieirl:
### Description (include screenshots)
Currently, the CF discovery flow is dependent on the `@sap/cf-tools` to retrieve a list of ABAP instances based on a predefined and known list, for example;

https://github.com/SAP/open-ux-tools/blob/3e0f5d4dd51ffe9e281abf3f67f0e972b60599ba/packages/inquirer-common/src/prompts/cf-helper.ts#L19

During the CF flow, as shown below, the user might want to expose other types of ABAP instances that might not be defined using the predefined list as above;

![Image](https://github.com/user-attachments/assets/26b3506c-3f9b-448c-a7ec-e0b059f2b630)

In some instances, customers might want to expose other known service types. To support this, the CF discovery should allow additional service types to be added to the filtered list, for example;

Load from `.env` file;

```env
ABAPEnvServiceTypes=internal,test
```

Load from shell;

```bash
ABAPEnvServiceTypes="internal,test" node app.js
# or pre-scripting your env variables
export ABAPEnvServiceTypes='internal,test"
```

### Value
Provides an additional way to load ABAP instance types.

### Architecture Elaboration
No

### Tasks
- [ ] Read additional service types from `process.env`
- [ ] Add new unit tests

longieirl:
Validated in tools-suite on BAS and VSC (Mac and Windows);

```bash
echo 'export ABAPEnvServiceTypes="abap-int-tests-staging"' >> ~/.bashrc
# Reload shell
source ~/.bashrc
# Refresh page
```

PR #2735, opened 2025-01-06
2734/allow support to apend abap filter types

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/2734

Add support to read append additional ABAP service types during the CF discovery flow.


packages/inquirer-common/src/prompts/cf-helper.ts
```diff
@@ -26,7 +26,14 @@ export async function getCFAbapInstanceChoices(
             AbapEnvType.ABAP_STAGING,
             AbapEnvType.ABAP_INTERNAL_STAGING
         ];
-        const serviceInstanceInfo: ServiceInstanceInfo[] = await getServicesInstances(filteredInstances);
+        // Load additional ABAP service types to extend the filtered instance list
+        const envFilteredInstances = process.env.ABAPEnvServiceTypes
```
IainSAP, line 30: Can we pass as an option instead of loading from the env var? The consuming generator should load and pass.

longieirl: Based on further investigation, the existing filtered list is not static and can change from different subaccount. For now, the approach to use an env variable to update the list will be required and a TBI will be generated to investigated how best to handle this without requiring any change from the user.

IainSAP (approved):

cianmSAP (approved):
Updates are clear and doc'd
Changeset

Issue #2712, opened 2024-12-17
BUG - `odata-service-inquirer` : multiple bugs


IainSAP:
### Description

Following testing a number of issues were discovered:

- Some validation messages need improving
- Credentials prompts are being displayed for destinations are already configured with auth (validation message to be improved)
- Updating of backend system did not work when empty uname/pword already persisted

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2736, opened 2025-01-06
fix(inquirer-common): Updates destination connection error messages

IainSAP:
Additional fixes for https://github.com/SAP/open-ux-tools/issues/2712

- Rename GA link property name to be correct
- Updates specific 404 message for destinations that are returning not found.


longieirl (approved):
Changes look good.
Changeset is good.

kjose90 (approved):
Changes looks good. Changeset ✅

Issue #2694, opened 2024-12-12
TBI - refactor object clones (JSON.parse(JSON.stringify(...)))


heimwege:
### Description (include screenshots)
instead of `JSON.parse(JSON.stringify(...)))` we should rather use [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm)

### Value
Using `JSON` has several limitations w.r.t circular objects, Map, Set, Date, RegEx

### Architecture Elaboration
Does it requires architecture elaboration? 🤷🏻 

### Notes
`structuredClone` might not be supported on older browser versions. See https://caniuse.com/?search=structuredClone

### Tasks
- [x] search and replace

heimwege:
@Klaus-Keller you thoughts?

Klaus-Keller:
@heimwege, I agree, most of the code runs on node and using `structuredClone()` shouldn't cause any issue. Let's switch.

heimwege:
fixed in #2738

PR #2738, opened 2025-01-07
fix: refactor object clones

heimwege:
https://github.com/SAP/open-ux-tools/issues/2694

### DIFF
--- a/packages/fiori-annotation-api/src/cds/preprocessor.ts
+++ b/packages/fiori-annotation-api/src/cds/preprocessor.ts
@@ -112,7 +112,7 @@ class ChangePreprocessor {
             ) {
                 continue;
             }
-            const newChange = JSON.parse(JSON.stringify(change));
+            const newChange = structuredClone(change);
             newChange.index = 0;
             this.commands.set(i, {
                 type: 'replace',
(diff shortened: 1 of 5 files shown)


packages/fiori-annotation-api/src/cds/preprocessor.ts
```diff
@@ -112,7 +112,7 @@ class ChangePreprocessor {
             ) {
                 continue;
             }
-            const newChange = JSON.parse(JSON.stringify(change));
+            const newChange = structuredClone(change);
             newChange.index = 0;
```
heimwege, line 116: @voicis currently `newChange` is type any, but after my adjustment its type `CDSDocumentChange`. And `CDSDocumentChange` does not have a property `index`. Is this property really needed or is this a leftover/copy&paste/whatever that can be deleted?

image

voicis: Yes, it is needed. If the tests are passing then we might be missing a test case for this. Typescript can't correctly infer the `change` type after the previous if statement, because it can't interpret `includes`. Line `115` should only run with insert change objects which have an optional `index` property. Here we set it to 0, if the element where we want to insert doesn't have any children (there used to be an issue if it was `undefined` in this case).

heimwege: Thanks a lot. I adjusted the type in 6f05741a76070d69401c5645f9410bc528628e55.

voicis (approved):
- Code changes looks good
- Excellent test coverage
- Did not test locally

Klaus-Keller (approved):
Thanks @heimwege!
* update of README and switch to `structuredClone()` looks good
* changeset exists
* test coverage remains great
* did a visual review, no manual testing

Approved from my side

Issue #2739, opened 2025-01-07
BUG - generated library uses deprecated functions


docirl:
### Description
Generating a new SAP Reuse Library has files that reference deprecated functions in SAPUI5. Running the linting script on the generated library will show the errors:

![Image](https://github.com/user-attachments/assets/b89b9747-1ea1-43df-b120-50c37634cf9a)

### Steps to Reproduce
1. Generate a new Reuse Library using `@sap-ux/ui5-library-sub-generator`
2. Run linter (npm run lint)
3. See errors

### Expected results
No linting errors

### Actual results
Deprecated functions used and lint errors

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

docirl:
Done with https://github.com/SAP/open-ux-tools/pull/2741

PR #2741, opened 2025-01-07
fix(ui5-lib-writer): update deprecated code in generated library

docirl:
https://github.com/SAP/open-ux-tools/issues/2739

- generated library (using latest SAPUI5 version) uses deprecated functions and code
- update templates (JS and TS) for library writer to add version folders, as updates would not be backward compatible, writer will determine which template to use based on UI5 version selected during generation
- add test to cover JS and TS for new templates for >1.120.0

File update overview (file change tab is not very clear):
- created version folders in `common`
- updated files in 1.120.0 only, all files in 1.71.0 are the same as before

Files updated:
- ExampleRenderer.ts/js
- library.ts/js
- /test/Example.html
- ui5.yaml

Issues addressed:
image


packages/ui5-library-writer/src/index.ts
```diff
@@ -6,6 +6,8 @@ import cloneDeep from 'lodash/cloneDeep';
 import type { UI5LibConfig } from './types';
 import { enableTypescript } from './options';
 import { mergeWithDefaults } from './data';
+import type { UI5 } from '@sap-ux/ui5-application-writer';
+import { getTemplateVersionPath } from '@sap-ux/ui5-application-writer';
```
devinea, line 10: `@sap-ux/ui5-application-writer` shouldn't be a dependency for `ui5-library-writer` (Seems is is unused dependency before this and should be removed from the package.json. 
getTemplateVersionPath and UI5 should moved to common module or duplicated.

devinea (approved):
@docirl thanks for addressing the comments. 

changes look good. 
changeset 👍

cianmSAP (approved):
Code updates are clear
Full test coverage
Changeset
LGTM

PR #2742, opened 2025-01-07
feat(odata-service-inq): fetch credentials after system has been selected

cianmSAP:
**POC** 

Rather than using the 'hybrid' approach which fetch both the system details from `.fioritools/systems.json` & the credentials from the store

We would split up these calls, i.e
 1. Fetch the systems from the file using `getFilesystemStore(). getAll()` to be used in system selection prompt
 2. Use the `SystemService` API to and read the credentials of specific system

### DIFF
--- a/packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts
+++ b/packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts
@@ -37,57 +37,63 @@ export type CfAbapEnvServiceChoice = typeof CfAbapEnvServiceChoice;
  * @returns the validation result of the backend system connection
  */
 export async function connectWithBackendSystem(
-    backend: BackendSystem,
+    backendSystem: BackendSystem,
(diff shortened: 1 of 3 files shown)


packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts
```diff
@@ -43,41 +43,57 @@ export async function connectWithBackendSystem(
...
     let connectValResult: ValidationResult = false;
-    if (backendSystem) {
+    let backendSystemWithCreds: BackendSystem | undefined;
+    // Fetch the credentials for the system
+    try {
+        backendSystemWithCreds = await new SystemService(LoggerHelper.logger).read(
+            BackendSystemKey.from(backendSystem) as BackendSystemKey
+        );
+    } catch (error) {
+        LoggerHelper.logger.error(t('errors.storedSystemReError', { systemName: backendSystem.name, error }));
```
IainSAP, line 53: Does this occur when the user denies access, if so it should probably be a warning rather than an error?

PR #2746, opened 2025-01-08
Add Optional mem-fs-editor to `getProjectType` and underlying functions

kjose90:
#2762

Enhance the `getProjectType` function and its underlying logic by adding support for an optional `mem-fs` editor. This will allow to access and manipulate the in-memory file structure.


devinea (dismissed):
I have not tested manually. 
Changeset ✅

packages/project-access/src/project/access.ts
```diff
@@ -322,7 +322,14 @@ export async function createApplicationAccess(
...
             throw new Error(`Could not find app with root ${appRoot}`);
         }
-        const project = await getProject(app.projectRoot);
+        let memFs: Editor | undefined;
+        if (fs && 'fs' in fs) {
+            const { fs: fsFromOptions } = fs;
+            memFs = fsFromOptions;
+        } else if (fs) {
+            memFs = fs as Editor;
+        }
```
Klaus-Keller, line 331: there is a check for editor or options already in line 335-337. Would it be sufficient to move the check before call to `getProject()` and do something like
```typescript
const project = await getProject(app.projectRoot, options?.fs)
```
?

packages/project-access/src/project/cap.ts
```diff
@@ -63,30 +64,70 @@ export function isCapNodeJsProject(packageJson: Package): boolean {
...
+ * @returns {Promise<boolean>} - Resolves to `true` if files are found in the `srv` folder; otherwise, `false`.
+ */
+async function checkFilesInSrvFolder(srvFolderPath: string, memFs?: Editor): Promise<boolean> {
+    if (!memFs) {
+        return await fileExists(srvFolderPath);
+    }
+    // Load the srv folder and its files into mem-fs
+    // This is necessary as mem-fs operates in-memory and doesn't automatically include files from disk.
+    // By loading the files, we ensure they are available within mem-fs.
+    if (fs.existsSync(srvFolderPath)) {
```
Klaus-Keller, line 93: why are we using sync file access functions here and not async functions from `file/file-access.ts` and, if not there from `fs/promises`?

Klaus-Keller (commented):
Hi @kjose90, changes to support mem-fs-editor look good, please see my comments

packages/project-access/src/project/access.ts
kjose90: Thanks for pointing this out. fix [here](https://github.com/SAP/open-ux-tools/pull/2746/files#diff-a09e853a6ffc443564c0b3fa401de2f1d65fa410f1ca436b42ac7aba8e25b1a7R329).

packages/project-access/src/project/cap.ts
kjose90: Yes - I replaced fs sync functions using async functions from `file/file-access.ts `
[See here](https://github.com/SAP/open-ux-tools/pull/2746/files#diff-96d0f09954dec56941c14f476ad806708f321e3d7885ea00e9c58c8388f2f6c8R92).

Klaus-Keller (approved):
Thanks @kjose90, all my comments are addressed
* code changes to enhance support of mem-fs-editor in project access look good
* changeset is correct
* test coverage is great

Approved from my side.

devinea (approved):
Re-approving

PR #2747, opened 2025-01-08
fix: outline

voicis:
There is a bug where getting properties for table macro using a path to a different entity than the current page throws an error.

Added additional error handling, so in case of an error only a single node in outline is not displayed instead of the whole outline.


vadson71 (approved):
Changes are simple and straightforward
Didn't test locally

Jimmy-Joseph19 (approved):
Changes looks good. Approving. I Haven't tested locally

PR #2748, opened 2025-01-08
fix(@sap-ux/jest-environment-ui5): Do not reassign global.navigator

drktfl:
Node.js introduced a readonly `global.navigator` property with Node.js 21. That is, attempting to (re-)assign the property now breaks.

It turns out that the `global.navigator` object is not used in the Jest environment. `window.navigator` _is_, but that property is unaffected by this change.


drktfl:
Replaced by #2750

PR #2754, opened 2025-01-08
fix: Handle mainService model for already existing mainService

broksy:
If `mainService` already exists in manifest models section and new service is being added with model `mainService`, then model is changed to "".


devinea (approved):
Changes are clear to reuse the model for "mainService" if already exists.
Changeset ✅

815are (approved):
checked:
1. code straight forward
2. new test is added
3. changelog persists

PR #2618, opened 2024-11-24
add quick action for create new annotation

Jimmy-Joseph19:
- Add New quick action for V2 and V4 app for creation of new annotation file.
- Update middleware glob pattern to read subdirectories from changes folder as well handling of deletion when changes are in subfolder
- Api for reading and writing annotation file

PR #2755, opened 2025-01-09
fix: update quick action on file change on workspace

Jimmy-Joseph19:
Contains fix for updating the quick action list when changes are made externally via generators/or manually by the user in workspace. For eg: There are generators to add or replace odata service so the running app should react to the change and update the quick action list.

Also covers the comments uncovered in PR
https://github.com/SAP/open-ux-tools/pull/2618

### DIFF
--- a/packages/adp-tooling/src/writer/changes/writers/annotations-writer.ts
+++ b/packages/adp-tooling/src/writer/changes/writers/annotations-writer.ts
@@ -37,7 +37,7 @@ export class AnnotationsWriter implements IWriter<AnnotationsData> {
             annotationsInsertPosition: 'END',
             dataSource: {
                 [annotationNameSpace]: {
-                    uri: `../annotations/${fileName}`,
+                    uri: `../changes/annotations/${fileName}`,
                     type: 'ODataAnnotation'
                 }
             }
--- a/packages/adp-tooling/src/writer/changes/writers/annotations-writer.ts
+++ b/packages/adp-tooling/src/writer/changes/writers/annotations-writer.ts
@@ -37,7 +37,7 @@ export class AnnotationsWriter implements IWriter<AnnotationsData> {
             annotationsInsertPosition: 'END',
             dataSource: {
                 [annotationNameSpace]: {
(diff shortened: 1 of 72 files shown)


packages/control-property-editor/test/unit/ws-middleware.test.ts
```diff
@@ -0,0 +1,52 @@
...
+        next = jest.fn();
+
+        const middleware = webSocketMiddleware(mockStore);
+        invoke = (action: AnyAction) => middleware(next)(action);
+
+        mockWebSocket = {
+            addEventListener: jest.fn()
+        } as unknown as WebSocket;
+
+        global.WebSocket = jest.fn(() => mockWebSocket) as any;
```
voicis, line 31: This does not get cleaned up after tests.

nikmace (approved):
Changes look good
Coverage is good
Did not test manually

packages/adp-tooling/src/writer/changes/writers/annotations-writer.ts
```diff
@@ -37,7 +37,7 @@ export class AnnotationsWriter implements IWriter<AnnotationsData> {
             annotationsInsertPosition: 'END',
             dataSource: {
                 [annotationNameSpace]: {
-                    uri: `../annotations/${fileName}`,
+                    uri: `../changes/annotations/${fileName}`,
```
nikmace, line 40: Actually, just understood that this change was not tested and we don't know how it will behave when deployed. Therefore, we will have a sync with a relevant team today to discuss this. If we are in a rush to merge this it is better this change to not be included in this PR.

voicis: This change is required, otherwise the quick action or command is pointless, because the annotation file then would be ignored since it can't be located.

nikmace (dismissed):
Changes look good
Coverage is good
Comments addressed
Did not test manually

voicis (approved):
- Code changes looks good
- Good test coverage
- Review comments addressed
- Tested locally

Issue #2756, opened 2025-01-09
BUG - Hide `semantic object` & `action` FLP prompts when there are inbounds for an Adaptation Project


nikmace:
### Description
Currently, `semantic object` & `action` prompts from `flp-inquirer` are shown when there are inbound IDs present in the merged manifest. The user is forced to enter them, but then they are not used by the flp writer. Thus, there is no need to show them to the user, valid only for adaptation projects.

### Steps to Reproduce
1. Execute CLI command for adding inbound config to a project that has inbound IDs.
2. In the terminal you will be prompted to choose an inbound ID and then the two `semantic object` & `action` prompts

### Expected results
`semantic object` & `action` prompts are not shown when there are existing inbound IDs

### Actual results
`semantic object` & `action` are visible and user is forced to enter data

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2758, opened 2025-01-09
fix: Hide `semantic object` & `action` FLP prompts when there are inbounds for an Adaptation Project

nikmace:
Fix for #2756.
- Adds the when condition for `semantic object` & `action` prompts from `flp-inquirer` module.
- Fixes incorrect subTitle property when writing i18n file content for FLP configuration.

### DIFF
--- a/.changeset/grumpy-shirts-complain.md
+++ b/.changeset/grumpy-shirts-complain.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/flp-config-sub-generator': patch
----
-
-tidy up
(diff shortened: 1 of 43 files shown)


cianmSAP (dismissed):
Updates look good
From a Fiori gen FLP standpoint, `inboundId` prompt is hidden and therefore will always be falsy

I don't think patch bump is required for `@sap-ux/create` this will be handled automatically as this inquirer is a dependency of `@sap-ux/create`

cianmSAP (approved):
Re-approving

GDamyanov (approved):
Code changes look good from ADP prospective.
Excellent test coverage.
Did not tested manually.

cianmSAP (dismissed):

GDamyanov (approved):
Re-approving

cianmSAP (approved):

Issue #2702, opened 2024-12-12
TBI - Expose functionality to facilitate the creation of SAP BTP subaccount destinations during the CF discovery flow


longieirl:
### Description (include screenshots)
Currently, during the CF discovery flow on BAS, when you select an ABAP environment, it will show you the appropriate ABAP service to allow you to connect to your ABAP cloud instance. 

When you select this ABAP environment system, it will generate a SAP BTP subaccount destination or update the existing destination using the following naming convention;

```
abap-cloud-<abap-environment-name-<cf-target-org>-<cf-target-space>
```

The SAP BTP destination is created using `OAuth2UserTokenExchange` and will append the appropriate UAA settings to allow connectivity to the backend.

The goal is to leverage the [cf-tools](https://www.npmjs.com/package/@sap/cf-tools) to generate the destination as the flow is only required for BAS.

Existing flow;

![Screenshot 2024-12-12 at 19 53 53](https://github.com/user-attachments/assets/f36eef3f-69c4-4c61-91c1-d9ab497349db)

For more information on this specific destination type;

https://help.sap.com/docs/connectivity/sap-btp-connectivity-cf/oauth-user-token-exchange-authentication

### Value
- remove dependency on internal module
- more maintainable code

### Tasks
- [x] Expose new functionality from axios-extension module
- [x] Leverage cf-tools and axios modules to validate and generate a new SAP BTP destination using the BAS `H2O_URL` proxy to call the create API endpoint

longieirl:
Change has been merged into bas-sdk, waiting on release https://www.npmjs.com/package/@sap/bas-sdk?activeTab=versions v3.11.2 or higher to consume the changes.

longieirl:
Validated in the latest release of SAP Fiori tools and working as expected.

PR #2760, opened 2025-01-09
2702/expose new btp destination creation flow 

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/2702

- changes applied to btp-utils
- leverage cf-tools to obtain UAA credentials, used to generate a new OAuth2TokenExchange BTP destination
- changes validated on BAS

Note:
- There is no exception thrown if the destination is already created on the target subaccount
- Exception thrown if the user is not logged into cf
- Exception thrown if subaccount doesn't have a dev space created
- If the SAP BTP destination already exists, existing properties are not updated if already present on the target destination


cianmSAP (dismissed):
Tests indicate new functionality clearly
Excellent test coverage
Would this new functionality not warrant a minor version bump?

longieirl:
Closing this PR, changes made to @sap/bas-sdk now reflect this flow.

PR #2768, opened 2025-01-13
fix(abap-deploy): updates for exposing package and transport prompts

cianmSAP:
https://github.com/SAP/open-ux-tools/issues/2769

- Exposes two 'groups' of prompts - `getPackagePrompts()`, `getTransportRequestPrompts`
- Adds parameters to exposed functions 
        - `useStandalone`: if the prompts are being used standalone, defaulting to true and set to false internally. This helps to determine whether the backendTarget should be used and any values for system configuration in the prompt state should be ignored
        - `isYUI`: like `getPrompts` pass in whether the prompts are ran in a YUI context
- Updates `packages/abap-deploy-config-inquirer/src/service-provider-utils/abap-service-provider.ts` to use a class with static props/methods


packages/abap-deploy-config-inquirer/src/prompts/validators.ts
```diff
@@ -469,40 +486,48 @@ async function handleListExistingTransportChoice(
...
  * @param input - transport choice input
  * @param previousAnswers - previous answers
  * @param validateInputChanged - if the input has changed
  * @param prevTransportInputChoice - previous transport input choice
  * @param backendTarget - backend target
+ * @param ui5AbapRepoName - ui5 app repository name derived from AbapDeployConfigPromptOptions[ui5AbapRepo]
  * @returns boolean or error message as a string
  */
 export async function validateTransportChoiceInput(
+    useStadalone: boolean,
```
longieirl, line 499: Spelling!

cianmSAP: Thanks, updated

IainSAP (approved):
Thanks @cianmSAP lgtm

- Test coverage is great
- Updates are clear
- Not tested locally

longieirl (approved):
Changes look good.
Changeset looks good.
Appropiate test coverage.
Will be test when consumed into tools-suite.

Issue #2770, opened 2025-01-13
BUG - UISections. Refresh sections/splitter sizes, when external sizes are changed


815are:
### Steps to Reproduce
1. We have UISections like:
```
const [firstSectionSize, setFirstSectionSize] = useState(200);
<UISections sizes={[firstSectionSize, undefined, 200]] />
```
2. Initial size of `firstSectionSize` is `200`;
3. Then we change `firstSectionSize` from `200` to `300` using `setFirstSectionSize(300)`;

Problem - size of sections/splitter is not recalculated.

### Expected results
Size of sections/splitter is recalculated when external sizes are changed.

### Actual results
Size of sections/splitter is not recalculated when external sizes are changed.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2771, opened 2025-01-13
fix(ui-components): recalculate sections sizes when external property sizes are changed

815are:
#2770


Klaus-Keller (approved):
Thanks @815are!
* code to recalculate sizes looks good
* change set exists
* test coverage is great

Approved from my side

PR #2791, opened 2025-01-16
Update pipeline.yml with actions/upload-artifact@v4

IainSAP:
Testing just updating version. Dont think `overwrite` option is required as the playwright reports seem to be retained and therefore must be using a rolling file name.


cianmSAP (approved):
Thanks @IainSAP

PR #2793, opened 2025-01-17
feat(app-config-writer): remove sap-client from start-variants-management script url parameters

heimwege:
- remove sap-client from start-variants-management script url parameters
- refactor min version management (introduce semver)


ytpo-lyne (approved):
Thanks for the changes, @heimwege!

- code deletions for sap-client are fine
  - sap-client should not be set over scripts and is overruled by yaml config
-  changeset present
- code coverage adjusted

Approving.

devinea (approved):
Changes are nice clean up.

PR #2794, opened 2025-01-17
fix: Setting default name for the service model

broksy:
Added additional check for service name in `setDefaultServiceModel`.
Additional check makes sure that `""` is used only for `mainService`, but currently `""` model is used for every service that is being added. Check screenshot:
image

New service is being added. Name of it is `STTA_SALES_ORDER_ND_SRV_01` (not `mainService`). `mainService` already exists and `""` is going to be used for model which is not correct behavior in this case. Service name `STTA_SALES_ORDER_ND_SRV_01` should be used instead.


devinea (approved):
changeset ✅

815are (approved):
checked:
1. code - change simple and straight forward;
2. new test is added
3. changeset presented

Issue #2798, opened 2025-01-20
BUG - Aggregation alias not supported for ALP V4 entities


cianmSAP:
### Description
During generation with a V4 Odata service, the entity list for ALP template type is filtered to ensure only those that are suitable for ALP are displayed. The current logic is:

Validate that the entity has an "Aggregation.ApplySupported" annotation defined.
This annotation has a Property value of "Transformations".

### Steps to Reproduce
- Go to BAS
- Launch the Fiori generator
- Choose ALP as your template type.
- Choose a SAP System
- Choose a service that has an alias that can be used instead of the annotation term defined above

### Expected results
Service is usable 

### Actual results
Service is not usable and prevent at the entity selection step

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

cianmSAP:
Fix is merged

PR #2800, opened 2025-01-20
fix(odata-inq): add check on entity type for aggregation annotation

cianmSAP:
#2798

As `@sap-ux/annotation-converter` resolves the alias, the only additional logic is needed is to add a check on the entity type for aggregation apply supported annotations (as a fallback).


longieirl (approved):
Changes look good.
Tests cover change.
Changeset is good.

IainSAP (approved):
Thanks @cianmSAP 

- Minimal change to address issue
- Test to cover
- Not tested locally

PR #2802, opened 2025-01-21
fix: Getting of webapp path

broksy:
Improved `getWebappPath` for projects that has relative webapp path defined in ui5.yaml.
For example, project with structure like this:
.app
..app1
...webapp
....ui5.yaml (`webapp: app/app1/webapp`)

`getWebappPath('app/app1')` previously returned: `app/app1/app/app1/webapp`.


packages/project-access/src/project/ui5-config.ts
```diff
@@ -1,25 +1,46 @@
...
+    let currentDir = resolve(projectRoot);
+    while (currentDir !== parse(currentDir).root) {
+        const packageJSONPath = join(currentDir, FileName.Package);
+        if (await fileExists(packageJSONPath, memFs)) {
+            return currentDir;
+        }
+        currentDir = dirname(currentDir);
+    }
+    return projectRoot;
+}
```
Klaus-Keller, line 24: @broksy, perhaps we could reuse function [`findFileUp()`](https://github.com/SAP/open-ux-tools/blob/960b79abb117ce0ac846338360235b8596edbdd7/packages/project-access/src/file/file-search.ts#L137C23-L137C33) for this?

broksy: Yes. Thanks. Almost identical function, but shorter.

Klaus-Keller (approved):
Thanks @broksy,
* code change to find root with `package.json` looks good
* change set exists
* test coverage is great

Approved from my side.

815are (approved):
checked:
1. code changes simple to understand and meets expectation
2. new test is added
3. changelog exists

PR #2806, opened 2025-01-22
feat: add support for external HTTP calls

nlunets:
When doing component based testing it might be useful to refer to an actual mockserver. 

Furthermore if we want to leverage more complex snapshot (including screenshots), being able to load CSS is desired, so an option was added for that.


voicis (approved):
- Changes looks good
- Changeset present
- Did not test locally

longieirl (approved):
Changes look good.
Changeset is good.
Tests are good.

PR #2807, opened 2025-01-22
fix: enabled missing quick actions on alp v2

vadson71:
Enabled quick actions available on LRP, but missing on ALP in v2 apps


nahmed22 (approved):
changes looks good
Did not test locally
changeset maintained.

Issue #2813, opened 2025-01-22
FEATURE - Add stable IDS for all components in AddFragment and ControllerExtension forms


GDamyanov:
_As a **developer**,
I want to **write automation tests for CPE for ADP use case and I want AddFragment and ControllerExtension forms to have stable IDS**,
so that I **I can use simple selectors**_

### Description

AddFragment and ControllerExtension for should be enhanced withs table IDS.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a ...
**when** I execute ...
**then** ... will happen.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #2816, opened 2025-01-22
feat: Add stable Ids in AddFragment and ControllerExtension forms

GDamyanov:
feat for #2813 

- Add stable Ids in AddFragment and ControllerExtension forms


voicis (commented):
We probably should also make similar changes to the TableColumn dialog https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/src/adp/ui/AddTableColumnFragments.fragment.xml

testojs (approved):
* simple and easy to understand change
* manually tested - works as expected

testojs:
> We probably should also make similar changes to the TableColumn dialog https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/src/adp/ui/AddTableColumnFragments.fragment.xml

Yes, it makes sense to add the IDs. Also, with the current implementation, the first link does not point to the respective select control (`for` attribute, even not explicitly set, will point to HBox now, which is wrong). This change also improves form semantics.

voicis (approved):
- changes looks good
- changeset present
- did not test locally

PR #2823, opened 2025-01-23
Adds annotations warning to odata service urls prompts

IainSAP:
WIP. Suggested updates to support odata service url annot warns.


IainSAP:
Done elsewhere

PR #2825, opened 2025-01-24
fix(flp-inquirer): fix hiding flp prompts

cianmSAP:
Related to https://github.com/SAP/open-ux-tools/issues/2756

To prevent default methods being ran, and answers automatically populated, if the `hide` prompt option is set for a specific prompt it should not be returned from `getPrompts`

This will also enhance performance in YUI so that conditions e.g `when` `default` are not ran everytime

### DIFF
--- a/packages/flp-config-inquirer/src/prompts/prompts.ts
+++ b/packages/flp-config-inquirer/src/prompts/prompts.ts
@@ -58,33 +58,12 @@ export function getQuestions(
         [promptNames.createAnotherInbound]: getCreateAnotherInboundPrompt(isCLI)
     };
 
-    const questions: FLPConfigQuestion[] = [];
-
-    if (promptOptions?.[promptNames.inboundId]?.hide !== true) {
-        questions.push(keyedPrompts[promptNames.inboundId]);
-    }
-
-    if (promptOptions?.[promptNames.emptyInboundsInfo]?.hide !== true) {
-        questions.push(keyedPrompts[promptNames.emptyInboundsInfo]);
-    }
-
-    questions.push(
-        ...[
-            keyedPrompts[promptNames.semanticObject],
-            keyedPrompts[promptNames.action],
-            keyedPrompts[promptNames.overwrite],
-            keyedPrompts[promptNames.title],
-            keyedPrompts[promptNames.subTitle]
-        ]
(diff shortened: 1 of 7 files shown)


longieirl (dismissed):
Changes look good.
Change is good.
Test coverage is good.

packages/flp-config-inquirer/src/prompts/prompts.ts
```diff
@@ -58,27 +54,37 @@ export function getQuestions(
...
+        ]
+    );
+
+    if (promptOptions?.[promptNames.additionalParameters]?.hide !== true) {
+        questions.push(keyedPrompts[promptNames.additionalParameters]);
+    }
+
+    if (promptOptions?.[promptNames.createAnotherInbound]?.hide !== true) {
+        questions.push(keyedPrompts[promptNames.createAnotherInbound]);
+    }
```
nikmace, line 87: (optional) Alternatively, would it be better to map them and not repeat the same if statements?

```suggestion
    const questions: FLPConfigQuestion[] = Object.entries(keyedPrompts)
        .filter(([promptName]) => {
            const option = promptOptions?.[promptName as promptNames];
            return !hasHideOption(option) || !option.hide;
        })
        .map(([_, prompt]) => prompt);
```

And then the helper method:

```ts
/**
 * Type guard to check if the provided object has a `hide` property.
 *
 * @param option - The object to check.
 * @returns True if the object has a `hide` property of type `boolean`, otherwise false.
 */
function hasHideOption(option: any): option is { hide: boolean } {
    return option && typeof option.hide === 'boolean';
}
```

Here options are `any` type. If that is not okay that would mean you need to write all interfaces of the promptNames options, or create a union one.

nikmace (dismissed):
Changes look good
Changeset is good
Test coverage is good
Did not test manually

packages/flp-config-inquirer/src/prompts/prompts.ts
cianmSAP: I've taken a slightly different approach but with the same result i.e not having to write if statements for each one.

nikmace (approved):
Re-approving

longieirl (approved):
Re-approving

PR #2826, opened 2025-01-24
fix: "add custom table action" quick action in ALP v4 projects with multiple toolbars in tables

vadson71:
Fixed CPE `Add Custom Table' quick action bug in ALP v4 projects with multiple toolbars in chart and tables


packages/preview-middleware-client/src/adp/quick-actions/fe-v4/create-table-action.ts
```diff
@@ -24,28 +25,27 @@ export class AddTableActionQuickAction extends TableQuickActionDefinitionBase im
...
-                        this.context.rta,
-                        DialogNames.ADD_FRAGMENT,
-                        undefined,
-                        {
-                            aggregation: 'actions',
-                            title: 'QUICK_ACTION_ADD_CUSTOM_TABLE_ACTION'
-                        }
-                    );
-                }
+        const actionToolbars = getRelevantControlFromActivePage(this.context.controlIndex, this.context.view, [
```
voicis, line 28: Why are we using toolbar and not the tables `actions` aggregation? According to the documentation everything under content aggregation should not be modified as it may change and break when using a different UI5 version. We probably should avoid creating such changes.

packages/preview-middleware-client/src/adp/quick-actions/fe-v4/table-quick-action-base.ts
```diff
@@ -19,6 +20,7 @@ export abstract class TableQuickActionDefinitionBase extends QuickActionDefiniti
     isClearButtonEnabled = false;
     children: NestedQuickActionChild[] = [];
     tableMap: Record<string, number> = {};
+    smartTables: UI5Element[] = [];
```
voicis, line 23: Instead of making a new array, it would be better to use `tableMap` like in https://github.com/SAP/open-ux-tools/blob/408a1c5e36714a5fb99ce6602cf1c845d3ee8222/packages/preview-middleware-client/src/adp/quick-actions/table-quick-action-base.ts#L61-L74 or even better use that base class instead.

vadson71: Refactored

voicis (approved):
- Code changes looks good
- Review comments addressed
- Tested locally

Jimmy-Joseph19 (approved):
Changes looks good.
Test coverage good.
Didn't test locally
Changeset found for affected package

PR #2828, opened 2025-01-27
feat(preview-middleware): reroute missing url param viewCache

ytpo-lyne:
The `sap-ui-xx-viewCache=false` URL parameter ensures that changes made during development are reflected in the applications preview. Some projects and scripts are missing this parameter - In that case a reroute will happen and the parameter will be added.


heimwege (approved):
code changes are straightforward
changeset ok
excellent code coverage
did NOT test manually but witnessed manual test 🐱

Issue #2829, opened 2025-01-27
BUG - Cannot bind i18n models different than 'i18n'


GDamyanov:
### Description
Cannot bind i18n models different than 'i18n'

### Steps to Reproduce
1. Start CPE on some ADP project which has List Report and Object Page
2. In Object Page i18n.properties add some key with value
3. Find some control to which you can bind model
4. Fill property value with: {i18n|sap.suite.ui.generic.template.ObjectPage|SEPMRA_C_PD_PRODUCT>{key added in i18n properties}}

### Expected results
Control text should be changed to value in i18n

### Actual results
Control text is empty

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

heimwege:
@GDamyanov is that one still open or has it been fixed with #2830 ?

heimwege:
Will close as there has been no response for quite a while. So assumption is it has been fixed with #2830

PR #2830, opened 2025-01-27
fix: Bind i18n models with namespace

GDamyanov:
Fix for #2829 

- Bind i18n models that contains namespace

### DIFF
--- a/packages/preview-middleware-client/test/unit/cpe/changes/validator.spec.ts
+++ b/packages/preview-middleware-client/test/unit/cpe/changes/validator.spec.ts
@@ -11,31 +11,45 @@ describe('vaildateBindingModel', () => {
             )
         })
     };
-    test('should throw when invalid binding model string is provided', () => {
+    test('should throw when invalid binding model string is provided', async () => {
         try {
-            validateBindingModel(mockModifiedcontrol as unknown as UI5Element, '{}');
+            await validateBindingModel(mockModifiedcontrol as unknown as UI5Element, '{}');
         } catch (error) {
             expect(error).toBeInstanceOf(SyntaxError);
             expect(error.message).toBe('Invalid binding string.');
         }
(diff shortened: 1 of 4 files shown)


nikmace (approved):
Change looks good
Coverage is good
Did not test manually

packages/preview-middleware-client/test/unit/cpe/changes/validator.spec.ts
```diff
@@ -11,32 +11,46 @@ describe('vaildateBindingModel', () => {
...
             );
         }
     });
+
+    test('should throw error when invalid binding model is provided', async () => {
+        const control = {
+            getModel: jest.fn().mockReturnValue(undefined)
+        };
+        try {
+            await validateBindingModel(control as unknown as UI5Element, '{ i18n|namespace>test }');
```
nlunets, line 48: Reading the initial issue it seems to me that you wanted to support this exact syntax, isn't that the case ?

GDamyanov: Yes, there are couple of i18n models with namespaces per application which you can use. Before my change we were supporting only the main "i18n" model.

nlunets: But your test expect that an error happens here correct ?
if not then your test is confusing, you should check that an error is not triggerred instead

voicis: Actually all of these tests looks a bit odd, since `expect` statements are in the catch block and would not be called if error is not thrown and the test will still pass.

nikmace (approved):
Re-approving
Tests adjusted

nikmace (approved):
Re-approving

PR #2831, opened 2025-01-27
feat: Delete one backend per service

broksy:
Replaced deletion logic from `filter` in `removeBackendFromFioriToolsProxydMiddleware` method to `findIndex` and `splice` combination to allow to delete only one backend instance for `fiori-tools-proxy middleware`.

Reason: Possible duplicated backend entries for multiple services from same system and filter deletes all of them.

### DIFF
--- a/.changeset/rude-lobsters-punch.md
+++ b/.changeset/rude-lobsters-punch.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui5-config': patch
+---
+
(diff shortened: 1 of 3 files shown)


devinea (approved):
Code changes look good. 
I haven't tested locally.
changeset ✅

815are (approved):
checked:
1. code changes - only single first entry is removed
2. new test is added
3. changeset is presented

PR #2833, opened 2025-01-27
Fix/preview middleware/ui5 version request protocol for karma

heimwege:
The `headers` object of a [connect](https://github.com/senchalabs/connect) API request does not know the `host` property. In this case we extract it from the `referer`.


heimwege:
@tobiasqueck this one should solve those connect API vs. express API issues once and for all. It includes tests for connect API and adjusts the types of the relevant routes (RTA will never be used by karma so we can leave them as it).

devinea (approved):
code changes seem straightforward.

PR #2834, opened 2025-01-27
fix: update the availability matix with the new action

nahmed22:
update the availability matrix with the new actions

1.  Enable Variant Management in Tables and Charts


Jimmy-Joseph19 (approved):
Changes looks good.
Approving

vinayhospete (approved):

PR #2837, opened 2025-01-28
[fix]: allow reentrance url alias

tobiasqueck:


packages/axios-extension/src/auth/reentrance-ticket/index.ts
```diff
@@ -38,9 +38,8 @@ export async function getReentranceTicket({
 
         // Open browser to handle SAML flow and return the reentrance ticket
         const scenario = process.env.FIORI_TOOLS_SCENARIO ?? 'FTO1';
-        const url = `${backend.uiHostname()}${ADT_REENTRANCE_ENDPOINT}?scenario=${scenario}&redirect-url=${redirectUrl(
-            redirectPort
-        )}`;
+        const endpoint = process.env.FIORI_TOOLS_REENTRANCE_ENDPOINT ?? ADT_REENTRANCE_ENDPOINT;
```
heimwege, line 41: I didn't find any other usage of `FIORI_TOOLS_REENTRANCE_ENDPOINT`. Do I get it right that this is a new env variable that is not yet being used?

devinea (approved):

Klaus-Keller (approved):
Thanks @tobiasqueck,
* change to configure reentrance endpoint look good
* change set exists

Approved from my side.

PR #2838, opened 2025-01-28
fix: make sure the canvas dependency is not required

nlunets:
With the previous change I removed the possibility to run the test without having `canvas` installed.
This is now enabled again, and `canvas` is only required when `allowCSS` is set to true


voicis (approved):

heimwege (approved):
changeset ok
excellent test coverage
code changes are straightforward
did NOT test locally

PR #2839, opened 2025-01-28
correct command type for add annotation file

Jimmy-Joseph19:
- PR includes fix for

1. undo/redo - for newAnnotation file change  (Issue was wrong command type).
2. Changed pending change tile from `Add Annotations To O Data` to `Add New Annotation File`


packages/adp-tooling/src/preview/adp-preview.ts
```diff
@@ -227,29 +227,33 @@ export class AdpPreview {
...
+                            logger
+                        );
+                    }
+                    break;
+                default:
+                    // no need to handle delete changes
+                    break;
+            }
+        } catch (e) {
+            throw e;
```
voicis, line 256: Why wrap this in try-catch if we're not doing anything with the exception?

nikmace (approved):
Change looks good
Coverage is ok
Did not test manually

PR #2845, opened 2025-01-29
fix: add check for initialized fiori integrity

Klaus-Keller:
When using module `@sap-ux/project-integrity`, there is currently no way to check if a Fiori project has enabled integrity check. The only way is with try and error. 

This pull request adds function `isFioriProjectIntegrityInitialized(projectRoot: string): boolean` to check, whether integrity has been enabled for a Fiori project without the need of `try`/`catch`.

### DIFF
--- a/.changeset/late-plants-accept.md
+++ b/.changeset/late-plants-accept.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-integrity': minor
+---
+
+Add function to check if Fiori project integrity is initialized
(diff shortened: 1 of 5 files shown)


marufrasully (approved):
Thanks @Klaus-Keller 
* New function looks god
* Great test coverage
* Did not test it locally

kjose90 (approved):
Changes are clear.
Tests covered.
Changeset ✅

PR #2859, opened 2025-02-04
Rename SAP FIORI TOOLS.svg to sap-fiori-tools.svg

gpeuc:
Just file rename


donal-tobin-sap (approved):
name change looks good

PR #2860, opened 2025-02-04
Update README.md - added logo

gpeuc:
Added official SAP Fiori tools logo to the file so there's visual connection


donal-tobin-sap (approved):
Looks good and renders nice on dark background

PR #2862, opened 2025-02-04
fix: Default model naming

broksy:
Fixes model naming.
For example OVP uses a default model named `mainModel`: https://github.com/SAP/open-ux-tools/blob/d90b2747a58c4c0e228cdc072cf64a2a623702ea/packages/fiori-elements-writer/src/data/defaults.ts#L146

During testing of odata-service-writer in fiori-elements-writer we found out that `mainModel` is overwritten and `""` is used for default model name.
We should not overwrite defined models.


IainSAP (approved):
Thanks @broksy 

- Changes seem appropriate
- Tests updated

devinea (approved):
Thanks @broksy . looks good.
Haven't tested locally.

PR #2871, opened 2025-02-05
Chore - Improve sonar scanner performance

devinea:
to improve scanner performance ~17minutes down to ~1 minute by removing mode_modules before scanning per https://community.sonarsource.com/t/sonarcloud-action-scan-is-slow-with-node-modules/78784


heimwege (approved):
Lgtm

donal-tobin-sap (approved):
Looks good.

Note we should review the time taken to delete as deletion on a docker fs (overlay fs) is slow and individual multiple deletion processes from find are slow, the exec process parallelism can be controlled - and ensure that the find is working breath first - so you do not bother attempting to delete sub folders called node_modues of the parent. rimraf with a glob is normally faster in most of my experience.

devinea:
> Looks good.
> 
> Note we should review the time taken to delete as deletion on a docker fs (overlay fs) is slow and individual multiple deletion processes from find are slow, the exec process parallelism can be controlled - and ensure that the find is working breath first - so you do not bother attempting to delete sub folders called node_modues of the parent. rimraf with a glob is normally faster in most of my experience.

I had tried rimraf without success https://github.com/SAP/open-ux-tools/pull/2871/commits/aea326cdbb63d99678c72e668629d69783cbe247 
Can you see what is wrong?

donal-tobin-sap:
> > Looks good.
> > Note we should review the time taken to delete as deletion on a docker fs (overlay fs) is slow and individual multiple deletion processes from find are slow, the exec process parallelism can be controlled - and ensure that the find is working breath first - so you do not bother attempting to delete sub folders called node_modues of the parent. rimraf with a glob is normally faster in most of my experience.
> 
> I had tried rimraf without success [aea326c](https://github.com/SAP/open-ux-tools/commit/aea326cdbb63d99678c72e668629d69783cbe247) Can you see what is wrong?

Actually, looking at the run is was in the region of 1s which is more than fast enough, no optimization required, so I would not worry. So fast in fact that I wondered if it ran at all, but there looks like there is no reference in the sonar step so I expect it is fine.

PR #2874, opened 2025-02-05
Fix/32779 bug resource path mapping is not always respected

Vladislavs-Silins-SAP:
The function `validateBasePath` now computes the manifestPath using the `webapp` path from `manifest.json`, rather than the previously hardcoded value 'webapp'.

Since `getWebappPath` is asynchronous, asynchronous calls were added wherever this function is used.


Klaus-Keller (approved):
Thanks @Vladislavs-Silins-SAP!
* changes to use `getWebappPath()` look good
* changeset exists
* test coverage is  great

Approved from my side

packages/fe-fpm-writer/src/common/validate.ts
```diff
@@ -43,12 +44,16 @@ export function validateDependenciesLibs(manifest: Manifest, dependencies: strin
...
+    basePath: string,
+    fs?: Editor,
+    dependencies = ['sap.fe.templates']
+): Promise<boolean> {
     if (!fs) {
         fs = create(createStorage());
     }
 
-    const manifestPath = join(basePath, 'webapp', 'manifest.json');
+    const manifestPath = join(await getWebappPath(basePath, fs), 'manifest.json');
```
815are, line 56: maybe we can call existing function `getManifestPath`, which does same is in code above? ->https://github.com/SAP/open-ux-tools/blob/fix/32779-bug-resource-path-mapping-is-not-always-respected/packages/fe-fpm-writer/src/common/utils.ts#L80

815are (approved):
checked:
1. code changes are straight forward - only noticed that existing method `getManifestPath` can be used to avoid small code duplication;
2. changeset persists

I did not test manually

packages/fe-fpm-writer/src/common/validate.ts
Vladislavs-Silins-SAP: Thanks. I've done this.

815are (approved):

Klaus-Keller (approved):
Re-approving after small improvement to use `getManifestPath()`

PR #2876, opened 2025-02-05
fix(cf-gen): fix unstable windows test

cianmSAP:
The `D:` is seems to be intermittently included in the path, and failing the expected test

image


longieirl (approved):
lgtm

Issue #2873, opened 2025-02-05
FEATURE - Scroll into view when clicking on a `control` that is not currently visible in the `iframe`


nikmace:
_As a **developer**,
I want to **click on a control in the outline tree**,
so that I **can see the control if it is not currently visible**_

### Description

_When clicking in the outline on a control that is currently not visible in the iframe the iframe does not scroll to locate this control rather you have to find it yourself._

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have started an Adaptation Editor
**when** I click on a control that is not currently visible (for example a column list in a table)
**then** the iframe scrolls to that dom element

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #2877, opened 2025-02-06
feat: Scroll into view when clicking on a `control` that is not currently visible in the `iframe`

nikmace:
Feat for #2873.
- Calls the `scrollIntoView` method on the dom element of the control.


mmilko01 (approved):
- Changeset OK
- Code looks good
- Did not test manually, but witnessed a manual test

Jimmy-Joseph19 (approved):
Changes looks good.
Changeset found for the respective change.
Did not test locally
Approving

PR #2884, opened 2025-02-06
fix: cds compile error with memfs

voicis:
#2890


vadson71 (approved):
Changes look reasonable
Test coverage is excellent
Didn't test locally

marufrasully (approved):
Thanks @voicis 
* Changes looks good to me
* Secured with unit test
* Did not test it locally

PR #2894, opened 2025-02-07
feat(preview-middleware): introduce 'editors' property on preview config root level

heimwege:
As there might be other editors than RTA in the future we deprecate the current RTA configuration and introduce a new one where `editors` is on the root level.

Fun fact: I think that test snap I adjusted has been wrong all the time. Don't know why nobody noticed it so far 🤷🏻


packages/preview-middleware/README.md
```diff
@@ -13,27 +13,32 @@ When this middleware is used together with the `reload-middleware`, then the ord
...
+| `adp.ignoreCertErrors`  | `boolean` | `false`          | Optional setting to ignore certification validation errors when working with e.g. development systems with self signed certificates                                                                                                                       |
+| `rta`                   |           |                  | 🚫 *Deprecated: use 'editors.rta' instead* <br/> Optional configuration allowing to add mount points for runtime adaptation                                                                                                                               |
+| `rta.layer`             | `string`  | `(calculated)`   | Optional property for defining the runtime adaptation layer for changes (default is `CUSTOMER_BASE` or read from the project for adaptation projects)                                                                                                     |
```
tobiasqueck, line 30: I'd remove the two entries `rta.layer` and `rta.editors` here and just keep `rta` with the current description. If people already have it in their configs fine, otherwise, they don't need to know how to use a deprecated config.

heimwege: also remove the rta.editors section? https://www.npmjs.com/package/@sap-ux/preview-middleware#configuration-option-rtaeditors

Klaus-Keller (approved):
Thanks @heimwege,
* implementation of new config with backward compatibility looks good
* changeset exists
* test coverage is great

Approved from my side.

PR #2899, opened 2025-02-10
fix: bump playwright and use chrome as browser

marufrasully:
This PR
* Bump playwright version
* Adapt Github action to only install Chrome browser - as this is the only browser UI integration tests are running.


devinea (approved):
Thanks @marufrasully

PR #2902, opened 2025-02-11
Fix: The adaptive card manifest should not be saved.

Sharmavksh:
This change will prevent saving of an empty adaptive card manifest.


SRKPendyala (approved):
Looks good for me

Klaus-Keller (approved):
Thanks @Sharmavksh.
* code to not write `adaptive-` file looks good
* changeset exists
* tests are adjusted

Approved from my side

Issue #2589, opened 2024-11-18
TBI - use project-acces->getWepappPath instead of 'webapp’


heimwege:
### Description (include screenshots)
Currently not everywhere the `getWebappPath` method is being used. Instead there's `'webapp'` used in some places. As the `webapp `folder does not necessarily need to be named `'webapp'` we should clean this up.

Occurrences:
-	~~odata-service-writer/src/index.ts - 174~~ #2931
-	~~odata-service-writer/src/updates.ts - 330, 337~~ #2931
-	~~i18n/readme.md - 47~~ #2931
-	~~mockserver-config-writer/src/mockserver-config - 50, 63~~ #2931
-	~~preview-middleware/readme.md - 191~~ #2857
-	~~preview-middleware/src/base/flp.ts - 423, 426~~ #2857
-	~~adp-tooling/src/base/helper.ts - 16, 27, 79~~ #2931
-	~~adp-tooling/src/prompts/add-annotations-to-odata/index.ts - 75~~ #2931
-	~~ui5-config/src/ui5config.ts – 468, 611~~ #2931
-	~~fe-fpm-writer/src/common/validate.ts - 51~~ #2874
-	~~fe-fpm-writer/src/page/custom.ts - 107~~ #2874
-	~~telemetry/src/tooling-telemetry/data-processor.ts - 274, 275~~ #2931

### Value
Consistent support for webapp folder named different than `'webapp'`.

### Architecture Elaboration
Does it requires architecture elaboration?

### Notes
@Klaus-Keller 

### Tasks
- [x] search for `'webapp'` and replace if necessary

Vladislavs-Silins-SAP:
@heimwege, @Klaus-Keller  
In addition to direct 'webapp' string constants mentioned in the original list, there are also occurrences of `DirName.Webapp` enum usage that need attention. This enum is defined in `packages/project-access/src/constants.ts` with the value 'webapp'. Here are the additional occurrences that should be considered:

- `packages/adp-tooling/src/base/change-utils.ts` - lines 43, 99, 225
- `packages/launch-config/src/project-discovery/project.ts` - line 63
- `packages/project-access/src/project/ui5-config.ts` - line 16

These instances should also be reviewed to ensure consistent usage of the `getWebappPath` method.

heimwege:
Fixed in the linked PRs

PR #2904, opened 2025-02-11
feat: Service update support

broksy:
Currently only `generate` and `remove` functions are supported.
Adds additional `update` function to `odata-service-writer` to support service updates.

Features:
- Added update function for service related data updates (manifest.json dataSources, annotation files, changes in YAML files),
- In case of service update, package.json will not be updated,
- In case of service update, annotations files will be removed and rewritten (remote annotations could have changed for the service).
- Mockserver configuration during update will only be added if ui5-mock.yaml already exists,
- Missing unit test checks for remove to check if remote annotation files are removed.


packages/odata-service-writer/src/index.ts
```diff
@@ -269,6 +272,35 @@ async function generate(basePath: string, service: OdataService, fs?: Editor): P
...
+ * @param {Editor} [fs] - the memfs editor instance
+ * @throws {Error} - if required UI5 project files are not found
+ * @returns {Promise<Editor>} the updated memfs editor instance
+ */
+async function update(basePath: string, service: OdataService, fs?: Editor): Promise<Editor> {
+    if (!fs) {
+        fs = create(createStorage());
+    }
+    const paths = await findProjectFiles(basePath, fs);
+    ensureExists(basePath, ['webapp/manifest.json'], fs);
```
johannes-kolbe, line 289: please no hardcoded webapp path

packages/odata-service-writer/src/updates.ts
```diff
@@ -274,9 +280,18 @@ export async function updateManifest(basePath: string, service: OdataService, fs
             t('error.requiredProjectPropertyNotFound', { property: `'${appProp}'.id`, path: manifestPath })
         );
     }
-    // Check and update existing services
+    // Throw if required property is not found manifest.json in case of update and dataSource is not defined
+    if (updateService && service.name && !manifest[appProp]?.dataSources?.[service.name]) {
```
devinea, line 284: Service manager supports Adding a service to a freestyle app that was created with no datasource and adding additional services. Won't this check block these flows as I assume Service manager always calls update rather than generate.

broksy: @devinea ,
`updateService` parameter makes sure it's for the update only.
Currently Service Manager calls generate for both - update and create which later creates some inconsistencies - like `package.json` is updated unnecessarily etc.
After this, Service Manager will call `update` for service update, `generate` for services creation.

packages/odata-service-writer/src/index.ts
broksy: @johannes-kolbe ,
`ensureExists` is pretty specific function that checks an array of relative paths against `basePath` and throws an error.
If I run `ensureExists(basePath, ['webapp/manifest.json'], fs)` it throws "Error: Invalid project structure. Cannot find required file webapp/manifest.json".
If I run `ensureExists(webappPath, ['manifest.json'], fs)` it throws "Error: Invalid project structure. Cannot find required file manifest.json".

I think second error is a bit better just because path is clearer. What do you think?

johannes-kolbe: @broksy is it possible to consider https://github.com/SAP/open-ux-tools/issues/2589 in that context?

broksy: I can change `ensureExists(basePath, ['webapp/manifest.json'], fs)` call to `ensureExists(webappPath, ['manifest.json'], fs)`.

johannes-kolbe: It is also used below in [`updateManifest`](https://github.com/SAP/open-ux-tools/pull/2904/files#diff-11060fa3e2b31c89f7a31e60dfe2c764b0f9b32db732156920d506581236a4c4R300)

johannes-kolbe (dismissed):
- Code changes and refactorings lg
- Unit tests provided
- Fokus on new `update` function and tools-suite consumption
- Changeset provided

devinea (approved):
Code changes look good. 
Snapshot updates are only json ordering so ok. 
changeset ✅ 
Checked with service modeler and working well for legacy projects using UI5 mockserver in browser.

johannes-kolbe (approved):
Reapproving

PetrStribrny:
ux-tools-mockserver-config-writer reviwers asked for review

packages/odata-service-writer/src/data/annotations.ts
```diff
@@ -1,6 +1,265 @@
...
+ * @param {CdsAnnotationsInfo} annotations - The cds annotations info.
+ * @param {Editor} fs - The memfs editor instance
+ * @returns {Promise<void>} A promise that resolves when the cds files have been updated.
+ */
+export async function updateCdsFilesWithAnnotations(
+    annotations: CdsAnnotationsInfo | CdsAnnotationsInfo[],
+    fs: Editor
+): Promise<void> {
+    if (Array.isArray(annotations)) {
+        for (const annotationName in annotations) {
```
unseen1980, line 48: This iterates over the keys, why not ```for (const annotation of annotations)``` so that you work with the objects?

packages/odata-service-writer/src/data/annotations.ts
```diff
@@ -1,6 +1,265 @@
...
+ */
+async function removeCdsIndexOrServiceFile(fs: Editor, annotations: CdsAnnotationsInfo): Promise<void> {
+    const dirPath = join(annotations.projectName, 'annotations');
+    const annotationPath = normalize(dirPath).split(/[\\/]/g).join(posix.sep);
+    const annotationConfig = `\nusing from './${annotationPath}';`;
+    // Get index and service file paths
+    const indexFilePath = join(annotations.projectPath, annotations.appPath ?? '', 'index.cds');
+    const serviceFilePath = join(annotations.projectPath, annotations.appPath ?? '', 'services.cds');
+    // Remove annotation config from index or service file
+    if (indexFilePath && fs.exists(indexFilePath)) {
```
unseen1980, line 94: This is probably enough?
```fs.exists(indexFilePath)```

File: packages/odata-service-writer/src/data/annotations.ts [hunk 5]
```diff
@@ -1,6 +1,265 @@
...
+                'webapp',
+                'localService',
+                serviceName,
+                `${annotation.technicalName}.xml`
+            );
+            if (fs.exists(pathToAnnotationFile)) {
+                fs.delete(pathToAnnotationFile);
+            }
+        }
+    } else if (edmxAnnotations?.xml) {
```
unseen1980, line 217: I think you can remove the ```?```. You already checked in line 203

packages/odata-service-writer/src/data/manifest.ts
```diff
@@ -0,0 +1,528 @@
...
+ *
+ * @param basePath - the root path of an existing UI5 application
+ * @param service - name of the OData service instance
+ * @param fs - the memfs editor instance
+ */
+export function deleteServiceFromManifest(basePath: string, service: OdataService, fs: Editor): void {
+    const serviceName: string = service.name ?? 'mainService';
+    const manifestPath = join(basePath, 'webapp', 'manifest.json');
+    // Get component app id
+    const manifest = fs.readJSON(manifestPath) as unknown as Manifest;
```
unseen1980, line 450: the readJson might throw an error if manifestPath for some reason does not exist. Maybe add a try ctach or something like this?
```
if (fs.exists(manifestPath)) {
    const manifest = fs.readJSON(manifestPath) as unknown as Manifest;
} else {
    throw new Error(`Manifest file not found at ${manifestPath}`);
}```

**Review by unseen1980 — APPROVED** [review]
Some small comments that can be fixed in a following PR. Thanks

File: packages/odata-service-writer/src/data/manifest.ts [hunk 7]
```diff
@@ -0,0 +1,528 @@
...
+): void {
+    const settings = dataSource.settings;
+    if (settings) {
+        // "localService/metadata.xml"
+        const localUri = settings.localUri;
+        // -> ["localService", "metadata.xml"]
+        const localUriParts = localUri ? localUri.split('/') : undefined;
+        if (localUriParts && localUriParts[0] === DirName.LocalService && localUriParts.length === 2) {
+            const localFileName = localUriParts[localUriParts.length - 1];
+            settings.localUri = `${DirName.LocalService}/${dataSourceKey}/${localFileName}`;
```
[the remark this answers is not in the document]
devinea: this one always uses `/`

packages/odata-service-writer/src/data/annotations.ts
PetrStribrny: As agreed a new refactoring PR will be created so we can proceed with this one.

broksy:
> Some small comments that can be fixed in a following PR. Thanks

I will create a separate pull requests with those change requests.

PR #2908, opened 2025-02-12
fix: undo redo issue for v2 manifest changes

Jimmy-Joseph19:
When v2 manifest changes created via quick actions are undone(by undo) they still get written to file system on save(rta save).

### DIFF
--- a/packages/preview-middleware-client/src/adp/quick-actions/fe-v2/utils.ts
+++ b/packages/preview-middleware-client/src/adp/quick-actions/fe-v2/utils.ts
@@ -5,6 +5,24 @@ import CommandFactory from 'sap/ui/rta/command/CommandFactory';
 import { QuickActionContext } from '../../../cpe/quick-actions/quick-action-definition';
 import { getUi5Version, isLowerThanMinimalUi5Version, isVersionEqualOrHasNewerPatch } from '../../../utils/version';
 import { Manifest } from 'sap/ui/rta/RuntimeAuthoring';
+import Component from 'sap/ui/core/Component';
+import AppComponent from 'sap/suite/ui/generic/template/lib/AppComponent';
+import ManagedObject from 'sap/ui/base/ManagedObject';
+import TemplateComponent from 'sap/suite/ui/generic/template/lib/TemplateComponent';
+
+/**
+ * Gets app component of a v2 project.
+ *
+ * @param control - ManagedObject.
+ * @returns AppComponent.
+ */
+export function getAppComponent(control: ManagedObject): AppComponent | undefined {
+    const ownerComponent = Component.getOwnerComponentFor(control);
+    if (ownerComponent?.isA<TemplateComponent>('sap/suite/ui/generic/template/lib/TemplateComponent')) {
+        return ownerComponent.getAppComponent();
+    }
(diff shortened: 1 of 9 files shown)


vadson71 (dismissed):
Code changes look good
Covered by test
Tested locally

packages/preview-middleware-client/src/adp/quick-actions/fe-v2/utils.ts
```diff
@@ -5,6 +5,25 @@ import CommandFactory from 'sap/ui/rta/command/CommandFactory';
 import { QuickActionContext } from '../../../cpe/quick-actions/quick-action-definition';
 import { getUi5Version, isLowerThanMinimalUi5Version, isVersionEqualOrHasNewerPatch } from '../../../utils/version';
 import { Manifest } from 'sap/ui/rta/RuntimeAuthoring';
+import Component from 'sap/ui/core/Component';
+import AppComponent from 'sap/suite/ui/generic/template/lib/AppComponent';
+import ManagedObject from 'sap/ui/base/ManagedObject';
+import TemplateComponent from 'sap/suite/ui/generic/template/lib/TemplateComponent';
```
nikmace, line 11: These three are only used as types

nikmace (changes requested):
Overall the changes look good

packages/preview-middleware-client/src/adp/quick-actions/fe-v2/utils.ts
```diff
@@ -5,6 +5,25 @@ import CommandFactory from 'sap/ui/rta/command/CommandFactory';
...
+
+/**
+ * Gets app component of a v2 project.
+ *
+ * @param control - ManagedObject.
+ * @returns AppComponent.
+ */
+export function getAppComponent(control: ManagedObject): AppComponent | undefined {
```
mmilko01, line 19: I saw that a very similar function already exists https://github.com/SAP/open-ux-tools/blob/8883d9a67a52a5e3b4a0b8467bf68eef92a4d545/packages/preview-middleware-client/src/utils/fe-v4.ts#L19-L25 . Does it make sense to reuse it in both cases?

Jimmy-Joseph19: types are different for v2 and v4 for AppComponent/Template Component
v4 is from sap/fe/core and v2 is sap/ui/generic

mmilko01: Yes, I noticed, but still this might a parameter so we could reuse the rest of the logic which is identical. Only a suggestion, if this is difficult it can be left like this.

mmilko01: After discussion, using the same function for both scenarios makes the implementation complex, it is sufficient to rename the functions.

mmilko01 (approved):
Changeset OK
Comments addressed
Good test coverage
Did not test manually

nikmace (approved):
Changes look good
Coverage is good
Did not test manually

vadson71 (approved):
Code changes look good.
Comments addressed

Issue #2511, opened 2024-10-30
FEATURE - Info Center for different type of messages


IvoSG:
As  a developerwhen use the Adaptation EditorI want to be able to see all relevant information, warnings and errors in a specific section in the bottom right corner of the editor so that I can have every message relevant for me consolidated in one place 

### Description

`@sap-ux/control-property-editor`  The whole InfoCenter box should be placed in the right panel, below the properties` box, snapped to the bottom and and if needed, may take up to 50 %  from the bottom of the page- i.e. the box should 'rise' with every new message inside of it up to these 50 %. If the messages are too many to fit in the available space, a scrollbar should appear and the developer should be able to scroll them down or up.

The InfoCenter should have draggable divider on its top that would be able to change its size. 

The title of the whole box should be "Info Center" (in the example it says "messages). This box with its title should become visible only when there is a message inside of it and not visible otherwise.

Newer messages should show on the top of the messages list.

Three types of messages are needed:

Information - blue info icon - Closable with an X button in the top right corner of the respective message box

Warning - yellow warning icon - Closable

Error - red error icon - Non closable

There should be a Close All button on the top right corner of the Info Center box that should close all infos but not the ones that are errors.

A 'message' should have a title and a description, shown with different font faces and be separated from each other with lines, as shown in the example image

![image](https://github.com/user-attachments/assets/07e38aac-a77e-4e22-a794-e894f755e2b7)

### Technical Design
_What modules need to be created or enhanced?_ 
_How would the functionality be exposed and how can it be consumed?_

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a ...
**when** I execute ...
**then** ... will happen.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #2909, opened 2025-02-12
feat: Info Center for different type of messages

IvoSG:
Feat for https://github.com/SAP/open-ux-tools/issues/2511

- add the InfoCenter message component to the RigthPanel.
- Enhances in UIMessageBar component
- Introduce new icon for clearAll
- Log message in info center from preview-middleware-client in adaptation-project


packages/control-property-editor/src/panels/info-center/InfoCenter.tsx
```diff
@@ -0,0 +1,115 @@
...
+                                    isExpandable &&
+                                    <Text className='more-less' onClick={() => dispatch(toggleExpandMessage(index))}>
+                                        {isExpanded ? 'Less' : 'More'}
+                                    </Text>
+                                }
+                                {
+                                    message.details &&
+                                    <Text className='message-details' onClick={() => dispatch(toggleModalMessage(index))}>
+                                        {'View Details'}
+                                    </Text>
```
nikmace, line 96: This doesn't need to be in curly braces and quotes

packages/control-property-editor/src/panels/info-center/InfoCenter.tsx
```diff
@@ -0,0 +1,115 @@
...
+                            iconProps={{ iconName: UiIcons.TextGrammarDismiss }}
+                        />
+                    </Stack.Item>
+                </div>
+                <Stack className='info-center-items'>
+                    {messages.map((info, index) => {
+                        const { expandable: isExpandable, expanded: isExpanded, read: isRead, modal: isOpenedModal, message, type } = info;
+                        return (
+                            <Stack.Item key={index} className={`message-bar ${getMessageType(type)} ${isRead && 'message-read'}`} onMouseOver={() => dispatch(readMessage(index))}>
+                                <UIMessageBar messageBarType={type as MessageBarType}>
```
nikmace, line 72: A couple of things here:

1. The `map` function inside JSX is large and could be extracted into a separate component for better readability, for example `InfoMessageItem.tsx`.

```tsx
{messages.map((info, index) => (
    <InfoMessageItem key={info.id} info={info} index={index} />
))}
```

```tsx
const InfoMessageItem: React.FC<InfoMessageProps> = ({ info, index }) => {
    const dispatch = useDispatch();
    const { expandable, expanded, read, modal, message, type } = info;
    
    ...etc
```

2. Using an `index` as a `key` can cause issues when items are added or removed dynamically. You should avoid using index as key in map functions as per official documentation: https://legacy.reactjs.org/docs/lists-and-keys.html#keys. Instead, it would be nice to generate a unique id for each message.

packages/control-property-editor/src/panels/info-center/InfoCenter.tsx
```diff
@@ -0,0 +1,115 @@
...
+    function getMessageType(type: MessageBarType) {
+        switch (type) {
+            case MessageBarType.error:
+                return 'error';
+            case MessageBarType.warning:
+                return 'warning';
+            default:
+                return 'info';
+        }
+    }
```
nikmace, line 32: It would be good to extract this function outside of the component, you will see below my suggestion for the new component. It will avoid re-creation on every render.

packages/control-property-editor/src/slice.ts
```diff
@@ -446,6 +457,48 @@ const slice = createSlice<SliceState, SliceCaseReducers<SliceState>, string>({
...
             .addMatcher(
+                showInfoCenterMessage.match,
+                (state: SliceState, action: ReturnType<typeof showInfoCenterMessage>): void => {
+                    state.infoCenterMessages.unshift(action.payload);
+                }
+            )
+            .addMatcher(
+                clearInfoCenterMessage.match,
+                (state: SliceState, action: ReturnType<typeof clearInfoCenterMessage>): void => {
+                    state.infoCenterMessages = state.infoCenterMessages.filter((_, index) => index !== action.payload);
```
nikmace, line 468: All of these need to work with unique ids a not to rely on the order of elements in the array.

packages/control-property-editor-common/src/api.ts
```diff
@@ -283,6 +283,28 @@ export interface NestedQuickActionExecutionPayload {
...
+}
+
+export enum MessageBarType {
+    /** Info styled MessageBar */
+    info = 0,
+    /** Error styled MessageBar */
+    error = 1,
+    /** Warning styled MessageBar */
+    warning = 5
+}
```
nikmace, line 306: Can we reuse `MessageBarType` from `@fluentui/react`? Because it is essentially the same thing.

```tsx
import { MessageBarType } from '@fluentui/react';
```

nikmace (changes requested):

packages/adp-tooling/package.json
```diff
@@ -43,6 +43,7 @@
         "@sap-ux/project-input-validator": "workspace:*",
         "@sap-ux/system-access": "workspace:*",
         "@sap-ux/ui5-config": "workspace:*",
+        "@sap-ux-private/control-property-editor-common": "workspace:*",
```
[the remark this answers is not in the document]
IvoSG: Done

packages/control-property-editor/src/panels/info-center/InfoCenter.scss
```diff
@@ -0,0 +1,131 @@
...
+        border-left: 2px solid var(--vscode-editorError-foreground);
+
+        .message-title {
+            color: var(--vscode-editorError-foreground);
+        }
+    }
+    .ms-MessageBar-icon, .ms-MessageBar-text {
+        margin: 0;
+    }
+    :first-child {
```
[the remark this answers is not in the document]
IvoSG: Should be like it is

packages/control-property-editor/src/panels/info-center/InfoCenter.scss
```diff
@@ -0,0 +1,131 @@
...
+    &:hover {
+        text-decoration: unset;
+        cursor: pointer;
+    }
+}
+
+.ms-Dialog {
+    .ms-Dialog-main {
+        width: 700px;
+        max-width: 700px;
```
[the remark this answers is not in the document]
IvoSG: Done

packages/control-property-editor-common/src/api.ts
IvoSG: The native implementation in **@fluentui/react** includes additional types that are irrelevant to this scenario.

packages/control-property-editor/src/panels/info-center/InfoCenter.tsx
```diff
@@ -0,0 +1,115 @@
...
+        document.querySelectorAll('.message-description').forEach((element) => {
+            const { scrollHeight, clientHeight, accessKey } = element as HTMLElement;
+            if (scrollHeight > clientHeight) {
+                dispatch(expandableMessage(Number(accessKey)));
+            }
+        });
+    });
+
+    return (
+        <>
```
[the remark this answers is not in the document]
IvoSG: Done

packages/control-property-editor-common/src/api.ts
```diff
@@ -283,6 +283,28 @@ export interface NestedQuickActionExecutionPayload {
...
+    type: MessageBarType;
+    message: {
+        title: string;
+        description: string;
+        details?: string;
+    };
+    expandable?: boolean;
+    expanded?: boolean;
+    read?: boolean;
+    modal?: boolean;
```
voicis, line 296: We only need to define the data types that are required for communication between the client and the editor. All the UI state should be defined in `control-property-editor`.

packages/control-property-editor-common/src/api.ts
```diff
@@ -464,7 +486,14 @@ export const executeQuickAction = createExternalAction<QuickActionExecutionPaylo
...
     createExternalAction<ContextMenuActionExecutionPayload>('execute-context-menu-action');
 export const setApplicationRequiresReload = createExternalAction<boolean>('set-application-requires-reload');
+export const showInfoCenterMessage = createExternalAction<InfoCenterMessage>('show-info-center-message');
+export const clearInfoCenterMessage = createExternalAction<number>('clear-info-center-message');
+export const clearAllInfoCenterMessages = createExternalAction<void>('clear-all-info-center-message');
 export const externalFileChange = createExternalAction<string>('external-file-change');
+export const toggleExpandMessage = createExternalAction<number>('toggle-expand-message');
+export const readMessage = createExternalAction<number>('read-message');
+export const expandableMessage = createExternalAction<number>('expandable-message');
+export const toggleModalMessage = createExternalAction<number>('toggle-modal-message');
```
voicis, line 496: All of these look like UI actions, so they should also be only defined in the `control-property-editor`.

packages/preview-middleware-client/src/cpe/outline/service.ts
```diff
@@ -61,9 +66,12 @@ export class OutlineService extends EventTarget {
                 sendAction(outlineChanged(outlineNodes));
                 if (reuseComponentsIds.size > 0 && scenario === SCENARIO.AdaptationProject && !hasSentWarning) {
                     sendAction(
-                        showMessage({
-                            message,
-                            shouldHideIframe: false
+                        showInfoCenterMessage({
+                            message: {
+                                title: 'Reuse components detected',
```
voicis, line 71: Title should also use texts from message bundle.

packages/ui-components/src/components/UISection/UISplitter.scss
```diff
@@ -19,6 +19,7 @@ $splitter-focus-border: 1px solid var(--vscode-focusBorder);
     background-color: var(--vscode-panelSection-border);
     position: absolute;
     outline: none;
+    z-index: 2;
```
voicis, line 22: Why was this needed? It seems that info center panel overlaps a bit the splitter, but that seems like an issue with info center panels styling not the splitter, because other components are not affected.

packages/control-property-editor/src/panels/info-center/InfoCenter.tsx
```diff
@@ -0,0 +1,145 @@
...
+                                accessKey={index.toString()}
+                                block={true}
+                                className={`message-description ${isExpanded && 'expanded'} ${
+                                    isExpandable && 'expandable'
+                                }`}>
+                                {message.description}
+                            </Text>
+                            {isExpandable && (
+                                <Text className="more-less" onClick={() => dispatch(toggleExpandMessage(index))}>
+                                    {isExpanded ? 'Less' : 'More'}
```
voicis, line 117: All the UI texts should use i18n.

packages/control-property-editor/src/panels/info-center/InfoCenter.tsx
```diff
@@ -0,0 +1,145 @@
...
+                            )}
+                            {message.details && (
+                                <Text className="message-details" onClick={() => dispatch(toggleModalMessage(index))}>
+                                    {'View Details'}
+                                </Text>
+                            )}
+                            <UIDialog
+                                hidden={!isOpenedModal}
+                                dialogContentProps={{
+                                    title: 'Error Details'
```
voicis, line 128: Same as above.

packages/control-property-editor/src/panels/info-center/InfoCenter.tsx
```diff
@@ -0,0 +1,115 @@
...
+                <div className='info-center-header'>
+                    <Stack.Item>
+                        <Label
+                            data-aria-label={t('INFO CENTER')}
+                            style={{
+                                color: 'var(--vscode-foreground)',
+                                fontSize: sectionHeaderFontSize,
+                                fontWeight: 'bold',
+                                padding: 0
+                            }}>
```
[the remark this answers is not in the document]
IvoSG: Done

packages/control-property-editor/src/panels/info-center/InfoCenter.tsx
IvoSG: Changed to text from message bundle

packages/control-property-editor/src/panels/info-center/InfoCenter.tsx
IvoSG: Used unique id's instead of index's

packages/control-property-editor/src/panels/info-center/InfoCenter.tsx
IvoSG: Done

packages/control-property-editor-common/src/api.ts
IvoSG: Moved

packages/control-property-editor-common/src/api.ts
IvoSG: Done

packages/ui-components/src/components/UISection/UISplitter.scss
IvoSG: Reverted and fixed in the info center styles

packages/preview-middleware-client/src/cpe/outline/service.ts
IvoSG: Done for all texts

packages/control-property-editor/src/panels/info-center/InfoCenter.tsx
IvoSG: Done

packages/control-property-editor/src/panels/info-center/InfoCenter.tsx
IvoSG: Done

815are (approved):
checked `ui-components`:
1. code changes are easy to understand - I did not notice issue
2. tests are added for additional message types

packages/preview-middleware-client/src/messagebundle.properties
```diff
@@ -35,7 +35,7 @@ ADP_ADD_FRAGMENT_NOTIFICATION = Note: The "{0}.fragment.xml" fragment will be cr
 ADP_ADD_FRAGMENT_WITH_TEMPLATE_NOTIFICATION = Note: The "{0}.fragment.xml" fragment will be created once you save the changes.
 ADP_ADD_TWO_FRAGMENTS_WITH_TEMPLATE_NOTIFICATION = Note: The "{0}.fragment.xml" and "{1}.fragment.xml" fragments will be created once you save the changes.
 ADP_SYNC_VIEWS_MESSAGE = Synchronous views are detected for this application. Controller extensions are not supported for such views and will be disabled.
-ADP_REUSE_COMPONENTS_MESSAGE = Reuse components are detected for some views in this application. Controller extensions, adding fragments and manifest changes are not supported for such views and will be disabled.
+ADP_REUSE_COMPONENTS_MESSAGE = Reuse components detected | Reuse components are detected for some views in this application. Controller extensions, adding fragments and manifest changes are not supported for such views and will be disabled.
```
voicis, line 38: This is rather confusing, looking at this text it is not clear that it is actually used for two different things.

Please add a new entry for the title instead of combining two and then splitting.

packages/preview-middleware-client/src/cpe/outline/service.ts
```diff
@@ -61,9 +67,10 @@ export class OutlineService extends EventTarget {
                 sendAction(outlineChanged(outlineNodes));
                 if (reuseComponentsIds.size > 0 && scenario === SCENARIO.AdaptationProject && !hasSentWarning) {
                     sendAction(
-                        showMessage({
-                            message,
-                            shouldHideIframe: false
+                        showInfoCenterMessage({
+                            type: MessageBarType.warning,
+                            title: title.trim(),
```
voicis, line 72: We shouldn't need to perform any operations on the texts. Changes should be done in .properties files.

packages/preview-middleware-client/src/cpe/outline/service.ts
```diff
@@ -35,6 +40,7 @@ export class OutlineService extends EventTarget {
         const resourceBundle = await getTextBundle();
         const key = 'ADP_REUSE_COMPONENTS_MESSAGE';
         const message = resourceBundle.getText(key) ?? key;
```
voicis, line 42: This probably can be converted into one liner passing the key without variable, `getText` should already return the value of `key` if the text is not found.

packages/control-property-editor/src/panels/info-center/InfoCenter.tsx
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * This effect selects all elements with the class "message-description" (which represent message descriptions)
+     * and checks if their content is clamped (i.e. truncated) by comparing each element's scrollHeight with its clientHeight.
+     * If an element's content is clamped and it is not already marked as expandable, the effect dispatches an action
+     * to mark the corresponding message as expandable.
+     *
+     * @returns {void} This effect does not return a value.
+     */
+    useEffect(() => {
+        document.querySelectorAll('.message-description').forEach((element) => {
```
voicis, line 32: Please add a comment why this non standard way of interacting with DOM is needed and typical approach using refs did not work. In react 19 you no longer need to use `forwardRef`, so perhaps we can remove this workaround then. https://react.dev/reference/react/forwardRef

nikmace (changes requested):

packages/control-property-editor/src/panels/info-center/InfoMessageItem.tsx
```diff
@@ -0,0 +1,83 @@
...
+ * React element for info center item.
+ *
+ * @param item
+ * @returns ReactElement
+ */
+export function InfoMessageItem(item: Readonly<InfoCenterItem>): ReactElement {
+    const { t } = useTranslation();
+    const dispatch = useDispatch();
+    const { expandable: isExpandable, expanded: isExpanded, read: isRead, modal: isOpenedModal, message, id } = item;
+
```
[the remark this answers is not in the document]
voicis: @nikmace unfortunately it looks like getting `ref` for the `Text` component does not work. When using a `span` element instead it does work, so there seems to be some kind of limitation with this particular component.

As for checking if it is expandable or not, I don't see how we can detect it before it gets rendered, as this is done via CSS.

packages/control-property-editor/src/slice.ts
```diff
@@ -445,6 +466,61 @@ const slice = createSlice<SliceState, SliceCaseReducers<SliceState>, string>({
...
+                clearInfoCenterMessage.match,
+                (state: SliceState, action: ReturnType<typeof clearInfoCenterMessage>): void => {
+                    state.infoCenterMessages = state.infoCenterMessages.filter(
+                        (message) => message.id !== action.payload
+                    );
+                }
+            )
+            .addMatcher(clearAllInfoCenterMessages.match, (state: SliceState): void => {
+                state.infoCenterMessages = state.infoCenterMessages.filter(
+                    (info) => info.message.type === MessageBarType.error
```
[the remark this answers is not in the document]
IvoSG: It's a feature requirement described in [2511](https://github.com/SAP/open-ux-tools/issues/2511).
**There should be a Close All button on the top right corner of the Info Center box that should close all infos but not the ones that are errors.**

packages/control-property-editor/src/panels/info-center/InfoCenter.scss
```diff
@@ -0,0 +1,143 @@
...
+
+.message-title {
+    font-size: 13px;
+    font-weight: bold;
+    color: var(--vscode-editor-foreground);
+    padding-left: 5px;
+}
+
+.message-description {
+    display: -webkit-box;
```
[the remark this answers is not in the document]
IvoSG: Added comment description

packages/control-property-editor/src/panels/info-center/InfoCenter.scss
```diff
@@ -0,0 +1,143 @@
...
+.message-bar:hover .remove-message {
+    display: block;
+    background-color: var(--vscode-activityBar-background);
+}
+
+.remove-message:hover {
+    background-color: var(--vscode-actionBar-toggledBackground) !important;
+}
+
+.message-title {
```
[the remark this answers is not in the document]
IvoSG: Divided into separate files

packages/preview-middleware-client/src/messagebundle.properties
IvoSG: A new entry for the title was added

packages/preview-middleware-client/src/cpe/outline/service.ts
IvoSG: Removed

packages/preview-middleware-client/src/cpe/outline/service.ts
IvoSG: Done

packages/control-property-editor/src/panels/info-center/InfoCenter.tsx
IvoSG: A comprehensive explanation for the current implementation has been added.

packages/control-property-editor/src/panels/info-center/InfoMessageItem.tsx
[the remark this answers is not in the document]
IvoSG: A comprehensive explanation for the current implementation has been added.

packages/control-property-editor/src/panels/info-center/InfoMessageItem.tsx
```diff
@@ -0,0 +1,83 @@
...
+                data-index={id}
+                block={true}
+                className={`message-description ${isExpanded && 'expanded'} ${isExpandable && 'expandable'}`}>
+                {message.description}
+            </Text>
+            {isExpandable && (
+                <Text className="more-less" onClick={() => dispatch(toggleExpandMessage(id))}>
+                    {t(isExpanded ? 'LESS' : 'MORE')}
+                </Text>
+            )}
```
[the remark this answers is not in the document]
IvoSG: Changed to using state instead of redux in the component

.changeset/yellow-hairs-peel.md
```diff
@@ -0,0 +1,8 @@
+---
+'@sap-ux/ui-components': major
+'@sap-ux-private/control-property-editor-common': minor
+'@sap-ux-private/preview-middleware-client': major
+'@sap-ux/control-property-editor': minor
+---
+
+feat: Info Center for different type of messages
```
nikmace, line 8: I am not sure if the `@sap-ux-private/preview-middleware-client` & `@sap-ux/ui-components` should be major. They both have little changes and could be patch instead as well as `@sap-ux-private/control-property-editor-common`. What do you think @voicis?

packages/control-property-editor/src/panels/info-center/InfoMessageItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * @param item
+ * @returns ReactElement
+ */
+export function InfoMessageItem(item: Readonly<InfoCenterItem>): ReactElement {
+    const { t } = useTranslation();
+    const dispatch = useDispatch();
+    const [isExpanded, setIsExpanded] = useState<boolean>();
+    const [isRead, setIsRead] = useState<boolean>();
+    const [isOpenedModal, setIsOpenedModal] = useState<boolean>();
```
nikmace, line 40: To avoid unintended behavior these need to have default values which in our case are false.

```suggestion
    const [isExpanded, setIsExpanded] = useState<boolean>(false);
    const [isRead, setIsRead] = useState<boolean>(false);
    const [isOpenedModal, setIsOpenedModal] = useState<boolean>(false);
```

nikmace (changes requested):

.changeset/yellow-hairs-peel.md
IvoSG: In` @sap-ux-private/preview-middleware-client`, the change is marked as a major because it consumes **cpe** however, I might be mistaken. Regarding `@sap-ux/ui-components`, it's also considered a major update since I've introduced a new icon.  
@voicis could you please confirm the correctness of this approach?

voicis: Since the majority of the packages are with 0.x.x versions, we should not use the major version without a good reason. For breaking or bigger changes we use minor version, for other changes patch. So in this case I'd use patch version for `@sap-ux/ui-components` and minor for other ones here. Also it is missing patch for `@sap-ux/preview-middleware` here, since that is the package through which `@sap-ux-private/preview-middleware-client` changes gets published.

https://semver.org/

IvoSG: Updated

voicis (approved):
- Code changes looks good
- Review comments addressed
- Good test coverage
- Tested locally

Note: there is an issue impacting this change #2971

815are (approved):

lfindlaysap (approved):
Looks good to me, @IvoSG.

PR #2910, opened 2025-02-12
feat(project-access): export `getI18nBundles` function

815are:
export [getI18nBundles](https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/project/i18n/read.ts#L31) function


vadson71 (approved):
Straightforward changes

devinea (approved):

Issue #2916, opened 2025-02-12
BUG - FLP config not correctly spelt


longieirl:
### Description
The FLP is using an incorrect spelling for the prompting name.

### Expected results

Correct spelling when defined prompts.

### Actual results

Incorrect promt name defined.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Fix and on the latest SAP Fiori tools release.

PR #2917, opened 2025-02-12
fix(flp): fix spelling prompt name

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/2916

### DIFF
--- a/packages/deploy-config-generator-shared/src/prompts/index.ts
+++ b/packages/deploy-config-generator-shared/src/prompts/index.ts
@@ -5,7 +5,7 @@ import type { Question } from 'inquirer';
  * Enumeration of prompt names
  */
 enum promptNames {
-    confirmConfigUpate = 'confirmConfigUpate'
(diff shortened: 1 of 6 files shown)


kjose90 (approved):
Patch to fix spelling mistake  ✅

packages/deploy-config-generator-shared/src/prompts/index.ts
```diff
@@ -18,8 +18,8 @@ export function getConfirmConfigUpdatePrompt(configType?: string): Question[] {
     return [
```
IainSAP, line 18: Im not sure these prompts should be here as they are extension generator specific?

IainSAP (approved):
- Typo fixed

packages/deploy-config-generator-shared/src/prompts/index.ts
longieirl: Going to resolve conversation and update internal ticket with comments.

PR #2922, opened 2025-02-14
fix(app-config-writer): wrong prerequisites check in convert preview-config

heimwege:
`cds-plugin-ui5` is currently being searched in `package.json` of the app to be converted but must be searched for in CAP root `package.json`.

### DIFF
--- a/.changeset/long-planets-sing.md
+++ b/.changeset/long-planets-sing.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/app-config-writer': patch
+---
+
+fix: wrong convert preview-config prerequisites check for usage of cds-plugin-ui5
--- a/packages/app-config-writer/src/preview-config/prerequisites.ts
+++ b/packages/app-config-writer/src/preview-config/prerequisites.ts
@@ -1,6 +1,6 @@
 import { join } from 'path';
 import type { Editor } from 'mem-fs-editor';
-import type { Package } from '@sap-ux/project-access';
+import { type Package, findCapProjectRoot } from '@sap-ux/project-access';
 import type { ToolsLogger } from '@sap-ux/logger';
 import { satisfies, valid } from 'semver';
 
@@ -85,14 +85,22 @@ export async function checkPrerequisites(
         prerequisitesMet = false;
     }
 
+    let cdsPluginUi5Exists = false;
+    const capProjectRootPath = await findCapProjectRoot(basePath, false, fs);
+    if (capProjectRootPath) {
+        const capRootPackageJsonPath = join(capProjectRootPath, 'package.json');
+        const capRootPackageJson = fs.readJSON(capRootPackageJsonPath) as Package | undefined;
+        cdsPluginUi5Exists =
+            !!capRootPackageJson?.devDependencies?.['cds-plugin-ui5'] ||
+            !!capRootPackageJson?.dependencies?.['cds-plugin-ui5'];
+    }
+
     const ui5MiddlewareMockserverExists =
         !!packageJson?.devDependencies?.['@sap-ux/ui5-middleware-fe-mockserver'] ||
         !!packageJson?.dependencies?.['@sap-ux/ui5-middleware-fe-mockserver'];
-    const cdsPluginUi5Exists =
-        !!packageJson?.devDependencies?.['cds-plugin-ui5'] || !!packageJson?.dependencies?.['cds-plugin-ui5'];
     if (!ui5MiddlewareMockserverExists && !cdsPluginUi5Exists) {
         logger?.error(
-            "Conversion from 'sap/ui/core/util/MockServer' is not supported. You must migrate from '@sap-ux/ui5-middleware-fe-mockserver'. For more information, see https://www.npmjs.com/package/@sap-ux/ui5-middleware-fe-mockserver."
+            "Conversion from 'sap/ui/core/util/MockServer' or '@sap/ux-ui5-fe-mockserver-middleware' is not supported. You must migrate from '@sap-ux/ui5-middleware-fe-mockserver'. For more information, see https://www.npmjs.com/package/@sap-ux/ui5-middleware-fe-mockserver."
         );
         prerequisitesMet = false;
     }
--- a/packages/app-config-writer/test/unit/preview-config/prerequisites.test.ts
+++ b/packages/app-config-writer/test/unit/preview-config/prerequisites.test.ts
@@ -3,12 +3,14 @@ import { create } from 'mem-fs-editor';
 import { create as createStorage } from 'mem-fs';
 import { join } from 'path';
 import { ToolsLogger } from '@sap-ux/logger';
+import * as ProjectAccess from '@sap-ux/project-access';
 
 describe('prerequisites', () => {
     const logger = new ToolsLogger();
     const errorLogMock = jest.spyOn(ToolsLogger.prototype, 'error').mockImplementation(() => {});
     const warnLogMock = jest.spyOn(ToolsLogger.prototype, 'warn').mockImplementation(() => {});
     const basePath = join(__dirname, '../../fixtures/preview-config');
+    jest.spyOn(ProjectAccess, 'findCapProjectRoot').mockImplementation(() => Promise.resolve(basePath));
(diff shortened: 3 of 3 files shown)


packages/app-config-writer/src/preview-config/prerequisites.ts
```diff
@@ -35,6 +45,26 @@ function isLowerThanMinimalVersion(
     return !satisfies(minVersionInfo, versionInfo);
 }
 
+/**
```
Klaus-Keller, line 48: Hi @heimwege, there is already a check for cds plugin in module `@sap-ux/cap-config-writer` `checkCdsUi5PluginEnabled()`, see:
https://github.com/SAP/open-ux-tools/blob/280c82284a4faea9e2947a782069d32ff201f059/packages/cap-config-writer/src/cap-config/index.ts#L92-L122

I don't think it is a good idea to have a dependency from `@sap-ux/app-config-writer` to `@sap-ux/cap-config-writer`, my proposal would be to move the function to `@sap-ux/project-access` and call it from there in all places.

heimwege: @Klaus-Keller done in https://github.com/SAP/open-ux-tools/pull/2922/commits/104f6942aad4af4e03f7663b9737554454477fc9. Please check carefully as it was a lot of methods and tests to be adjusted

packages/project-access/src/index.ts
```diff
@@ -47,7 +47,13 @@ export {
     readUi5Yaml,
     refreshSpecificationDistTags,
     toReferenceUri,
-    updatePackageScript
+    updatePackageScript,
+    hasCdsPluginUi5,
```
Klaus-Keller, line 51: Do we need to export this? Is it not also part of `checkCdsUi5PluginEnabled`?

packages/project-access/src/index.ts
```diff
@@ -47,7 +47,13 @@ export {
...
+    getWorkspaceInfo,
+    getWorkspacePackages,
```
Klaus-Keller, line 53: Can we combine `getWorkspaceInfo` and `getWorkspacePackages` or add them to results to `checkCdsUi5PluginEnabled`?

packages/project-access/src/index.ts
```diff
@@ -47,7 +47,13 @@ export {
...
+    minCdsVersion,
```
Klaus-Keller, line 54: Perhaps we add this to `constants.ts` and change it to Uppercase? Also a more speakable name might help, like `MinCdsVersionUi5Plugin`?

packages/project-access/src/index.ts
```diff
@@ -47,7 +47,13 @@ export {
...
+    hasMinCdsVersion,
```
Klaus-Keller, line 55: Do we need a separate export for this, info is also included in `checkCdsUi5PluginEnabled `

packages/project-access/src/project/cap.ts
```diff
@@ -821,3 +827,151 @@ export async function deleteCapApp(appPath: string, memFs?: Editor, logger?: Log
...
+}
+
+/**
+ * Check if devDependency to cds-plugin-ui5 is present in package.json.
+ *
+ * @param packageJson - the parsed package.json
+ * @returns true: devDependency to cds-plugin-ui5 exists; false: devDependency to cds-plugin-ui5 does not exist
+ */
+export function hasCdsPluginUi5(packageJson: Package): boolean {
+    return !!packageJson.devDependencies?.['cds-plugin-ui5'];
```
Klaus-Keller, line 955: We have already
https://github.com/SAP/open-ux-tools/blob/4859d13f3bf9a2bc71fb8d66523f73bb7f5dd874/packages/project-access/src/project/dependencies.ts#L15-L16
which checks for both: `dependencies` and `devDependencies` which I think is even better, if someone decides to have `cds-plgin-ui5` as dependency we still should enable tooling for it

packages/project-access/src/types/cap/index.ts
```diff
@@ -235,3 +235,22 @@ export interface CdsVersionInfo {
     version: string;
     root: string;
 }
+
+export type CdsUi5PluginInfo = {
+    /**
+     * Convienience property. The CDS UI5 plugin is considered enabled if `hasCdsUi5Plugin`, `hasMinCdsVersion`, `isWorkspaceEnabled` are all true.
```
Klaus-Keller, line 241: typo
```suggestion
     * Convenience property. The CDS UI5 plugin is considered enabled if `hasCdsUi5Plugin`, `hasMinCdsVersion`, `isWorkspaceEnabled` are all true.
```

packages/cap-config-writer/src/index.ts
```diff
@@ -1,4 +1,4 @@
-export { checkCdsUi5PluginEnabled, enableCdsUi5Plugin, satisfiesMinCdsVersion } from './cap-config';
```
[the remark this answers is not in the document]
heimwege: I adjusted all occurrences of `checkCdsUi5PluginEnabled` in open-ux-tools but you are right there might be usages in tools-suite as well. let's keep it.
Or we do it in two steps. First have both in then adjust tools-suite and ensure everything works and then have a follow up PR that gets rid of the import/export in cap-config-writer.

.changeset/long-planets-sing.md
```diff
@@ -0,0 +1,5 @@
+---
```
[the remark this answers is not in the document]
heimwege: I totally forgot to adjust the changeset. I should create it at the very end when the PR is ready to not fall for this again 🙈

packages/project-access/src/index.ts
heimwege: Currently it's additionally being used in cap-config-writer

https://github.com/SAP/open-ux-tools/blob/f63da1fff238397fea4c848b7f3293928b69f692/packages/cap-config-writer/src/cap-config/package-json.ts#L53-L58

packages/project-access/src/index.ts
heimwege: Same as `hasCdsPluginUi5`. Currently it's additionally being used in cap-config-writer

https://github.com/SAP/open-ux-tools/blob/f63da1fff238397fea4c848b7f3293928b69f692/packages/cap-config-writer/src/cap-config/package-json.ts#L17-L22

packages/project-access/src/index.ts
heimwege: I combined them. Was not sure about adding everything into `checkCdsUi5PluginEnabled` as it's being used in differente locations of cap-config-writer.

packages/project-access/src/project/cap.ts
heimwege: `hasDependency` was not yet exported outside of project-access. I exported it and consumed it in `app-config-writer convert preview config` as well as `cap-config-writer`.
Might be worth refactoring other occurrences as well to have this in a central place.

heimwege:
@Klaus-Keller ...and yes w.r.t. the failing unit test I might need your help. The cards-editor-middleware unit test fails because `mem-fs-editor` was added to cap.ts and I cannot get it running

image

packages/cap-config-writer/src/index.ts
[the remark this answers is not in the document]
Klaus-Keller: Perhaps unlikely, but there might be modules outside our control that use this. But then again, according to semver, changes like this are tolerated in `0.x.y` versions. Otherwise it would be a breaking change that requires mayor version bump.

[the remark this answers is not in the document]
heimwege: hm. yeah. You're right. Should I add an additional comment to the export so that nobody deletes it by mistake?

Klaus-Keller (approved):
Thanks @heimwege!
* code changes and moving of CDS check look good
* all my comments have been addressed
* changeset exists
* code coverage is good
* visual review, no manual testing done

Approved from my side.

ytpo-lyne (approved):
- fix for prerequisite check in CAP project looks good
- changeset exists
- code coverage met 
- did a manual test, no issues found 

Thanks @heimwege, approving.

PR #2923, opened 2025-02-14
fix(project-access): check for files in srv folder treats directories as files and throws exception

heimwege:
#2922

project-access [checkFilesInSrvFolder](https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/project/cap.ts#L97) currently throws an exception in case the CAP srv folder contains subfolders because those are currently being treated like files and `readFiles` dumps on folders.

@kjose90 would the proposed fix work for you?


packages/project-access/src/project/cap.ts
```diff
@@ -86,27 +86,24 @@ async function checkFilesInSrvFolder(srvFolderPath: string, memFs?: Editor): Pro
...
-    // By loading the files, we ensure they are available within mem-fs.
+    // Check for files in the file system
     if (await fileExists(srvFolderPath)) {
         const fileSystemFiles = await readDirectory(srvFolderPath);
         for (const file of fileSystemFiles) {
             const filePath = join(srvFolderPath, file);
             if (await fileExists(filePath)) {
-                const fileContent = await readFile(filePath);
-                memFs.write(filePath, fileContent);
+                return true;
```
heimwege, line 95: Only downside with this approach is that we would accept files that exist in the file system but have been deleted in memFS as a existing file.
But as this has not been checked in the erroneous version of the code either it should not harm 🤷🏻

Klaus-Keller: @heimwege, could we create a map from `memFs.dump()` before line 90 and check if the file is deleted in memFs here?

heimwege:
@Klaus-Keller I refactored to the approach I would prefer because for me this would be more reusing what we already have instead of doing the same thing over again at a different place. Can you please check?

@kjose90 Any comment from your side?

packages/project-access/src/types/file/index.ts
Klaus-Keller, line 1: I've added types for finder2 recently with https://github.com/SAP/open-ux-tools/pull/2926 🙈

Klaus-Keller:
Hi @heimwege, while reviewing this pull request I found some issues with `findBy()` when used with memFs.

Here is what happened, I wanted to test the two scenarios for determining CAP project type using memFs.

1. The `srv` folder was there but has been deleted in memFs
2. There was no `srv` folder on disk but files in `srv` folder in memFs

While the first test was fine, the second test failed. Reason is, that `findit2` uses fs.lstat to traverse paths, in case of `srv` being in memory, this throws an error here: https://github.com/andrewrk/node-findit/blob/44259da7851f4198714581d6fd7c89b256c2256d/index.js#L43.

However, the callback function passed to `finder.on('end', ..) ` is still executed. So I thought, we could not immediately `reject()` in case an error occurs, but collect errors and filter out those, that are for missing files when they exist in memFs.

While debugging this, I found another issue with the code to filter changes:
```typescript
    const filteredChanges = Object.keys(changes).filter(
        (f) =>
            fileNames.includes(basename(f)) ||
            extensionNames.includes(extname(f)) ||
            (fileNames.length === 0 && extensionNames.length === 0 && f.includes(root))
```

While you addition (fileNames.length === 0 && extensionNames.length === 0 && f.includes(root)) works, the other two cases `fileNames.includes(basename(f))` and `extensionNames.includes(extname(f))` do not consider the root path. This means, the result contains files from memFs, which are not inside the root or a subfolder of the root. 
I've fixed the issues I've found, which led to the fact, that function `findBy()` exceeded cognitive complexity of 15, which is a Sonar code smell. So I had to extract the `finder.on('end'...)`  handling into an own function `handleOnEnd()` 🙈.

As I've ended up spending quite some time on it, I took the freedom to commit the changes to this branch, if you disagree we can revert them:

https://github.com/SAP/open-ux-tools/pull/2923/commits/f905921a222478ad145f699c080eae6b52353539: 
* tests for CAP projects with memFs
* tests for `findBy()` with memFs

https://github.com/SAP/open-ux-tools/pull/2923/commits/2725963ebdb98ab2a7c13a3218c9571913dc533f: 
* ignore errors for files/folders that do not exist on physical file system but in memFs
* fix issues with files from memFs that or not inside the path the search was for
* remove type definition `Finder` in favor for `findit2.d.ts`, see https://github.com/SAP/open-ux-tools/pull/2923#discussion_r1964061474


Please check the changes @heimwege and @kjose90.

Klaus-Keller:
> @Klaus-Keller did the tests run for you? I get a lot or errors with the current version of the branch
> 
> image
> 
> Seems like there was the well known windows vs. mac path separator issue. I enforced the path.posix.sep for the memFs check and at least the tests are now green.

Sorry, missed that. For me the tests where green. Thanks for fix.

packages/project-access/src/file/file-search.ts
```diff
@@ -68,21 +113,33 @@ export function findBy(options: {
...
+                options.root,
+                errors,
+                options.memFs
+            );
+
+            if (fatalErrors.length === 0) {
+                resolve(searchResult);
+            } else {
+                // eslint-disable-next-line prefer-promise-reject-errors
+                reject(fatalErrors);
```
Klaus-Keller, line 138: @heimwege, need you opinion on this, because I was a bit struggling. You think it is of to return an array of errors? Or should be check if it just one and handle it somehow different in case multiple fatal errors exist.

Klaus-Keller (approved):
Thanks @heimwege for looking into this.
* changes to check files in virtual folder look good
* changeset exists
* test coverage is good
* visual check and manual tests done

Approved from my side.

Klaus-Keller:
@kjose90, please review this pull request.

packages/project-access/src/file/file-search.ts
```diff
@@ -10,17 +10,23 @@ import { promises as fs } from 'fs';
  * @param changes - memfs editor changes, usually retrieved by fs.dump()
  * @param fileNames - array of file names to search for
  * @param extensionNames - array of extensions names to search for
+ * @param root - path to root folder
  * @returns - array of deleted and modified files filtered by query
```
kjose90, line 14: Can you add a small example to js docs - something like below?
```ts/**
 * @example
 * const changes = {
 *   'src/app.js': { state: 'modified' },
 *   'src/test.css': { state: 'deleted' },
 *   'src/index.ts': { state: 'modified' }
 * };
 * const result = getMemFsChanges(changes, ['app.js'], ['.css'], 'src');
 * Output: { deleted: ['src/test.css'], modified: ['src/app.js'] }
 */
 ```

kjose90 (approved):
Thanks @heimwege. Changes looks great & tests to cover. ✅ 
Just a small example would be nice `getMemFsChanges`  if its not too much work.

Issue #2924, opened 2025-02-14
FEATURE - Add stable IDS for all components in AddFragment and ControllerExtension forms


mmilko01:
As a developer,
I want to write automation tests for CPE for ADP use case and I want ExtensionPoint form to have stable IDS,
so that I I can use simple selectors

### Description

ExtensionPoint for should be enhanced withs table IDS.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a ...
**when** I execute ...
**then** ... will happen.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #2925, opened 2025-02-14
feat: add stable ids to form elements

mmilko01:
#2924

- Add stable Ids in ExtensionPoint form


testojs (approved):
* small and simple change
* looks good to me

PR #2929, opened 2025-02-17
chore: Project access dependency in mockserver-config-writer

broksy:
Moved project-access to dependencies.


johannes-kolbe (approved):
- 👍 Manual test with packaged consumption side 
- changeset provided

unseen1980 (approved):
Thanks @broksy

PR #2935, opened 2025-02-18
fix(ui-components): UIButtonAction underline in normal mode

eouin:
- [x] Action button should be inverted so it's underline by default, and on-hover removed underline and add a background.


815are (approved):
checked:
1. style changes fits described behaviour
2. changeset presented

I did not test manually

Thomas-Neuhauser (approved):
lgtm

Issue #2938, opened 2025-02-18
BUG - @sap-ux/odata-service-writer - update does not write metadata.xml


docirl:
### Related Feature
Feature request: _issue number_

### Description
When `update` is called it does not write/update the metadata.xml to the localService folder. 
https://github.com/SAP/open-ux-tools/blob/f1f9cbec149678130044a3c6efe134ae8a0ea2d1/packages/odata-service-writer/src/index.ts#L113

The annotation files are created/updated as expected even if they did not exist before
The expectation calling `update` is that both the metadata and annotations will be updated or created if not existing.

### Steps to Reproduce
1. Call [update](https://github.com/SAP/open-ux-tools/blob/f1f9cbec149678130044a3c6efe134ae8a0ea2d1/packages/odata-service-writer/src/index.ts#L113) for a project that does not contain a metadata.xml already
2. See no metadata.xml file created

### Expected results
metadata.xml file is created or updated

### Actual results
no metadata.xml is written when calling `update`

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2939, opened 2025-02-18
fix(odata-service-writer): write metadata.xml from update

docirl:
https://github.com/SAP/open-ux-tools/issues/2938

### DIFF
--- a/packages/odata-service-writer/src/data/annotations.ts
+++ b/packages/odata-service-writer/src/data/annotations.ts
@@ -169,11 +169,9 @@ export async function writeLocalServiceAnnotationXMLFiles(
     templateRoot: string,
     service: EdmxOdataService
 ): Promise<void> {
-    // mainService should be used in case there is no name defined for service
-    fs.write(
-        join(webappPath, 'localService', service.name ?? 'mainService', 'metadata.xml'),
-        prettifyXml(service.metadata, { indent: 4 })
-    );
+    // Write metadata.xml file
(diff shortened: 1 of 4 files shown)


broksy (approved):
Issue case with `metadata.xml` file not getting created during update process is fixed.
Good test coverage.
Approved.

packages/odata-service-writer/src/data/annotations.ts
```diff
@@ -169,11 +169,9 @@ export async function writeLocalServiceAnnotationXMLFiles(
...
     service: EdmxOdataService
 ): Promise<void> {
-    // mainService should be used in case there is no name defined for service
-    fs.write(
-        join(webappPath, 'localService', service.name ?? 'mainService', 'metadata.xml'),
-        prettifyXml(service.metadata, { indent: 4 })
-    );
+    // Write metadata.xml file
+    await writeMetadata(fs, webappPath, service);
+
```
devinea, line 174: This doesn't seem like the best place IMO. 
Maybe `writeMetadata` should be called in `addServicesData` and `updateServicesData` or in the common function `writeRemoteServiceAnnotationXmlFiles`

devinea (approved):
Changes look good. 
1 comment that can be addressed in later PR.
Changeset ✅

packages/odata-service-writer/src/data/annotations.ts
docirl: https://github.com/SAP/open-ux-tools/issues/2944

PR #2941, opened 2025-02-18
fix: create custom table action v2 app

nahmed22:
v2 table custom action  "QA" not working when headerToolBar is undefined/null

### DIFF
--- a/packages/preview-middleware-client/test/unit/adp/quick-actions/fe-v2.test.ts
+++ b/packages/preview-middleware-client/test/unit/adp/quick-actions/fe-v2.test.ts
@@ -552,9 +552,12 @@ describe('FE V2 quick actions', () => {
                     tableType: M_TABLE_TYPE
                 },
                 {
-                    tableType: SMART_TABLE_TYPE
+                    tableType: SMART_TABLE_TYPE,
+                    overflowTable: true,
+                    headerToolbar: false
                 },
-                { tableType: SMART_TABLE_TYPE, headerToolbar: true }
+                { tableType: SMART_TABLE_TYPE, headerToolbar: true },
+                { tableType: SMART_TABLE_TYPE, showTooltip: true }
             ];
             test.each(testCases)('initialize and execute action (%s)', async (testCase) => {
                 const pageView = new XMLView();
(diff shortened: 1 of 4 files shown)


vadson71 (approved):
Changes look good
Covered in tests
Didn't test locally

packages/preview-middleware-client/test/unit/adp/quick-actions/fe-v2.test.ts
```diff
@@ -615,69 +672,92 @@ describe('FE V2 quick actions', () => {
...
+
+                        'sap.m.NavContainer': [
+                            {
+                                controlId: 'NavContainer'
+                            } as any
+                        ]
+                    });
+                }
+
+                const expectedActions = (showTooltip: boolean) => [
```
vadson71, line 705: why do you need this wrapper function?

packages/preview-middleware-client/test/unit/adp/quick-actions/fe-v2.test.ts
```diff
@@ -615,69 +672,92 @@ describe('FE V2 quick actions', () => {
...
-                    mockOverlay,
-                    rtaMock,
-                    'AddFragment',
-                    undefined,
-                    {
-                        aggregation: 'content',
-                        title: 'QUICK_ACTION_ADD_CUSTOM_TABLE_ACTION'
-                    }
-                );
+                if (!testCase.showTooltip) {
```
vadson71, line 749: This is wrong way to check expectation.
With this if condition you are just bypassing the call check and reducing test reliability.
You have to do it like expect(DialogFactory.createDialog).toHaveBeenCalledTimes(testCase.showTooltip ? 0 : 1) followed by more detailed call arguments check

showTooltip option also is not explaining test case.
Better to name it like isTableWithoutToolbar?: boolean

packages/preview-middleware-client/test/unit/adp/quick-actions/fe-v2.test.ts
nahmed22: changed.

Jimmy-Joseph19 (approved):
Changes looks good.
Chageset found for packages.
Test coverage good.
Tested locally and works as expected.
Approving.

vadson71 (approved):
Code changes look good
Excellent test coverage
Comments addressed
Didn't test locally

PR #2943, opened 2025-02-18
fix(app-modeler):  change clicking outside the toggle

nafis0077:


packages/ui-components/src/components/UIToggle/UIToggle.tsx
```diff
@@ -235,7 +236,10 @@ export class UIToggle extends React.Component<UIToggleProps, {}> {
                         marginRight: 10
                     }),
                     ...(labelFlexGrow && {
-                        flexGrow: 1
+                        flexShrink: 1,
+                        flexBasis: 'auto',
+                        cursor: 'pointer',
+                        margin: 0
```
815are, line 242: it is not necessary to add margin: 0`` here please remove or there is usecase why it is there?

packages/ui-components/src/components/UIToggle/UIToggle.tsx
```diff
@@ -235,7 +236,10 @@ export class UIToggle extends React.Component<UIToggleProps, {}> {
                         marginRight: 10
                     }),
                     ...(labelFlexGrow && {
-                        flexGrow: 1
+                        flexShrink: 1,
+                        flexBasis: 'auto',
```
815are, line 240: `flexBasis: 'auto'` and `flexShrink: auto` seems also is not required here

packages/ui-components/src/components/UIToggle/UIToggle.tsx
```diff
@@ -235,7 +236,10 @@ export class UIToggle extends React.Component<UIToggleProps, {}> {
...
+                        cursor: 'pointer',
```
815are, line 241: `cursor: pointer` should be applied for all cases - currently it applies only when `labelFlexGrow` is set

packages/ui-components/src/components/UIToggle/UIToggle.tsx
nafis0077: yes it's not necessary, removed it. Thanks

packages/ui-components/src/components/UIToggle/UIToggle.tsx
nafis0077: removed it. Thanks

815are (commented):
@nafis0077 
changes looks ok, but there is missing changelog file,
to generate changelog file run `pnpm cset` in root folder of `open-ux-tools`

815are (approved):
checked:
1. style changes are straightforward
2. changelog persists
3. tested locally with storybook

vadson71 (approved):
Changes look good
Didn't test locally

PR #2949, opened 2025-02-20
fix: Create Custom Column quick action generates incorrect fragment for Grid and Tree table 

vadson71:
Fixed bug - Create Custom Column quick action generates incorrect column fragment for Grid and Tree tables

### DIFF
--- a/packages/adp-tooling/templates/rta/common/grid-tree-custom-column.xml
+++ b/packages/adp-tooling/templates/rta/common/grid-tree-custom-column.xml
@@ -2,7 +2,7 @@
 <core:FragmentDefinition xmlns="sap.m" xmlns:table="sap.ui.table" xmlns:smartfield="sap.ui.comp.smartfield"	xmlns:core="sap.ui.core">
 	<table:Column
 		autoResizable="true"
-        width="150px" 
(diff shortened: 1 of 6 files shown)


packages/adp-tooling/templates/rta/common/grid-tree-custom-column.xml
```diff
@@ -0,0 +1,19 @@
+<!-- Use stable and unique id's!-->
+<core:FragmentDefinition xmlns="sap.m" xmlns:table="sap.ui.table" xmlns:smartfield="sap.ui.comp.smartfield"	xmlns:core="sap.ui.core">
```
Jimmy-Joseph19, line 2: smartfiled namespace can be removed as it is not used

Jimmy-Joseph19 (approved):
Changes looks good.
Test coverage good.
Tested locally
Approving

packages/adp-tooling/templates/rta/common/grid-tree-custom-column.xml
vadson71: True. Will remove it in the next PR.

Issue #2821, opened 2025-01-23
FEATURE - Generate a CAP mta.yaml configuration, supporting managed or standalone


longieirl:
_As a **persona**,
As a Fiori developer using a Fiori application within a NodeJS CAP project, I'd like to be able to add deployment configuration even if there isn't already an MTA.yaml file in my CAP project.

Description
Currently, the deployment config generator prohibits users from adding deployment config if there isn't already an MTA.yaml file in the associated CAP project. The requirement is to allow the Fiori deployment generator to optionally allow the user to create an MTA file on the fly as part of the deployment configuration, if none already exists, and then permit the user to add the Fiori deployment configuration to that MTA file.

### Description

Generate an MTA configuration and supporting deployment configurations, similar to the sample CAP apps shown here;

https://github.com/SAP-samples/fiori-tools-samples/tree/main/cap

There is an existing method called `generateBaseConfig` which will generate a base configuration using predefined templates and scripts. We need to leverage the `cds` package to generate and append services, for example;

```bash
cds init captestproject 
cds add mta xsuaa destination html5-repo
```
This will generate a `mta.yaml` with specific resources solely used for CAP projects. For example, the destination service will append a destination instance with the CAP server API exposed to allow HTML5 apps to consume the CAP OData services when deployed to CF. 

To invoke this method `generateCAPConfig`, the user should be provided with a new prompt rather than existing the process showing the error that you cant add a deployment configuration since its missing MTA.

Existing error shown;

```
[2025-02-05 15:27:25] ERROR: The SAP Fiori application is within a CAP project and deployment should be configured as part of the CAP project. Please ensure you have a mta.yaml file defined for this project.
```

Here is a sample screenshot of the new prompt;

![Image](https://github.com/user-attachments/assets/64aa3c3a-e274-494e-a08a-4f05422b9c07)

The sample text;
```
There is no mta.yaml file defined for this project. In order to add deployment configuration for this application, this file must be present. Do you want to create an mta.yaml to continue?
```

### Technical Design
- enhance `cf-config-writer`
- enhance `@sap-ux/cf-deploy-config-sub-generator`
- expose a new generate method to specifically handle CAP mta
- expose a new prompt to handle CAP project and missing MTA

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have an existing CAP project with no existing mta.yaml
**when** I execute the CAP generate method, selecting either managed or standalone approuter
[...]

PR #2956, opened 2025-02-24
fix(cf-deploy-sub-gen): handle flow when loaded as sub-gen

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/2821

- found when testing locally in tools-suite
- init should only be called when run standalone
- add tests to cover this change


IainSAP (approved):
Thanks @longieirl 

- Updates are clear, separate function
- Tests to cover
- Not tested locally

Issue #2962, opened 2025-02-26
BUG - Unable to create/edit destinations when user is missing CF privileges


longieirl:
### Description
During the selection of an ABAP instance, the user selects a system but the process exists showing the root cause of the exception;

![Image](https://github.com/user-attachments/assets/32f590dd-7853-4c88-bed8-862306c84668)

![Image](https://github.com/user-attachments/assets/e897fe18-e96d-4fc9-b771-b6c6069cf78c)

### Steps to Reproduce
1. User is missing the subaccount admin role on their BTP subaccount
2. Create a new project, selecting any template
3. User selects an ABAP instance

### Expected results

User is shown the error why they are unable to proceed.

### Actual results

Wizard flow exists and user has to start again

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Validated in tools-suite.

PR #2963, opened 2025-02-26
fix(odata-service-inq): handle exception if user is missing cf privileges to create CF destinations

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/2962


IainSAP (approved):
Thanks @longieirl 

- Appropriate fix for uncaught error
- Test added to cover
- Not tested locally

Issue #2964, opened 2025-02-27
BUG - TypeScript template causes the iframe to break when controller extension is loaded


nikmace:
### Related Feature
Feature request: https://github.com/SAP/open-ux-tools/issues/2811

### Description
Template causes the Application to break because `this` points to the browser and not to the controller extension in the `onInit` method. Refactored the `onInit` to be the same way as Fiori controller extensions.

### Steps to Reproduce
1. Start an Adaptation Editor from a TypeScript project
2. Add XML fragment and controller extension (for example a button in the toolbar)
3. Reload the application, and see error

### Expected results
Application iframe load correctly

### Actual results
Application iframe shows error that target was not found because of a problem with the ts template.

### Screenshots
If applicable, add screenshots to help explain the problem.

<img width="447" alt="Image" src="https://github.com/user-attachments/assets/98def95a-f1ba-46fd-a1b8-9607d67e3ddf" />

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2965, opened 2025-02-27
fix: TypeScript template causes the iframe to break when controller extension is loaded

nikmace:
Fix for #2964.
- Refactores `onInit` method to accept a `this` parameter that causes problems during controller extension loading.


devinea (approved):

heimwege (approved):
lgtm

testojs (approved):
LGTM

PR #2968, opened 2025-02-27
Feat/preview middleware/2942 support browser locale in generated application

korotkovao:
Remove English language config from HTML templates to let it adapt to customers' browser language.


packages/cards-editor-middleware/templates/flpSandbox.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE HTML>
 <!-- Copyright (c) 2015 SAP AG, All Rights Reserved -->
-<html lang="en">
+<html>
```
heimwege, line 3: @donal-tobin-sap didn't we discuss in one of that last architecture syncs that the language property is needed here for accessibility if I remember it correctly?

donal-tobin-sap: Yes see: https://dequeuniversity.com/rules/axe/4.2/html-has-lang
Required by WCAG standards.

korotkovao: Thank you for the review.
The language config is back.

packages/preview-middleware/templates/flp/sandbox.html
```diff
@@ -53,7 +53,6 @@
         data-sap-ui-preload="async"
         data-sap-ui-theme="<%- ui5.theme %>"
         data-sap-ui-compatVersion="edge"
-        data-sap-ui-language="en"
```
heimwege, line 56: At least for preview-middleware this introduces the following regression issue:

Given the setup that you use the preview-middleware (or fiori-tools-preview) to run your OPA5 tests on virtual endpoints (locally or in the pipeline). Currently the tests run in language EN. Always.

This PR would change the language to browser locale by default, so it might be that the OPA5 tests start failing because of a different language being used.

To prevent this you need to know that you can pass the `sap-ui-language` parameter directly in your journey ([sample](https://github.com/heimwege/cap-node-fe-ts-sample/blob/main/app/samples/webapp/test/integration/ListReportJourney.ts#L17)).
But I doubt that this is common knowledge.

@tobiasqueck @Klaus-Keller your thoughts?

Klaus-Keller: @heimwege, good point!
@korotkovao, where is this requirement coming from? As from this pull request I can see changes for `flpSandbox.html`, only. This file is not for productive usage, it is for developers to check, preview, and test the application during development time. Here it might be of advantage to have a stable language setting, which can still be controlled and overwritten using URL parameter `sap-language=XY`.

korotkovao:
Closing this PR as there is configuration already available.

PR #2969, opened 2025-02-27
fix: compile error when new .cds files are added to memfs

voicis:
Follow up to #2884. That solution was not sufficient for the case when there is no file in the `app` folder for `cds.resolve` to pick up.


vadson71 (approved):
Changes look good
Test coverage is good
Didn't test locally

marufrasully (approved):
Thanks @voicis 
* code changes look good
* secured with unit test
* did not test it locally

Issue #2972, opened 2025-02-28
BUG - Unable to append managed approuter when executed from CLI


longieirl:
### Related Feature
Feature request: _issue number_

### Steps to Reproduce
1. run npx fiori add deploy-config
2. select CF
3. select add to managed approuter

### Expected results

mta.yaml is generated with managed approuter

### Actual results

mta.yaml is missing managed approuter

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Work around;

Step1. Load the CF App router

![Image](https://github.com/user-attachments/assets/b07afc7c-e70c-4955-b9f6-6fff5bd49d87)


Step2. Create your app using managed approuter;

![Image](https://github.com/user-attachments/assets/daa65834-fa73-4e91-8eef-f0dc6ea9d7d9)

Step 3. Append a HTML5 app to the existing mta.yaml

![Image](https://github.com/user-attachments/assets/d26e7df0-40ac-41f9-84a0-36dba594e253)

You can then follow the existing steps in your tutorial. The only change will be where the question to append to a managed approuter wont be present since it already has a managed approuter.

PR #2973, opened 2025-02-28
fix(cf-deploy-sub-generator): ensure addmanagedapprouter is appended

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/2972

- managed approuter not being appended when selected


kjose90 (approved):
Changes looks good ✅ 
Thanks.

Issue #2975, opened 2025-02-28
BUG - Reference lib inquirer throws error when .library not found at path of library.(js|ts)


cianmSAP:
### Related Feature
Feature request: https://github.com/SAP/open-ux-tools/issues/1800

### Description
It is assumed a `.library` file will be found at the same location as the `library.js` file. There has been examples of `library.js` files within test folders (and no `.library` file) and as a result it crashes the reference library generator

### Steps to Reproduce
1. Add a library project to a folder
2. Add a mock folder with a `library.js` file
3. Use the Reference library inquirer

### Expected results
The inquirer (or project-access) just ignores this error and does not add the path to the library

### Actual results
It throws an error and crashes the generator

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #2976, opened 2025-02-28
fix(project-access): ignore missing .library file error

cianmSAP:
#2975

If the `.library` file is not found in the same directory as the `library.js/ts` file then the error should be ignored


devinea (approved):
Thanks changes look safe enough to ignore error.
Changeset 👍

PR #2977, opened 2025-02-28
Chore(preview middleware): fix eslint warnings

heimwege:


heimwege:
So if the type assertion is in, Sonar complains about it being redundant

image

but if it's out eslint complains about an unsafe assignment of an error typed value.

image

Can this problem be solved at all? 🤔

PR #2978, opened 2025-02-28
Export `patchTls` from `sap-ux/axios-extension`

kjose90:
Export `patchTls` from `sap-ux/axios-extension`  to improve reusability


longieirl (approved):
Changes look good.

Issue #2971, opened 2025-02-28
BUG - UISections. The content is "jumping" around when show/hide sections


IvoSG:
### Steps to Reproduce
1. We have UISections like:
```
const [firstSectionSize, setFirstSectionSize] = useState(200);
<UISections sizes={[firstSectionSize, undefined, 200]] />
```
2. Initial size of `firstSectionSize` is `200`;
3. Then we change `firstSectionSize` from `200` to `300` using `setFirstSectionSize(300)`;
4. 

Problem - The issue is that when the sizes are recalculated and one becomes zero, the element gets hidden based on the condition, reappearing when its size increases again. This results in strange jumping behavior and suspicious splitter positioning.

### Expected results
Size of sections/splitter is recalculated when external sizes are changed or hidden/show.

### Actual results
The size of sections/splitter is not recalculated properly with hidden/show sections.

### Screenshots

https://github.com/user-attachments/assets/0e6d86ab-d9da-4c0f-924c-4e8b1407f5ee

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

815are:
PR with fix https://github.com/SAP/open-ux-tools/pull/2979
closing issue, please correct if there is still issue with latest version

PR #2979, opened 2025-03-02
fix(ui-components): UISections. Fixed incorrect position calculations when hiding sections with three or more sections present.

815are:
Issue #2971

Issue happens when we have 3 sections and we toggle visibility of sections:

https://github.com/user-attachments/assets/ce8fe2e2-d5a4-4e60-b375-b4c50c66fcb8

After fix:

https://github.com/user-attachments/assets/9dbaa0b4-ede4-46ae-85a6-1d66fbf2f8b3


nikmace:
Does this need a patch for `@sap-ux/control-property-editor` because it is consumed there?

815are:
@nikmace 

Update changelog to include `@sap-ux/control-property-editor` into changelog - https://github.com/SAP/open-ux-tools/pull/2979/commits/7b1878b9a48dfcaef042770597114c0969a3fd05

PR #2980, opened 2025-03-03
fix inserting target for complex type

Jimmy-Joseph19:


packages/cds-odata-annotation-converter/src/printer/csdl-to-cds.ts
```diff
@@ -128,9 +128,23 @@ export const printTarget = (target: Target): string => {
             //        <term> <qualifier>: <value>
             //     );
             // }
-            result = `annotate ${rootElementName} with {\n${childSegments[0]} ${
-                target.terms.length > 1 ? `@(${result})` : `@${result}`
-            }};\n`;
+            const terms = target.terms.length > 1 ? `@(${result})` : `@${result}`;
+            if (isStructureTargetType) {
+                let closingBraces: string[] = [];
+                const complexTarget = childSegments[0].split('_').reduce((acc, item, index, array) => {
```
voicis, line 134: We should not try to rebuild the structure here, because we don't have all the information here. Instead the structure should be passed as a parameter.

voicis (changes requested):
Please also add API level tests for this case in `fiori-service.test.ts`.

packages/cds-odata-annotation-converter/src/printer/csdl-to-cds.ts
Jimmy-Joseph19: adapted

voicis (dismissed):
- Code changes looks good
- Review comments addressed
- Good code coverage
- Did not test locally

PR #2983, opened 2025-03-03
chore: update dependencies

voicis:
Update `vscode-languageserver-textdocument` to match language server.

This update was causing tests to fail in `preview-middleware-client`, because `vscode-languageserver-textdocument` now includes `exports` field in package.json which is used to determine package entry points. When using `jsdom` environment Jest uses `browser` entry point which defaults to ESM, but that it is not compatible with the current test setup, so we need to override the module import logic. 

https://jest-archive-august-2023.netlify.app/docs/28.x/upgrading-to-jest28/#packagejson-exports

### DIFF
--- a/.changeset/beige-days-applaud.md
+++ b/.changeset/beige-days-applaud.md
@@ -1,0 +1,6 @@
+---
+'@sap-ux/fiori-annotation-api': patch
+'@sap-ux/i18n': patch
+---
+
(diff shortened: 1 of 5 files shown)


marufrasully (approved):
@voicis thanks a lot.
LGTM

Jimmy-Joseph19 (approved):
Update looks good.
Changeset for packages found.
Approving

PR #2994, opened 2025-03-04
feat: use webappPath for mockserver generation

broksy:
Adds `webappPath` and `basePath` parameters to resolve service paths during mockserver generation/update.
Fixes cases when custom webapp paths are used.

### DIFF
--- a/.changeset/neat-grapes-design.md
+++ b/.changeset/neat-grapes-design.md
@@ -1,0 +1,6 @@
+---
+'@sap-ux/mockserver-config-writer': patch
+'@sap-ux/ui5-config': patch
+---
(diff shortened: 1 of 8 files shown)


devinea (dismissed):
passing parameters for webapp path seems to allow using custom webapp folder names. 
changeset ✅ 
not immediately obvious that tests include an example with a custom webapp path.

devinea (approved):
Thanks for update. Re-approving.

815are (approved):
checked:
1. code changes look straightforward to allow pass and allow pass webapp path
2. tests are maintained
3. changelog persists

Issue #2999, opened 2025-03-06
BUG - getModule fails to load module from cache if first installation did not finish or it installed broken version


815are:
### Description
`getModule` fails to load module from cache if first installation did not finish or it installed broken version

### Steps to Reproduce
Steps to simulate the behavior:
1. Navigate to `.fioritools\module-cache`
2. Use any dependency/module
3. For example you should have `.fioritools\module-cache\your_dep\1.124.11\node_modules\your_dep`;
4. Delete folder `.fioritools\module-cache\your_dep\1.124.11\node_modules\your_dep`;
5. Try to load module again using `getModule`

Problem - loading fails.
One way to fix it is to delete folder `.fioritools\module-cache\your_dep`

### Expected results
Module is loaded

### Actual results
Module is not loaded

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3000, opened 2025-03-06
fix(project-access): Improve `getModule` to attempt module reinstallation if loading fails

815are:
Issue #2999 

It was reported multiple times by different persons that loading of `specification` module fails and during investigation we noticed that it fails to load from `.fioritools\module-cache`.

While checking one of this case - we noticed:
1. Folder with dependency exists in `.fioritools\module-cache\test` and there is `package.json` and `package-lock.json`
image
2. Folder with dependencies like `fioritools\module-cache\your_dep\1.124.11\node_modules\your_dep` does not have `package.json` and also not all js sources are there:
image

As fix:
1. if loading of module fails, then try to reinstall it:
1.1. If `package-lock.json` exists run `npm ci`
1.2. If `package-lock.json` does not exist rerun `npm i ....@version..`


Klaus-Keller (approved):
Thanks @815are!
* code to retry installation looks good
* changeset exists
* test coverage is great

approved from my side.

broksy (approved):
Changeset reflects to the changes made and are fully covered by unit tests.
Thanks, @815are ! Approved from my side.

PR #3007, opened 2025-03-10
fix: set timeout in before all

marufrasully:
Problem:
Timeout was set in config which effects all tests timeout.[More on timeout](https://playwright.dev/docs/test-timeouts). 
If there was any failure, playwright would re-try one more time after waiting `5 min + 30 second` and this caused timeout in pipeline which has only [30 min](https://github.com/SAP/open-ux-tools/blob/main/.github/workflows/pipeline.yml#L15). Playwright report artifacts were not uploaded due to pipeline time out.

Possible solution:
Add time out only in `beforeAll` which affects only before all. [More on this check.](https://playwright.dev/docs/test-timeouts#change-timeout-for-beforeallafterall-hook).

### DIFF
--- a/packages/preview-middleware/playwright.config.ts
+++ b/packages/preview-middleware/playwright.config.ts
@@ -6,7 +6,6 @@ import type { PlaywrightTestConfig } from '@sap-ux-private/playwright';
  */
 import 'dotenv/config';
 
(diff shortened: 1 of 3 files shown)


devinea (approved):

heimwege (approved):
lgtm

PR #3008, opened 2025-03-10
fix(deploy-gen): fix undefined error when no backend config found

cianmSAP:
When the deployment generator is ran standalone, against a project with no existing backend configuration, the generator is throwing an undefined error when trying to read `backendConfig.destination`


kjose90 (approved):
Bug fix is clear, tests and change request to cover ✅

longieirl (approved):
Changes are good.
Tests to coverage chanage.
Changeset is good.

Issue #3010, opened 2025-03-10
BUG - fails to populate deployment destination - ADT deep link


longieirl:
### Description
When the following deep link is called, the ABAP destination is not prepopulated when the deployment generator is loaded;

```
https://mystaging.somelink.int.applicationstudio.cloud.sap/index.html?arg-service-url=https%3A%2F%2Finternallink.sap.corp%3A44300%2Fsap%2Fopu%2Fodata4%2Fdmo%2Fui_travel_d_d_o4%2Fsrvd%2Fdmo%2Fui_travel_d_d%2F0001%2F%24metadata%3Fsap-client%3D000&arg-primary-entity=Booking&ds-id=ws-xw6p8&action-id=sap.ux.appGenerator.adt.launch&action-type=COMMAND&action-name=sap.ux.appGenerator.adt.launch&arg-destination-name=DEST_TEST&folder=/home/user/projects#ws-xw6p8
```

### Steps to Reproduce
1. Load the above deep link into a known dev workspace
2. Select yes to adding deployment config

### Expected results

the ABAP target should be selected by default and the destination `DEST_TEST` should be prepopuated

### Actual results

The destination is not prepopulated.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Validated in tools-suite, working as expected.

PR #3011, opened 2025-03-10
fix(deploy-config-sub-generator): append destination to backendconfig

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/3010

- ensure the destination passed from another generator is appended to backend config
- add unit tests to cover change


kjose90 (approved):
lgtm. tests and change set ✅

PR #3014, opened 2025-03-11
fix(ui-components): UISections. Error `Cannot read properties of undefined (reading 'size')` is throws for section on initial rendering 

815are:
Part of #2971
Regression after https://github.com/SAP/open-ux-tools/pull/2979

There was missing check for `undefined`:
```
const size = sizes[i].size ?? 0;
```

In case if previous/original sizes from state are not set yet.

### DIFF
--- a/.changeset/itchy-students-retire.md
+++ b/.changeset/itchy-students-retire.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui-components': patch
+---
+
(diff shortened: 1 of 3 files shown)


devinea (approved):
Changes are clear and tested. 
Changeset 👍

AlinaGovoruhina (approved):
Change is clear, changeset added.

PR #3020, opened 2025-03-12
Feat(UI components): issue #32314 - UITranslationInput inverted theme

eouin:
fix for issue #[32314](https://github.wdf.sap.corp/ux-engineering/tools-suite/issues/32314)

- [x] update the text of the translation input callout dialog following story requirement
- [x] fix the CSS accordingly
- [x] fix CSS for colour high contrast theme and other colour theme to have the inverted dialog colour


815are (approved):
checked:
1. code change and style updates are clear - do not see issue
2. changeset is presented

I did not test manually

packages/ui-components/src/components/UITranslationInput/defaults.ts
```diff
@@ -1,16 +1,17 @@
 import type { TranslationInputStrings } from './UITranslationButton.types';
 
 export const defaultTranslationInputStrings: TranslationInputStrings = {
-    acceptButtonLabel: 'Apply',
+    acceptButtonLabel: 'Substitute',
```
815are, line 4: Only not sure about this label `Substitute` - I think it is not always `Substitute`

packages/ui-components/src/components/UITranslationInput/defaults.ts
```diff
@@ -1,16 +1,17 @@
...
     cancelButtonLabel: 'Cancel',
     i18nKeyMissingTooltip: 'Text key or value is not available in i18n file.',
     i18nKeyMissingDescription: 'Generate a text key {{{key}}} with value {{{value}}} in i18n file.',
     i18nValueMissingTooltip: 'Text key or value for {{{value}}} is not available in i18n file.',
-    i18nValueMissingDescription: 'Generate a text key {{{key}}} in i18n file and substitute {{{value}}} by {{{i18n}}}.',
+    i18nValueMissingDescription:
+        'Generate a text key {{{key}}} in i18n file and substitute {{{value}}} with {{{i18n}}}.',
```
hitesh-parmar, line 10: @gpeuc Shouldn't we also add a question mark at the end here?

packages/ui-components/src/components/UITranslationInput/defaults.ts
```diff
@@ -1,16 +1,17 @@
...
     i18nReplaceWithExistingTooltip:
-        'Text key {{{key}}} for value {{{value}}} is available in i18n file. \nConsider substituting {{{value}}} by {{{i18n}}}.',
+        'Text key {{{key}}} for value {{{value}}} is available in i18n file. \nConsider substituting {{{value}}} with {{{i18n}}}.',
```
hitesh-parmar, line 12: @eouin Do you know a scenario when this text is used? This text looks almost same to the one below.

eouin: One of them is used for the popup message and the other one for the tooltip

CleanShot 2025-03-20 at 13 50 42@2x

packages/ui-components/src/components/UITranslationInput/defaults.ts
hitesh-parmar, line 4: Another question @eouin is there a reason not to use i18n .json files for texts?

eouin: @hitesh-parmar Don't know what to answer, I haven't created that component in the first place, maybe you should ask @815are as he's the one who created it.

815are: As I remember idea was that we follow approach how it is done in fluent-ui, for example -> https://github.com/microsoft/fluentui/blob/master/packages/react/src/components/Calendar/defaults.ts

lfindlaysap: Thanks, @815are. Is this widespread or is this the only file?

PR #3012, opened 2025-03-10
fix(deploy-config): fix errors running standalone

cianmSAP:
- Remove references to `confirmConfigUpdate` as the prompt has been removed
- Remove responsibility of setting standalone header/description out of the subgen
- Fix logging error message

PR #3023, opened 2025-03-13
fix(deploy-gen): support running deploy generator standalone

cianmSAP:
Revert some changes made in https://github.com/SAP/open-ux-tools/pull/3012 to fully support running this generator standalone in YUI, without the requirment for a 'wrapper' generator

### DIFF
--- a/packages/deploy-config-sub-generator/src/app/index.ts
+++ b/packages/deploy-config-sub-generator/src/app/index.ts
@@ -8,18 +8,19 @@ import {
     TargetName,
     getExtensionGenPromptOpts
 } from '@sap-ux/deploy-config-generator-shared';
-import { parseTarget } from './utils';
+import { parseTarget, getYUIDetails } from './utils';
 import {
     getApiHubOptions,
     getEnvApiHubConfig,
     t,
     generatorNamespace,
     getBackendConfig,
-    getSupportedTargets
+    getSupportedTargets,
+    generatorTitle
 } from '../utils';
+import { AppWizard, Prompts } from '@sap-devx/yeoman-ui-types';
 import { promptDeployConfigQuestions } from './prompting';
 import { promptNames } from '../prompts/deploy-target';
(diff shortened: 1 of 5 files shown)


packages/deploy-config-sub-generator/src/app/index.ts
```diff
@@ -69,6 +70,11 @@ export default class extends DeploymentGenerator implements DeployConfigGenerato
         this.target = parseTarget(args, opts);
         this.vscode = opts.vscode;
 
+        if (!this.env.isPackageRegistered(this.genNamespace)) {
```
cianmSAP, line 73: There's no real way to test this afaik, the yeoman test environment registers the namespace `gen:test` by default when running the generator. I can change the namespace but can't deregister it

IainSAP (approved):
Thanks @cianmSAP 

- Changes seem appropriate for  use standalone or with `composeWith`

IainSAP (approved):

packages/deploy-config-sub-generator/test/utils.test.ts
```diff
@@ -55,4 +55,11 @@ describe('Test utils - Deploy', () => {
         delete process.env['API_HUB_TYPE'];
         expect(getEnvApiHubConfig()).toBeUndefined();
     });
+
+    it('should return prompt step details for yui usage', () => {
+        const yuiDetails = getYUIDetails('/path/to/app/project1');
```
kjose90, line 60: good to use path.join() ... windows may cause an issue?

cianmSAP: It's safe in this context because how it is used, the path is passed and then it's called with `basename()` and I only assert the basename e.g `project1`

kjose90 (dismissed):
Looks good 👍

kjose90 (approved):
re approving

IainSAP (approved):
Re-approving

PR #3024, opened 2025-03-13
fix: Improved add, delete and get methods for backends

broksy:
Improved add, delete and get methods for backends.
The goal is to restrict the usage of multiple backends with same path.

- Changed deletion logic to delete items by path (`removeBackendFromFioriToolsProxydMiddleware`) and always leave entry with `/sap` path.
- Added check for existing backends by path before backend addition (`addBackendToFioriToolsProxydMiddleware`).
- Added new get method (`getBackendConfigFromFioriToolsProxydMiddleware`) to return matching backend entry by path.
- Enhanced preview setting determination in `enhanceData`


devinea (approved):
Code changes look good. 
Tested locally in service manager add and remove service from different backends (url and client) work as expected. 
Covered by tests
changeset ✅

johannes-kolbe (approved):
- Changes lg and address described behaviour
- unit tests provided
- changelog provided
- ✅ manual consumption and test

PR #3026, opened 2025-03-13
tbi(fiori-freerstyle-writer): Export required type

IainSAP:
Export a type required by the consuming generator


cianmSAP (approved):
Looks good to me

PR #3029, opened 2025-03-14
fix(ui-components): UISections. Issue with window resize hiding section on fullscreen toggle event

815are:
issue/regression after https://github.com/SAP/open-ux-tools/pull/2979
Scenario of issue:
1. Two sections with initial size `[400, available_space]` and with min sizes `minSectionSize={[400, 400]}` and window size `1000px x 600px`;
2.  Make screens size less than 800px in width to trigger full screen toggle - hide one of section;
3. Then make screen wider to toggle non-fullscreen mode back and restore visibility of section back;

Problem - size of restored section is wrong.

### DIFF
--- a/.changeset/early-ligers-beam.md
+++ b/.changeset/early-ligers-beam.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui-components': patch
+---
+
+UISections. Issue with window resize hiding section on fullscreen toggle event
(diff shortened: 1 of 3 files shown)


broksy (approved):
Changes corresponds to the changes made and are covered by unit test.
Approved from my side.

eouin (approved):

devinea (approved):

PR #3032, opened 2025-03-17
WIP feat: partial V3 migration - proxy middleware

heimwege:


packages/ui5-proxy-middleware/src/ui5/middleware.ts
```diff
@@ -19,9 +19,9 @@ import type { ReaderCollection } from '@ui5/fs';
  */
 function createProxyOptions(logger: ToolsLogger, config: UI5ProxyConfig): Options {
     return {
-        secure: config.secure !== undefined ? !!config.secure : true,
-        logLevel: config.debug ? 'debug' : 'info',
-        logProvider: () => logger
+        secure: config.secure !== undefined ? !!config.secure : true
+        // ToDo: logger
+        // logger
     };
```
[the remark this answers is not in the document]
heimwege: fixed

packages/backend-proxy-middleware/src/middleware.ts
```diff
@@ -20,27 +21,31 @@ function formatProxyForLogging(proxy: string | undefined): string | undefined {
             proxy = proxy.replace(proxy.slice(forwardSlashIndex + 2, atIndex), '***:***');
         }
     }
+    // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
     return proxy || 'none';
```
nlunets, line 25: why not write proxy ?? 'none' ?

nlunets (commented):
changeset ok
small comment

packages/backend-proxy-middleware/src/middleware.ts
heimwege: `||` also works when the proxy string is empty (`""`) and returns `none` whereas the `??` would return `""` in that case.

I actually did the change you suggested but then we had that empty string case and wondered why it is being logged as `""` instead of `none`. So I reverted my adjustment and added the esline disable statement.

Issue #3035, opened 2025-03-18
BUG - @sap-ux/ui-service-inquirer - service name request not using correct package value


docirl:
### Description
The api request sent to get suggested content and service name uses the package name provided in the previous prompt. Due to incorrect condition determining the value to use for package, when using the option 'choose from existing' the request is sent with an empty string. 

This causes the response to contain an empty/invalid config, which returns a blocking validation error on service name

### Steps to Reproduce
1. Open ui service sub generator
3. Select system and object, go to config step
4. Enter `$TMP` as 'choose from existing' package prompt
5. See validation error on service name

### Expected results
Using a value in 'choose from existing' package returns a valid suggested service name

### Actual results
Using a value in 'choose from existing' package is returning an error that blocks service generation.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

docirl:
Done with https://github.com/SAP/open-ux-tools/pull/3036

PR #3036, opened 2025-03-18
fix(ui-service-inq): ensure correct package value is used for getContent request

docirl:
https://github.com/SAP/open-ux-tools/issues/3035

- update condition logic to ensure correct value is passed to getContent request
- update test to cover flow


cianmSAP (approved):
Looks good to me
Update for known issue with empty strings
Changeset

longieirl (approved):
Changeset is good.
Updated tests.
Small code change.

PR #3039, opened 2025-03-20
fix: align mem-fs versions

longieirl:


packages/abap-deploy-config-sub-generator/package.json
```diff
@@ -51,7 +53,6 @@
         "@sap-ux/store": "workspace:*",
         "@sap-ux/system-access": "workspace:*",
         "memfs": "3.4.13",
```
IainSAP, line 55: Shouldn't `memfs` be removed?

PR #3041, opened 2025-03-20
fix(ui-service-gen): hide draftEnabled for abap cds views

docirl:
https://github.com/SAP/open-ux-tools/issues/3112

- remove `draft enabled` prompt from UI Service Generator when ABAP CDS view is selected
- update and add tests to cover new code
- added options used to hide draftEnabled and launchAppGen prompts (other prompts are mandatory)


packages/ui-service-sub-generator/src/app/index.ts
```diff
@@ -153,6 +153,11 @@ export default class extends Generator {
         // prompt service configuration
         const configPrompts = await getConfigPrompts(
             this.systemSelectionAnswers,
+            {
+                useDraftEnabled: !(
+                    this.answers.objectType === ObjectType.CDS_VIEW || this.options.data?.type === BAS_OBJECT.CDS
+                )
```
cianmSAP, line 159: You don't need to anything for `useLaunchGen` here no?

cianmSAP (approved):
Comments addressed based on conversation offline
Updates look good

packages/ui-service-sub-generator/src/app/index.ts
docirl: Not specifically, I added the option while I was there as it is the only other prompt that could be considered conditional. Currently there is no requirement to hide it.

devinea (approved):
Code changes seems clear and well covered by tests. 
changeset ✅

Issue #3015, opened 2025-03-12
FEATURE - Add Adaptation Project generator


nikmace:
_As a **developer**,
I want to **create an Adaptation Project**,
so that I **can adapt base applications**_

### Description

_The **first iteration** will focus on creating the new yeoman package. Additionally, we’ll add the first page with system and application selection prompts. The remaining prompts will have default values. This setup is sufficient to generate an Adaptation Project similar to the CLI._

_The **second iteration** will include the second page from the original creation generator. This page will contain all the configuration prompts, such as name, title, namespace, target folder, UI5 version, enable TypeScript, add deploy and flp configs._

### Technical Design

Which new packages should be created (naming)? 
**@sap-ux/generator-adp** (yeoman generator package)

How should the new package folder be named? 
packages/**adp-generator**

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a transferred all of the code
**when** I run the generator
**then** I am able to create an Adaptation Project

### Notes

This is a general issue for tracking all of the progress regarding the code transfer.

### Tasks
- [x] Create the new yeoman generator package with the prompts from the first page which will create an adaptation project. 
- [x] Enhance the newly created package with the rest of the prompts from the original creation generator
- [ ] Improve ADP generator writer by refactoring its logic
- [ ] Consume flp and deploy sub-generators

PR #3049, opened 2025-03-26
fix(generator-adp): Generator running in the CLI context

nikmace:
Fix for #3015.
- Fixes generator flow when running with `yo` command in the CLI context.
- Improves error handling for system prompt.
- Fixed incorrect custom config generation.
- Refactor classes to be standalone methods for less complexity.


testojs (approved):
* code is improved
* changeset OK
* issue with the CLI resolved
* tested manually and works as expected

PR #3051, opened 2025-03-26
fix(cf-deploy-config-writer): spawn options required for windows

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/2821

- when generating mta on windows getting an error;
```
Error: Something went wrong creating mta.yaml!
```
The `cds` binary is not being inherited from the shell.

### DIFF
--- a/.changeset/tasty-islands-nail.md
+++ b/.changeset/tasty-islands-nail.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/cf-deploy-config-writer': patch
+---
+
+add spawn options for windows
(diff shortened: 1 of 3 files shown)


devinea (approved):
Changes look appropriate.
Could consider using https://github.com/SAP/open-ux-tools/blob/main/packages/nodejs-utils/src/commandRunner.ts ?

longieirl:
> Changes look appropriate. Could consider using https://github.com/SAP/open-ux-tools/blob/main/packages/nodejs-utils/src/commandRunner.ts ?

Thanks @devinea I will create a [TBI](https://github.com/SAP/open-ux-tools/issues/3057) as this will require some further testing.

cianmSAP (approved):
Update looks good
This issue has come up before as @devinea pointed out

PR #3052, opened 2025-03-27
add check-dependency-version-consistency script

devinea:
Useful script to check for npm dependencies that are using different versions across modules.


donal-tobin-sap (approved):
New check script looks good.

PR #3060, opened 2025-03-31
fix(ui-components): Remove font-family definitions from individual components and apply it globally.

815are:
Recently we introduced new API method to apply theme and time to time we still find that some components uses default fluent-ui font family.
As solution - apply global font family and remove font-family definitions from individual components.


Vladislavs-Silins-SAP (approved):
The code looks good and the changeset is present. I haven't tested the changes in the extension.

packages/ui-components/src/components/UIInput/UITextInput.tsx
```diff
@@ -85,8 +85,7 @@ export class UITextInput extends React.Component<UITextInputProps> {
         return {
             ...{
                 root: {
-                    height: 'auto',
-                    fontFamily: 'var(--vscode-font-family)'
```
tborychowski, line 89: as inputs don't inherit font from the parent, should we leave at least `font: inherit` here?

packages/ui-components/src/components/UIInput/UITextInput.tsx
```diff
@@ -154,11 +153,9 @@ export class UITextInput extends React.Component<UITextInputProps> {
                         fontWeight: 'normal',
                         boxSizing: 'border-box',
                         borderRadius: COMMON_INPUT_STYLES.borderRadius,
-                        fontFamily: 'var(--vscode-font-family)',
```
tborychowski, line 157: and here

packages/ui-components/src/components/UISearchBox/UISearchBox.tsx
```diff
@@ -30,7 +30,6 @@ export class UISearchBox extends React.Component<ISearchBoxProps, {}> {
             ...{
                 root: [
                     {
-                        fontFamily: 'var(--vscode-font-family)',
```
tborychowski, line 33: and probably here as well

packages/ui-components/src/components/UIInput/UITextInput.tsx
tom-sap, line 89: as inputs don't inherit `font` from the parents, should we leave at least `font: inherit` here?

packages/ui-components/src/components/UIInput/UITextInput.tsx
tom-sap, line 157: and here?

packages/ui-components/src/components/UISearchBox/UISearchBox.tsx
tom-sap, line 33: and probably here as well (if search input doesn't inherit from some common input component).

packages/ui-components/src/components/UIInput/UITextInput.tsx
815are: As i checked, then we do not need it, fluent-ui still sets font-family to component itself, but instead of using default/fluent-ui font, then our font-family is used:
image

packages/ui-components/src/components/UIInput/UITextInput.tsx
815are: as checked - I do not see reason to leave it, same explanation as - https://github.com/SAP/open-ux-tools/pull/3060#discussion_r2020847135

packages/ui-components/src/components/UISearchBox/UISearchBox.tsx
815are: as checked - I do not see reason to leave it, same explanation as - https://github.com/SAP/open-ux-tools/pull/3060#discussion_r2020847135

packages/ui-components/src/components/UIInput/UITextInput.tsx
tborychowski: this is a div, not an input. 
But if the `input` has the same styling, than great :-) 
(just wanted to make sure)

tom-sap: this is a div, not an input.
But if the input has the same styling, than great :-)
(just wanted to make sure)

PR #3061, opened 2025-03-31
fix (ui-components): standardize SwitchOff and SwitchOn icons to 16x16

Vladislavs-Silins-SAP:
The icons <i>SwitchOff</i> and <i>SwitchOn</i> were standardized to 16×16 size, and the UIToggle component was adjusted to use these icons.


815are (approved):
checked:
1. code changes looks correct - icons are centered and has standard size and colors. Component code is simplified to use generic approach to center icon;
2. Tests are adjusted
3. Changeset persists

eouin (approved):

devinea (approved):

gpeuc (commented):
I have just seen this, we should not be forcefully resizing icons. The pixel size they are designed at has to stay that size. If we need bigger or smaller icon, I will design is for the target size.

PR #3062, opened 2025-03-31
fix: call `initTheme` API method to initialize theme for `ui-components`

815are:
There can be issues after - https://github.com/SAP/open-ux-tools/pull/3060
We need call `initTheme` API method to initialize theme for `ui-components` with correct vscode variables


voicis (approved):
- Code changes looks good
- Good test coverage
- Did not test manually

Issue #3066, opened 2025-03-31
BUG - destination URL returns 404 to due lower case service path


cianmSAP:
### Related Feature
Feature request: _issue number_

### Description
Issues have arisen where backends are returning 404 due to the service path being updated to use lower case

### Steps to Reproduce
Use a partial or full service URL which relies on a backend in which the case will not be resolved.

### Expected results
The service path remains the same case as it is entered

### Actual results
Is it updated to be lower case

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3067, opened 2025-03-31
fix(odata-inq): prevent changing service path casing

cianmSAP:
#3066 

Change only lowers the case of the destination name within the URL and leave the service path as it is. 
Some backends return a 404 for the service path as lower case

### DIFF
--- a/packages/odata-service-inquirer/src/prompts/connectionValidator.ts
+++ b/packages/odata-service-inquirer/src/prompts/connectionValidator.ts
@@ -601,7 +601,7 @@ export class ConnectionValidator {
         this.resetConnectionState();
         this.resetValidity();
         // Get the destination URL in the BAS specific form <protocol>://<destinationName>.dest. This function lowercases the origin.
-        const destUrl = getDestinationUrlForAppStudio(destination.Name, servicePath).toLowerCase();
+        const destUrl = getDestinationUrlForAppStudio(destination.Name.toLowerCase(), servicePath);
         // Get the destination URL in the portable form <protocol>://<host>:<port>.
         // We remove trailing slashes (up to 10, infinite would allow DOS attack) from the host to avoid double slashes when appending the service path.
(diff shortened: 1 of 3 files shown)


packages/odata-service-inquirer/src/prompts/connectionValidator.ts
```diff
@@ -601,7 +601,7 @@ export class ConnectionValidator {
         this.resetConnectionState();
         this.resetValidity();
         // Get the destination URL in the BAS specific form <protocol>://<destinationName>.dest. This function lowercases the origin.
-        const destUrl = getDestinationUrlForAppStudio(destination.Name, servicePath).toLowerCase();
+        const destUrl = getDestinationUrlForAppStudio(destination.Name.toLowerCase(), servicePath);
```
IainSAP, line 604: This is deliberate lower casing of the host. Other issues will arise.

IainSAP (changes requested):
'

packages/odata-service-inquirer/src/prompts/connectionValidator.ts
IainSAP: Looks like the path case is respected by `URL` :
```
new URL('http://AbCd:999/UpperCase')
URL {origin: 'http://abcd:999', protocol: 'http:', username: '', password: '', host: 'abcd:999', …}
new URL('http://AbCd:999/UpperCase').toString()
'http://abcd:999/UpperCase'
```

So should be ok.

IainSAP: Maybe add a comment so its clear whats happening

IainSAP (approved):
Thanks @cianmSAP 

- Update seems appropriate
- Test updated to cover

longieirl (approved):
Changeset is good.
Tests are appropriate.

PR #3068, opened 2025-04-01
fix(abap-inq): init transport config for scp systems

cianmSAP:
Current implementation does not initiliase the abap service provider for SCP systems at selection,
but only created if they are needed later in the flow - for packages, and transports

However with the change to validate packages (if they are local or not) coming in regardless if the package is entered manually, it makes sense to start initialising the transport when we do the init for on prem

This prevents multiple browsers opening for BTP systems (for every character entered in package manual) as the service provider will already be in the state


longieirl (approved):
Changeset is good.
Tests to cover change.
Description describes the issue.

PR #3075, opened 2025-04-02
feat: Integration of ExtendControllerPlugin in Adaptation Editor

mmilko01:
#3117

* Extends @sap-ux/adp-tooling to handle `codeExt` changes
* Adds integration of the new ControllerExtensionPlugin in @sap-ux-private/preview-middleware-client


packages/preview-middleware-client/src/adp/extend-controller.ts
```diff
@@ -0,0 +1,49 @@
...
+import CommandFactory from 'sap/ui/rta/command/CommandFactory';
+import { Deferred, createDeferred } from './utils';
+import { DialogFactory, DialogNames } from './dialog-factory';
+import ExtendController from 'sap/ui/rta/plugin/ExtendControllerPlugin';
+
+export interface ExtenControllerData {
+    deferred: Deferred<DeferredExtendControllerData>;
+}
+
+export type DeferredExtendControllerData = {
```
nikmace, line 12: The name seems odd, would it better for the name to be: `ControllerExtModifiedValue`, `ExtendControllerRef` or something else of that sort? Maybe the same way the plugin calls this variable after the promise is resolved?

packages/preview-middleware-client/src/adp/extension-point.ts
```diff
@@ -84,9 +84,9 @@ export default class ExtensionPointService {
                 await this.fragmentHandler(overlay, info)
         });
 
-        const defaultPlugins = this.rta.getDefaultPlugins();
-        defaultPlugins.addXMLAtExtensionPoint = plugin;
-        this.rta.setPlugins(defaultPlugins);
+        const plugins = this.rta.getPlugins();
+        plugins.addXMLAtExtensionPoint = plugin;
+        this.rta.setPlugins(plugins);
```
nikmace, line 89: Q. Why is this change needed? Does `getDefaultPlugins` return incorrect results or not all of the results? I see below that you set default plugins initially, but aren't they set by default already, so when you can `getPlugins` you will get all of the plugins, so it seems redundant, or am I missing something that RTA colleagues have provided as best practice?

packages/preview-middleware-client/src/adp/init.ts
```diff
@@ -34,6 +39,8 @@ export default async function (rta: RuntimeAuthoring) {
     const applicationType = getApplicationType(rta.getRootControlInstance().getManifest());
     const quickActionRegistries = await loadDefinitions(applicationType);
 
+    
+
```
nikmace, line 43: Empty rows, some of the files in the client folder are not prettier'd, could you please check?

packages/preview-middleware-client/test/__mock__/sap/ui/rta/plugin/ExtendControllerPlugin.ts
```diff
@@ -0,0 +1,5 @@
+export default class ExportControllerPlugin{
```
nikmace, line 1: `ExportControllerPlugin` -> `ExtendControllerPlugin`

packages/preview-middleware-client/test/unit/adp/extend-controller.test.ts
```diff
@@ -0,0 +1,62 @@
+import { initExtendControllerPlugin } from '../../../src/adp/extend-controller';
+import type RuntimeAuthoring from 'sap/ui/rta/RuntimeAuthoring';
+import type UI5Element from 'sap/ui/core/Element';
+import CommandFactory from 'sap/ui/rta/command/CommandFactory';
+import ExtendControllerPlugin from 'sap/ui/rta/plugin/ExtendControllerPlugin';
+import { DialogFactory, DialogNames } from '../../../src/adp/dialog-factory';
+import { createDeferred } from '../../../src/adp/utils';
+
+jest.mock('sap/ui/rta/command/CommandFactory');
+jest.mock('sap/ui/rta/plugin/ExtendControllerPlugin');
```
nikmace, line 10: We are already mocking `CommandFactory` and `ExtendControllerPlugin`, you can import the needed mocks from the `__mock__` folder.

packages/preview-middleware-client/test/unit/adp/extend-controller.test.ts
```diff
@@ -0,0 +1,62 @@
...
+describe('AddFragmentService', () => {
+    let mockRta: jest.Mocked<RuntimeAuthoring>;
+    let mockOverlay: jest.Mocked<UI5Element>;
+
+    beforeEach(() => {
+        mockRta = {
+            getFlexSettings: jest.fn().mockReturnValue({}),
+            getPlugins: jest.fn().mockReturnValue({}),
+            setPlugins: jest.fn()
+        } as unknown as jest.Mocked<RuntimeAuthoring>;
```
nikmace, line 25: The correct way of mocking RTA is with the `__mock__` import of `RuntimeAuthoringMock` and instantiating a class then, you can mock each of its methods. Since you are not returning anything in any of those methods you can do this:

```ts
const rtaMock = new RuntimeAuthoringMock({} as RTAOptions);
```
And then pass it to the plugin: 
```ts
rtaMock as unknown as RuntimeAuthoring
```

packages/preview-middleware-client/test/unit/adp/extend-controller.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+            expect(ExtendControllerPlugin).toHaveBeenCalledWith({
+                commandFactory: mockCommandFactory,
+                handlerFunction: expect.any(Function)
+            });
+
+            // Test the handler function
+            const handlerFunction = (ExtendControllerPlugin as jest.Mock).mock.calls[0][0].handlerFunction;
+            const mockDeferred = { promise: Promise.resolve('mockDeferredData') };
+            (createDeferred as jest.Mock).mockReturnValue(mockDeferred);
```
nikmace, line 47: It is good to extract this into its own variable since you are mocking it, this will be more readable and easier to maintain: 

```ts
const createDeferredMock = createDeferred as jest.Mock;
```

packages/preview-middleware-client/test/unit/adp/extend-controller.test.ts
```diff
@@ -0,0 +1,62 @@
...
+            (DialogFactory.createDialog as jest.Mock).mockResolvedValue(undefined);
```
nikmace, line 48: Maybe it is better to use a spy here and not cast it as a mock?

packages/preview-middleware-client/types/sap.ui.rta.d.ts
```diff
@@ -80,6 +80,20 @@ declare module 'sap/ui/rta/plugin/AddXMLAtExtensionPoint' {
...
+
+    interface Arguments {
+        commandFactory: CommandFactory;
+        handlerFunction: (overlay: UI5Element, extensionPointInfo: uknown) => Promise<void | object>;
+    }
+
+    export default class AddXMLAtExtensionPoint {
+        constructor(_: Arguments) {}
+    }
+}
```
nikmace, line 94: This does not make sense, you are declaring an `ExtendControllerPlugin` plugin and then you are configuring `AddXMLAtExtensionPoint` class.

packages/preview-middleware-client/types/sap.ui.rta.d.ts
```diff
@@ -277,6 +291,7 @@ declare module 'sap/ui/rta/RuntimeAuthoring' {
         getService: <T>(name: 'outline' | 'controllerExtension' | string) => Promise<T>;
         getSelection: () => ElementOverlay[];
         getDefaultPlugins: () => { [key: string]: uknown; contextMenu: ContextMenu };
+        getPlugins: () => { [key: string]: uknown; contextMenu: ContextMenu };
```
nikmace, line 294: `uknown` -> `unknown`

nikmace (changes requested):

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
```diff
@@ -46,11 +47,14 @@ type ControllerModel = JSONModel & {
  * @namespace open.ux.preview.client.adp.controllers
  */
 export default class ControllerExtension extends BaseDialog<ControllerModel> {
-    constructor(name: string, overlays: UI5Element, rta: RuntimeAuthoring, telemetryData?: QuickActionTelemetryData) {
+    public readonly data?: ExtenControllerData;
```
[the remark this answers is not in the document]
voicis: I'd also prefer using full words. We don't really need to save characters in type names and it is clearer.

packages/preview-middleware-client/src/adp/extension-point.ts
mmilko01: This was discussed with RTA colleagues. There was an issue with using `getDefaultPlugins` as that list always stays the same. When we need to add multiple plugins (e.g. addXMLPlugin, ExtendControllerPlugin, etc), we override the list each time making the previously added plugins to disappear. This is why I initially get the default plugins and set the in the plugins list, so we could mutate that list later in the flow.

lfindlaysap (approved):

packages/preview-middleware-client/src/adp/extend-controller.ts
mmilko01: I used the same naming for the type as used in `extension-point.ts`. Do you think it should be changed in both cases as I think it is best to have it the same? https://github.com/SAP/open-ux-tools/blob/4b73d5ff8ee692902ddd5090f9b692d259b137f9/packages/preview-middleware-client/src/adp/extension-point.ts#L18

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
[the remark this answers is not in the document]
mmilko01: This was due to a typo.
Fixed in: ca785d8d4824f366020f429f21ac0a50db00b26a

packages/adp-tooling/src/preview/change-handler.ts
```diff
@@ -234,12 +244,49 @@ export function addXmlFragment(basePath: string, change: AddXMLChange, fs: Edito
...
+    basePath: string,
+    change: CodeExtChange,
+    fs: Editor,
+    logger: Logger
+): Promise<void> {
+    const { codeRef } = change.content;
+    const isTsSupported = isTypescriptSupported(rootPath, fs);
+    const fileExtension = extname(codeRef);
+    const fileName = basename(codeRef, fileExtension);
+    const fullName = `${basename(codeRef, fileExtension)}.${isTsSupported ? 'ts' : 'js'}`;
```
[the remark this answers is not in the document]
mmilko01: Fixed in: ca785d8d4824f366020f429f21ac0a50db00b26a

packages/preview-middleware-client/test/unit/adp/extend-controller.test.ts
mmilko01: Fixed in: ca785d8d4824f366020f429f21ac0a50db00b26a

packages/preview-middleware-client/test/unit/adp/extend-controller.test.ts
mmilko01: Fixed in: ca785d8d4824f366020f429f21ac0a50db00b26a

packages/preview-middleware-client/test/unit/adp/extend-controller.test.ts
mmilko01: Fixed in: ca785d8d4824f366020f429f21ac0a50db00b26a

packages/preview-middleware-client/types/sap.ui.rta.d.ts
mmilko01: Copy/paste error.
Fixed in: ca785d8d4824f366020f429f21ac0a50db00b26a

packages/preview-middleware-client/src/adp/utils.ts
```diff
@@ -60,19 +62,55 @@ export function createDeferred<T>(): Deferred<T> {
...
+
+    function getNestedProperty(obj: ChangeDefinition, path: string): string | undefined {
+        return path.split('.').reduce((acc: unknown, key) => {
+            return (acc as Record<string, unknown>)?.[key];
+        }, obj) as string | undefined;
+    }
+
+    const nestedProperty = getNestedProperty(change, propertyPath);
+    return typeof nestedProperty === 'string' ? nestedProperty.includes(propertyValue) : false;
 }
```
nikmace, line 114: 1. (Optional) Not sure if we need some safety here if a different command is passed somehow and does not have a `getPreparedChange`:
```ts
const change = command.getPreparedChange()?.getDefinition?.();
if (!change) return false;
```

2. Extract `getNestedProperty` outside the function as it's a generic utility and could be reused elsewhere maybe (and is cluttering your main function a bit). Also, you can make the object property type (`ChangeDefinition`) more generic.

packages/preview-middleware-client/src/messagebundle.properties
```diff
@@ -37,9 +37,13 @@ ADP_ADD_SUB_PAGE_DIALOG_PAGE_TYPE_LABEL = Page Type
 ADP_ADD_SUB_PAGE_DIALOG_NAVIGATION_LABEL = Navigation
 ADP_ADD_FRAGMENT_NOTIFICATION = Note: The `{0}.fragment.xml` fragment will be created once you save the change.
 ADP_ADD_FRAGMENT_WITH_TEMPLATE_NOTIFICATION = Note: The `{0}.fragment.xml` fragment will be created once you save the changes.
+ADP_CONTROLLER_EXTENSION_EXISTS = An existing controller extension has been found. Please use the previously created extension controller
```
nikmace, line 40: A dot at the end of the sentence is missing as I see the ones below have it at the end.

nikmace (changes requested):

packages/preview-middleware-client/src/adp/utils.ts
```diff
@@ -60,19 +62,55 @@ export function createDeferred<T>(): Deferred<T> {
...
         return false;
     }
-    const change = command.getPreparedChange().getDefinition() as unknown as FragmentChange;
-    return change.content?.fragmentPath?.includes(`${fragmentName}.fragment.xml`) || false;
+    const change = command.getPreparedChange().getDefinition();
+
+    function getNestedProperty(obj: ChangeDefinition, path: string): string | undefined {
+        return path.split('.').reduce((acc: unknown, key) => {
+            return (acc as Record<string, unknown>)?.[key];
+        }, obj) as string | undefined;
```
voicis, line 109: This is not a safe assertion, because  it could also be number or boolean.

packages/preview-middleware-client/src/adp/utils.ts
```diff
@@ -60,19 +62,55 @@ export function createDeferred<T>(): Deferred<T> {
...
-export function matchesFragmentName(command: FlexCommand, fragmentName: string): boolean {
+export function matchesChangeProperty(command: FlexCommand, propertyPath: string, propertyValue: string): boolean {
     if (typeof command.getPreparedChange !== 'function') {
         return false;
     }
-    const change = command.getPreparedChange().getDefinition() as unknown as FragmentChange;
-    return change.content?.fragmentPath?.includes(`${fragmentName}.fragment.xml`) || false;
+    const change = command.getPreparedChange().getDefinition();
+
+    function getNestedProperty(obj: ChangeDefinition, path: string): string | undefined {
```
voicis, line 106: I don't see any dependencies to the `matchesChangeProperty` scope, so it could also be defined outside.

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
```diff
@@ -183,8 +237,10 @@ export default class ControllerExtension extends BaseDialog<ControllerModel> {
 
         const content = this.dialog.getContent();
 
-        const form = content[0] as SimpleForm;
-        form.setVisible(false);
+        const form0 = content[0] as SimpleForm;
+        form0.setVisible(false);
```
voicis, line 241: We shouldn't manipulate the view directly from the controller, model should be used instead.

packages/preview-middleware-client/src/adp/utils.ts
mmilko01: Fixed in: 42a8fbb6175c18568171aff4baf1da16e9623027

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
mmilko01: Fixed in: 42a8fbb6175c18568171aff4baf1da16e9623027

packages/preview-middleware-client/src/adp/utils.ts
mmilko01: Fixed in: 42a8fbb6175c18568171aff4baf1da16e9623027

packages/preview-middleware-client/src/adp/utils.ts
mmilko01: Fixed in: d3a3bde43ff3218390c93174b0d45af83a768e81

nikmace (changes requested):
There are also two comments from the previous review and two Sonar issues, one of which is a circular dependency.

packages/preview-middleware-client/src/adp/init.ts
```diff
@@ -28,7 +28,12 @@ export default async function (rta: RuntimeAuthoring) {
     const defaultPlugins = rta.getDefaultPlugins();
     rta.setPlugins(defaultPlugins);
 
-    await initDialogs(rta, syncViewsIds, ui5VersionInfo);
+    if (isLowerThanMinimalUi5Version(ui5VersionInfo, { major: 1, minor: 136 })) {
+        await initDialogs(rta, syncViewsIds, ui5VersionInfo);
+    } else {
+        (await import('open/ux/preview/client/adp/add-fragment')).initAddXMLPlugin(rta);
+        (await import('open/ux/preview/client/adp/extend-controller')).initExtendControllerPlugin(rta);
```
voicis, line 35: Now we have quite a few dependent requests that has to execute one after another, but we can fix it separately as this is not directly related to this PR.

packages/preview-middleware-client/src/adp/utils.ts
```diff
@@ -60,19 +61,64 @@ export function createDeferred<T>(): Deferred<T> {
...
+ */
+export function checkForExistingChange(
+    rta: RuntimeAuthoring,
+    commandName: string,
+    propertyPath: string,
+    propertyValue: string
+): boolean {
+    const allCommands = rta.getCommandStack().getCommands();
+
+    return allCommands.some((command: FlexCommand) => {
```
voicis, line 80: This does not seem to work correctly, because it does not take into account undo/redo (after undo it still thinks that change exists, even though when no change will be saved).

packages/preview-middleware-client/types/sap.ui.core.d.ts
```diff
@@ -11,3 +11,17 @@ declare module 'sap/ui/core/util/reflection/JsControlTreeModifier' {
     const JsControlTreeModifier: JsControlTreeModifier;
     export default JsControlTreeModifier;
 }
+
+declare module 'sap/ui/core/mvc/XMLView' {
```
voicis, line 15: Maybe we should just update the `@sapui5/types` version?

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
```diff
@@ -95,6 +111,13 @@ export default class ControllerExtension extends BaseDialog<ControllerModel> {
 
         const fileExists = controllerList.some((f) => f.controllerName === `${controllerName}.js`);
 
+        const pendingChangeExists = checkForExistingChange(
+            this.rta,
+            'codeExt',
+            'content.codeRef',
+            `${controllerName}.js`
```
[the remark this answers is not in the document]
mmilko01: Yes, in the `.change` files we always have `.js` extension for the controllers as it is meant to be applied at runtime where only the JS files are used.

packages/preview-middleware-client/src/adp/init.ts
mmilko01: You are correct, but I think this can be handled separately.

packages/preview-middleware-client/types/sap.ui.core.d.ts
mmilko01: I tried bumping the version to the latest 1.135, as it contains this `getControllerModuleName`, but it caused multiple issues during the build and it will require some refactoring to fix. Do you think that this needs to be done in this PR?

packages/preview-middleware-client/test/__mock__/sap/ui/rta/plugin/ExtendControllerPlugin.ts
mmilko01: Fixed in: 8e91d1e2a1b998d31dfbc163918b8e2f8d2f48cd

packages/preview-middleware-client/test/unit/adp/extend-controller.test.ts
mmilko01: Fixed in: 8e91d1e2a1b998d31dfbc163918b8e2f8d2f48cd

packages/preview-middleware-client/test/unit/adp/utils.test.ts
```diff
@@ -60,31 +67,148 @@ describe('utils', () => {
...
+                                })
+                            })
+
+                        }
+                    ])
+                })
+            }
+            const result = checkForExistingChange(mockRta as unknown as RuntimeAuthoring, 'addXML', 'content.fragmentPath', 'testFragment.fragment.xml');
+            expect(result).toBe(true);
+        });
```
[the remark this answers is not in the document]
mmilko01: I addressed the two comments. The Sonar issues were accepted. And I changed the mock as suggested.
Changed in: ddde85bb568fac2f261572fb38bc6c8b5667df12

packages/preview-middleware-client/src/adp/utils.ts
mmilko01: As discussed I created a new issue regarding this, as it was pre-existing - https://github.com/SAP/open-ux-tools/issues/3211

voicis (dismissed):
- Code changes looks good
- Review comments addressed
- Good test coverage
- Tested manually with the latest UI5 version

nikmace (dismissed):
Changes look good
Coverage is good
Did not test manually

packages/preview-middleware-client/src/messagebundle.properties
```diff
@@ -39,9 +39,13 @@ ADP_ADD_SUB_PAGE_DIALOG_PAGE_TYPE_LABEL = Page Type
...
 ADP_ADD_FRAGMENT_WITH_TEMPLATE_NOTIFICATION = Note: The `{0}.fragment.xml` fragment will be created once you save the changes.
+ADP_CONTROLLER_EXTENSION_EXISTS = An existing controller extension has been found. Please use the previously created extension controller.
+ADP_CONTROLLER_PENDING_CHANGE_EXISTS = A pending change for controller extension has been found. Please use the previously created extension controller after saving the change or delete the pending change.
+ADP_CREATE_CONTROLLER_EXTENSION = Note: The `{0}` controller extension will be created once you save the change.
 ADP_ADD_TWO_FRAGMENTS_WITH_TEMPLATE_NOTIFICATION = Note: The `{0}.fragment.xml` and `{1}.fragment.xml` fragments will be created once you save the changes.
 ADP_SYNC_VIEWS_MESSAGE = Synchronous views are detected for this application. Controller extensions are not supported for such views and will be disabled.
 ADP_QUICK_ACTION_DIALOG_OPEN_MESSAGE = This action is disabled because a dialog is already open.
+ADP_QUICK_ACTION_CONTROLLER_PENDING_CHANGE_EXISTS =This action is disabled because a pending change for controller extension has been found.
```
lfindlaysap, line 48:
```suggestion
ADP_QUICK_ACTION_CONTROLLER_PENDING_CHANGE_EXISTS =This action is disabled because a pending change for a controller extension has been found. 
```

lfindlaysap (changes requested):
@mmilko01, one small change.

packages/preview-middleware-client/src/messagebundle.properties
mmilko01: Fixed with: 5d3cc4d9eba1b0a669caa841fd5a95645058587e

lfindlaysap (approved):

nikmace (approved):
Changes look good
Changeset looks good
Coverage is good
Did not test manually

Issue #3078, opened 2025-04-02
BUG - Cannot create a Fragment and Controller Extension when clicking create button in the dialog


nikmace:
### Related Feature
Feature request: #3030 

### Description
When invoking an asynchronous base class method (using `await super.onCreateBtnPressHandler()`) inside our `onCreateBtnPress` handler, the original event object loses its context (i.e. properties such as source, parameters, etc.). This causes our code to behave unexpectedly when trying to access event properties after the asynchronous call.

### Steps to Reproduce
1. Open Adaptation Editor with UI5 Version 1.124.11 or lower
2. Try to Add Fragment or create Controller Extension
3. Click Create button
4. See error in the console or while digging the controller

### Expected results
The event object should retain its properties after the asynchronous call, or at least, the necessary data should be captured before the asynchronous operation begins so that subsequent operations can use the captured data reliably.

### Actual results
After the await call to the base class method, the event object's context is lost. This is likely due to the event pooling mechanism in the framework (e.g., SAP UI5), where event objects are reused for performance reasons. As a result, any attempt to access the event properties after the asynchronous call fails or returns undefined.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3079, opened 2025-04-02
fix: Cannot create a Fragment and Controller Extension when clicking create button in the dialog

nikmace:
Fix for #3078.
- Changes the order of execution inside the handler to capture the event before calling the base class.


IvoSG (approved):
- Simple and easy to understand change.
- Not tested manually.

Jimmy-Joseph19 (approved):
Thanks for the fix. Tested manually and issue is fixed.

PR #3081, opened 2025-04-02
fix: CPE Add Subpage quick action bug fix in v4 projects

vadson71:
Fixed CPE Add Subpage quick action bug. The action is not shown when current page has context path defined in page settings in manifest


voicis (approved):
- Code changes looks good
- Good test coverage
- Did not test manually

Jimmy-Joseph19 (approved):
Changes looks fine.
Tested locally and fix works
Changeset found.
Covered with unit test

PR #3083, opened 2025-04-03
fix: Cannot create a Extension Point fragment when clicking create button in the dialog

nikmace:
Fix for #3078.
- Changes the order of execution inside the handler to capture the event before calling the base class.


IvoSG (approved):
- Simple and easy to understand change.
- Not tested manually.

marufrasully (approved):
LGTM - did not test it locally.

PR #3085, opened 2025-04-04
fix: Generic Annoation file missing explicit record type for flattened record

Jimmy-Joseph19:
While converting flattened structire to generic annotation file format the explicit type information missing in final annotation file


packages/cds-odata-annotation-converter/src/transforms/annotation/handlers/annotation.ts
```diff
@@ -1,20 +1,28 @@
...
 import type { VisitorState } from '../visitor-state';
 import { createQualifierAttribute, createTermAttribute } from '../creators';
+import { createRange } from '../range';
 
 export const annotationHandler: NodeHandler<Annotation> = {
     type: ANNOTATION_TYPE,
     convert,
     getChildren(state: VisitorState, annotation: Annotation): AnnotationNode[] {
-        if (annotation.value) {
+        if (annotation.value && !state.context.isFlattened$Type) {
```
voicis, line 25: This logic should not relay on context, but annotation value itself. Context is something that is coming from from nodes higher up in the tree. If we want to avoid recomputing some data between `convert` and `getChildren` calls we could think of some other mechanism. However, currently I don't think checking if the the annotation term contains $Type again would be expensive.

packages/cds-odata-annotation-converter/src/transforms/annotation/handlers/annotation.ts
```diff
@@ -161,8 +169,24 @@ function handleFlattenedStructure(state: VisitorState, annotation: Annotation, e
...
+                [Edm.Type]: createAttributeNode(Edm.Type, value.value, flattenedSegments[0].range, value.range)
+            }
+        });
+        element.content.push(recordWithType);
+        const currentState = state.popContext();
+        state.pushContext({
+            ...currentState,
+            isFlattened$Type: true
+        });
+    } else if (flattenedSegments.length) {
```
voicis, line 189: This logic should be done in `convertFlattenedPath` function, because I suspect we would have a similar issue with cases where flattened syntax starts from a record property.

packages/cds-odata-annotation-converter/src/transforms/annotation/visitor-state.ts
```diff
@@ -7,6 +7,7 @@ export interface Context {
      */
     recordType?: string;
     termType?: string;
+    isFlattened$Type?: boolean;
```
voicis, line 10: This should not be needed, because $Type is a leaf node and we don't need to pass context deeper.

packages/cds-odata-annotation-converter/src/transforms/annotation/handlers/annotation.ts
Jimmy-Joseph19: adapted

packages/cds-odata-annotation-converter/src/transforms/annotation/handlers/annotation.ts
Jimmy-Joseph19: adapted

packages/cds-odata-annotation-converter/src/transforms/annotation/visitor-state.ts
Jimmy-Joseph19: removed

packages/cds-odata-annotation-converter/src/transforms/annotation/flattened.ts
```diff
@@ -46,18 +47,29 @@ export function convertFlattenedPath(
...
-            parent.contentRange = record.range ?? undefined;
-            parent.content.push(record);
+            if (expandedStructure.kind === 'record-type') {
+                if (expandedStructure.element.attributes['Type'].nameRange?.start && parent.contentRange?.end) {
+                    const range: Range = {
+                        start: expandedStructure.element.attributes['Type'].nameRange?.start,
+                        end: parent.contentRange?.end
+                    };
+                    parent.contentRange = copyRange(range);
+                }
```
voicis, line 60: We already calculate the correct range when creating record, we could just reuse that.

```suggestion
                if (expandedStructure.element.range) {
                    parent.contentRange = copyRange(expandedStructure.element.range);
                }
```

packages/cds-odata-annotation-converter/src/transforms/annotation/flattened.ts
```diff
@@ -144,30 +159,48 @@ function convertToExpandedStructure(
...
+            const flatProperty: Element = createElementNode({
+                name: Edm.Record,
+                attributes: {
+                    [Edm.Type]: createAttributeNode(Edm.Type, value.value, segment.range, nodeRange(value, false))
+                },
+                range: propertyRange
+            });
+            if (expandedStructure.length > 0) {
+                const last = expandedStructure[i - 1];
+                last.element.contentRange = createRange(segment.range?.start, valueRange?.end);
```
voicis, line 172: We already do this in line 59, I don't think we need to do it here as well.

packages/cds-odata-annotation-converter/test/data/parser/flattened-nested-record/assignment.txt
```diff
@@ -0,0 +1,4 @@
+UI : {
+    HeaderInfo.Title.$Type : 'UI.DataFieldForAnnotation',
+    HeaderInfo.Title.Target : '@Communication.Contact'
+}
```
voicis, line 4: We should have a bit more test cases here:
1. where we have also the $Type directly on annotation (valid)
2. where it does not have a string value (invalid)
3. where $Type is in the middle of a flattened structure e.g `HeaderInfo.$Type.Title` (invalid)

Maybe you can think of some other scenarios which could have issues.

packages/cds-odata-annotation-converter/src/transforms/annotation/flattened.ts
```diff
@@ -144,30 +159,48 @@ function convertToExpandedStructure(
...
             continue;
         }
-
-        const flatProperty: Element = createElementNode({
-            name: Edm.PropertyValue,
-            range: propertyRange,
-            contentRange: propertyRange,
-            attributes: {
-                [Edm.Property]: createPropertyAttribute(segment.value, segment.range)
+        if (segment.value === ReservedProperties.Type && value?.type === STRING_LITERAL_TYPE) {
```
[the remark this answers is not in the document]
Jimmy-Joseph19: adapted

packages/cds-odata-annotation-converter/src/transforms/annotation/flattened.ts
Jimmy-Joseph19: adapted

packages/cds-odata-annotation-converter/src/transforms/annotation/flattened.ts
Jimmy-Joseph19: removed

packages/cds-odata-annotation-converter/test/data/parser/flattened-nested-record/assignment.txt
Jimmy-Joseph19: added

packages/cds-odata-annotation-converter/src/i18n/i18n.json
```diff
@@ -4,5 +4,8 @@
   "Wrong_element_0_Did_you_mean_1": "`{{currentValue}}` is incorrect. Did you mean `{{proposedValue}}`?",
   "Mandatory_property_not_provided_0": "The mandatory `{{property}}` property was not provided.",
   "Path_escaping_brackets_should_be_used": "Path escaping brackets `![]` must be used.",
-  "Wrong_path_separator_0_Did_you_mean_1": "`{{currentValue}}` is the wrong path separator. Did you mean `{{proposedValue}}`?"
+  "The_name_attribute_value_must_be_a_string_literal": "The `{{name}}` attribute value must be a string literal.",
+  "Wrong_path_separator_0_Did_you_mean_1": "`{{currentValue}}` is the wrong path separator. Did you mean `{{proposedValue}}`?",
+  "Segment_is_not_allowed_here": "Segment(s) not allowed after `$Type`",
```
lfindlaysap, line 9: We shouldn't use (s). Can we have a separate string for the plural?

lfindlaysap (commented):
@Jimmy-Joseph19, one comment from me.

packages/cds-odata-annotation-converter/src/i18n/i18n.json
Jimmy-Joseph19: Done

packages/cds-odata-annotation-converter/src/i18n/i18n.json
```diff
@@ -4,5 +4,9 @@
   "Wrong_element_0_Did_you_mean_1": "`{{currentValue}}` is incorrect. Did you mean `{{proposedValue}}`?",
   "Mandatory_property_not_provided_0": "The mandatory `{{property}}` property was not provided.",
   "Path_escaping_brackets_should_be_used": "Path escaping brackets `![]` must be used.",
-  "Wrong_path_separator_0_Did_you_mean_1": "`{{currentValue}}` is the wrong path separator. Did you mean `{{proposedValue}}`?"
+  "The_name_attribute_value_must_be_a_string_literal": "The `{{name}}` attribute value must be a string literal.",
+  "Wrong_path_separator_0_Did_you_mean_1": "`{{currentValue}}` is the wrong path separator. Did you mean `{{proposedValue}}`?",
+  "Segment_is_not_allowed_here": "Segment not allowed after `$Type`",
+  "Segments_are_not_allowed_here": "Segments not allowed after `$Type`",
```
lfindlaysap, line 10:
```suggestion
  "Segments_are_not_allowed_here": "Segments not allowed after `$Type`.",
```

packages/cds-odata-annotation-converter/test/data/parser/flattened-nested-record/generic.json
```diff
@@ -0,0 +1,336 @@
...
+                "attributes": {},
+                "contentRange": "[(6,30)..(6,34)]",
+                "content": [
+                    {
+                        "type": "text",
+                        "range": "[(6,30)..(6,34)]",
+                        "text": "test"
+                    }
+                ]
+            }
```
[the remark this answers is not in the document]
Jimmy-Joseph19: removed

packages/cds-odata-annotation-converter/test/data/parser/flattened-nested-record/generic.json
```diff
@@ -0,0 +1,336 @@
...
+                "attributes": {},
+                "contentRange": "[(8,29)..(8,51)]",
+                "content": [
+                    {
+                        "type": "text",
+                        "range": "[(8,29)..(8,51)]",
+                        "text": "UI.ChartDefinitionType"
+                    }
+                ]
+            }
```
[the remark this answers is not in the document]
Jimmy-Joseph19: removed

packages/cds-odata-annotation-converter/src/transforms/annotation/handlers/annotation.ts
```diff
@@ -20,7 +20,7 @@ export const annotationHandler: NodeHandler<Annotation> = {
     type: ANNOTATION_TYPE,
     convert,
     getChildren(state: VisitorState, annotation: Annotation): AnnotationNode[] {
-        if (!annotation.term.value.endsWith(ReservedProperties.Type) && annotation.value) {
+        if (!annotation.term.value.includes(ReservedProperties.Type) && annotation.value) {
```
voicis, line 23: This would also include false positive case like `Chart.My$Type`. We should check if there are any segments with the .`$Type`. Please also add a case for this.

voicis (approved):
- Code changes looks good
- Review comments addressed
- Good test coverage
- Did not test manually

PR #3087, opened 2025-04-04
fix(abap-inq): check cloud systems are connected before making backend call

cianmSAP:
Add check that connection has been made for cloud systems before validating package (otherwise validator will run for every character entered, which in turns means a browser window opens for every character opened)


longieirl (approved):
Changeset is good
Tests to cover change.
Havent tested locally.
No issues found.

kjose90 (approved):
Thanks, changes are clear, tests and changeset ✅

PR #3099, opened 2025-04-09
fix: ADP Generator integration test mocking

nikmace:
Fix for #3015.
- Fixes the mocking of package.json retrieval and the snapshot having dynamic value instead of mocked.


mmilko01 (approved):
Code looks good
Code coverage is good
Changeset OK
Did not test manually

GDamyanov (approved):
- Changeset present.
- Code looks good.
- Excellent test coverage.

Issue #3063, opened 2025-03-31
TBI - Expose new UI `App Frontend Service` as a new router module during Cloud Foundry deployment generation


longieirl:
As a Fiori developer, I'd like to be able to use AppFront as a deployment option when deploying to Cloud Foundry.

Description
AppFront is a simplified CF deployment method that is launching in 2025. The cf inquirer module needs to be updated to support this deployment option.

New deployment Options:
```
None
Add application to managed application router
Add application to Application Frontend service
```
If the user chooses the new application frontend service option, provide the following warning to the user;

```
You need to have the Application FrontEnd service enabled for your subaccount.
```

This warning notification is required as it will block their CF deployment.

Under the prompt `Add router module`, a new option `Application Frontend Service` will be appended.

Acceptance Criteria
```
Given I choose to add deployment configuration to my Fiori project, either during generation or afterwards, and I have the feature toggle enabled, if I choose CF as my deploy target, and I don't already have an MTA.yaml file defined, then I will be offered the following 3 deployment options:

None
Add application to managed application router
Add application to Application Frontend service
```

### Value
Support new router configuration

### Tasks
- [x] User will now have three options when they select `Cloud Foundry` as their target system
- [x] Should be hidden behind a toggle until feature is released
- [x] Update unit tests

longieirl:
Toggle is not required if all being merged into the same release.

longieirl:
Validated on tools-suite and is being pushed for release.

PR #3104, opened 2025-04-09
fix(cf-deploy-config-inq): add support for toggle

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/3063

- enable toggle to disable App Frontend service


longieirl:
Not required for now.

PR #3107, opened 2025-04-09
fix: add telemetry tracking for context menu

KamalKumarSardiwal:


packages/preview-middleware-client/src/cpe/context-menu-service.ts
```diff
@@ -33,8 +37,20 @@ export class ContextMenuService {
...
+                    const controlName = getControlById(controlId)?.getMetadata().getName();
+                    const versionInfo = await getUi5Version();
+                    await reportTelemetry({
+                        category: 'OutlineContextMenu',
+                        actionName,
+                        controlName,
+                        ui5Version: `${versionInfo.major}.${versionInfo.minor}.${versionInfo.patch}`
+                    });
+                } catch (err) {
+                    Log.error('Error in reporting Telemetry:', err);
```
voicis, line 52: The message is misleading, the error could also be caused by executing action. If the goal was to avoid telemetry call throwing, then move the action execution out of the try/catch.

KamalKumarSardiwal: Adapted.

voicis (approved):
- Straight forward changes
- Review comments addressed
- Tests are passing
- Did not test manually

vadson71 (approved):
Code changes look good
Good test coverage
Didn't test locally

PR #3110, opened 2025-04-10
Support Downloading ABAP Deployed Fiori App from BSP Repository

kjose90:
- Added new sub generator `@sap-ux/bsp-app-download-sub-generator`
- Added support in `axios-extension` to download base 64 encoded apps.


packages/bsp-app-download-sub-generator/src/app/index.ts
```diff
@@ -0,0 +1,351 @@
...
+            Object.assign(this.answers, answers);
+        }
+    }
+
+    /**
+     *
+     * @param quickDeployedAppConfig - The configuration for the quick deployed app.
+     * @param targetFolder - The target folder where the app will be downloaded.
+     */
+    private async _handleQuickDeployedAppDownload(
```
cianmSAP, line 135: Just trying to understand the flow here. 
So if `quickDeployedAppConfig` has been passed in throught the options, the user is shown a target folder prompt, and then a backend call is made to fetch the app(s). 

What is the likelihood of `appList` not returning an array  from `fetchAppListForSelectedSystem` 

Would it make more sense to fetch the `appList` before showing the prompts - and in that case, the user can select an app (if there is an array of apps possible), and there is none, they can be shown a meaningful error?

packages/bsp-app-download-sub-generator/src/app/index.ts
```diff
@@ -0,0 +1,351 @@
...
+    ): Promise<void> {
+        const appList = await fetchAppListForSelectedSystem(
+            quickDeployedAppConfig.serviceProvider,
+            quickDeployedAppConfig.appId
+        );
+        if (!appList.length) {
+            BspAppDownloadLogger.logger?.error(
+                t('error.quickDeployedAppDownloadErrors.noAppsFound', { appId: quickDeployedAppConfig.appId })
+            );
+            throw new Error();
```
cianmSAP, line 147: I just wonder in practice, if the user has gone through the step of selecting the location for their new app to be saved, at that point I think the assumption would be that the details have been fetched, but it would throw an error after they click `Finish` 

If POs are ok with this, I think the error needs to be more descriptive

packages/bsp-app-download-sub-generator/src/app/index.ts
```diff
@@ -0,0 +1,351 @@
...
+
+    /**
+     * Writes the configuration files for the project, including deployment config, and README.
+     */
+    public async writing(): Promise<void> {
+        if (isValidPromptState(this.answers.targetFolder, this.answers.selectedApp.appId)) {
+            this.projectPath = join(this.answers.targetFolder, this.answers.selectedApp.appId);
+            this.extractedProjectPath = join(this.projectPath, extractedFilePath);
+            // Trigger app download
+            await downloadApp(this.answers.selectedApp.repoName, this.extractedProjectPath, this.fs);
```
cianmSAP, line 162: This doesn't feel correct in the writing phase. Would this not make more sense within the validation of whatever the appropriate prompt is? 
And the user can see a validation error in the UI if something goes wrong?

packages/bsp-app-download-sub-generator/src/app/index.ts
```diff
@@ -0,0 +1,351 @@
...
+            });
+        }
+    }
+
+    /**
+     * Finalises the generator process by creating launch configurations and running post-generation hooks.
+     */
+    async end(): Promise<void> {
+        try {
+            this.appWizard.showInformation(t('info.bspAppDownloadCompleteMsg'), MessageType.notification);
```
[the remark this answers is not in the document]
kjose90: Yeoman does show `Files generated` message. I had a requirement to include a toast message `The selected application has been downloaded and updated to support SAP Fiori tools.`. So I added that in as well? Not sure if there is an option in yeoman where I can default `Files generated` message to display the above custom message instead?

packages/bsp-app-download-sub-generator/src/app/types.ts
```diff
@@ -0,0 +1,140 @@
...
+ * Quick deploy app config are applicable only in scenarios where an application
+ * deployed via ADT Quick Deploy is being downloaded from a BSP repository.
+ */
+export interface QuickDeployedAppConfig {
+    /**  application Id to be downloaded. */
+    appId: string;
+    /**  appUrl is the URL pointing to the application */
+    appUrl?: string;
+    /**  service provider is used to identify the system from which the app is downloaded from. */
+    serviceProvider: AbapServiceProvider;
```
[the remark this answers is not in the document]
kjose90: so there are two distinct flows:

- Command Palette Flow:
In this flow, the user accesses the app download via the command palette. The user is prompted to choose the system, select the apps available in that system, and specify the project path where the app should be downloaded.

- Quick Deployment Flow:
This flow is for users who want to download a quickly deployed ADT app. Here, all the information (appId, appUrl, and service url) is already available via the URI. The same BSP app downloader is used, but the user is only prompted to specify the project path for the download. In this case, the user does not have the option to choose the system or apps.

Given these differences, I thought it would be more logical to group the relevant configurations for the quick deployment flow under `QuickDeployedAppConfig`. What do you think?

packages/bsp-app-download-sub-generator/src/app/index.ts
kjose90: In the Quick Deployment Flow, the `appId` is extracted from the URI and passed to the `fetchAppListForSelectedSystem` function where the `appId` is used to filter deployed apps based on the provided `appId` and the source template.
If the filtered `appId` is not found in the list, there is a problem - and this [message](https://github.com/SAP/open-ux-tools/pull/3110/files#diff-d336408ce992c12dad1e29c20c37f4edad3772882d1cbaa545604399438d9873R41) is displayed. Maybe I can make this message for clear 🤔 
Since the app is essential for proceeding with the download, the process cannot continue without it. 
See [fetchAppListForSelectedSystem](https://github.com/SAP/open-ux-tools/pull/3110/files#diff-60fc3269c4d0049f646962a5984b1933af66437319320f1cc7226a976dc9b50eR78).

packages/bsp-app-download-sub-generator/src/app/index.ts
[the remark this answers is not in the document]
cianmSAP: Ok I understand, so it's showing both? No we can't customise that message AFAIK.

packages/bsp-app-download-sub-generator/src/app/types.ts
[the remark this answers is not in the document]
cianmSAP: Maybe logically these can be grouped, it could be just a naming thing.

packages/bsp-app-download-sub-generator/src/app/index.ts
kjose90: I initially had the `downloadApp` process in the prompt phase for the Command Palette Flow, as the user is prompted to choose the system and apps. In the Quick Deployment Flow, there are no such prompts since the required information is already provided. so  I moved the download process to the writing phase instead. But maybe your right ...

lfindlaysap (dismissed):
Approved. To review these texts in a separate PR as discussed.

devinea (commented):
New module set-up looks good. 
a few small comments.

packages/repo-app-download-sub-generator/README.md
```diff
@@ -0,0 +1,16 @@
+# @sap-ux/repo-app-download-sub-generator
+
+## Features
+
+The SAP App download sub-generator enables users to download a basic LROP App from an ABAP repository. The sub-generator will download the app from the repository and add it to the workspace.
+
+## Installation
+
+The SAP App download sub-generator  sub-generator is installed as part of the [@sap/generator-fiori](https://www.npmjs.com/package/@sap/generator-fiori) generator and cannot be used stand alone.
```
cianmSAP, line 9: `sub-generator` is repeated here

packages/repo-app-download-sub-generator/package.json
```diff
@@ -0,0 +1,85 @@
...
+        "@sap-ux/abap-deploy-config-writer": "workspace:*",
+        "@sap-ux/btp-utils": "workspace:*",
+        "@sap-ux/ui5-info": "workspace:*",
+        "@sap-ux/axios-extension": "workspace:*",
+        "@sap-ux/store": "workspace:*",
+        "adm-zip": "0.5.10",
+        "i18next": "23.5.1",
+        "inquirer": "8.2.6",
+        "yeoman-generator": "5.10.0",
+        "inquirer-autocomplete-prompt": "2.0.1"
```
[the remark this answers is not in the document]
kjose90: removed `@sap-ux/i18n` and moved `inquirer-autocomplete-prompt` to dev dependency

packages/bsp-app-download-sub-generator/README.md
```diff
@@ -0,0 +1,16 @@
+# @sap-ux/bsp-app-download-sub-generator
+
+## Features
+
+The SAP App download sub-generator enables users to download a basic LROP App from BSP repository. The sub-generator will download the app from the BSP repository and add it to the workspace.
+
+## Installation
+
+The SAP App download sub-generator  sub-generator is installed as part of the [@sap/generator-fiori](https://www.npmjs.com/package/@sap/generator-fiori) generator and cannot be used stand alone.
```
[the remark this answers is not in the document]
kjose90: [fixed](https://github.com/SAP/open-ux-tools/pull/3110/files#diff-8cc91a2d4b392076b876b9958236e282b231f9ae6abe4a1611870d05e3a51546R9)

packages/bsp-app-download-sub-generator/src/app/index.ts
kjose90: So as discussed I have moved download app to prompting phase with the new workflow

packages/repo-app-download-sub-generator/src/prompts/prompts.ts
```diff
@@ -0,0 +1,154 @@
...
+                }
+            }
+        }
+
+        let appList: AppIndex = [];
+        const appSelectionPrompt = [
+            {
+                when: async (answers: RepoAppDownloadAnswers): Promise<boolean> => {
+                    if (answers[PromptNames.systemSelection]) {
+                        if (quickDeployedAppConfig?.appId) {
```
cianmSAP, line 101: is this if/else necessary, could you not just call 

```
appList = await fetchAppListForSelectedSystem(
                                systemQuestions.answers.connectedSystem?.serviceProvider as AbapServiceProvider,
                                quickDeployedAppConfig.appId
                            );
```
The second param `appId` for `fetchAppListForSelectedSystem` is optional so no need to check if it exists up front

cianmSAP (changes requested):
Couple of small comments but looks good

packages/repo-app-download-sub-generator/src/prompts/prompts.ts
kjose90: Yes good spot 
[here](https://github.com/SAP/open-ux-tools/pull/3110/files#diff-45152bcdb20e78376760e010ea25844db92c69936efdd022fd2ee0da158a0d75R101)

packages/repo-app-download-sub-generator/src/prompts/prompts.ts
```diff
@@ -0,0 +1,154 @@
...
+                type: 'list',
+                name: PromptNames.selectedApp,
+                default: () => (quickDeployedAppConfig?.appId ? 0 : undefined),
+                guiOptions: {
+                    mandatory: !!appList.length,
+                    breadcrumb: t('prompts.appSelection.breadcrumb')
+                },
+                message: t('prompts.appSelection.message'),
+                choices: (): { name: string; value: AppInfo }[] => (appList.length ? formatAppChoices(appList) : []),
+                validate: async (answers: AppInfo): Promise<string | boolean | undefined> => {
```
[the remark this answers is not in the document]
kjose90: I can do this in a follow up PR ...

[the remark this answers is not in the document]
cianmSAP: Ok, just to note aswell I don't know why `| undefined` is a possible return type for this

packages/repo-app-download-sub-generator/src/prompts/prompts.ts
```diff
@@ -0,0 +1,148 @@
...
+                        return t('prompts.appSelection.noAppsDeployed');
+                    }
+
+                    // Valid app selected, try to download
+                    if (answers?.appId) {
+                        try {
+                            await downloadApp(answers.repoName);
+                            return true;
+                        } catch (error) {
+                            throw new Error(t('error.appDownloadErrors.appDownloadFailure', { error: error.message }));
```
cianmSAP, line 136: Should you just return the `t('error.appDownloadErrors.appDownloadFailure', { error: error.message }` so it's a validation message? Throwing the error will crash the generator

cianmSAP (approved):
Thanks for the updates @kjose90 
Comments addressed
Few follow ups needed but ok for now
Changeset looks good

devinea (approved):
re-approving.

lfindlaysap (approved):

Issue #3113, opened 2025-04-11
BUG - issue with latest cf-writer module


longieirl:
### Description
A number of issues have been identified when testing the latest version of `cf-deploy-config-writier`.

Issue 1.
When `generateCAPConfig` is called to generate the CAP deployment configuration, the `package.json` is not being updated with the devDependency `@sap/cds-dk`. This is added as part of the `cds add mta` command which generates the `mta.yaml` and also updates the package.json and is required when the `before-all` script is executed;
```
build-parameters:
  before-all:
    - builder: custom
      commands:
        - npm ci
        - npx cds build --production
```

Issue 2. 
If there is more than one HTML5 app in the CAP `apps` folder, when `generateCAPConfig` is executed, it appends the HTML5 apps to the mta.yaml using the following structure;

```
  - name: captestprojectproject2
    type: html5
    path: app/project2
    build-parameters:
      build-result: dist
      builder: custom
      commands:
        - npm ci
        - npm run build
      supported-platforms:
        []
```
This module type uses a different naming convention i.e.  `captestprojectproject2` instead of `project2` and the command `npm run build` is not supported for SAP Fiori generated apps. To support SAP Fiori generated apps in the mta.yaml, the command needs to be changed to ` npm run build:cf`.

Issue 3.
In some instances, when consuming `generateCAPConfig` or `generateAppConfig` you might want to delay the `npm install` step being executed. For example, you might want to append additional `scripts` to the or append new dependencies or devDependencies to the `package.json`. To support, this these methods need to allow a flag to skip the install phase.

Delaying the install step is important, as the mta.yaml contains a before-all script `npm ci` which will fail since the package.json and package-lock.json are not in sync.  In the case of the SAP Fiori tools generator, the install lifecycle method, will instead invoke the required `npm install` step which will align the required files to support `npm ci`.

### Steps to Reproduce
Sample code;
```
const fs1 = await generateCAPConfig({ mtaId: 'mymtaproject', routerType: RouterModuleType.Managed, mtaPath }, null, logger);
const ui5AppPath = join(__dirname, '../../captestproject/app/project1');
const fs2 = await generateAppConfig({appPath: ui5AppPath, addManagedAppRouter: true, destinationName: 'fiori-default-srv-api'}); 
  return new Promise((resolve) => {
     fs1.commit(resolve);
     fs2.commit(resolve);
  });
}
```
[...]

PR #3114, opened 2025-04-11
fix(cf-deploy-config-writer): improvements in cf-writer flows

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/3113

- CAP flow improvement around adding devDependency
- update tests
- append new tests to cover known flow with multiple HTML5 apps added by `cds add mta` where the HTML name is different.

A number of the snapshot tests are showing `stateCleared` this is because `fs.commit` is being run when skipInStall is false during the unit tests.


packages/cf-deploy-config-writer/src/translations/cf-deploy-config-writer.i18n.json
```diff
@@ -14,9 +14,12 @@
         "cleanupModules": "Cleanup modules.",
         "creatingMta": "Creating MTA using {{- cdsParams }}...",
         "html5AppAdded": "HTML5 application added to MTA configuration {{- appName}}",
+        "html5AlreadyExists": "HTML5 application {{- appName}} already exists in the MTA configuration.",
         "addXsuaaService": "Adding XSUAA service to MTA configuration.",
         "mtaReadWithPrefix": "Read mta.yaml with prefix {{- prefix }}.",
-        "errorReadingMta": "Error reading mta.yaml file {{- error }}."
+        "errorReadingMta": "Error reading mta.yaml file {{- error }}.",
```
lfindlaysap, line 20:
```suggestion
        "errorReadingMta": "An error when reading the `mta.yaml` file. {{- error }}.",
```
What should users do to try to fix this error? View the logs?

packages/cf-deploy-config-writer/src/translations/cf-deploy-config-writer.i18n.json
```diff
@@ -14,9 +14,12 @@
...
+        "unableToReadPackageJson": "Unable to read package.json file {{- error }}.",
```
lfindlaysap, line 21:
```suggestion
        "unableToReadPackageJson": "Unable to read the `package.json` file {{- error }}.",
```
What should users do to try to fix this error? View the logs?

lfindlaysap (changes requested):
@longieirl, here are my comments.

packages/cf-deploy-config-writer/src/translations/cf-deploy-config-writer.i18n.json
longieirl: There is nothing we can do! Normally its related to incorrectly formatted JSON which is rare but this would fail the build process much earlier anyway. A 2nd reason possibly is related to a missing package.json, again really ad-hoc to have a project without a package.json.

packages/cf-deploy-config-writer/src/cf-writer/app-config.ts
```diff
@@ -238,17 +247,26 @@ async function generateDeployConfig(cfAppConfig: CFAppConfig, fs: Editor): Promi
     if (config.isMtaRoot) {
         await updateRootPackage({ mtaId: config.mtaId ?? config.appId, rootPath: config.rootPath }, fs);
     }
+    if (!skipInstall) {
+        // When installing, we need to ensure that the package.json is written to disk before running npm install
+        await new Promise((resolve) => fs.commit(resolve));
```
IainSAP, line 252: Not a good idea to call `commit` as its managed by the Yeoman environment lifecycle.

cianmSAP (dismissed):
Updates look good
Have not tested

packages/cf-deploy-config-writer/src/cf-writer/app-config.ts
longieirl: This entire flow is now removed.

IainSAP (approved):
Thx @longieirl 

- Tests added to cover
- Not manually tested

cianmSAP (approved):
re-approving

lfindlaysap (approved):

Issue #3116, opened 2025-04-11
BUG - Odata Service inquirer - Validation error for failed V4 request blocking service selection step


cianmSAP:
### Related Feature
Feature request: _issue number_

### Description
If connecting to a destination, the validation error is persisting on the system selection prompt, appears to be due to the 404 error returned from the V4 catalog request

### Steps to Reproduce
Connect to a NO_AUTH destination which does not have an active V4 catalog. 
V2 catalog will return 401 and V4 will return 404

### Expected results
Basic auth prompts are shown due to 401 error from V2 catalog and no other blocking message

### Actual results
Basic auth prompts are shown but a validation error from the 404 returned from the V4 catalog request blocks progressing from the step

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

IainSAP:
Fixed

PR #3118, opened 2025-04-11
fix(odata-service-inq): handle auth errors for catalog requests

cianmSAP:
#3116 

If the V2 request returns an auth error, we run the V4 catalog request
However if this also fails, we should return the original auth error so it can be handled accordingly


packages/odata-service-inquirer/src/prompts/connectionValidator.ts
```diff
@@ -438,34 +438,40 @@ export class ConnectionValidator {
...
         }
     }
 
     /**
-     * Check if we should attempt to use the v4 catalog service as a fallback.
+     * Depending on the initial error, it will run the catalog request and handle the error if it occurs.
      *
-     * @param statusCode http status code, if not provided will return false as we cannot determine the reason for v2 catalog request failure
-     * @returns true if we should attempt the v4 catalog service
+     * @param error - error
```
IainSAP, line 448: Determine if a v4 catalog request should be made based on the specified error and if the error originated from a v2 or v4 catalog request. If the error originated from a v2 catalog request and it is of type 401/403/404 then we try the v4 catalog. If the v4 catalog request fails, for any reason, and the v2 catalog request failed with 401/403 we will throw that error, ultimately resulting in a basic auth prompt for the end-user. Otherwise the v4 catalog request is thrown, which may also be a 401/403, again resulting in a basic auth prompt and re-validation. If both catalogs have returned 404 nothing else can be done, the error is thrown and reported.

@param error An error returned from either a v2 or v4 catalog `listServices` request.

cianmSAP: Updated, thank you

IainSAP (dismissed):
Thanks @cianmSAP 

- Bug fix seems appropriate to cover these edge cases 
- Test added to cover

longieirl (dismissed):
Changes look good.
Changeset is good.
Tests to cover change are good.

IainSAP:
@cianmSAP fyi: This needs to go ASAP: https://github.com/SAP/open-ux-tools/pull/3120

IainSAP (dismissed):
Re-approving...but there is another pending VH fix.

IainSAP (approved):
Re-approving

longieirl (approved):
Re-approving

PR #3120, opened 2025-04-11
fix(odata-service-inquirer): `removeCircularFromServiceProvider` undefined ref

IainSAP:
Fixes reference to `serviceProvider.services` and optionally chains in case of undefined in function `removeCircularFromServiceProvider `


cianmSAP (approved):
Fix looks good
Has been tested from previous implementation
Changeset

longieirl (approved):
Changes look good.
Tests are validating change.

PR #3121, opened 2025-04-14
Dont load osName() with module as it can cause test instabilities

IainSAP:
- Addresses an intermittent test failure where osName() throws during module loading.


cianmSAP (approved):

longieirl (approved):
Changes look good.

PR #3123, opened 2025-04-14
chore: update @sap/bas-sdk due to axios dependency

Klaus-Keller:
Update version of `@sap/bas-sdk@3.11.6` to address transient `axios` dependency.


heimwege (approved):
Version bump lgtm

donal-tobin-sap (approved):
Upgrade looks good.
Already resolved in tools-suite.

longieirl (approved):
changes look good.
Changeset is good.

PR #3124, opened 2025-04-15
fix: nested quick action flattening

voicis:
Quick Action flattening did not work correctly in case there are multiple sections, but only one table.

- Moved all path calculation to the `preview-middleware-client` so it is only done once.
- Separated logic to remove nodes with only one child from rendering

### DIFF
--- a/.changeset/small-jars-invent.md
+++ b/.changeset/small-jars-invent.md
@@ -1,0 +1,8 @@
+---
+'@sap-ux-private/control-property-editor-common': patch
+'@sap-ux-private/preview-middleware-client': patch
+'@sap-ux/control-property-editor': patch
+'@sap-ux/preview-middleware': patch
+---
+
+fix: nested Quick Actions not working if there are sections with only one child (e.g Change Table Columns)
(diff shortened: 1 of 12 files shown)


vadson71 (approved):
Code changes look good.
Good test coverage
Tested locally

vadson71 (changes requested):
@voicis  Some actions still need adaptation. In initialize() method, in cases when this.children is recursively processed, we need to use child.path instead if child index when we build tableMap key.
Please revise change-table-actions, create-table-action, enable-empty-row mode, enable-variant-management (for both v2 and v4)

vadson71 (approved):
Comments addressed
Tested locally 
Reapproving...

nikmace (approved):
Changes look good
Coverage is good
Did not test manually

vadson71 (dismissed):

nikmace (dismissed):
Re-approving

vadson71 (approved):

nikmace (approved):
Re-approve

PR #3128, opened 2025-04-15
fix: Existing services

broksy:
Updates for `@sap-ux/odata-service-writer`:

1. Generate unique service name if necessary (service with such name already exists)
2. Throws error in case service with the same URI exists


packages/odata-service-writer/src/data/defaults.ts
```diff
@@ -18,15 +18,42 @@ function setDefaultServicePath(service: OdataService): void {
     service.path = service.path?.endsWith('/') ? service.path : (service.path ?? '') + '/';
 }
 
+/**
+ * Generates a unique name for the service.
+ *
+ * @param {OdataService} dataSources - The service object whose path needs to be set or modified.
```
IainSAP, line 24: The param descriptions might need to be updated?

packages/odata-service-writer/src/translations/odata-service-writer.i18n.json
```diff
@@ -2,6 +2,7 @@
     "error": {
         "requiredProjectFileNotFound": "Invalid project structure. Cannot find required file {{-path}}",
         "requiredProjectPropertyNotFound": "Required project property: {{-property}} was not found in file: {{-path}}",
+        "requiredServiceAlreadyExists": "Service with URI: {{-uri}} already exists",
```
lfindlaysap, line 5:
```suggestion
        "requiredServiceAlreadyExists": "A service with the {{-uri}} URI already exists. Choose another service.",
```

lfindlaysap (changes requested):
@broksy, here is my suggestion. Please let me know if "Choose another service" is correct.

heimwege (approved):
changeset ok
excellent code coverage
did NOT test manually

packages/odata-service-writer/test/unit/index.test.ts
```diff
@@ -554,19 +621,19 @@ describe('generate', () => {
...
                 });
                 const configCopy = cloneDeep(Object.assign({}, config, { name: 'mainService' }));
-                await enhanceData('', configCopy, fs);
+                await enhanceData('', configCopy, fs, true);
                 expect(configCopy).toMatchInlineSnapshot(`
                     Object {
                       "model": "",
                       "name": "mainService",
-                      "path": "/V2/Northwind/Northwind.svc/",
+                      "path": "/V2/Northwind/Northwind.svc",
```
devinea, line 636: not sure the trailing / should be removed?

broksy: @devinea ,
It's because of this: https://github.com/SAP/open-ux-tools/pull/3128/files#diff-15f0a9bfab30f8f41214d7df999fa2001924044c022bbfc261beb183b78faabcL158

As you can see from the description, the test has changed from testing add to testing update.

In this case the `enhanceData` runs during the update.

devinea (approved):

PR #3131, opened 2025-04-16
fix(fiori-annotation-api): empty path values

voicis:
Fixes writing empty path values in CDS files.

CDS does not support empty paths, so we need to use empty string values which later gets translated into empty path value by the compiler.


marufrasully (approved):
thanks @voicis 
* changes looks
* great test coverage
* did not test it locally

Jimmy-Joseph19 (approved):
Changes looks good
Changeset found 
Covered with unit test

Issue #3022, opened 2025-03-13
TBI - Expose new writing functionality to generate app frontend service mta and supporting deployment configuration


longieirl:
As a Fiori developer, I'd like to be able to use AppFront as a deployment option when deploying to Cloud Foundry.

Description
AppFront is a simplified CF deployment method that is launching in 2025. The cf-writer needs to be updated to support this deployment.

Acceptance Criteria
```
Given I choose to add deployment configuration to my Fiori project, either during generation or afterwards, and I have the feature toggle enabled, if I choose CF as my deploy target, and I don't already have an MTA.yaml file defined, then I will be offered the following 3 deployment options:

None
Add application to managed application router
Add application to Application Frontend service
```

Please note, as it stands, `cds` changes haven't been released so the writer needs to proceed with generating a custom template.

### Value
Support new router configuration

### Notes

https://help.sap.com/http.svc/login?time=1744289938854&url=%2Fdocs%2Fapplication-frontend-service%2Fapplication-frontend-service-internal%2Fwhat-is-application-frontend-service-beta%3Fstate%3DDRAFT

### Tasks
- [x] Allow cf-writer to generate app frontend router configuration for CAP
- [x] Allow cf-writer to generate app frontend router configuration for HTML5 application
- [x] Use a custom template to generate the mta (until CDS is released)
- [x] Extend test cases

longieirl:
Validated on tools-suite, part of next release.

PR #3133, opened 2025-04-16
fix(cf-deploy-config-sub-generator): issues found when testing

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/3022

- text label using incorrect property
- options passed to subgenerator not being used
- pass logger to cf-inquirer


cianmSAP (approved):
Updates are small and clear
Changeset looks good

kjose90 (approved):
Clear updates, test, changeset ✅

Issue #3136, opened 2025-04-16
BUG - fiori-app-sub-generator: flp app id is incorrectly formed


IainSAP:
### Description

The flp app id generated by `fiori-app-sub-generator` should be in the form <namespace><name> but is instead in the format <name><namespace>

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3137, opened 2025-04-16
fix(fiori-app-sub-generator): Flp App Id form is incorrect

IainSAP:
https://github.com/SAP/open-ux-tools/issues/3136

- Reverses namespace/name to be correctly ordered in flp app id string.
- Updates test


cianmSAP (approved):
Updates are clear
Tests to cover
Changeset looks good

PR #3147, opened 2025-04-17
fix: Space between empty table and add button

broksy:
Fixes space between empty table and add button.
Occurs when `noDataText` is undefined and `rows` length is 0. 

From:

To:


815are (approved):
checked:
1. style changes meets expected/described behaviour
2. changeset persists

devinea (approved):

PR #3173, opened 2025-04-22
Update i18next versions to v24.2.3

UserBread:
* Update all packages using i18next to `v24.2.3`
* Migrate `i18n.ts` files to be compatible with `v24.2.3`
* Update snapshots for `ui-service-inquirer` and `odata-service-inquirer`
* Update tests


packages/fiori-app-sub-generator/src/fiori-app-generator/fioriAppGenerator.ts
```diff
@@ -330,6 +330,7 @@ export class FioriAppGenerator extends Generator {
     }
 
     async writing(): Promise<void> {
+        await initI18nFioriAppSubGenerator();
```
IainSAP, line 333: Not sure why this is now necessary?

packages/flp-config-sub-generator/test/app.test.ts
```diff
@@ -517,7 +517,7 @@ describe('flp-config generator', () => {
         ).resolves.not.toThrow();
 
         expect(showWarningSpy).toHaveBeenCalledWith(
-            t('warning.updatei18n', { path: `.${OUTPUT_DIR_PREFIX}/app1/webapp/manifest.json` }),
+            expect.stringContaining(t('warning.updatei18n')),
```
IainSAP, line 520: Shouldnt the path also be tested?

packages/odata-service-writer/test/unit/annotations.test.ts
```diff
@@ -34,7 +34,7 @@ describe('metadata parsing', () => {
     it('getAnnotationNamespaces: metadata parsing', () => {
         expect(() => {
             getAnnotationNamespaces({ metadata: invalidEdmx });
-        }).toThrow(t('error.unparseableXML'));
+        }).toThrow(t('error.unparseableXML', { error: "Error: boolean attribute '<Schema' is not allowed.:13:5" }));
```
IainSAP, line 37: This error will have to be continously updated as it refers specifically to a loc. Probably should not check that detail.

PR #3174, opened 2025-04-22
fix(fiori-sub-gen): check enable typescript answer to determine use of npm workspaces

cianmSAP:
If enable typescript prompt is answered as true, it is required that the CAP project will be updated to use NPM workspaces


longieirl (approved):
Changeset is good.
Change is small and tests to cover change.

docirl (approved):
Code changes look ok, npmWorkspaces are required for typescript so updated condition makes sense
Test added

PR #3175, opened 2025-04-22
fix(cf-deploy-config-writer): refactor code to align two templates into one

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/3022

- remove `xs-app-no-destination.json` no longer required since the properties passed to the template can now be handled by remaining template `xs-app-destination.json`


cianmSAP (approved):
Updates look good
Tests to cover

Issue #3178, opened 2025-04-23
BUG - Consumption of isReuseComponent API from RTA library is incorrect


mmilko01:
### Related Feature
Feature request: https://github.com/SAP/open-ux-tools/issues/3045

### Description
Reuse component detection is done via `isReuseComponent` API from RTA library in UI5 versions above 1.134.
Currently, the API is not invoked correctly in the Adaptation Editor.

### Steps to Reproduce
1. Create adaptation project of type Cloud for UI5 version >= 1.134
2. Right Click anywhere on the Application iFrame
4. Notice the context menu doesn't appear and also `Add Controller to Page` quick actions is not visible

### Expected results

- Context Menu appears on right click and `Extend with Controller` is available
-  `Add Controller to Page` quick actions is visible

### Actual results

-  Context Menu does not appear on right click and `Extend with Controller` is not available
-  `Add Controller to Page` quick actions is not visible

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3180, opened 2025-04-23
fix: reuse component api consumption in Adaptation Editor

mmilko01:
#3178

- fixes the call for `isReuseComponent` API from RTA library


nikmace (approved):
Changes look good
Coverage is good
Changeset is present
Did not test manually

PR #3181, opened 2025-04-23
fix(ui-service-gen): show draft enabled when valid for business objects

docirl:
- only show the `Draft Enabled` prompts for business object that are compatible with draft
- response of suggest 'content' for business object will include `useDraft` if object can use draft


devinea (approved):
code changes are clear and covered by tests
changeset ✅

longieirl (approved):
Changeset is good.
Tests are appropiate to cover change.

PR #3187, opened 2025-04-25
fix(ui5-app-inquirer): CAP checks on virtual endpoints prompt

cianmSAP:
- Adds additional criteria to show the `enableVirtualEndpoints` prompt for cap projects
   - i.e if cds plugin is enabled or if the enable typescript question is true

### DIFF
--- a/packages/ui5-application-inquirer/src/prompts/prompts1.ts
+++ b/packages/ui5-application-inquirer/src/prompts/prompts1.ts
@@ -281,18 +281,15 @@ export function getAddFlpConfigPrompt(
 /**
  * Get the `enableVirtualEndpoints` prompt.
  *
- * @param isCapProject is this a CAP project - additional checks performed if true
  * @param capCdsInfo CDS UI5 plugin information
  * @returns the `enableVirtualEndpoints` prompt
  */
-export function getEnableVirtualEndpoints(
-    isCapProject: boolean,
-    capCdsInfo?: CdsUi5PluginInfo
-): UI5ApplicationQuestion {
+export function getEnableVirtualEndpoints(capCdsInfo?: CdsUi5PluginInfo): UI5ApplicationQuestion {
     return {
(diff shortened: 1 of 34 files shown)


longieirl (dismissed):
Changeset is good.
Code change is small and covered by tests.

packages/ui5-application-inquirer/src/prompts/prompts1.ts
```diff
@@ -281,10 +281,21 @@ export function getAddFlpConfigPrompt(
...
  */
-export function getEnableVirtualEndpoints(): UI5ApplicationQuestion {
+export function getEnableVirtualEndpoints(
+    isCapProject: boolean,
+    capCdsInfo?: CdsUi5PluginInfo
+): UI5ApplicationQuestion {
     return {
+        when: (answers: UI5ApplicationAnswers): boolean => {
+            if (isCapProject) {
+                return capCdsInfo?.isCdsUi5PluginEnabled || !!answers.enableTypeScript;
```
IainSAP, line 295: Should l we be checking for `enableTypescript` as this implies knowledge about other flows and tightly couples to the typescript flow. I had previously removed `enableTypescript` checks and rely on the cds info only.

devinea (dismissed):
Thanks @cianmSAP

packages/ui5-application-inquirer/src/prompts/prompts1.ts
cianmSAP: As discussed, this needs following up as the writers don't currently support the assumption that cds ui5 plugin will be enabled if the mid cds version is met

IainSAP (approved):
Thanks @cianmSAP 

- Updates as expected
- Agreed to follow-up to review the auto update of cds plugin and workspaces separately

longieirl (approved):
re-approving

PR #3190, opened 2025-04-25
feat(fiori-annotation-api): relax metadata reference check

voicis:
UI5 runtime also works with annotations files without references to the metadata, so we should not prevent tools from working in such cases either. We should still report this as warning since it does not conform with the OData spec, but currently we can only throw errors.


marufrasully (approved):
thanks @voicis 
* changes are straightforward
* secured with unit tests
* did not test it locally

vadson71 (approved):
Code changes look good
Test coverage is good
Didn't test locally

Issue #3191, opened 2025-04-25
TBI -


longieirl:
### Description (include screenshots)
Currently, if a BSP app is already deployed and the user decides to define a different package name to update the BSP app, the deployment task is successful but the package defined in the request will not contain the changes in the deployment task. 

This is expected since `PUT` requests i.e. changes to an existing app don't validate or use the package defined in the deployment task. The package is only considered for POST requests.
In the customers experience, they were not aware of this and are expecting the package they have defined, to have the BSP changes.

### Value
Provides more context to developer if they expect the package name to be referenced during a PUT.

### Architecture Elaboration
Does it requires architecture elaboration?

### Notes
- Found when investigating a deployment issue where it was expected the package name used, would contain the  development objects that were changed.

### Tasks
- [ ] If an app is already deployed, validate if the BSP package and the deployed package are different, if not then show a warning;
```
this.log.warn(`Cannot change package assignment from ${info.Package} to ${bsp.package}.`);
```
- [ ] Update test cases

longieirl:
Validated on latest release of ui5-tooling, showing message to indicate that TR is not being used if using a PUT request.

PR #3192, opened 2025-04-25
fix(axios-ext): Append warning message if BSP properties do not match

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/3191


IainSAP (approved):
Thanks @longieirl 

- Small update to add lo
- Agreed to look at improving the logger control from consumers as a TBI

cianmSAP (approved):
Small update looks good
Changeset and test to cover

Issue #3194, opened 2025-04-25
BUG - `abap-deploy-config-inquirer`: re-authentication should not be required


IainSAP:
Description
In BAS, after successful connection to a system with credentials, deploy config requires re-authentication

<img width="1048" alt="Image" src="https://github.com/user-attachments/assets/55807a2e-48c6-423f-8035-b292343e01c6" />

Steps to Reproduce
Using a BAS devspace
Generate Fiori app using a system that requires authentication (e.g. `NoAuthentication` on prem destination)
Continue to Deployment Configuration step

Expected results
On Deployment Configuration step, re-authentication is not required

Actual results
On Deployment Configuration step, re-authentication is required

PR #3196, opened 2025-04-25
fix(abap-deploy-config-inquirer): Re-auth should not be required

IainSAP:
#3194 

- Replace `??` with logical OR `||` for correct behaviour (thx Sonar 😞  )
- Update test to replicate real world scenario where the bug occurred


cianmSAP (approved):
Thanks @IainSAP
Small but important change
Tests and changeset 👍🏻

longieirl (approved):
Small change and covered with unit test.

PR #3197, opened 2025-04-28
fix(odata-service-writer): avoid double slashes when adding new odata service 

815are:
Problem happens if we pass service like:
```
{
  annotations: [
    {
      technicalName: '/TEST/DUMMY_SERVICE'
    }
  ]
}
```
then double forward slash is added here:
image

As fix - sanitize trailing slashes before creating uri for `settings.localUri` - [trimSlashes](https://github.com/SAP/open-ux-tools/blob/fix/avoidDoubleSlashesWhenAddingNewDataSource/packages/odata-service-writer/src/data/manifest.ts#L536-L544)

### DIFF
--- a/packages/odata-service-writer/src/data/manifest.ts
+++ b/packages/odata-service-writer/src/data/manifest.ts
@@ -232,6 +232,21 @@ function removeUnusedAnnotations(
 }
 
 /**
+ * Resolves the local service URI by joining the service name and technical name,
+ * ensuring that no double slashes appear.
+ *
+ * @param serviceName - the name of the service (may contain slashes).
+ * @param technicalName - the technical name or path of the service (may contain slashes).
+ * @returns The resolved local service URI in the format `localService/{serviceName}/{technicalName}.xml`.
+ */
+function resolveLocalServiceUri(serviceName: string, technicalName: string): string {
+    // Remove beginning and end forward slashes
+    serviceName = serviceName.replace(/^\/+|\/+$/g, '');
+    technicalName = technicalName.replace(/^\/+|\/+$/g, '');
+    return `localService/${serviceName}/${technicalName}.xml`;
+}
+
+/**
  * Adds remote annotations to manifest dataSources and removes unused annotations by the service.
  *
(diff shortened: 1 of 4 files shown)


packages/odata-service-writer/src/data/manifest.ts
```diff
@@ -275,7 +260,7 @@ function addRemoteAnnotationDataSources(
                     )}',Version='0001')/$value/`,
                     type: 'ODataAnnotation',
                     settings: {
-                        localUri: resolveLocalServiceUri(serviceName, remoteAnnotation.technicalName)
+                        localUri: `localService/${serviceName}/${trimSlashes(remoteAnnotation.technicalName)}.xml`
```
IainSAP, line 263: Why not fix the existing code rather than adding only a selective change. Dont we need to sanitize the service name also?

packages/odata-service-writer/src/data/manifest.ts
```diff
@@ -541,3 +526,19 @@ export async function updateManifest(
...
 }
+
+/**
+ * Trims leading and trailing slashes from the input string.
+ *
+ * @param input - the string from which leading and trailing slashes will be removed.
+ * @returns The input string with leading and trailing slashes removed.
+ */
+function trimSlashes(input: string): string {
+    while (input.startsWith('/')) {
```
IainSAP, line 537: Could this be used as an attack vector?

815are: regex was giving following sonar issue for me:
image

IainSAP: Yeah I think you would need to add a character limit anyway or Sonar would complain about DOS attack. Previously we needed to limit the '/' check.
```
// We remove trailing slashes (up to 10, infinite would allow DOS attack) from the host to avoid double slashes when appending the service path.
        this._destinationUrl = servicePath
            ? destUrl.replace(
                  `https://${destination.Name.toLowerCase()}.dest`,
                  destination.Host.replace(/\/{1,10}$/, '')
              )
            : destination.Host;
```

PR #3200, opened 2025-04-29
Fix App Select Prompt Display Issue When Switching Systems

kjose90:
This PR fixes an issue in the `@sap-ux/repo-import-app-sub-generator` package where the App Select Prompt incorrectly displays `[Object] [Object]` instead of Click to Display List when switching systems after selecting an app. The issue occurs because the prompt state is not properly reset when switching systems.

### DIFF
--- a/.changeset/dull-cheetahs-float.md
+++ b/.changeset/dull-cheetahs-float.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/repo-app-import-sub-generator': patch
+---
+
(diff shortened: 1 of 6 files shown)


cianmSAP (approved):
Change is small and clear
Changeset looks good
Have not tested

lfindlaysap (approved):

devinea (approved):
lgtm. Thanks.

Issue #3198, opened 2025-04-29
BUG - Issues for Adaptation Projects using TypeScipt


nikmace:
### Description
**Adaptation Editor**
- The handler for reading all controller extension files from the workspace currently only takes into account `.js` files. Which then in turn allows you to press the 'Create' button in the Controller Extension dialog when:
  - A `code_ext` change is made, and we try to add another `code_ext` change for the object page with the same name. Although that is code by the server and a `409` is returned, we still need to prevent it in the UI and disable the 'Create' button.
- The text for the existing controller extension is never shown in the dialog for TS projects. The handler for checking for an existing controller cannot be called because the `controllerName` now includes slashes (i.e. `module:sap/ui/demoapps/rta/freestyle/controller/ProductDetail.controller`).

----------

**ADP Generator**
When generating a TypeScript adaptation project and choosing snapshot as the SAPUI5 version, the wrong types package is being put in the `package.json` and the `tsconfig.json`. Using snapshot as the UI5 Version is not recognized by the semver versioning npm package in the code that decides which types package to use. That is a bug or something that was overlooked before.
These two methods are causing problems:
```ts
const typesVersion = getEsmTypesVersion(data.ui5?.version);
const typesPackage = getTypesPackage(typesVersion);
```
Using SAPUI5 version 1.136.0 puts the types package as `"@sapui5/types": "~1.136.0"`

_Note: This change will come in a separate PR under this issue._

### Steps to Reproduce
Described the steps for the problems in the description.

### Expected results
Creating a second controller extension change with the same name **is not** allowed.
Trying to create a controller extension for the same view shows the message for the existing controller.

### Actual results
Creating a second controller extension change with the same name **is** allowed.
Trying to create a controller extension for the same view shows the message for the existing controller fails because the request cannot be processed or the controllerName is not found.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3206, opened 2025-04-29
fix: Issues for Adaptation Projects using TypeScipt

nikmace:
Fix for #3198.
- Changes the two handlers.
  - Handler for reading all controllers now reads `.js` and `.ts` files and returns only their names without the file extension. Then these names are compared on the UI with the user input. This way, if the user tries to create a controller with the same name that exists in the workspace, an error will be shown.
  - Handler for checking existing controller info now accepts `controllerName` in the URL query (`?name=<controllerName>`). Sanitization is removed because the new `controllerName` syntax includes ":" and "/" which get removed when sanitized.


mmilko01 (approved):
Changeset OK
Code looks good
Great test coverage
Did not test manually

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -130,11 +130,13 @@ export default class RoutesHandler {
...
-            const files = await this.readAllFilesByGlob('/**/changes/coding/*.js');
+            const files = await this.readAllFilesByGlob('/**/changes/coding/*.{js,ts}');
 
-            const fileNames = files.map((file) => ({
-                controllerName: file.getName()
-            }));
+            const fileNames = files.map((file) => {
+                const fullName = file.getName();
+                const name = fullName.replace(/\.(js|ts)$/, '');
+                return { controllerName: name };
```
heimwege, line 138: You could also re-use the new type here (if you don't define it as array) and ensure type safety by using `return { controllerName: name } satisfies ControllerListButNotAsArray`

packages/preview-middleware-client/src/adp/api-handler.ts
```diff
@@ -171,7 +171,10 @@ export async function getDataSourceAnnotationFileMap(): Promise<AnnotationDataSo
  * @returns {CodeExtResponse} Returns path to existing controller if found
  */
 export async function getExistingController(controllerName: string): Promise<CodeExtResponse> {
-    return request<CodeExtResponse>(`${ApiEndpoints.CODE_EXT}/${controllerName}` as ApiEndpoints, RequestMethod.GET);
+    return request<CodeExtResponse>(
+        `${ApiEndpoints.CODE_EXT}?name=${controllerName}` as ApiEndpoints,
```
voicis, line 175: Query parameters should be encoded using `encodeURIComponent`.

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
```diff
@@ -34,7 +34,7 @@ interface ControllerExtensionService {
 }
 
 type ControllerModel = JSONModel & {
-    getProperty(sPath: '/controllersList'): { controllerName: string }[];
+    getProperty(sPath: '/controllersList'): { controllerName: /** File name without extension */ string }[];
```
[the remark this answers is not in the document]
nikmace: Changed in: [483af3b](https://github.com/SAP/open-ux-tools/pull/3206/commits/483af3b55a3e0e69f46f6e89a9f603ded3dafeca)

packages/preview-middleware-client/src/adp/api-handler.ts
nikmace: Changed in [483af3b](https://github.com/SAP/open-ux-tools/pull/3206/commits/483af3b55a3e0e69f46f6e89a9f603ded3dafeca). Decided to use URLSearchParams, which is more robust if in the future we will wanna add more query parameters.

packages/adp-tooling/src/preview/routes-handler.ts
nikmace: If I wanna reuse that type, we have to add `adp-tooling` to `preview-middleware-client` as a dependency. Currently, we are not using `adp-tooling` code in `preview-middleware-client`. If you want, I can do like in this commit: [271add8](https://github.com/SAP/open-ux-tools/pull/3206/commits/271add81946410d67ba8b7e26c7035d5758f568d). Although, I still don't see much value in the satisfies here.

mmilko01 (approved):
re-approving

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -41,6 +41,8 @@ export interface AnnotationDataSourceResponse {
     annotationDataSourceMap: AnnotationDataSourceMap;
 }
 
+type ControllerInfo = { controllerName: string };
```
heimwege, line 44: My comment might have been misleading/incomplete. Sorry for that. Let me rephrase: If we actually use the same type as in https://github.com/SAP/open-ux-tools/pull/3206/files#diff-42b325143d008e05622468e3a6a4948d9f46a1cceaeae85522c16c9f2f04e5deR36 instead of re-create a new type we can ensure that changes to the type are verified at both locations where the type is being used.

Example: You rename `controllerName` here to `MyControllerName` the TS typecker will not complain. But actually the code will no longer work because `ControllerExtension.controller.ts` now has an issue because it uses a non existing property `controllerName`.
But if ControllerExtension.controller.ts and routes-handler.ts use the very same type, the change to  `MyControllerName`  will lead to a typecheck error in `ControllerExtension.controller.ts` because there still `controllerName` is being used.

so we need a setup similar to

```ts
type ControllerInfo = {
    /**
     * File name without extension
     */
    controllerName: string;
};
type ControllerList = ControllerInfo[];
```

heimwege: @Klaus-Keller seems like this is yet another demand for centralized types that can be re-used across packages w/o the need for dependencies on other packages just to be able to re-use existing types.

@nikmace I don't want to block this PR because this issue occurs in other placers as well so I'll approve

heimwege (approved):
changes are straightforward
changeset exists
test coverage is great
did NOT test manually

voicis (approved):
- Review comments addressed
- Did not test manually

PR #3207, opened 2025-04-29
fix(odata-service-writer): Write service annotations using `name` instead of `technicalName` to avoid sub folders

815are:
We noticed that service annotation files are written with sub folder like:
image

after correction it looks following:
image

### DIFF
--- a/.changeset/light-dryers-jump.md
+++ b/.changeset/light-dryers-jump.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/odata-service-writer': patch
+---
+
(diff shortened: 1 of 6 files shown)


devinea (approved):
changes look good and snapshot updates are as expected. 
changeset ✅

devinea (approved):

ytpo-lyne (approved):
- code changes to avoid subfolders look good
- tests updated accordingly
- changeset exists 
- tested manually, no issues found

Thanks for the fix @815are, approving.

PR #3189, opened 2025-04-25
Remove UI Theme Belize from the generator for UI5 Version over `1.136`

kjose90:
- Filter Belize Theme for Versions Greater Than `1.136.0`
- For versions below `1.136.0` Belize theme is displayed as `Belize (deprecated)` 
- Unit tests

PR #3217, opened 2025-05-01
Fix/remove UI theme belize label

kjose90:
Related to [PR](https://github.com/SAP/open-ux-tools/pull/3189)

This PR updates the UI5 theme label based on the selected UI5 version:

- `Belize (deprecated)` is shown when `ui5Version >= 1.120 and < 1.136.0`
- `Belize` is shown when `ui5Version < 1.120`

### DIFF
--- a/packages/ui5-info/src/ui5-theme-info.ts
+++ b/packages/ui5-info/src/ui5-theme-info.ts
@@ -69,27 +69,26 @@ export function getUi5Themes(ui5Version: string = defaultVersion): UI5Theme[] {
     const ui5VersionSince = ui5Version.replace('snapshot-', '');
     const cleanSemVer = coerce(ui5VersionSince);
     if (cleanSemVer) {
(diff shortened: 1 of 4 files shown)


packages/ui5-info/src/ui5-theme-info.ts
```diff
@@ -68,18 +69,26 @@ export function getUi5Themes(ui5Version: string = defaultVersion): UI5Theme[] {
...
+        return Object.values(ui5Themes)
+            .map((ui5Theme) => {
+                const theme = { ...ui5Theme };
 
-            // Include themes that are within their version support range
-            if (isValidSinceVersion && isValidUntilVersion) {
-                return true;
-            }
+                if (
+                    theme.id === ui5ThemeIds.SAP_BELIZE &&
```
IainSAP, line 77: Can we please change the implementation so any theme gets marked as deprecated if its outside the range

devinea (approved):
Changes lg. Thanks.

PR #3221, opened 2025-05-02
fix: use memFs if passed while application access creation

Klaus-Keller:
Fix to use `memFs` in functions `updatePackageJSON()` and `updateManifestJSON()` in case it has been passed when `ApplicationAccess` instance was created using `createApplicationAccess()`.

### DIFF
--- a/.changeset/quick-rings-invite.md
+++ b/.changeset/quick-rings-invite.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': patch
+---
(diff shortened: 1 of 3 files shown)


heimwege (approved):
Changeset ok
Excellent test coverage
Changes are straightforward
Did not test manually

marufrasully (approved):
* straightforward changes
* secured with unit tests
* did not test it locally

PR #3222, opened 2025-05-02
fix(ui5-app-inquirer): fix 'when' condition for Typescript prompt

docirl:
- there was a bug when a project has the CdsUi5Plugin installed and hasMinCdsVersion but was not getting the option for Typescript
- update typescript prompt condition to show prompt when only hasMinCdsVersion is true
- current logic was hiding the prompt when the CdsUi5Plugin was installed but workspaces were not enabled. User should be given the option to select Typescript here as they have the plugin and we enable workspaces at generation.
- Warning is still shown to state we enable workspaces when Typescript is selected, if workspaces are not already enabled.
- Tests mostly cover the flow already, added an explicit test for `when` if only `hasMinCdsVersion ` is true (the flow for the bug)

### DIFF
--- a/packages/ui5-application-inquirer/src/prompts/prompts1.ts
+++ b/packages/ui5-application-inquirer/src/prompts/prompts1.ts
@@ -188,7 +188,7 @@ export function getEnableTypeScriptPrompt(capCdsInfo?: CdsUi5PluginInfo): UI5App
     return {
         when: (): boolean => {
             if (capCdsInfo) {
-                return capCdsInfo.isCdsUi5PluginEnabled || capCdsInfo.hasMinCdsVersion;
+                return capCdsInfo.hasMinCdsVersion;
             }
             return true;
         },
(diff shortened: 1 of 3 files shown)


longieirl (approved):
Changeset is good.
Minor changes and tests to cover change.

packages/ui5-application-inquirer/src/prompts/prompts1.ts
```diff
@@ -188,13 +188,13 @@ export function getEnableTypeScriptPrompt(capCdsInfo?: CdsUi5PluginInfo): UI5App
     return {
         when: (): boolean => {
             if (capCdsInfo) {
-                return capCdsInfo.isCdsUi5PluginEnabled || (capCdsInfo.hasMinCdsVersion && !capCdsInfo.hasCdsUi5Plugin);
+                return capCdsInfo.isCdsUi5PluginEnabled || capCdsInfo.hasMinCdsVersion;
```
cianmSAP, line 191: Is there any point in checking `isCdsUi5PluginEnabled` ?
If `isCdsUi5PluginEnabled` is true, then `hasMinCdsVersion` has to be true. 

If all that matters is `hasMinCdsVersion` is true then we could just check that in isolation?

docirl: Thanks, I've update it to only check `hasMinCdsVersion`

cianmSAP (approved):
Updates look good
Comments addressed
Changeset ✅

devinea (approved):
Thanks @docirl

PR #3091, opened 2025-04-07
refactor:  Enhance XML Fragment context menu control with addXMLPlugin Integration

mmilko01:
#3093

* For newer UI5 versions (>1.136) the RTA supplied `addXMLPlugin` is used.
* The plugin provides the same functionality, but also gives better control on the context menu as it can be configured based on the designtime metadata of the UI element.

PR #3223, opened 2025-05-02
fix(preview-middleware): new column is not visible after using `Add Custom Table Column` Quick Action

voicis:
Fix regression introduced by #3091.

Move `templateName` data to the new data transfer mechanism also for adding table column dialog.

We can't use plugin, because it only creates one fragment, but we need to create two at the same time.


nikmace (approved):
Changes look good
Did not test manually
Coverage is excellent

PR #3226, opened 2025-05-06
feat: Update of backend

broksy:
Adds `updateBackendToFioriToolsProxydMiddleware` to support backend updates.
Backends are matched by path and updated with the new data.
Includes updated tests to use more realistic test data.


packages/odata-service-writer/src/index.ts
```diff
@@ -117,7 +117,7 @@ async function update(basePath: string, service: OdataService, fs?: Editor): Pro
     const paths = await findProjectFiles(basePath, fs);
     const webappPath = await getWebappPath(basePath, fs);
     ensureExists(webappPath, ['manifest.json'], fs);
-    await enhanceData(basePath, service, fs);
+    await enhanceData(basePath, service, fs, true);
```
815are, line 120: shouldn't be param for update? As I understand in our UI we have two options - with update and without update

815are (approved):
checked:
1. code to pass `update` param to support save with/without yaml update - is easy to understand;
2. new tests are added and adjusted


I did not test manually

ytpo-lyne (approved):
- changes to update backend look good
- changeset exists 
- coverage met 
- tested manually, no further issues found

Approving.

heimwege (approved):
code coverage is excellent
changes are straightforward
did NOT test manually
changeset ok

Issue #3228, opened 2025-05-07
BUG(fiori-app-sub-generator): `ignoreCertError` is not passed to writers and is always `false`


IainSAP:
### Description
Regardless of the user setting `ignoreCertErrors` is always output in yaml files as `false`:

<img width="1118" alt="Image" src="https://github.com/user-attachments/assets/a96eafe3-9f4a-420b-a762-01ebca14222b" />

<img width="1023" alt="Image" src="https://github.com/user-attachments/assets/3282d8c2-3f85-4365-932e-9079e999568f" />

Note: Technically the property `ignoreCertErrors` is not exported from the `odata-service-inquirer` interface but should be.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3229, opened 2025-05-07
fix(fiori-app-sub-generator): Set 'ignoreCertErrors'

IainSAP:
Fix for: https://github.com/SAP/open-ux-tools/issues/3228

- Pass option as expected
- Adds to odata service answers interface
- Update tests


packages/odata-service-inquirer/src/types.ts
```diff
@@ -91,6 +92,11 @@ export interface OdataServiceAnswers {
      *
      */
     connectedSystem?: ConnectedSystem;
+
+    /**
+     * If the user chose to ignore the certificate error when connecting to the service the value will be true.
+     */
+    [serviceUrlInternalPromptNames.ignoreCertError]?: boolean;
```
cianmSAP, line 99: This looks odd but I guess it's the most robust approach, covers the unlikely event the prompt name was changed?

cianmSAP (approved):
Updates look good
Test updates covering the change
Changeset for fix ✅

PR #3232, opened 2025-05-07
fix(odata-service-inquirer): Set `ignoreCertError` to PromptState

IainSAP:
Follow-up to fix for: https://github.com/SAP/open-ux-tools/issues/3228

- Add `ignoreCertError` property to prompt state in case if this is referenced by a caller instead of using `prompt()` function which returns the answers object including all prompt answers
- Updates test


cianmSAP (approved):
Small  change
Changeset ✅

longieirl (approved):
changes look good to me.
Changeset is good.
Test to cover change.

PR #3233, opened 2025-05-08
fix: Incorrect types package put during adaptation project generation

nikmace:
Fix for #3198.
- Adds check for `snapshot` & `snapshot-untested` when trying to decide which types package should be used because semver does not recognize both of these versions (it does recognize "1.120.0-snapshot" for example).


packages/ui5-config/src/utils.ts
```diff
@@ -66,6 +66,10 @@ export function getTypesVersion(minUI5Version?: string) {
  * @returns string representing the types package name.
  */
 export function getTypesPackage(ui5Version?: string) {
+    if (ui5Version?.toLowerCase() === 'snapshot' || ui5Version?.toLowerCase() === 'snapshot-untested') {
+        return UI5_DEFAULT.TYPES_PACKAGE_NAME;
+    }
+
     const version = semVer.coerce(ui5Version) ?? semVer.coerce(UI5_DEFAULT.TYPES_VERSION_BEST);
```
nikmace, line 73: `semver.coerce` does not recognize versions like `snapshot` and `snapshot-untested`.
But it does recognize versions like `1.136.0-snapshot` and parse them to `1.136.0`.

_Another proposal to the one in the PR's changed code_:
**getTypesPackage**: We could check if the version includes a `snapshot` word; if it does, it will default to the TYPES_PACKAGE_NAME - `@sapui5/types` package. 
**getTypesPackage**: We don't have to check explicitly whether the version is `snapshot` because semver will return null and we will get TYPES_VERSION_BEST - 1.108.0.

If we do this change the tests will fail for versions like `1.102-snapshot`, which default to `UI5_DEFAULT.TS_TYPES_ESM_PACKAGE_NAME`. 

As far as I know, such `-snapshot` versions are only available internally and are removed after the version gets released. Therefore, testing UI5 versions lower than 1.113 with `-snapshot` does not make sense, and these tests should be removed from [utils.test.ts]. If you think we should catch such version with `.includes('snapshot')`, then I suggest we remove UI5 versions lower than 1.113 with `-snapshot` and put higher versions with `-snapshot`.

I am also not sure about the use case for this functionality in Fiori Tools, maybe I don't understand something here. What do you think about this @devinea?

devinea (dismissed):
changeset ✅

packages/ui5-config/src/utils.ts
```diff
@@ -66,6 +66,10 @@ export function getTypesVersion(minUI5Version?: string) {
  * @returns string representing the types package name.
  */
 export function getTypesPackage(ui5Version?: string) {
+    if (ui5Version?.toLowerCase() === 'snapshot' || ui5Version?.toLowerCase() === 'snapshot-untested') {
```
mmilko01, line 69: Do you think that changing this condition to check whether the version contains `snapshot` would be better? For internal versions we have several different identifiers as `nightly`, `nightly1136`, `monthly`, etc. We usually don't use those in adaptation project, but it might be case we need to use them in the future. 
My proposal is to change the conditions as:
```
ui5Version?.toLowerCase().contains('snapshot')
```

packages/ui5-config/src/utils.ts
nikmace: I've proposed the same solution here @mmilko01.

devinea (approved):
Changeset ✅

PR #3239, opened 2025-05-14
fix: resuse abapProvider

Jimmy-Joseph19:
When creating a new annotation file via quick action in a BAS environment where single sign-on is not enabled, it prompts for the system username and password again after they were initially provided to start the adaptation editor.


voicis (approved):
- Code changes looks good
- Changeset present
- Did not test manually

packages/adp-tooling/src/preview/change-handler.ts
```diff
@@ -345,11 +348,25 @@ export async function addAnnotationFile(
...
+ * @param {AbapServiceProvider} provider - abap provider.
  * @returns Promise<ManifestService>
  */
-async function getManifestService(basePath: string, logger: Logger): Promise<ManifestService> {
+async function getManifestService(
+    basePath: string,
+    logger: Logger,
+    provider?: AbapServiceProvider
+): Promise<ManifestService> {
+    if (!provider) {
```
mmilko01, line 359: I saw that the coverage of the code is 100%, but should there be a unit test for this branching?

mmilko01 (commented):
Just couple of small questions, I don't see an issue linked to this PR in the description. As far as I know we should create and link issues to the PRs.

packages/adp-tooling/src/preview/change-handler.ts
Jimmy-Joseph19: I have adapted the code now

mmilko01 (approved):
Code looks good
Excellent test coverage
Comments addressed
Changeset OK
Did not test manually

voicis (approved):

Issue #3238, opened 2025-05-13
BUG - Errors During Deployment in Business Application Studio when the package name contains slashes


avasilev-sap:
### Description
When we deploy an adaptation project with BAS IDE with cloud package containing slashes we encounter validation errors in the deployment wizard.

### Steps to Reproduce
Just fill a package name with slashes within the deployment wizard, e.g. `/AUIC0/PRODUIP`.

### Expected results
The deployment wizard should accept package names containing slashes.

### Actual results
Validation errors are displayed within the wizard.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3241, opened 2025-05-14
fix(axios): Disable System info params encoding

avasilev-sap:
Fix for: #3238

* Disable internally the encoding of url parameters set in the `get system info` api call.
* Make a small refactoring - in the `merge app descriptor variant` api call we also disable the encoding.

Testing done:
* Add unit tests.
* Manually verified.

### DIFF
--- a/packages/axios-extension/src/abap/lrep-service.ts
+++ b/packages/axios-extension/src/abap/lrep-service.ts
@@ -1,13 +1,14 @@
-import type { Service } from '../base/service-provider';
-import type { AxiosResponse } from 'axios';
+import type { AxiosResponse, CustomParamsSerializer } from 'axios';
 import { Axios } from 'axios';
-import { LogLevel } from '@sap-ux/logger';
-import type { Logger } from '@sap-ux/logger';
 import { readFileSync } from 'fs';
-import { isAxiosError } from '../base/odata-request-error';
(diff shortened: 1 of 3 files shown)


lfindlaysap (approved):

packages/axios-extension/src/abap/lrep-service.ts
```diff
@@ -133,6 +134,14 @@ function isBuffer(input: string | Buffer): input is Buffer {
     return (input as Buffer).BYTES_PER_ELEMENT !== undefined;
 }
 
+/**
+ * Decodes url parameters.
+ *
+ * @param params An object containing the parameters to be decoded.
+ * @returns The deocded parameters as a string.
```
testojs, line 141:
```suggestion
 * @returns The decoded parameters as a string.
```

testojs (changes requested):

packages/axios-extension/src/abap/lrep-service.ts
avasilev-sap: Fixed in https://github.com/SAP/open-ux-tools/pull/3241/commits/81bdc0b24a4d93456c38d9bdda3be0a6b623c3e2

testojs (approved):
* code looks good
* excellent test coverage
* changeset OK
* not tested manually

PR #3242, opened 2025-05-14
fix(fiori-annotation-api): new cds parser support

voicis:
- Updated test setup to support testing against multiple CDS versions.
- Adapted code to support token structures from the new CDS parser.


Jimmy-Joseph19 (approved):
Overall changes looks good.
Covered with unit test.
Unit test converage due to compiler versions
Changeset found
Did not test it locally
Approving

vadson71 (approved):
Code changes look good
Covered with tests
Didn't test locally

PR #3243, opened 2025-05-14
chore - drop node18 support as it is out of maintenance

devinea:
chore - drop node18 support as it is out of maintenance


donal-tobin-sap (approved):
Changes look good to me.

Tsconfig / es version levels can also be reviewed now with this update.
I believe there may also be esm support changes available.

Klaus-Keller (approved):
Thanks @devinea, changes to node version >=20 look good.

PR #3244, opened 2025-05-15
fix: missing change indicator for control change kind and scrollbar issue

Jimmy-Joseph19:
Fixes for
1. Restricting controls only for change kind 'generic' caused missing change indicators on outline for other unknown control changes which has control id.
2. Quick action scrollbar lost styling when focus from app is lost
Issue 2


815are (approved):
checked `ui-components`:
1. change to pass external class name to section body is correct;
2. changelog persis
3. test is added

I did not test manually

packages/control-property-editor/src/panels/RightPanel.tsx
```diff
@@ -42,6 +42,7 @@ export function RightPanel(): ReactElement {
                 scrollable={true}
                 layout={UISectionLayout.Standard}
                 className="editor__quickactions"
+                bodyClassName="app-panel-scroller"
```
voicis, line 45: There are additional styles also coming with this class and it breaks spacing 
image

Having to set scrollbar styles for every single element seems error prone. For example, styles for properties panel are missing. 
image

It would be better to set them globally. The drawback is that the we can't add hover effects, but the web views in vscode function the same way.

voicis (approved):
- Simple code changes
- Review comments addressed
- Tested manually

PR #3009, opened 2025-03-10
feat(preview-middleware):refactor lrep connectors in preview middleware

heimwege:
Refactoring of the lrep connectors. Needed for #2967

PR #3248, opened 2025-05-19
fix(preview middleware): custom connector for flex changes

heimwege:
#3009 refactored the lrep connectors for preview. This is yet another refactoring.

Instead of using a connector but replacing the callbacks to some non existing endpoint (`FioriToolsNonexistentConnector`) for UI5 versions < 1.72, this PR deletes the whole connector. And the UI5 version is increased to < 1.78 to be aligned with the Fiori Elements writer.

As a bonus in case of a CAP project the ABAP connector (to `/sap/bc/lrep`) will no longer be used to get rid of an error message in the console. This connector is not of use for CAP projects.


devinea (approved):
changes look good and aligns with Fiori Elements writer https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-elements-writer/templates/common/add/webapp/test/flpSandbox.html#L79-L80

heimwege:
@nikmace could you please verify that this adjustment works for you as well?

heimwege:
@devinea @Klaus-Keller there's also the following connector

`{\\"connector\\":\\"LrepConnector\\",\\"layers\\":[],\\"url\\":\\"/sap/bc/lrep\\"}`

(see snapshot). Can we delete this one as well for BTP native (CAP) projects as there will be no calls to `/sap*`?
I always get
image
when starting my playground. Can this connector be the reason? Or does this message just tell me that there are no flex changes? But why is it an error then... 🤔

Klaus-Keller:
> @devinea @Klaus-Keller there's also the following connector
> 
> `{\\"connector\\":\\"LrepConnector\\",\\"layers\\":[],\\"url\\":\\"/sap/bc/lrep\\"}`
> 
> (see snapshot). Can we delete this one as well for BTP native (CAP) projects as there will be no calls to `/sap*`? I always get image when starting my playground. Can this connector be the reason? Or does this message just tell me that there are no flex changes? But why is it an error then... 🤔

Sounds plausible. In the preview middleware, do you know the project type, like CAP?

heimwege:
> > @devinea @Klaus-Keller there's also the following connector
> > `{\\"connector\\":\\"LrepConnector\\",\\"layers\\":[],\\"url\\":\\"/sap/bc/lrep\\"}`
> > (see snapshot). Can we delete this one as well for BTP native (CAP) projects as there will be no calls to `/sap*`? I always get image when starting my playground. Can this connector be the reason? Or does this message just tell me that there are no flex changes? But why is it an error then... 🤔
> 
> Sounds plausible. In the preview middleware, do you know the project type, like CAP?

@Klaus-Keller I thought of using `getProjectType` from project access like

```ts
// this is done in the init of the preview middleware
this.projectType = await getProjectType(process.cwd());
[...]
//and this in the router callback as part of checkDeleteCustomConnector
if (this.projectType === 'CAPJava' || this.projectType === 'CAPNodejs') {
    this.templateConfig.ui5.flex?.splice(0, 1);
    this.logger.debug(`The ABAP connector is not being used because the current project is a CAP project.`);
} else {
    this.logger.debug(`The ABAP connector is being used.`);
}
```

What do you think?

packages/preview-middleware/src/base/flp.ts
```diff
@@ -233,14 +236,31 @@ export class FlpSandbox {
...
      *
+     * @param ui5VersionMajor - the major version of UI5
+     * @param ui5VersionMinor - the minor version of UI5
      * @private
      */
-    private overrideCustomConnector(): void {
-        (this.templateConfig.ui5.flex?.[1] as CustomConnector).applyConnector = 'FioriToolsNonexistentConnector';
-        (this.templateConfig.ui5.flex?.[1] as CustomConnector).writeConnector = 'FioriToolsNonexistentConnector';
+    private checkDeleteConnectors(ui5VersionMajor: number, ui5VersionMinor: number): void {
+        if (ui5VersionMajor === 1 && ui5VersionMinor < 78) {
```
Klaus-Keller, line 248: I'm wondering whether we should also reuse function utility function `isLowerThanMinimalUi5Version()` here

heimwege: I agree but that function is from package `preview-middleware-client` and an import seems to cause issues. The client is copied over to the `preview-middleware` package during build (/dist/client). Not sure what would be needed to re-use that function. Any idea?

```ts
import { isLowerThanMinimalUi5Version } from '@private/preview-middleware-client';
//                                            ^
// TS2307: Cannot find module @private/preview-middleware-client or its corresponding type declarations.
// ESLint: Unable to resolve path to module '@private/preview-middleware-client'. (import/no-unresolved)
```

heimwege: resolving this comment for now but will discuss offline

PR #3249, opened 2025-05-19
feat(generator-adp): Consume FLP generator and Deployment generators

nikmace:
Feat for #3015.
- Adds two confirm prompts for adding flp and deployment configuration.
- Call the two generators from the main ADP generator.


packages/generator-adp/src/translations/generator-adp.i18n.json
```diff
@@ -3,7 +3,11 @@
         "configurationName": "System and Application Selection",
         "configurationDescr": "Configure the system and select an application.",
         "projectAttributesName": "Project Attributes",
-        "projectAttributesDescr": "Configure the main project attributes."
+        "projectAttributesDescr": "Configure the main project attributes.",
+        "flpConfigName": "FLP Configuration",
```
lfindlaysap, line 7:
```suggestion
        "flpConfigName": "SAP Fiori Launchpad Configuration",
```

packages/generator-adp/src/translations/generator-adp.i18n.json
```diff
@@ -3,7 +3,11 @@
...
+        "flpConfigDescr": "Configure Fiori Launchpad settings for {{- projectName}}",
```
lfindlaysap, line 8:
```suggestion
        "flpConfigDescr": "Configure SAP Fiori launchpad settings for {{- projectName}}.",
```

packages/generator-adp/src/translations/generator-adp.i18n.json
```diff
@@ -33,6 +37,8 @@
         "projectFolderPath": "Project Path",
         "ui5VersionLabel": "SAPUI5 Version",
         "ui5VersionTooltip": "Select the SAPUI5 version you want to use to preview your app variant.",
+        "addDeployConfig": "Add deployment configuration",
```
lfindlaysap, line 40:
```suggestion
        "addDeployConfig": "Add deployment configuration",
```
Is this a title,  a button, placeholder?

packages/generator-adp/src/translations/generator-adp.i18n.json
```diff
@@ -33,6 +37,8 @@
...
+        "addFlpConfig": "Add FLP configuration",
```
lfindlaysap, line 41:
```suggestion
        "addFlpConfig": "Add FLP configuration",
```
Is this a title,  a button, placeholder?

packages/generator-adp/src/translations/generator-adp.i18n.json
```diff
@@ -59,6 +65,7 @@
         "appParameterMissing": "Could not create an extension project because the application parameters are missing. Please select a different application.",
         "destinationInfoMissing": "Could not create an extension project because the destination info is missing. Please select a different system.",
         "systemNotFound": "System not found or not supported: {{system}}. Please check the system exists.",
-        "applicationNotFound": "Application not found or not supported: {{appName}}. Please check the application exists."
+        "applicationNotFound": "Application not found or not supported: {{appName}}. Please check the application exists.",
+        "callingDeployGen": "Failed to invoke the Fiori deployment sub-generator."
```
lfindlaysap, line 69:
```suggestion
        "callingDeployGen": "Failed to invoke the SAP Fiori deployment sub-generator."
```
Is there something the user can do? Check the logs? Try again? Contact support?

lfindlaysap (changes requested):
@nikmace, some suggestions and questions.

packages/generator-adp/src/translations/generator-adp.i18n.json
nikmace: It is a label for the confirm prompt: Yes / No

packages/generator-adp/src/translations/generator-adp.i18n.json
lfindlaysap: Thanks. Updated suggestion.

lfindlaysap (changes requested):
@nikmace, thanks for the comments. Updated suggestions.

packages/generator-adp/src/translations/generator-adp.i18n.json
nikmace: This was removed totally

packages/generator-adp/src/translations/generator-adp.i18n.json
nikmace: Changed in: [b8a4488](https://github.com/SAP/open-ux-tools/pull/3249/commits/b8a44880c058526a7c1f1a0f54dc5d5ca4a02569)

packages/generator-adp/src/translations/generator-adp.i18n.json
nikmace: Changed in: https://github.com/SAP/open-ux-tools/pull/3249/commits/765a1e73370efd03a3d0257eab438da57159ad0e#diff-220e9b6afe70b6af8b7e99cfd0d4fb2469fcddcda2f8d60f300120c9cc2ec5c1L4-L71

lfindlaysap (dismissed):

packages/generator-adp/src/utils/subgenHelpers.ts
```diff
@@ -0,0 +1,156 @@
...
+ * @param {ILogWrapper} logger - Logger for info and error output
+ * @param {AppWizard} appWizard - Optional AppWizard instance for displaying UI messages
+ * @param {boolean} debug - Debug flag which composes a generator from OSR.
+ */
+export function addDeployGen(
+    { projectName, targetFolder, client, connectedSystem, destinationName }: DeployGenOptions,
+    composeWith: Generator['composeWith'],
+    logger: ILogWrapper,
+    appWizard: AppWizard,
+    debug: boolean = false
```
cianmSAP: Under what circumstances will this be set to true?

nikmace: For example, you want to debug the **FLP** or the **Deploy** generators from **open-ux-tools** without doing links, or all that stuff to have the latest code from the source and not the bundled `@sap/fiori` generator. Maybe I am wrong here, and it is better to link the FLP or Deploy generators to **tools-suite** and then link `@sap/fiori` generator and then call it with compose from **open-ux-tools**.

So the short answer is that when you want to debug it, you change the false to true manually, and that way you skip the problematic testing in **tools-suite**.

The sub-generators are added as devDependencies.

cianmSAP: I think this could lead to problems. My assumption is this case would be if debug is set that you will have additional low level logs. But we are currently not passing any value from the caller.

Also, as you need to manually change this in the code, at the point could you not just manually update the `composeWith` statement to use `@sap-ux/deploy-config-sub-generator/generators/app`.

If the concern is to make it easy for future developement, I wonder could we look at creating a launch config, in which we add the generator param, and default to `@sap/fiori:deploy-config` if it is not passed,

What do you think?

nikmace: I agree that this is not a good solution with the debug argument. And I will remove this part and just leave `@sap/fiori:deploy-config`.

> just manually update the composeWith statement to use @sap-ux/deploy-config-sub-generator/generators/app

You would also need to point the generator in the `package.json` dependencies: `@sap-ux/deploy-config-sub-generator`, otherwise the require.resolve won't be able to find it.

nikmace: @cianmSAP Removed here: [43fc37b](https://github.com/SAP/open-ux-tools/pull/3249/commits/43fc37bc5262c9ebc389c3faf09e4fffe21836a6)

cianmSAP (approved):
Updateslook good from my side
Adds support to show single set of prompts from specific target (ABAP in this case) in a sub gen flow 
Tests updated

testojs (approved):
* Code looks good
* Changeset OK
* Good test coverage
* Did not tested manually

cianmSAP (approved):
Re-approving

nikmace:
Hello @lfindlaysap, could you please re-review? There were no text changes since your last review.

lfindlaysap (approved):

PR #3253, opened 2025-05-19
fix(preview-middleware) fix template spacing

devinea:
fix(preview-middleware) fix template spacing (avoids breaking test snapshots in tools-suite)

follow-on from https://github.com/SAP/open-ux-tools/pull/3167


Sharmavksh (approved):
=> The change looks fine.
=> Tested the enablement of card generation in local works as expected.

PR #3258, opened 2025-05-21
feat(odata-inq): service filter prompt option

cianmSAP:
- In some cases, the list of services returned from the catalog may require filtering, the new property `serviceFilter` on `ServiceSelectionPromptOptions` will facilitate the filtering of services returned from the catalog, based on the service ID.

### DIFF
--- a/packages/odata-service-inquirer/src/types.ts
+++ b/packages/odata-service-inquirer/src/types.ts
@@ -331,7 +331,7 @@ export type ServiceSelectionPromptOptions = {
      */
     showCollaborativeDraftWarning?: boolean;
     /**
-     * A list of service ids which will be used to filter the service returned from the catalog service
+     * A list of service ids used to filter the catalog results
      */
     serviceFilter?: string[];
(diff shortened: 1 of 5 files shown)


IainSAP (dismissed):
Thanks @cianmSAP 

- Adds new `ServiceSelectionPromptOption` as stated
- Adds test to cover

packages/odata-service-inquirer/src/types.ts
```diff
@@ -330,6 +330,10 @@ export type ServiceSelectionPromptOptions = {
      * This is used to indicate that the service does not support collaborative draft.
      */
     showCollaborativeDraftWarning?: boolean;
+    /**
+     * A list of service ids used to filter the catalog results
```
IainSAP, line 334: Can we add an example here as `service id` is not a property thats specified in the returned service object so a dev coming to this may not know what that it?

IainSAP (approved):
Re-approve

longieirl (approved):
Changes look good.
Changeset is good.
Tests to cover change.

PR #3259, opened 2025-05-21
fix(fiori-annotation-api): replace text content for cds paths

voicis:
Correctly replace `/` with `.` for CDS paths when they are used for text replacement, because there conversion logic from `@sap-ux/cds-odata-annotation-converter` is not called.

### DIFF
--- a/.changeset/smart-onions-mate.md
+++ b/.changeset/smart-onions-mate.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fiori-annotation-api': patch
+---
+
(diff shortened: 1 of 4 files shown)


vadson71 (dismissed):
Code changes look good
Covered by tests
Changeset detected
Approving

vadson71 (approved):
Reapproving additional test

Jimmy-Joseph19 (approved):
Changes looks good.
Covered with unit test
Tested locally. Annotation File updated correctly
Changeset found
Thanks Davis

PR #3260, opened 2025-05-21
Fix/preview middleware/cards generator with static flp html

heimwege:
Currently the virtual card generator endpoint does not work in case the preview uses a (non virtual) local flp html file. This is fixed in this PR.


devinea (approved):
Changes make sense. 👍 
Changeset ✅

PR #3263, opened 2025-05-23
(WIP) feat: Enhance ADP FLP configuration generator

mmilko01:


packages/adp-flp-config-sub-generator/src/app/index.ts
```diff
@@ -186,25 +197,33 @@ export default class extends Generator {
...
+    private async _getAppInbounds(): Promise<ManifestNamespace.Inbound | undefined> {
+        const lrepService = this.provider.getLayeredRepository();
+        const inbounds = (await lrepService.getSystemInfo(undefined, undefined, this.appId)).inbounds as Inbound[];
+
+        if (!inbounds?.length) {
+            return undefined;
+        }
+
+        return filterAndMapInboundsToManifest(inbounds);
+    }
```
nikmace: Could we extract this method to a file in the same package? The reason is to avoid cluttering the generator code with private methods and keep it clean. It could accept a provider as an argument. This could improve testability

packages/adp-flp-config-sub-generator/src/app/index.ts
```diff
@@ -186,25 +197,33 @@ export default class extends Generator {
...
-        const provider = await createAbapServiceProvider(target, requestOptions, false, this.toolsLogger);
-        const variant = await getVariant(this.projectRootPath);
-        const manifestService = await ManifestService.initMergedManifest(
-            provider,
-            this.projectRootPath,
-            variant,
-            this.toolsLogger
-        );
-        return manifestService.getManifest();
+        return await createAbapServiceProvider(target, requestOptions, false, this.toolsLogger);
```
nikmace: This could potentially also be extracted somewhere and just pass all the needed parameters. Also, you do not need this extra `await` because you are not catching anything here

packages/adp-flp-config-sub-generator/src/app/index.ts
```diff
@@ -302,35 +329,20 @@ export default class extends Generator {
...
-                this._abortExecution(t('error.systemNotFoundInStore', { systemUrl: url }));
-                return;
+            } else {
+                this._abortExecution(t('error.systemNotFoundInStore'));
             }
+            return undefined;
         }
 
-        return configuredSystem;
+        return endpoint.Name;
```
nikmace: You could move SystemLookup to a class property and instantiate once:
```ts
private readonly systemLookup = new SystemLookup(this.toolsLogger);
```

For clarity you can simplify this block:
```ts
if (!endpoint?.Name) {
    const message = isBas ? t('error.destinationNotFound') : t('error.systemNotFoundInStore');
    this._abortExecution(message);
    return undefined;
}
```

Also, do you think it would be better to extract `isBas ? target.destination : target.url` outside of this method? So you could just pass the system to `_findSystemName(system: string)`?

packages/adp-flp-config-sub-generator/src/app/index.ts
```diff
@@ -387,6 +399,84 @@ export default class extends Generator {
...
+            if (!this.launchAsSubGen && flpConfigurationExists(this.variant)) {
+                promptOptions.existingFlpConfigInfo.hide = false;
+            }
+            const tileSettingsPrompts = getTileSettingsQuestions(promptOptions);
+            return (await this.prompt(tileSettingsPrompts)) as TileSettingsAnswers & FLPConfigAnswers;
+        }
+        return undefined;
+    }
+
+    private _setTileSettingsPrompts(): void {
```
nikmace, line 376: Missing JSDoc for this method and another one below.

packages/adp-flp-config-sub-generator/src/app/index.ts
```diff
@@ -387,6 +399,84 @@ export default class extends Generator {
...
+                }
+            };
+            if (!this.launchAsSubGen && flpConfigurationExists(this.variant)) {
+                promptOptions.existingFlpConfigInfo.hide = false;
+            }
+            const tileSettingsPrompts = getTileSettingsQuestions(promptOptions);
+            return (await this.prompt(tileSettingsPrompts)) as TileSettingsAnswers & FLPConfigAnswers;
+        }
+        return undefined;
+    }
```
nikmace: Could we improve and streamline this method a bit?
```suggestion
    private async _promptTileActions(): Promise<(TileSettingsAnswers & FLPConfigAnswers) | undefined> {
    if (!this.inbounds) {
        return undefined;
    }

    this._setTileSettingsPrompts();

    const existingFlpConfig = !this.launchAsSubGen && flpConfigurationExists(this.variant);
    const promptOptions = {
        existingFlpConfigInfo: {
            hide: !existingFlpConfig
        }
    };

    const tileSettingsPrompts = getTileSettingsQuestions(promptOptions);
    return await this.prompt(tileSettingsPrompts);
}
```

Also, I do not understand why we need `FLPConfigAnswers` in the return type if you are getting only `TileSettingsAnswers`? And why we cast here to the same joint type?

packages/adp-flp-config-sub-generator/src/app/appWizzardCache.ts
```diff
@@ -0,0 +1,82 @@
+import { getHostEnvironment, hostEnvironment, type ILogWrapper } from '@sap-ux/fiori-generator-shared';
+import type { AdpAppWizardCache } from './types';
+import type { AppWizard } from '@sap-devx/yeoman-ui-types';
+import { t } from '../utils/i18n';
+
+const ADP_FLP_CONFIG_CACHE = '$adp-flp-config-cache';
+// Type alias for the AppWizard object with the cache object.
+export type AppWizardCache = AppWizard & { [ADP_FLP_CONFIG_CACHE]?: AdpAppWizardCache };
+const hostEnv = getHostEnvironment();
+
```
nikmace: If we are following the app-gen colleagues' implementation here, there are a couple of things:
- The file name should be just with one 'z' - `appWizardCache.ts`.
- It should be moved under the utils folder.
- The `AdpAppWizardCache` interface is only applicable to this generator. So we can remove adp part. We also know the context - app wizard so that is redundant, maybe. What I am trying to say you can name it `State` like in other packages which have app wizard cache.

packages/flp-config-sub-generator/src/app/index.ts
```diff
@@ -143,12 +143,10 @@ export default class extends Generator {
         }
 
         const silentOverwrite = this.options.overwrite;
-        let questions: Question[] = (await getPrompts(inbounds, undefined, {
+        let questions: Question[] = (await getPrompts(inbounds, {
             silentOverwrite,
             inboundId: { hide: true },
-            emptyInboundsInfo: { hide: true },
-            additionalParameters: { hide: true },
-            createAnotherInbound: { hide: true }
```
cianmSAP, line 151: If `emptyInboundsInfo` & `createAnotherInbound` have been removed, can their replacements or any additional prompts be added here so they are hidden?

packages/flp-config-inquirer/src/prompts/questions/basic.ts
```diff
@@ -53,10 +64,28 @@ export function getActionPrompt(isCLI: boolean, options?: ActionPromptOptions):
...
+            }
+
+            if (!inbounds || !answers.semanticObject) {
+                return true;
+            }
+
+            const isDuplicate = Object.values(inbounds).some(
+                (inbound: any) => inbound.semanticObject === answers.semanticObject && inbound.action === val
+            );
+            return isDuplicate ? t('validators.duplicateInbound') : true;
```
cianmSAP, line 104: Correct me if I'm wrong, but this appears to be the same logic with which we show the overwrite prompt? 
So this would be a blocking change for the user in our case

Does your adp generator use the overwrite prompt, or hide it? 

Can we put this additional logic behind an option so it is not ran by default, but set in adp generator so you run this logic if it is required

packages/flp-config-inquirer/src/prompts/questions/basic.ts
```diff
@@ -157,7 +191,36 @@ export function getSubTitlePrompt(
...
+    };
+}
+
+/**
+ * Creates the 'icon' prompt for FLP configuration.
+ *
+ * @param {IconPromptOptions} [options] - Optional configuration for the icon prompt, including default values.
+ * @returns {FLPConfigQuestion} The prompt configuration for the icon.
+ */
+export function getIconPrompt(options?: IconPromptOptions): FLPConfigQuestion {
```
cianmSAP: Can this go to `advanced.ts`, we do not have a requirement for this prompt

packages/flp-config-inquirer/src/prompts/questions/basic.ts
```diff
@@ -157,7 +191,36 @@ export function getSubTitlePrompt(
...
+    return {
+        name: promptNames.icon,
+        type: 'input',
+        guiOptions: {
+            breadcrumb: t('prompts.icon')
```
nikmace: This can just be true as it is the same as message property

packages/i18n/src/write/properties/create.ts
```diff
@@ -22,11 +22,46 @@ export async function createPropertiesI18nEntries(
...
+ * @param fs Optional mem-fs-editor instance for file operations.
+ * @returns A promise that resolves to the written content or void.
+ */
+async function createNewI18nFile(i18nFilePath: string, root?: string, fs?: Editor): Promise<string | void> {
+    let content = '# Resource bundle \n';
+    if (root) {
+        content = `# This is the resource bundle for ${basename(root)}\n`;
+    }
+    return writeFile(i18nFilePath, content, fs);
+}
```
nikmace: Could these be updated to follow the style of the rest of the jsdocs in the repo? Example below
An option is to add a type to the jsdoc property.
```suggestion
/**
 * Creates or overwrites i18n entries in the `i18n.properties` file.
 *
 * @param i18nFilePath - Absolute path to the `i18n.properties` file.
 * @param newI18nEntries - New i18n entries to be added or updated.
 * @param keysToRemove - Keys to be removed from the properties file.
 * @param root - Project root, optionally used in the comment if the file is newly generated.
 * @param fs - Optional `mem-fs-editor` instance. If provided, its API is used instead of Node's `fs`.
 * @returns Promise that resolves when the operation is complete.
 */
export async function createOrReplaceI18nEntries(
    i18nFilePath: string,
    newI18nEntries: NewI18nEntry[],
    keysToRemove: string[] = [],
    root?: string,
    fs?: Editor
): Promise<void> {
    if ((!fs && !(await doesExist(i18nFilePath))) || (fs && !fs.exists(i18nFilePath))) {
        await createNewI18nFile(i18nFilePath, root, fs);
    }
    await replaceI18nProperties(i18nFilePath, newI18nEntries, keysToRemove, fs);
}

/**
 * Creates a new i18n.properties file with a default or root-specific comment.
 *
 * @param i18nFilePath Absolute path to the i18n.properties file.
 * @param root Optional project root, used in the comment if provided.
 * @param fs Optional mem-fs-editor instance for file operations.
 * @returns A promise that resolves to the written content or void.
 */
async function createNewI18nFile(i18nFilePath: string, root?: string, fs?: Editor): Promise<string | void> {
    let content = '# Resource bundle \n';
    if (root) {
        content = `# This is the resource bundle for ${basename(root)}\n`;
    }
    return writeFile(i18nFilePath, content, fs);
}
```

packages/flp-config-sub-generator/src/app/index.ts
mmilko01: I added the two new prompts here (existingFlpConfigInfo, icon) with commit: 122d49024480d64ca9a2be675c3d03aea66713bd

packages/flp-config-inquirer/src/prompts/questions/basic.ts
mmilko01: Thanks for pointing this out, I missed that this would collide with the `overwrite` logic. I added a new option that needs to be set explicitly to  true for this validation to be executed: 122d49024480d64ca9a2be675c3d03aea66713bd

packages/flp-config-inquirer/src/prompts/questions/basic.ts
mmilko01: Moved with commit: 122d49024480d64ca9a2be675c3d03aea66713bd

nikmace (changes requested):

packages/adp-tooling/package.json
```diff
@@ -35,6 +35,8 @@
         "!dist/**/*.map"
     ],
     "dependencies": {
+        "@sap-ux/fiori-generator-shared": "workspace:*",
```
nikmace: This is not used anywhere in `adp-tooling` and should be removed. Don't think `adp-tooling` should have a dep to this package in any case (possible circular dep)

mmilko01: Removed with commit: bd356d36d9056a8cc227fb43567d5bc2241d275f

packages/i18n/src/write/properties/create.ts
mmilko01: Fixed with commit: bd356d36d9056a8cc227fb43567d5bc2241d275f

packages/flp-config-inquirer/src/prompts/questions/basic.ts
mmilko01: Fixed with commit: bd356d36d9056a8cc227fb43567d5bc2241d275f

packages/generator-adp/src/app/index.ts
```diff
@@ -424,6 +438,33 @@ export default class extends Generator {
...
+
+        if (!deployPageExists) {
+            updateWizardSteps(
+                this.prompts,
+                getDeployPage(),
+                t('yuiNavSteps.projectAttributesName'),
+                this.attributeAnswers.addDeployConfig
+            );
+        }
+    }
```
nikmace: I think this should be moved to the same file: `steps.ts` and pass `prompts['items']` as an argument. The interface for this is `IPrompt[]`

cianmSAP (dismissed):
looks good from my side - focused on flp config inquirer

cianmSAP (approved):
Re-approving

packages/i18n/src/write/properties/create.ts
```diff
@@ -1,16 +1,16 @@
...
 import type { Editor } from 'mem-fs-editor';
 
 /**
  * Creates new i18n entries in `i18n.properties` file.
  *
- * @param i18nFilePath absolute path to `i18n.properties` file
- * @param newI18nEntries new i18n entries that will be maintained
- * @param root project root optionally used in comment if file is newly generated
- * @param fs optional `mem-fs-editor` instance. If provided, `mem-fs-editor` api is used instead of `fs` of node
+ * @param {string} i18nFilePath absolute path to `i18n.properties` file
```
voicis: Types in JSDoc are redundant.

packages/i18n/src/write/utils/index.ts
```diff
@@ -33,3 +33,54 @@ export async function writeToExistingI18nPropertiesFile(
...
+/**
+ * Overwrites an existing i18n.properties file by removing specified keys and adding new i18n entries.
+ *
+ * @param i18nFilePath - The path to the i18n file.
+ * @param newI18nEntries - Array of new i18n entries to add.
+ * @param keysToRemove - Array of keys to remove from the file.
+ * @param fs - Optional mem-fs-editor instance.
+ * @returns Promise<void>
+ */
+export async function replaceI18nProperties(
```
voicis: I don't see a reason to have 2 different internal functions here, because they are very similar. It would be better to extend `writeToExistingI18nPropertiesFile` to also allow key deletion and `createPropertiesI18nEntries` pass an empty value used for deletion.

packages/i18n/src/write/utils/index.ts
```diff
@@ -33,3 +33,54 @@ export async function writeToExistingI18nPropertiesFile(
...
+            // Skip this line (the key) and any empty lines below
+            i++;
+            while (i < lines.length && lines[i].trim() === '') {
+                i++;
+            }
+        } else {
+            filteredLines.push(lines[i]);
+            i++;
+        }
+    }
```
voicis: This will not remove more complicated key/value pair constructs and probably there are other potential issues. You should not reimplement parsing logic and use parsed content from `parseProperties` (that should be sufficient, but another alternative is `propertiesToI18nEntry` function) and build text edits and then apply them similarly to the code here https://github.com/SAP/open-ux-tools/blob/065f0ccee15e36fbf3e8f9e19378922d9a02b1cf/packages/i18n/src/write/cap/json.ts#L65-L91

packages/i18n/src/write/properties/create.ts
```diff
@@ -22,11 +22,47 @@ export async function createPropertiesI18nEntries(
...
+ *
+ * @param {string} i18nFilePath - Absolute path to the `i18n.properties` file.
+ * @param {NewI18nEntry[]} newI18nEntries - New i18n entries to be added or updated.
+ * @param {string []} keysToRemove - Keys to be removed from the properties file.
+ * @param {string} root -Project root, optionally used in the comment if the file is newly generated.
+ * @param {Editor} fs - Optional `mem-fs-editor` instance. If provided, its API is used instead of Node's `fs`.
+ * @returns Promise that resolves when the operation is complete.
+ * @description If `i18n.properties` file does not exits, it tries to create.
+ */
+export async function createOrReplaceI18nEntries(
```
voicis: `replace` is a bit misleading as it implies that something will be added at the same place, which is not the case here. Maybe something similar to `removeAndCreateI18nEntries` would be better fit.

voicis (changes requested):
This was not modified in this PR, but`getFlpI18nKeys` should follow best practices and also include the text type annotations for the new entries.

packages/create/src/cli/add/navigation-config.ts
```diff
@@ -95,25 +98,29 @@ async function addInboundNavigationConfig(basePath: string, simulate: boolean):
...
     basePath: string,
     isAdp: boolean,
     fs: Editor,
     logger: ToolsLogger
-): Promise<Manifest> {
+): Promise<ManifestNamespace.Inbound | undefined> {
     if (isAdp) {
-        return retrieveMergedManifest(basePath, logger);
+        const appId = (await getVariant(basePath)).reference;
+        const { target, ignoreCertErrors = false } = await getAdpConfig(basePath, join(basePath, FileName.Ui5Yaml));
```
heimwege: Why hard code to UI5.yaml? Can't I use another yaml config file via `[path]` variable of the create command?

packages/create/package.json
```diff
@@ -47,6 +47,7 @@
         "@sap-ux/system-access": "workspace:*",
         "@sap-ux/ui5-config": "workspace:*",
         "@sap-ux/flp-config-inquirer": "workspace:*",
+        "@sap-ux/axios-extension": "workspace:*",
```
heimwege: Why is the `axios-extension` a new dependency? I don't see where it is being used in the `create` package.

packages/create/tsconfig.json
```diff
@@ -21,6 +21,9 @@
     {
       "path": "../app-config-writer"
     },
+    {
+      "path": "../axios-extension"
+    },
```
heimwege: Why is the `axios-extension` a new dependency? I don't see where it is being used in the `create` package.

heimwege (commented):
see comments for create package

packages/flp-config-inquirer/src/translations/flp-config-inquirer.i18n.json
```diff
@@ -1,21 +1,30 @@
...
-        "emptyInboundsInfo": "Original application target mapping and the default parameters can be found in Fiori reference library under Implementation Information -> Configuration -> Target Mapping(s) in",
         "semanticObject": "Semantic Object",
         "action": "Action",
         "title": "Title",
         "subTitle": "Subtitle",
         "additionalParameters": "Parameters",
-        "createAnotherInbound": "Do you want to add or edit another inbound?"
+        "createAnotherInbound": "Do you want to add or edit another inbound?",
+        "icon": "Icon",
+        "existingFLPConfig": "Keep in mind that this project already has created a Fiori Launchpad Configuration. If you create a new configuration, it will completely overwrite the existing one.",
```
lfindlaysap:
```suggestion
        "existingFLPConfig": "This project already has a SAP Fiori launchpad configuration. If you create a new configuration, it will overwrite the existing one.",
```

packages/flp-config-inquirer/src/translations/flp-config-inquirer.i18n.json
```diff
@@ -1,21 +1,30 @@
...
+        "tileHandlingAction": "Choose a Tile Handling Action",
+        "copyFromExisting": "Copy Configurations From an Existing Inbound?"
```
lfindlaysap:
```suggestion
        "copyFromExisting": "Copy Configurations from an Existing Inbound?"
```

packages/flp-config-inquirer/src/translations/flp-config-inquirer.i18n.json
```diff
@@ -1,21 +1,30 @@
...
     },
     "tooltips": {
         "inboundId": "Choose the inbound ID that will be used to set the FLP tile configuration",
-        "additionalParameters": "Enter the value for the parameters in the Target Mappings section from the URL mentioned above"
+        "additionalParameters": "Enter the value for the parameters in the Target Mappings section from the URL mentioned above",
+        "semObjectActionDuplication": "When generating a new tile, you must modify the Semantic Object value or the Action value, or both. You can use the default suggestions, which include a '_New' suffix for each value, or choose to introduce your specific designations."
```
lfindlaysap:
```suggestion
        "semObjectActionDuplication": "When generating a new tile, you must modify the Semantic Object value, the Action value, or both. You can use the default suggestions, which include a '_New' suffix for each value or use your own designations."
```

packages/flp-config-inquirer/src/translations/flp-config-inquirer.i18n.json
```diff
@@ -1,21 +1,30 @@
...
     },
     "validators": {
         "inboundConfigKeyExists": "An inbound configuration with the key: '{{inboundKey}}' is already defined.",
+        "invalidParameterString": "Parameter string is not a valid JSON.",
```
lfindlaysap:
```suggestion
        "invalidParameterString": "The parameter string is not valid JSON.",
```

packages/flp-config-inquirer/src/translations/flp-config-inquirer.i18n.json
```diff
@@ -1,21 +1,30 @@
...
         "flpConfigOverwrite": "Overwrite existing configuration?",
-        "noInboundKeysAreFound": "Original application target mapping and the default parameters can be found in Fiori reference library under Implementation Information -> Configuration -> Target Mapping(s) in application page."
+        "duplicateInbound": "The base application has an existing inbound that matches the values for Semantic Object and Action. To create a new tile you must modify the Semantic Object value or  the Action value, or both."
```
lfindlaysap:
```suggestion
        "duplicateInbound": "The base application has an existing inbound that matches the values for Semantic Object and Action. To create a new tile you must modify the Semantic Object value, the Action value, or both."
```

packages/flp-config-inquirer/src/translations/flp-config-inquirer.i18n.json
```diff
@@ -1,21 +1,30 @@
...
+    },
+    "choices": {
+        "replaceOriginalTile": "Replace Base Application's Existing Tile(s)",
```
lfindlaysap:
```suggestion
        "replaceOriginalTile": "Replace Base Application's Existing Tiles",
```
We must either have a separate text for singular and plural or decide on one.

packages/generator-adp/src/translations/generator-adp.i18n.json
```diff
@@ -4,8 +4,8 @@
         "configurationDescr": "Configure the system and select an application.",
         "projectAttributesName": "Project Attributes",
         "projectAttributesDescr": "Configure the main project attributes.",
-        "flpConfigName": "SAP Fiori Launchpad Configuration",
-        "flpConfigDescr": "Configure SAP Fiori launchpad settings for {{- projectName}}.",
+        "flpConfigName": "Fiori Launchpad Configuration",
```
```suggestion
        "flpConfigName": "SAP Fiori Launchpad Configuration",
```

packages/generator-adp/src/translations/generator-adp.i18n.json
```diff
@@ -4,8 +4,8 @@
...
+        "tileSettingsName": "Fiori Launchpad Configuration - Tile Settings for {{- projectName}}",
```
```suggestion
        "tileSettingsName": "SAP Fiori Launchpad Configuration - Tile Settings for {{- projectName}}",
```

lfindlaysap (changes requested):
@mmilko01, here are my suggestions.

lfindlaysap (approved):

voicis (approved):
- Changes to `i18n` package looks good
- Review comments addressed
- Good test coverage
- Did not test manually

nikmace (approved):
Changes look good
Changeset looks good
Great coverage
Did not test manually, but @testojs did
🎆 👏🏻

cianmSAP (approved):

heimwege (approved):
re-approve

PR #3266, opened 2025-05-23
feat(fiori-sub-gen): adds appGenInfo.json file

cianmSAP:
- Adds `appGenInfo.json` to all generated apps 

@todo add more specific tests
update telemetry to read `appGenInfo.json` instead of README

### DIFF
--- a/packages/fiori-app-sub-generator/src/types/common.ts
+++ b/packages/fiori-app-sub-generator/src/types/common.ts
@@ -1,3 +1,4 @@
+import type { ReadMe } from '@sap-ux/fiori-generator-shared';
 import type {
     DatasourceType,
     OdataServicePromptOptions,
@@ -53,4 +54,22 @@ export interface GenerateLaunchConfigOptions {
     flpAppId?: string;
     /** Optional OData service version (e.g. v2 or v4). */
     odataVersion?: OdataVersion;
+}
+
+/**
+ * Metadata for an ABAP CSN (Core Schema Notation)
+ */
+export interface AbapCsn {
+    packageUri: string;
+    csnName: string;
+    serviceNameCsn: string;
+    datasourceKey?: string;
+}
+
+/**
+ * Interface for .appGenInfo file
+ */
+export interface AppGenInfo {
+    generationParameters?: Partial<ReadMe>;
+    abapCSN?: AbapCsn[];
 }
--- a/packages/fiori-app-sub-generator/src/types/state.ts
+++ b/packages/fiori-app-sub-generator/src/types/state.ts
@@ -10,7 +10,7 @@ import type {
     OdataVersion
 } from '@sap-ux/odata-service-inquirer';
 import type { ApiHubType, SapSystemSourceType } from '../types/constants';
-import type { Script } from './common';
+import type { Script, AbapCsn } from './common';
 
 export interface Project {
     targetFolder: string;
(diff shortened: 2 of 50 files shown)


packages/fiori-app-sub-generator/src/types/state.ts
```diff
@@ -125,6 +125,10 @@ export interface State {
      * The template selected by the user, maybe FE or FF
      */
     floorplan: Floorplan;
+    /**
+     * ABAP CSN (Core Schema Notation) - for writing to .appGenInfo.json
+     */
+    abapCsn?: AbapCsn[];
```
cianmSAP, line 131: My assumption is this should not be part of the state - @IainSAP can you recommend a more appropriate home for this. 
For context, it is passed to app gen (from adaptors) and written directly to the `.appGenInfo.json` file

IainSAP: What is .appGenInfo.json, where is it defined and what is it purpose? Is there a writer for it?

packages/fiori-app-sub-generator/src/types/common.ts
```diff
@@ -54,3 +55,21 @@ export interface GenerateLaunchConfigOptions {
     /** Optional OData service version (e.g. v2 or v4). */
     odataVersion?: OdataVersion;
 }
+
+/**
+ * Metadata for an ABAP CSN (Core Schema Notation)
+ */
+export interface AbapCsn {
```
IainSAP, line 62: Are we really using this in this module. Shouldnt it be defined in the writer?

IainSAP:
@cianmSAP Can you add a link to the features mentioned in the description please

packages/fiori-app-sub-generator/src/types/common.ts
```diff
@@ -54,3 +55,21 @@ export interface GenerateLaunchConfigOptions {
...
+    packageUri: string;
+    csnName: string;
+    serviceNameCsn: string;
+    datasourceKey?: string;
+}
+
+/**
+ * Interface for .appGenInfo file
+ */
+export interface AppGenInfo {
```
IainSAP, line 72: Maybe this should be part of the read me API in fiori-generator-shared.

packages/fiori-app-sub-generator/src/types/state.ts
IainSAP: On review, I think this is actually app gen info additional entries. I would propose that we define a new state property `appGenInfo` based on the existing read me type (we used to have a readMe property for this exact case where adaptors or extensions wanted to add additional information). Then adaptors can add entries (use `additionalEntries`) into the state via this `appGenInfo` property. This property `abapCsn` appears to be for informational purposes only, we dont use it to generate anything AFAIK and so should be added in the adaptor and then will get written via the `generateAppGenInfo` ( maybe renamed from `generateReadMe`)  without any code  in the generators.

packages/fiori-generator-shared/src/app-gen-info.ts
```diff
@@ -0,0 +1,28 @@
...
+    // N.B. This function must stay at this level in the directory structure, i.e one level below 'templates'
+    // Apply the configuration to generate the README file
+    const templateSourcePath = join(__dirname, '../templates/README.md');
+    const templateDestPath = `${destPath}/README.md`;
+
+    const { additionalEntries, ...appGenInfoCore } = appGenInfo;
+
+    fs.writeJSON(`${destPath}/.appGenInfo.json`, {
+        generationParameters: appGenInfoCore,
+        ...additionalEntries
```
IainSAP, line 23: Wondering should these be at the root or under `additionalEntries` or some such property : `additionalParameters` ?

PR #3267, opened 2025-05-26
fix: service addition error

broksy:
Do not stop service from being added in case if existing services are without annotation `dataSources` in manifest.
Previously error was thrown: `Cannot read properties of undefined (reading 'settings')`.


devinea (approved):
Code changes are clear and covered by test.
changeset ✅

PR #3269, opened 2025-05-27
fix: Remove the already generated dependencies for card generator

Sharmavksh:
=> Remove the '@sap-ux/cards-editor-middleware' from dev dependencies of package.json in case if card generation is already enabled for an application.
=> Remove the "sap-cards-generator" middleware from custom middleware for applications having card generation enabled already.


packages/app-config-writer/src/cards-config/index.ts
```diff
@@ -67,9 +76,16 @@ async function updatePackageJson(basePath: string, fs: Editor, yamlPath?: string
...
         '/test/flpCardGeneratorSandbox.html';
     const cliForPreview = await getCLIForPreview(basePath, yamlPath ?? '', fs);
+    const dependencyName = '@sap-ux/cards-editor-middleware';
 
     packageJson.scripts ??= {};
     packageJson.scripts['start-cards-generator'] = `${cliForPreview} --open '${cardGeneratorPath}${intent}'`;
+
+    if (packageJson.devDependencies?.[dependencyName]) {
+        delete packageJson.devDependencies[dependencyName];
+        console.log(`Dependency ${dependencyName} removed from devDependencies.`);
```
chaudhary1-rohit, line 86: Can you please replace console.log with logger?

chaudhary1-rohit (changes requested):

packages/app-config-writer/src/cards-config/index.ts
heimwege: Also please enhance the log message so that the user knows that this is no issue but a service from us. So there's no need to worry 😃 
Something like
`Dependency ${dependencyName} removed from devDependencies as it is no longer needed.`

packages/app-config-writer/src/cards-config/index.ts
```diff
@@ -25,6 +26,13 @@ async function updateMiddlewareConfigWithGeneratorPath(
 ): Promise<void> {
     const ui5YamlFile = yamlPath ? basename(yamlPath) : FileName.Ui5Yaml;
     const ui5YamlConfig = await readUi5Yaml(basePath, ui5YamlFile, fs);
+
+    try {
+        ui5YamlConfig.removeCustomMiddleware('sap-cards-generator');
```
heimwege, line 31:
```suggestion
        ui5YamlConfig.removeCustomMiddleware('sap-cards-generator');
        logger?.info(`Removed 'sap-cards-generator' middleware configuration from ${ui5YamlFile}. This has been integrated into fiori-tools-preview / preview-middleware and is no longer needed`);
```

chaudhary1-rohit (approved):
Thanks for making the change, looks good to me.

Issue #3274, opened 2025-05-28
BUG - fiori-app-sub-generator : headless throws when specifying destination name


IainSAP:
### Description

Using the an app config file with the `service.destination` property set @sap/fiori:headless throws an error.

Since only the destination name is set in headless input files and not the rest of the destination properties, some of which are mandatory, an undefined ref throws when determining the destination type for telemetry.

Since connected system is not support by headless we should not set this property at all as the assumption in the generator is that this contains an active `ServiceProvider` and all the mandatory destination properties.

Steps to Reproduce

Run yo @sap/fiori:headless with the above config file.
An error is thrown and no output files generated.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3275, opened 2025-05-28
fix(fiori-app-sub-generator): Setting destination in headless app config throws

IainSAP:
https://github.com/SAP/open-ux-tools/issues/3274

- Removes setting of `ConnectedSystem` from headless transform as most of the mandatory properties are not set. Instead only the destination name is considered]
- Adds test to cover


docirl (dismissed):
Changes look good, thanks for the fix
tests updated to cover fix
⚠️ changeset is missing

PR #3277, opened 2025-05-29
feat: poc as example for search and loader using combobox

815are:
poc as example for search and loader using combobox


broksy (approved):
Changeset reflects to the changes made. Good test coverage.
Approved from my side (did not test manually).

PR #3278, opened 2025-05-29
fix(fiori-gen-shared): loop through workspace folders to find correct default

cianmSAP:
- Update `getDefaultTargetFolder` function to return the first folder found in the workspace which meets the criteria i.e `uri.scheme === 'file'`, rather than checking the first folder and going to the default if it is not suitable

### DIFF
--- a/.changeset/serious-mice-leave.md
+++ b/.changeset/serious-mice-leave.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/fiori-generator-shared': patch
+---
+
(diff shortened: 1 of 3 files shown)


IainSAP (approved):
Thanks @cianmSAP 

- Updates are clear
- Tests updated to cover additional checks

docirl (approved):
Changes look good
Tests added/updated

devinea (approved):
clear and tested. Changeset ✅

Issue #3264, opened 2025-05-23
FEATURE - Support `NODE_TLS_REJECT_UNAUTHORIZED=0` for selected `odata-service-inquirer` and `abap-deploy-config-inquirer` prompts


IainSAP:
### Description

In some cases customer are unable to develop Fiori applications using systems that do not yet have properly configured certificates. In some limited cases we should allow the generation of apps to continue if the environment variable  `NODE_TLS_REJECT_UNAUTHORIZED` is set to `0`. 

This should include a clear and obvious warning about the risk of using unauthorized certificates.

The setting should be warned and applied to the following prompts :

- `@sap-ux/odata-service-inquirer` 
    - system selection (Abap on Premise)
    - service url

-  `@sap-ux/abap-deploy-config-inquirer`
    - package selection from backend (Abap on prem)
    - transport selection from backend

For Deploy Config, warnings will only be shown in the logs when the setting is applied. When cert errors are returned from the package/transport calls, a validation GA link will be shown to inform the user about how to fix the cert issue.

### Technical Design

Enhance the `odata-serice-inquirer` module to explicitly apply the `NODE_TLS_REJECT_UNAUTHORIZED` env setting where applicable,

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I am attempting to use the odata service inquirer prompts to connect to an Abap on premise system or service url endpoint which are returning certificate errors and I have configured the node env var `NODE_TLS_REJECT_UNAUTHORIZED=0`
**when** I enter the URL
**then** A detailed warning is shown that the cert errors are being ignored but have been logged.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

IainSAP:
Reopening as an edge case (updating system credentials) does not show the warning.

PR #3285, opened 2025-05-30
fix(odata-service-inquirer): Adds node TLS warn to password prompt

IainSAP:
https://github.com/SAP/open-ux-tools/issues/3264

- Adds the node tls cert warning to the password prompt for display in an edge scenario where credentials are being updated. Also prevents showing twice if new system flow for Abap on Prem
- Adds tests to cover


kjose90 (approved):
Ty - update looks good - tests & changeset ✅

PR #3286, opened 2025-06-02
fix: path issue for windows

Jimmy-Joseph19:
CDS Compiling fails after node update in windows due to path issue


packages/project-access/src/index.ts
```diff
@@ -58,3 +58,4 @@ export { execNpmCommand } from './command/npm-command';
 export * from './types';
 export * from './library';
 export { hasDependency } from './project';
+export * from './path';
```
Klaus-Keller: `export *` is a bit dangerous on root level `index.ts`, later added exported functions in `path.ts` are automatically exposed on the public interface of this module. 
Also, can we move this file to folder `file`, then export it from `file/index.ts` and then add it here to `export { getFilePaths } from './file';`?

PR #3287, opened 2025-06-02
fix(app-config-writer): cards generator log message

heimwege:
Do not show warning in log in case the old card generator middleware has not been found


Sharmavksh (approved):
The change looks good to me.

ytpo-lyne (approved):
LGTM.

- changes straight forward 
- changeset exists 
- did not test manually

Issue #3291, opened 2025-06-03
BUG - UI Service Generator - content endpoint does not use correct accept header


docirl:
### Related Feature
Feature request: _issue number_

### Description
The generator service in axios extension, when creating the service instance, is using a hard coded accept header for the request. This value has now changed on backend systems and requests are failing.

The accept content value should be used from the service generator config that is returned for the selected business object.

### Steps to Reproduce
1. Open UI Service Generator from Fiori tools
2. Select a system with 2508 release
3. Requests to get suggested content and generation fail 

### Expected results
Api requests and service generation succeed as expected

### Actual results
Api requests fail due to wrong content type used in accept header

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

docirl:
done with https://github.com/SAP/open-ux-tools/pull/3292

PR #3292, opened 2025-06-03
fix(ui-service-generator): use accept header for content request from generator config

docirl:
https://github.com/SAP/open-ux-tools/issues/3291

- hardcoded content type is no longer valid and has changed
- use content type from generator config retrieved in generator flow

### DIFF
--- a/packages/axios-extension/src/abap/abap-service-provider.ts
+++ b/packages/axios-extension/src/abap/abap-service-provider.ts
@@ -250,7 +250,7 @@ export class AbapServiceProvider extends ServiceProvider {
      * @returns the type of the config link from config
      */
     private getContentAcceptType(config: GeneratorEntry): string {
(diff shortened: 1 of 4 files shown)


packages/axios-extension/src/abap/abap-service-provider.ts
```diff
@@ -243,6 +243,19 @@ export class AbapServiceProvider extends ServiceProvider {
...
     }
 
+    /**
+     * Get the content type from the generator config.
+     *
+     * @param config - generator config
+     * @returns the type of the content link from service generator config
+     */
+    private getContentType(config: GeneratorEntry): string {
+        const contentEndpoint = config.link.find(
```
longieirl: Do you need to make this `config.link?.find` safe?

docirl: It this config is returned from earlier in the flow it will contain an array of links, but good to add ? to be safe. Thanks

PR #3296, opened 2025-06-04
feat(fiori-annotation-api): write using new annotation syntax

voicis:
Write CDS annotations using new annotation syntax (without `![]` escaping).

Also fixed issues with indentation and added trailing comma for inserts.


vadson71 (approved):
Re-approving sonar fixes

PR #3298, opened 2025-06-04
fix(fiori-gen): update externalParameters on appgeninfo type

cianmSAP:
On `AppGenInfo`, the `externalParameters` property should be just the type `ExternalParameters` not an array


IainSAP (approved):
THanks @cianmSAP 

- Small change in output structure
- Tests updated to align

PR #3299, opened 2025-06-04
Reuse existing sap.cloud.service when adding config

schiwekM:
Hi colleagues,

currently when adding a Fiori app and the managed approuter config is not yet present, Fiori tools adds the config. However when there is already an existing destination with 'sap.cloud.service' this is not reused but instead the name for the new destinations is generated.

This PR fixes this gap, that first an existing `sap.cloud.service` name will be used before generating a new one. 

In addition the generation now strips away dashes ('-') from the name as those are apparently not allowed.

BR,
Marten


longieirl (commented):
Changes look good. Just add a changeset by running `pnpm changeset` and I can approve.
Thanks.

PR #3301, opened 2025-06-05
fix(preview-middleware): only parse changes

voicis:
Preview middleware was logging the following warnings when trying to parse non change files as JSON.
```
warn fiori-tools-preview Unexpected token '<', "<edmx:Edmx"... is not valid JSON
warn fiori-tools-preview Unexpected token '<', "<!-- Use s"... is not valid JSON
```

### DIFF
--- a/.changeset/chilled-hounds-travel.md
+++ b/.changeset/chilled-hounds-travel.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/preview-middleware': patch
+---
+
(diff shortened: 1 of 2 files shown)


marufrasully (approved):
thanks @voicis
* changes is straightforward
* did not test it locally

815are (approved):
checked:
1. code changes easy to understand
2. changelog persists

I did not test manually

Issue #3220, opened 2025-05-01
TBI - Align @ui5/cli with latest v4 release


longieirl:
### Description (include screenshots)
V4 of @ui5/cli is now available, https://sap.github.io/ui5-tooling/v4/ there are changes required to support this change.

1. update package.json to consume "@ui5/cli": "^4.0.0"
2. Specification `specVersion: "3.1"` in the respective ui5 yaml configurations can remain at v3

For more info;

https://sap.github.io/ui5-tooling/v4/updates/migrate-v4/ 

Other impacts;

- Plugins like cds-plugin-ui5 , ui5-task-flatten-library need to be updated.
- Fiori pipeline impacts (internal dependency on FIORI-PIPELINE)

### Value
Align with latest version.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [x] PoC of migrating existing project generated using SAP Fiori tools and validating if any issues found
- [x] Update templates and default ui5 versions
- [x] Ensure specification version is not changed
- [x] Update snapshots/tests

longieirl:
One issue found;

When a project is bumped to the latest using;
```bash
npm i --save-dev @ui5/cli@latest
```

and the command `npm run build` is run, there is a new warning regarding locales `warn builder:processors:manifestEnhancer`, as shown;

```
info ProjectBuilder Preparing build for project project4
info ProjectBuilder   Target directory: dist
info ProjectBuilder Cleaning target directory...
info Project 1 of 1: ❯ Building application project project4...
info project4 › Running task escapeNonAsciiCharacters...
info project4 › Running task replaceCopyright...
info project4 › Running task replaceVersion...
info project4 › Running task minify...
info project4 › Running task enhanceManifest...
warn builder:processors:manifestEnhancer /resources/project4/manifest.json: Generated supported locales ('') for bundle 'i18n/i18n.properties' do not contain default fallback locale 'en'. Either provide a properties file for 'en' or configure another available fallbackLocale
warn builder:processors:manifestEnhancer /resources/project4/manifest.json: Generated supported locales ('') for bundle 'i18n/i18n.properties' do not contain default fallback locale 'en'. Either provide a properties file for 'en' or configure another available fallbackLocale
info project4 › Running task generateFlexChangesBundle...
info project4 › Running task generateComponentPreload...
info ProjectBuilder Build succeeded in 609 ms
```

[...]

PR #3302, opened 2025-06-05
fix: align @ui5/cli version

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/3220


packages/cf-deploy-config-writer/src/constants.ts
```diff
@@ -30,7 +30,7 @@ export const MbtPackageVersion = '^1.2.29';
 export const UI5TaskZipperPackage = 'ui5-task-zipper';
 export const UI5TaskZipperPackageVersion = '^3.4.x';
 export const UI5Package = '@ui5/cli';
-export const UI5PackageVersion = '^3.9.2';
+export const UI5PackageVersion = '^4.0.16';
```
devinea, line 33: Out of curiosity, any reason for `^4.0.16` and `^4.0.0` in application-writer?

devinea (dismissed):
Changes default version changes lg (minor comment) 
Snapshot updates as expected. 
changset ✅

packages/cf-deploy-config-writer/src/constants.ts
longieirl: I think I was just following the same existing syntax but I've now applied the base as 4.0.16, soo its consistent across all packages. Thanks.

longieirl:
@nikmace can you please review? I've added a change to ADP to align the version.

nikmace (approved):
Changes look good from `adp-tooling` side
Thanks

PR #3305, opened 2025-06-05
fix(deploy-gen): updates for abap headless deploy config

cianmSAP:
- Update to the ABAP sub generator to follow CF's pattern - moves the private init and writing methods into the yeoman writing phase when launched as a sub gen. This eliminates the requirement to have `conflicter.force` set to true. Also speeds up writing as it occurs in the correct order
- Also updates `appRootPath` so it contains the name of the project and not just the target folder


packages/abap-deploy-config-sub-generator/src/app/index.ts
```diff
@@ -86,10 +85,6 @@ export default class extends DeploymentGenerator {
             watchTelemetrySettingStore: false
         });
 
-        if ((this.env as unknown as YeomanEnvironment).conflicter) {
-            (this.env as unknown as YeomanEnvironment).conflicter.force = this.options.force ?? true;
```
cianmSAP, line 90: No longer required as writing phase is in the correct place and yeoman does not prompt for overwriting files

PR #3311, opened 2025-06-06
fix(preview-middleware): wrong property path

voicis:
Fixed an issue where incomplete property path was used for configuration changes created from property panel e.g "Show Counts" for table.

### DIFF
--- a/packages/preview-middleware-client/src/utils/fe-v4.ts
+++ b/packages/preview-middleware-client/src/utils/fe-v4.ts
@@ -103,13 +103,24 @@ export async function createManifestPropertyChange(
     if (!overlay) {
         return undefined;
     }
-    const overlayData = overlay?.getDesignTimeMetadata().getData();
+    const overlayData = overlay.getDesignTimeMetadata().getData();
+    const settings = overlayData.manifestSettings(modifiedControl);
     let manifestPropertyPath = overlayData.manifestPropertyPath(modifiedControl);
     if (propertyPathExtraSegments) {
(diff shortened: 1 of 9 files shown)


packages/preview-middleware-client/src/utils/fe-v4.ts
```diff
@@ -103,13 +103,24 @@ export async function createManifestPropertyChange(
...
+        if (setting) {
+            adjustedChanges[setting.path ?? setting.id] = value;
+        } else {
+            adjustedChanges[key] = value;
+        }
+        
+    }
     const [manifestPropertyChange] = overlayData.manifestPropertyChange(
-        propertyChanges,
+        adjustedChanges,
```
Jimmy-Joseph19, line 123: will this be downported and is the patch already avaialble for the version we supported. Will it cause any issue if we merge?

packages/preview-middleware-client/test/unit/cpe/__snapshots__/control-data.test.ts.snap
```diff
@@ -162,6 +162,22 @@ Object {
       "ui5Type": undefined,
       "value": "USA table",
     },
+    Object {
+      "documentation": Object {
```
Jimmy-Joseph19, line 166: What caused this updated

packages/preview-middleware-client/src/utils/fe-v4.ts
voicis: If there is no `path` then we'll just use the `id`, it shouldn't cause any issues.

PR #3317, opened 2025-06-08
fix(preview-middleware): missing variant changes

voicis:
After restricting what files are processed as changes in #3301 variant changes were excluded and were no longer shown in Adaptation Editor.


nikmace (approved):
Changes look good
Changeset present
Did not test manually

vadson71 (approved):
Straightforward change

PR #3003, opened 2025-03-06
Centralize and increase TypeScript version and introduce erasableSyntaxOnly

heimwege:
Update TypeScript version to `5.8.2`, delete local typescript devDependencies in packages and introduce new compiler option `erasableSyntaxOnly` for preview-middleware

As SonarCloud does not yet support the new property this PR introduces a sonar specific tsconfig for preview-middleware w/o this property.

PR #3318, opened 2025-06-08
Chore remove local typescript package. 

devinea:
should be loaded from the root
no changeset needed as they are devDeps.

aligning with changes made in https://github.com/SAP/open-ux-tools/pull/3003


longieirl (approved):
Changes look good.
No changeset required.

PR #3320, opened 2025-06-09
Enhancements to `getUi5Themes` Logic

kjose90:
- Improved handling of the `ui5Version` parameter.
- When `ui5Version` is not explicitly provided or set to `"Latest"`, the function now defaults to the latest supported UI5 version instead of returning all available themes.
- This ensures `sap_belize` theme does not appear when no ui5 version is provided ...


packages/ui5-info/src/ui5-version-fallback.ts
```diff
@@ -10,6 +10,10 @@ export const supportState = {
 
 // Updated Apr-25-2025 from https://ui5.sap.com/versionoverview.json
```
IainSAP: Please update the comment, and verify that the reset of the entries are valid.

kjose90: [updated](https://github.com/SAP/open-ux-tools/pull/3320/files#diff-be0ab748ee0e5ef03c53ef0fc8e6f46c8b84b68113d3b3cbdeb320820cc4b735R11)

packages/ui5-info/src/ui5-theme-info.ts
```diff
@@ -85,8 +86,13 @@ function isSupported(theme: UI5Theme, cleanSemVer: SemVer): boolean {
  * @returns UI5 themes array
  */
 export function getUi5Themes(ui5Version: string = defaultVersion): UI5Theme[] {
-    const ui5VersionSince = ui5Version.replace('snapshot-', '');
-    const cleanSemVer = coerce(ui5VersionSince);
+    // Handle 'Latest' versions by using the latest supported UI5 version
+    const isUsingDefaultVersion = ui5Version === defaultVersion;
```
IainSAP:
```suggestion
    const isLatestVersion = ui5Version === latestVersionString;
```

The assumption that `defaultVersion` === `latestVersionString` but that may not always be the case

Also the tsdoc needs to be updated. The behaviour is changing to returning only themes supported by the latest available version when `ui5Version` is not specified.
```
* @param [ui5Version] - optional, restrict the returned themes to only those supported by specified UI5 version. If not specified, only themes supported by the latest UI5 version fallback version will be returned.
 * If the passed version is not a valid semantic version all themes will be returned.
 ```

File: packages/ui5-info/src/ui5-theme-info.ts [hunk 3]
```diff
@@ -85,8 +86,13 @@ function isSupported(theme: UI5Theme, cleanSemVer: SemVer): boolean {
...
+    const resolvedUi5Version = isUsingDefaultVersion
+        ? supportedUi5VersionFallbacks[0].version
```
Klaus-Keller: Do we now use the fallback also for cases where we would have access to the real data at https://ui5.sap.com/versionoverview.json? My understanding was, to use fallbacks in cases we are not able to read the live data, only.

packages/ui5-info/src/ui5-theme-info.ts
kjose90: Added check for [latestVersionString](https://github.com/SAP/open-ux-tools/pull/3320/files#diff-0c6f81e6fb16854470c89216e1e8eba18132c04f44db94df737895fc80c197d9R94)

packages/ui5-info/src/ui5-theme-info.ts
IainSAP: `getLatestUI5Version` does not use the cache and we want to avoid additional unnecessary network calls so I propose we update to support cache use and call `getLatestUI5Version(useCache = true)`. Here is my proposal: 

https://github.com/SAP/open-ux-tools/commit/091aece046fe44b390edcbeba5998af7763900e2#r159653844

Klaus-Keller: > `getLatestUI5Version` does not use the cache and we want to avoid additional unnecessary network calls so I propose we update to support cache use and call `getLatestUI5Version(useCache = true)`. Here is my proposal:
> 
> [091aece#r159653844](https://github.com/SAP/open-ux-tools/commit/091aece046fe44b390edcbeba5998af7763900e2#r159653844)

Sorry, I'm not fully getting this. Cache is not the same as fallback, right? 
With the new implementation it looks like we use `supportedUi5VersionFallbacks` whenever `isUsingDefaultVersion` is true. Perhaps we can reuse the cache in this case as well?

IainSAP: Yes I propose we dont use fallbacks we call `getLatestUI5Version` but update it to also consider the cache. The point is if the cache is loaded theres no point making another call to get the latest version if the cache is set.

BTW we will create a TBI to update `getLatestUI5Version` to use the proper API `getUI5Versions` as this internally handles cache and fallbacks and in the end makes the same network call.

packages/ui5-info/src/ui5-version-info.ts
```diff
@@ -137,55 +137,55 @@ async function parseUI5VersionsAndSupport(
...
- * @returns Array of UI5 versions or UI5VersionSupport objects
+ * @param {'official' | 'snapshot' | 'support'} type - The type of UI5 versions to retrieve.
+ * @param {boolean} [useCache] - If true, attempts to return cached versions first.
+ * If false, or if cached versions are not available/ empty, it fetches data from https://ui5.sap.com.
+ * If fresh data is fetched and useCache is true, the cache will be updated.
+ * @param {string} [snapshotUrl] - The URL to fetch snapshot versions from. Required if type is 'snapshot'.
+ * @returns {Promise<string[] | UI5VersionSupport[]>} A promise that resolves to an array of UI5 version strings
+ * Returns an empty array if type is snapshot and snapshotUrl is not provided.
  */
 const retrieveUI5VersionsCache = async (
```
IainSAP: Theres no need to touch this function, it works perfectly as is. Its already utilising the cache.

packages/ui5-application-inquirer/src/prompts/index.ts
```diff
@@ -71,7 +71,7 @@ export function getQuestions(
         [promptNames.addFlpConfig]: getAddFlpConfigPrompt(promptOptions?.addFlpConfig),
         [promptNames.enableVirtualEndpoints]: getEnableVirtualEndpoints(capCdsInfo),
         [promptNames.showAdvanced]: getShowAdvancedPrompt(),
-        [promptNames.ui5Theme]: getUI5ThemePrompt(),
+        [promptNames.ui5Theme]: await getUI5ThemePrompt(),
```
IainSAP: This shouldnt be awaited

packages/ui5-application-inquirer/src/prompts/prompts2.ts
```diff
@@ -29,7 +29,7 @@ export function getShowAdvancedPrompt(): UI5ApplicationQuestion {
  *
  * @returns The `ui5Theme` prompt
  */
-export function getUI5ThemePrompt(): UI5ApplicationQuestion {
+export async function getUI5ThemePrompt(): Promise<UI5ApplicationQuestion> {
```
IainSAP: Please remove this async

packages/ui5-application-inquirer/src/index.ts
```diff
@@ -35,7 +35,7 @@ async function getPrompts(
     };
     const ui5Versions = await getUI5Versions(filterOptions);
 
-    return getQuestions(ui5Versions, promptOptions, capCdsInfo, isYUI);
+    return await getQuestions(ui5Versions, promptOptions, capCdsInfo, isYUI);
```
IainSAP: Please remove await

packages/ui5-application-inquirer/src/prompts/index.ts
```diff
@@ -35,12 +35,12 @@ import type { CdsUi5PluginInfo } from '@sap-ux/project-access';
  * @param [isYUI] - optional, default is `false`. Changes the behaviour of some validation since YUI does not re-validate prompts that may be inter-dependant.
  * @returns the prompts
  */
-export function getQuestions(
+export async function getQuestions(
```
IainSAP: await can be removed

packages/ui5-info/src/ui5-theme-info.ts
```diff
@@ -78,15 +79,22 @@ function isSupported(theme: UI5Theme, cleanSemVer: SemVer): boolean {
 }
 
 /**
- * Return supported UI5 themes.
+ * Returns the list of supported UI5 themes for a given UI5 version.
  *
- * @param [ui5Version] - optional, restrict the returned themes to only those supported by specified UI5 version
- * If the passed version is not a valid semantic version all themes will be returned.
- * @returns UI5 themes array
+ * - If ui5Version is Latest, the latest supported UI5 version is used.
```
IainSAP:
```suggestion
 * - If ui5Version is `Latest`, the latest supported UI5 version is used.
```

packages/ui5-info/src/ui5-version-info.ts
```diff
@@ -378,12 +376,29 @@ export async function getUI5Versions(filterOptions?: UI5VersionFilterOptions): P
 }
 
 /**
- * Method retrieves latest SAPUI5 version by sending HTTP GET request to "https://ui5.sap.com/version.json'".
+ * Retrieves the latest supported UI5 version.
```
IainSAP:
```suggestion
 * Retrieves the latest supported published UI5 version.
```

packages/ui5-info/test/ui5-theme-info.test.ts
```diff
@@ -1,18 +1,17 @@
...
     const allExpectedThemes: UI5Theme[] = Object.values(ui5Themes);
-    test('getUi5Themes', () => {
-        expect(getUi5Themes()).toEqual(allExpectedThemes);
-        expect(getUi5Themes('not-a-valid-version')).toEqual(allExpectedThemes);
-        expect(getUi5Themes(defaultMinUi5Version)).toEqual(allExpectedThemes.slice(0, 2));
-        expect(getUi5Themes('1.71')).toEqual(allExpectedThemes.slice(0, 2));
-        expect(getUi5Themes('1.72')).toEqual(allExpectedThemes.slice(0, 3));
-        expect(getUi5Themes('1.101')).toEqual(allExpectedThemes.slice(0, 3));
-        expect(getUi5Themes('1.102')).toEqual(allExpectedThemes);
+    test('getUi5Themes', async () => {
```
IainSAP, line 63: Since `getUi5Themes` now makes a network call we will have to nock the requests. See other tests in the is module for examples. Thx.

IainSAP (commented):
Thanks @kjose90 

- Additional caching support for `getUI5Themes` is welcome
- Tests to cover

kjose90:
As discussed - returning `await getUI5Versions` from [getLatestUi5Themes](https://github.com/SAP/open-ux-tools/pull/3320/files#diff-a61e1f718d63875fc8d476a327ee58a3f40d130b9d136881a168b2cc901a791aR390)

packages/ui5-info/src/ui5-theme-info.ts
```diff
@@ -78,15 +79,22 @@ function isSupported(theme: UI5Theme, cleanSemVer: SemVer): boolean {
...
- * @param [ui5Version] - optional, restrict the returned themes to only those supported by specified UI5 version
- * If the passed version is not a valid semantic version all themes will be returned.
- * @returns UI5 themes array
+ * - If ui5Version is `Latest`, the latest supported UI5 version is used.
+ * - If ui5Version is a snapshot version (e.g., 'snapshot-1.137.0'), the snapshot prefix is removed.
+ * - If ui5Version is invalid or cannot be parsed into a semantic version, all themes are returned.
+ * - Themes that are deprecated for the given version are marked with "(deprecated)" in their label.
+ *
+ * @param [ui5Version] - The UI5 version to filter themes by. Defaults to `defaultVersion`.
+ * @returns An array of UI5 themes supported for the specified version.
```
longieirl: Comment needs to be updated to reflect await/async.

packages/ui5-info/test/ui5-version-info.test.ts
```diff
@@ -360,6 +361,32 @@ describe('getUI5Versions: npm listed versions', () => {
...
+            ...originalConstants,
+            ui5VersionsCache: {
+                officialVersions: [],
+                snapshotsVersions: [],
+                support: []
+            }
+        };
+    });
+
+    beforeEach(() => {
```
IainSAP, line 386: Might be worth resetting the cache before each test (rather than once before all tests) so the tests are isolated?

kjose90: [added](http://github.com/SAP/open-ux-tools/pull/3320/files#diff-b67be0a9209d6236009cdbf6d58c8d85292564bdaa630d1ed6add7be985e25e7R388)

Klaus-Keller (approved):
Thanks @kjose90,

* improvements for `getUi5Themes()` look good
* switch to `async` is a breaking change which would usually require a major version update, but as we have a 0.x.y version it is fine
* focused review on module `@sap-ux/ui5-info`
* test coverage is great

Approved from my side

PR #3321, opened 2025-06-09
fix(deploy-gens): re-add conflicter.force prop for yui standalone

cianmSAP:
Yeoman environment property `conflicter.force` is still required when running deploy config standalone

### DIFF
--- a/packages/abap-deploy-config-sub-generator/src/app/index.ts
+++ b/packages/abap-deploy-config-sub-generator/src/app/index.ts
@@ -10,7 +10,8 @@ import {
     YUI_EXTENSION_ID,
     YUI_MIN_VER_FILES_GENERATED_MSG,
     sendTelemetry,
-    TelemetryHelper
+    TelemetryHelper,
+    YeomanEnvironment
 } from '@sap-ux/fiori-generator-shared';
 import { getPackageAnswer, getTransportAnswer, reconcileAnswers } from '@sap-ux/abap-deploy-config-inquirer';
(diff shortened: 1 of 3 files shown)


packages/abap-deploy-config-sub-generator/src/app/index.ts
```diff
@@ -85,6 +86,9 @@ export default class extends DeploymentGenerator {
             watchTelemetrySettingStore: false
         });
 
+        if ((this.env as unknown as YeomanEnvironment).conflicter) {
```
longieirl, line 91: Do you want to add a dev note?

packages/cf-deploy-config-sub-generator/src/app/index.ts
```diff
@@ -105,6 +106,10 @@ export default class extends DeploymentGenerator {
 
         DeploymentGenerator.logger?.debug(t('cfGen.debug.initTelemetry'));
 
+        if ((this.env as unknown as YeomanEnvironment).conflicter) {
```
longieirl, line 111: Same here, a dev note?

packages/abap-deploy-config-sub-generator/src/app/index.ts
cianmSAP: Thanks, added

IainSAP (approved):
Thanks @cianmSAP

packages/cf-deploy-config-sub-generator/src/app/index.ts
cianmSAP: Added

PR #3326, opened 2025-06-09
Mockserver Config Writer Texts

lfindlaysap:
Reviewed `packages/mockserver-config-writer/src/translations/mockserver-config-writer.i18n.json`.

### DIFF
--- a/packages/mockserver-config-writer/src/translations/mockserver-config-writer.i18n.json
+++ b/packages/mockserver-config-writer/src/translations/mockserver-config-writer.i18n.json
@@ -1,6 +1,6 @@
 {
     "questions": {
         "pathToMock": "Path to mocked service",
-        "overwrite": "Overwrite Services"
+        "overwrite": "Overwrite services"
     }
 }
--- a/packages/mockserver-config-writer/src/translations/mockserver-config-writer.i18n.json
+++ b/packages/mockserver-config-writer/src/translations/mockserver-config-writer.i18n.json
@@ -1,6 +1,6 @@
 {
     "questions": {
-        "pathToMock": "Path to mocked service",
-        "overwrite": "Overwrite services"
+        "pathToMock": "Path to Mocked Service",
+        "overwrite": "Overwrite Services"
     }
 }
--- a/packages/mockserver-config-writer/src/translations/mockserver-config-writer.i18n.json
+++ b/packages/mockserver-config-writer/src/translations/mockserver-config-writer.i18n.json
@@ -1,6 +1,6 @@
 {
     "questions": {
-        "pathToMock": "Path to Mocked Service",
+        "pathToMock": "Path to mocked service",
         "overwrite": "Overwrite Services"
     }
 }
--- a/examples/fe-fpm-cli/CHANGELOG.md
+++ b/examples/fe-fpm-cli/CHANGELOG.md
@@ -1,5 +1,23 @@
 # @sap-ux/fe-fpm-cli
 
+## 0.1.7
+
+### Patch Changes
+
+-   @sap-ux/fe-fpm-writer@0.35.3
+
+## 0.1.6
+
+### Patch Changes
+
+-   @sap-ux/fe-fpm-writer@0.35.2
+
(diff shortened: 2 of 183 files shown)


packages/mockserver-config-writer/src/translations/mockserver-config-writer.i18n.json
```diff
@@ -1,6 +1,6 @@
 {
     "questions": {
-        "pathToMock": "Path to mocked service",
-        "overwrite": "Overwrite services"
+        "pathToMock": "Path to Mocked Service",
+        "overwrite": "Overwrite Services"
```
Klaus-Keller: These are questions to the user, not sure why we should have them all capital letters?

It is similar to these texts here:
https://github.com/SAP/open-ux-tools/blob/1a1d780c7c8ca06b0a110d9c48bb1acea3d11681/packages/ui5-library-inquirer/src/translations/ui5-library-inquirer.i18n.json#L10

packages/mockserver-config-writer/src/translations/mockserver-config-writer.i18n.json
```diff
@@ -1,6 +1,6 @@
 {
     "questions": {
-        "pathToMock": "Path to mocked service",
-        "overwrite": "Overwrite services"
+        "pathToMock": "Path to Mocked Service",
```
korotkovao:
```suggestion
        "pathToMock": "Path to mocked service",
```

packages/mockserver-config-writer/src/translations/mockserver-config-writer.i18n.json
korotkovao:
```suggestion
        "overwrite": "Overwrite services"
```

korotkovao: Thank you, I have removed the capitalisation.

JannaLisa:
Hi @Klaus-Keller,
I'm unfortunately unable to open the file properly.
But from what I see, it's really difficult for me to understand how these are questions??
I had a similar issue with Ian and Cian yesterday and we managed to resolve it.
Could you please maybe just show me/explain to me where this comes up for a user on the UI?

Klaus-Keller:
> Hi @Klaus-Keller, I'm unfortunately unable to open the file properly. But from what I see, it's really difficult for me to understand how these are questions?? I had a similar issue with Ian and Cian yesterday and we managed to resolve it. Could you please maybe just show me/explain to me where this comes up for a user on the UI?

When you open a terminal and are in the folder of a Fiori elements project, you can run following command to add a mockserver configuration:
```
npx @sap-ux/create add mockserver-config -i
```

Using `-i` which means "interactive" the following questions will be asked: 
```
✔ Path to mocked service › mainService: /sap/opu/odata/sap/SERVICE_NAME/
✔ Overwrite services … no
```

I don't think these should be capitalized.

JannaLisa:
> > Hi @Klaus-Keller, I'm unfortunately unable to open the file properly. But from what I see, it's really difficult for me to understand how these are questions?? I had a similar issue with Ian and Cian yesterday and we managed to resolve it. Could you please maybe just show me/explain to me where this comes up for a user on the UI?
> 
> When you open a terminal and are in the folder of a Fiori elements project, you can run following command to add a mockserver configuration:
> 
> ```
> npx @sap-ux/create add mockserver-config -i
> ```
> 
> Using `-i` which means "interactive" the following questions will be asked:
> 
> ```
> ✔ Path to mocked service › mainService: /sap/opu/odata/sap/SERVICE_NAME/
> ✔ Overwrite services … no
> ```
> 
> I don't think these should be capitalized.

thanks for the explanation. Klaus also checked other similar examples, which are all very brief and don't use capitalization.
So we agreed to write the texts without capitalization.
"Path to mocked service" and "Overwrite services" would be correct.
@korotkovao: would you please update accordingly? Many thanks!

korotkovao:
Thank you for clarifying, removed the capitalisation.

PR #3342, opened 2025-06-10
fix(preview-middleware): enhancedhomepage loading issues from UI5 v1.137

ui5red:
The latest ushell changes break enhanced homepage loading when using UI5 version 1.137 and above. This fix ensures custom CDN-related ushell configuration is provided before the container is required, following ushell team recommendations.

- Addresses breaking changes introduced in recent ushell updates
- Maintains compatibility with UI5 1.137+ versions
- Implements ushell-recommended configuration order

### DIFF
--- a/packages/preview-middleware-client/src/flp/init.ts
+++ b/packages/preview-middleware-client/src/flp/init.ts
@@ -305,9 +305,12 @@ export async function init({
     enhancedHomePage?: boolean | null;
     enableCardGenerator?: boolean
 }): Promise<void> {
+    // Set CDM configuration before importing ushell container
+    // to ensure proper configuration pickup during bootstrap
     if (enhancedHomePage) {
-        await initCdm();
+        initCdm();
     }
+
     const urlParams = new URLSearchParams(window.location.search);
     const container = sap?.ushell?.Container ??
         (await import('sap/ushell/Container')).default as unknown as typeof sap.ushell.Container;
--- a/packages/preview-middleware-client/src/flp/init.ts
+++ b/packages/preview-middleware-client/src/flp/init.ts
@@ -305,6 +305,9 @@ export async function init({
     enhancedHomePage?: boolean | null;
     enableCardGenerator?: boolean
 }): Promise<void> {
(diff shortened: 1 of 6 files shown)


packages/preview-middleware-client/src/flp/init.ts
```diff
@@ -305,6 +305,9 @@ export async function init({
     enhancedHomePage?: boolean | null;
     enableCardGenerator?: boolean
 }): Promise<void> {
+    if (enhancedHomePage) {
+        await initCdm();
```
heimwege: Why having this in a separate `if` statement? Can't we do this in the one we already have below? If this is needed here please add a comment so that nobody will refactor this later on because it is not obvious why this needs to be done here.

packages/preview-middleware-client/src/flp/init.ts
```diff
@@ -389,7 +392,7 @@ export async function init({
     registerSAPFonts();
 
     if (enhancedHomePage) {
-        await initCdm(container);
+        await container.init('cdm');
```
heimwege, line 398: Can we do it like this?

```suggestion
        await initCdm();
        await container.init('cdm');
```

heimwege (commented):
see comment

packages/preview-middleware-client/src/flp/init.ts
ui5red: ushell has made some changes in their bootstrapping logic from v1.137.0 onwards where they now pick up the custom ushell configuration (`window['sap-ushell-config]`) right when the Container module is imported.

Since we were previously providing our custom configuration _(containing the cdm configuration)_ after the Container was imported, it was not considered anymore during bootstrapping.

So yes, it should be a separate `if` block. I've added a comment as well for the same.

packages/preview-middleware-client/src/flp/init.ts
ui5red: unfortunately, nope 🥲

heimwege (approved):
Thanks for adding tests. Re-approve

PR #3344, opened 2025-06-10
fix(axios-ext): disable proxy for bas

longieirl:
Fixes : https://github.com/SAP/open-ux-tools/issues/3345

- setting `providerConfig.proxy = false;` on BAS is failing the API request fails, this flow was never executed on BAS before, soo it should be disabled.


IainSAP (approved):
thanks @longieirl 

- Use Axios proxy on BAS to fix broken CLI requests (non-resolved `*.dest` usage)

packages/axios-extension/src/factory.ts
```diff
@@ -73,7 +74,7 @@ function createInstance<T extends ServiceProvider>(
         rejectUnauthorized: !providerConfig.ignoreCertErrors
     };
     const localProxy = getProxyForUrl(config.baseURL);
```
IainSAP, line 76: Should we add a comment here explaining why we are using this "localProxy"  and that using it breaks the BAS proxy host resolution? Its not clear from the code....

packages/axios-extension/test/factory.test.ts
```diff
@@ -73,6 +77,7 @@ test('create', async () => {
 });
 
 test('create with proxy', async () => {
```
IainSAP, line 79: Can we update this test name to reflect that its a BAS specific test ?

IainSAP: Code is cross platform now

packages/axios-extension/src/factory.ts
```diff
@@ -80,7 +80,7 @@ function createInstance<T extends ServiceProvider>(
             agentOptions as HttpsProxyAgentOptions<string>
         );
         providerConfig.httpAgent = new HttpProxyAgent(localProxy);
-        providerConfig.proxy = false;
+        delete providerConfig.proxy;
```
IainSAP: Could we add a note?

packages/axios-extension/src/factory.ts
longieirl: BAS is no longer in play, remove the `proxy` property from the provider works.

packages/axios-extension/test/factory.test.ts
longieirl: Tests updated.

kjose90 (approved):
Changes looks good ✅ 
Thanks

PR #3353, opened 2025-06-12
fix: abap-deploy-config-inquirer - Use more specific GA link texts for cert errors.

IainSAP:
https://github.com/SAP/open-ux-tools/issues/3264

- Replace the generic cert error types with the actual error so the specific texts can be generated
- Updates tests
- Update `inquirer-common` `ErrorHandler` log prefix to be configurable
- Updates all instances of `new ErrorHandler` where a logger is not provided, to use the consumer module name as prefix

### DIFF
--- a/packages/abap-deploy-config-inquirer/src/prompts/validators.ts
+++ b/packages/abap-deploy-config-inquirer/src/prompts/validators.ts
@@ -1,44 +1,44 @@
-import { PromptState } from './prompt-state';
-import { type Destinations, isS4HC, isAbapEnvironmentOnBtp } from '@sap-ux/btp-utils';
-import {
-    createTransportNumber,
-    getTransportList,
-    isEmptyString,
-    isValidClient,
-    isValidUrl,
-    isAppNameValid
-} from '../validator-utils';
-import { DEFAULT_PACKAGE_ABAP } from '../constants';
-import { getTransportListFromService, getSystemInfo, isAbapCloud } from '../service-provider-utils';
-import { t } from '../i18n';
-import {
-    findBackendSystemByUrl,
-    initTransportConfig,
-    getPackageAnswer,
-    queryPackages,
-    getSystemConfig
-} from '../utils';
-import { handleTransportConfigError } from '../error-handler';
-import { AuthenticationType } from '@sap-ux/store';
+import type { IValidationLink } from '@sap-devx/yeoman-ui-types';
+import { AdaptationProjectType } from '@sap-ux/axios-extension';
+import { isAbapEnvironmentOnBtp, isS4HC, type Destinations } from '@sap-ux/btp-utils';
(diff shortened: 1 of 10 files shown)


packages/abap-deploy-config-inquirer/src/prompts/validators.ts
```diff
@@ -1,44 +1,44 @@
...
-import { type Destinations, isS4HC, isAbapEnvironmentOnBtp } from '@sap-ux/btp-utils';
-import {
-    createTransportNumber,
-    getTransportList,
-    isEmptyString,
-    isValidClient,
-    isValidUrl,
-    isAppNameValid
-} from '../validator-utils';
+import type { IValidationLink } from '@sap-devx/yeoman-ui-types';
```
IainSAP, line 1: Ran a format/reorder on the imports hence the diffs

packages/deploy-tooling/src/base/deploy.ts
```diff
@@ -44,7 +44,7 @@ async function handleError(
     archive: Buffer
 ): Promise<void> {
     if (ErrorHandler.isCertError(error)) {
-        const gaLink = ErrorHandler.getHelpForError(ERROR_TYPE.CERT)?.toString();
+        const gaLink = new ErrorHandler().getValidationErrorHelp(error)?.toString();
```
longieirl: Why are we instantiating only in some places? How do you know when use either?
Also, just noticed the logger for the ErrorHandler is `@sap-ux/odata-service-inquirer` when it should be `@sap-ux/inquirer-common`, doesn't need to be changed here.

IainSAP: Thats how the error handler has been implemented. Some functions internally access the state and so the a new instance is required. `getValidationErrorHelp`  references error state in some cases (not here) and does analysis of the error and so gives more accurate messages. But this probably could be updated for convenience (a static version). `How do you know when use either?` - you can only use some of the functions statically, and all functions are well doc'd.

WIll update the log prefix in the ErrorHandler. Thanks.

longieirl (dismissed):
Changeset is good.
Changes look good.
Tests are appropiate.

packages/abap-deploy-config-inquirer/src/prompts/validators.ts
```diff
@@ -596,7 +600,11 @@ export async function validateTransportChoiceInput({
             );
         } catch (error) {
             if (ErrorHandler.isCertError(error)) {
-                return ErrorHandler.getHelpForError(ERROR_TYPE.CERT) ?? true;
+                return (
+                    new ErrorHandler(undefined, undefined, '@sap-ux/abap-deplu-config-inquirer').getValidationErrorHelp(
```
longieirl: Spelling?

Issue #3360, opened 2025-06-16
BUG - preview middleware config does not point to flpSandbox.html file in non-virtual scenario


cianmSAP:
### Description
In a non-virtual endpoint scenario for preview, the current configuration uses the default settings. The default settings point to a virtual endpoint is served at flp.html and intent app-preview

However our package.json still (correctly) points to the physical flpSandbox.html in webapp/test so the virtual endpoint that is being served, is not used.

### Steps to Reproduce
1. Generate a fiori app and select no for using virtual endpoints
2. See `fiori-tools-preview` middleware config in ui5-yaml

### Expected results
Config contains the path & intent e.g.           
```
          path: test/flpSandbox.html 
          theme: sap_fiori_3 
          intent:
            object: project1 
            action: tile
```
### Actual results
The preview config does not contain `path` or `intent`

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3361, opened 2025-06-16
fix(fiori-gen): correct preview config for non virtual scenario

cianmSAP:
#3360

Updates preview config in non virtual endpoint scenario

- Move `getFlpId` and `getSemanticObject` to fiori gen shared - for use by the writers
- Passes `flpAction` to the writers, and let them create the `flpAppId` (the writers still accepts the `flpAppid` for backward compatibility). This is because the values of the app id and action are needed separately for yaml configs, and avoids passing duplicate data
- Adds path and intent to flp config for non virtual scenario


packages/fiori-app-sub-generator/src/fiori-app-generator/transforms.ts
```diff
@@ -316,10 +315,7 @@ function getBaseAppConfig(
             id: appId,
             title: project.title,
             description: project.description,
-            flpAppId: getFlpId(
-                appId,
-                floorplan === FloorplanFF.FF_SIMPLE ? defaultNavActionDisplay : defaultNavActionTile
-            ),
+            flpAction: floorplan === FloorplanFF.FF_SIMPLE ? defaultNavActionDisplay : defaultNavActionTile,
```
IainSAP: Can remove this property from the transform as the difference is writer specific and the defaults can be assigned in the writer along with the flp id generation?

PR #3363, opened 2025-06-16
Fix/preview middleware/lrep connector

heimwege:
Adjust the min UI5 version for the usage of the workspace connector (`ObjectStorageConnector`) to `1.76`. The `ObjectStorageConnector` does not exist in UI5 `1.71` ([link](https://github.com/SAP/openui5/tree/rel-1.71/src/sap.ui.fl/src/sap/ui/fl/apply/_internal/connectors)) but it already exists in `1.76` ([link](https://github.com/SAP/openui5/tree/rel-1.76/src/sap.ui.fl/src/sap/ui/fl/apply/_internal/connectors)).

Doing so fixes a bug where running an app with UI5 sources from npmjs using the lowest available version (`1.76.0`) will not work.

Additional fixes:
- Instead of blindly deleting certain indexes from an array we should actually look if it is the correct entry we want to delete.


heimwege:
UPDATE:
ok now I got it. it's for sure a false positive. It's not even related to my code. Sonar complains about a duplication because of a card generator part that is identical compared to the cards-editor-middleware. That's actually true, but cards-editor-middleware is deprecated and this is the successor. Any besides that the duplication was in cards-editor-middleware. preview middleware owns the original code. So everything's fine 👍🏻 

@Klaus-Keller @devinea I don't geht where these duplicated lines should come from. It's just a type guard. Is this a false positive? Did I miss something or can we overrule this? 🤔 
Everything was fine and then I moved the type guard from `flp.ts` to `config.ts` and the duplication issue appeared.

Klaus-Keller:
Should we add a test for both cases depending on the version or is there already one, but not with the new threshold of versions?

heimwege:
> Should we add a test for both cases depending on the version or is there already one, but not with the new threshold of versions?

There are tests for [1.84.50](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/test/unit/flp/init.test.ts#L254) and [1.71.60](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/test/unit/flp/init.test.ts#L278). Should we add another one for `1.76`? But that would be a copy of the `1.84.50` one

Klaus-Keller:
> > Should we add a test for both cases depending on the version or is there already one, but not with the new threshold of versions?
> 
> There are tests for [1.84.50](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/test/unit/flp/init.test.ts#L254) and [1.71.60](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/test/unit/flp/init.test.ts#L278). Should we add another one for `1.76`? But that would be a copy of the `1.84.50` one

Perhaps change `1.84.50` to `1.76`? Not too important, just thought as we now have certainty about the exact threshold we could check for it.

heimwege:
> > > Should we add a test for both cases depending on the version or is there already one, but not with the new threshold of versions?
> > 
> > 
> > There are tests for [1.84.50](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/test/unit/flp/init.test.ts#L254) and [1.71.60](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/test/unit/flp/init.test.ts#L278). Should we add another one for `1.76`? But that would be a copy of the `1.84.50` one
> 
> Perhaps change `1.84.50` to `1.76`? Not too important, just thought as we now have certainty about the exact threshold we could check for it.

fixed in ecaadf7e018426e28c7cad8d87441fa3a166c813

devinea:
> @Klaus-Keller @devinea I don't geht where these duplicated lines should come from. It's just a type guard. Is this a false positive? Did I miss something or can we overrule this? 🤔
> Everything was fine and then I moved the type guard from flp.ts to config.ts and the duplication issue appeared.

@heimwege It is picking up `getUI5Libs` is duplicated in 
https://github.com/SAP/open-ux-tools/blob/7921990b559690180c7ca9707f57a3093b91ae4c/packages/cards-editor-middleware/src/index.ts#L62-L83
and 
https://github.com/SAP/open-ux-tools/blob/7921990b559690180c7ca9707f57a3093b91ae4c/packages/cards-editor-middleware/src/index.ts#L62-L83

PR #3364, opened 2025-06-17
fix: flattened annotation deletion

voicis:
- Updated `@sap/ux-cds-compiler-facade` dependency
- Updated test debug config to only use one worker for better performance
- Fixed issue where flattened structures were kept when their "parent" annotation was deleted


packages/fiori-annotation-api/test/unit/fiori-service.test.ts
```diff
@@ -2498,6 +2498,211 @@ rating : Rating;
...
+                            }
+                        ],
+                        'IncidentService',
+                        fsEditor,
+                        false
+                    );
+
+                    expect(text).toMatchSnapshot();
+                });
+                test.skip('with qualifiers and embedded annotation', async () => {
```
vadson71: Is skipped test needed?

voicis: While adding tests for this issue I found a different issue. Technically it is not needed for this PR, but I thought it would be a good starting point when working on the fix for the other issue, but you are probably right it's better to keep it separate and I can still add a link to the test in this PR.

PR #3366, opened 2025-06-17
fix(app-config-writer): preview middleware dependency

heimwege:
Currently there's a devDependency to preview-middleware in app-config-writer but actually it is needed at runtime. This PR fixes the dependency.

### DIFF
--- a/.changeset/hot-cows-provide.md
+++ b/.changeset/hot-cows-provide.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/app-config-writer': patch
+---
+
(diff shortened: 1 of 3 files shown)


cianmSAP (approved):
Update looks good, should resolve issue with consuming app-config-writer externally
Thanks @heimwege

Klaus-Keller (approved):
Thanks @heimwege!
 
* removing dependency to `@sap-ux/preview-middleware` by adding local `sanitizeRtaConfig` looks good
* changeset exists
* test coverage is great

Approved from my side

PR #3370, opened 2025-06-18
tbi(fiori-gen): updates followUpCommadn to add additional params for post generation tasks

cianmSAP:
Internal TBI #34739

- Updates followUpCommand to be an object containing `cmdName` and `cmdParams` which can be used in the post generation task


packages/fiori-app-sub-generator/src/fiori-app-generator/end.ts
```diff
@@ -44,21 +44,23 @@ async function saveApiHubApiKey(apiKey: string, logger: Logger & ILogWrapper): P
...
+ * @param vscode - vscode instance
+ * @param followUpCommand - post generation command which will be ran instead of default
+ * @param addProjectToWorkspace - adds project to the workspace automatically
  */
 async function runPostGenHooks(
     projectPath: string,
     logger: ILogWrapper,
     vscode?: VSCodeInstance,
-    followUpCommand?: string
+    followUpCommand?: string,
```
IainSAP: Why not use a command "addProjectToWorkspace" do we really need another parameter? The follup command was added for this reason I would think.

packages/fiori-app-sub-generator/src/fiori-app-generator/fioriAppGeneratorOptions.ts
```diff
@@ -100,4 +100,10 @@ export interface FioriAppGeneratorOptions extends Generator.GeneratorOptions, Fi
      * VSCode Workspace folders
      */
     workspaceFolders?: string[];
+    /**
+     * Used for the post generation task, if set to true, the project will automatically be added to the workspace.
+     *
+     * @default false
+     */
+    addProjectToWorkspace?: boolean;
```
IainSAP: Why not have an option `followupCommand`, this would be re-usable in other cases.

cianmSAP: Enhanced `followupCommand` as discussed

packages/fiori-app-sub-generator/src/fiori-app-generator/end.ts
cianmSAP: Updated to use followup command (with params)

kjose90 (approved):
LGTM ✅ Thanks!

PR #3372, opened 2025-06-19
fix(fiori-gen): adds preview settings to external service state object

cianmSAP:
Internal bug #31013

- Updates the external `Service` state object to allow configuration of the `previewSettings` i.e `scp` and `authenticationType`


IainSAP (approved):
Thanks @cianmSAP 

- Adds support for correct preview configs when no connected system (adaptors, headless)
- Tests updated to cover

PR #3374, opened 2025-06-19
fix(ui-components): UIDialog: Override FluentUI overlay background with VSCode theme variable

815are:
Problem - overlay background for dialogs used buildin color from fluentui instead of vscode theme variable.
Appllied corrected version proposed by Goran:
```
  background: 'var(--vscode-editor-background)',
  opacity: 0.8
```

Light theme:
image

Dark theme:
image

### DIFF
--- a/.changeset/early-boxes-vanish.md
+++ b/.changeset/early-boxes-vanish.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui-components': patch
+---
+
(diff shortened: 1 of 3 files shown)


Klaus-Keller (approved):
Thanks @815are!

* change to background looks like VSCode now
* changeset exists
* test have been added
* did a manual test in Storybook, works as expected

Approved from my side

Issue #3341, opened 2025-06-10
BUG - Warn user when Windows path exceeds 256 characters


korotkovao:
### Related Feature
Feature request: _issue number_

### Description
During the project generation and deployment configuration, there is no warning for Windows users when the paths and names are too long and exceed 256 characters.

### Steps to Reproduce
Steps to reproduce the behaviour:
1. Go to Fiori: Open CF Application Router Generator
2. Enter MTA ID field value that is very long

Alternatively
1. Open the application generator 
2. Enter a name or namespace that is very long 
3. Enter the target path

### Expected results

The Windows user is warned the long name and path might result in issues during generation/deployment.

### Actual results

No warning.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [x] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

korotkovao:
Complete

PR #3375, opened 2025-06-20
fix(cf-deploy-config-inquirer): update error message and regex

korotkovao:
#3341

Update MTA ID error message and regex.


packages/cf-deploy-config-inquirer/src/prompts/validators.ts
```diff
@@ -86,7 +86,8 @@ export function validateMtaId(input: string, previousAnswers: CfAppRouterDeployC
     if (input.length > 100) {
         return t('errors.invalidMtaIdError');
```
cianmSAP: Would it not be more accurate to have a specific error for this constraint? 
`invalidMtaIdError` is quite long. You could have an error which is direct, something like `Cannot exceed 100 characters`

korotkovao: Thank you, I have pushed the update

longieirl (dismissed):
Changeset is good.
Tests are good.
Change is good, thanks for addressing feedback.

cianmSAP (dismissed):
Changes look good to me 
Comment addressed 
Changeset is appropriate

packages/cf-deploy-config-inquirer/src/translations/cf-deploy-config-inquirer.i18n.json
```diff
@@ -34,7 +34,8 @@
         "destinationNameLengthError": "Destination name cannot contain more than 200 characters",
         "folderDoesNotExistError": "Folder path does not exist: {{- filePath}}",
         "noMtaIdError": "MTA ID cannot be empty",
-        "invalidMtaIdError": "The ID can only contain letters, numbers, dashes, periods, and underscores (but no spaces).",
+        "mtaIdLengthError": "The MTA ID must be under 100 characters in length.",
```
lfindlaysap:
```suggestion
        "mtaIdLengthError": "The MTA ID must not exceed 100 characters.",
```

packages/cf-deploy-config-inquirer/src/translations/cf-deploy-config-inquirer.i18n.json
```diff
@@ -34,7 +34,8 @@
...
+        "invalidMtaIdError": "The MTA ID can only contain letters, numbers, dashes, periods, and underscores. It cannot contain spaces or begin with a number.",
```
lfindlaysap:
```suggestion
        "invalidMtaIdError": "The MTA ID must only contain letters, numbers, dashes, periods, and underscores. It cannot contain spaces or begin with a number.",
```

lfindlaysap (changes requested):
@korotkovao, two suggestions from me.

lfindlaysap (approved):

cianmSAP (approved):
Re-approving, recent commits look good

korotkovao:
Thank you All

Issue #3257, opened 2025-05-21
BUG - Undefined being written to xs-app.json for CF deployment


longieirl:
### Description
Generating a deployment configuration using a Basic template when no OData service is selected results in the following `xs-app.json` being generated with `undefined`. 

```json
    {
      "source": "^undefined/(.*)$",
      "target": "undefined/$1",
      "destination": "mydestination",
      "authenticationType": "xsuaa",
      "csrfProtection": false
    },
```

Since the project was generated with no OData service, the source and target cannot be populated, however, in some instances, developers are unaware they need to update this field if they are introducing their own custom AJAX calls. Developers are then deploying their custom code and are expecting the destination to be triggered.

Suggestion is update `undefined` to `<apply-service-segment-path>` to make it more explicit they need to update this value with to handle specific API flows.

For example, if they are calling `/myendpoint`, then the `xs-app.json` should be updated to;

```json
    {
      "source": "^/myendpoint/(.*)$",
      "target": "/myendpoint/$1",
      "destination": "mydestination",
      "authenticationType": "xsuaa",
      "csrfProtection": false
    },
```

### Steps to Reproduce
1. Go to BAS
2. Generate an app using Basic Template
3. Select none for data source
4. Select an existing destination on the CF properties page

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Validated in tool-suite.

PR #3378, opened 2025-06-23
fix: handle when project is created with no datasource

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/3257

### DIFF
--- a/packages/cf-deploy-config-writer/src/cf-writer/app-config.ts
+++ b/packages/cf-deploy-config-writer/src/cf-writer/app-config.ts
@@ -119,7 +119,8 @@ async function getUpdatedConfig(cfAppConfig: CFAppConfig, fs: Editor): Promise<C
         destinationAuthentication: cfAppConfig.destinationAuthentication || destinationAuthentication,
         isDestinationFullUrl: cfAppConfig.isDestinationFullUrl ?? destinationIsFullUrl,
         apiHubConfig: cfAppConfig.apiHubConfig,
-        firstServicePathSegment: firstServicePathSegmentUI5Config || firstServicePathSegment,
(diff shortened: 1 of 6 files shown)


packages/cf-deploy-config-writer/src/cf-writer/app-config.ts
```diff
@@ -119,7 +119,8 @@ async function getUpdatedConfig(cfAppConfig: CFAppConfig, fs: Editor): Promise<C
         destinationAuthentication: cfAppConfig.destinationAuthentication || destinationAuthentication,
         isDestinationFullUrl: cfAppConfig.isDestinationFullUrl ?? destinationIsFullUrl,
         apiHubConfig: cfAppConfig.apiHubConfig,
-        firstServicePathSegment: firstServicePathSegmentUI5Config || firstServicePathSegment,
+        firstServicePathSegment:
+            firstServicePathSegmentUI5Config ?? firstServicePathSegment ?? '<apply-service-segment-path>',
```
IainSAP, line 123: Should we reference this tag in the read me...with an example. Its a little obscure?

longieirl: This is a really edge case since, its only where a basic app is created with no datasource but they have selected a CF destination. The issue has arisen where developers have subsequently appended a dataSource or some custom AJAX API call and expected the routing to work. 
I'll append a developers note to the readme. Thanks.

cianmSAP (approved):
Updates look good and are clear eihth README.md update
Changeset looks good
Snapshot shows update

docirl (approved):
Changes look good
Sample project and test updated

PR #3379, opened 2025-06-24
fix(fiori-gen): ensure service runtimeName is upper case to match service id

cianmSAP:
`serviceId` from the chosen service is upper case and the `runtimeName` passed in the abapCSN object must be converted in order to succesfully match


longieirl (approved):
Changeset and tests are good.
Small change, lgtm.

PR #3380, opened 2025-06-24
fix(fiori-gen): correct launch config start scripts for virtual endpoints

cianmSAP:
Currently the launch config start scripts e.g `Start project1` have outdated `args` when virtual endpoints are used

This corrects the configs when virtual endpoints are used so they are : 

```
            "args": [
                "--open",
                "test/flp.html#app-preview"
            ]
```

### DIFF
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.32
+
+### Patch Changes
+
+-   @sap-ux/fiori-elements-writer@2.5.4
+-   @sap-ux/fiori-freestyle-writer@2.4.15
+
 ## 1.1.31
 
 ### Patch Changes
--- a/examples/simple-generator/package.json
+++ b/examples/simple-generator/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/generator-simple-fe",
-    "version": "1.1.31",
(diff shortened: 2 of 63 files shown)


packages/fiori-app-sub-generator/test/unit/utils/common.test.ts
```diff
@@ -268,6 +276,7 @@ describe('Test utils', () => {
         });
 
         it('should generate correct launch config for OData v2', async () => {
```
IainSAP, line 278: This test name is a little confusing. It appears to test that `createLaunchConfig ` is called rather than the config is actually created and Im not sure the odata version is relevant? Maybe we can rename?

cianmSAP: Yeah, not sure why we are calling `createLaunchConfig`, I assume the test is ensuring `createLaunchConfig` is called with the correct params based on what is passed to `generateLaunchConfig`.
So that's why I've updated this. 

I assume it was to increase the coverage for the conditions of v2 vs v4

IainSAP (approved):

packages/fiori-app-sub-generator/test/unit/utils/common.test.ts
kjose90: yea looks like it was put to increase coverage ..

kjose90 (approved):
changes lgtm.
Havent tested.
changeset & tests ✅

Issue #1156, opened 2023-08-09
TBI - ABAP deployment should prevent crendentials being entered in clear text 


longieirl:
### Description (include screenshots)
Found when testing https://github.com/SAP/open-ux-tools/issues/1150

Clear text credentials should not be allowed using either ui5-deploy.yaml or via the cli.

The credentials should be validated before being accepted, for example;

```JS
export function validateCredential(credentials: Credentials): void {
    function validateUseEnv(property: string): boolean {
        return !property || (property && property.startsWith('env:'));
    }

    if (credentials && (!validateUseEnv(credentials.username) || !validateUseEnv(credentials.password))) {
        throw new Error(i18next.t('ERROR_USER_PASSWORD_PLAIN'));
    }
}
```

Invalid CLI param;
```bash
 fiori deploy --noConfig --username XYZ_USER --password XYZ_PASSWORD --description 'My Test App' --url 'https://XYZ.sap-system.corp:44311' --client '001' --package '$tmp' --name 'MyProject'
```

Invalid YAML config;
```
...
    configuration:
      target:
        url: https://XYZ.sap-system.corp:44311
        client: 200
        auth: basic
      credentials:
        username: XYZ_USER
        password: XYZ_PASSWORD
```

### Value
Supplying and using cleartext passwords is a bad practice.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Update the ABAP deployment flow to prevent credentials being read from a yaml config or via the CLI
- [ ] Update test cases
- [ ] Provide working examples

Valid YAML config;
```
...
    configuration:
      target:
        url: https://XYZ.sap-system.corp:44311
        client: 200
        auth: basic
      credentials:
        username: env:XYZ_USER
        password: env:XYZ_PASSWORD
```

longieirl:
Related issues https://github.com/SAP/open-ux-tools/issues/1148

https://www.npmjs.com/package/@sap/ux-ui5-tooling handles username and password being loaded from the ui5-deploy.yaml or via cli. For example, if a username doesnt contain `env:` in the config or via the CLI, it's appended by `@sap/ux-ui5-tooling` since this is a known flow for CI/CD pipeline jobs that needs to be supported. 
This is important since all params passed to deploy-tooling will have `env:` appended, so username and passwords will never be in the clear.

PR #3384, opened 2025-06-25
fix: prevent plain text credentials

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/1156


longieirl:
Fixed as part of https://github.com/SAP/open-ux-tools/pull/4539

PR #3390, opened 2025-06-25
tbi(fiori-gen): update fiori gen to fetch credentials only upon system selection

cianmSAP:
Internal ticket # 32588

#### @sap-ux/store 
- Adds serizable property to `BackendSystem` : `systemType` 
- Adds new option to `getAll` - `includeSensitiveData` which will determine whether the secure store is accessed when retrieving backend systems 
- Adds temp method to update all backend systems which do not have `systemType` 
- Adds `partialUpdate` to `DataAccess` and implemented in `FilesystemStore`

#### @sap-ux/odata-service-inquirer 
- Updates Fiori gen to use new option `includeSensitiveData` set to false when retrieving the list of systems
- When a system is selected, we make an isolated call to the secure store to fetch the credentials

### DIFF
--- a/packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts
+++ b/packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts
@@ -10,7 +10,7 @@ import {
 } from '@sap-ux/btp-utils';
 import { ERROR_TYPE } from '@sap-ux/inquirer-common';
 import type { OdataVersion } from '@sap-ux/odata-service-writer';
-import type { BackendSystem } from '@sap-ux/store';
+import type { BackendSystem, DataAccess } from '@sap-ux/store';
 import { BackendSystemKey, getFilesystemStore, SystemService } from '@sap-ux/store';
 import { getBackendSystemType } from '../prompt-helpers';
 import type { ListChoiceOptions } from 'inquirer';
@@ -189,19 +189,20 @@ function getBackendSystemTypeName(systemType?: string): string {
 /**
  * Adds the system type to backend systems for all systems.
  * Should be removed in a number of sprints.
+ *
+ * @param fileSystemStore - the file system access for backend systems
  */
-async function addSystemTypeToBackendSystems(): Promise<void> {
+async function addSystemTypeToBackendSystems(fileSystemStore: DataAccess<BackendSystem>): Promise<void> {
     const allSystems = await new SystemService(LoggerHelper.logger).getAll();
     for (const system of allSystems) {
         // Set the system type based on the authentication type
         const systemType = getBackendSystemType(system);
         // Update the backend system in the systems.json file
-        await new SystemService(LoggerHelper.logger).partialUpdate(
-            BackendSystemKey.from(system as BackendSystem) as BackendSystemKey,
-            {
-                systemType
-            }
-        );
+        await fileSystemStore.partialUpdate({
+            entityName: 'system',
+            id: BackendSystemKey.from(system).getId(),
+            entity: { systemType }
+        });
     }
 }
 
@@ -307,8 +308,8 @@ export async function createSystemChoices(
         // to be removed in a number of sprints
         if (requiresSystemTypeMigration(backendSystems)) {
             // there are backend systems without a system type, so we perform a one-time migration to set the system type
-            LoggerHelper.logger.info(t('info.systemTypeMigration'));
-            await addSystemTypeToBackendSystems();
+            LoggerHelper.logger.debug(t('info.systemTypeMigration'));
+            await addSystemTypeToBackendSystems(fileSystemStore);
             backendSystems = await fileSystemStore.getAll({ entityName: 'system' });
         }
 
--- a/packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts
+++ b/packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts
@@ -10,9 +10,8 @@ import {
 } from '@sap-ux/btp-utils';
 import { ERROR_TYPE } from '@sap-ux/inquirer-common';
 import type { OdataVersion } from '@sap-ux/odata-service-writer';
-import type { BackendSystem, DataAccess } from '@sap-ux/store';
-import { BackendSystemKey, getFilesystemStore, SystemService } from '@sap-ux/store';
-import { getBackendSystemType } from '../prompt-helpers';
+import type { BackendSystem } from '@sap-ux/store';
+import { type BackendSystemKey, SystemService } from '@sap-ux/store';
 import type { ListChoiceOptions } from 'inquirer';
 import { t } from '../../../../i18n';
 import type { ConnectedSystem, DestinationFilters } from '../../../../types';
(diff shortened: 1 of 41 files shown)


packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -186,6 +186,9 @@
             "hint": "Enables page level variant."
         }
     },
+    "info": {
+        "systemTypeMigration": "Performing a one-time migration to set system type for all backend systems."
```
```suggestion
        "systemTypeMigration": "Performing a one-time migration to set the system type for all back-end systems..."
```

lfindlaysap (changes requested):
@cianmSAP, my suggestion.

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
cianmSAP: Thanks, have remove `backend` as it's actually possibly misleading anyway. Corrected the rest

packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts
```diff
@@ -157,14 +163,51 @@ export async function connectWithDestination(
...
+/**
+ * Returns the formatted system type name for the given backend system.
+ *
+ * @param systemType the system type to get the name for
+ * @returns system type name formatted as a string, e.g. " (BTP)" or " (S4HC)".
+ */
+function getBackendSystemTypeName(systemType?: string): string {
+    let systemTypeName = ''; // for on prem we do not show the system type
+    if (systemType === 'BTP' || systemType === 'S4HC') {
+        systemTypeName = ` (${systemType})`;
```
IainSAP: I think names should be externalised, as was previously done, they have changed in the past and we dont want to have to update the stored types again! :
```suggestion
        systemTypeName = systemType === 'S4HC'` ? t('texts.systemTypeS4HC'} : t('texts.systemTypeBTP'};
```

packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts
```diff
@@ -157,14 +163,51 @@ export async function connectWithDestination(
...
+    return systemTypeName;
+}
+
+/**
+ * Adds the system type to backend systems for all systems.
+ * Should be removed in a number of sprints.
+ *
+ * @param fileSystemStore - the file system access for backend systems
+ */
+async function addSystemTypeToBackendSystems(fileSystemStore: DataAccess<BackendSystem>): Promise<void> {
```
IainSAP: Im not sure odata-service-inquirer is the right place for this work. I think it should be in the `store` on access of backend systems. So they will be updated once any consumer accesses.

packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts
```diff
@@ -157,14 +163,51 @@ export async function connectWithDestination(
...
-    return `${system.name}${systemTypeName}${userDisplayName}`;
+}
+
+/**
+ * Simple check to see if the backend systems require a system type migration.
+ *
+ * @param systems - the list of backend systems to check
+ * @returns - true if any of the backend systems do not have a system type, false otherwise
+ */
+function requiresSystemTypeMigration(systems?: BackendSystem[]): boolean {
```
IainSAP: Same comment, Im not sure odata service inquirer is the right home for this code as this will end up being copied into all consumers of backend systems.

packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts
```diff
@@ -252,7 +295,18 @@ export async function createSystemChoices(
             };
         }
     } else {
-        const backendSystems = await new SystemService(LoggerHelper.logger).getAll();
+        // Fetch backend systems from systems.json file, credential are fetched once the system is selected
+        const fileSystemStore = getFilesystemStore<BackendSystem>(LoggerHelper.logger);
```
IainSAP: Accessing the file store should only be done in the `store` module. I think all the migration code should be there as its not relevant to odata-service-inquirer and what about other users of the `store` like SAP Systems Panel. It would not benefit from the migration at all.

cianmSAP: It would if the SAP Systems panel also takes the same approach and uses the file for the initial loading of systems

packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts
cianmSAP: Not necessary?

IainSAP: ~~Shouldnt system type be a string type literal which can be translated into an end user system type label~~ I see it is..thx. At the moment, the system type and label are the same, but in the past the label was something else and so decided to externalise it to isolate changes. If S4HC gets rebranded for example we would have to update the type and not just the label!

cianmSAP: I've reverted to use the labels again

packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts
cianmSAP: Moved the logic to the store module

packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts
cianmSAP: SAP Systems updates will be minimal now, similar to odata serivce inquirer

packages/odata-service-inquirer/src/prompts/datasources/sap-system/validators.ts
```diff
@@ -11,7 +11,7 @@ import { readFileSync } from 'fs';
  * @returns true if the system name is already in use, otherwise false
  */
 async function isSystemNameInUse(systemName: string): Promise<boolean> {
-    const backendSystems = await new SystemService(LoggerHelper.logger).getAll();
+    const backendSystems = await new SystemService(LoggerHelper.logger).getAll({ includeSensitiveData: false });
```
IainSAP, line 14: 👍

IainSAP (approved):
Thanks @cianmSAP 

- odata-service-inquirer updates as expected
- valuable features to enhance security to store to hide sensitive data
- migration trigger for relevant store access as agreed
- tests to cover

devinea (approved):
code is clear and well tested. 
was a little concerned at first glance that migrated is being called in frequenty used function `getAll`. But is seems to be coded that it should not be an expense process after the first call (assuming all updates are successful). 
I didn't test locally. 
Changeset ✅

cianmSAP:
> code is clear and well tested. 
> 
> was a little concerned at first glance that migrated is being called in frequenty used function `getAll`. But is seems to be coded that it should not be an expense process after the first call (assuming all updates are successful). 
> 
> I didn't test locally. 
> 
> Changeset ✅ 

Thanks @devinea, yes the call to the store and partial updates will only be required once. We can look to eventually remove the migration in a few sprints

PR #3391, opened 2025-06-26
fix(launch-config): fix start mock args for launch config

cianmSAP:
Ensure `Start xxx Mock` launch config has the correct args (html file) in a virtual endpoint scenario


kjose90 (approved):
Changes looks correct ✅ 
Test to cover

PR #3393, opened 2025-06-26
fix: Various bugs in FLP config generator for Adaptation Project

mmilko01:
#3397

* Flex layer is not taken from `manifest.appdescr_variant` in case of standalone FLP generator, or passed as a parameter from paranet generator in case of subgenerator scenario. Based on the layer, the correct prefix is added to the `InboundID`
* Introduces a temporary fix to filter out Base Application's parameters in case if find invalid ones. This is a temporary fix until e permanent fix is introduced in ABAP Platform Cloud 2508.

### DIFF
--- a/examples/fe-fpm-cli/CHANGELOG.md
+++ b/examples/fe-fpm-cli/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/fe-fpm-cli
 
+## 0.1.13
+
+### Patch Changes
+
+-   @sap-ux/fe-fpm-writer@0.35.9
+
 ## 0.1.12
 
 ### Patch Changes
--- a/examples/fe-fpm-cli/package.json
+++ b/examples/fe-fpm-cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/fe-fpm-cli",
-    "version": "0.1.12",
+    "version": "0.1.13",
     "description": "A simple CLI to prompt required information to create a building block using the fe-fpm-writer module's prompt and generate functions.",
     "license": "Apache-2.0",
     "private": true,
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,14 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.37
+
+### Patch Changes
+
+-   @sap-ux/fiori-elements-writer@2.5.9
+-   @sap-ux/fiori-freestyle-writer@2.4.20
+-   @sap-ux/axios-extension@1.22.3
+-   @sap-ux/system-access@0.6.9
+
 ## 1.1.36
 
 ### Patch Changes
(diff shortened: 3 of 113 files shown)


packages/create/src/cli/add/navigation-config.ts
```diff
@@ -106,20 +113,21 @@ async function addInboundNavigationConfig(basePath: string, simulate: boolean, y
...
-    logger: ToolsLogger
+    logger: ToolsLogger,
+    variant?: DescriptorVariant
 ): Promise<ManifestNamespace.Inbound | undefined> {
     if (isAdp) {
-        const appId = (await getVariant(basePath)).reference;
         const { target, ignoreCertErrors = false } = await getAdpConfig(basePath, yamlPath);
         const provider = await createAbapServiceProvider(target, { ignoreCertErrors }, true, logger);
-        return getBaseAppInbounds(appId, provider);
+        return getBaseAppInbounds(variant?.reference as string, provider);
```
heimwege: `variant?.reference` is `string | undefined`. So the type cast to `string` is actually not correct. If you want a `string` you need to do something like `variant?.reference as string | undefined ?? ''`

packages/create/src/cli/add/navigation-config.ts
```diff
@@ -81,7 +88,7 @@ async function addInboundNavigationConfig(basePath: string, simulate: boolean, y
         }
 
         if (isAdp) {
-            const config = getAdpFlpInboundsWriterConfig(answers, tileSettingsAnswers);
+            const config = getAdpFlpInboundsWriterConfig(answers, variant?.layer as FlexLayer, tileSettingsAnswers);
```
heimwege: `variant?.layer` is `FlexLayer | undefined`. Type cast to `FlexLayer` could cause issues at runtime.

heimwege (commented):
Checked the create package. See comments

packages/create/src/cli/add/navigation-config.ts
mmilko01: As discussed the existence of the variant depends on the scenario. To handle this I added a new type that can handle both cases and can serve as a type guard.
Changed in: 0e037d629bdb457f647427e5d0a2190eae2bcd32

packages/create/src/cli/add/navigation-config.ts
mmilko01: Same as above.
Changed in: 0e037d629bdb457f647427e5d0a2190eae2bcd32

heimwege (approved):
checked the create package

- changeset ok
- test coverage great
- changes are straightforward
- review comments have been addressed
- did NOT test manually

nikmace (dismissed):
Changeset is good
Changes look good
Coverage is excellent
Did not test manually

nikmace (dismissed):
re-approving

heimwege (dismissed):
re-approve

heimwege (approved):
re-approve

PR #3395, opened 2025-06-26
fix(cf-deploy-config-inquirer): long path warning

korotkovao:
## #3341

Edit the error message for long Windows paths.

### DIFF
--- a/packages/cf-deploy-config-inquirer/test/validators.test.ts
+++ b/packages/cf-deploy-config-inquirer/test/validators.test.ts
@@ -230,7 +230,12 @@ describe('validateMtaId long Windows path', () => {
         };
         const combinedLength = `${longPath}\\${input}`.length;
         const result = validateMtaId(input, previousAnswers);
-        expect(result).toBe(t('error.windowsMtaIdPathTooLong', { length: combinedLength }));
+        expect(result).toBe(
+            t(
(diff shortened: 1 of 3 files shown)


longieirl (approved):
Changes look good.
Changeset is good.

packages/cf-deploy-config-inquirer/test/validators.test.ts
```diff
@@ -230,7 +230,12 @@ describe('validateMtaId long Windows path', () => {
         };
         const combinedLength = `${longPath}\\${input}`.length;
         const result = validateMtaId(input, previousAnswers);
-        expect(result).toBe(t('error.windowsMtaIdPathTooLong', { length: combinedLength }));
+        expect(result).toBe(
+            t(
```
cianmSAP: I couldn't understand how this was working, as you are still using the `t` for translating but add a full string. 

I checked out the branch and you are mocking the `t` function, everything is returning `MTA ID too long` so the tests are not correct

cianmSAP (changes requested):
Test covers

cianmSAP (approved):
Approving as I requested the change but we should get another reviewer

longieirl (approved):
Latest changes look good.

korotkovao:
Thank you @cianmSAP @longieirl

Issue #3399, opened 2025-06-27
TBI - Update standalone router configuration


longieirl:
### Description (include screenshots)
The latest approuter configuration, generated by cds v8 and v9 is out of sync with the cf generated router.

V8 version of CDS generated approuter;
```bash
cds init standaloneAppCAPJLProject && cd  standaloneAppCAPJLProject && npm install && cds add tiny-sample && cds add hana mta xsuaa connectivity destination html5-repo approuter && cds build --for hana && cp gen/db/package.json db && npm i
```
The router is generated wit the following properties;
```json
{
  "name": "approuter",
  "dependencies": {
    "@sap/approuter": "^21.0.0"
  },
  "scripts": {
    "start": "node node_modules/@sap/approuter/approuter.js"
  }
}

```

### Value
Alignment with external tools.

### Tasks
- [ ] Update router configuration
- [ ] Update tests
- [ ] Update readme if required

PR #3400, opened 2025-06-27
fix(deloy-config-sub-generator): update standalone approuter

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/3399

Align the standalone approuter with the latest generated from CDS v8 and v9.


packages/cf-deploy-config-writer/templates/router/package.json
```diff
@@ -1,18 +1,10 @@
 {
     "name": "app-router",
-    "private": true,
```
cianmSAP, line 3: Was this always incorrect or why the change?

cianmSAP (commented):
Updates look good
One query & I don't think `@sap-ux/cf-deploy-config-sub-generator` needs a changeset

packages/cf-deploy-config-writer/templates/router/package.json
longieirl: The approuter configurations have always been a replication of the route folder generated by cds and in the latest version, this property is now removed. Here are some notes that I've taken from different forums;
Cleaner Build Pipelines:
Some SAP Cloud SDK and CAP deployment setups involve using tools like npm pack, npm ci, or custom scripts to bundle and deploy services (including approuter). Having "private": true would block packaging steps in these cases.

CDS Tooling Evolution:
SAP’s Yeoman-based generators and cds add approuter have evolved toward being more flexible and universal. Not enforcing private: true reduces friction and lets developers decide whether they want that flag based on their own packaging and deployment needs.

Issue #3401, opened 2025-06-27
BUG - UI Service Gen - namespace in service data not used for metadata retrieval


docirl:
### Related Feature
Feature request: _issue number_

### Description
In the UI Service Generator, if a package uses a namespace (and therefore the service generated in that namespace) when adding the service to the Storyboard project the request to retrieve the service metadata will fail as it call the wrong url.

### Steps to Reproduce
1. Create a Storyboard project in BAS and click to add a datasource, pick a system and BO
2. Enter package with a namespace and fill other prompts as needed
3. Generate service
4. When service is generated, it will fail to add to project (the service is created successfully in backend)

### Expected results
Service is added to the Storyboard project when a namespace is used

### Actual results
Service is generated but not added to the project

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

docirl:
fixed by https://github.com/SAP/open-ux-tools/pull/3402

PR #3402, opened 2025-06-27
fix(ui-service-gen): handle namespace for storyboard metadata retrieval

docirl:
https://github.com/SAP/open-ux-tools/issues/3401

- use namespace from service data correctly in metadata retrieval
- update tests


packages/ui-service-sub-generator/src/app/utils.ts
```diff
@@ -108,12 +108,16 @@ export async function writeBASMetadata(
  */
 export function getRelativeUrlFromContent(content: string): string {
     const contentJson = JSON.parse(content);
-    let serviceNamespace = SAP_NAMESPACE;
-    if (contentJson.general?.namespace?.startsWith('/')) {
-        // set namespace to value from suggested content, removing leading and trailing slashes
-        serviceNamespace = contentJson.general.namespace.replace(/^\/|\/$/g, '');
+    let serviceBinding = contentJson.businessService.serviceBinding.serviceBindingName.replace(/^\/|\/$/g, '');
```
longieirl: Should you make this null safe?

docirl: Done, although this value wont be null as it is checked earlier in prompt phase and blocks if empty, and this code is post gen.

PR #3412, opened 2025-07-02
Feat/deprecate cards editor middlewares

heimwege:
Deprecate `cards-editor-*` packages as the functionality has been integrated int `preview-middleware` and the config has been integrated into `app-config-writer` in #3167

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,6 +11,38 @@
     ],
     "configurations": [
         {
+            "request": "launch",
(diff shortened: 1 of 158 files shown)


chaudhary1-rohit (approved):
Thanks for the change 😄

johannes-kolbe (approved):
Updates to `create` package and removal of unused `@sap-ux/cards-editor-config-writer` lgtm

PR #3383, opened 2025-06-25
feat(general): improve Fiori Tools UI Texts

korotkovao:
## #3382

Improve Fiori Tools UI Texts

PR #3413, opened 2025-07-03
Align text on mta message and non mta deploy config

devinea:
Follow-on from https://github.com/SAP/open-ux-tools/pull/3383


cianmSAP (approved):
Text updates look good
Appropriate test and changeset

Issue #3415, opened 2025-07-03
BUG - UI Service Gen - show only abap destinations


docirl:
### Description
Only abap destinations should be shown the system selection dropdown, showing other destination types will lead to errors and inconsistencies in prompting flow.

Service Center is already filtering like this, so this will align both tools.

### Steps to Reproduce
1. Go to BAS and create dev space
2. Open UI Service Gen
3. Expand system selection dropdown
4. See destinations that are not abap backend, such as Northwind

### Expected results
Only abap backend destinations are displayed

### Actual results
All destinations are displayed

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

docirl:
Done with https://github.com/SAP/open-ux-tools/pull/3416

PR #3416, opened 2025-07-03
fix(ui-service-gen): show only abap destinations

docirl:
https://github.com/SAP/open-ux-tools/issues/3415

- add filter to only show abap destinations
- this aligns with Service Center, now the same destinations are returned in both places
- tests in odata-service-inq already cover the filtering logic


longieirl (approved):
Changeset is good.
Thanks for adding tests.
Change looks good.

cianmSAP (approved):
Update is clear and covered by test
Changeset is appropriate for the change

PR #3417, opened 2025-07-03
fix(ui5-config): export getPreviewMiddlewareConfig and utilise existing ui5 config methods

cianmSAP:
Removes `addFioriToolsPreviewMiddleware` as `getPreviewMiddlewareConfig` is now exported, meaning `addFioriToolsPreviewMiddleware` does not provide any benefit, and allows consumers to use existing `updateCustomMiddleware` method.

### DIFF
--- a/packages/ui5-config/test/index.test.ts
+++ b/packages/ui5-config/test/index.test.ts
@@ -389,46 +389,6 @@ describe('UI5Config', () => {
                     .trim()
             );
         });
-
-        test('Should update preview middleware when it is exists', () => {
-            ui5Config.updateCustomMiddleware({
-                name: 'fiori-tools-preview',
-                afterMiddleware: 'fiori-tools-appreload',
-                configuration: {
-                    flp: {
-                        theme: 'sap_fiori_3'
-                    }
-                }
-            });
-            ui5Config.updateCustomMiddleware({
-                name: 'fiori-tools-preview',
-                afterMiddleware: 'fiori-tools-appreload',
-                configuration: {
-                    flp: {
-                        theme: 'sap_fiori_3',
-                        path: 'test/flpSandbox.html',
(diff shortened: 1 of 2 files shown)


packages/ui5-config/test/index.test.ts
```diff
@@ -357,10 +361,94 @@ describe('UI5Config', () => {
...
+                            intent:
+                                object: testapp
+                                action: display
+                `
+                    .replace(/\s+/g, ' ')
+                    .trim()
+            );
+        });
+
+        test('Should not preview middleware when it is exists and update is set to false', () => {
```
devinea:
```suggestion
        test('Should not update preview middleware when it is exists and update is set to false', () => {
```
typo?

devinea (dismissed):

packages/ui5-config/src/ui5config.ts
```diff
@@ -278,27 +278,30 @@ export class UI5Config {
...
+    ): this {
+        let existingMiddleware: CustomMiddleware<FioriPreviewConfig> | undefined;
+        try {
+            existingMiddleware = this.findCustomMiddleware<FioriPreviewConfig>(fioriToolsPreview);
+        } catch {
+            // no existing fiori preview middleware found
+        }
+
+        if (existingMiddleware) {
+            if (!update) {
```
kjose90:
```
if (existingMiddleware && !update){
    // don't update existing middleware
    return this;
}
```

cianmSAP: Thanks, I originally had logic outside outside of the `if (!update) {` block but I realised `updateCustomMiddleware` handled it

kjose90 (approved):
Re approving 
Looks good after reusing existing UI5Config methods 👍

devinea (approved):
re-approving after refactoring. 👍

PR #3420, opened 2025-07-04
feat: Add optional  isActive parameter to  isExtensionInstalled function

kjose90:
- Enhanced the `isExtensionInstalled` function to include an optional `isActive` parameter.
- The `isActive` parameter defaults to `false` and allows checking if the extension is active.


devinea (approved):
code changes are clear and covered by tests.
changeset ✅

PR #3430, opened 2025-07-09
Fix/add virtual preview adt quick deploy

kjose90:
Add `useVirtualPreviewEndpoints` true  to App Configuration for ADT Quick Deploy Apps by default to enable virtual preview


cianmSAP (commented):
Is there any tests to actually check the output has will be updated with this change?

kjose90:
> Is there any tests to actually check the output has will be updated with this change?

Once you pass `useVirtualPreviewEndpoints: true` to `fiori-elements-writer`, it is expected to be covered in the tests located [here](https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-elements-writer/test/lrop.test.ts). There are already a couple of tests in this file that validate the behaviour when `useVirtualPreviewEndpoints` is enabled, which should be sufficient ?

cianmSAP:
> Once you pass `useVirtualPreviewEndpoints: true` to `fiori-elements-writer`, it is expected to be covered in the tests located [here](https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-elements-writer/test/lrop.test.ts). There are already a couple of tests in this file that validate the behaviour when `useVirtualPreviewEndpoints` is enabled, which should be sufficient ?

Those are FE writer tests though - we can't really rely on other modules to test the functionality. There is a function `verifyGeneratedFiles` in this module, it would be best to verify atleast the flpSandbox.html for example is not generated, the start script in the `package.json` is correct. It does not have to be full integration test. It is now the default to use virtual endpoints, so it should have some sort of sanity test that the physical files produced by this generator are correct

kjose90:
> > Once you pass `useVirtualPreviewEndpoints: true` to `fiori-elements-writer`, it is expected to be covered in the tests located [here](https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-elements-writer/test/lrop.test.ts). There are already a couple of tests in this file that validate the behaviour when `useVirtualPreviewEndpoints` is enabled, which should be sufficient ?
> 
> Those are FE writer tests though - we can't really rely on other modules to test the functionality. There is a function `verifyGeneratedFiles` in this module, it would be best to verify atleast the flpSandbox.html for example is not generated, the start script in the `package.json` is correct. It does not have to be full integration test. It is now the default to use virtual endpoints, so it should have some sort of sanity test that the physical files produced by this generator are correct

Ah ok - thats what you meant ...  - ty 
https://github.com/SAP/open-ux-tools/pull/3430/files#diff-238845d2cef532fc00921e084d453001094272e5bb279737085e172c14a10798R240

cianmSAP (approved):

cianmSAP:
Comments addressed,
Thanks @kjose90

PR #3432, opened 2025-07-11
Fix/preview middleware/no default for rta layer

heimwege:
According to the documentation `CUSTOMER_BASE` is the default rta layer, but currently it's not. This is fixed in this PR.


ytpo-lyne (approved):
- code changes for rta layer are straight forward
- changeset exist 
- greate coverage
- did not test manually

Thanks, @heimwege. Approving.

PR #3433, opened 2025-07-11
fix(ui5-app-inq): adds a validation for fiori app folder before target folder is prompted

cianmSAP:
Internal issue 35025

Update to add check for fiori app to before the suggested default target folder is passed to the prompt.
Prevents suggesting a folder which will immediately return a validation error


docirl (approved):
Changes are clear
Tests updated
Changeset ok

PR #3435, opened 2025-07-11
Fix(preview middleware): adjust default test pattern to avoid picking up wrong files

heimwege:
The old pattern used `.*` which could lead to e.g. `.html` files being picked up. The pattern is adjusted to only consider .js or .ts files in this PR.


heimwege:
@petermuessig FYI

Klaus-Keller (approved):
Thanks @heimwege!

* change to select ts and js files looks good
* changeset exists
* visual review, no manual test
* README adjusted

Approved from my side

Issue #3436, opened 2025-07-14
BUG - '@sap-ux/project-access' - method `getSpecificationPath` returns corrupted path when local specification is linked


815are:
### Steps to Reproduce
1. Setup fiori application to point to local `@sap/ux-specification'` source - use `nom link' or ` "@sap/ux-specification": "file:./....",` approach
2. Call function `getSpecificationPath`

Problem, following wrong path is returned:

<img width="1384" height="476" alt="Image" src="https://github.com/user-attachments/assets/b127bd56-d57f-4f39-aef4-71629cd88080" />

### Expected results
Specification folder path is resolved when local specification is linked

### Actual results
Specification folder path is wrong when local specification is linked

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

815are:
PR with fix - https://github.com/SAP/open-ux-tools/pull/3437

PR #3437, opened 2025-07-14
fix(project-access): method getSpecificationPath returns corrupted path when local specification is linked

815are:
Issue #3436

### Steps to Reproduce
1. Setup fiori application to point to local `@sap/ux-specification'` source - use `nom link' or ` "@sap/ux-specification": "file:./....",` approach
2. Call function `getSpecificationPath`

Problem, following wrong path is returned:

Image

Changes:
1. As fix - if original approach with `node_modules/@sap/ux-specification` path fails, then resolve path using `packages/specification/dist` path as target.
2. Throw error if path is failed to resolve;

### DIFF
--- a/.changeset/eleven-rabbits-hammer.md
+++ b/.changeset/eleven-rabbits-hammer.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': patch
+---
+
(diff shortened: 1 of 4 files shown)


815are:
issue addressed in https://github.com/SAP/open-ux-tools/pull/4894

Issue #3438, opened 2025-07-14
BUG - Remove temporary filtering of parameter properties for inbounds


mmilko01:
### Related Feature
Feature request: #3319

### Description

The temporary code introduced to filter some properties of the inbound parameters needs to be removed as this is now handled by the API.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3440, opened 2025-07-14
fix: remove filtering of parameters for inbounds

mmilko01:
#3438

* Removed temporary code for filtering inbound parameters as it is done by the API
* Changed condition for filtering inbounds based on `hideLauncher` property. Only inbounds with explicit `hideLauncher=true` should be filtered out


nikmace (approved):
Changeset looks good
Changes look good
Did not test manually

IvoSG (dismissed):
Changes look good
Enhanced tests
Did not test manually

packages/flp-config-inquirer/src/translations/flp-config-inquirer.i18n.json
```diff
@@ -14,15 +14,13 @@
     },
     "tooltips": {
         "inboundId": "Choose the inbound ID to be used to set the SAP Fiori launchpad tile configuration.",
-        "additionalParameters": "Enter the value for the parameters in JSON format."
+        "additionalParameters": "Enter the value for the parameters in JSON format.",
+        "semObjectActionDuplication": "When generating a new tile, you must modify the Semantic Object value, the Action value, or both. You can use the default suggestions, which include a '_New' suffix for each value or use your own designations."
```
lfindlaysap:
```suggestion
        "semObjectActionDuplication": "When generating a new tile, you must modify the 'Semantic Object' value, the 'Action' value, or both. You can use the default suggestions, which include a '_New' suffix for each value or use your own designations."
```

lfindlaysap (changes requested):
@mmilko01, one small suggestion.

packages/flp-config-inquirer/src/translations/flp-config-inquirer.i18n.json
mmilko01: Fixed with: 7b9d28edbbf666f2f443964e2f550db86edd1777

lfindlaysap (approved):

IvoSG (approved):
Re-approval

PR #3441, opened 2025-07-14
Feat(preview-middleware): change file type of templates to ejs

heimwege:
Currently the template files are .html which leads to the IDE showing errors because the templates are no correct HTML. If we change the file type to ejs we get a proper IDE support.


johannes-kolbe (dismissed):
Lgtm,
- changeset provided
- file name change makes sense
- did not test manually

heimwege:
@johannes-kolbe can you please re-check? I had to solve a merge conflict which dismissed your review. Thanks

johannes-kolbe (approved):
reapproving

IainSAP (approved):
Thanks @heimwege 

- renames as appropriate
- no test changes as expected (output filenames are unchanged)

Issue #3442, opened 2025-07-14
TBI - Remove archive.zip from deploy package.json script


korotkovao:
### Description (include screenshots)
Improve the ability for developers to pass CLI params to the deploy task, specifically for the ABAP flow

An ABAP generated HTML5 application with deployment configuration added, is appended with `rimraf archive.zip` in the `deploy` task, for example;

```json
 "deploy": "npm run build && fiori deploy --config ui5-deploy.yaml && rimraf archive.zip",
```

This blocks users from passing cli params when running the `npm run deploy` command,  for example 
```bash
npm run deploy --strict-ssl false
```

In order to support additional CLI params being pased, the user has to update the `package.json` deploy script manually, for example;

```json
 "deploy": "npm run build && fiori deploy --config ui5-deploy.yaml --strict-ssl false && rimraf archive.zip",
```

The suggested improvement is to remove `rimraf archive.zip` from the `deploy` task;
```json
 "deploy": "npm run build && fiori deploy --config ui5-deploy.yaml",
```

This change would allow a developer to run the `deploy` task and pass any CLI param they want;
```bash
npm run deploy --strict-ssl false
```

Extra notes;
The `archive.zip` is already appended to `.gitignore` which is generated each time the deploy task is executed.

### Value
More versability in being able to run CLI commands.

korotkovao:
Complete

PR #3443, opened 2025-07-14
feat(abap-deploy-config-sriter): remove archive.zip from deploy #3442

korotkovao:


.changeset/sharp-worms-chew.md
```diff
@@ -0,0 +1,8 @@
+---
+'@sap-ux/abap-deploy-config-sub-generator': minor
+'@sap-ux/deploy-config-sub-generator': minor
```
cianmSAP: `@sap-ux/deploy-config-sub-generator` & `@sap-ux/abap-deploy-config-sub-generator` do not need to be added to the changeset, they only have test updates

packages/abap-deploy-config-writer/src/index.ts
```diff
@@ -60,7 +60,7 @@ async function generate(
 
     // package.json
     // deploy script
-    const deployScript = `${BUILD_SCRIPT} && ${DEPLOY_SCRIPT} --config ${deployConfigFile} && ${RIMRAF} archive.zip`;
+    const deployScript = `${BUILD_SCRIPT} && ${DEPLOY_SCRIPT} --config ${deployConfigFile}`;
```
cianmSAP, line 63: What happens to the `archive.zip` after deployment? Is it agreed that the archive.zip will no longer be removed?

cianmSAP (changes requested):

packages/abap-deploy-config-writer/src/index.ts
cianmSAP: Query answered here https://github.wdf.sap.corp/ux-engineering/tools-suite/issues/34368#issuecomment-14257310

korotkovao: We are not removing `dist` either.` Archive.zip` is treated the same as the `dist` folder, it's generated each time a user tries to deploy, it's in the `gitignore` file already and shouldn't be checked into a source control. 
And we enable users to pass the additional `parameters` without having to edit the` package.json`

cianmSAP (approved):
Comments addressed and answered
Changeset and test updates look good
Looks good from my side

packages/abap-deploy-config-sub-generator/test/app.test.ts
```diff
@@ -135,7 +135,7 @@ describe('Test abap deploy configuration generator', () => {
         );
 
         expect(pkgJson.scripts).toStrictEqual({
-            'deploy': 'npm run build && fiori deploy --config ui5-deploy.yaml && rimraf archive.zip',
+            'deploy': 'npm run build && fiori deploy --config ui5-deploy.yaml',
```
IvoSG, line 138:
```suggestion
            'deploy': 'npm run build && fiori deploy --config ui5-deploy.yaml',
            'postdeploy': 'rimraf archive.zip',
```

korotkovao: Thank you @IvoSG, 
As a follow-up to our discussion, we agreed just to remove `rimraf archive.zip`
I have reverted the `postdeploy` script additions.

kjose90 (approved):
use of `post-deploy` looks better - Thanks @korotkovao

korotkovao:
As a follow-up to our discussion, we agreed to just remove rimraf archive.zip
I have reverted the` postdeploy` script additions.

cianmSAP (approved):
Re-approving

korotkovao:
Thank you All

PR #3444, opened 2025-07-15
fiori-sub-gen: update `FioriGeneratorPromptExtension` types

cianmSAP:
- Updates all prompt in the two sub gens - Deploy & FLP - to additionally support `CommonPromptOptions`.
- Corrects prompt options logic in CF Inquirer to use `hide` property to determine whether a prompt is shown.
- Deprecates `FioriAppGeneratorOptions.extensions`
- Enhances `FioriAppGeneratorPromptSettings` used by `FioriAppGeneratorOptions.promptSettings`

Note - the goal is to eventually remove [this](https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-app-sub-generator/src/fiori-app-generator/prompting.ts#L329-L337) logic, and for all extensions and adaptors to use `FioriAppGeneratorOptions.promptSettings`


packages/fiori-app-sub-generator/src/types/common.ts
```diff
@@ -21,15 +21,15 @@ export type WorkspaceFolder = {
     path: string;
 };
 
-type DeploySubGenPromptOptions = {
-    targetName: CommonPromptOptions;
-};
+export type DeploySubGenPromptOptions =
```
IainSAP: Shouldnt this type be defined in `@sap-ux/deploy-config-sub-generator`. Anyway we should already have that dep since we `composeWith` from this gen?

cianmSAP: Yes I'd have preferred to all of this in `@sap-ux/deploy-config-sub-generator` but it wasn't easily doable due to circular dependencies.
`@sap-ux/@sap-ux/fiori-app-sub-generator` is already a dev dep in `@sap-ux/deploy-config-sub-generator` because of the `AppConfig` needed for headless.

Moving `AppConfig` into the shared module also comes with it's issues

cianmSAP: Should be able to resolve once https://github.com/SAP/open-ux-tools/pull/3466 goes in

cianmSAP: The type is now consumed from `@sap-ux/deploy-config-sub-generator`

kjose90 (approved):
Thanks @cianmSAP - Nice work on aligning the types
changeset & tests ✅

Issue #3448, opened 2025-07-16
FEATURE - @sap-ux/lnodejs-utils: Add support for a passed logger reference


IainSAP:
### Description

To promote reuse of `CommandLogger` we should add an option to the `run` function to support passing of a logger.

### Technical Design
_What modules need to be created or enhanced?_ 
_How would the functionality be exposed and how can it be consumed?_

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a ...
**when** I execute ...
**then** ... will happen.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #3452, opened 2025-07-17
feat(nodejs-utils): Add logging support to `CommandRunner`

IainSAP:
https://github.com/SAP/open-ux-tools/issues/3448

- Adds logger support to the command runner
- Removes trailing carriage returns/newline since typically command runner will be piping log statements from a chile process which already includes carriage returns/newline
- Adds test to cover


kjose90 (approved):
Changes are clear.
Changeset and tests ✅

PR #3458, opened 2025-07-18
test(preview-middleware): add adaptation editor integration tests

voicis:
Add a new workflow for running Adaptation Editor integration tests. In these tests ABAP backend is mocked and applications are generated on the fly for each test case. Currently only one SAP Fiori Elements for OData V2 project is supported, but more can be added and customized if that is required in test cases. 

UI5 versions used in tests are dynamically loaded from cache and can be updated using `Update UI5 Maintenance Versions` workflow. For each version a different job is triggered and runs in parallel, later all the results are merged back together and the combined Playwright report can be downloaded and viewed.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,0 +1,98 @@
+name: Integration Tests
+
+on:
+    workflow_dispatch:
+    # workflow_run:
+    #     workflows: ["CI/CD Pipeline"]
+    # workflow_call:
+    #     # workflows: ["CI/CD Pipeline"]
+    #     secrets:
+    #         personal_access_token:
+    #             required: true
+        
+jobs:
+    ui5-versions:
+        runs-on: ubuntu-latest
+        timeout-minutes: 5
+        outputs:
+            ui5_versions: ${{ steps.output-ui5-versions.outputs.ui5_versions }}
+        steps:
+            - name: Read UI5 versions from cache
+              uses: actions/cache/restore@v4
+              with:
+                path: |
+                  tests/integration/adaptation-editor/versions.json
+                key: ui5-versions
+                restore-keys: ui5-versions-
+            - name: Output UI5 versions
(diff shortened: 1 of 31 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,90 @@
+name: Integration Tests
+
+on:
+    workflow_dispatch:   
+jobs:
+    ui5-versions:
+        runs-on: ubuntu-latest
```
Jimmy-Joseph19, line 7: should we also run it on windows? or will have a seperate job for it

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,90 @@
+name: Integration Tests
+
+on:
+    workflow_dispatch:
```
Jimmy-Joseph19, line 4: is it not for every PR?

Jimmy-Joseph19: something like
on:
    workflow_dispatch:
    pull_request:
        paths:
            - 'packages/preview-middleware-client/**'
            - 'packages/control-property-editor/**'
            - 'packages/adp-preview/**'

.github/workflows/integration-tests.yml
voicis: We can look into this in the future, but I'm not sure if it is worth it to run all the UI tests on windows. OS differences should be handled by non UI tests.

.github/workflows/integration-tests.yml
voicis: No, not yet. The tests are not stable enough yet and we don't want to block PRs due to that. Also triggering logic should be more sophisticated - check dependencies because they could have an impact on the test results.

devinea (approved):
New GH Action setup seems like a good approach to run in parallel to main build and just  on 1 operating system.
New test/ module  looks good.

PR #3463, opened 2025-07-22
feat: enable jest environment to resolve reusable libraries

whydrae:
Given:

1. TS UI5 application which consumes the reusable library.
2. TS UI5 reusable library.

When trying to test a controller, which imports a reusable library dependency, the jest environment fails, because cannot resolve the dependency path.

Example PoC: https://github.com/whydrae/ui5-jest-library-poc.

This PR does fixes the following:
1. Strips the namespace from the reusable library path. If we import the dependency as `my/reusable/library/Control`, then we only leave the `Control` part and join it with the `sourcePath`.
2. Handle the `targetPath` the same we do for the root project.

Side notes: I had to use the `babel.config.json` configuration instead of the `.babelrc.json`, otherwise the reusable library files are not getting transpiled.


whydrae:
Fixed via https://github.com/SAP/open-ux-tools/pull/3487.

PR #3464, opened 2025-07-22
fix: provide templates path parameter to handle different runtime context

mmilko01:
#3465

* Optional parameter is introduced, so the templates path can be specified based on the runtime context


mmilko01:
Integration test execution:
image

PR #3466, opened 2025-07-22
move headless AppConfig type and associated types to fiori gen shared module

cianmSAP:
Moves `AppConfig` type used by headless generators into `@sap-ux/fiori-generator-shared`

This allows the removal of the `@sap-ux/fiori-app-sub-generator` dep from the deploy config sub gen


IainSAP (approved):
Thanks @cianmSAP 

- Reorg of types to avoid circular deps

PR #3467, opened 2025-07-22
feat(odata-service-inq): set analytical table when supported by main entity

docirl:
Internal issue 33754

- update defaulting logic for table type in LROP and Worklist template types to set Analytical table type if the main entity supports it
- this is determined by reusing the existing logic to filter for entities for ALP template type, it is called with the main entity for LROP and Worklist and will set the table type to analytical if the entity supports it
- the table type will always default/reset to Analytical when the entity is changed and supports Analytical, otherwise the current type is unchanged
- information message added to indicate that Analytical type is being set because the main entities supports it.
- tests added to cover new flow and snapshots updated


packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -151,7 +151,8 @@
             "choiceGrid": "Grid",
             "choiceAnalytical": "Analytical",
             "choiceResponsive": "Responsive",
-            "choiceTree": "Tree"
+            "choiceTree": "Tree",
+            "analyticalTableDefault": "The Analytical table type is chosen by default, as the selected entity supports it."
```
```suggestion
            "analyticalTableDefault": "The 'Analytical' table type is chosen by default because the selected entity supports it."
```

lfindlaysap (changes requested):
@docirl, here is my suggestion.

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
docirl: Thanks, I've updated to use this

packages/odata-service-inquirer/src/prompts/edmx/questions.ts
```diff
@@ -257,6 +272,63 @@ function getEdmxSizeInKb(edmx: string): number {
...
+ *
+ * @param templateType the template type of the application to be generated from the prompt answers
+ * @param prevAnswers the previous answers from the prompt, used to determine the main entity
+ * @param metadata the metadata (edmx) string of the service, used to determine if there are aggregate transformations
+ * @param odataVersion the OData version of the service,
+ * @returns the default table type based on the template type and previous answers or undefined if no default is set
+ */
+function getDefaultTableType(
+    templateType: TemplateType,
+    prevAnswers: EntitySelectionAnswers & TableConfigAnswers,
```
IainSAP: Looks like we can just pass `mainEntitySetName` and `tableType` rather than the entire previous answers?

packages/odata-service-inquirer/src/prompts/edmx/questions.ts
```diff
@@ -257,6 +272,63 @@ function getEdmxSizeInKb(edmx: string): number {
...
 }
 
+/**
+ * Get the default table type based on the template type and previous answers.
+ *
+ * @param templateType the template type of the application to be generated from the prompt answers
+ * @param prevAnswers the previous answers from the prompt, used to determine the main entity
+ * @param metadata the metadata (edmx) string of the service, used to determine if there are aggregate transformations
+ * @param odataVersion the OData version of the service,
+ * @returns the default table type based on the template type and previous answers or undefined if no default is set
```
IainSAP: Not sure this comment is correct: "and previous answers"

packages/odata-service-inquirer/src/prompts/edmx/questions.ts
```diff
@@ -206,18 +208,31 @@ function getTableLayoutQuestions(
...
                 hint: t('prompts.tableType.hint'),
-                breadcrumb: true
+                breadcrumb: true,
+                applyDefaultWhenDirty: true // set table type on entity selection change
             },
             choices: tableTypeChoices,
-            default: tableTypeDefault
+            default: (prevAnswers: EntitySelectionAnswers & TableConfigAnswers) => {
+                const tableTypeDefault = getDefaultTableType(templateType, prevAnswers, metadata, odataVersion);
+                setAnalyticalTableDefault = tableTypeDefault.setAnalyticalTableDefault;
```
IainSAP, line 233: Do we need this? `setAnalyticalTableDefault`. Cant we call `isAnalyticalTTSupported`, if its true set the TT to `AnalyticalTable` and if true output the message. Otherwise apply the default as was previously done?

packages/odata-service-inquirer/src/prompts/edmx/questions.ts
```diff
@@ -257,6 +272,63 @@ function getEdmxSizeInKb(edmx: string): number {
...
+ *
+ * @param metadata The metadata (edmx) string of the service.
+ * @param entitySetName The entity set name to check for aggregate transformations.
+ * @returns true if the entity set has aggregate transformations, false otherwise.
+ */
+function hasAggregateTransformationsForEntity(metadata: string, entitySetName?: string): boolean {
+    if (!entitySetName) {
+        return false;
+    }
+    return getEntityChoices(metadata, {
```
IainSAP: Dont we already call this earlier, so essentially we are parsing the metadata multiple times? This func is doing a lot more than just searching for entity sets, it also creates choices. Is that appropriate for this usage? Why not call `filterAggregateTransformations(convertedMetadata.entitySets);` directly with a cached  `convertedMetadata` to avoid reparsing

packages/odata-service-inquirer/src/prompts/edmx/questions.ts
docirl: Thanks, updated this

packages/odata-service-inquirer/src/prompts/edmx/questions.ts
docirl: Yes, I've updated to only pass those values

packages/odata-service-inquirer/src/prompts/edmx/questions.ts
docirl: I've updated to use `convertedMetadata` as it was available from the parent function, and call `filterAggregateTransformations` directly. Thanks

packages/odata-service-inquirer/src/prompts/edmx/questions.ts
docirl: Do you mean to call isAnalyticalTTSupported from both the default and additionalMessages? There is also the applyDefaultWhenDirty in play here, so I'm trying to limit the resetting of the value for other flows than this US. Thats why I have it like this currently. Maybe a short sync to discuss would be best.

docirl:
@lfindlaysap can you please re-review, I've made the text change. Thanks

lfindlaysap (approved):

PR #3469, opened 2025-07-23
Fix/35136/form data upgrades

donal-tobin-sap:


heimwege (approved):
- version bump lgtm
- changeset ok
- did not test manually

.github/workflows/pipeline.yml
```diff
@@ -19,9 +21,10 @@ jobs:
               with:
                   fetch-depth: 0
             - name: Setup pnpm
-              uses: pnpm/action-setup@v4
+              uses: pnpm/action-setup@v4.1.0
               with:
-                  run_install: true
+                  run_install: |
+                    - args: [--ignore-scripts]
```
Klaus-Keller, line 27: @donal-tobin-sap, what happens if any module or transient dependency has any install script that is required?

donal-tobin-sap: This is a first install with pnpm and it keeps giving native installer errors.
It is probably not needed to install with pnpm.
The install with scripts is done with the frozen lock file in the later step.
(yes teh install is done twice)

donal-tobin-sap: (the native installer errors are because macos14 and macos15/latest are now darwin arm64 and a number of native binaries are missing for this architecture resulting in attempts to natively compile which willl not work because the mac compilation support does not exist in the images)
MacOS is the fastest and fails first so the success of the other platforms is not seen, hense the fail fast entry that I had to add to determine the issues/ also commented out are the 3 masos architectures.

Klaus-Keller (approved):

donal-tobin-sap:
Security fix with 3 reviews

Issue #3422, opened 2025-07-07
BUG - @sap-ux/preview-middleware - Not resolving i18n texts for manifest.json


pwasem:
### Description

The ` "@sap-ux/preview-middleware": "^0.20.51",` does not resolve i18n texts for manifest.json.
In particular `sap.app/title` and `sap.app/description` are not being resolved for tile in local flp sandbox.

### Steps to Reproduce

1. **manifest.json**:

```json
{
  "sap.app": {
    "i18n": "i18n/i18n.properties",
    "title": "{{App.Title}}",
    "description": "{{App.Description}}",
    "subTitle": "{{App.SubTitle}}"
  }
}
```

1. **18n/i18n.properties** / **webapp/i18n/i18n_en.properties**

```text
# This is the resource bundle for files

#Texts for manifest.json

App.Title=Games
App.SubTitle=Manage
App.Description=Manage your games
```

**ui5.yaml**

```yaml
specVersion: "4.0"
metadata:
  name: some-app
  allowSapInternal: true
type: application
framework:
  name: SAPUI5
  version: 1.136.2
server:
  customMiddleware:
    - name: preview-middleware
      afterMiddleware: compression
      configuration:
        debug: true
        flp:
          theme: sap_horizon_dark
          intent:
            object: Games
            action: manage

```

```shell
debug preview-middleware Configured apps: {"Some-app":{"title":"{{App.Title}}","description":"{{App.Description}}","additionalInformation":"SAPUI5.Component=some.app","applicationType":"URL","url":".."}}
```

### Expected results

Resolve i18n textes for manifest.json

### Actual results

Texts not being resolved.

### Screenshots

<img width="1820" height="1191" alt="Image" src="https://github.com/user-attachments/assets/4c51a9d7-418d-45a8-8f14-d63ad3d139e3" />

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

pwasemsap:
Root cause is the invalid regexp 

https://github.com/SAP/open-ux-tools/blob/19dadc2c9d84ae98638c8222439adb30376ddf1e/packages/preview-middleware/src/base/config.ts#L333

[...]

PR #3470, opened 2025-07-24
fix(preview-middleware): regex for resolving i18n

heimwege:
fix for #3422.

Credits to @pwasem for reporting the bug as well as the adjustment needed to fix it 👍🏻 I had to slightly adjust the suggestion to avoid a `polynomial regular expression used on uncontrolled data` issue.

### DIFF
--- a/.github/workflows/ui5-maintenance-versions.yml
+++ b/.github/workflows/ui5-maintenance-versions.yml
@@ -43,8 +43,11 @@ jobs:
             - name: List previously cached UI5 maintenance versions
               id: previous-ui5-versions
               run: |
-                cat tests/integration/adaptation-editor/versions.json
-                echo "versions=$(cat tests/integration/adaptation-editor/versions.json)" >> $GITHUB_OUTPUT
+                if [[ -f tests/integration/adaptation-editor/versions.json ]]; then
+                  cat tests/integration/adaptation-editor/versions.json
+                else
+                  echo "No previous versions found."
+                fi
             - name: Load UI5 maintenance versions
               id: current-ui5-versions
               run: |
--- a/packages/abap-deploy-config-inquirer/CHANGELOG.md
+++ b/packages/abap-deploy-config-inquirer/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/abap-deploy-config-inquirer
 
+## 1.6.0
+
+### Minor Changes
+
+-   6e2c5aa: fix overwrite prompt conflict
+
 ## 1.5.2
 
 ### Patch Changes
--- a/packages/abap-deploy-config-inquirer/package.json
+++ b/packages/abap-deploy-config-inquirer/package.json
@@ -6,7 +6,7 @@
         "url": "https://github.com/SAP/open-ux-tools.git",
         "directory": "packages/abap-deploy-config-inquirer"
     },
-    "version": "1.5.2",
+    "version": "1.6.0",
     "license": "Apache-2.0",
     "main": "dist/index.js",
     "scripts": {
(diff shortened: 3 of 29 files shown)


packages/preview-middleware/src/base/config.ts
```diff
@@ -329,8 +329,8 @@ async function getI18nTextFromProperty(
     propertyValue: string | undefined,
     logger: Logger
 ): Promise<string | undefined> {
-    //i18n model format could be {{key}} or {i18n>key}
-    if (!projectRoot || !propertyValue || propertyValue.search(/{{\w+}}|{i18n>\w+}/g) === -1) {
+    //i18n model format could be {{key}} or {i18n>key}; limit max length of key to 100 characters to avoid performance issues
```
Klaus-Keller: @heimwege, what happens if i18n key is longer than 100 characters?
I've just checked some apps and I've found keys with 60 characters, so I assume there are even longer keys.

heimwege: I adjusted the code so that the limit is no longer needed.

heimwege:
@marufrasully is there any specific reason the `extractI18nKey` function so far was not able to handle double curly brackets input? Would this adjustment be fine from you pov?

nlunets (approved):
Code looks like code
Changeset exists
Tests looks decent

packages/i18n/src/utils/key.ts
```diff
@@ -8,9 +8,14 @@ import type { I18nBundle, I18nEntry } from './../types';
...
  */
 export function extractI18nKey(input: string, key = 'i18n'): string {
-    input = input.trim();
-    const regPattern = new RegExp(`^({@?${key}(>|&gt;))`, 'g');
-    return input.replace(regPattern, '').replace(/\}$/gm, '').trim();
+    let sanitizedInput = input.trim();
+
+    if (doesDoubleCurlyBracketsExist(sanitizedInput)) {
+        sanitizedInput = '{' + (extractDoubleCurlyBracketsKey(sanitizedInput) ?? '') + '}';
+    }
```
voicis: Currently this function was only used for UI5 binding expressions and I don't think we should mix these two use cases, because they are context dependent. In some cases only one of the syntax variants are valid and we should check that.

packages/preview-middleware/src/base/config.ts
```diff
@@ -329,11 +330,10 @@ async function getI18nTextFromProperty(
     propertyValue: string | undefined,
     logger: Logger
 ): Promise<string | undefined> {
-    //i18n model format could be {{key}} or {i18n>key}
-    if (!projectRoot || !propertyValue || propertyValue.search(/{{\w+}}|{i18n>\w+}/g) === -1) {
+    if (!projectRoot || !propertyValue?.startsWith('{')) {
         return propertyValue;
     }
-    const propertyI18nKey = propertyValue.replace(/i18n>|[{}]/g, '');
+    const propertyI18nKey = extractI18nKey(propertyValue);
```
voicis: AFAIK the values in the manifest `sap.app` part does not use UI5 models and should not have `i18n>` prefix, so we could just use `extractDoubleCurlyBracketsKey` here.

voicis (changes requested):

packages/preview-middleware/src/base/config.ts
heimwege: That was a good find. The [ui5 manifest schema](https://github.com/SAP/ui5-manifest/blob/main/schema.json#L2904) states: `"^\\{\\{[\\w][\\w\\.\\-]*\\}\\}$"`

_Represents a title (mandatory); to make this property language dependent (recommended), use a key in double curly brackets '{{key}}'_

So I'm afraid you are right @voicis 😄👍🏻 I'll re-work this.

packages/i18n/src/utils/key.ts
heimwege: adjusted in 646342fbc725c9798cb5449de8132d4a6a7f8a37

packages/preview-middleware/src/base/config.ts
heimwege: adjusted in https://github.com/SAP/open-ux-tools/commit/646342fbc725c9798cb5449de8132d4a6a7f8a37

marufrasully (approved):
@heimwege thanks.
* changes looks good
* secured with unit test
* did not test it locally

voicis (approved):
- Code changes looks good
- Review comments addressed
- Did not test manually

PR #3474, opened 2025-07-24
fix(fiori-gen): updates FioriAppGeneratorPromptSettings type

cianmSAP:
Removes specific key reference from FioriAppGeneratorPromptSettings type and makes necessary updates in Fiori sub gen

This is so the `FioriAppGeneratorPromptSettings` has the prompt options at the appropriate level, and not with UI5 app prompt options and odata service inquirer at the root level


IainSAP (approved):
Thanks @cianmSAP 

- Aligns prompt settings across adaptors and extensions

Issue #3472, opened 2025-07-24
TBI - `axios-extension`: Improve performance of odata v4 `fetchServices` when large paged datasets


IainSAP:
### Description (include screenshots)

Current implementation uses the `@nextlink` , waiting for each page before using the next page link. Instead the page requests can be calculated based on the total record count (`ServiceGroups`) returned by the first request. All remaining pages can then be requested in parallel.

### Value
Significant improvement in the overall services loading time.

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

PR #3478, opened 2025-07-25
tbi(axios-extension): Improves v4 services load times

IainSAP:
Addresses : #3472 

- Adds calculated paging approach for v4 service group requests, in addition to serial `@nextlink` approach
- New implementation of on by defalt but previous implementation remains available as an option
- Updates tests to cover
- Adds performance logging to `odata-service-inquirer`


cianmSAP (approved):
Updates are clear
test added to cover

Issue #3479, opened 2025-07-25
FEATURE - Store fragment parent control info in fragment body


IvoSG:
As  a developer

when I create each fragment

I want to have the data for its parent control type and target aggregation stored in the fragment`s xml body as commented code

so that I can relate to it further in time

### Description

 Information for the parent control type and target aggregation of a given fragment should be stored in its own xml so the developer can have it as information there, as well as it can be used further by AI capabilities.

### Out of Scope 
adding AI capabilities to it

### Acceptance Criteria
Parent control type, view name and target aggregation are now stored in the fragments xml body.

### Tasks
- [x] Unit Tests added
- [ ] Test Cases defined

PR #3482, opened 2025-07-28
feat: Store fragment parent control info in fragment body

IvoSG:
Feat for [#3479.](https://github.com/SAP/open-ux-tools/issues/3479)
- Store targetAggregation and parent controlType in the fragment body


packages/adp-tooling/templates/rta/fragment.xml
```diff
@@ -1,4 +1,7 @@
 <!-- Use stable and unique IDs!-->
-<core:FragmentDefinition xmlns:core='sap.ui.core' xmlns='sap.m'>
+<core:FragmentDefinition xmlns:core='sap.ui.core' xmlns='sap.m'><% if (targetAggregation && controlType) { %>
+    <!-- targetAggregation: <%= targetAggregation %> -->
+    <!-- controlType: <%= controlType %> --> <% } %>
```
voicis: It is a minor thing, but shouldn't control type be before aggregation, because aggregation depends on the control type. The second one should be more readable.
```
targetAggregation: content
controlType: Toolbar
```

```
controlType: Toolbar
targetAggregation: content
```

packages/adp-tooling/test/unit/preview/descriptor-change-handler.test.ts
```diff
@@ -166,27 +165,51 @@ describe('change-handler', () => {
 
         it('should create the XML fragment and log information if it does not exist', () => {
             mockFs.exists.mockReturnValue(false);
+            mockFs.read.mockReturnValue(`
```
voicis: Shouldn't we actually use the content from the file? Now it doesn't even match and in the future you can introduce unexpected change in the template file and tests would still pass.

packages/adp-tooling/templates/rta/fragment.xml
IvoSG: Fixed in [231e6](https://github.com/SAP/open-ux-tools/pull/3482/commits/2c85b742353ddc6ae20d94826f8b32ed3fa231e6)

packages/adp-tooling/test/unit/preview/descriptor-change-handler.test.ts
IvoSG: Fixed in [231e6](https://github.com/SAP/open-ux-tools/pull/3482/commits/2c85b742353ddc6ae20d94826f8b32ed3fa231e6)

mmilko01 (dismissed):
Changeset OK
Excellent test coverage
Changes look good
Did not test manually

avasilev-sap (commented):
lgtm

avasilev-sap (dismissed):

voicis (approved):
- Code changes looks good
- Review comments addressed
- Good test coverage
- Did not test manually

mmilko01 (approved):
Re-approving

avasilev-sap (approved):

PR #3492, opened 2025-07-29
fix(deploy-config): fix conflict of overwrite prompt name

cianmSAP:
Internal high bug 35240

Reverts change made here https://github.com/SAP/open-ux-tools/pull/3444/files#diff-24301133b22fc485c7fb65d11b3e76c2150400123d73802940fe378e8b378ebdR13

`overwrite` prompt causing conflict as there is an existing prompt in ABAP deploy config inquirer with same name

Added a note to help prevent such case in the future

### DIFF
--- a/packages/abap-deploy-config-inquirer/src/types.ts
+++ b/packages/abap-deploy-config-inquirer/src/types.ts
@@ -51,6 +51,9 @@ export interface AbapSystemChoice {
 
 /**
  * Enumeration of prompt names used by
+ *
+ * N.B. as these prompts are merged with CF prompts (see `promptNames` in packages/cf-deploy-config-inquirer/src/types.ts),
+ * ensure that the names are unique across both files.
  */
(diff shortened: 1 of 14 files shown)


packages/abap-deploy-config-inquirer/src/types.ts
```diff
@@ -51,6 +51,9 @@ export interface AbapSystemChoice {
 
 /**
  * Enumeration of prompt names used by
+ *
+ * N.B. as these prompts are merged with CF prompts (see `promptNames` in packages/cf-deploy-config-inquirer/src/types.ts),
+ * ensure that the names are unique across both files.
```
IainSAP, line 56: Should the overwrite prompt be renamed here also for clarity and future proofing? `overwriteAbap...`

cianmSAP: Done

PR #3493, opened 2025-07-30
ci: do not crash on version cache miss

voicis:
Fixes workflow in case there is no versions cached.


devinea (approved):
Changes look appropriate.

Issue #3496, opened 2025-07-31
BUG - Pass the same prompt options when running deployment generator as a step or standalone


nikmace:
### Description
Many prompt validations do not work when running the deployment generator as a step in the ADP generator, unlike when running the generator as a standalone. Also, some cloud-related prompts in the OnPremise case are shown because of that.

### Steps to Reproduce
1. Start by creating a new ADP project in the ADP generator
2. Select the Add Deploy Config step and click next
3. See differences in validations, etc

### Expected results
Deployment generator prompt validations work the same way when running the deployment gen as a step or standalone, only for the ADP scenario.

### Actual results
Validations happen only when running the deployment generator as standalone.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3503, opened 2025-08-01
fix: Pass the same prompt options when running deployment generator as a step or standalone

nikmace:
Fix for #3496.
- Passes correct prompt options when `add deploy config` generator is called as a step in the ADP generator, that way the behavior is similar to when the `add deploy config` generator is run as a standalone.


mmilko01 (approved):
Changeset OK
Changes look good
Excellent test coverage
Did not test manually

testojs (approved):
* code looks good
* changeset OK
* good test coverage
* did not test manually

PR #3506, opened 2025-08-01
add an alert red color uidefaultbutton

eouin:
Add a red color for UIDefaultButton, property is called `tiertary`

7391714a-f1ed-4f22-bc3e-449bf2fc0013


Klaus-Keller (approved):
Thanks @eouin,

* new button type `alert` looks good
* changeset exists
* test case and storybook entry added
* tried it locally in storybook

Approved from my side

devinea (approved):

PR #3507, opened 2025-08-01
fix changesets

voicis:
Revert unintended major version bump.

### DIFF
--- a/.changeset/cool-parents-search.md
+++ b/.changeset/cool-parents-search.md
@@ -1,7 +1,8 @@
 ---
-'@sap-ux-private/control-property-editor-common': major
-'@sap-ux-private/preview-middleware-client': major
-'@sap-ux/control-property-editor': major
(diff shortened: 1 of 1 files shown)


packages/fe-fpm-writer/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/fe-fpm-writer",
     "description": "SAP Fiori elements flexible programming model writer",
-    "version": "0.35.15",
+    "version": "0.35.14",
```
devinea: `0.35.15` is already on npmjs https://www.npmjs.com/package/@sap-ux/fe-fpm-writer

voicis: ok, we'll have to do a partial revert only for the non published packages. It should still be possible as the only public package failed to publish.

PR #3509, opened 2025-08-01
feat(ui-prompting): Support index-based default values for 'list' type questions.

815are:
Currently in questions we set default value like -> https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/building-block/prompts/questions/table.ts#L83
where default is actual value.
Request was that we should allow to pass index based value for `list` type question, like:
```
return {
        type: 'list',
        name: 'aggregationPath',
        defaultIndex: 1,
        choices: project
}
```

### DIFF
--- a/.changeset/strong-adults-mix.md
+++ b/.changeset/strong-adults-mix.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/ui-prompting': patch
+---
+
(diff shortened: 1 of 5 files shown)


devinea (approved):
Adding defaultIndex is clear and covered by tests
changeset ✅

PR #3515, opened 2025-08-05
fix(telemetry): unstable test

IainSAP:
Fixes unstable telem test

- Replace `node -p process.versions` with `node -v` as only the node version was extracted
- Return a valid string `unknown` otherwise
- Replace custom command runnner with `CommandRunner`

### DIFF
--- a/packages/create/CHANGELOG.md
+++ b/packages/create/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/create
 
+## 0.13.93
+
+### Patch Changes
+
+-   Updated dependencies [ad49c30]
+    -   @sap-ux/preview-middleware@0.21.2
+    -   @sap-ux/app-config-writer@0.6.29
+
 ## 0.13.92
 
 ### Patch Changes
--- a/packages/create/package.json
+++ b/packages/create/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/create",
     "description": "SAP Fiori tools module to add or remove features",
-    "version": "0.13.92",
+    "version": "0.13.93",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/preview-middleware-client/CHANGELOG.md
+++ b/packages/preview-middleware-client/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux-private/preview-middleware-client
 
+## 0.15.3
+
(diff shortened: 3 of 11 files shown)


devinea (dismissed):
Thanks @IainSAP refactoring looks good.

packages/telemetry/src/tooling-telemetry/data-processor.ts
```diff
@@ -14,17 +15,15 @@ import yaml from 'yaml';
  * packages consuming telemetry need to add dependencies manually (for ex: sap/cds)
  * Further, size of generated .vsix will be increased by esbuild
  */
-import { getAppProgrammingLanguage, getProjectType, getAppType } from '@sap-ux/project-access/dist/project/info';
-import { findProjectRoot } from '@sap-ux/project-access/dist/project/search';
+import { getWebappPath } from '@sap-ux/project-access';
 import { isCapJavaProject } from '@sap-ux/project-access/dist/project/cap';
+import { getAppProgrammingLanguage, getAppType, getProjectType } from '@sap-ux/project-access/dist/project/info';
```
heimwege: Is there a reason to use a deep/subpath import instead of importing from `@sap-ux/project-access`? That's actually an anti-pattern. I know you did not introduce it  but I saw it and I am a good boy scout and want to leave the code a better place than how I found it 😄
So If the exports on root level of project access are missing we can do this in a separate PR.

heimwege (commented):
First of all thanks for fixing this annoying issue 🥳 
I only have one comment that is also not really related to this PR.

packages/telemetry/src/tooling-telemetry/data-processor.ts
IainSAP: Theres a comment relating to this although Im not sure if this makes any practical difference. https://github.com/SAP/open-ux-tools/blob/main/packages/telemetry/src/tooling-telemetry/data-processor.ts#L8-L16

IainSAP: I have removed these deep imports and verified a couple of bundling cases that there is no increase in bundle size as a result.

heimwege: Oh that sounds interesting. I wasn't aware of that being possible. At least it does not seem to help in this case 😸

devinea (approved):
Thanks @IainSAP 
changes look good.

PR #3516, opened 2025-08-06
fix: we improved the support for ui5 legacy free

nlunets:
This changes tweak a bit the way we load third parties to ensure that legacy free loading behaves properly.
On top of it it adds an option to shim manifest which might be needed in some scenarios.


packages/jest-environment-ui5/src/env/ui5loader.js
```diff
@@ -1766,11 +1764,16 @@
 					}
                     const output = oModule.data.call(__global);
 					if(output && mShims[oModule.name] && mShims[oModule.name].exports && Object.keys(output).length > 0 && !global[mShims[oModule.name].exports]) {
-						// if(global[mShims[oModule.name].exports] ) {
-						// 	global[mShims[oModule.name].exports] = {...global[mShims[oModule.name].exports], ...output}
-						// } else {
+						if(global[mShims[oModule.name].exports] ) {
+							global[mShims[oModule.name].exports] = {...global[mShims[oModule.name].exports], ...output}
```
voicis, line 1768: This looks like unreachable code, because the condition `!global[mShims[oModule.name].exports]` in line 1767 is true, then `global[mShims[oModule.name].exports]` in the following line can't be true.

nlunets: true :)

nlunets: fixed

voicis (approved):
- Changes looks good
- Review comments addressed
- Did not test manually

heimwege (approved):
- Changes ok
- Test Coverage ok
- Code changes ok (it's plain JS, can't get a better rating than "ok" 😄)
- Did not test manually

PR #3523, opened 2025-08-11
Feat/set default aggregation path when macro page

kjose90:
Linked open source PR https://github.com/SAP/open-ux-tools/pull/3509
https://github.com/SAP/open-ux-tools/pull/3520

- Introduced support for a [checked](https://github.com/SAP/open-ux-tools/pull/3523/files#diff-b76a61a01c9bdd992807705a419c87838dc5950f0bb47153b4c28fba743b6caeR30) value on choices used by the Select component.
- When a choice in the choices array has checked: true, the Select component will automatically preselect that option as the default value.
- This enables marking a default option directly in the choices array
- Sets [default AggregationPath prompt](https://github.com/SAP/open-ux-tools/pull/3523/files#diff-cca1db6d02995dd35490e22d87425fa5113ebea58607454f7f62095e2a302036R282) to macro page in fe-fpm-writers  if macro Page is present in choices

### DIFF
--- a/packages/fe-fpm-writer/src/building-block/prompts/utils/questions.ts
+++ b/packages/fe-fpm-writer/src/building-block/prompts/utils/questions.ts
@@ -245,7 +245,7 @@ export function getAggregationPathPrompt(
                           join(appPath, viewOrFragmentPath),
                           fs
                       );
-                      const key = `/mvc:View/${pageMacroDefinition}`;
+                      const key = Object.keys(inputChoices).find((k) => k.endsWith(`/${pageMacroDefinition}`));
                       const choices = transformChoices(inputChoices, false, key);
                       if (!choices.length) {
                           throw new Error('Failed while fetching the aggregation path.');
--- a/packages/fe-fpm-writer/src/building-block/prompts/utils/questions.ts
+++ b/packages/fe-fpm-writer/src/building-block/prompts/utils/questions.ts
@@ -241,10 +241,11 @@ export function getAggregationPathPrompt(
             ? (answers: Answers) => {
                   const { viewOrFragmentPath } = answers;
                   if (viewOrFragmentPath) {
-                      const choices = transformChoices(
-                          getXPathStringsForXmlFile(join(appPath, viewOrFragmentPath), fs),
-                          false
+                      const { inputChoices, pageMacroDefinition } = getXPathStringsForXmlFile(
+                          join(appPath, viewOrFragmentPath),
+                          fs
                       );
+                      const choices = transformChoices(inputChoices, false, pageMacroDefinition);
                       if (!choices.length) {
                           throw new Error('Failed while fetching the aggregation path.');
                       }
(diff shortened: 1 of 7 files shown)


packages/fe-fpm-writer/src/building-block/prompts/utils/questions.ts
```diff
@@ -266,12 +267,23 @@ export function getAggregationPathPrompt(
...
  * @param obj - object to be converted to choices
  * @param sort - apply alphabetical sort(default is "true")
+ * @param pageMacroDefinition - the page macro definition to check against
  * @returns the list of choices.
  */
-export function transformChoices(obj: Record<string, string> | string[], sort = true): PromptListChoices {
+export function transformChoices(
+    obj: Record<string, string> | string[],
+    sort = true,
+    pageMacroDefinition?: string
```
815are: please use different naming for parameter. `transformChoices` method is used for other questions like `entitySets`, `qualifier`, `filterBarId`.
I think parameter name could be `defaultKey` and check should be `key === defaultKey` instead of `key.endsWith(`/${pageMacroDefinition}`)`. I think you can find precise key after calling `getXPathStringsForXmlFile`, but before calling `transformChoices`

packages/fe-fpm-writer/src/building-block/prompts/utils/questions.ts
```diff
@@ -266,12 +267,23 @@ export function getAggregationPathPrompt(
...
+): PromptListChoices {
     let choices: PromptListChoices = [];
     if (!Array.isArray(obj)) {
-        choices = Object.entries(obj).map(([key, value]) => ({ name: key, value }));
+        choices = Object.entries(obj).map(([key, value]) => {
+            // Add checked if value matches Page macro choice definition example: `/mvc:View/macro:Page/`
+            if (key.endsWith(`/${pageMacroDefinition}`)) {
```
815are: see comment above

packages/ui-prompting/src/components/Inputs/Select/Select.tsx
```diff
@@ -26,11 +26,15 @@ export const Select = (props: SelectProps) => {
             return options[0].data?.value;
         }
 
-        // Handle numeric default that isn't a key = it could be an index
-        if (props.defaultIndex !== undefined && options[props.defaultIndex]) {
-            return options[props.defaultIndex].data?.value;
+        // Use the first checked option as default
+        const checkedOption = options.find((opt) => opt.data && 'checked' in opt.data && (opt.data as any).checked);
```
815are: could you please try to avoid `any`?

packages/fe-fpm-writer/src/building-block/prompts/utils/questions.ts
kjose90: [agreed](https://github.com/SAP/open-ux-tools/pull/3523/files#diff-cca1db6d02995dd35490e22d87425fa5113ebea58607454f7f62095e2a302036R277)

packages/fe-fpm-writer/src/building-block/prompts/utils/questions.ts
kjose90: passed like [this](https://github.com/SAP/open-ux-tools/pull/3523/files#diff-cca1db6d02995dd35490e22d87425fa5113ebea58607454f7f62095e2a302036R248) instead

packages/ui-prompting/src/components/Inputs/Select/Select.tsx
kjose90: https://github.com/SAP/open-ux-tools/pull/3523/files#diff-b76a61a01c9bdd992807705a419c87838dc5950f0bb47153b4c28fba743b6caeR30

Issue #3524, opened 2025-08-11
BUG - S4 - LROP v4 - Absolute paths used in HTML Test files error causing failure on the S4 pipeline


fionan-SAP:
### Description
After completion of the s4 tests and the pull requests to the S4 Pipeline, s4-lrop-v4-odata is failing on the S4 pipeline with the following error

```
[ERROR] 
[ERROR] Quality issue: HTML_ABSOLUTE_PATH_ERROR:Very High!
[ERROR] In file test/integration/opaTests.qunit.html (0:0)
[ERROR] Test :: Absolute path for SCRIPT/LINK tag is not allowed as it may prevent ui5 bootstrapping.
```

### Steps to Reproduce

1. After the s4 tests complete, click on the check the PRs from the link in the e2e-jenkins-failures-app-gen slack channel
2. PRs show failure for s4-lrop-v4-odata
3. Click on the details for the failure

### Expected results
- S4-pipeline PR passes

### Actual results
- S4-pipeline PR fails

### Screenshots
If applicable, add screenshots to help explain the problem.

<img width="3253" height="1310" alt="Image" src="https://github.com/user-attachments/assets/adbf91fc-9253-4993-b3d6-2216eb80e392" />

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3525, opened 2025-08-11
fix: use relative resource paths in UI5 OPA test HTML files

fionan-SAP:
Issue: #3524

Updating all UI5 OPA test runner HTML files (`opaTests.qunit.html`) to use relative resource paths instead of absolute paths. This resolves the `HTML_ABSOLUTE_PATH_ERROR` that prevented proper UI5 bootstrapping.


kjose90 (approved):
Changes looks good 
tests & changeset ✅

PR #3527, opened 2025-08-13
feat(ui5-test-writer): use self contained journeys for OPA tests

docirl:
Internal issue 34871

Currently the generator creates a central journey runner instance used for all opa5 journeys. With virtual endpoints it's much better to go for self contained journeys that bring their own journey runner because then the virtual endpoint can collect all journeys and start them one by one instead of needing one central instance that creates the journey runner and starts it.

- refactor OPA test templates to use self contained journeys
- update tests/snapshots


packages/ui5-test-writer/templates/v4/integration/FirstJourney.js
```diff
@@ -1,38 +1,37 @@
...
-            });
-        }
+        test("Teardown", function (Given, When, Then) { 
+            // Cleanup
+            Given.iTearDownMyApp();
+        });
     }
 
-    return Journey;
+    JourneyRunner.run({}, run)
```
heimwege: To me JourneyRunner.run({}, run) as @tobiasqueck named it looks odd. That's why I named the function `journey` in my sandbox (link) because `runner.run({}, [journey])` sound much more like what actually is being done.

According to the FE documentation [run](https://ui5.sap.com/#/api/sap.fe.test.JourneyRunner%23methods/run) expects an settings object and an array of Journeys.
The documentation is misleading/wrong because [this one](https://ui5.sap.com/#/api/sap.fe.test.JourneyRunner%23methods/sap.fe.test.JourneyRunner.run) states a list of journey params but links also to the one that requires an array of Journeys 🤷🏻

packages/ui5-test-writer/templates/v4/integration/opaTests.qunit.js
```diff
@@ -1,24 +1,9 @@
...
-            },
-            opaJourney.run
-        );
-    }
-);
\ No newline at end of file
+    function () {
+        "use strict";
+        
+        QUnit.start();
```
heimwege: I'm missing setting the QUnit autostart to false. It also doesn't seem to be done in the respective [html template](https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-test-writer/templates/v4/integration/opaTests.qunit.html) (See [preview middleware template](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/templates/test/qunit.js#L18)).

Not stopping the [QUnit autostart](https://qunitjs.com/api/config/autostart/) can lead to QUnit executing tests before they were actually loaded (that's also why it is done before the async require). So everything is green but only because no test was executed.

Also calling `QUnit.start()` w/o stopping the autostart does not make sense because it would start automatically anyway.

packages/ui5-test-writer/templates/v4/integration/FirstJourney.js
```diff
@@ -1,38 +1,37 @@
 sap.ui.define([
-    "sap/ui/test/opaQunit"
-], function (opaTest) {
+    "sap/ui/test/opaQunit",
+    "./pages/JourneyRunner"
+], function (test, JourneyRunner) {
```
heimwege: I'd keep the name `opaTest` for `sap/ui/test/opaQunit`.

```suggestion
], function (opaTest, JourneyRunner) {
```

packages/ui5-test-writer/templates/v4/integration/FirstJourney.js
heimwege: @BjoernDederichs could you please follow up on the FE documentation issue?

docirl: Thanks, it works passing a list of journeys but I'll check passing an array also (from your sample it looks to work also) but I can update to pass an array. I'll update the naming also.

packages/ui5-test-writer/templates/v4/integration/opaTests.qunit.js
docirl: Thanks, wasnt aware of the autostart logic for QUnit. I've updated to use this same config to set autostart false and require QUnit, from my local testing it is needed

packages/ui5-test-writer/templates/v4/integration/FirstJourney.js
docirl: I've updated to pass an array. I was testing with array and non array with different setup/importing of qunit and I was getting some inconsistent behaviour (e.g. like failing to load first time, loading on refresh). Would be good to see the fix documentation for `run` and what should be used.

Using the updated code below for autostart works consistently and as expected.

packages/ui5-test-writer/templates/v4/integration/FirstJourney.js
docirl: done

packages/fiori-app-sub-generator/test/int/fiori-elements/expected-output/alp_v4/webapp/test/integration/pages/JourneyRunner.js
```diff
@@ -0,0 +1,21 @@
+sap.ui.define([
+    "sap/fe/test/JourneyRunner",
+	"testNameSpace/alpv4/test/integration/pages/SalesOrderItemList",
+	"testNameSpace/alpv4/test/integration/pages/SalesOrderItemObjectPage",
+	"testNameSpace/alpv4/test/integration/pages/MaterialDetailsObjectPage"
+], function (JourneyRunner, SalesOrderItemList, SalesOrderItemObjectPage, MaterialDetailsObjectPage) {
+    'use strict';
+
+    var runner = new JourneyRunner({
+        launchUrl: sap.ui.require.toUrl('testNameSpace/alpv4') + '/index.html',
```
heimwege, line 10: I'm a bit surprised. Why are we referring to `/index.html` here? shouldn't we use the virtual or non-virtual endpoint at `test/flp.html`, `test/flpSandbox.html` or however it is called?

For this specific app (`testNameSpace.alpv4`) I would have expected [/test/flpSandbox.html](https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-app-sub-generator/test/int/fiori-elements/expected-output/alp_v4/webapp/test/flpSandbox.html)

That also matches the open-ux-tools-playground sample and [my own playground](https://github.com/heimwege/cap-node-fe-ts-sample/blob/main/app/samples/webapp/test/integration/pages/JourneyRunner.ts#L57)

[index.html](https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-app-sub-generator/test/int/fiori-elements/expected-output/alp_v4/webapp/index.html) does not bootstrap a flp sandbox. It just launches the app directly because this file can be deployed and used productively as well.

packages/ui5-test-writer/templates/v4/integration/opaTests.qunit.js
heimwege: According to the discussions in https://github.com/SAP/open-ux-tools/pull/1813, I suggest to use the `Qunit` from the `sap.ui.require` import, not via global for the `Qunit.start` (please ignore that it was a `sap.ui.define` in the linked PR. It has meanwhile been changed to `sap.ui.require`).

packages/fiori-app-sub-generator/test/int/fiori-elements/expected-output/alp_v4/webapp/test/integration/pages/JourneyRunner.js
docirl: This matches the previous expected output [file](https://github.com/SAP/open-ux-tools/pull/3527/files#diff-64bcd1dbdc65c74baa4d2bd1dd5c560fa1424970bcb2031c6e5ef2b4a3b1c98eL13), but I had a look into the code and this is controlled by [generateIndexHtml](https://github.com/SAP/open-ux-tools/blob/5529c63b6b9a5e1fc98909f2da82148b1bc9da7e/packages/fiori-app-sub-generator/src/fiori-app-generator/transforms.ts#L194) value in the app state.

`generateIndexHtml` is used to determine the html target [here](https://github.com/SAP/open-ux-tools/blob/06d3ba3d50987beaca924f2e64ef98302b90d7ca/packages/fiori-elements-writer/src/index.ts#L80), if it is true (which is default) then `index.html` is used. This is passed to the test writer then as `htmlTarget` used by the template.

This was the existing behaviour, is this not correct?

packages/ui5-test-writer/templates/v4/integration/opaTests.qunit.js
docirl: Is this not what I'm doing already? https://github.com/SAP/open-ux-tools/pull/3527/files#diff-0e602f7cfb4ecae0b74f1079da9310f6d1b1522c788522c146a7e0406b105a86R22

packages/fiori-app-sub-generator/test/int/fiori-elements/expected-output/alp_v4/webapp/test/integration/pages/JourneyRunner.js
heimwege: Thanks for the explanation. If it is not related to your changes that's good news (for this PR 😏)

Yet I think using `index.html` is wrong because you don't test in an flp. Should I open a new issue for that one to follow up?

packages/ui5-test-writer/templates/v4/integration/opaTests.qunit.js
heimwege: Oh sorry, 🙈 the inner `require` tricked me. But this leads me to the question: why are there two requires needed? Can't you just

```js
sap.ui.require(
  [
    "sap/ui/thirdparty/qunit-2",
    "sap/ui/qunit/qunit-junit",
    "sap/ui/qunit/qunit-coverage",
    "<%- appPath %>/test/integration/<%- opaJourneyFileName %>",
  ], function (QUnit) {
    "use strict";
    QUnit.start();
});
```

docirl: Thanks, yes this is better. I've updated with this change.

packages/fiori-app-sub-generator/test/int/fiori-elements/expected-output/alp_v4/webapp/test/integration/pages/JourneyRunner.js
docirl: Yes I think it might be better to keep that issue separate to this for tracking, there's already a lot of updates here

docirl:
@heimwege I've made updates based on the latest comments, can you re-review please?

packages/fiori-app-sub-generator/test/int/fiori-elements/expected-output/alp_v4/webapp/test/integration/pages/JourneyRunner.js
heimwege: opened an internal ticket to follow up on this one

packages/ui5-test-writer/templates/v4/integration/opaTests.qunit.js
heimwege: ...and I just realized that the preview middleware qunit template also has this weird nested require. I need to fix it in the template as well 🙈

heimwege:
## 1.138.x

👍🏻 

## 1.108.45

🙅🏻 

```
2025-08-20 09:53:28.378399 Started URL '../../index.html?sap-ui-log-level=ERROR&sap-ui-xx-viewCache=true' in iFrame - FAILURE
Failure in Opa success function
Exception thrown by the testcode:'TypeError: Cannot read properties of undefined (reading 'show')
TypeError: Cannot read properties of undefined (reading 'show')
    at Object.stubMessageToast (Stubs-dbg.js:50:69)
    at Object.stubAll (Stubs-dbg.js:20:9)
    at BaseArrangements-dbg.js:69:13
    at constructor.<anonymous> (OpaBuilder-dbg.js:132:17)
    at u.success (Opa5-dbg.js:799:24)
    at constructor.<anonymous> (Opa-dbg.js:579:24)
    at n (Opa-dbg.js:52:18)
    at _ (Opa-dbg.js:46:3)
    at Opa-dbg.js:101:4
    at t (http://localhost:8080/resources/sap/ui/test/autowaiter/_timeoutWaiter.js:6:1504)' - Opa sap.ui.test.Opa
```

Not exactly sure where this comes from.

## 1.96.42

🙅🏻 

image

(No Journey started.)

Did not try lower versions after that.

Issue might also exist for the virtual test endpoints. Was not yet able to check this.

docirl:
Thanks for testing @heimwege 

I'm looking into these errors and decided to check 1.108.45 with release version of the generator and I'm getting the same error on the test:

image


So while it's not related to this change, it's another issue with these tests. 

I'll have a look what I can find

docirl:
Follow up issue will be created to investigate test errors with lower UI5 versions

devinea (approved):
Template updates lg. 
coverage and changeset ✅

Issue #3528, opened 2025-08-13
TBI - Create a ulity method to determine if there is a mis-match between VSCode proxy settings and system proxy settings


longieirl:
### Description (include screenshots)
Customers may find it confusing that saved systems and the integrated VS Code terminal rely on VS Code proxy settings, while npm calls use the npm proxy configuration. In practice, npm inherits its proxy values from the HTTP_PROXY / HTTPS_PROXY environment variables, not from the VS Code proxy configuration.

Proposed Improvement
We should implement a check that compares the environment variables (HTTP_PROXY / HTTPS_PROXY) against the VS Code proxy settings. If they are misaligned, display a clear warning to the user to prevent connectivity issues.

### Value
Maintain consistency between VSCode and CLI flows.

### Architecture Elaboration
Does it requires architecture elaboration? No

### Tasks
- [ ] Create a new method that can be compare against the environment variables. 
- [ ] Append unit tests

longieirl:
For now, the check is not required but the implementation to validate this is completed and should be merged.

PR #3529, opened 2025-08-13
fix: add new method to compare proxy environment settings

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/3528


packages/nodejs-utils/src/proxyCheck.ts
```diff
@@ -0,0 +1,100 @@
+export enum ProxyValidationStatus {
```
IainSAP: I think we should avoid exported enums from modules going forward. Please consider converting to a const like so : https://github.com/SAP/open-ux-tools/blob/94b35206e123e79254348c5cf812266d96602a12/packages/store/src/entities/backend-system.ts#L4-L11 

This allows module consumers the choice of using a dependency or dev dependency where only the type is needed (its probably not really relevant in this case but I think should be adopted as general practice)

longieirl: Done. I've also created a feature to introduce a custom rule to block the export of enums.

https://github.com/SAP/open-ux-tools/issues/3686

IainSAP (approved):
Thanks @longieirl 

- Comments addressed
- Not tested manually

PR #3533, opened 2025-08-15
fix: resolve path issue in annotationFile creation

Jimmy-Joseph19:
corrected path parameter to use project path instead of web app path as getVariant now calculates webapp path of the project


voicis (approved):
- Changes looks good
- Review comments addressed
- Tested manually

nikmace (approved):
Changes look good
Changeset present
Did not test manually

(One test file is not formatted, though)

PR #3522, opened 2025-08-08
updates writers to enable cds plugin and use workspaces

cianmSAP:
Internal issue 29990

- Deprecates writer option `enableCdsUi5Plugin` as if `enableNPMWorkspaces` it will also be added by default
- Defaults `enableNPMWorkspaces` in the writer 
- Updates any existing cds watch scripts to use the new format with the app id
- Updates ui5 app inquirer - removes checks for min cds version, removes additional message for typescript prompt

PR #3534, opened 2025-08-15
fix(ui5-app-inquirer): fix typescript and virtual endpoint prompts for cap projects

cianmSAP:
Correction required from https://github.com/SAP/open-ux-tools/pull/3522

Typescript and Virtual endpoint prompts should not be shown for CAP JAVA. This is acheived by checking the mid cds version - as described in the updated comment in `packages/fiori-app-sub-generator/src/utils/common.ts`

### DIFF
--- a/packages/ui5-application-inquirer/src/prompts/index.ts
+++ b/packages/ui5-application-inquirer/src/prompts/index.ts
@@ -72,14 +72,14 @@ export async function getQuestions(
         [promptNames.description]: getDescriptionPrompt(),
         [promptNames.targetFolder]: getTargetFolderPrompt(targetDir, shouldValidateFioriAppFolder),
         [promptNames.ui5Version]: getUI5VersionPrompt(ui5Versions, promptOptions?.ui5Version),
-        [promptNames.enableTypeScript]: getEnableTypeScriptPrompt(),
+        [promptNames.enableTypeScript]: getEnableTypeScriptPrompt(capCdsInfo),
         [promptNames.addDeployConfig]: getAddDeployConfigPrompt(
             targetDir,
             promptOptions?.addDeployConfig,
             isCapProject
         ),
         [promptNames.addFlpConfig]: getAddFlpConfigPrompt(promptOptions?.addFlpConfig),
-        [promptNames.enableVirtualEndpoints]: getEnableVirtualEndpoints(),
+        [promptNames.enableVirtualEndpoints]: getEnableVirtualEndpoints(capCdsInfo),
         [promptNames.showAdvanced]: getShowAdvancedPrompt(),
         [promptNames.ui5Theme]: getUI5ThemePrompt(),
         [promptNames.enableEslint]: getEnableEsLintPrompt(),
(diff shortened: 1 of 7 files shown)


longieirl (dismissed):
Changeset is good.
Appropiate tests to cover change.
Changes look good.

packages/ui5-application-inquirer/src/prompts/index.ts
```diff
@@ -72,14 +72,14 @@ export async function getQuestions(
         [promptNames.description]: getDescriptionPrompt(),
         [promptNames.targetFolder]: getTargetFolderPrompt(targetDir, shouldValidateFioriAppFolder),
         [promptNames.ui5Version]: getUI5VersionPrompt(ui5Versions, promptOptions?.ui5Version),
-        [promptNames.enableTypeScript]: getEnableTypeScriptPrompt(),
+        [promptNames.enableTypeScript]: getEnableTypeScriptPrompt(capCdsInfo),
```
IainSAP: since the `capCdsInfo` is known already when calling getPrompts() we dont need when clauses but instead can move that condition out of the prompt to help with the rendering performance. Small improvement over previous implementation.

```
if (capCdsInfo?.hasMinCdsVersion) {
    ...insert prompt into object at the correct position
}
```

cianmSAP: I've updated the `hidePrompts` function to add the logic there

packages/ui5-application-inquirer/src/prompts/prompt-helpers.ts
```diff
@@ -84,7 +86,16 @@ export function hidePrompts(
...
-                !([promptNames.targetFolder, promptNames.enableEslint].includes(promptNames[promptKey]) && isCapProject)
+                !(
+                    [promptNames.targetFolder, promptNames.enableEslint].includes(promptNames[promptKey]) &&
+                    isCapProject
+                ) &&
+                // `enableTypeScript` and `enableVirtualEndpoints` should not be shown for certain CAP projects i.e CAP Java
+                !(
+                    [promptNames.enableTypeScript, promptNames.enableVirtualEndpoints].includes(
+                        promptNames[promptKey]
+                    ) && (capCdsInfo ? !capCdsInfo?.hasMinCdsVersion : false)
```
IainSAP, line 97: Why are we checking `hasMinCdsVersion` should be checking for Java

cianmSAP: We don't pass the specific info whether it's node or Java to the inquirer
Also basing it off of your comment here https://github.com/SAP/open-ux-tools/blob/5bc021197a7868e4ca2405b1e7adabfa636caaa3/packages/fiori-app-sub-generator/src/utils/common.ts#L118

IainSAP: So it seems this is somehow how we check for cap java...hmm a little obtuse. And we also have this:
https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-app-sub-generator/src/fiori-app-generator/prompting.ts#L285

IainSAP (approved):
Thx @cianmSAP 

- reapproving

PR #3535, opened 2025-08-15
Fix/preview middleware/lrep connector

heimwege:
Follow-up to #3363

Error: Preview with UI5 1.76 from npmjs still not working because of a 404 for `ObjectStorageConnector`.

Solution: We need to use dynamic imports because `ObjectStorageConnector` is located at a different place in UI5 1.76 (and uses slightly different property names).

### DIFF
--- a/.changeset/plenty-foxes-cheat.md
+++ b/.changeset/plenty-foxes-cheat.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux-private/preview-middleware-client': patch
+---
+
+fix: error when loading ObjectStorageConnector from npmjs for UI5 version 1.76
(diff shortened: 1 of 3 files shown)


heimwege:
@voicis do you have an Idea why I get the `error TS1378: Top-level 'await' expressions are only allowed when the 'module' option is set to 'es2022', 'esnext', 'system', 'node16', 'node18', 'node20', 'nodenext', or 'preserve', and the 'target' option is set to 'es2017' or higher.` when running the tests?
I don't get a TS error in my IDE when writing to code. And `ES2022` is being used in [.tsconfig.json](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/tsconfig.json#L15) 🤷🏻 

Update 1: Seems like jest uses the root tsconfig that has `module` set to `commonjs`. But how to make it use the tsconfig from preview-middleware-client package?

Update 2: I'm giving up on that one and rather get rid of the top level await

heimwege:
Decision: Let's close this one in favor of #3557

PR #3537, opened 2025-08-18
Fix Select Prompt Default Value Handling

kjose90:
When a default value is preselected for a <Select /> prompt, the user is still required to change the option before the value is considered valid. If the user does not change the selection and submits the form, an alert incorrectly states that the input is required.

Added [logic](https://github.com/SAP/open-ux-tools/pull/3537/files#diff-96e63d642cf5734e3c17201bf821d69c22fc17bca987de29806182a701239ed7R96) -  to take updated answers into consideration when a default option is provided in the Select prompt.


815are (approved):
checked:
1. code changes are easy to read - avoids overwrite during conflicting change callback;
2. test is added with new scenario
3. changelog persists

devinea (approved):

PR #3538, opened 2025-08-18
fix(annotation-generator): annotations are not generated if navigation property can't be resolved

voicis:
There was a type mismatch to what `targetType` of a navigation property was resolved to by `@sap-ux/annotation-converter`. It was returning empty object instead of `EntityType`.


Jimmy-Joseph19 (approved):
Changes looks good.
Covered the scenario with unit test.
Changeset found.
Approving.

Did not test locally.

PR #3539, opened 2025-08-18
chore: update TS target to ES2023

heimwege:
- update TS `compilerOptions` `target` to ES2023
- delete target from packages where equal to root tsconfig
- adjust TS `compilerOptions` `lib` to ES2023


heimwege:
`yeomanTest` in @sap-ux/repo-app-import-sub-generator seem to be a bit flaky 🤷🏻

voicis:
I think we should also remove the target overrides in packages e.g. https://github.com/SAP/open-ux-tools/blob/2ea8282f24893878c2c3186866d4be31bef786d4/packages/cds-annotation-parser/tsconfig.json#L6-L7

heimwege:
> I think we should also remove the target overrides in packages e.g.
> 
> https://github.com/SAP/open-ux-tools/blob/2ea8282f24893878c2c3186866d4be31bef786d4/packages/cds-annotation-parser/tsconfig.json#L6-L7

That's a good idea 👍🏻 For the `lib` at least the arrays differ, so override might make sense but using the same target... I deleted them and noticed there are other equal properties as well. Might clean this up later in another PR.

devinea (dismissed):
Thanks for the update @heimwege

packages/cds-annotation-parser/src/parser/parser.ts
```diff
@@ -262,7 +262,7 @@ export class AnnotationParser extends CstParser {
      * https://chevrotain.io/docs/guide/performance.html#caching-arrays-of-alternatives
      */
     v?: IOrAlt<void>[];
-    private CST_STACK: AssignmentCstNode[];
+    private CST_STACK: AssignmentCstNode[] = [];
```
voicis, line 265: Starting with ES2022 [useDefineForClassFields](https://www.typescriptlang.org/tsconfig/#useDefineForClassFields) is set to `true` by default which emits this field in the class declaration to confirm with ECMAScript standard. In our case it caused problems because it was initialized by the superclass and we were setting it to undefined again afterwards.

heimwege:
@marufrasully or @vinayhospete could you maybe review for ux-tools-annotations?

PR #3540, opened 2025-08-18
fix(cds-annotation-converter): crash on empty flattened annotation term conversion

voicis:
Frequently during code completion annotation values are not complete, but we should still be able to process it without crashing. 

In this case we need gracefully handle case where there is no following segment for annotation term.


Jimmy-Joseph19 (approved):
Changes looks good.
Covered the scenario with unit test
Changeset found. 
Did not test locally
Approving

Issue #3542, opened 2025-08-18
BUG - We do not want to show in the info center an error when we select a component in the Outline tree which does not have related api.json..


avasilev-sap:
### Description

Whenever we select an UI5 component from the outline tree in the ADP Visual Editor which ui5 lib does not provide api.json, we got an error in the info center which creates a noise in the message log. The api.json is used to create the list of properties related to the component in the right panel. For the described set of components the list in the right panel is properly rendered as an empty one.

### Steps to Reproduce
1. Go to the outline tree filter and type EasyFilterBarContainer
2. Click on the corresponding tree node representing the EasyFilterBarContainer component.

### Expected results
We must not display in the info center the `Documentation error`. This error is shown because the api.json fetch responds with status 404 not found.

### Actual results
The error is displayed which is unwanted noice in the message log.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3543, opened 2025-08-18
fix: [ADP] For ui5 components which do not provide api.json we do not display Documentation error in the info center.

avasilev-sap:
#3542

When we select an UI5 component from the Components outline tree in the Visual Editor we create a list of properties for the object in the right panel based on the result of the documentation api call.

**Documentation api call:** We do fetch with url `/test-resources/<ui5 lib name with slashes>/designtime/api.json`.
For libraries which do not provide api.json file e.g. `sap.fe.controls.easyFilter`(EasyFilterBarContainer component) or `sap.fe.macros.Table` the api responds with 404 not found. In case the file is not found we do not display error in the info center. Any other http error during the api.json fetch will be displayed.


voicis (dismissed):
- Code changes looks good
- Good test coverage
- Tested manually

.changeset/chilly-deers-count.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux-private/preview-middleware-client': minor
```
nikmace, line 3: Here we are missing a @sap-ux/preview-middleware where it is consumed.

cc @voicis

nikmace (changes requested):
The rest looks OK

.changeset/chilly-deers-count.md
avasilev-sap: fixed in 104bfe243eb50cd1d91acab066e51ce80ef18855

nikmace (approved):
Changeset OK
Changes look good
Test coverage is good
Did not test manually

Issue #3546, opened 2025-08-20
BUG - [ADP][Info center] Remove all messages from the info center which overlaps with messages added with the ui5 Log library.


avasilev-sap:
### Description
Reduce the information populated in the info center.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3549, opened 2025-08-20
fix: Remove from info center messages which overlap with messages added with the Log library.

avasilev-sap:
#3546

Remove the described messages + update unit tests.

Testing done:
* Run staging job with ADP e2e tests.
* Manually verified.


voicis:
There still seems to be a case where info message is sent together with log.
https://github.com/SAP/open-ux-tools/blob/3a8eaa399f82c42087420064be9f02623754468a/packages/preview-middleware-client/src/cpe/documentation.ts#L173-L184
https://github.com/SAP/open-ux-tools/blob/3a8eaa399f82c42087420064be9f02623754468a/packages/preview-middleware-client/src/utils/version.ts#L55-L63

lfindlaysap (approved):

avasilev-sap:
Hi @voicis i synced with @hristotsolev those messages could stay as is, they could be seen on a rare occasions so we can make an exception for them.

Issue #3550, opened 2025-08-20
BUG - ADP Generator shows misleading additional message when system requires auth


nikmace:
### Description
An addiotional_info error message is shown in the case where the system requires authentication (we show the username and password prompt), and we have not yet fetched flexUISystem this error is shown:
```ts
The system that you have selected is not an ABAP On-Premise system. Adaptation projects are only supported on those systems. Please choose an ABAP On-Premise system which supports flexibility and DTA_FOLDER deployment.
```

### Steps to Reproduce
1. Open ADP Generator in BAS and select a system with NoAuthentication
2. See the error message on the first page, which misleads the user

### Expected results
Error is not shown.

### Actual results
Error is shown

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3551, opened 2025-08-20
fix: ADP Generator shows misleading additional message when system requires auth

nikmace:
Fix for #3550.
- Add a check that prevents additional info error message from showing.


IvoSG (approved):
Small and simple change
Code looks good
Adapted tests

avasilev-sap (approved):

PR #3552, opened 2025-08-21
fix: Restrict Page Building Block support to UI5 version > 1.136.0 during custom app generation

kjose90:
- Ensured the page building block feature is only enabled when the minimum UI5 version is `1.136.0 `or higher by setting `minUi5Version` to `1.136.0 `when the user selects `Add Page Building Block` option during entity selection for custom app generation [here](https://github.com/SAP/open-ux-tools/pull/3552/files#diff-46aa777b9463fc95164b52047582481d96483996adc92cf372f1149ced3066a1R227). This enforces filtering of the available UI5 versions in the project attribute prompts, allowing users to select only supported versions for page macros functionality.

- Added a tooltip and an additional warning message to the page building block prompt, informing users about the UI5 version restriction [here](https://github.com/SAP/open-ux-tools/pull/3552/files) .


packages/fiori-app-sub-generator/src/fiori-app-generator/prompting.ts
```diff
@@ -195,35 +207,80 @@ export async function promptOdataServiceAnswers(
...
+ * @param {boolean} addPageBuildingBlock - Whether the page building block feature is enabled.
+ * @returns {string} The minimum required UI5 version.
+ */
+function getMinUi5Version(
+    promptSettings: UI5ApplicationPromptOptions | undefined,
+    service: Partial<Service>,
+    floorplan: Floorplan,
+    addPageBuildingBlock: boolean = false
+): string {
+    if (addPageBuildingBlock && floorplan === FloorplanFE.FE_FPM) {
```
kjose90: Do I need to explicitly check for `FloorplanFE.FE_FPM` here? since we display this page building block only when [condition](https://github.com/SAP/open-ux-tools/blob/e1850a6631f7d0386315c02655911fafe10d270d/packages/fiori-app-sub-generator/src/fiori-app-generator/fioriAppGenerator.ts#L232) is met so the check is already done previously? 🤔

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -146,6 +146,8 @@
         },
         "pageBuildingBlock": {
             "message": "Add a Page Building Block to the Application?",
+            "tooltip": "Page Building Block is supported from SAPUI5 version 1.136.  By choosing this option, you will be restricting the minimum SAPUI5 versions available for the application to be 1.136 and above.",
```
```suggestion
            "tooltip": "The Page building block is supported from SAPUI5 version 1.136.  By choosing this option, you will be restricting the minimum SAPUI5 versions available for the application to be 1.136 and above.",
```

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -146,6 +146,8 @@
...
+            "warning": "Page Building Block is supported with SAPUI5 version 1.136 and above.",
```
```suggestion
            "warning": "The Page building block is supported with SAPUI5 version 1.136 and above.",
```

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -146,6 +146,8 @@
         },
         "pageBuildingBlock": {
             "message": "Add a Page Building Block to the Application?",
+            "tooltip": "The Page building block is supported from SAPUI5 version 1.136.  By choosing this option, you will be restricting the minimum SAPUI5 versions available for the application to be 1.136 and above.",
```
```suggestion
            "tooltip": "The Page building block is supported from SAPUI5 version 1.136. By choosing this option, you will be restricting the minimum SAPUI5 versions available for the application to be 1.136 and above.",
```

lfindlaysap (dismissed):

packages/fiori-app-sub-generator/src/fiori-app-generator/prompting.ts
devinea: I think it is clearer to keep it that it is FPM only related here. 
However it could be nice update `getMinSupportedUI5Version` instead of wrapping it with this new function `getMinUi5Version`
Not sure with the null coalescing before `getMinSupportedUI5Version` if it is sufficient to update it.

@IainSAP could you take a look?

cianmSAP: I've discussed with Kitty, we can set the `minUi5Version` prompt option directly, rather than adding a new `addPageBuildingBlock` option. It should simplify the code.

IainSAP: I agree, use the existing options, add a prompt option that accepts the ui5 version filter that already exists for ui5 info. This function is far too opinionated to this use case.

This can be set in this function rather than outside, in the intereset of keep ing the code nicely separated:

https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-app-sub-generator/src/fiori-app-generator/prompting.ts#L258

kjose90: Thank yous - will revert and ping once ready 😉

packages/odata-service-inquirer/src/prompts/edmx/questions.ts
```diff
@@ -212,7 +212,16 @@ function getPageBuildingBlockQuestions(): Question<PageBuildingBlockAnswers>[] {
         message: t('prompts.pageBuildingBlock.message'),
         default: false,
         guiOptions: {
-            breadcrumb: true
+            breadcrumb: true,
+            hint: t('prompts.pageBuildingBlock.tooltip')
+        },
+        additionalMessages: (answers: boolean) => {
```
IainSAP: Please rename, this is not `answers`, its the value of the input so `addPageBB` ?

packages/fiori-app-sub-generator/src/fiori-app-generator/fioriAppGenerator.ts
```diff
@@ -264,6 +265,14 @@ export class FioriAppGenerator extends Generator {
             }
             // get project information
             if (hasStep(this.fioriSteps, STEP_PROJECT_ATTRIBUTES)) {
+                let promptSettings = generatorOptions.promptSettings?.['@sap/generator-fiori'];
+                if (
+                    this.state.floorplan === FloorplanFE.FE_FPM &&
```
cianmSAP: Rather than adding more complexity here in the main generator code, can you move the logic into [createUI5ApplicationPromptOptions](https://github.com/SAP/open-ux-tools/blob/635102f462f746f5e4c729bcf701a0c3cb383d7b/packages/fiori-app-sub-generator/src/fiori-app-generator/prompting.ts#L212), thre is some handling their already for ui5 version. Would be good to keep the logic in one place

kjose90: so then I would require to pass additional parameter's to `promptUI5ApplicationAnswers` & `createUI5ApplicationPromptOptions` leading to my previous [commits](https://github.com/SAP/open-ux-tools/pull/3552/commits/09edbe8a651a06ddd1fd6b226d7e920455270fb1#diff-46aa777b9463fc95164b52047582481d96483996adc92cf372f1149ced3066a1R114) leading to more code and test changes changes like in [commit](https://github.com/SAP/open-ux-tools/pull/3552/commits/09edbe8a651a06ddd1fd6b226d7e920455270fb1#diff-46aa777b9463fc95164b52047582481d96483996adc92cf372f1149ced3066a1R277) 

I was thinking we only need to call `getFPMPromptSettings` when its  FPM project and the user has chosen to add page building blocks; in all other cases, the function call is unnecessary and the original prompt settings can be used directly? 🤷‍♀️

cianmSAP: It's not so much about the lines of code or passing params, but the main issue was adding a new option `addPageBuildingBlock` that was then only used to set the min ui5 version, when there is a prompt option to do this.

If you look at `createUI5ApplicationPromptOptions`, it's current use case aligns with what you want to do. It already accepts the floorplan as a property, it would be reasonable to also pass the `entityRelatedConfig` property to this function also. 

In `createUI5ApplicationPromptOptions` function we set the `minUI5Version`. See `getMinSupportedUI5Version` already checks the floorplan - logically your checks would be better suited here. 

As pointed out previously, currently if `promptSettings?.[ui5AppInquirerPromptNames.ui5Version]?.minUI5Version` is set (e.g by an adaptor or extension gen), it will take precedence over a ui5 version set by the template (or building block in this case), this could lead to bugs (I know this is not your code), it would be nice to compare these values and take the highest one.

IainSAP: I'd suggest we do not filter or block selection of ui5 versions. Theres also system versions to consider. Instead I would add a warning under the pbb prompt that warns about version compatibility and then if a non-compatible version is selected (or otherwise set) when calling the fpm writer it simply does not add that feature. That should be   contained within the fpm writer itself. ?

packages/fiori-app-sub-generator/src/fiori-app-generator/prompting.ts
```diff
@@ -195,35 +208,81 @@ export async function promptOdataServiceAnswers(
...
+ * @param service - The service configuration.
+ * @param floorplan - The selected floorplan type.
+ * @param entityRelatedConfig - Entity-related configuration.
+ * @returns {Promise<string>} The resolved minimum UI5 version.
+ */
+async function resolveMinUI5Version(
+    promptSettings: UI5ApplicationPromptOptions | undefined,
+    service: Partial<Service>,
+    floorplan: Floorplan,
+    entityRelatedConfig?: Partial<EntityRelatedAnswers>
```
IainSAP, line 225: same comment

packages/fiori-app-sub-generator/src/fiori-app-generator/prompting.ts
```diff
@@ -195,35 +208,81 @@ export async function promptOdataServiceAnswers(
...
+): Promise<string> {
+    let minUI5Version =
+        promptSettings?.[ui5AppInquirerPromptNames.ui5Version]?.minUI5Version ??
+        getMinSupportedUI5Version(service.version ?? OdataVersion.v2, floorplan);
+
+    if (floorplan === FloorplanFE.FE_FPM && entityRelatedConfig?.addPageBuildingBlock) {
```
IainSAP: Please move this block into `getMinSupportedUI5Version` as this is what this logic is also doing. Also we need to leave the existing prompt option override as some extensions dont set a valid version deliberately.

packages/fiori-app-sub-generator/test/unit/fiori-app-generator/fioriAppGenerator-lifecycle1.test.ts
```diff
@@ -406,90 +407,116 @@ describe('Test FioriAppGenerator', () => {
...
-                    Version: '1.0.0',
-                    Uri: 'testUri'
-                }
-            ],
-            source: DatasourceType.capProject
-        };
-        // Note: These are not real-world combinations of options and state, they are for testing as many branches and options as possible
-        (options.appWizard as AppWizardCache)!['$fiori-cache'] = {
-            service: mockCachedService
+    describe('Test FioriAppGenerator for FPM projects', () => {
```
IainSAP: Why was the existing test replaced? Can you please restore it and add new tests for your specific use case, it looks like you should create a new test file for your custom page related tests.

packages/fiori-app-sub-generator/test/unit/fiori-app-generator/fioriAppGenerator-lifecycle1.test.ts
```diff
@@ -406,90 +407,116 @@ describe('Test FioriAppGenerator', () => {
...
-        options.preselectedEntityName = 'TestPreSelectedEntity';
-        options.showLayoutPrompts = false;
-        // Skipping the project attributes step for this test
-        options.fioriSteps = FIORI_STEPS.filter((step) => step.key !== STEP_PROJECT_ATTRIBUTES);
 
-        const mockEntityRelatedAnswers: EntityRelatedAnswers = {
-            mainEntity: {
-                entitySetName: 'SEPMRA_C_PD_Product',
-                entitySetType: 'SEPMRA_C_PD_ProductType'
+        const cases = [
```
IainSAP: Can you move these to a new test file. These are not lifecycle tests.

packages/fiori-app-sub-generator/test/unit/fiori-app-generator/fioriAppGenerator-lifecycle1.test.ts
```diff
@@ -406,90 +407,116 @@ describe('Test FioriAppGenerator', () => {
...
+            {
+                title: 'without building block and minUI5Version not set',
+                promptSettings: { '@sap/generator-fiori': { ui5Version: { minUI5Version: undefined } } },
+                expectedPromptSettings: { ui5Version: { minUI5Version: undefined } },
+                addPageBuildingBlock: false
+            }
+        ];
+
+        test.each(cases)(
+            'prompts for entity and UI5 application answers for FE_FPM custom page $title',
```
IainSAP: FPM specific tests should be somewhere else rather than in the main lifecycle test file.

IainSAP (changes requested):
Thanks @kjose90 

- Updates are clear
- Tests should be moved somewhere else please. Also please restore the existing test.

kjose90:
Thanks @IainSAP removed those tests from `fioriAppGenerator-lifecycle1.test`

packages/fiori-app-sub-generator/test/unit/utils/common.test.ts
```diff
@@ -68,6 +69,11 @@ jest.mock('@sap-ux/fiori-generator-shared', () => ({
     generateAppGenInfo: jest.fn()
 }));
 
+jest.mock('semver', () => ({
```
IainSAP: Do we need to mock this?

IainSAP (approved):
Thanks @kjose90 

- All comments addressed
- Tests added to cover

kjose90:
- Agreed that the minUI5Version via prompt settings would take precedence over `getMinSupportedUI5Version` being called [here](https://github.com/SAP/open-ux-tools/pull/3552/files#diff-46aa777b9463fc95164b52047582481d96483996adc92cf372f1149ced3066a1R249)
- If page building block feature is enabled then simply return 1.136.0 for custom app
- So we dont require semver any more for this PR.

cianmSAP (approved):
Updates look good,
comments addressed 
Changeset and tests are appropriate

Issue #3553, opened 2025-08-21
BUG - ADP Generator shows project type even if request for system info is failing


nikmace:
### Description
An additional message is shown: Project type: OnPremise under the system prompt when switching to a different system that has authentication. The additional message is still shown because the previous version selection state is not cleared. We should not show this additional message at all because we have not logged in

### Steps to Reproduce
1. Go to the SAPUX staging BAS instance and choose the latest master build
2. Try to create an ADP project with the bundled ADP generator
3. Choose a system with no authentication and then choose a system with authentication
4. See that Project type: onPremise is shown

### Expected results
Project type: onPremise is not shown at all

### Actual results
Project type: onPremise is shown

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3554, opened 2025-08-21
fix: ADP Generator shows project type even if request for system info is failing

nikmace:
Fix for #3553.
- Clears state when switching systems to avoid misleading info messages in info center.


mmilko01 (approved):
Straight forward change
Excellent test coverage
Changeset OK
Did not test manually

Issue #3555, opened 2025-08-21
BUG - Deploy Config generator: Flashes of prompts and busy spinner when entering description


korotkovao:
### Related Feature
Feature request: _issue number_

### Description
Typing in Deployment Description in the Deployment Configuration step with the prompts' state as below results in constant re-rendering unnecessarily.

<img width="559" height="826" alt="Image" src="https://github.com/user-attachments/assets/5c6749fa-1cbb-4db3-97de-4732b36662cf" />

### Steps to Reproduce
1. Go to the Generator, select the Add Deployment Configuration radio button
2. In the Deployment Configuration step, fill in all the values as shown in the screenshot
3. Go back to the Deployment Description and start editing the text

### Expected results

No busy indicator for the Deployment Description, no flashing prompt

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3556, opened 2025-08-21
fix(abap-deploy-config-inquirer): don't validate other fields unless …

korotkovao:
…package changes

## #3555

YUI retriggers validation for all fields when there is a change.
Don't validate other fields unless the package changes.

### DIFF
--- a/packages/abap-deploy-config-inquirer/src/prompts/questions/config/package.ts
+++ b/packages/abap-deploy-config-inquirer/src/prompts/questions/config/package.ts
@@ -34,6 +34,7 @@ export function getPackagePrompts(
 ): Question<AbapDeployConfigAnswersInternal>[] {
     let packageInputChoiceValid: boolean | string | IValidationLink;
     let morePackageResultsMsg = '';
+    let packageName: string | undefined;
     PromptState.isYUI = isYUI;
 
     const questions: Question<AbapDeployConfigAnswersInternal>[] = [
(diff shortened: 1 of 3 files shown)


packages/abap-deploy-config-inquirer/src/prompts/questions/config/package.ts
```diff
@@ -128,17 +129,21 @@ export function getPackagePrompts(
...
-                    options.packageAutocomplete,
-                    options.ui5AbapRepo,
-                    options.backendTarget
-                );
+                if (input !== packageName) {
+                    // Autocomplete can the entire choice object as the answer, so we need to extract the value
+                    const pkgValue: string = (input as ListChoiceOptions)?.value
+                        ? (input as ListChoiceOptions).value
+                        : input;
+                    packageName = pkgValue;
```
IainSAP: Shouldnt the packageName only be assigned once the validation is true. Here we are assigning it regardless and then returning true on subsequent calls so it wont get validated.

korotkovao: Thank you, I have added the change

cianmSAP: I didn't like the look of th fix in the first place, but with this change, the fix no longer works. i.e it was ensuring the validation only occurred if the package changed, now it will start to revalidate every time again, or atleast until the validation returns true. This original fix was also not ideal as we need to re-validate if the ui5 abap repo name changes

cianmSAP: @korotkovao Does the bug occur only for `packageAutocomplete` prompt, or also  `packageManual`?

korotkovao: @cianmSAP It occurs with both `packageAutocomplete` prompt and `also packageManual`
When you then go back to edit the description (or the repo) field, you see the bug. I guess it's not very common to start editing the previous input.

packages/abap-deploy-config-inquirer/src/prompts/questions/config/package.ts
```diff
@@ -128,17 +129,33 @@ export function getPackagePrompts(
...
+                // Compare current answers with previous answers
+                if (prevAnswers != null) {
+                    const keys = Object.keys(answers) as (keyof AbapDeployConfigAnswersInternal)[];
+                    // Defensive: prevAnswers is defined, so prevAnswers[key] is safe
+                    const onlyDescriptionChanged =
+                        keys.every((key) => key === 'description' || prevAnswers![key] === answers[key]) &&
+                        prevAnswers!.description !== answers.description;
+                    if (onlyDescriptionChanged) {
+                        prevAnswers = { ...answers };
+                        return true;
```
cianmSAP: We'd also need to keep track of the previous validation result for `packageAutocomplete`. Returning true just because only the description has changed, could be a false positive. For `packageAutocomplete` this is highly unlikely if not impossible, but `packageManual` would need to be considered more

korotkovao: Thank you, I have added it.

packages/abap-deploy-config-inquirer/src/prompts/questions/config/package.ts
korotkovao: Added a check for the prevAnswers, and if only the description has updated, don't run the validation.

PR #3557, opened 2025-08-21
Feat(preview-middleware): disable flex changes for ui5 from npmjs lower than 1.84

heimwege:
Follow-up to https://github.com/SAP/open-ux-tools/pull/3363

Error: Preview with UI5 1.76 from npmjs still not working because of a 404 for ObjectStorageConnector.

Solution: 
- Disable flex changes for preview with virtual endpoints using ui5 sources from npmjs `< 1.84.x` and log warning to terminal
- Adjust min local sapui5 version (e.g. used by generator for `ui5-local.yaml`) to `1.84.54`

Alternative to #3535

Decision:
- check for `SAPUI5` vs `OpenUI5` will not be taken into account

### DIFF
--- a/packages/adp-flp-config-sub-generator/CHANGELOG.md
+++ b/packages/adp-flp-config-sub-generator/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/adp-flp-config-sub-generator
 
+## 0.1.82
+
+### Patch Changes
+
+-   Updated dependencies [e015869]
+    -   @sap-ux/flp-config-inquirer@0.4.31
+    -   @sap-ux/adp-tooling@0.15.35
+
 ## 0.1.81
 
 ### Patch Changes
--- a/packages/adp-flp-config-sub-generator/package.json
+++ b/packages/adp-flp-config-sub-generator/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/adp-flp-config-sub-generator",
     "description": "Generator for adding FLP configuration to an Adaptation Project",
-    "version": "0.1.81",
+    "version": "0.1.82",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/adp-tooling/CHANGELOG.md
+++ b/packages/adp-tooling/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/adp-tooling
 
+## 0.15.35
+
+### Patch Changes
+
+-   e015869: chore: patch inquirer dependency
+
 ## 0.15.34
 
 ### Patch Changes
--- a/packages/adp-tooling/package.json
+++ b/packages/adp-tooling/package.json
@@ -9,7 +9,7 @@
     "bugs": {
         "url": "https://github.com/SAP/open-ux-tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug+label%3Aadp-tooling"
     },
-    "version": "0.15.34",
+    "version": "0.15.35",
     "license": "Apache-2.0",
     "author": "@SAP/ux-tools-team",
     "main": "dist/index.js",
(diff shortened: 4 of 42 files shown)


packages/ui5-config/src/defaults.ts
```diff
@@ -2,7 +2,7 @@ export const enum UI5_DEFAULT {
     DEFAULT_UI5_VERSION = '',
     DEFAULT_LOCAL_UI5_VERSION = '1.95.0',
     MIN_UI5_VERSION = '1.60.0',
-    MIN_LOCAL_SAPUI5_VERSION = '1.76.0',
+    MIN_LOCAL_SAPUI5_VERSION = '1.84.54',
```
devinea, line 5: Seems ok as 1.84.x is the closest currently supported UI5 version.
https://ui5.sap.com/versionoverview.html

packages/preview-middleware-client/src/flp/initConnectors.ts
```diff
@@ -11,7 +11,8 @@ import { getUi5Version, isLowerThanMinimalUi5Version } from '../utils/version';
  * intiConnectors(); // Simply call the function without any arguments.
  */
 export default async function initConnectors(): Promise<void> {
-    if (isLowerThanMinimalUi5Version(await getUi5Version(), { major: 1, minor: 76 })) {
+    const ui5Version = await getUi5Version();
```
devinea, line 14: jsdoc above should be updated 
```
 * For UI5 versions below 1.78, this function dynamically requires and executes a FakeLrepConnector.
 * For UI5 versions 1.78 and above, a local connector that reuses the WorkspaceConnector is being defined in preview-middleware/src/base/config.ts.
```

packages/preview-middleware-client/src/utils/version.ts
```diff
@@ -51,7 +54,9 @@ function checkVersionInfo(versionInfo: Ui5VersionInfo): void {
  * @returns Ui5VersionInfo
  */
 export async function getUi5Version(library: string = 'sap.ui.core'): Promise<Ui5VersionInfo> {
-    let version = ((await VersionInfo.load({ library })) as SingleVersionInfo)?.version;
+    const versionInfo = await VersionInfo.load() as { name: string; libraries: SingleVersionInfo[] } | undefined;
+    let version = versionInfo?.libraries?.find((lib) => lib.name === library)?.version;
+    const isCdn = versionInfo?.name === 'SAPUI5 Distribution';
     if (!version) {
         Log.error('Could not get UI5 version of application. Using version: 1.130.0 as fallback.');
         version = '1.130.0';
```
devinea, line 62: Unrelated. But `1.130.0` is out of maintenance. 
Shouldn't this use the latest fallback version from ui5-info?

packages/ui5-config/src/defaults.ts
heimwege: @devinea I was just wondering if we now must update the patch version on a regular basis? 🤷🏻

packages/preview-middleware-client/src/utils/version.ts
heimwege: @devinea are you referring to the patch version? How to retrieve this from `ui5-info`? I just see [1.130.*](https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-info/src/ui5-version-fallback.ts#L38) but I'm not sure how this is supposed to help finding the actual patch version 🤷🏻 And sending yet another request to the very same url that actually lead to the fallback... I'm pretty sure the request would result in the same error.

packages/preview-middleware-client/src/flp/initConnectors.ts
heimwege: done

packages/preview-middleware-client/src/utils/version.ts
devinea: My mistake I thought we had fallback versions including patches. 
The fallback here should probably be an LTS version  e.g. `1.136.9`  https://ui5.sap.com/versionoverview.html

packages/ui5-config/src/defaults.ts
devinea: No. I expect ui5-info will return the available versions above the `MIN_LOCAL_SAPUI5_VERSION` value

Issue #3562, opened 2025-08-25
BUG - Deploy Config generator does not show cloud-specific prompts for ADP when login is required


nikmace:
### Description
When you run the deploy config generator as a step in the ADP generator or standalone, and select a **CloudReady** destination in BAS (that requires authentication with credentials), only the prompts for the **OnPremise** flow are shown - meaning there are no AbapUi5Repo, Transport prompts shown.

This happens because the destination prompt does not correctly update the prompt state `isS4HC`. The first request to isAbapCloud fails because the destination requires authentication so the `isS4HC` prompt state is set incorrectly. We need to update the `isS4HC` prompt state after we have input the credentials in the password prompt validation.

<img width="1019" height="685" alt="Image" src="https://github.com/user-attachments/assets/daa19df8-ada8-4f7a-9e90-69a1c3855bb8" />

### Steps to Reproduce
1. Go to sap-ux staging instance and open the ADP generator 
2. Choose the only cloud destination there that requires credentials
3. Complete the system and attribute pages and click add deploy config
4. Your destination will be preselected. After entering the credentials and authenticating, you will not see missing cloud-specific prompts.

### Expected results
Cloud-related prompts show for CloudReady systems.

### Actual results
OnPremise flow is shown for CloudReady system that requires authentication.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3563, opened 2025-08-25
fix: Deploy Config generator does not show cloud-specific prompts for ADP when login is required

nikmace:
Fix for #3562.
- Enhances the password prompt validation (when in BAS and a flag is passed) to update the isS4HC prompt state. This state update makes the cloud-specific prompts visible.

Also, includes this fix:
- Temporarily hide application variants from the application list. Due to an issue in the App Index API, application variants are listed as available for the Adaptation Project. This will cause issues if users select such an application variant.

### DIFF
--- a/packages/abap-deploy-config-inquirer/src/prompts/validators.ts
+++ b/packages/abap-deploy-config-inquirer/src/prompts/validators.ts
@@ -1,6 +1,6 @@
 import type { IValidationLink } from '@sap-devx/yeoman-ui-types';
 import { AdaptationProjectType } from '@sap-ux/axios-extension';
-import { isAbapEnvironmentOnBtp, isS4HC, type Destinations } from '@sap-ux/btp-utils';
+import { isAbapEnvironmentOnBtp, isAppStudio, isS4HC, type Destinations } from '@sap-ux/btp-utils';
 import { ErrorHandler } from '@sap-ux/inquirer-common';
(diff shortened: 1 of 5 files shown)


packages/abap-deploy-config-inquirer/src/prompts/validators.ts
```diff
@@ -302,6 +304,11 @@ export async function validateCredentials(
         }
     });
 
+    if (isAppStudio() && shouldCheckSystemType) {
```
longieirl, line 307: I'm wondering if this should be part of this method? Would you move this logic to the `getPasswordPrompt` where `validateCredentials` is being called, soo there is a separation?

nikmace: @longieirl I am not sure where this would go here, but we need it exactly here in the validate when we have retrieved all of the credentials and we can make the request. If we put this logic in `getPasswordPrompt` method, the request will still fail and the problem will persist. Screenshot for reference.

Screenshot 2025-08-25 at 14 34 54

nikmace: Also, in other validation methods, it was done the same way

longieirl (approved):
CHanges look good.
Comments addressed.
Needs review from @cianmSAP

cianmSAP (approved):

PR #3566, opened 2025-08-26
Feat/annotation based features

marufrasully:


packages/fiori-mcp-server/src/tools/functionalities/functionalities.ts
```diff
@@ -2,16 +2,21 @@ import type { FunctionalityHandlers } from '../../types';
...
-    CREATE_CONTROLLER_EXTENSION_FUNCTIONALITY
+    CREATE_CONTROLLER_EXTENSION_FUNCTIONALITY,
+    MODIFY_ANNOTATION_FUNCTIONALITY
 ];
 
 export const FUNCTIONALITIES_HANDLERS: Map<string, FunctionalityHandlers> = new Map([
     [ADD_PAGE_FUNCTIONALITY.id, addPageHandlers],
     [DELETE_PAGE_FUNCTIONALITY.id, deletePageHandlers],
     [GENERATE_FIORI_UI_APP.id, generateFioriUIAppHandlers],
+    [CREATE_CONTROLLER_EXTENSION_FUNCTIONALITY.id, createControllerExtensionHandlers],
```
heimwege, line 19: This is a duplicate. See line 21

packages/fiori-mcp-server/src/tools/functionalities/functionalities.ts
```diff
@@ -2,16 +2,21 @@ import type { FunctionalityHandlers } from '../../types';
...
+    [MODIFY_ANNOTATION_FUNCTIONALITY.id, modifyAnnotationHandlers],
```
heimwege, line 20: As discussed yesterday, I'm wondering what is the benefit in not sticking to the current concept but wrapping a set of functionalities in an own list and serve it not in step `list-functionality` but in step `get-functionality-details`.

Current concept:
1. `list-functionality` -> List of single functionalities available for the app
2. `get-functionality-details` -> get details for one functionality from step 1.
3. `execute-functionality` -> execute functionality according to the details that have been provided (e.g. parameter details)

how you integrated it:
1. `list-functionality` -> List of single functionalities available for the app **and one functionality where the list of single functionalities for annotation based features can be requested**
2. `get-functionality-details` -> get details for one functionality from step 1 **or list of functionalities in case the feature is annotation based**
3. `execute-functionality` -> execute functionality according to the details that have been provided (e.g. parameter details) **or get documentation of one functionality from step 2 in case of annotation feature.**

So basically this is the same as introducing another tool `list-functionality-for-annotations` that somehow does the same as `list-functionality` but for a specific set of (annotation based) features.

I would suggest to keep the current concept (described at the very top) unless there are real benefits for doing otherwise. Currently I don't see any but I'm open for a discussion.
In case we merge your suggestion step 2 of the workflow (`get-functionality-details`) needs to be renamed from singular to plural `get-functionalities-details` because it is no longer information about one functionality of the app but multiple in case of annotation based features.

This would however introduce another interesting question:
In case we only reply documentation in the `get-functionality-details` tool, there will be no need for a `execute-functionality` step. So we need to adjust the description of the affected workflow steps to not confuse the agent/LLM with providing a three step workflow in case it's only a two step workflow in some cases.

So we would end up with something like:
1. `list-functionality` -> List of single functionalities available for the app (including single annotation features)
2. `get-functionality-details` -> get details for one functionality from step 1.
3. (optional) `execute-functionality` -> execute functionality according to the details that have been provided (e.g. parameter details) and that are related to something that can be executed (in case of annotation based features this will not be the case)

packages/fiori-mcp-server/src/tools/functionalities/annotation/annotation_features.json
```diff
@@ -0,0 +1,254 @@
...
+      "cdsExample": "annotate schema.RootEntities with{\n    contact @(Common : {\n        Text            : contact.name,\n        TextArrangement : #TextOnly,\n        ValueList       : {\n            Label          : '{i18n>customer}',\n            CollectionPath : 'Contacts',\n            Parameters     : [\n                {\n                    $Type             : 'Common.ValueListParameterInOut',\n                    ValueListProperty : 'ID',\n                    LocalDataProperty : contact_ID\n                }\n            ]\n        }\n    });\n};",
+      "ui5Version": []
+    }
+  },
+  {
+    "id": "highlighting-line-items-based-on-criticality",
+    "name": "Highlighting Line Items Based on Criticality",
+    "description": "Line items can be highlighted based on their criticality with the annotation ![@UI.Criticality] as part of the @UI.LineItem annotation.",
+    "category": ["list-report"],
+    "area": "content",
```
heimwege, line 43: Do we also need something like cds version information that supports this feature (similar to `ui5Version`)?

packages/fiori-mcp-server/src/tools/functionalities/annotation/features.ts
```diff
@@ -0,0 +1,18 @@
+import { readFileSync } from 'fs';
+import { FioriFeature } from './types';
+import { join } from 'path';
+
+const getAnnotationFeatures = (): FioriFeature[] => {
+    try {
+        // first try in `.fiori-ai`
+        const featuresJson = readFileSync(join(__dirname, '.fiori-ai', 'annotation_features.json'), 'utf-8');
```
heimwege, line 8: Why do we try to read it from there when we never write it to that location? Or did I miss that in the code?

packages/fiori-mcp-server/src/tools/functionalities/annotation/search.ts
```diff
@@ -0,0 +1,156 @@
+import { readFile } from 'fs/promises';
+import { FioriFeature } from './types';
+
+const rulesMsg = `
+RULES:
+    - Content of **Expected annotation pattern:** are sample example. You MUST learn from **Expected annotation pattern:**  and You MUST use Targets and **Available Dynamic Property Names:** from **Available Metadata Targets:** to annotation Targets with any dynamic property names where meaningful and possible.
+    - You MUST learn from **Expected annotation pattern:** and decides if a Target MUST be with fully qualified name or name without namespace should be used.
```
heimwege, line 7:
```suggestion
    - You MUST learn from **Expected annotation pattern:** and decide if a Target MUST be with fully qualified name or name without namespace should be used.
```

packages/fiori-mcp-server/src/tools/functionalities/annotation/search.ts
```diff
@@ -0,0 +1,156 @@
...
+    let result = `## Feature Analysis: ${feature.name}\n\n`;
+    result += `**Feature ID:** ${feature.id}\n\n`;
+    result += `**Description:** ${feature.description}\n\n`;
+    result += `**Search Path:** ${appPath}\n\n`;
+    result += `**Found ${cdsFiles.length} CDS file(s)**\n\n`;
+
+    // Analyze each CDS file for feature matches
+    const matches: Array<{ file: string; matches: string[]; content?: string }> = [];
+
+    for (const filePath of cdsFiles) {
```
heimwege, line 96: Instead of searching cds files we should first try to ask the cds mcp and only use this as fallback (compare to [here](https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-app/generate-fiori-ui-app.ts#L13)). Also using `cds compile '*' --to serviceinfo` or similar would be an option (depending on what information you need; see [capire](https://cap.cloud.sap/docs/node.js/cds-compile#cds-compile-to))

heimwege (commented):
see comments

packages/fiori-mcp-server/src/tools/functionalities/annotation/search.ts
```diff
@@ -0,0 +1,156 @@
+import { readFile } from 'fs/promises';
+import { FioriFeature } from './types';
+
+const rulesMsg = `
+RULES:
+    - Content of **Expected annotation pattern:** are sample example. You MUST learn from **Expected annotation pattern:**  and You MUST use Targets and **Available Dynamic Property Names:** from **Available Metadata Targets:** to annotation Targets with any dynamic property names where meaningful and possible.
+    - You MUST learn from **Expected annotation pattern:** and decides if a Target MUST be with fully qualified name or name without namespace should be used.
+    - If Target could not be determined from user query, ask user to select target from **Available Metadata Targets:**. Suggest list of targets to user.
```
heimwege, line 8: Instead of requesting information from a user via rules/descriptions we should rather use the [elicitation feature](https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation) from the MCP specification that covers exactly this use case.

marufrasully:
This PR bring some narrow context to support annotation and useful incase of token usages, but since embedding is place, and covers the features which this PR tried to achieve, so I mark this as obsolete and close it.

PR #3570, opened 2025-08-28
feat(fiori-mcp-server): add generate-fiori-ui-odata-app functionality

tom-sap:
add the `generate-fiori-ui-odata-app` functionality that allows the user to generate a non-cap project & app from a service URL.

### DIFF
--- a/packages/fiori-mcp-server/src/tools/functionalities/functionalities.ts
+++ b/packages/fiori-mcp-server/src/tools/functionalities/functionalities.ts
@@ -2,9 +2,12 @@ import type { FunctionalityHandlers } from '../../types';
 import { ADD_PAGE_FUNCTIONALITY, addPageHandlers, DELETE_PAGE_FUNCTIONALITY, deletePageHandlers } from './page';
 import { CREATE_CONTROLLER_EXTENSION_FUNCTIONALITY, createControllerExtensionHandlers } from './controller-extension';
 import { GENERATE_FIORI_UI_APP, generateFioriUIAppHandlers } from './generate-fiori-ui-app';
+import { GENERATE_FIORI_UI_ODATA_APP, generateFioriUIodataAppHandlers } from './generate-fiori-ui-odata-app';
+
 export const FUNCTIONALITIES_DETAILS = [
     ADD_PAGE_FUNCTIONALITY,
     GENERATE_FIORI_UI_APP,
+    GENERATE_FIORI_UI_ODATA_APP,
     DELETE_PAGE_FUNCTIONALITY,
     CREATE_CONTROLLER_EXTENSION_FUNCTIONALITY
 ];
@@ -13,5 +16,6 @@ export const FUNCTIONALITIES_HANDLERS: Map<string, FunctionalityHandlers> = new 
     [ADD_PAGE_FUNCTIONALITY.id, addPageHandlers],
     [DELETE_PAGE_FUNCTIONALITY.id, deletePageHandlers],
     [GENERATE_FIORI_UI_APP.id, generateFioriUIAppHandlers],
+    [GENERATE_FIORI_UI_ODATA_APP.id, generateFioriUIodataAppHandlers],
     [CREATE_CONTROLLER_EXTENSION_FUNCTIONALITY.id, createControllerExtensionHandlers]
 ]);
--- a/packages/fiori-mcp-server/package.json
+++ b/packages/fiori-mcp-server/package.json
@@ -24,6 +24,7 @@
         "format": "prettier --write '**/*.{js,json,ts,yaml,yml}' --ignore-path ../../.prettierignore",
         "lint": "eslint . --ext .ts",
         "lint:fix": "eslint . --ext .ts --fix",
+        "start": "npx -y supergateway --port 9881 --sessionTimeout 30000 --stdio \"node ./dist/index.js\"",
         "test": "jest --ci --forceExit --detectOpenHandles --colors"
     },
     "files": [
--- a/packages/fiori-mcp-server/src/constant.ts
+++ b/packages/fiori-mcp-server/src/constant.ts
@@ -1,4 +1,5 @@
 export const LATEST_UI5_VERSION = '1.136.7';
 export const GENERATE_FIORI_UI_APP_ID = 'generate-fiori-ui-app';
+export const GENERATE_FIORI_UI_ODATA_APP_ID = 'generate-fiori-ui-odata-app';
 export const ADD_PAGE = 'add-page';
 export const DELETE_PAGE = 'delete-page';
--- a/packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-odata-app/command.ts
+++ b/packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-odata-app/command.ts
@@ -1,0 +1,75 @@
+/* eslint-disable no-console */
+import { promises } from 'fs';
+import { promisify } from 'util';
+import { exec as execAsync } from 'child_process';
+import { dirname, join } from 'path';
+import type { ExecuteFunctionalitiesInput, ExecuteFunctionalityOutput } from '../../../types';
+import { GENERATE_FIORI_UI_ODATA_APP_ID } from '../../../constant';
+const exec = promisify(execAsync);
+
+/**
+ * Method to generate fiori app.
+ *
+ * @param params Input parameters for application generation.
+ * @returns Application generation execution output.
+ */
+export async function command(params: ExecuteFunctionalitiesInput): Promise<ExecuteFunctionalityOutput> {
+    const { appGenConfig = {} } = params.parameters;
+    let { projectPath = '' } = params.parameters;
+    if (!projectPath) {
+        projectPath = params.appPath;
+    }
+
+    console.log('Starting Fiori UI generation...');
+    console.log('Project path is:' + projectPath);
+
+    if (!projectPath || typeof projectPath !== 'string') {
+        throw new Error('Please provide a valid path to the non-CAP project folder.');
+    }
+    if (!appGenConfig || typeof appGenConfig !== 'object') {
+        throw new Error('Invalid appGenConfig. Please provide a valid configuration object.');
+    }
+    const project = 'project' in appGenConfig && typeof appGenConfig.project === 'object' ? appGenConfig.project : {};
+    const appName = project && 'name' in project && typeof project.name === 'string' ? project.name : 'default';
+    const appPath = join(projectPath, 'app', appName);
+    try {
+        const targetDir = projectPath;
+        const generatorConfig = appGenConfig;
+
+        const configPath = `.fiori-ai/${appName}-generator-config.json`;
+        const outputPath = join(targetDir, configPath);
+        const content = JSON.stringify(generatorConfig, null, 4);
+        const configDir = dirname(outputPath);
+
+        await promises.mkdir(configDir, { recursive: true });
+        await promises.writeFile(outputPath, content, { encoding: 'utf8' });
+        const command = `npx -y yo@4 @sap/fiori:headless ${configPath} --force --skipInstall`.trim();
+
+        const { stdout, stderr } = await exec(command, { cwd: targetDir });
+        console.log(stdout);
+        if (stderr) {
+            console.error(stderr);
+        }
+    } catch (error) {
+        console.error('Error generating application:', error);
+        return {
+            functionalityId: GENERATE_FIORI_UI_ODATA_APP_ID,
+            status: 'Error',
+            message: 'Error generating application: ' + error.message,
+            parameters: params.parameters,
+            appPath,
+            changes: [],
+            timestamp: new Date().toISOString()
+        };
+    }
+
+    return {
+        functionalityId: GENERATE_FIORI_UI_ODATA_APP_ID,
+        status: 'Success',
+        message: 'Generation completed successfully: ' + appPath,
+        parameters: params.parameters,
+        appPath,
+        changes: [],
+        timestamp: new Date().toISOString()
+    };
+}
--- a/packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-odata-app/generate-fiori-ui-odata-app.ts
+++ b/packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-odata-app/generate-fiori-ui-odata-app.ts
@@ -1,0 +1,218 @@
+import { command } from './command';
+import { GENERATE_FIORI_UI_ODATA_APP_ID, LATEST_UI5_VERSION } from '../../../constant';
+import type {
+    ExecuteFunctionalitiesInput,
+    ExecuteFunctionalityOutput,
+    FunctionalityHandlers,
+    GetFunctionalityDetailsOutput
+} from '../../../types';
+
+export const GENERATE_FIORI_UI_ODATA_APP: GetFunctionalityDetailsOutput = {
+    id: GENERATE_FIORI_UI_ODATA_APP_ID,
+    name: 'Generate SAP Fiori UI Application for non-CAP Projects',
+    description: `Creates (generates) a new SAP Fiori UI application within an existing project (RAP or other non-CAP).
+                Crucially, you must first construct the appGenConfig JSON argument.
+                If not provided - you **MUST** ask the user for the servicePath and host of the OData service they want to use.
+                Then, you **MUST** query the service metadata endpoint to retrieve the list of available entities.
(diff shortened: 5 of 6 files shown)


heimwege:
Tool name must be adjusted to generate_fiori_ui_odata_app because the tool names have been aligned to snake_case in #3655

tom-sap:
@heimwege 
> Tool name must be adjusted to generate_fiori_ui_odata_app because the tool names have been aligned to snake_case in #3655

Not sure if I understand what you mean.
`generate-fiori-ui-odata-app` is not an MCP tool name, but rather a "functionality" name.
And it follows the other functionality names using kebab-case:

https://github.com/SAP/open-ux-tools/blob/5a753a6c2bd2c3832f39a6501cd1bc789e3a9106/packages/fiori-mcp-server/src/constant.ts#L2-L5

packages/fiori-mcp-server/src/tools/functionalities/fetch-service-metadata/details.ts
```diff
@@ -0,0 +1,26 @@
+import type { GetFunctionalityDetailsOutput } from '../../../types';
```
heimwege, line 1: Can this new functionality be used standalone or only in combination with `generate-fiori-ui-odata-app`? In case of the second we should not offer it as standalone functionality. I'm afraid agents might use it for non intended purposes and will fail / be confused.

packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-application/details.ts
```diff
@@ -0,0 +1,26 @@
+import type { GetFunctionalityDetailsOutput } from '../../../types';
+import { generatorConfigODataJson as parameters } from '../../schemas';
+
+export default {
+    functionalityId: 'generate-fiori-ui-odata-app',
+    name: 'Generate SAP Fiori UI Application for OData Projects (non-CAP)',
+    description: `Creates (generates) a new SAP Fiori UI application within an existing project (RAP or other non-CAP). Steps:
+
+        1. Crucially, you must first construct the appGenConfig JSON argument.
+            - If the user has not provided a valid servicePath and host (URL) of the OData service they want to use, you **MUST** ask for it.
```
heimwege, line 10: Should we use [elicitation](https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation) here?

packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-application/details.ts
```diff
@@ -0,0 +1,26 @@
...
+
+            - If the user provided a reference to a SAP system (name, host or full URL), or provided a full URL as a service,
+                you **MUST** try to retrieve the service's metadata using the "fetch-service-metadata" functionality with "execute_functionality" tool.
```
heimwege, line 13: That's basically skipping the three step workflow (1. list 2. get details 3.execute). Not sure if this will motivate the agent to not adhere to that workflow in other cases as well

packages/fiori-mcp-server/src/tools/functionalities/fetch-service-metadata/service-metadata.ts
```diff
@@ -0,0 +1,160 @@
+import type { BackendSystem } from '@sap-ux/store';
```
heimwege, line 1: Typo in filename `serevice-metadata.ts`

packages/fiori-mcp-server/src/utils/utils.ts
```diff
@@ -0,0 +1,23 @@
+import type { PackageInfo } from '@sap-ux/nodejs-utils';
+
+import { promisify } from 'util';
+import { exec as execAsync } from 'child_process';
```
heimwege: see sonar findings

```suggestion
import { exec as execAsync } from 'node:child_process';
```

packages/fiori-mcp-server/src/tools/functionalities/fetch-service-metadata/details.ts
tom-sap: 🤔, so:
1. I think it can be used separately
2. If the agent's provide the correct parameters - they will get the (clearly defined) result, so not sure what do you mean by "non intended"
3. The description of the functionality clearly states when, how and by what other functionality this is needed. 
4. I feel that merging this together with the odata app generation may actually add a lot of complexity to the LLM "steering". If you follow the logic defined in both descriptions - it's already quite complicated (lots of conditions).
5. Lastly, this seem to be running very smoothly (no follow up questions from LLM, and no getting lost, which is not easy to achieve. Can we maybe try that first and fix when there are problems?

packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-application/details.ts
tom-sap: I don't see a reason. We don't use it for the main prompt, so why have it for part of it?
Also - we don't expect just one fixed thing (like name or a number). User can provide:
- a service url (if no auth is needed)
- sap system host and service path
- sap system name and service path
- only service path (if e.g. they have 1 sap system stored and that service belongs to it)

Giving them a form is limiting and IMHO, a step back from the natural language.

packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-application/details.ts
tom-sap: No, it's not. This just steers the LLM to use the other functionality to get the data it needs.
The LLM still follows the 3 steps: gets the list of all functionalities at the beginning, and always gets the details before it executes.

packages/fiori-mcp-server/src/utils/utils.ts
tom-sap: fixed

packages/fiori-mcp-server/src/tools/functionalities/fetch-service-metadata/service-metadata.ts
tom-sap: fixed

packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-application/details.ts
heimwege: _using the "fetch-service-metadata" functionality with "execute_functionality" tool_ sounded to me like "do not use the list functionalities (strep 1) and find a suitable one, then get the details of that functionality (step 2) but directly execute `fetch-service-metadata` (step 3)".
But I might have also misinterpreted the description. If the agent doesn't show that behavior everything is fine 😸

packages/fiori-mcp-server/src/tools/functionalities/fetch-service-metadata/details.ts
heimwege: Hm. Maybe my first thoughts were wrong. The output of that is needed as input of the generator, correct? Then we anyway need that additional roundtrip.
And the agent could also make use of that functionality to update the `metadata.xml` at a later point in time, right? Did you already try if a prompt like _Please update the metadata.xml of my app_ will work?
Maybe a slight update in the description would be needed to indicate that updating `metadata.xml` is supported as well?
Something like _[...] Finally, it stores the fetched metadata in the appPath folder as metadata.xml or updates and existing one_

tom-sap: Hi @heimwege,

> Did you already try if a prompt like Please update the metadata.xml of my app will work?

No, I haven't tried that.
I'm not sure why would the user do that (can you describe a use case?).

Here `metadata.xml` file is a temporary one, that is used in the app-gen config to generate the app. After that this file is removed.

heimwege: Maybe we're not talking about the same file. I thought this is about `webapp/localService/metadata.xml` used by e.g. the fe mockserver, Fiori tools Application Modeler, Service Modeler, etc.

tom-sap: Yes, this is a different `metadata.xml` file 🙂
This one is basically a response for `.../$metadata` request to a service endpoint, and is used in the app-gen config (but also LLM uses it to figure out the entities).

heimwege: I'm still confused because what you are describing sounds exactly like the service `metadata.xml` I'm referring to as well.
Maybe using a real metadata.xml from an existing app helps.. Can you please have a look at https://github.com/heimwege/cap-node-fe-ts-sample/blob/main/app/samples/webapp/localService/metadata.xml and tell me if this is such a `metadata.xml` file that would be loaded from the backend (in my sample it's a cap node backend, let's imagine it would be ABAP)? And if yes why does the generator create its own copy of the service metadata that is during the generation stored in `localService/metadata.xml` and then the generator owned copy is deleted?

devinea: The `fetch-service-metadata`  tool downloads the service $metadata locally (in the appPath) as a temporary file.
The `generate-fiori-ui-odata-app` internally reads the temporary file and passes it in with the params to the headless generator which in turn eventually writes the contents to the generated project's `localService/mainService/metadata.xml` . 
The temporary file is deleted. 

So yes theoretically `fetch-service-metadata` could be used to update `localService/mainService/metadata.xml` directly in an existing project. We could look into that in a follow-up as there maybe other considerations such as updates to the manifest.json, ui5-mock.yaml etc.. that is normally covered by the odata-service-writer.

heimwege: Thanks for the clarification, that would be a cool enhancement 👍🏻

.changeset/healthy-walls-stare.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-ux/fiori-mcp-server": patch
+---
+
+feat(fiori-mcp-server): add generate-fiori-ui-application functionality
```
heimwege: As this text is only in the changeset of the fiori mcp I think we can skip naming it in the changeset. But we should list renaming the existing tool.

```suggestion
feat: add generate-fiori-ui-application functionality and rename existing generate-fiori-ui-app to generate-fiori-ui-application-cap
```

packages/fiori-mcp-server/src/tools/functionalities/functionalities.ts
```diff
@@ -1,23 +1,34 @@
...
 
 export const FUNCTIONALITIES_HANDLERS: Map<string, FunctionalityHandlers> = new Map([
     [ADD_PAGE, addPageHandlers],
     [DELETE_PAGE, deletePageHandlers],
-    [GENERATE_FIORI_UI_APP_ID, generateFioriUIAppHandlers],
-    [CREATE_CONTROLLER_EXTENSION_FUNCTIONALITY_ID, createControllerExtensionHandlers]
+    [GENERATE_FIORI_UI_APPLICATION_CAP_ID, generateFioriUIApplicationCapHandlers],
+    [CREATE_CONTROLLER_EXTENSION_FUNCTIONALITY_ID, createControllerExtensionHandlers],
+    [generateFioriUIApplication.id, generateFioriUIApplication.handlers],
+    [fetchServiceMetadata.id, fetchServiceMetadata.handlers]
```
heimwege, line 33: Consistency: Is there a reason Why this is implemented differently than the existing functionalities that export dedicated upper case constants variables for ID and handler functions (e.g. `[CREATE_CONTROLLER_EXTENSION_FUNCTIONALITY_ID, createControllerExtensionHandlers]`)?

packages/fiori-mcp-server/test/integration/scenarios/test-dev.yaml
```diff
@@ -2,11 +2,11 @@
 providers:
     - id: file://../provider/index.ts
       config:
-          model: 'gpt-4o'
+          model: 'gpt-5-mini'
```
heimwege: @815are do we want to change our promptfoo config model?

heimwege (commented):
Just some nit-pick comments but in general I think this PR looks quite good and is ready to go. Thanks a lot @tom-sap

packages/fiori-mcp-server/test/integration/scenarios/test-dev.yaml
tom-sap: yes. it looks like the 4o cannot even provide the input params correctly.
(and 5-mini is also 5-10 times cheaper)

packages/fiori-mcp-server/src/tools/functionalities/functionalities.ts
tom-sap: The existing structure is extremely convoluted, requires changes in several files (instead of 1), uses a const for a value (functionalityId) that is used only once in a single place (and should be reused from there) and uses less-than-ideal data structures.
It took me a lot of time to figure out why is this that way (because there is no reason to make it this complex).

This could (and should) be done much simpler, but even with the current setup, this here is an improvement.

Ideally: `functionalities` folder should export an array of functionalities and the wrappers should only need that to provide mcp tools (list, details, execute).
Currently there are several "lists" where a functionality must be manually added.

My hope is that this can gradually evolve into a better structure. But if not, I can force my code to match the rest 😉

packages/fiori-mcp-server/test/integration/scenarios/test-dev.yaml
heimwege: But we can't just update the model we use for our tests either just because the current model does something wrong for one new functionality.
I can take the question about changing our baseline model to our mcp sync and report back.

tom-sap: ok. it seems to work fine now (with the updated prompt). 
So I reverted the model change.

815are: I am not sure, can we enhance descriptions to make 'gpt-4o'  work? for me `gpt-4o` seems worked ok for my tests so far(but required small adjustment in description for input params at beginning), but not sure about this particular case

packages/fiori-mcp-server/src/tools/functionalities/functionalities.ts
heimwege: Refactoring the others is of course good as well 👍🏻
I'll create a TBI for that.

packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-application/execute-functionality.ts
```diff
@@ -0,0 +1,88 @@
...
+        }
+        if (existsSync(metadataPath)) {
+            await FSpromises.unlink(metadataPath);
+        }
+    }
+
+    return {
+        functionalityId: details.functionalityId,
+        status: 'Success',
+        message: `Generation completed successfully. You must run \`npm install\` in ${appPath} first, and then run the application using \`npm run start-mock\`.`,
```
devinea:
```suggestion
        message: `Generation completed successfully. You must run \`npm install\` in ${appPath} first, and then run the application using \`npm run start\`.`,
```
I think since we are generating an app from real service then preview should use the real rather than mock. Any reason not too?

tom-sap: no reasons 🙂 
I suppose mock demoes better, as it just works, whereas a real service will ask for credentials, which i need to find and copy & paste.

tom-sap:
> First call to execute ` "functionalityId": "generate-fiori-ui-application",` returns `Error: appPath parameter is required`
> 
> mcp client populates and it proceeds successfully. Have you seen it?

Hi @devinea,
No, I haven't. For me it always passes the `appPath`. 

Which model were you using?

packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-application/execute-functionality.ts
devinea: If using SAP Systems the preview should automatically use the credentials from the store.

devinea:
> > First call to execute ` "functionalityId": "generate-fiori-ui-application",` returns `Error: appPath parameter is required`
> > mcp client populates and it proceeds successfully. Have you seen it?
> 
> Hi @devinea, No, I haven't. For me it always passes the `appPath`.
> 
> Which model were you using?

`Claude-4-Sonnet`

PR #3572, opened 2025-08-28
fix(security): axios upgrade fix for publish

donal-tobin-sap:
- fix issues with napi-postinstall
- ensure that the lock file is not updated during publish

Issue: tools-suite#35136


donal-tobin-sap:
- fixes to the publish steps
- - security fixes for supply chain vulnerabilities

PR #3573, opened 2025-08-28
fix(test): version used in test cannot be hardcoded

donal-tobin-sap:
Skip test that is using a hard coder version number

- each patch version will change the version and this may come from other modules direct and indirect


donal-tobin-sap:
In discussion with @devinea reset version to 0.0.0 
Skip of test has to be kept as offical publish would fail on it when done

devinea (approved):
Thanks @donal-tobin-sap 
As
Discussed We should keep version on 0.0.0 for now and fix test.

donal-tobin-sap:
using admin as this is blocking PRs

PR #3576, opened 2025-08-29
fix(fiori-mcp): delete config file used for headless generation

heimwege:


Klaus-Keller (approved):
Thanks @heimwege, clean up after generation looks good.

devinea (approved):
Cleanup and tests lg. 
No changeset expected 👍

PR #3582, opened 2025-09-01
feat: rename property `id` to `functionalityId`

815are:
We noticed that sometimes tools are called with functionality id. As enhancement - we can be more precise and rename property `id` to `functionalityId` for tool `list-functionality` and `get-functionality-details`.


heimwege (approved):
Thanks for the harmonization @815are  👍🏻

Klaus-Keller (approved):
Thanks! Rename of id property looks good.

PR #3583, opened 2025-09-01
fix(store): correction for returning recovered backend systems

cianmSAP:
Ensures the recovered properties are returned on the initial `getAll` call


longieirl (approved):
Changeset is good.
Tests to cover change.
Changes are good.

PR #3585, opened 2025-09-02
feat(serve-static-middleware): make servestatic ignore cachebuster in URLs

heimwege:
Add a new config option `keepCacheBusterInUrl` to the serve-static-middleware to revert the new default that ignores the cache buster part of the URL.

### DIFF
--- a/.changeset/flat-suits-allow.md
+++ b/.changeset/flat-suits-allow.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/serve-static-middleware': minor
+---
+
+feat: by default the cache buster part of the URL will be ignored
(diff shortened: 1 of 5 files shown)


Klaus-Keller (commented):
Hi @heimwege, 

I did a local test and this works for me. I can proxy a connection to S/4 public cloud system and overwrite a specific SAP Fiori artifact (in my case a Fiori launchpad plugin) with local sources. As `path` in the `ui5.yaml` config I can specify the path without cache id, for instance:

Instead of 
```
[..]
- path: /sap/bc/ui5_ui5/plugin/~5A28...C03~5
```

we now can use
```
[..]
- path: /sap/bc/ui5_ui5/plugin/
```

Klaus-Keller (approved):
Thanks @heimwege,

* addition to filter cache buster by default looks good
* changeset exists
* test coverage is great
* did a visual review and local test with the code, works for me

Approved from my side

Issue #3594, opened 2025-09-02
BUG - Deploy Config generator does not allow package with CloudReady and OnPremise project types


nikmace:
### Description
The customers are unable to deploy the Adaptation Project from SAP Business Application Studio onto on-premise system. In the Deployment Configuration Generator wizard, after selecting the destination and the package, they see the error `The provided package is not intended for on-premise deployments.` and are unable to proceed with the deploy config generation.

<img width="678" height="421" alt="Image" src="https://github.com/user-attachments/assets/601d85b9-a9ee-4b65-b734-71303146efa5" />

This behavior is happening because the package has both CloudReady and OnPremise project types. When making a call to system info with the provided package, we get back two project types. Then, the validation for the package fails, and we see the error above.

### Steps to Reproduce
1. Open the ADP Generator
2. Select an OnPremise system and go through the steps
3. Select Add Deploy Config, your system will be preselected
4. Enter you package which has both cloud_dev, and standard (cl_abap_language_version)

### Expected results
Customers are able to deploy, the message does not appear when there are two project types for the package

### Actual results
Package has two project types, which causes the validation to fail and dissalow customer to create the deploy config and continue with the project deployment.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3595, opened 2025-09-02
fix: Deploy Config generator does not allow package with CloudReady and OnPremise project types

nikmace:
Fix for #3594.
- Adds a check when the package has two project types. Then, we allow the user to continue with deployment.

Also,
- Removes this temporary fix, hiding application variants from the application list. Due to an issue in the App Index API (which is now fixed), application variants are listed as available for the Adaptation Project. This will cause issues if users select such an application variant.

### DIFF
--- a/.changeset/itchy-crabs-bake.md
+++ b/.changeset/itchy-crabs-bake.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/abap-deploy-config-inquirer': patch
+---
+
+fix: Deploy Config generator does not allow package with CloudReady and OnPremise project types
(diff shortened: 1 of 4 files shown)


IvoSG (approved):
Code looks good
Excellent coverage
Changeset is ok
Did not test manually

GDamyanov (approved):
Changeset present.
Code looks good.
Excellent code coverage.

cianmSAP (approved):
Updates look good
The change does not touch fiori gen as it's behind `shouldValidatePackageType` option
Changeset and tests to cover ✅

PR #3598, opened 2025-09-03
fix: add missing referential constraints

voicis:
Added missing referential constraints for navigation properties in `MetadataElement` interface.
Added missing referential constraints when converting to AVT types in Fiori Annotation API.

### DIFF
--- a/packages/xml-odata-annotation-converter/src/parser/metadata.ts
+++ b/packages/xml-odata-annotation-converter/src/parser/metadata.ts
@@ -1,6 +1,11 @@
 import type { XMLDocument, XMLElement } from '@xml-tools/ast';
 
-import type { MetadataElementProperties, MetadataElement, TargetKind } from '@sap-ux/odata-annotation-core-types';
+import type {
+    MetadataElementProperties,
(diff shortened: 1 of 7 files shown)


packages/xml-odata-annotation-converter/src/parser/metadata.ts
```diff
@@ -364,6 +398,114 @@ function createMetadataElementNodeForType(
     };
 }
 
+/**
+ * Converts string value to number.
```
marufrasully: jsdoc is mis matching.

packages/xml-odata-annotation-converter/src/parser/metadata.ts
```diff
@@ -364,6 +398,114 @@ function createMetadataElementNodeForType(
...
+ *
+ * @param value Input
+ * @returns Parsed number or undefined otherwise
```
marufrasully: same here.

marufrasully (commented):
Thanks @voicis 
- Code wise looks good
- Secured with unit test - great code coverage
- Test it locally

PR #3600, opened 2025-09-03
fix(mcp): add timeout configuration for MCP server and ensure absolute service paths

devinea:
feat: add timeout configuration for MCP server and ensure absolute service paths

### DIFF
--- a/packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-app/command.ts
+++ b/packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-app/command.ts
@@ -72,7 +72,12 @@ export async function command(params: ExecuteFunctionalitiesInput): Promise<Exec
     if (!projectPath || typeof projectPath !== 'string') {
         throw new Error('Please provide a valid path to the CAP project folder.');
     }
-
+    if (generatorConfig?.service.servicePath) {
+        // Add leading slash if missing as cds-plugin-ui5 expects absolute service paths
+        generatorConfig.service.servicePath = generatorConfig?.service.servicePath?.startsWith('/')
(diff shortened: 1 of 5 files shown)


johannes-kolbe (approved):
- Changes lgtm,
- Did not test manually
- changeset provided

packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-app/command.ts
```diff
@@ -72,7 +72,12 @@ export async function command(params: ExecuteFunctionalitiesInput): Promise<Exec
     if (!projectPath || typeof projectPath !== 'string') {
         throw new Error('Please provide a valid path to the CAP project folder.');
     }
-
+    if (generatorConfig?.service.servicePath) {
+        // Add leading slash if missing as cds-plugin-ui5 expects absolute service paths
```
heimwege: Actually that comment is not correct. It's not the `cds-plugin-ui5` that expects something it's CAP that by default serves at `odata/v4/<servicename>` and if you deviate from this you need to add a `@path` annotation to your service.
So maybe it's the plugin that should adjust the cap path for a service as well or something. I need to find out.
I addressed that question to @petermuessig.

heimwege (approved):
Workaround looks good to me. FYI: see my comment for comment in the code

packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-app/command.ts
devinea: I'll remove the comment as it is a bit misleading.

Klaus-Keller (approved):

heimwege (approved):
re-approve

PR #3601, opened 2025-09-03
fix: make variant management script intent dynamic

fionan-SAP:


fionan-SAP:
Preview launches successfully:
Screenshot 2025-09-10 at 14 19 36
Screenshot 2025-09-10 at 14 18 40

packages/fiori-generator-shared/src/npm-package-scripts/getPackageScripts.ts
```diff
@@ -63,10 +63,14 @@ function buildStartCommand(localOnly: boolean, params: string, startFile?: strin
  * When virtual endpoints are used, the search parameters are injected at runtime.
  *
  * @param {boolean} addSearchParams - Indicates whether to include search parameters in the command.
+ * @param projectName
```
IainSAP: Is this the correct terminology? There is no `projectName` property in the application interfaces.  It should be `flpAppId` I think for consistency and clarity?

Also could you add a description that indicates its usage please.

PR #3608, opened 2025-09-05
fix: update primitve value

voicis:
Fixed pointer conversion logic to correctly handle values for shortcut annotation syntax.


vadson71 (approved):
Changes look good
Good test coverage
Didn't test locally

Issue #389, opened 2022-03-14
FEATURE - retrieve OData metadata for `ValueListReferences`


Klaus-Keller:
As a **consumer of the `@sap-ux/axios-extension`** (former `odata-client`),
I want to **be able to download metadata of referenced value help services**,
so that I **can create mock data for them**.

### Description
Starting with SAP Fiori elements for OData V4, value helps of applications are modeled as separate OData service endpoints. In order to write automated tests, there might be the need for mocking value help data. To be able to create mock data for value helps, we need to know the data structure behind them. This data structure can be retrieved from metadata of each value help service endpoint.

### Technical Design
The module `@sap-ux/axios-extension` (former `odata-client`) allows structured access to OData services. It provides functionality to retrieve the metadata. For an OData V4 service, this metadata might contain one or multiple annotations `SAP__common.ValueListReferences` that points to the URL of a value help service. Instead of parsing the metadata for value help references by each consumer, module `@sap-ux/axios-extension` could provide this functionality allowing convenient access to the metadata and metadata of all included value help references.

### Acceptance Criteria
**Given** I have a OData V4 service including value help references
**when** I execute the function to retrieve value help metadata
**then** I'll be able to store metadata of all referenced value help.

### Notes
**Example**
Metadata for V4 List Report Object Page service contains:
```
...
<Annotation Term="SAP__common.ValueListReferences">
  <Collection>
    <String>relative/path/to/value/list/$metadata</String>
  </Collection>
</Annotation>
...
```

This means, the metadata for value help values can be found at:
`relative/path/to/value/list/$metadata`

### Tasks
- [ ] Unit Tests added
- [ ] Test Cases defined

tobiasqueck:
@Klaus-Keller I am not sure this should go into the `@sap-ux/axios-extension` since it would require that it parses the metadata. So far, the module only simplifies getting the metadata but does not do anything with it. Your scenario could be easily implemented outside of the module with.

```typescript
import { ODataServiceProvider } from '@sap-ux/axios-extension';

const provider = ODataServiceProvider.create({
    baseURL: 'https://sap.example',
    params: { 'sap-client': client }
});
const service = provider.service('/ns/my_service');
const metadata = await service.metadata();
const valueListRefs = yourParser(metadata);

valueListRefs.forEach((ref) => {
   valueListRefs[ref] = provider.service(ref).metdata();
});

```

tobiasqueck:
[...]

PR #3609, opened 2025-09-05
feat: collect and cache ValueListReferences

voicis:
#389 

Check metadata and collect `ValueListReferences` annotations in order to store the value help metadata.

### DIFF
--- a/packages/fiori-elements-writer/src/index.ts
+++ b/packages/fiori-elements-writer/src/index.ts
@@ -5,6 +5,7 @@ import type { App, Package } from '@sap-ux/ui5-application-writer';
 import { generate as generateUi5Project } from '@sap-ux/ui5-application-writer';
 import {
     generate as addOdataService,
+    generateValueListReferences,
     OdataVersion,
     ServiceType,
     type OdataService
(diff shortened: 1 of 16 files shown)


packages/fiori-elements-writer/src/index.ts
```diff
@@ -102,13 +104,15 @@ function shouldAddTest(service: Partial<OdataService>, addTests?: boolean): bool
...
  * @param data - configuration to generate the Fiori elements application
+ * @param provider - The service provider to fetch metadata and annotations
  * @param fs - an optional reference to a mem-fs editor
  * @param log - optional logger instance
  * @returns Reference to a mem-fs-editor
  */
 async function generate<T extends {}>(
     basePath: string,
     data: FioriElementsApp<T>,
+    provider: AbapServiceProvider | undefined,
```
IainSAP, line 115: There should not be any service provider references in writers.

IainSAP (changes requested):
Writers must only take inputs and write outputs. No heavy lifting, network calls or user input can be required. This is to maintain the generator architecture separating the phases.

voicis:
Closing this in favor of #3794

PR #3611, opened 2025-09-07
Disable change table column QA when variant management not enabled

Jimmy-Joseph19:
When variantmanagement of table is set to false change table column quick action must be disabled as the variant cannot be stored when the setting is false

### DIFF
--- a/packages/generator-adp/CHANGELOG.md
+++ b/packages/generator-adp/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/generator-adp
 
+## 0.4.0
+
+### Minor Changes
+
+-   e36b965: fix(adp-generator): In case the base app inbounds api call fails for the selected application we display the error next to the application prompt.
+
 ## 0.3.78
 
 ### Patch Changes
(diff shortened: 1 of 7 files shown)


packages/preview-middleware-client/src/messagebundle.properties
```diff
@@ -83,6 +83,7 @@ HIGHER_LAYER_CHANGES_TITLE = Higher Layer Changes Detected
 HIGHER_LAYER_CHANGES_INFO_MESSAGE=The preview of the project was reloaded due to detected changes in higher layer than the one used in your project.
 
 TABLE_CUSTOM_COLUMN_ACTION_NOT_AVAILABLE=This action has been disabled because the table rows are not available. Please load the table data and try again
+TABLE_CHANGE_COLUMN_ACTION_NOT_AVAILABLE=This action has been disabled because the variant management is disabled. Please enable and try again.
```
lfindlaysap:
```suggestion
TABLE_CHANGE_COLUMN_ACTION_NOT_AVAILABLE=This action has been disabled because variant management is disabled. Enable variant management and try again.
```

lfindlaysap (approved):

voicis (approved):

Issue #3610, opened 2025-09-05
BUG - Extensibility generator cannot be resolved in the bundled ADP generator


nikmace:
### Description
The customer reported an error while attempting to create an Extension Project in BAS with the ADP generator.
The error message displayed: "The Extensibility Project generator plugin was not found in your dev space and it is required for this action. To proceed please install the  extension.” 

This is happening because the bundled ADP generator cannot resolve the node paths of the current dev space. Therefore, no extensibility generator is found (but checking manually it does exist).

### Steps to Reproduce
1. Open a dev space of your choice and open ADP generator
2. Select an OnPremise system and choose an application that is not supported
3. You will see an error message after being prompted with yes / no question (whether you want to create an extension project)
4. See the error

### Expected results
The user is able to create an extension project

### Actual results
Users cannot create an extension project because extensibility generator is not found

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

jun-wu:
what is "bundled ADP generator "?

PR #3612, opened 2025-09-09
fix: Extensibility generator cannot be resolved in the bundled ADP generator

nikmace:
Fix for #3610.
- Registers the extensibility generator and directly composes with its namespace.
- Adds a check for whether the extensibility extension is installed in the dev space.


packages/generator-adp/src/app/questions/helper/validators.ts
```diff
@@ -19,17 +18,17 @@ interface JsonInputParams {
  * @param {boolean} value - A confirm flag indicating whether user wants to continue creating an extension project.
  * @param {boolean} isApplicationSupported - Whether the selected application is supported.
  * @param {boolean} hasSyncViews - Whether synchronized views exist for the app.
+ * @param {boolean} isExtensibilityGenInstalled - Whether the extensibility sub-generator is installed.
  * @returns {boolean | string} Returns true if app is supported and contains sync views, or an error message if not.
  */
 export function validateExtensibilityGenerator(
```
testojs: There are too many boolean parameters here. You could consider using a single options/configuration object instead.

nikmace: Yes, I agree with this. Changed here: [febba64](https://github.com/SAP/open-ux-tools/pull/3612/commits/febba6429c43102a30ff07d5d87c80d567bd1c71)

nikmace:
@SAP/ux-tools-text-reviewers, there is no change in texts, just the variable name has changed.

testojs (approved):
* code looks good
* comments addressed
* excellent test coverage
* changeset OK
* not tested manually

mmilko01 (approved):
Changeset OK
Code change looks good
Excellent test coverage
Did not test manually

PR #3613, opened 2025-09-09
feat(fiori-mcp-server): Use Zod schemas to generate input/output JSON schemas and TypeScript types

815are:
In previous version we had issue that we needed maintain json schemas and typescript interfaces manually - we were required to sync between json schema and typescript interfaces, which required double maintenance.
Example:
1. typescript interface `ListFunctionalitiesOutput ` -> https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-mcp-server/src/types.ts#L33
2. json schema of `ListFunctionalitiesOutput ` -> https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-mcp-server/src/tools/output-schema/list-functionality.json

Proposal is to use zod and define schema in zod and then:
1. generate json schema using `zod.toJSONSchema`;
3. resolve typescript type using `zod.infer`;

In result we only need to maintain zod schemas defined here -> [basic.ts](https://github.com/SAP/open-ux-tools/blob/feat/mcpUnifyInputOutputInterfacesAndSCHEMAS/packages/fiori-mcp-server/src/types/basic.ts), [input](https://github.com/SAP/open-ux-tools/blob/feat/mcpUnifyInputOutputInterfacesAndSCHEMAS/packages/fiori-mcp-server/src/types/input.ts), [output](https://github.com/SAP/open-ux-tools/blob/feat/mcpUnifyInputOutputInterfacesAndSCHEMAS/packages/fiori-mcp-server/src/types/output.ts)


heimwege:
Should we move the generator zod stuff to the new structure as well and adjust to the new concept? (command.ts and generate-fiori-ui-app.ts)

815are:
> Should we move the generator zod stuff to the new structure as well and adjust to the new concept? (command.ts and generate-fiori-ui-app.ts)

@heimwege
I think yes, but I would like to little bit discuss/clarify beforehand and I think better handle it separate.
Main goal of this PR was to avoid double maintenance between typescript interfaces and json schemas for tools, where if we manually do not correctly maintain both sources, then we can see output about invalid schema.
`command.ts and generate-fiori-ui-app.ts` is more about functionalities and interface of `parameters` - perhaps we can define it with zod as well and details return converted json schema, but I think we need try and discuss this before

heimwege:
> > Should we move the generator zod stuff to the new structure as well and adjust to the new concept? (command.ts and generate-fiori-ui-app.ts)
> 
> @heimwege
> I think yes, but I would like to little bit discuss/clarify beforehand and I think better handle it separate.
> Main goal of this PR was to avoid double maintenance between typescript interfaces and json schemas for tools, where if we manually do not correctly maintain both sources, then we can see output about invalid schema.
> `command.ts and generate-fiori-ui-app.ts` is more about functionalities and interface of `parameters` - perhaps we can define it with zod as well and details return converted json schema, but I think we need try and discuss this before

@815are the generator stuff it is already defined in zod. That's why I'm asking. But it makes sense to do this separately 👍🏼

devinea (approved):
Thanks @815are approach seems great.

Issue #3619, opened 2025-09-12
BUG - If user has no rights, selecting XID/080 and clicking Next leads to crash : @sap/fiori:adp generator failed - env.on(error) Request failed with status code 403


avasilev-sap:
### Related Feature
Feature request: _issue number_

### Description
If user has no rights, selecting XID/080 and clicking Next leads to crash : @sap/fiori:adp generator failed - env.on(error) Request failed with status code 403

### Steps to Reproduce
1. Start the adaptation project generator.
2. Select the system XID/080.
3. Enter credentials.
4. Select any application and click Next.

### Expected results
The exception thrown from the base application inbounds api call(the user has no privileges to do the call) need to be handled in the generator. The user should be prevented from proceeding with the project generator. The Next button need to be disabled.

### Actual results
The generator crashes.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

nikmace:
Missing labels, bug type, scope is not needed for bugs

PR #3626, opened 2025-09-15
fix(adp-generator): Move inbounds fetch inside the app prompt validation

avasilev-sap:
#3619

Move the base app inbound api call inside the select application prompt validator. In case any error occur the prompt display the error message and the user cannot proceed with the project generation. The api call has a single parameter applicationId so seems reasonable to display the error message there.

Testing done: manually verified in BAS with cloud application and user which has no privileges to fetch app inbounds. The error message appears in the wizard and in the cli scenario.

Screenshot 2025-09-15 at 18 25 19
Screenshot 2025-09-15 at 18 23 41


nikmace (changes requested):
A couple of things are missing:
- Changeset for the changed package.
- Assignee and labels in the right column of the PR

avasilev-sap:
> A couple of things are missing:
> 
> * Changeset for the changed package.
> * Assignee and labels in the right column of the PR

Fixed in fe1cdf60aaea4133f6ee5fb843d4d9f8dd3fdf43

packages/generator-adp/src/app/questions/configuration.ts
```diff
@@ -1,41 +1,43 @@
...
     isSyncLoadedView,
     isV4Application,
-    getRelevantVersions,
-    fetchPublicVersions,
-    checkSystemVersionPattern,
-    getAch,
-    getFioriId
+    loadApps,
+    SourceManifest
 } from '@sap-ux/adp-tooling';
```
nikmace, line 26: I can't quite understand the reason for changing all of the imports in all of the files you have. It seems as if the AI/editor is doing these changes, and I see them in every PR.  I believe these changes should not be implemented for the following reasons:

- The actual code contribution in this instance is minimal, while the number of modified lines is significantly higher. This cluttering of the PR provides no tangible benefit, as I have not observed any. I could comprehend these changes if the intention was to enhance code aesthetics and structure imports in a specific order, such as by length or package name.
- The review process becomes more challenging due to the increased size of the diff, making it difficult to pinpoint the actual changes.

Could you please clarify if you are employing a formatter that is responsible for these modifications?

.changeset/spicy-flies-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/generator-adp': minor
+---
+
+fix: [ADP][generator] In case the base app inbounds api call fails for the selected application we display the error next to the application prompt.
```
nikmace: I noticed a couple of things about the commit/PR metadata:
- We’ve talked before about code style and commit conventions for this repo.
- For the scope: instead of `fix: [ADP][generator]`, it should be `fix(adp-generator):`.
Using the scope is optional, but when you do add it, please follow the conventional format.
- Could you clarify whether you’ve seen other PRs in this repo using the `[ADP][…]` style? I want to make sure we stay consistent.

👉 The **PR title** and the **changeset message** should be updated to match this convention before merging.

packages/generator-adp/src/app/questions/configuration.ts
```diff
@@ -492,24 +506,29 @@ export class ConfigPrompter {
             return t('error.selectCannotBeEmptyError', { value: 'Application' });
         }
 
-        const validationResult = await this.validateAppData(app);
-
-        if (!isAppStudio()) {
-            return validationResult;
-        }
+        let validationAppDataResult = await this.validateAppData(app);
```
nikmace: Not sure that making this `validationResult` variable longer is an improvement.

packages/generator-adp/src/app/questions/configuration.ts
```diff
@@ -492,24 +506,29 @@ export class ConfigPrompter {
...
+            try {
+                this.baseApplicationInbounds = await getBaseAppInbounds(app.id, this.provider);
+                return true;
+            } catch (error) {
+                return error.message;
+            }
         }
 
-        return validationResult;
+        return validationAppDataResult;
```
nikmace: If you take into account the comment above:
- `isKnownUnsupported` is string || string, why do you need to cast it?
- It would be nice to add a generic message for the catch block so we know exactly what request failed (i.e. `Fetching base application inbounds failed: {e.message}`).

Thinking about this, if an application is unsupported, why do we need to fetch the base app inbounds? The user cannot continue with creating an ADP; only an extension project. Then, you can return true in this block because the logic works only for an extension project.
```ts
const validationResult = await this.validateAppData(app);

const isKnownUnsupported =
            validationResult === t('error.appDoesNotSupportManifest') ||
            validationResult === t('error.appDoesNotSupportFlexibility');

if (isAppStudio() && isKnownUnsupported) {
    this.logger.error(validationResult);
    this.appValidationErrorMessage = validationResult;
    this.isApplicationSupported = false;
    return true; // continue to next prompt for extension project
} 

if (typeof validationResult === 'string') {
    // Any other validation error -> stop.
    return validationResult;
}

if (this.isCloud) {
    try {
         this.baseApplicationInbounds = await getBaseAppInbounds(app.id, this.provider);
    } catch (e) {
         return t('error.fetchBaseInboundsFailed', { error: e.message });
    }
}

return true;
```
In the end, the only valid scenario where you can fetch inbounds is:
- System is **Cloud**
- `validationResult` is `true`
- `isApplicationSupported` is `true`

cc @mmilko01

packages/generator-adp/test/unit/questions/configuration.test.ts
```diff
@@ -370,6 +380,53 @@ describe('ConfigPrompter Integration Tests', () => {
...
+
+            expect(result).toEqual(true);
+            expect(configPrompter.isCloud).toBe(true);
+            expect(configPrompter.baseAppInbounds).toEqual(baseAppInbounds);
+
+            expect(getBaseAppInboundsMock).toHaveBeenCalledWith(app.id, provider);
+        });
+
+        it('cloud application prompt validate should return error message if base app inbounds api call fails', async () => {
+            const baseAppInboundsError = new Error('Fail to load app inbounds');
```
nikmace: `Fail to load app inbounds` -> `Failed to load app inbounds`

packages/generator-adp/src/app/questions/configuration.ts
avasilev-sap: We discussed this issue with imports offline will leave this as is for this change will fix in further change.

.changeset/spicy-flies-divide.md
avasilev-sap: Fixed in 449e0166885ec939db255fcd47a0771810fa6db4

packages/generator-adp/src/app/questions/configuration.ts
avasilev-sap: Fixed in 449e0166885ec939db255fcd47a0771810fa6db4

packages/generator-adp/src/app/questions/configuration.ts
avasilev-sap: Fixed in 449e0166885ec939db255fcd47a0771810fa6db4

packages/generator-adp/test/unit/questions/configuration.test.ts
avasilev-sap: Fixed in 449e0166885ec939db255fcd47a0771810fa6db4

lfindlaysap (approved):

mmilko01 (approved):
Changeset OK
Changes look good
Excellent test coverage
Did not test manually

PR #3630, opened 2025-09-16
fix: align mta and package.json changes

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/3399                   
                                                                                                                                                                                                                                                                                                                            
  - Align UAA service during MTA generation                                  
  - During MTA generation, the cds binary executes and generates CAP configuration files; additional attributes are added to package.json which were being lost during the final writing phase. A deep merge is now used to keep disk and in-memory package.json in sync (replacing the previous lodash-based approach.


packages/cf-deploy-config-writer/test/unit/__snapshots__/cap.test.ts.snap
```diff
@@ -277,16 +277,16 @@ builder:
   },
   "package.json": Object {
     "contents": "{
-    \\"cds\\": {},
```
longieirl, line 280: Nothng has changed, just formatting changes since its now using a different writer method to update the package.json.

PR #3632, opened 2025-09-17
fix: message not correctly updated in pagemap

Vladislavs-Silins-SAP:
Enhanced the Project type with the capCustomPaths property to store information about custom paths in CAP projects. These custom paths should be considered during performance optimization checks.


packages/project-access/src/project/info.ts
```diff
@@ -32,13 +32,15 @@ export async function getProject(root: string, memFs?: Editor): Promise<Project>
         throw new Error(`The project root folder '${root}' is not a Fiori project. No 'package.json' found.`);
     }
     const capProjectType = await getCapProjectType(root);
+    const capCustomPaths = await getCapCustomPaths(root);
```
815are: should we call/return always  even if it is EDMX project?
we can see it in updated test:
```
'root': projectRoot,
            'projectType': 'EDMXBackend',
            'apps': {}
            'apps': {},
            'capCustomPaths': {
                'app': 'app/',
                'db': 'db/',
                'srv': 'srv/'
            }
```
I think w eshould not return `capCustomPaths` for EDMX project, please correct if you think differently

Klaus-Keller: No, we should not add this for EDMX projects, @Vladislavs-Silins-SAP could you please make this conditional?

Vladislavs-Silins-SAP: Fixed.

815are (approved):
checked:
1. code changes straightforward - comment is addressed
2. test extended to check new property for cap project
3. changelog persists

I did not test manually

Klaus-Keller (approved):
Thanks @Vladislavs-Silins-SAP!
* code to add cap project folders to to project info looks good
* changeset exists
* code coverage is great

Approved from my side.

Issue #3633, opened 2025-09-17
BUG - (VSCode)(ADP Wizard) - SAP Fiori Launchpad Configuration step appears multiple times


avasilev-sap:
### Description
When the Project name is edited (with selected FLP Config), for every Project Name new step is added on the left (on selecting Add SAP Fiori Launchpad Configuration to `No` these don't disappear).

### Steps to Reproduce
1. Start the adaptation project generator.
2. Select a cloud system.
3. From the second page(attributes page) select `Yes` radio button in the `Add SAP Fiori Launchpad Configuration` group.
4. Enter different names inside the project name input.

### Expected results
Fiori Launchpad Configuration step should not be multiplied for each new project name. We only have to see one page with the latest project name.

### Actual results
For each project name we entered, a separate step is added flooding the breadcrumb with wizard steps.

### Screenshots

<img width="1850" height="1522" alt="Image" src="https://github.com/user-attachments/assets/997460fb-df5e-487f-8236-379a60d58c65" />

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3635, opened 2025-09-17
fix(adp)(vscode): SAP Fiori Launchpad Configuration step appears multiple times.

avasilev-sap:
#3633 

There is a drawback in the way we manage wizard pages. When we add/remove pages we filter existing pages by name which is wrong. The name is localized string which can have parameters, practically makes the name dynamic value. In our case when the user changes the project name, the `SAP Fiori Launchpad Configuration` step receives a new name so the previous page with the older name is treated as entirely different page. As a result for every new name we enter we end up with new page.

Add a notion for a stable id for a wizard page. When we add/remove pages we filter pages by the id not the name. The convention for the id is `<packageName>:<localPageId>`. The packageName is included to avoid collisions since a wizard can have pages from different packages with same localPageIds.

steps

Testing done:
* Unit tests.
* Manually verified.

dynamic-page

### DIFF
--- a/.changeset/thick-yaks-juggle.md
+++ b/.changeset/thick-yaks-juggle.md
@@ -1,0 +1,7 @@
+---
+'@sap-ux/adp-flp-config-sub-generator': patch
+'@sap-ux/generator-adp': patch
+'@sap-ux/adp-tooling': patch
+---
+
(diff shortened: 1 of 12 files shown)


avasilev-sap:
Discard this PR as not needed we will put dynamic content inside page description.

Issue #3634, opened 2025-09-17
BUG - Wrong npm deploy scripts are created for ADP onPremise projects


nikmace:
### Description
When using the Deployment Configuration generator for an ADP on-premise project, the wrong scripts are generated, which are causing issues during deployment.

There is now a build script, which is not needed. In this case, only the webapp folder should be directly deployed without any build steps:

Wrong Scripts:
```ts
"build": "ui5 build --exclude-task generateFlexChangesBundle generateComponentPreload minify --clean-dest", <-- not needed for Javascript
"deploy": "npm run build && fiori deploy --config ui5-deploy.yaml", <-- for JavaScript, no build is needed
"undeploy": "npm run build && fiori undeploy --config ui5-deploy.yaml", <-- does not need an build for undeploy
```

### Steps to Reproduce
1. Create a Deployment Config for ADP project
2. Open package.json and see wrong scripts

### Expected results
Scripts are adapted for project type CloudReady/OnPremise and whether it is JS or TS.

### Actual results
Scripts are causing deployment errors

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3637, opened 2025-09-18
fix: Wrong npm deploy scripts are created for ADP onPremise projects

nikmace:
Fix for #3634.
- Removes deploy scripts from ADP OnPremise projects; all other projects' scripts stay untouched.
- Added new tests for testing the deploy config and package.json scripts.


longieirl (dismissed):
Changeset is good.
Changes to cover tests are good.
Reason for change is evident.
Changes look good.

packages/abap-deploy-config-writer/src/scripts.ts
```diff
@@ -0,0 +1,62 @@
...
+        // undeploy script
+        const undeployScript = `${UNDEPLOY_SCRIPT} --config ${deployConfigFile}`;
+        await updatePackageScript(basePath, 'undeploy', undeployScript, fs);
+        // test mode script
+        const deployTestModeScript = `${DEPLOY_SCRIPT} --config ${deployConfigFile} --testMode true`;
+        await updatePackageScript(basePath, 'deploy-test', deployTestModeScript, fs);
+
+        // dependencies
+        await addPackageDevDependency(basePath, RIMRAF, RIMRAF_VERSION, fs);
+    }
```
mmilko01: I think the code will look a bit better if the `else` statement is removed and we have a `return` in the `if`

```suggestion
    if (shouldAddBuildScript) {
        await updateScripts(basePath, deployConfigFile, fs);
        return;
    }
        // deploy script
        const deployScript = `${DEPLOY_SCRIPT} --config ${deployConfigFile}`;
        await updatePackageScript(basePath, 'deploy', deployScript, fs);
        // undeploy script
        const undeployScript = `${UNDEPLOY_SCRIPT} --config ${deployConfigFile}`;
        await updatePackageScript(basePath, 'undeploy', undeployScript, fs);
        // test mode script
        const deployTestModeScript = `${DEPLOY_SCRIPT} --config ${deployConfigFile} --testMode true`;
        await updatePackageScript(basePath, 'deploy-test', deployTestModeScript, fs);

        // dependencies
        await addPackageDevDependency(basePath, RIMRAF, RIMRAF_VERSION, fs);
```

packages/abap-deploy-config-writer/src/file.ts
```diff
@@ -117,3 +118,27 @@ export async function writeUi5RepositoryIgnore(fs: Editor, path?: string): Promi
...
+ * Checks if the project is a TypeScript project.
+ *
+ * @param fs - the memfs editor instance
+ * @param basePath - the base path
+ * @returns true if the project is a TypeScript project, false otherwise
+ */
+export function isTsProject(fs: Editor, basePath: string): boolean {
+    const tsconfigPath = join(basePath, FileName.Tsconfig);
+    return fs.exists(tsconfigPath) || existsSync(tsconfigPath);
+}
```
mmilko01, line 132: Is it possible to reuse the method from `@sap-ux/adp-tooling` instead? https://github.com/SAP/open-ux-tools/blob/ec605962576a185d9a375c3ade07b772f671d5a9/packages/adp-tooling/src/base/helper.ts#L52C1-L62C2

packages/abap-deploy-config-writer/src/file.ts
```diff
@@ -117,3 +118,27 @@ export async function writeUi5RepositoryIgnore(fs: Editor, path?: string): Promi
...
+ * Checks if the project is an ADP project.
+ *
+ * @param fs - the memfs editor instance
+ * @param basePath - the base path
+ * @returns true if the project is an ADP project, false otherwise
+ */
+export function isAdpProject(fs: Editor, basePath: string): boolean {
+    const variantPath = join(basePath, 'webapp', FileName.ManifestAppDescrVar);
+    return fs.exists(variantPath) || existsSync(variantPath);
+}
```
mmilko01: Is it possible to have this method defined in the helpers of `@sap-ux/adp-tooling` and have it as a reusable method?

packages/abap-deploy-config-writer/src/file.ts
nikmace: First, we don't have a dependency to `adp-tooling` in this package and I think people wouldn't agree for us to add it here. I also think, that for now it does not make sense to couple it to `adp-tooling`.

Second, I'll have to call this method twice like this:
```ts
isTypescriptSupported(..., fs) || isTypescriptSupported(...)
```
We don't know whether we are running as standalone or part of a generator, so I need to check both cases.

packages/abap-deploy-config-writer/src/file.ts
nikmace: >First, we don't have a dependency to adp-tooling in this package and I think people wouldn't agree for us to add it here. I also think, that for now it does not make sense to couple it to adp-tooling.

Same as above. @longieirl do you mind if we add a dependency to `@sap-ux/adp-tooling` in this package: `abap-deploy-config-writer`?

packages/abap-deploy-config-writer/src/scripts.ts
mmilko01: My suggestion just follows the 'Return early' pattern and the idea is to improve readability. It is an optional suggestion, I can approve the PR without this change as well.

nikmace: Changed in: [c7eed39](https://github.com/SAP/open-ux-tools/pull/3637/commits/c7eed394dcffadef9d8403a6299b3456a9fc47ea)

packages/abap-deploy-config-writer/src/file.ts
```diff
@@ -117,3 +118,27 @@ export async function writeUi5RepositoryIgnore(fs: Editor, path?: string): Promi
...
 }
+
+/**
+ * Checks if the project is an ADP project.
+ *
+ * @param fs - the memfs editor instance
+ * @param basePath - the base path
+ * @returns true if the project is an ADP project, false otherwise
+ */
+export function isAdpProject(fs: Editor, basePath: string): boolean {
```
cianmSAP: Is this information not determined in the deploy config generator?

cianmSAP (changes requested):
Can we add an option to the writer to add the build script and determine this in the consumer?

We would prefer to keep the writer generator agnostic. `updateScripts` and `updateScriptsForAdp` seems like there is room for consolidation rather than two functions

packages/abap-deploy-config-writer/src/file.ts
nikmace: We need to determine if the project is ADP, JS, and OnPremise - only then will we not add the build script. I don't think now that we even need to know which project it is, and we can remove isAdpProject method from here.

Screenshot 2025-09-19 at 12 30 40

nikmace: Here are the changes: [4fb5466](https://github.com/SAP/open-ux-tools/pull/3637/commits/4fb546624a8ef8bcd08a5cbd14d06e1f902c2723)

mmilko01 (approved):
Changeset OK
Changes look good
Excellent test coverage
Did not test manually

packages/abap-deploy-config-writer/src/file.ts
cianmSAP: Thanks @nikmace

Issue #3638, opened 2025-09-18
BUG - Deploy Config generator should not validate the package for OnPremise systems for ADP


nikmace:
### Description
The Deploy Config generator should not validate the package for OnPremise systems for ADP. Due to the unavailability of a certain API that checks the package project types on older ABAP systems, this check API call needs to be removed for OnPremise systems.

### Steps to Reproduce
1. Open Deploy Config generator for an ADP
2. Choose an OnPremise system and enter your package
3. See error under the package input prompt
4. Open output channel for the generator and see the logs with the failing request for system_info (404 not found)
NOTE: This issue is only reproducible on an older ABAP system.

### Expected results
Package validation is skipped for OnPremise systems

### Actual results
Error is shown and the user cannot continue.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3639, opened 2025-09-18
fix: Deploy Config generator should not validate the package for OnPremise systems for ADP

nikmace:
Fix for #3638.
- Returns true when validating a package for an OnPremise system. Extra validation now happens only for packages when system is CloudReady.


packages/abap-deploy-config-inquirer/src/prompts/validators.ts
```diff
@@ -647,14 +647,10 @@ export function validateConfirmQuestion(overwrite: boolean): boolean {
  */
 async function validatePackageType(input: string, backendTarget?: BackendTarget): Promise<boolean | string> {
     const isS4HC = PromptState?.abapDeployConfig?.isS4HC;
-    if (isS4HC === false && input === DEFAULT_PACKAGE_ABAP) {
+    if (!isS4HC) {
```
longieirl, line 650: This is not a good fit for this flow; just becuse its not an S4HC system, doesnt mean its onpremise. The only way to know is the destination exposed, has a proxy type of `Onpremise`. There is a method in btp-utils called `isOnPremiseDestination` that shows the logic.
Infact, the isS4HC needs to be updated to be more specific since ABAP Cloud envs using SAMLAssertion for cross subaccount communication will be tagged incorrectly. 
The best option is to look for the ProxyType.

nikmace: `isS4HC` property in our case is based on a call of `isAbapCloud` method of axios-extension, which we all currently rely on. Therefore, I think it's okay for now to use it. I know that `isS4HC` property name is not really correct, and should be changed to something like `isCloud`. We relied on this property before, as you can see 
```ts
if (isS4HC === false && input === DEFAULT_PACKAGE_ABAP) {
```

longieirl: I'm wonderig if its better then to introduce a new type isAbapCloud to address this? Soo then we know exactly what we are dealing with? 
For example, the method `updateDestinationPromptState` also sets isS4HC could set an incorrect value.

longieirl: I created the following bug as this has happened a few times, where a system is tagged as s4hc.

https://github.com/SAP/open-ux-tools/issues/3641

nikmace: @longieirl but isn't Iain O'Farrell working on this already, to introduce a new type?

nikmace: In both cases, we are using `isAbapCloud` to determine the system type. We are passing `shouldRestrictDifferentSystemType` in prompt options for ADP projects. So I think in our cases we are determining it correctly. I am sure the issue you pointed out is related to this PR. But I agree that this is an issue and some customers could have Cloud prompts shown for their OnPremise system which is incorrect.

Screenshot 2025-09-18 at 17 38 28

mmilko01: @longieirl I'm not sure if I understand completely the points you raised, but I think that determining the system type from the Destination configuration is not reliable. For example, we have internal setups where the Destination is of type `OnPremise`, but the system is Cloud. This is why we think that is best to send a request to the system to determine the type.

mmilko01 (approved):
Changeset OK
Changes look good
Excellent test coverage
Did not test manually

packages/abap-deploy-config-inquirer/src/prompts/validators.ts
longieirl: @mmilko01 can you give me an example of where the system is onPremise but is a cloud system? The only example I can think of is, where its a private cloud customer where they are using an on-premise connection to access it.

longieirl: Based on offline discussions, happy to accept change.

PR #3640, opened 2025-09-18
fix(fiori-annotation-api): annotation merging

voicis:
In CAP CDS annotations with shortcut syntax can occur before actual values, but this was not correctly handled when merging, because we expected that all annotations should have values which is not correct.

- Fixed annotation value/collection/record not being merged if it was not the first reference for that target/term/qualifier in that file 
- Fixed issue where origins were not correctly updated when merging annotations
- Replaced `console.log` with proper logger calls


marufrasully (approved):
@voicis thanks
* code changes looks good
* secured with unit test
* did not test it locally

PR #3642, opened 2025-09-18
Fix issues with publish embeddings

devinea:
Refactored embeddings to reduce table size and update mcp server to read DB tables.


devinea:
Merging to fix main build

PR #3658, opened 2025-09-22
fpm-writer - Update ControllerExtension.d.ts for easier usage

devinea:


packages/fe-fpm-writer/templates/controller-extension/ControllerExtension.d.ts
```diff
@@ -1,11 +1,38 @@
...
+  import Paginator from 'sap/fe/core/controllerextensions/Paginator';
+  import Recommendations from 'sap/fe/core/controllerextensions/Recommendations';
+  import Routing from 'sap/fe/core/controllerextensions/Routing';
+  import Share from 'sap/fe/core/controllerextensions/Share';
+  import ViewState from 'sap/fe/core/controllerextensions/ViewState';
+  import CoreElement from 'sap/ui/core/Element';
+ 
+  export default class ControllerExtension<ExtensionAPI> {
+    override?: Overrides;
+    base: {
```
heimwege, line 20: Actually I use
```ts
import type PageController from "sap/fe/core/PageController";
//[...]
protected base!: PageController;
```
as type in [my controller extension](https://github.com/heimwege/cap-node-fe-ts-sample/blob/main/app/samples/webapp/ext/controller/CommentsSection.controller.ts#L25). I don't think you need to list all [PageController](https://ui5.sap.com/#/api/sap.fe.core.PageController) functions if you just re-use that type. And it also comes with less maintenance in case the PageController type changes later on.

packages/fe-fpm-writer/templates/controller-extension/ControllerExtension.d.ts
```diff
@@ -1,11 +1,38 @@
...
+      getExtensionAPI(): ExtensionAPI;
+      getView(): View;
+      getAppComponent(): AppComponent;
+      byId(id: string): CoreElement
+    };
+  }
+ 
+  export type Overrides<GenericController extends Controller = Controller> = {
```
heimwege, line 28: For the overrides I use 

```ts
import ViewState from "sap/fe/core/controllerextensions/ViewState";
import EditFlow from "sap/fe/core/controllerextensions/EditFlow";
import Routing from "sap/fe/core/controllerextensions/Routing";
import Controller from "sap/ui/core/mvc/Controller";

type ControllerExtensionOverrides = Partial<Pick<typeof Controller["prototype"], "onAfterRendering" | "onBeforeRendering" | "onExit" | "onInit">>;

/**
 * Types for Fiori Elements Controller Extension Overrides (sap.ui.core.mvc.ControllerExtension.override).
 */
type FioriElementsControllerExtensionOverrides = ControllerExtensionOverrides & {
    /**
     * experimental
     */
    onPageReady?: () => void;
    editFlow?: Partial<Pick<typeof EditFlow["prototype"], "onAfterCreate" | "onAfterDelete" | "onAfterDiscard" | "onAfterEdit" | "onAfterSave" | "onBeforeCreate" | "onBeforeDelete" | "onBeforeDiscard" | "onBeforeEdit" | "onBeforeSave">>;
    routing?: Partial<Pick<typeof Routing["prototype"], "onAfterBinding" | "onBeforeBinding" | "onBeforeNavigation">>;
    viewState?: Partial<Pick<typeof ViewState["prototype"], "adaptBindingRefreshControls" | "adaptBindingRefreshHandler" | "adaptControlStateHandler" |"adaptStateControls" | "applyAdditionalStates" | "applyInitialStateOnly" | "applyNavigationParameters" | "onAfterStateApplied" | "onBeforeStateApplied" | "retrieveAdditionalStates">>;
};
```

in [my controller extension](https://github.com/heimwege/cap-node-fe-ts-sample/blob/main/app/samples/webapp/ext/controller/CommentsSection.controller.ts#L30).
My definition might be incomplete though. But I think just having every property optional actually is not the correct type for overrides because you cannot override everything from that type. So code completion gives you wrong suggestions. Downside is of course that newly introduced properties are missing in the type definition. 
To overcome this nowadays I probably would introduce a utility type to filter on e.g. event handers that start with `on` and pick those.

```ts
/**
 * A utility type that extracts the keys of an object `T` that correspond to methods
 * starting with the prefix "on".
 */
type OnMethodKeys<T> = {
  [K in keyof T]: K extends `on${string}` // 1. If the key starts with "on"...
    ? T[K] extends (...args: any[]) => any // 2. ...and the property is a function...
      ? K // ...keep the key.
      : never
    : never; // 3. Discard all other keys.
}[keyof T]; // 4. Collect the kept keys into a union type.

//[...]

editFlow?: Partial<Pick<typeof EditFlow["prototype"], OnMethodKeys<typeof EditFlow["prototype"]>>>;
```

@MarcelWaechter But actually I would prefer if Fiori elements would provide those types (UI5 version dependent). This would be a great feature, though I understand the JSDoc-based type generation is a hurdle 🙈 

(Please note that `onPageReady` is experimental and should maybe not be part in an official type definition.)

heimwege (commented):
See comments. Coming with proper types out-of-the box would really be a great benefit for TS based controller extensions. So thanks for this PR 👍🏻

packages/fe-fpm-writer/templates/controller-extension/ControllerExtension.d.ts
MarcelWaechter: yeah onPageReady is not a public one. Recommendations also not yet public.

EditFlow, IntentBasedNavigation, MessageHandler, Paginator, Routing, Share and ViewState is fine. 

Yeah let's see how we can improve as well, also adding @nlunets .
Now as we want to provide the Fiori Dev Portal examples with TypeScript, we have the same pain 😆

heimwege: I further refined this. What could be handy is something like:

```ts
import ViewState from "sap/fe/core/controllerextensions/ViewState";
import EditFlow from "sap/fe/core/controllerextensions/EditFlow";
import Routing from "sap/fe/core/controllerextensions/Routing";
import Controller from "sap/ui/core/mvc/Controller";

/**
 * A utility type that extracts the keys of an object `T` that correspond to methods
 * starting with the specified prefix `P`.
 * 
 * @template T - The object type to extract method keys from
 * @template P - The string prefix to match (e.g., "on", "adapt", "apply", "retrieve")
 */
type MethodKeys<T, P extends string> = {
  [K in keyof T]: K extends `${P}${string}` // 1. If the key starts with the prefix...
    ? T[K] extends (...args: any[]) => any // 2. ...and the property is a function...
      ? K // ...keep the key.
      : never
    : never; // 3. Discard all other keys.
}[keyof T]; // 4. Collect the kept keys into a union type.

/**
 * A utility type that combines method keys with multiple prefixes.
 * 
 * @template T - The object type to extract method keys from
 * @template P - A union of string prefixes to match
 */
type MultiPrefixMethodKeys<T, P extends string> = P extends string 
  ? MethodKeys<T, P> 
  : never;

/**
 * Types for UI5 Controller Extension Overrides (sap.ui.core.mvc.ControllerExtension.override).
 */
type ControllerExtensionOverrides = Partial<Pick<typeof Controller["prototype"], MethodKeys<typeof Controller["prototype"], "on">>>;

/**
 * Types for Fiori Elements Controller Extension Overrides (sap.ui.core.mvc.ControllerExtension.override).
 */
type FioriElementsControllerExtensionOverrides = ControllerExtensionOverrides & {
    /**
     * experimental
     */
    onPageReady?: () => void;
    editFlow?: Partial<Pick<typeof EditFlow["prototype"], MethodKeys<typeof EditFlow["prototype"], "on">>>;
    routing?: Partial<Pick<typeof Routing["prototype"], MethodKeys<typeof Routing["prototype"], "on">>>;
    viewState?: Partial<Pick<typeof ViewState["prototype"], MultiPrefixMethodKeys<typeof ViewState["prototype"], "on" | "adapt" | "apply" | "retrieve">>>;
};
```

If Fiori elements keeps following those function naming pattern this should be fine.

packages/fe-fpm-writer/templates/controller-extension/ControllerExtension.d.ts
MarcelWaechter: Another thing that came into my mind. Actually **base** can be the actual controller that is extended (ListReportController or ObjectPageController), that itself returns the respective ExtensionAPI (LR or OP ExtensionAPI), and some additional APIs.

https://ui5.sap.com/#/api/sap.fe.templates

The ObjectPageController is missing from our public API, but I'd say we should make it public as well (as it can also be extended already) - any objections @nlunets ?

packages/fe-fpm-writer/templates/controller-extension/ControllerExtension.d.ts
MarcelWaechter: not sure, the fact a method starts with on doesn't say it's public and can be overridden. That's defined in the decorators, like

@publicExtension()
	@extensible("AfterAsync")
	async onBeforeSave(_mParameters?: { context?: Context }): Promise<void> {
		// to be overridden
		return Promise.resolve();
	}

I guess you can't read our decorators... but it's there in the metadata, just checked quickly

image

but yes not great, it shall be ideally accessible through the TS types...

PR #3664, opened 2025-09-23
Add instructions for VS Code global MCP server file

KevinRiedelsheimer:
This PR adds instructions for the configuration of the MCP server within the `mcp.json` file which can be used as global MCP server list within Visual Studio Code.

### DIFF
--- a/packages/fiori-mcp-server/README.md
+++ b/packages/fiori-mcp-server/README.md
@@ -53,6 +53,26 @@ Then, configure your MCP client to start the server with command `fiori-mcp`. He
 }
 ```
 
+You can create a global configuration [mcp.json](https://code.visualstudio.com/docs/copilot/customization/mcp-servers) within Visual Studio Code:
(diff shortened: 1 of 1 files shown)


packages/fiori-mcp-server/README.md
```diff
@@ -53,6 +53,26 @@ Then, configure your MCP client to start the server with command `fiori-mcp`. He
 }
 ```
 
+You can create a global configuration [mcp.json](https://code.visualstudio.com/docs/copilot/customization/mcp-servers) within Visual Studio Code:
+GitHub Copilot uses the `mcp.json` file as source for it's Agent mode.
+```json
```
heimwege, line 58: Hey @KevinRiedelsheimer, thanks for the contribution.
What is the difference between the snippet you want to add and already existing one [here](https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-mcp-server/README.md#method-1-npx)?
I'm not sure if it really adds value just because what is called `servers` in Cline is called `mcpServers` in VSC.

PR #3665, opened 2025-09-24
feat(fpm-writer): table building block custom columns

docirl:
Internal issue 35222

- adds support for custom columns to be added to table building blocks


packages/fe-fpm-writer/src/building-block/index.ts
```diff
@@ -53,13 +59,22 @@ export async function generateBuildingBlock<T extends BuildingBlock>(
     }
 
     // Read the view xml and template files and update contents of the view xml file
+    let updatedAggregationPath;
     const xmlDocument = getUI5XmlDocument(basePath, viewOrFragmentPath, fs);
+    if (config.buildingBlockData?.buildingBlockType === BuildingBlockType.CustomColumn) {
+        // check xmlDocument for macrosTable element
+        const hasTableColumn = xmlDocument.getElementsByTagName('macros:columns').length > 0;
```
815are: please do not use  hardcoded `macros:` - it can differ:
```
<core:FragmentDefinition xmlns:core="sap.ui.core" xmlns="sap.m" xmlns:bb="sap.fe.macros">
	<bb:Tablr />
</core:FragmentDefinition>
```

docirl: thanks, updated

packages/fe-fpm-writer/src/building-block/types.ts
```diff
@@ -409,6 +412,22 @@ export interface Page extends BuildingBlock {
...
+    customColumnFragmentPath: string;
+    customColumnFragmentName: string;
+    appName?: string;
+    hasTableColumns?: boolean;
+    fragmentFile?: string;
+    folder?: string;
+    columnKey?: string;
+    content?: string;
+    width?: string;
+    eventHandler?: true | string | EventHandlerConfiguration;
```
815are: instead duplicating property definition, maybe just extend from existing interface `EventHandler`?:
```
export interface CustomColumn extends BuildingBlock, EventHandler {
```

btw, seems there is also type `CustomFragment`, but not so commonly used as `EventHandler`

packages/fe-fpm-writer/src/building-block/types.ts
```diff
@@ -409,6 +412,22 @@ export interface Page extends BuildingBlock {
     description?: string;
 }
 
+export interface CustomColumn extends BuildingBlock {
+    title: string;
+    customColumnFragmentPath: string;
```
815are: as I understand it is path to directory where to check fragment?
Just little bit unsure about name - by checking property I though I should pass path to fragment file. But by checking code and tests - there should be path to folder where fragment is stored
for example there is already property [viewOrFragmentPath](https://github.com/SAP/open-ux-tools/blob/3ba03a3988a9ef9fe443356dfa210f1ea3b81da0/packages/fe-fpm-writer/src/building-block/types.ts#L457) - this property point to fragment file, not folder -> https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/test/unit/building-block.test.ts#L23C11-L23C31
maybe we can use same namings like for other custom extensions?
In other cases seems property `folder` is used -> https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/common/types.ts#L24

also maybe there should be additional interface which can be reused later by other building block, which will create additional fragment file.

Also maybe fragment related parts should be in separate/resuable interface?
Like:
```
export interface Dummy extends BuildingBlock {
    /**
     * The title attribute/property of Dummy BB
     */
    title?: string;

    /**
     * The description attribute/property of Dummy BB
     */
    description?: string;

    embededFragment?: EmbededFragment;
}
```

not fully sure about naming `EmbededFragment` - just idea to separate building block attribute properties from fragment separation logic

packages/fe-fpm-writer/test/unit/building-block.test.ts
```diff
@@ -838,6 +838,325 @@ describe('Building Blocks', () => {
...
+</mvc:View>`;
+
+            const basePath = join(testAppPath, 'test-custom-column-detection');
+            const customColumnData: CustomColumn = {
+                id: 'testCustomColumn',
+                buildingBlockType: BuildingBlockType.CustomColumn,
+                title: 'CustomColumnTitle',
+                customColumnFragmentPath: 'ext/fragments',
+                customColumnFragmentName: 'CustomColumnTitle.fragment.xml',
+                folder: 'ext/fragments'
```
815are: no sure, why we need both `folder: 'ext/fragments'` and `customColumnFragmentPath: 'ext/fragments',`?

packages/fe-fpm-writer/src/building-block/types.ts
```diff
@@ -409,6 +412,22 @@ export interface Page extends BuildingBlock {
...
+    customColumnFragmentName: string;
+    appName?: string;
```
815are: we do not pass appname through building block or custom extension properties. Path to app is received through generation function, param `basePath` -> https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/src/building-block/index.ts#L48

packages/fe-fpm-writer/src/building-block/types.ts
docirl: I've extended from these, thanks

packages/fe-fpm-writer/src/building-block/types.ts
docirl: I've reviewed this based on the call and updated. Please review again please.

packages/fe-fpm-writer/test/unit/building-block.test.ts
docirl: I've updated the test data based on the code updates and to be more clear what data is passed.

packages/fe-fpm-writer/src/building-block/types.ts
docirl: I've removed appName now

packages/fe-fpm-writer/src/building-block/index.ts
```diff
@@ -100,6 +110,77 @@ export async function generateBuildingBlock<T extends BuildingBlock>(
...
+        );
+        if (!fs.exists(viewPath)) {
+            fs.copyTpl(getTemplatePath('common/Fragment.xml'), viewPath, buildingBlockData.embededFragment);
+        }
+        // check xmlDocument for macrosTable element
+        const xpathSelect = xpath.useNamespaces((xmlDocument.firstChild as any)._nsMap);
+        const hasTableColumn = xpathSelect("//*[local-name()='columns']", xmlDocument);
+        if (hasTableColumn && Array.isArray(hasTableColumn) && hasTableColumn.length > 0) {
+            buildingBlockData.embededFragment.hasTableColumns = true;
+            updatedAggregationPath = aggregationPath + '/macros:columns';
```
815are: I can be different than `macros:` - I think should not be hardcoded.  For example case like:
```
<core:FragmentDefinition xmlns:core="sap.ui.core" xmlns="sap.m" xmlns:bb="sap.fe.macros">
    <VBox>
        <bb:Table id="Table" metaPath="@com.sap.vocabularies.UI.v1.LineItem"/>
    </VBox>
</core:FragmentDefinition>
```
there is function `getOrAddNamespace` to detect prefix for namespace

packages/fe-fpm-writer/templates/building-block/custom-column/View.xml
```diff
@@ -0,0 +1,5 @@
+<% if (!data.embededFragment.hasTableColumns) { %><macros:columns>
+    <% } %><macrosTable:Column header="<%- data.title %>" importance="High" <% if (data.width) { %>width="<%- data.width %>"<% } %>  <% if (data.position.anchor) { %>anchor="<%- data.position.anchor %>"<% } %>  <% if (data.position.placement) { %>placement="<%- data.position.placement %>"<% } %>  key="<%- data.columnKey || 'myColumnKey' %>">
+        <core:Fragment fragmentName="<%- data.embededFragment.ns %>.<%- data.embededFragment.name %>" type="XML" />
+    </macrosTable:Column>
```
815are: It can be different than `macrosTable`, for example:

```
<core:FragmentDefinition xmlns:core="sap.ui.core" xmlns="sap.m" xmlns:bb="sap.fe.macros" xmlns:table="sap.fe.macros.table">
    <VBox>
        <bb:Table id="Table" metaPath="@com.sap.vocabularies.UI.v1.LineItem">
            <table:Column />
        </bb:table>
    </VBox>
</core:FragmentDefinition>
```

we have method `getOrAddNamespace` to retrieve such prefix.
As example also can be used other template like -> https://github.com/SAP/open-ux-tools/blob/main/packages/fe-fpm-writer/templates/building-block/chart/View.xml#L1
```
<<%- macrosNamespace %>:Chart
```

packages/fe-fpm-writer/src/common/types.ts
```diff
@@ -195,6 +195,13 @@ export interface FragmentContentData {
     requireAttribute?: string;
 }
 
+/**
+ * Interface represents table columns data for fragments.
+ */
+export interface FragmentTableColumnsData {
+    hasTableColumns?: boolean;
```
815are: seems unused anymore after adding TemplateConfig?

packages/fe-fpm-writer/src/building-block/types.ts
```diff
@@ -409,6 +424,24 @@ export interface Page extends BuildingBlock {
...
+    width?: string;
+    columnKey?: string;
+    position?: Position;
+    embededFragment?: EmbededFragment;
+}
+
+export interface EmbededFragment
+    extends EventHandler,
+        CustomFragment,
+        InternalCustomElement,
```
815are: could you please remove `InternalCustomElement` from public interface as InternalCustomElement contains internal/private properties.
Could you please check method `setCommonDefaults` and can you reuse it instead of duplicating logic here? -> https://github.com/SAP/open-ux-tools/blob/feat/fpm-writer/building-block-custom-column/packages/fe-fpm-writer/src/building-block/index.ts#L178

packages/fe-fpm-writer/src/common/types.ts
docirl: Thanks, removed

packages/fe-fpm-writer/src/building-block/types.ts
docirl: Updated to remove internal interface and use setCommonDefaults

packages/fe-fpm-writer/src/building-block/types.ts
```diff
@@ -409,6 +417,18 @@ export interface Page extends BuildingBlock {
...
+export interface CustomColumn extends BuildingBlock {
+    title: string;
+    width?: string;
+    columnKey?: string;
+    position?: Position;
+    embededFragment?: EmbededFragment;
+}
+
+export interface EmbededFragment extends EventHandler, CustomFragment, CustomElement, FragmentContentData {
+    customColumnFragmentPath: string;
```
815are: seems after latest changes - `customColumnFragmentPath` is unused, could you please remove it or please point where it is used?

packages/fe-fpm-writer/test/unit/__snapshots__/building-block.test.ts.snap
```diff
@@ -1,5 +1,63 @@
...
+    "contents": "<mvc:View xmlns:core=\\"sap.ui.core\\" xmlns:mvc=\\"sap.ui.core.mvc\\" xmlns=\\"sap.m\\" xmlns:html=\\"http://www.w3.org/1999/xhtml\\" controllerName=\\"com.test.myApp.ext.main.Main\\" xmlns:macros=\\"sap.fe.macros\\" xmlns:macrosTable=\\"sap.fe.macros.table\\">
+    <Page title=\\"Main\\">
+        <content>
+            <macros:Table>
+            </macros:Table>
+            <macros:columns>
+                <macrosTable:Column header=\\"CustomColumnTitle2\\" importance=\\"High\\" placement=\\"After\\" key=\\"myColumnKey\\">
+                    <core:Fragment fragmentName=\\"my.test.App.ext.fragment.CustomColumnTitle2\\" type=\\"XML\\"/>
+                </macrosTable:Column>
+            </macros:columns>
```
815are: aggregationPath in test seems is not correct? example in test seems wrong, I would expect:
```
<macros:Table>
    <macros:columns>
                <macrosTable:Column header=\\"CustomColumnTitle2\\" importance=\\"High\\" placement=\\"After\\" key=\\"myColumnKey\\">
                    <core:Fragment fragmentName=\\"my.test.App.ext.fragment.CustomColumnTitle2\\" type=\\"XML\\"/>
                </macrosTable:Column>
            </macros:columns>
</macros:Table>
          
```

packages/fe-fpm-writer/test/unit/building-block.test.ts
```diff
@@ -838,6 +839,324 @@ describe('Building Blocks', () => {
...
+            // by mocking or using the internal functions directly
+            const { DOMParser } = await import('@xmldom/xmldom');
+            const xmlDocument = new DOMParser().parseFromString(xmlViewWithColumns);
+
+            // Test the getElementsByTagName functionality directly - this is what the code checks
+            const hasTableColumn = xmlDocument.getElementsByTagName('macros:columns').length > 0;
+            expect(hasTableColumn).toBe(true);
+
+            // Test that content property gets set
+            await writeFilesForDebugging(fs);
```
815are, line 919: not fully sure got idea of test - it does not call any fpm-writer function?

packages/fe-fpm-writer/src/building-block/types.ts
815are: cf8bc4d919303f5ae710b8ff258537fbb91eeb5c

packages/fe-fpm-writer/test/unit/__snapshots__/building-block.test.ts.snap
815are: d73b37edae4473ca423b6854c3b37bfdc4a9a509

packages/fe-fpm-writer/test/unit/building-block.test.ts
815are: added some steps to generate custom column bb and check snapshot - 07343be598ceb50b3eaf262ce5ca8b541591e785

packages/fe-fpm-writer/src/building-block/index.ts
```diff
@@ -60,10 +64,17 @@ export async function generateBuildingBlock<T extends BuildingBlock>(
...
     }
 
+    const { path: manifestPath, content: manifest } = await getManifest(basePath, fs);
+
     // Read the view xml and template files and update contents of the view xml file
     const xmlDocument = getUI5XmlDocument(basePath, viewOrFragmentPath, fs);
-    const { content: manifest } = await getManifest(basePath, fs);
-    const templateDocument = getTemplateDocument(buildingBlockData, xmlDocument, fs, manifest);
+    const { updatedAggregationPath, processedBuildingBlockData, hasAggregation, aggregationNamespace } =
+        processBuildingBlock(buildingBlockData, xmlDocument, manifestPath, manifest, aggregationPath, fs);
```
kjose90, line 72: Currently, the function `processBuildingBlock` is called for every building block, regardless of type. However, its logic is only relevant for custom columns with embedded fragments. 

- Why not only call `processBuildingBlock` when the building block is a custom column and has an embedded fragment and process only when required?
- Also maybe rename to `handleCustomColumnBuildingBlock`?

docirl: The function was created to handle additional logic/processing for the building block, and possibly be used by other building blocks in future. Most of the processing in the function for custom column is behind a conditional, but I agree the check for custom column could be hoisted to the function call if this will be not be used by other flows.

PR #3668, opened 2025-09-25
fix: In Cloud ADP project with selected unsupported application we should stop the ADP generator.

avasilev-sap:
…tion prompt must fail

### DIFF
--- a/packages/generator-adp/src/app/questions/configuration.ts
+++ b/packages/generator-adp/src/app/questions/configuration.ts
@@ -516,8 +516,9 @@ export class ConfigPrompter {
             this.logger.error(validationResult);
             this.appValidationErrorMessage = validationResult;
             this.isApplicationSupported = false;
-            // Continue to the next prompt for extension project.
-            return true;
+            // Continue to the next prompt for extension project only if the project
+            // is NOT cloud project.
+            return this.isCloud ? validationResult : true;
         }
 
         if (typeof validationResult === 'string') {
--- a/packages/generator-adp/test/unit/questions/configuration.test.ts
+++ b/packages/generator-adp/test/unit/questions/configuration.test.ts
@@ -480,6 +480,23 @@ describe('ConfigPrompter Integration Tests', () => {
             expect(result).toEqual(t('error.manifestCouldNotBeValidated'));
         });
 
+        it('application prompt validate should return string when  application does NOT support adaptation and the project is cloud', async () => {
+            const error = new Error(t('error.appDoesNotSupportManifest'));
+            getManifestSpy.mockRejectedValue(error);
+            mockIsAppStudio.mockReturnValue(true);
+            configPrompter['isCloudProject'] = true;
+            const prompts = configPrompter.getPrompts();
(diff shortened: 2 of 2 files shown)


nikmace (changes requested):
Things that are currently missing:
- In the right pane of the PR: Assignee (which I've added), Labels
- Changeset
- PR is currently a draft. Please tag it ready for review

avasilev-sap:
> Things that are currently missing:
> 
> * In the right pane of the PR: Assignee (which I've added), Labels
> * Changeset
> * PR is currently a draft. Please tag it ready for review

Sorry i've just waited for the pipeline to complete before making the change ready for review. Change set and labels added.

packages/generator-adp/src/app/questions/configuration.ts
```diff
@@ -516,8 +516,9 @@ export class ConfigPrompter {
             this.logger.error(validationResult);
             this.appValidationErrorMessage = validationResult;
             this.isApplicationSupported = false;
-            // Continue to the next prompt for extension project.
-            return true;
+            // Continue to the next prompt for extension project only if the project
+            // is NOT cloud project.
+            return this.isCloud ? validationResult : true;
         }
```
nikmace, line 521: I think the _easier_ solution would be to check in the `if block` if we are in a cloud scenario. Entering this **extension project flow** `if block` is not correct because you are setting properties that may cause side effects. We should skip this if block as a whole and directly go to the next `if block` where the string will be returned. What do you think?

```suggestion

if (isAppStudio() && isKnownUnsupported && !this.isCloud) {
   this.logger.error(validationResult);
   this.appValidationErrorMessage = validationResult;
   this.isApplicationSupported = false;
   // Continue to the next prompt for extension project
   return true;
}
```

nikmace (changes requested):

packages/generator-adp/src/app/questions/configuration.ts
avasilev-sap: Are you sure about this, it changes the logic a bit? The `appValidationErrorMessage`, `isApplicationSupported` won't be set, i tested manually and it works - the validation message is correct and the next button is disabled.

nikmace: As discussed: `appValidationErrorMessage` and `isApplicationSupported` are needed for the extension prompt. In our case, we will return the error message for the application prompt. Thus, other prompts will not be shown and the `next` button will be disabled. And we will also not set `appValidationErrorMessage` and `isApplicationSupported`, which might cause side-effect or problems with the tests in the future.

avasilev-sap: fixed in e995a497efacb1d1d3d7b3cf9e6ece20144bf926

nikmace (approved):
Changes look good
Coverage is good, functionality is covered by a test
Did not test manually, but I see screenshots from testing attached
Changeset looks good

mmilko01 (approved):
Changeset OK
Change looks good
Excellent test coverage
Did not test manually

PR #3669, opened 2025-09-25
fix(ui5-proxy-middleware): support pathRewrite for UI5 resources 

ytpo-lyne:
Supports the pathReplace over the ui5-proxy-middleware in order to allow ui5.yaml configurations to load UI5 from a different resource path that is not e.g. `/1.96.25/resources` by default.


packages/ui5-proxy-middleware/test/base/proxy.test.ts
```diff
@@ -166,6 +167,24 @@ describe('proxy', () => {
...
+            url: 'https://example.example',
+            version: '1.0.0'
+        };
+
+        ui5Proxy(config);
+        const proxyConfig = createProxyMiddlewareSpy.mock.calls[0][1];
+        if (typeof proxyConfig?.pathRewrite === 'function') {
+            proxyConfig?.pathRewrite(config.path, {} as any);
+            expect(getPathPeplaceSpy).toHaveBeenCalledTimes(1);
+            expect(getPathPeplaceSpy).toHaveBeenCalledWith(config.path, {} as any);
```
heimwege: You just thest that re replace is called but not the actual result. We should also test the result to be the expected path

heimwege: Can we avoid using `any`?

packages/ui5-proxy-middleware/README.md
```diff
@@ -5,9 +5,11 @@ The `@sap-ux/ui5-proxy-middleware` is a [Custom UI5 Server Middleware](https://s
 ## Configuration Options
 | Option       | Default Value | Description |
 | ------------ | ------------- | ----------- |
-| `ui5`        | `object`      | List of mount paths and target urls that should be handled by the proxy. If not provided then `/resources` and `/test-resources` are proxied to `https://ui5.sap.com` |
-| `version`    | `undefined`   | The UI5 version. If this property is not defined, then the `minUI5Version` from the `manifest.json` will be used |
-| `secure`     | true          | Defines if SSL certs should be verified |
+| `path`        | `/resources`, `/test-resources`       | Path that is to be proxied |
+| `url`        | `https://ui5.sap.com`      | URL pointing to the resources |
+| `pathReplace`        | `undefined` | If provided then the path will be replaced with this value before forwarding |
```
heimwege: Do we need to mention that this does override the `version` property?

packages/ui5-proxy-middleware/test/base/proxy.test.ts
ytpo-lyne: This is the case. Please check [utils.test.ts](https://github.com/SAP/open-ux-tools/pull/3669/files#diff-d9ce3bd77626043e6af3f89581a93e4f4f63308ef95b8d3400492e2092abd6d7R551). Anyway, test updated: 2ffc1493dc15e737e5a351a71a4c62fa41d84fb2

ytpo-lyne: Sure. Done in 2ffc1493dc15e737e5a351a71a4c62fa41d84fb2

packages/ui5-proxy-middleware/README.md
ytpo-lyne: Are we sure this is the case? Need to first verify.

packages/ui5-proxy-middleware/README.md
```diff
@@ -54,10 +57,27 @@ server:
...
+Use the following syntax for proxying different UI5 requests to different URLs
+
+```Yaml
+server:
+  customMiddleware:
+  - name: fiori-tools-proxy
+    afterMiddleware: compression
+    configuration:
+      ui5:
+        paths:
```
heimwege: As discussed this option does not exist in ui5-proxy-middleware

longieirl (approved):
Changeset is good.
A locally built version consuming these changes was validated on BAS and working as expected.
Tests are good.
Overall changes look good.

PR #3672, opened 2025-09-25
fix(abap deploy config inquirer): 33258 asterisk for transport request

korotkovao:
Related to the internal issue 33258.

Show an asterisk next to the Transport Request (mandatory) input.

image

### DIFF
--- a/packages/abap-deploy-config-inquirer/src/prompts/questions/config/transport.ts
+++ b/packages/abap-deploy-config-inquirer/src/prompts/questions/config/transport.ts
@@ -120,14 +120,11 @@ export function getTransportRequestPrompts(
                 ),
             type: 'input',
             name: promptNames.transportManual,
-            message: () =>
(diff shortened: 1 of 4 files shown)


packages/abap-deploy-config-inquirer/src/prompts/questions/config/transport.ts
```diff
@@ -126,7 +126,8 @@ export function getTransportRequestPrompts(
                     : t('prompts.config.transport.common.transportRequest'),
```
cianmSAP: There was also an ask to remove the `(mandatory)` 
So this condition is no longer needed

cianmSAP (changes requested):

cianmSAP (dismissed):
Thanks for the updates
Much cleaner UI now

cianmSAP (approved):

korotkovao:
Thank you All

PR #3673, opened 2025-09-25
feat: use schema approach for input parameters in functionality details

815are:
Changes:
1. changed interface of output for `get-functionality-details`, `parameters` now is returned as JSON Schema instead of out custom  interface [Parameter[]](https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-mcp-server/src/types/basic.ts#L86)
2. avoiding duplication of defining zod schema and parameters. Like problem -> [zod schema](https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-app/command.ts#L12-L44); same with [parameters](https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-app/generate-fiori-ui-app.ts#L20-L167)
3. all static functionalities are validated with build in zod schema validation -> [validateWithSchema](https://github.com/SAP/open-ux-tools/blob/feat/schemasApproachForFunctionalities/packages/fiori-mcp-server/src/tools/utils.ts#L88)
4. JSON schema is returned based on zod schema -> [convertToSchema](https://github.com/SAP/open-ux-tools/blob/feat/schemasApproachForFunctionalities/packages/fiori-mcp-server/src/tools/utils.ts#L152)

### DIFF
--- a/examples/fe-fpm-cli/CHANGELOG.md
+++ b/examples/fe-fpm-cli/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/fe-fpm-cli
 
+## 0.1.40
+
+### Patch Changes
+
+-   Updated dependencies [99ec164]
+    -   @sap-ux/fe-fpm-writer@0.39.2
+
 ## 0.1.39
 
 ### Patch Changes
--- a/examples/fe-fpm-cli/package.json
+++ b/examples/fe-fpm-cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/fe-fpm-cli",
-    "version": "0.1.39",
+    "version": "0.1.40",
     "description": "A simple CLI to prompt required information to create a building block using the fe-fpm-writer module's prompt and generate functions.",
     "license": "Apache-2.0",
     "private": true,
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.83
+
+### Patch Changes
+
+-   @sap-ux/fiori-elements-writer@2.7.14
+
 ## 1.1.82
 
 ### Patch Changes
--- a/examples/simple-generator/package.json
+++ b/examples/simple-generator/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/generator-simple-fe",
-    "version": "1.1.82",
+    "version": "1.1.83",
     "description": "Simple example of a yeoman generator for Fiori elements.",
     "license": "Apache-2.0",
     "private": true,
--- a/examples/ui-prompting-examples/CHANGELOG.md
+++ b/examples/ui-prompting-examples/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux-private/ui-prompting-examples
 
+## 0.4.1
+
+### Patch Changes
+
+-   99ec164: Remove button group configuration from Rich Text Editor template for initial version
+
 ## 0.4.0
 
 ### Minor Changes
--- a/examples/ui-prompting-examples/package.json
+++ b/examples/ui-prompting-examples/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux-private/ui-prompting-examples",
-    "version": "0.4.0",
+    "version": "0.4.1",
     "description": "This project contains UI storybook stories with exampleS with prompt ui and FPM based building blocks.",
     "license": "Apache-2.0",
     "private": true,
--- a/examples/ui-prompting-examples/src/RichTextEditor.story.tsx
+++ b/examples/ui-prompting-examples/src/RichTextEditor.story.tsx
@@ -19,14 +19,6 @@ export const ExternalValues = (): JSX.Element => {
                     metaPath: {
                         bindingContextType: 'absolute',
                         entitySet: 'C_CustomerBankDetailsOP'
-                    },
-                    buttonGroup: {
-                        name: 'edit-actions',
-                        buttons: 'copy,paste,undo,redo',
-                        visible: false,
-                        priority: 20,
-                        customToolbarPriority: 5,
-                        row: 2
                     }
                 }
             }}
--- a/packages/fe-fpm-writer/CHANGELOG.md
+++ b/packages/fe-fpm-writer/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/fe-fpm-writer
 
+## 0.39.2
+
+### Patch Changes
+
+-   99ec164: Remove button group configuration from Rich Text Editor template for initial version
+
 ## 0.39.1
 
 ### Patch Changes
--- a/packages/fe-fpm-writer/package.json
+++ b/packages/fe-fpm-writer/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/fe-fpm-writer",
     "description": "SAP Fiori elements flexible programming model writer",
-    "version": "0.39.1",
+    "version": "0.39.2",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/fe-fpm-writer/src/building-block/prompts/questions/richTextEditor.ts
+++ b/packages/fe-fpm-writer/src/building-block/prompts/questions/richTextEditor.ts
@@ -6,8 +6,7 @@ import {
     getEntityPrompt,
     getTargetPropertiesPrompt,
     getAggregationPathPrompt,
-    getViewOrFragmentPathPrompt,
-    getBooleanPrompt
+    getViewOrFragmentPathPrompt
 } from '../utils';
 import type { PromptContext, Prompts } from '../../../prompts/types';
 import { BuildingBlockType, bindingContextAbsolute } from '../../types';
@@ -25,13 +24,6 @@ export async function getRichTextEditorBuildingBlockPrompts(
     context: PromptContext
 ): Promise<Prompts<RichTextEditorPromptsAnswer>> {
     const t = translate(i18nNamespaces.buildingBlock, 'prompts.richTextEditor.');
-    const defaultAnswers = {
-        buttonGroup: {
-            name: 'font-style',
-            visible: true,
-            buttons: 'bold,italic,underline'
-        }
-    };
 
     return {
         questions: [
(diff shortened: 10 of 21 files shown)


packages/fiori-mcp-server/src/tools/functionalities/controller-extension/index.ts
```diff
@@ -88,15 +49,16 @@ async function getFunctionalityDetails(input: GetFunctionalityDetailsInput): Pro
...
-        throw new Error('Missing or invalid parameter "controllerName"');
-    }
     if (project?.applicationAccess) {
         const { appId, root } = project;
         const ftfsFileIo = new SapuxFtfsFileIO(project.applicationAccess);
+        const parametersSchema = await buildControllerExtensionSchema(ftfsFileIo);
+
+        const creationParameters = validateWithSchema(parametersSchema, parameters);
+        const { pageId, controllerName, pageType } = creationParameters;
+
```
heimwege, line 61: Is this check/error no longer needed?

```
if (!controllerName || typeof controllerName !== 'string') {
        throw new Error('Missing or invalid parameter "controllerName"');
    }
```

packages/fiori-mcp-server/src/tools/functionalities/controller-extension/schema.ts
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Schema for creating a controller extension.
+ */
+export const ControllerExtensionCreationSchema: zod.ZodObject<{
+    pageType: zod.ZodDefault<zod.ZodEnum>;
+    controllerName: zod.ZodString;
+    pageId?: zod.ZodOptional<zod.ZodEnum>;
+}> = zod.object({
+    pageType: zod.enum(['ListReport', 'ObjectPage']).describe('Type of page').default('ListReport'),
```
heimwege, line 13: off-topic: Do we only support these two floor plans as of now?

packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-app/schema.ts
```diff
@@ -0,0 +1,70 @@
+import { LATEST_UI5_VERSION } from '../../../constant';
+import * as z from 'zod';
+
+export const GeneratorConfigSchemaCAP = z.object({
+    floorplan: z
+        .literal(['FE_LROP', 'FE_FEOP', 'FE_FPM', 'FE_OVP', 'FE_ALP', 'FE_WORKLIST', 'FF_SIMPLE'])
```
heimwege, line 6: off-topic (part 2): because here we offer much more floor plans

packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-app/schema.ts
```diff
@@ -0,0 +1,70 @@
...
+                    ]
+                }),
+            capType: z.optional(z.literal(['Node.js', 'Java']))
+        })
+    }),
+    entityConfig: z.object({
+        mainEntity: z.object({
+            entityName: z
+                .string()
+                .describe("The name of the main entity, e.g. 'SalesOrder'")
```
heimwege: Not sure if we should add examples into the description if the already `provide` examples in the `.meta`.
_Putting examples in two different places can lead to confusion for both developers and automated tools._

packages/fiori-mcp-server/src/tools/functionalities/page/add-page.ts
```diff
@@ -79,6 +74,22 @@ async function executeFunctionality(params: ExecuteFunctionalityInput): Promise<
...
+    const pageType = addPageParameters.pageType;
+    const parentPage = 'parentPage' in addPageParameters ? addPageParameters.parentPage : undefined;
+    const entitySet = 'entitySet' in addPageParameters ? addPageParameters.entitySet?.toString() : undefined;
+    const pageNavigation =
+        'pageNavigation' in addPageParameters ? addPageParameters.pageNavigation.toString() : undefined;
+    const viewName =
+        'pageViewName' in addPageParameters && addPageParameters.pageViewName ? addPageParameters.pageViewName : '';
+
+    if (pageType === PageTypeV4.CustomPage && !viewName) {
+        throw new Error('Missing value for parameter "pageViewName"');
```
heimwege: should we maybe add the information that `pageViewName` is mandatory for type `CustomPage` ? Just wondering if this would be a more helpful response for llm 🤷🏻

packages/fiori-mcp-server/src/tools/functionalities/page/schema.ts
```diff
@@ -0,0 +1,99 @@
...
+const childPageSchema = (page: string, navigations: AllowedNavigationOptions[]) =>
+    zod.object({
+        parentPage: zod
+            .literal(page)
+            .describe(
+                'Parent page is id/name of parent page. First try to extract parent page from user input in a format defined in example, if not possible suggest content defined in options'
+            ),
+        pageNavigation: zod.enum(navigations.map((n) => n.name)),
+        pageType: PageTypeEnum.describe(
+            `Type of page to be created. First try to extract page type from user input in a format defined in example, if not possible suggest content defined in options.`
```
heimwege: [...] _from user input in a format defined in example_ -> which example?

packages/fiori-mcp-server/src/tools/functionalities/page/schema.ts
```diff
@@ -0,0 +1,99 @@
...
+        ),
+        pageViewName: zod
+            .string()
+            .regex(PAGE_VIEW_NAME_PATTERN)
+            .optional()
+            .describe(
+                `Required if pageType is "CustomPage". Name of custom view file. First try to extract view name from user input that satisfies the pattern, if not possible ask user to provide view name`
```
heimwege: General remark because there are several occurrences. Shouldn't we use the [elicitation](https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation) for user input? But might be off-topic for this PR

packages/fiori-mcp-server/src/tools/get-functionality-details.ts
```diff
@@ -42,67 +42,43 @@ export async function getFunctionalityDetails(
...
+            parameters: prepatePropertySchema(property.name, schema)
         };
     } else if (node?.path.length) {
         // Node was found by path - list node properties
-        let parameters: Parameter[] = [];
-        for (const property of node.properties) {
-            parameters = parameters.concat(getParameters([property]));
-        }
+        const schema = resolveRefs(node.schema, rootSchema);
         details = {
```
heimwege: Can we remove the duplication for the `details` object? Having one central definition and then just adding the `parameters` property?

packages/fiori-mcp-server/src/tools/utils.ts
```diff
@@ -71,6 +72,90 @@ export const getDefaultExtensionFolder = (directory: string): string | undefined
     return subFolder;
 };
 
+/**
+ * Recursively resolves `$ref` references in a JSON Schem.
```
heimwege:
```suggestion
 * Recursively resolves `$ref` references in a JSON Schema.
```

heimwege (commented):
see comments

packages/fiori-mcp-server/src/tools/functionalities/controller-extension/index.ts
815are: Now it is part of schema(controller name is required field) and validation fails during schema validation -> https://github.com/SAP/open-ux-tools/blob/feat/schemasApproachForFunctionalities/packages/fiori-mcp-server/src/tools/utils.ts#L185
But there was missing test for it and also was missing regex for file name. I added missing tests and regex pattern within following commit -> 25aee7d2d322f27222482287894820d41e6fd629

here is test with validation message -> https://github.com/SAP/open-ux-tools/blob/feat/schemasApproachForFunctionalities/packages/fiori-mcp-server/test/unit/tools/functionalities/controller-extension/index.test.ts#L204-L213

packages/fiori-mcp-server/src/tools/functionalities/controller-extension/schema.ts
815are: As far as I know, for controller extensions - yes, only LR and OP.
Custom pages has custom code from beginning so ther eis no much benefit for controller extensions.
There is two setup options:
1. controller extension for specific page;
2. controller extension for specific floorplan, and here we support only LR and OP. Same also in page map

packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-app/schema.ts
815are: from my perspective those two different entry points:
1. generator - it uses floorplan to generate different manifest/pages structure?
2. controller extension - floorplan is used to extends `template` class, like following config in manifest.json:
```
"extensions": {
        "sap.ui.controllerExtensions": {
          "sap.fe.templates.ListReport.ListReportController": {
            "controllerName": "javascript.ext.controller.Ddddd"
          }
        }
      }
```

packages/fiori-mcp-server/src/tools/functionalities/page/schema.ts
815are: I agree we should check, but from my perspective that would be a separate topic. This PR mainly focuses on avoiding interface duplication and using json schema (instead of our custom interface) for parameter input

packages/fiori-mcp-server/src/tools/functionalities/page/add-page.ts
815are: @heimwege 
I updated it to include it into schema validation, commit - https://github.com/SAP/open-ux-tools/pull/3673/commits/7fffb0f2b7a5af03fdcc13136efe5b31a1f861e9

packages/fiori-mcp-server/src/tools/get-functionality-details.ts
815are: improvement eda37e24fb2c6a0ff75f9a106cb87d699a343787

packages/fiori-mcp-server/src/tools/utils.ts
815are: fixed, thanks

packages/fiori-mcp-server/src/tools/functionalities/page/schema.ts
815are: adjustment - 39f9d396c12ad4e2b1360511bba125e020667492

packages/fiori-mcp-server/src/tools/functionalities/generate-fiori-ui-app/schema.ts
815are: adjustment 39f9d396c12ad4e2b1360511bba125e020667492

packages/fiori-mcp-server/src/tools/functionalities/controller-extension/schema.ts
heimwege: Ah, now I see this is the controller extension. Sorry for being totally blind 😄

heimwege (dismissed):
- changeset exists
- review comments addressed
- test coverage is great
- did NOT test mahaully

packages/fiori-mcp-server/src/tools/get-functionality-details.ts
```diff
@@ -67,7 +67,8 @@ function getPropertyDetails(
     } else if (node?.path.length) {
         // Node was found by path - list node properties
         schema = node.schema;
-        name = node.path[node.path.length - 1].toString();
+        // Safe: path length > 0 ensures at least one segment
+        name = node.path.at(-1)!.toString();
```
heimwege: Not sure if we should rather do something like `(node.path.at(-1) ?? []).toString()`. Seems to be a bit safer that just insisting on `node.path.at(-1)` to exist.
Why do we know that `node.path.at(-1)` always exist?

815are: there is check for `node?.path.length` - so there must be at least one element in `path`:
```
else if (node?.path.length) {
        // Node was found by path - list node properties
        schema = node.schema;
        // Safe: path length > 0 ensures at least one segment
        name = node.path.at(-1)!.toString();
    }
```

but adjusted code with latest suggestion, but we can not write any test to cover it

815are: I pushed without usage of `!` -> c9eddb580ed9b9a3f4b0879d3d95d9bd8feb79b1

heimwege: oh or is it the if `node?.path.length` but if that's 0 🤔 the if does only ensure it's not undefined, right?

815are: it also ensures it has at least 1 entry, see some example run:
image

815are: @heimwege 
maybe this approach is more readable? -> https://github.com/SAP/open-ux-tools/pull/3673/commits/2dfc5bbc5c3f8d9cf105dbeb8a9ca4b47fd35cca

heimwege (approved):
re-approve

devinea (approved):
Code changes look like a nice approach.
Covered by tests. 
Changset 👍 
I haven't tested locally.

PR #3677, opened 2025-09-26
fix(security): upgrade tar-fs in lock file

donal-tobin-sap:
- https://github.com/SAP/open-ux-tools/security/dependabot/243

Issue: ux-engineering#35838


donal-tobin-sap:

- security fix patch update only in lock file

PR #3681, opened 2025-09-29
fix(ui-service-gen): escape information message

docirl:
Internal issue 35854

- message displayed after generation do not always display correctly, depending on the content (e.g. `/`)
- escape the text when used in i18n, so that are displaying in the YUI toast correctly

image


mikicvi-SAP (approved):
Did not test locally, but escaping translation is done as expected.
Changeset ☑️
LGTM ✅

Issue #3616, opened 2025-09-11
BUG - "Undeploy" does not work with the new ui5-deploy.yaml file for ADP


nikmace:
### Description
The "undeploy" does not work with the new `ui5-deploy.yaml` file because it is missing a property "lrep" in the "configuration" section of the file. 

To make the undeployment go through **LREP**, you need to add a new property “lrep” in the “configuration” section of the `ui5-deploy.yaml` with the `namespace` of your project. You can find this property in the `manifest.appdescr_variant` file of your project. Copy it and paste it next to “lrep: <your_namespace>”.

If you do not add this property, you will see an error similar to this: "Property app-name does not exist".

### Steps to Reproduce
1. Create an ADP with a deployment configuration
2. Deploy the application and then try to undeploy
3. See error: "Property app-name does not exist".

### Expected results
Undeployment is successful through LREP

### Actual results
Undeployment does not even start because an error is thrown for a missing property, which is not needed for On-Premise projects

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3687, opened 2025-10-01
fix: "lrep" property is not added when the deploy config is created from within the ADP Generator

nikmace:
Fix for #3616.
- Adds a check to determine whether we are running as a sub-generator or part of the ADP generator and read from disk or from mem-fs, respectively.

### DIFF
--- a/packages/abap-deploy-config-sub-generator/src/utils/project.ts
+++ b/packages/abap-deploy-config-sub-generator/src/utils/project.ts
@@ -24,19 +24,34 @@ export async function indexHtmlExists(fs: Editor, path: string): Promise<boolean
  *
  * @param path - The path to the project.
  * @param isS4HC - Whether the project is Cloud.
+ * @param launchAsSubGen - Whether the project is launched as a sub generator.
+ * @param fs - The file system editor.
  * @returns The variant namespace.
  */
-export async function getVariantNamespace(path: string, isS4HC: boolean): Promise<string | undefined> {
+export async function getVariantNamespace(
+    path: string,
+    isS4HC: boolean,
+    launchAsSubGen: boolean,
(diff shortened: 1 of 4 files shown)


packages/abap-deploy-config-sub-generator/src/utils/project.ts
```diff
@@ -24,19 +24,34 @@ export async function indexHtmlExists(fs: Editor, path: string): Promise<boolean
...
 
     try {
-        const webappPath = await getWebappPath(path);
-        const filePath = join(webappPath, FileName.ManifestAppDescrVar);
-        if (existsSync(filePath)) {
-            const descriptor = JSON.parse(readFileSync(filePath, 'utf-8'));
-            return descriptor.namespace;
+        if (launchAsSubGen) {
+            const filePath = join(await getWebappPath(path, fs), FileName.ManifestAppDescrVar);
+            // When launched as sub-generator, prioritize memory-based approach
```
IainSAP: It shouldnt be necessary to have different code for in mem or on disk. The same code should be used in both cases. `mem-fs` will read from disk if the file path is not in memory. We should always be referencing `mem-fs` instance during the generator lifecycle.

nikmace: @IainSAP You are right, here is the change [55a9fc8](https://github.com/SAP/open-ux-tools/pull/3687/commits/55a9fc88fa02f5f7e6ab7ac2de768d252762c620). I got confused because `@sap-ux/project-access` package uses both..

mmilko01 (approved):
Re-approving

IainSAP (approved):
Thanks @nikmace 

- Comment addressed
- Did not test manually

PR #3691, opened 2025-10-01
fix: Wrong navigation between steps in FLP Configuration generator for ADP

mmilko01:
#3692

* Introduces an additional property persisted in AppWizard cache that indicates if credential prompts were already shown. If the property is set to `true` the credential prompts are added as a step in Yeoman UI to prevent issues with the navigation.


nikmace (approved):
Code looks good
Coverage is good, tests adjusted
Changeset is good
Did not test manually

Issue #3578, opened 2025-08-29
TBI - `axios-extension`: Replace service-key auth with reentrance tickets for all Abap connections


IainSAP:
### Description (include screenshots)
SAP Cloud Abap systems will remove support for Service Keys as an authentication mechanism shortly. This TBI is to ensure these auth APIs continue to allow authentication with the reentrance ticket approach for all S4HC and BTP abap systems.

There are some other changes which are required for this to work correctly. TBD

### Value
Optional - details on how this technical backlog item provides value to end-users or dev team eg. Performance, Quality, Maintainability. This will help determine the ranking in relation to feature development

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

Issue #3641, opened 2025-09-18
BUG - isS4HC in btp-utils is incorrectly tagging destinations


longieirl:
### Description
When an SAP BTP destination is passed to the method `isS4HC`, in some instances, it can tag the destination incorrectly.

If the destination meets the following criteria, it's tagged as a `S4HC` system;

WebIDEUsage contains ODATA_ABAP
Authentication type is SAML_ASSERTION
ProxyType is Internet

https://github.com/SAP/open-ux-tools/blob/master/packages/btp-utils/src/destination.ts#L246

However, if a destination with similar properties is defined, for example, when ABAP Cloud environments are exposed to different subaccounts, the same properties are used thus the destination can be tagged as S4HC which is correct.

In the case of an ABAP Cloud env, it can contain other properties for example, `abap_cloud` so the `isS4HC` needs to be more strict on how it determines the type of system its handling.

### Steps to Reproduce
Create a destination using;
```
#clientKey=<< Existing password/certificate removed on export >>
#
tokenServiceURLType=Dedicated
audience=https\:///myabap-cloud-env.abap.hana.ondemand.com/
Description=SAML Assertion
authnContextClassRef=urn\:oasis\:names\:tc\:SAML\:2.0\:ac\:classes\:PreviousSession
WebIDEEnabled=true
URL=https\:///myabap-cloud-env.abap.hana.ondemand.com/
Name=mys4hc-destination
WebIDEUsage=odata_abap,dev_abap
Type=HTTP
HTML5.DynamicDestination=true
Authentication=SAMLAssertion
nameIdFormat=urn\:oasis\:names\:tc\:SAML\:1.1\:nameid-format\:emailAddress
HTML5.Timeout=60000
ProxyType=Internet
```

Load the app-gen wizard, select your new system and S4HC will shown in the drop down list.

### Expected results

System should be tagged with BTP.

### Actual results

System should be tagged with S4HC

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Not impacting on any flows, will close until raised as an issue or blocker.

PR #3693, opened 2025-10-01
Replace service key based connections with re-entrance tickets

IainSAP:
https://github.com/SAP/open-ux-tools/issues/3578

- Service Key support will be removed. Clients should align to use re-entrance tickets to create secure sessions.
- Removed guessing of API and UI hosts. Virtual hosts should now always be requested from the backend as they do not necessarily follow a naming convention. 
- Replaces the internal connection implementation but keeps the exported APIs as is for backward compatibility with clients in the short term
- Updates tests
- Replace use of backend system (`@sap-ux/store`) categorization as `BTP` or `S4HC` with `Abap Cloud` as the distinction was being made based on auth type which does not hold true. 
- Remove toggle for service key enablement in `@sap-ux/odata-service-inquirer`
- Continue to support `isSCP` property in existing yamls for backwards compatibility (preview middleware) however new apps that connect to cloud systems will not always use `authenticationType: reentranceTicket`
- Replaces previous implementation of `user()` from UAA with the endpoint `/sap/bc/adt/core/http/systeminformation` which also contains the logged on user info. However this will be the Abap user not the UAA (IDP) user.


packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -58,7 +58,9 @@
             "metadataFilePathNotValid": "The metadata file does not exist or is not accessible. Please specify a valid file path.",
             "capProjectNotFound": "The folder you have selected does not contain a valid CAP project. Please check and try again.",
             "warningCertificateValidationDisabled": "Certificate validation has been disabled by the user.",
-            "annotationsNotFound": "Annotations not found for the specified service."
+            "annotationsNotFound": "Annotations not found for the specified service.",
```
lfindlaysap, line 61:
```suggestion
            "annotationsNotFound": "There were no Annotations found for the specified service.",
```

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -58,7 +58,9 @@
...
+            "backendSystemExistsWarning": "A saved system connection entry: {{backendName}} already exists with the same URL. Please reuse the existing entry or remove it, using the systems panel, before adding a new connection.",
+            "reentranceTicketSystemHostOnly": "ABAP cloud system urls must only contain the hostname (e.g. 'https://123.abc.com')"
```
```suggestion
            "reentranceTicketSystemHostOnly": "ABAP cloud system urls must only contain the hostname. For example, https://123.abc.com."
```

packages/store/src/translations/ux-store.i18n.json
```diff
@@ -6,7 +6,8 @@
         "noPropertiesSpecified": "No properties specified for update",
         "systemMigrationFailed": "Migrating systems from secure store failed",
         "systemAlreadyExistsInHybridStore": "System with ID: [{{systemId}}] already exists in hybrid store. Not migrating",
-        "couldNotDeleteRefreshToken": "Could not delete refresh token for [{{- url}}]"
+        "couldNotDeleteRefreshToken": "Could not delete refresh token for [{{- url}}]",
```
lfindlaysap, line 9:
```suggestion
        "couldNotDeleteRefreshToken": "Cannot delete the refresh token for [{{- url}}]",
```

packages/store/src/translations/ux-store.i18n.json
```diff
@@ -6,7 +6,8 @@
...
+        "backendSystemEntityKeyExists": "Backend system entity was not written. Backend system entity key already exists: {{entityKey}}"
```
lfindlaysap:
```suggestion
        "backendSystemEntityKeyExists": "The back-end system entity was not written because the back-end system entity key already exists: {{entityKey}}"
```

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
IainSAP: Shouldnt this be:
```
ABAP cloud system urls must only contain the hostname, for example, https://123.abc.com.
```

packages/store/src/translations/ux-store.i18n.json
IainSAP: Not changed here

packages/store/src/translations/ux-store.i18n.json
IainSAP: `BackendSystem` is technical term referring to the TypeScript type. This is a technical log message. So should be:
```
"'BackendSystem' entity was not written. 'BackendSystem' entity key already exists: {{entityKey}}"
```

packages/store/src/translations/ux-store.i18n.json
lfindlaysap: What does this mean? Sorry, I'm confused.

packages/store/src/translations/ux-store.i18n.json
```diff
@@ -6,7 +6,8 @@
         "noPropertiesSpecified": "No properties specified for update",
         "systemMigrationFailed": "Migrating systems from secure store failed",
         "systemAlreadyExistsInHybridStore": "System with ID: [{{systemId}}] already exists in hybrid store. Not migrating",
-        "couldNotDeleteRefreshToken": "Could not delete refresh token for [{{- url}}]"
+        "couldNotDeleteRefreshToken": "Could not delete refresh token for [{{- url}}]",
+        "backendSystemEntityKeyExists": "'BackendSystem' entity was not written. 'BackendSystem' entity key already exists: {{entityKey}}"
```
lfindlaysap:
```suggestion
        "backendSystemEntityKeyExists": "The 'BackendSystem' entity was not written because a 'BackendSystem' entity key already exists: {{entityKey}}."
```

packages/store/src/translations/ux-store.i18n.json
IainSAP: This text is not changed in this PR...the diff is showing due to the newline.

packages/abap-deploy-config-inquirer/src/prompts/validators.ts
```diff
@@ -134,13 +134,12 @@ export async function updateDestinationPromptState(
         updatePromptState({
             url: dest?.Host,
             client: dest['sap-client'],
-            isS4HC: isS4HC(dest),
+            isAbapCloud: isS4HC(dest),
```
longieirl, line 137: I need to followup on this bug; https://github.com/SAP/open-ux-tools/issues/3641 to ensure the scope to determine if its S4HC is more strict. Currently, it's too broad, for example, setting up cross subaccount destinations to support ABAP Cloud using SAMLAssertion will return `true` which is incorrrect.

IainSAP: But `isS4HC` for destinations is much more specific.  What do I need to do here to resolve this comment? Generally we should not be using auth type to determine system type but rather use the auth type directly when its needed for connectivity.

IainSAP:
Back to draft it seems we still need to support UAA with credentials from the CLI without browser.

packages/backend-proxy-middleware/src/base/proxy.ts
```diff
@@ -270,9 +276,9 @@ export async function enhanceConfigForSystem(
             // sending a request to the backend to get token
             await provider.getAtoInfo();
         } else {
-            throw new Error('Cannot connect to ABAP Environment on BTP without service keys.');
+            throw new Error('Cannot connect to ABAP Environment on BTP using OAuth without service keys.');
         }
-    } else if (system?.authenticationType === AuthenticationType.ReentranceTicket) {
+    } else if (system && authType === 'reentranceTicket') {
```
heimwege: Please continue using a central constant for `reentranceTicket`. `AuthenticationType.ReentranceTicket`

packages/backend-proxy-middleware/src/base/proxy.ts
```diff
@@ -250,16 +256,16 @@ export async function enhanceConfigsForDestination(
...
  */
 export async function enhanceConfigForSystem(
     proxyOptions: Options & { headers: object },
     system: BackendSystem | undefined,
-    oAuthRequired: boolean | undefined,
+    authType: AuthenticationType,
     tokenChangedCallback: (refreshToken?: string) => void
 ): Promise<void> {
-    if (oAuthRequired) {
+    if (authType === 'oauth2') {
```
heimwege: Please use a central constant for `oauth2`. `AuthenticationType.OAuth2RefreshToken`

packages/backend-proxy-middleware/src/base/proxy.ts
```diff
@@ -397,6 +370,54 @@ export async function generateProxyMiddlewareOptions(
...
 }
 
+/**
+ * Determine the correct authentication configuration for connections from a non-BAS platform.
+ *
+ * @param backend the backend config loaded form the app ui5.yaml
+ * @param logger a logger instance
+ * @param proxyOptions additional proxy header, request and response settings
+ */
+async function updateVSCodeConfig(
```
heimwege: Thanks for doing the refactoring and giving this code its own method 👍🏻
Not using App Studio does not mean you use VS Code. I use IntelliJ as IDE and yet this function will be used. Please rename to something more general e.g. `updateLocalIDEConfig`

packages/backend-proxy-middleware/src/base/proxy.ts
```diff
@@ -397,6 +370,54 @@ export async function generateProxyMiddlewareOptions(
     return proxyOptions;
 }
 
+/**
+ * Determine the correct authentication configuration for connections from a non-BAS platform.
+ *
+ * @param backend the backend config loaded form the app ui5.yaml
```
heimwege: Actually it is the config from the UI5 configuration yaml file used when starting the UI5 server. This is not necessarily `ui5.yaml`.

```suggestion
 * @param backend the backend config loaded form the yaml config
```

packages/backend-proxy-middleware/src/base/proxy.ts
```diff
@@ -397,6 +370,54 @@ export async function generateProxyMiddlewareOptions(
...
+        )) ?? {
+            name: '<unknown>',
+            url: localBackend.url,
+            authenticationType: localBackend.authenticationType
+        };
+        // Auth type is determined from app config as we may have multiple stored systems with the same url/client using different auth types
+        await enhanceConfigForSystem(
+            proxyOptions,
+            system,
+            localBackend.authenticationType ?? (localBackend.scp ? 'oauth2' : 'basic'),
```
heimwege: Please use central constants. `AuthenticationType.OAuth2RefreshToken` and `AuthenticationType.Basic`

packages/backend-proxy-middleware/src/base/proxy.ts
```diff
@@ -397,6 +370,54 @@ export async function generateProxyMiddlewareOptions(
...
+    const localBackend = backend as LocalBackendConfig;
+    // check if system credentials are stored in the store
+    try {
+        const systemStore = await getService<BackendSystem, BackendSystemKey>({ logger, entityName: 'system' });
+        const system = (await systemStore.read(
+            new BackendSystemKey({ url: localBackend.url, client: localBackend.client })
+        )) ?? {
+            name: '<unknown>',
+            url: localBackend.url,
+            authenticationType: localBackend.authenticationType
```
heimwege, line 401: Do we need to check if `localBackend.authenticationType` actually contains a valid `authenticationType` and throw an exception if this is not the case?

packages/backend-proxy-middleware/src/base/proxy.ts
IainSAP: These are string literal types (not strings) so there is no need for a constant, its already a type and cannot be set to anything else. To import `AuthenticationType` requires an additional non-dev dependency. The point of defining string types in this way is to avoid dependencies where only the type is required. See there are 2 defs so the consumer decides which it needs: https://github.com/SAP/open-ux-tools/blob/6bdd60bc32967ebcb0a23e0bbd0c3e24d5048e67/packages/store/src/entities/backend-system.ts#L4-L11

IainSAP: I see we already using a dependency on `@sap-ux/store` so in this specific instance we can use the definitions directly (I will revert). But my point above is relevant FYI.

heimwege: Oh I totally overlooked that one 🙈

packages/backend-proxy-middleware/src/base/proxy.ts
IainSAP: Should we not rely on the type being correct?  `BaseBackendConfig` uses a proper type for `AuthenticationType`. If a bad value has got this far the issue is elsewhere and pre-exists. I think that might be for another PR. I think the loading of the yaml should be doing the validation, that would be an appropriate separation of concern?

heimwege: Agree could be a separate task.

PR #3696, opened 2025-10-02
feat: Store view name as part of additional info in fragment body

IvoSG:
Feat for [#3479.](https://github.com/SAP/open-ux-tools/issues/3479)
- Store viewName in the fragment body
- Add additional info in other xml fragment templates


nikmace (approved):
Changes look good
Changeset looks good
Coverage is good
Did not test manually

_Please don't forget to fill the right panel of the PR with the assignee and labels 👍🏻_

packages/preview-middleware-client/src/cpe/additional-change-info/add-xml-additional-info.ts
```diff
@@ -21,14 +22,24 @@ export type AddXMLChangeContent = {
...
+        if (templateName) {
+            result.templateName = templateName;
+        }
+        if (controlType && targetAggregation && viewName) {
+            result.targetAggregation = targetAggregation;
+            result.controlType = controlType;
+            result.viewName = viewName;
+        }   
+        return result;
+    }
```
voicis: This doesn't seem to scale well. Next time we need to add something else we'll need to add another set of conditions which are repeated in two places. Maybe we can just check if there are any keys set before we return?
```javascript
return Object.keys(result).length > 0 ? result : undefined;
```

packages/preview-middleware-client/src/utils/core.ts
```diff
@@ -71,6 +72,38 @@ export function hasParent(component: ManagedObject, parentIdToFind: string): boo
...
+ * @param control - Control instance  
+ * @returns View instance if found, undefined otherwise
+ */
+export function findViewByControl(control: Element | ManagedObject): View | undefined {
+    if (!control) {
+        return undefined;
+    }
+    if (isA('sap.ui.core.mvc.View', control)) {
+        return control as View;
+    }
```
voicis:
```suggestion
    if (isA<View>('sap.ui.core.mvc.View', control)) {
        return control;
    }
```

packages/preview-middleware-client/src/cpe/additional-change-info/add-xml-additional-info.ts
IvoSG: Changed in: [5053c37](https://github.com/SAP/open-ux-tools/pull/3696/commits/2470c199ee5cee97baf3df9da34c8f34c5053c37)

packages/preview-middleware-client/src/utils/core.ts
IvoSG: Changed in: [5053c37](https://github.com/SAP/open-ux-tools/pull/3696/commits/2470c199ee5cee97baf3df9da34c8f34c5053c37)

Issue #3701, opened 2025-10-06
BUG - Parameter `sap-appvar-id` should not be added in FLP Config for Adaptation Project changes


mmilko01:
### Description
Parameter `sap-appvar-id` should not be added as part of the FLP configuration for both `Add New` and `Replace` scenarios. Currently, the parameter is added in the `manifest.appdescr_variant` as: 
```
"sap-appvar-id": {
      "required": true,
      "filter": {
        "value": "<app.variant>",
        "format": "plain"
      },
      "launcherValue": {
        "value": "<app.variant>"
      }
    }
```

### Steps to Reproduce
1. Create an Adaptation Project of type CloudReady
2. Open `SAP Fiori Launchpad Configuration for Adaptation Project` wizard
3. Select either `Replace Base Application's Existing Tiles` or `Add a New Tile` and click `Finish`
4. See that `sap-appvar-id` parameter is added to the `manifest.appdescr_variant`

### Expected results
`sap-appvar-id` is not added to the `manifest.appdescr_variant`

### Actual results
`sap-appvar-id` is added to the `manifest.appdescr_variant`

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3702, opened 2025-10-06
fix: Remove sap-appvar-id parameter from FLP configuration changes for ADP

mmilko01:
#3701 

* Parameter `sap-appvar-id` is no longer added in parameter section of FLP configuration changes for both `Add a New Tile` or `Replace Base Application's Existing Tiles` scenarios.


IvoSG (approved):
Code looks good
Adapted tests
Did not tested manually

Issue #3698, opened 2025-10-03
BUG - Project type appears twice under system and password prompts when cloud system is selected


avasilev-sap:
### Steps to Reproduce
**Steps to reproduce:**
1. Select cloud system which does not require auth and wait for the system to load.
2. Select cloud system which requires auth.
3. Enter credentials and press login.
=> The project type appears twice beneath the system and password prompts.

**Another way to reproduce the issue:**
1. Login to a cloud system.
2. Press Start-over.
3. Start again the generator.
=> In this scenario the issue happens because the generator caches prompt data.

### Expected results
Project type should appear only once beneath the system prompt or password prompt in case the system requires authentication.

### Actual results
Project type appears beneath both prompts.

### Screenshots

<img width="1099" height="567" alt="Image" src="https://github.com/user-attachments/assets/ca68f6c8-55cc-46ce-86a0-5058eb637f0e" />

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3706, opened 2025-10-07
Project name appears twice under system and password prompts when cloud system is selected.

avasilev-sap:
#3698


mmilko01 (approved):
Changeset OK
Changes look good
Excellent test coverage
Did not test manually

PR #3707, opened 2025-10-07
chore: update  cds-compiler-facade dependency

Jimmy-Joseph19:
Consume new cds-compiler-facade


vadson71 (approved):
Dependency updates look good

PR #3711, opened 2025-10-08
fix(ui5-info): EINVALs error fix

korotkovao:
Related to internal issue 32726.

Update Windows spawn command for UI5 version check.


korotkovao:
Thank you All

PR #3712, opened 2025-10-08
Fix(ui5 library inquirer): 32726 versions sapui5 reuse libs gen

korotkovao:
Related to internal issue 32726.

Only show maintained UI5 versions in the Reusable Library Generator.
Show one patch per version.

### DIFF
--- a/packages/ui5-library-inquirer/src/index.ts
+++ b/packages/ui5-library-inquirer/src/index.ts
@@ -18,7 +18,10 @@ async function getPrompts(promptOptions?: UI5LibraryPromptOptions): Promise<Ques
         onlyNpmVersion: true
     };
     // Get only npm-available UI5 versions to avoid post-selection resolution
-    const ui5Versions = await getUI5Versions(filterOptions);
+    const allUI5Versions = await getUI5Versions(filterOptions);
+
+    // Filter for maintained versions only
+    const ui5Versions = allUI5Versions.filter((version) => version.maintained === true);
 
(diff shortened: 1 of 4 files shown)


packages/ui5-library-inquirer/src/index.ts
```diff
@@ -18,7 +18,10 @@ async function getPrompts(promptOptions?: UI5LibraryPromptOptions): Promise<Ques
         onlyNpmVersion: true
     };
     // Get only npm-available UI5 versions to avoid post-selection resolution
-    const ui5Versions = await getUI5Versions(filterOptions);
+    const allUI5Versions = await getUI5Versions(filterOptions);
+
+    // Filter for maintained versions only
+    const ui5Versions = allUI5Versions.filter((version) => version.maintained === true);
```
devinea, line 24: Would be preferable to filter the versions in `@sap-ux/ui5-info` based on patches returned in 
https://ui5.sap.com/version.json?sap-ui-config-patches=true&sap-ui-config-showall=true
```
"1.84": {
"version": "1.84.56",
"support": "Maintenance",
"lts": true,
"patches": [
"1.84.56",
"1.84.55",
"1.84.54",
"1.84.53",
"1.84.52",
"1.84.51",
"1.84.50"
]
},
```
versus current approach which returns ALL patches on npmjs.com for the major and minor version. 
In the example above 1.84.x we get an additional 50 patch versions returned.

FYI @IainSAP

korotkovao:
Thank you All

PR #3717, opened 2025-10-09
feat(fiori-mcp-server): add fiori_development_portal_extension.md, etc...

devinea:
- add fiori_development_portal_extension.md. Improved documentation for adding a analytical chart to a odata v4 list report. 
- update README with usage rules 
- new build scripts (minify the production release).

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,7 +1,9 @@
 name: Integration Tests
 
 on:
-    workflow_dispatch:   
+    schedule:
+        - cron: '30 1 * * 1-5' # At 01:30 on every day-of-week from Monday through Friday.
+    workflow_dispatch:
 jobs:
     ui5-versions:
         runs-on: ubuntu-latest


packages/fiori-mcp-server/README.md
```diff
@@ -70,6 +70,8 @@ The following rules help guide the LLM to use the server correctly:
 - When generating or modifying the SAP Fiori elements application on top of the CAP service use the Fiori MCP server if available.
 - When attempting to modify the SAP Fiori elements application like adding columns you must not use the screen personalization but instead modify the code of the project, before this first check whether an MCP server provides a suitable function.
 - When previewing the SAP Fiori elements application use the most specific script for the app in the `package.json`.
```
vinayhospete: preview app rule already exist in line 72

vinayhospete: can we also add below rule to avoid project creation issues ?

- When creating new CAP project ask cds mcp server. while using `cds init`,  **do not include any `--add` options** (e.g., `--add hana`, `--add sqlite`, `--add tiny-sample`, etc.). Only use the plain form: cds init <project-name>


Screenshot 2025-10-10 at 12 40 03 PM

vinayhospete:
Tested this PR for analytical chart, it works awesome !!!
https://github.tools.sap/ux-foundation/teched2025-CA261/tree/main/exercises/ex5

packages/fiori-mcp-server/README.md
devinea: added. https://github.com/SAP/open-ux-tools/pull/3717/commits/65c6977b692377d5814736386a2747691f125de0 Thanks

heimwege: I would not add rules to our mcp server that are basically rules for other mcp servers. That's why we did not add instructions for the cds mcp so far. In my opinion they should be added to the mcp they belong. In this case CDS. @cfg74 what do you think?

devinea: @heimwege 
Are you referring to this:
`When creating new CAP project ask cds mcp server. while using cds init, do not include any --add options (e.g., --add hana, --add sqlite, --add tiny-sample, etc.). Only use the plain form: cds init`

or also 
```
- When running CAP Node.js projects for development, you "MUST" use `cds watch` or `npm run watch-*` scripts. You "MUST NOT" use `npm start`, `npm run start`, `cds-serve`, or `cds serve` because:
  - `cds watch` provides live reload functionality
  - `npm run watch-*` scripts are optimized for development with hot reloading
  - `npm start` typically runs production builds without development features
```
As this one is very important for cds file changes from fiori mcp doc search responses.

heimwege: I was referring to all of them.

How to preview a fiori elements app is described in

 _When previewing the SAP Fiori elements application use the most specific script for the app in the `package.json`._

That rule should enforce picking the script from the generator. It uses cds watch but with --livereload false because as soon as cds livereload and our livereload middleware are running in parallel we end up in reload hell.

I thought we agreed to have only mcp specific rules with the respective mcp servers.
The changes from this pr would somehow bring fiori mcp in the position of an orchestrator that has cross mcp instructions.
Not sure if we want to do this. If we align to do so, fine for me. I just want this to be a conscious decision.

vinayhospete: we can discuss today in mcp daily. Rule to enforce `cds init` without sample can be dropped. I noticed today LLM creates sample bookshop app first and further replaces bookshop app with requested travel management app.

devinea: @heimwege, @vinayhospete , @cfg74  Updated based on discussion.

packages/fiori-mcp-server/README.md
```diff
@@ -70,6 +70,7 @@ The following rules help guide the LLM to use the server correctly:
...
+- When modifying CAP Node.js project you "MUST" only use cds watch using either the fiori application "watch-" script in the CAP project or "cds watch". You "MUST NOT" use "npm run start", "npm start", "cds-serve" or "cds serve".
```
vinayhospete: we need to remove separate rule for cds watch and modify existing rule to 
```
- When previewing the SAP Fiori elements application use the most specific `npm run watch-*` script for the app in the `package.json`.
```

vinayhospete (approved):
Changes looks good to me. manually tested

PR #3719, opened 2025-10-10
[FEATURE] Support Lean-App card generation for CAP based Fiori applications

Singham-24:
This update adds support for launching the Card Generation Dialog in CAP-based Fiori applications, aligning it with existing functionality available for RAP-based apps. CAP apps can now generate, save, and share cards using the same flow.

Summary of Changes :-

Enabled card generation trigger for CAP-based projects

Reused existing logic for dialog handling

Ensured consistent behavior across RAP and CAP apps

### DIFF
--- a/packages/preview-middleware/test/unit/base/flp.test.ts
+++ b/packages/preview-middleware/test/unit/base/flp.test.ts
@@ -965,7 +965,7 @@ describe('FlpSandbox', () => {
             const manifest = JSON.parse(readFileSync(join(fixtures, 'simple-app/webapp/manifest.json'), 'utf-8'));
             manifest['sap.app'].i18n = {
                 bundleUrl: 'i18n/i18n.properties',
-                supportedLocales: ['de', 'es'],
+                supportedLocales: ['de', 'es']
             };
             await flp.init(manifest);
 


heimwege:
@Singham-24 please don't use internal URLs in an external PR.
FYI I deleted the URL, you might want to rephrase the description.

.changeset/blue-crews-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/preview-middleware': major
+---
+
+i18n for CAP nased apps
```
heimwege: Typo

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1067,8 +1059,39 @@ export class FlpSandbox {
         try {
             this.fs = this.fs ?? create(createStorage());
             const webappPath = await getWebappPath(path.resolve(), this.fs);
-            const i18nPath = this.manifest['sap.app'].i18n as string;
-            const filePath = i18nPath ? join(webappPath, i18nPath) : join(webappPath, 'i18n', 'i18n.properties');
+            const i18nConfig = this.manifest['sap.app'].i18n;
```
heimwege, line 1062: General question: What does the i18n adjustments have to do with card generation for CAP based apps?

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1067,8 +1059,39 @@ export class FlpSandbox {
...
+            let supportedLocales: unknown[] = [];
+            let fallbackLocale: string | undefined;
+
+            if (typeof i18nConfig === 'string') {
+                i18nPath = i18nConfig;
+            } else if (typeof i18nConfig === 'object' && i18nConfig !== null && 'bundleUrl' in i18nConfig) {
+                const { bundleUrl } = i18nConfig as {
+                    bundleUrl: string;
+                };
+                i18nPath = bundleUrl;
```
heimwege: All of this type casting is not needed because of the statement `'bundleUrl' in i18nConfig`
```suggestion
                i18nPath = i18nConfig.bundleUrl;
```

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1067,8 +1059,39 @@ export class FlpSandbox {
...
+                supportedLocales = i18nConfig.supportedLocales ?? [];
+                fallbackLocale = i18nConfig.fallbackLocale;
+            } else {
+                i18nPath = 'i18n/i18n.properties';
```
heimwege, line 1075: In case there is no `i18n` defined (as string or object) in the manifest why using `i18n/i18n.properties` as fallback? This file might not even exists.

packages/preview-middleware/test/unit/base/flp.test.ts
```diff
@@ -820,20 +820,22 @@ describe('FlpSandbox', () => {
...
             await flp.init(manifest);
 
             const app = express();
             app.use(flp.router);
 
             server = supertest(app);
+            return { flp, app };
         };
-
+        let flpInstance: any;
```
heimwege: Please use proper types instead of `any`.

`flpInstance`seems to be set in `beforeAll` but not used at all. Can this be deleted?

packages/preview-middleware/test/unit/base/flp.test.ts
```diff
@@ -926,62 +928,60 @@ describe('FlpSandbox', () => {
...
-                cardGenerator: {
-                    path: 'test/flpCardGeneratorSandbox.html'
-                }
-            }
-        };
+        test('should handle string i18n path', async () => {
+            const newI18nEntry = [{ key: 'HELLO', value: 'Hello World' }];
 
-        let mockFsPromisesWriteFile: jest.Mock;
+            (flp as any).manifest = {
```
heimwege: There seems to be a test missing that has no `sap.app.i18n` defined at all to test the `i18nPath = 'i18n/i18n.properties';` fallback

packages/preview-middleware/test/unit/base/flp.test.ts
```diff
@@ -926,62 +928,60 @@ describe('FlpSandbox', () => {
...
+                'sap.app': { i18n: 'i18n/custom.properties' }
+            };
```
heimwege: Please don't use type `any`

```suggestion
            flp['manifest']['sap.app']['i18n'] = 'i18n/custom.properties';
```

packages/preview-middleware/test/unit/base/flp.test.ts
```diff
@@ -926,62 +928,60 @@ describe('FlpSandbox', () => {
...
-            expect(response.status).toBe(404);
-        });
+            (flp as any).manifest = {
+                'sap.app': {
+                    i18n: {
+                        bundleUrl: 'i18n/i18n.properties',
+                        supportedLocales: ['de', 'es']
+                    }
+                }
+            };
```
heimwege: Please don't use type `any`

```suggestion
            flp['manifest']['sap.app']['i18n'] = {
                bundleUrl: 'i18n/i18n.properties',
                supportedLocales: ['de', 'es']
            };
```

packages/preview-middleware/test/unit/base/flp.test.ts
```diff
@@ -926,62 +928,60 @@ describe('FlpSandbox', () => {
...
-            app.use(flp.router);
+            (flp as any).manifest = {
+                'sap.app': {
+                    i18n: {
+                        bundleUrl: 'i18n/i18n.properties',
+                        supportedLocales: ['de', 'es'],
+                        fallbackLocale: 'de'
+                    }
+                }
+            };
```
heimwege: Please don't use type `any`

```suggestion
            flp['manifest']['sap.app']['i18n'] = {
                bundleUrl: 'i18n/i18n.properties',
                supportedLocales: ['de', 'es'],
                fallbackLocale: 'de'
            };
```

packages/preview-middleware/src/base/flp.ts
Singham-24: Earlier, the implementation only handled the default i18n.properties file and did not account for scenarios where the manifest defined i18n configurations using a bundleUrl.

packages/preview-middleware/src/base/flp.ts

packages/preview-middleware/src/base/flp.ts
Singham-24: If the manifest does not define i18n, we fall back to i18n/i18n.properties for backward compatibility, since the previous implementation relied on this file as the default i18n resource.

packages/preview-middleware/test/unit/base/flp.test.ts
Singham-24: done

packages/preview-middleware/test/unit/base/flp.test.ts
Singham-24: added

packages/preview-middleware/test/unit/base/flp.test.ts
Singham-24: Patched the manifest before init()

packages/preview-middleware/test/unit/base/flp.test.ts
Singham-24: Patched the manifest before init()

packages/preview-middleware/test/unit/base/flp.test.ts
Singham-24: Patched the manifest before init()

packages/preview-middleware/src/base/flp.ts
heimwege: I understand that. My question was rather what the i18n adjustment has to do with general cards support for CAP apps. According to this PR because of the missing object support for the i18n property, CAP apps have not been supported. But that's not true. Card generation fro CAP apps has been disabled in https://github.com/SAP/open-ux-tools/pull/3349 because you cannot deploy the generated cards / the generated cards will be ignored by the FLP. That only works for ABAP based FLPs. So basically my questions are:
1) Did you verify that the non-ABAP FLP(s) now also support cards? If not, why delete the restriction in this PR?
2) Why is this i18n change related to CAP at all? The type of the i18n property has been `string | object` all the time. That's nothing CAP specific. So that part rather looks like a bugfix to me.

packages/preview-middleware/test/unit/base/flp.test.ts
```diff
@@ -820,18 +820,18 @@ describe('FlpSandbox', () => {
...
-            const flp = new FlpSandbox(mockConfig, mockProject, mockUtils, logger);
+            flp = new FlpSandbox(mockConfig, mockProject, mockUtils, logger);
             const manifest = JSON.parse(readFileSync(join(fixtures, 'simple-app/webapp/manifest.json'), 'utf-8'));
             await flp.init(manifest);
 
             const app = express();
             app.use(flp.router);
 
             server = supertest(app);
+            return { flp, app };
```
heimwege, line 833: The return parameters seem to be not use anywhere. Can this be deleted?

Singham-24:
not needed

PR #3724, opened 2025-10-10
Fix/35934/incorrect entity list rich text editor

kjose90:
Internal issue - 35934

- Disabled Relative binding context when `no suitable entities found` when creating rich text editor building blocks
- Added support for disabled and tool tip text in `ui-prompting`
- Added disabled style to `ui-components`


packages/fe-fpm-writer/src/prompts/translations/i18n.ts
```diff
@@ -219,6 +219,7 @@ const ns1 = {
                 'validation': 'A view or fragment is required to generate the Rich Text Editor building block.'
             },
             'bindingContextType': 'Binding Context',
+            'relativeBindingDisabledTooltip': 'No suitable entities available for relative binding',
```
lfindlaysap:
```suggestion
            'relativeBindingDisabledTooltip': 'There are no suitable entities available for relative binding.',
```

packages/fe-fpm-writer/src/building-block/prompts/utils/questions.ts
```diff
@@ -411,13 +411,13 @@ export function getBindingContextTypePrompt(properties: Partial<ListPromptQuesti
...
         name: 'buildingBlockData.metaPath.bindingContextType',
-        choices: [
+        choices: properties.choices ?? [
             { name: t('bindingContextType.option.absolute') as string, value: 'absolute' },
             { name: t('bindingContextType.option.relative') as string, value: 'relative' }
         ],
         guiOptions: {
             ...guiOptions,
-            selectType: 'static'
+            selectType: 'dynamic'
```
815are: maybe more correct logic would be `typeof properties.choices === 'function' ? 'dynamic' : 'static'`
?

eouin (approved):
lgtm

kjose90:
Thank you all for the reviews

Issue #3741, opened 2025-10-16
BUG - @sap-ux/fiori-generator-shared: Backend System label not returned for `AbapCloud`


IainSAP:
### Related Feature
#3564 

### Description
The function `@sap-ux/fiori-gneator-shared/getSystemTypeLabel` returns the an empty label for system type `AbapCloud` (there is not type for backend system system type)

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3742, opened 2025-10-16
fix(fiori-generator-shared): #3741 Abap cloud label not returned

IainSAP:
Fix for #3741 

- Backend system type is not a type but a string, an incorrect string was used when creating a string label for the type `AbapCloud`


cianmSAP (approved):
Update looks good for fix
Have not tested
Changeset present

longieirl (approved):
Changeset is good.
Tests to cover change.
Change is good.

PR #3753, opened 2025-10-20
fix(fe-fpm-writer): custom column bb fixes

docirl:
internal issue 36038

- use aggregation path correctlly for cc in object pages

### DIFF
--- a/packages/fe-fpm-writer/src/building-block/index.ts
+++ b/packages/fe-fpm-writer/src/building-block/index.ts
@@ -25,7 +25,7 @@ import { getMinimumUI5Version } from '@sap-ux/project-access';
 import { detectTabSpacing, extendJSON } from '../common/file';
 import { getManifest, getManifestPath } from '../common/utils';
 import { getDefaultFragmentContent, setCommonDefaults } from '../common/defaults';
-import { getOrAddNamespace } from './prompts/utils/xml';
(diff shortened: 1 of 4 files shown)


packages/fe-fpm-writer/src/building-block/index.ts
```diff
@@ -479,6 +477,66 @@ function updateViewFile(
...
+/**
+ * Selects nodes for the provided aggregation path. If the initial namespace-aware XPath returns
+ * no results, it retries by rewriting any unprefixed path segments into a namespace-agnostic
+ * local-name() predicate so that elements in a default namespace (e.g. <VBox>) can still be matched.
+ *
+ * @param {string} aggregationPath - Original aggregation XPath expression
+ * @param {Document} viewDocument - XML view/fragment document
+ * @returns {Node[]} Array of matched nodes (empty if none)
+ */
+export function selectTargetNodes(aggregationPath: string, viewDocument: Document): Node[] {
```
815are: maybe move to utils on same level as other xml related function like [getOrAddNamespace](https://github.com/SAP/open-ux-tools/blob/fix/fe-fpm-writer/bb-customcolumn-fixes/packages/fe-fpm-writer/src/building-block/prompts/utils/xml.ts#L172)?

docirl: thanks, moved to `xml.ts`

PR #3757, opened 2025-10-21
feat: Integrate a new ABAP api for retrieval of the UI5 framework version for a system in the ADP generator. 

avasilev-sap:
#3756

* Integrate a new ABAP api for retrieval of the UI5 version for a system. The new api all is part from the axios extension and is currently used in the ADP generator. If the api call fails due to connectivity issues, server error or if the backend returns a response without body we switch to the legacy api. If both apis fail to retrieve the UI5 version a warning is displayed in the attributes page in the ADP wizard under the UI5 version prompt. Whatever the result from the version retrieval is we still give the user the ability to choose a framework version.
Additional work:
* The translation key used to store the message for an undetected UI5 version was put by mistake in a different package than the generator package(see the screenshot).
* Change the severity for the `error.notDeployableNotFlexEnabledSystemError` error message.
* For some systems with internal user the bootstrap api returns the UI5 version in the format <major>.<minor>.<patch>-SNAPSHOT which breaks the version normalization logic. For internal users we manually set the snapshot prefix, as a result we store the system version with the snapshot label repeated twice.

Screenshot 2025-10-21 at 19 30 53

Screenshot 2025-10-23 at 13 46 08

Testing done: 
* Manually verified:
 - When we create ADP project with internal user and system, with snapshot UI5 version(reported from the bootstrap api) the project is successfully created, the Adaptation editor can be started and the SAP application is loaded.
 - If the bootstrap api throws then the UI5 version is loaded with the ADT api.
* Add unit tests.


IainSAP:
@avasilev-sap Isnt this already available from `@sap-ux/ui5-info`: https://github.com/SAP/open-ux-tools/blob/e10e4b6abbb6811c2fac6318736013d51ee802d7/packages/ui5-info/src/ui5-version-backend.ts#L11

avasilev-sap:
> @avasilev-sap Isnt this already available from `@sap-ux/ui5-info`:
> 
> https://github.com/SAP/open-ux-tools/blob/e10e4b6abbb6811c2fac6318736013d51ee802d7/packages/ui5-info/src/ui5-version-backend.ts#L11

It is preferable not to use this method since Axios applies a lot of stuff behind the scene to the actual request, like cookies headers etc. when we use a service through the service provider.

packages/adp-tooling/src/abap/client.ts
```diff
@@ -44,14 +44,22 @@ export async function getFlexUISupportedSystem(
...
+    try {
+        const ui5VersionService = provider.getUI5VersionService();
+        // We await here because in case of any exception from the new api we want
+        // to fallback to the legacy api - UI5RtVersionService.getUI5Version().
+        return await ui5VersionService.getUI5Version();
+    } catch {
+        const ui5RtVersionService = await provider.getAdtService<UI5RtVersionService>(UI5RtVersionService);
+        return ui5RtVersionService?.getUI5Version();
+    }
 }
```
nikmace, line 69: A couple of things here:
1. There is no need for that two-line comment. The method is short and self-explanatory. Removing it will make the code leaner without losing clarity.
2. It would be very good to add logging in the catch block. Right now, the error will be silently swallowed. You need to add a new message that could maybe also accept the error message.

packages/axios-extension/src/abap/ui5-version-service.ts
```diff
@@ -0,0 +1,67 @@
...
+import type { Service } from 'base/service-provider';
+
+/**
+ * Stores the version of the UI5 framework.
+ */
+interface UI5VersionInfo {
+    Version: string;
+}
+
+/**
```
nikmace, line 12: It would be more straightforward to throw the error directly

packages/axios-extension/src/abap/ui5-version-service.ts
```diff
@@ -0,0 +1,67 @@
...
+     */
+    public log: Logger;
+
+    /**
+     * Util method used to parse json string as a javascript plain object.
+     *
+     * @param data The json data as a string.
+     * @returns The parsed javascript object.
+     */
+    private readonly toJSON = (data: string): unknown => JSON.parse(data);
```
nikmace: _Optional_: I would just put this directly in the transformResponse to spare me some lines of code and JSdoc. And also spare writing extra tests below to test this method. But this approach is ok too.
```ts
transformResponse: (data: string): unknown => JSON.parse(data)
```

packages/axios-extension/test/abap/ui5-version-service.test.ts
```diff
@@ -0,0 +1,191 @@
+import type { Logger } from '@sap-ux/logger';
+import type { AxiosRequestConfig } from 'axios';
+import { UI5VersionService } from '../../src/abap/ui5-version-service';
+
+// Mock axios
```
nikmace: All of the comments are not needed here and below in the tests because they don't bring much value.

packages/axios-extension/src/abap/ui5-version-service.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * The method returns a valid version of the UI5 framework on the ABAP system
+     *  or throws an error if the version is not provided, malformed or there are connectivity
+     *  issues or server errors.
+     *
```
nikmace, line 31: There are some extra spaces here in the JSDoc right where the text starts

packages/axios-extension/test/abap/ui5-version-service.test.ts
```diff
@@ -0,0 +1,191 @@
...
+                statusText: 'OK',
+                headers: {},
+                config: {}
+            };
+            mockGet.mockResolvedValue(mockResponse);
+
+            // Act & Assert
+            await expect(service.getUI5Version()).rejects.toThrow('UI5 version not provided.');
+            expect(mockLogger.error).toHaveBeenCalledWith('Could not get UI5 Version.');
+        });
```
nikmace: It would probably be enough to leave 1-2 tests here because these three are checking the same thing

packages/generator-adp/src/app/questions/configuration.ts
```diff
@@ -689,7 +689,15 @@ export class ConfigPrompter {
...
      */
     private async loadUI5Versions(): Promise<void> {
-        const version = await getSystemUI5Version(this.abapProvider);
+        let version: string | undefined;
+        try {
+            version = await getSystemUI5Version(this.abapProvider);
+        } catch {
+            // Do nothing. The systemVersion field will be later set to undefined.
+            // If systemVersion field is undefined then the system version was not
+            // loaded successfuly, we display a warning in that case in the generator.
```
nikmace: You could also log here, or like I suggested in my first review comment.

nikmace (changes requested):
Overall, the change looks good 👍🏻

packages/generator-adp/src/translations/generator-adp.i18n.json
```diff
@@ -105,5 +105,8 @@
         "projectDoesNotExistMta": "Provide the path to the MTA project where you want to have your Adaptation Project created.",
         "noAdaptableBusinessServiceFoundInMta": "No adaptable business service found in the MTA.",
         "fetchBaseInboundsFailed": "Fetching base application inbounds failed: {{error}}."
+    },
+    "validators": {
+        "ui5VersionNotDetectedError": "The SAPUI5 version of the selected system cannot be determined. You will be able to create and edit adaptation projects using the newest version but it will not be usable on this system until the system`s SAPUI5 version is upgraded to version 1.71 or higher."
```
```suggestion
        "ui5VersionNotDetectedError": "The SAPUI5 version of the selected system cannot be determined. You are able to create and edit adaptation projects that use the latest SAPUI5 version, but they are not usable on this system until the system's SAPUI5 version is upgraded to version 1.71 or higher."
```

avasilev-sap: This is a legacy message which was part of the product before. Are you sure we need to change the text here? Can you confirm with PM/PO.

packages/adp-tooling/src/abap/client.ts
avasilev-sap: fixed in 82804dce2c5d5c291cd3ef3cc69d0b49966646a1

packages/axios-extension/src/abap/ui5-version-service.ts
avasilev-sap: fixed in 82804dce2c5d5c291cd3ef3cc69d0b49966646a1

packages/axios-extension/src/abap/ui5-version-service.ts
avasilev-sap: fixed in 82804dce2c5d5c291cd3ef3cc69d0b49966646a1

packages/axios-extension/test/abap/ui5-version-service.test.ts
avasilev-sap: fixed in 82804dce2c5d5c291cd3ef3cc69d0b49966646a1

packages/axios-extension/src/abap/ui5-version-service.ts
avasilev-sap: fixed in 82804dce2c5d5c291cd3ef3cc69d0b49966646a1

packages/axios-extension/test/abap/ui5-version-service.test.ts
avasilev-sap: fixed in 82804dce2c5d5c291cd3ef3cc69d0b49966646a1

packages/generator-adp/src/app/questions/configuration.ts
avasilev-sap: fixed in 82804dce2c5d5c291cd3ef3cc69d0b49966646a1

packages/generator-adp/src/translations/generator-adp.i18n.json
avasilev-sap: Since this is grammar check i think we are safe. Fixed in 6249ba90d7930b11fd839c586a7a3bca591a89b5, 10x.

IainSAP:
> > @avasilev-sap Isnt this already available from `@sap-ux/ui5-info`:
> > https://github.com/SAP/open-ux-tools/blob/e10e4b6abbb6811c2fac6318736013d51ee802d7/packages/ui5-info/src/ui5-version-backend.ts#L11
> 
> It is preferable not to use this method since Axios applies a lot of stuff behind the scene to the actual request, like cookies headers etc. when we use a service through the service provider.

Is this new implementation different to the existing UI5 version information provided by `@sap-ux/ui5-info`? Im not clear whats missing there as this works in all current scenarios supported by application generation, for example. 

Otherwise we have some handling of the ui5 versions that might also be reused? There are quite a few filter options: https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-info/src/types.ts#L3

avasilev-sap:
> > > @avasilev-sap Isnt this already available from `@sap-ux/ui5-info`:
> > > https://github.com/SAP/open-ux-tools/blob/e10e4b6abbb6811c2fac6318736013d51ee802d7/packages/ui5-info/src/ui5-version-backend.ts#L11
> > 
> > 
> > It is preferable not to use this method since Axios applies a lot of stuff behind the scene to the actual request, like cookies headers etc. when we use a service through the service provider.
> 
> Is this new implementation different to the existing UI5 version information provided by `@sap-ux/ui5-info`? Im not clear whats missing there as this works in all current scenarios supported by application generation, for example.
> 
> Otherwise we have some handling of the ui5 versions that might also be reused? There are quite a few filter options: https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-info/src/types.ts#L3

Hi Ian I include @mmilko01. Misho can you help us with a resolution here. Do we have to reuse the call from the ui5-info package, we do not have dependency to that package in the generator-adp package. Also it uses another axios instance and we do not pass through the service provider from the axios-extension. Is it safe to pass the publicUrl from the AbapServiceProvider to the ui5-info > getSapSystemUI5Version?

heimwege:
@avasilev-sap @mmilko01 I agree to @IainSAP. Is there any reason why the `ui5-info` package is not being used (missing feature, etc.)? And if it is not being used for specific reasons, is something preventing us from rather enhancing `ui5-info` instead of having yet another implementation that has more or less the same purpose?

devinea:
@avasilev-sap @mmilko01 Please reuse the function from `@sap-ux/ui5-info` rather than adding a new one to axios extension.

packages/axios-extension/src/abap/ui5-version-service.ts
```diff
@@ -0,0 +1,50 @@
+import type { Logger } from '@sap-ux/logger';
+import { Axios } from 'axios';
+import type { Service } from 'base/service-provider';
+
+/**
+ * Stores the version of the UI5 framework.
```
heimwege, line 6: If you load the UI5 sources from npmjs (via `ui5 serve` or `fiori run` w/o proxying the UI5 sources), the root level `version` property of the UI5 version info will contain the application version instead of the UI5 version. So actually the JSDoc is not quite right because that's only the case when loading UI5 sources from CDN or an ABAP backend. But as the context here is loading the UI5 version info from an ABAP backend (via proxy), this is fine.

avasilev-sap:
> @avasilev-sap @mmilko01 Please reuse the function from `@sap-ux/ui5-info` rather than adding a new one to axios extension.

Hi we had a call with @heimwege we agreed to merge the change as is and later to do a subsequent sync on how this api will be integrated. The change is with a high priority.

Issue #3758, opened 2025-10-21
BUG - App Generator does not update modified backend systems (`force` option required)


IainSAP:
### Related Feature
https://github.com/SAP/open-ux-tools/issues/3578

### Description
When updating existing saved backend systems it is now necessary to pass a `{ force: true }` option. This is not done and so the save does not happen.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3759, opened 2025-10-21
fix(odata-service-inquirer): Updates to backend system writes required `force` option

IainSAP:
Fix for: https://github.com/SAP/open-ux-tools/issues/3758

- Force option is required for updates to existing systems (was already applied to systems panel)
- Update test to cover


cianmSAP (approved):
Update looks good
Have not tested
Changeset looks good

PR #3761, opened 2025-10-21
chore - Rimraf upgrade

devinea:
Chore - upgrade rimraf to 6.0.1 (excluding ADP templates which will be a separate PR).


IainSAP (approved):
Dep only updates

longieirl (approved):
Re-approving.

PR #3767, opened 2025-10-22
feat(fpm-writer): Filter building block custom filter

eouin:
Internal issue 32954

- [x] adds support for custom filters to be added to filter building blocks

### DIFF
--- a/packages/fiori-mcp-server/CHANGELOG.md
+++ b/packages/fiori-mcp-server/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/fiori-mcp-server
 
+## 0.4.1
+
+### Patch Changes
+
+-   9544c24: fix: page creation fails in v2
+
 ## 0.4.0
 
 ### Minor Changes
--- a/packages/fiori-mcp-server/package.json
+++ b/packages/fiori-mcp-server/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/fiori-mcp-server",
     "description": "SAP Fiori - Model Context Protocol (MCP) server",
-    "version": "0.4.0",
+    "version": "0.4.1",
     "keywords": [
         "SAP Fiori tools",
         "SAP Fiori elements",
--- a/packages/fiori-mcp-server/src/tools/functionalities/page/add-page.ts
+++ b/packages/fiori-mcp-server/src/tools/functionalities/page/add-page.ts
@@ -10,6 +10,8 @@ import { convertToSchema, resolveApplication, validateWithSchema } from '../../u
 import { ADD_PAGE } from '../../../constant';
 import { SapuxFtfsFileIO, getServiceName } from '../../../page-editor-api';
 import { buildPageCreationSchema } from './schema';
+import { PageTypeV4 } from '@sap/ux-specification/dist/types/src';
+import { getFioriElementsVersion } from './utils';
 
 export const ADD_PAGE_FUNCTIONALITY: GetFunctionalityDetailsOutput = {
     functionalityId: ADD_PAGE,
@@ -40,11 +42,16 @@ async function getFunctionalityDetails(params: GetFunctionalityDetailsInput): Pr
     const appData = await ftfsFileIo.readApp();
     const serviceName = await getServiceName(applicationAccess);
     const application = new Application({ params, applicationAccess, serviceName, appId, appData });
-
     const navigationOptions = await application.getCreationNavigationOptions();
     return {
         ...ADD_PAGE_FUNCTIONALITY,
-        parameters: convertToSchema(buildPageCreationSchema(navigationOptions.navigations, navigationOptions.entities))
+        parameters: convertToSchema(
+            buildPageCreationSchema(
+                navigationOptions.navigations,
+                navigationOptions.entities,
+                getFioriElementsVersion(appData)
+            )
+        )
     };
 }
 
@@ -75,15 +82,23 @@ async function executeFunctionality(params: ExecuteFunctionalityInput): Promise<
     const application = new Application({ params, applicationAccess, serviceName, appId, appData });
     // Parse page creation parameters
     const navigationOptions = await application.getCreationNavigationOptions();
-    const schema = buildPageCreationSchema(navigationOptions.navigations, navigationOptions.entities);
+    const schema = buildPageCreationSchema(
+        navigationOptions.navigations,
+        navigationOptions.entities,
+        getFioriElementsVersion(appData)
+    );
     const addPageParameters = validateWithSchema(schema, parameters);
-    const pageType = addPageParameters.pageType;
+    const pageType = addPageParameters.pageType ?? PageTypeV4.ObjectPage;
     const parentPage = 'parentPage' in addPageParameters ? addPageParameters.parentPage : undefined;
     const entitySet = 'entitySet' in addPageParameters ? addPageParameters.entitySet?.toString() : undefined;
     const pageNavigation =
         'pageNavigation' in addPageParameters ? addPageParameters.pageNavigation.toString() : undefined;
     const viewName =
-        'pageViewName' in addPageParameters && addPageParameters.pageViewName ? addPageParameters.pageViewName : '';
+        'pageViewName' in addPageParameters &&
+        addPageParameters.pageViewName &&
+        typeof addPageParameters.pageViewName === 'string'
+            ? addPageParameters.pageViewName
+            : '';
     return application.createPage({
         pageType: pageType,
         parent: parentPage,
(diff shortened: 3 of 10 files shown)


packages/fe-fpm-writer/src/filter/types.ts
```diff
@@ -5,6 +5,8 @@ import type { CustomElement, InternalCustomElement, Position, EventHandler } fro
  * Properties to create a custom filter.
  */
 export interface CustomFilter extends CustomElement, EventHandler {
+    // Unique filter field ID.
+    id: string;
```
815are: it affects manifest based custom filter field. Is there reason to change interface for manifest based custom filter field? I see updated filter.test.ts, but snapshots stays same

packages/fe-fpm-writer/src/building-block/index.ts
```diff
@@ -211,6 +262,35 @@ function processBuildingBlock<T extends BuildingBlock>(
         aggregationNamespace = getOrAddNamespace(xmlDocument, 'sap.fe.macros.table', 'macrosTable');
     }
 
+    if (isCustomFilterField(buildingBlockData) && buildingBlockData.embededFragment) {
+        const embededFragment = setCommonDefaults(buildingBlockData.embededFragment, manifestPath, manifest);
+        const config = createCustomFilterConfig(buildingBlockData, embededFragment);
+
+        const viewPath = join(
```
kjose90: Duplicate code. [Lines](https://github.com/SAP/open-ux-tools/pull/3767/files#diff-35bea2eceb48eb617af76a763f4f3225008b39a925bddca01f770211b1fe50d9R229) from 229 to 262 does the same thing. Please move this logic to new function some thing like `generateFragmentFile`?

packages/fe-fpm-writer/src/building-block/index.ts
```diff
@@ -145,6 +158,44 @@ function updateAggregationPathForTableColumns(
...
+        };
+    } else {
+        const useDefaultAggregation = xpathSelect("//*[local-name()='FilterField']", xmlDocument);
+        if (useDefaultAggregation && Array.isArray(useDefaultAggregation) && useDefaultAggregation.length > 0) {
+            return { updatedAggregationPath: aggregationPath, hasFilterFields: true };
+        }
+    }
+
+    return { updatedAggregationPath: aggregationPath, hasFilterFields: false };
+}
```
815are, line 214: I see content is very similar to `updateAggregationPathForTableColumns`:
image

maybe unify both functions to more reusable function?

packages/fe-fpm-writer/src/filter/types.ts
eouin: Done

packages/fe-fpm-writer/src/building-block/index.ts
```diff
@@ -219,6 +299,31 @@ function processBuildingBlock<T extends BuildingBlock>(
...
 }
 
+/**
+ * Creates an InternalCustomFilter configuration from CustomFilterField building block data.
+ *
+ * @param {CustomFilterField} buildingBlockData - The building block data
+ * @param {InternalCustomElement & EventHandler & CustomFragment & CustomElement & FragmentContentData} embededFragment - The embedded fragment data with defaults applied
+ * @returns {InternalCustomFilter} The internal custom filter configuration
+ */
+function createCustomFilterConfig(
```
kjose90: not sure if you need an extra function call ? Its simply returning an object. you can avoid and write inline config construction [here](https://github.com/SAP/open-ux-tools/pull/3767/files#diff-35bea2eceb48eb617af76a763f4f3225008b39a925bddca01f770211b1fe50d9R267) or even define like below 
```
 const config = BUILDING_BLOCK_CONFIG[type];
// make a map or something - so we can add more building blocks
export const BUILDING_BLOCK_CONFIG: Partial<Record<BuildingBlockType, BuildingBlockTemplateConfig>> = {
    [BuildingBlockType.CustomColumn]: {
        nodes: { explicit: "columns", default: "Column" },
        templateType: "common",
        templateFile: "common/Fragment.xml",
        namespace: { uri: "sap.fe.macros.table", prefix: "macrosTable" }
    },
    [BuildingBlockType.CustomFilterField]: {
        nodes: { explicit: "filterFields", default: "FilterField" },
        templateType: "filter",
        templateFile: "filter/fragment.xml",
        namespace: { uri: "sap.fe.macros.filterBar", prefix: "macros" }
    }
};
```

packages/fe-fpm-writer/src/building-block/index.ts
```diff
@@ -145,6 +158,44 @@ function updateAggregationPathForTableColumns(
...
+ * @param {string} aggregationPath - The current aggregation path
+ * @param {CustomFilterField} buildingBlockData - The building block data with embedded fragment
+ * @returns {object} Object containing the updated aggregation path
+ */
+function updateAggregationPathForFilterBar(
+    xmlDocument: Document,
+    aggregationPath: string,
+    buildingBlockData: CustomFilterField
+): { updatedAggregationPath: string; hasFilterFields: boolean } {
+    if (!buildingBlockData.embededFragment) {
```
kjose90, line 195: Similar code like `updateAggregationPathForTableColumns` - maybe again make a common function like `updateAggregationPath`

packages/fe-fpm-writer/src/building-block/index.ts
eouin: maybe later if a new building block is needed. keep it simple for now

kjose90: It will look more cleaner. Also you dont need this extra function call as mentioned previously `createCustomFilterConfig`.. you can just pass config to use `BUILDING_BLOCK_CONFIG` based on the type.

packages/fe-fpm-writer/test/unit/filter.test.ts
```diff
@@ -61,6 +61,7 @@ describe('CustomFilter', () => {
 
         test('Create several new custom filters', async () => {
             const secondFilter = {
+                id: 'testID2',
```
815are: seems obsolete change or it is required?

eouin: Done

packages/fe-fpm-writer/test/unit/__snapshots__/building-block.test.ts.snap
```diff
@@ -75,6 +75,149 @@ Object {
...
 `;
 
+exports[`Building Blocks CustomFilterField building block CustomFilterField detects macros:filterFields elements correctly: generate-custom-filter-field-with-macros-filter-fields 1`] = `
+"<mvc:View xmlns:core=\\"sap.ui.core\\" xmlns:mvc=\\"sap.ui.core.mvc\\" xmlns=\\"sap.m\\" xmlns:html=\\"http://www.w3.org/1999/xhtml\\" controllerName=\\"com.test.myApp.ext.main.Main\\" xmlns:macros=\\"sap.fe.macros\\">
+    <Page title=\\"Main\\">
+        <content>
+            <macros:FilterBar>
+                <macros:filterFields>
+                    <macros:FilterField/>
+                    <macros:FilterField key=\\"\\" label=\\"Custom Filter Field 2\\" id=\\"testCustomFilterField2\\" placement=\\"After\\" required=\\"true\\" property=\\"testProperty\\">
```
815are: seems not fully well with key attribute, why we need generate empty key? should it be there at all or it should be populated, I see empty `key=""` in all snaposhots for `macros:FilterField`. In case of custom column - I see some snapshots with populated `key="myColumnKey"`

eouin: Done

packages/fe-fpm-writer/src/building-block/index.ts
eouin: we can add that if we add a new building block

815are (approved):

packages/fe-fpm-writer/src/building-block/index.ts
```diff
@@ -11,35 +18,93 @@ import {
...
+        embededFragment?: EmbeddedFragmentData
+    ) => void;
+}
+
+/**
+ * Configuration map for building block types.
+ */
+export const BUILDING_BLOCK_CONFIG: Partial<Record<BuildingBlockType, BuildingBlockTemplateConfig>> = {
+    [BuildingBlockType.CustomColumn]: {
+        nodes: { explicit: 'columns', default: 'Column' },
```
kjose90, line 77: perhaps rename  rename these to something more descriptive?

eouin: Will do.

packages/fe-fpm-writer/src/building-block/index.ts
eouin: done

PR #3768, opened 2025-10-22
feat(store): migrate backend system file from .fioritools to .saptools

cianmSAP:
Internal issue 35418

Backend systems are migrated from `.fioritools` dir to a new `.saptools` folder

Updates required to `@sap-ux/odata-service-inquirer` to not call `SystemService` directly and utilise `getService` as a single point of entry. Reducing requirement to check for migration in all crud methods

### DIFF
--- a/examples/fe-fpm-cli/CHANGELOG.md
+++ b/examples/fe-fpm-cli/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/fe-fpm-cli
 
+## 0.1.53
+
+### Patch Changes
+
+-   Updated dependencies [0cff26e]
+    -   @sap-ux/fe-fpm-writer@0.39.15
+
 ## 0.1.52
 
 ### Patch Changes
--- a/examples/fe-fpm-cli/package.json
+++ b/examples/fe-fpm-cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/fe-fpm-cli",
-    "version": "0.1.52",
+    "version": "0.1.53",
     "description": "A simple CLI to prompt required information to create a building block using the fe-fpm-writer module's prompt and generate functions.",
     "license": "Apache-2.0",
     "private": true,
--- a/examples/odata-cli/CHANGELOG.md
+++ b/examples/odata-cli/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/odata-cli
 
+## 0.17.18
+
+### Patch Changes
+
+-   Updated dependencies [cdeb18b]
+    -   @sap-ux/axios-extension@1.24.0
+
 ## 0.17.17
 
 ### Patch Changes
--- a/examples/odata-cli/package.json
+++ b/examples/odata-cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/odata-cli",
-    "version": "0.17.17",
+    "version": "0.17.18",
     "description": "Simple example CLI uing the @sap-ux/axios-extension module to fetch metadata and annotations from an SAP system.",
     "license": "Apache-2.0",
     "private": true,
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,19 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.101
+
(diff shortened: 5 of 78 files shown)


packages/odata-service-inquirer/test/unit/prompts/prompts.test.ts
```diff
@@ -16,10 +15,10 @@ jest.mock('@sap-ux/btp-utils', () => {
     };
 });
 
-jest.mock('@sap-ux/store', () => ({
+jest.mock('../../../src/utils/store', () => ({
```
longieirl: Why is this required?

cianmSAP: It's not, I went down the path of mocking the function that uses @sap-ux/store, rather than mocking @sap-ux/store itself. But on reflection it's better to mock the call to the store. Have updated

packages/store/src/utils/index.ts
```diff
@@ -1,5 +1,6 @@
 import { homedir } from 'node:os';
 import path from 'node:path';
```
longieirl: Should you be using the named import here? i.e. `import { join } from 'node:path';`

packages/odata-service-inquirer/test/unit/index-api.test.ts
```diff
@@ -7,6 +7,7 @@ import * as prompts from '../../src/prompts';
 import * as systemSelection from '../../src/prompts/datasources/sap-system/system-selection';
 import LoggerHelper from '../../src/prompts/logger-helper';
 import { PromptState } from '../../src/utils';
+import { Service } from '../../../fiori-mcp-server/src/tools/functionalities/page/service';
```
IainSAP: Not sure this is the correct import ? `fiori-mcp-server/` doesnt appear to be used.

```suggestion
```

packages/store/src/services/backend-system.ts
```diff
@@ -81,5 +90,31 @@ export class SystemService implements Service<BackendSystem, BackendSystemKey> {
...
+    }
     return new SystemService(logger, options);
 }
+
+/**
+ * Ensure settings are migrated from the old fiori tools directory to the new sap development tools directory.
+ */
+function ensureSettingsMigrated(): void {
+    const sapDevToolsDir = getSapDevToolsDirectory();
+    const migrationFlag = join(sapDevToolsDir, '.migrated');
```
devinea:
```suggestion
    const migrationFlag = join(sapDevToolsDir, '.systemsMigrated');
```
Can we make this more specific?

packages/store/src/services/backend-system.ts
```diff
@@ -81,5 +90,31 @@ export class SystemService implements Service<BackendSystem, BackendSystemKey> {
...
+        return;
+    }
+
+    const systemFileName = getEntityFileName(Entity.BackendSystem);
+    const legacyPath = join(getFioriToolsDirectory(), systemFileName);
+    const newPath = join(sapDevToolsDir, systemFileName);
+
+    if (existsSync(legacyPath)) {
+        mkdirSync(dirname(newPath), { recursive: true });
+        copyFileSync(legacyPath, newPath);
```
devinea: What if the new file already exists?
Read it and merge content?

cianmSAP: I don't follow, I check for `existsSync(migrationFlag)` above which will return early if the migration has occurred

packages/odata-service-inquirer/test/unit/index-api.test.ts
cianmSAP: Thanks, auto import

packages/store/src/services/backend-system.ts
cianmSAP: I think I follow now, so the scenario is they already have a `systems.json` in the new directory? Is that likely?

devinea: I think it could happen if consumers of store module are released/ installed by customers out of sync. Which seems very possible IMO.

cianmSAP: Thanks Austin, I was thinking it could be a one time migration and we could avoid continious file reads/checks. But it makes sense - a consumer using an old version could create a new system and this would not be reflected in a different consumer using the latest.

Issue #2683, opened 2024-12-10
BUG - preview middleware uses fixed list of ui5 libs although the list is version dependent


heimwege:
### Description
preview-middleware uses a [list](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/src/base/config.ts#L105)) of UI5 libs. That list seems to be UI5 version specific (compare [1.120.24](https://ui5.sap.com/1.120.24/#/api/sap) to [latest](https://ui5.sap.com/#/api/sap)). So it needs to be adjusted to a more dynamic list.

More Details see https://github.com/SAP/open-ux-tools/pull/2613/files#r1872645371

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

devinea:
@heimwege  Looking at this again I think we we can use `ui5Json.libraries.map(lib => lib.name)` to replace the static list of namespaces (which was a short cut to knowing the library names in the first place)

from https://ui5.sap.com/resources/sap-ui-version.json
returns
[...]

PR #3774, opened 2025-10-27
fix(preview-middleware): default flp.libs parameter to true

heimwege:
- default the `flp.libs` parameter of `preview-middleware` to `true` for edmx backends
- override the `flp.libs` parameter of `preview-middleware` to `false` for cap
- for compatibility reasons: enhance the `serve-static-middleware` to additionally serve paths that start with `/resources` w/o the `/resources` part (excluding `/resources` as standalone path) to ensure that the workaround to use servestatic on `/resources/my/custom/lib` because the `flp.libs` parameter was not known/used will still continue to work.

Sample: An existing `path` configuration `/resources/my/custom/lib` to serve a custom library locally w/o `flp.libs` `true` will have two servestatic paths.
1. `/resources/my/custom/lib` <-- the existing one remains unchanged, but will no longer be used because
2. `/my/custom/lib          ` <-- the new `flp.libs` default `true` will add the custom library and its respective ABAP backend url to the  UI5 loader configuration; That's why we add this `path` configuration on-the-fly to ensure compatibility with the new `flp.libs` `true` default

Sample: An existing `path` configuration `/resources` will remain unchanged and still only has one servestatic path so that the whole UI5 framework can still be served locally
1. `/resources` <-- the existing one remains unchanged, and will continue to work


packages/preview-middleware/src/base/flp.ts
```diff
@@ -546,17 +545,15 @@
...
         let isCdn = false;
         if (!host) {
             this.logger.error('Unable to fetch UI5 version: No host found in request header.');
         } else {
             try {
                 const versionUrl = `${protocol}://${host}${baseUrl}/resources/sap-ui-version.json`;
-                const responseJson = (await fetch(versionUrl).then((res) => res.json())) as
-                    | { name: string; libraries: { name: string; version: string }[] }
-                    | undefined;
+                const responseJson = (await fetch(versionUrl).then((res) => res.json())) as UI5VersionDetails | undefined;
```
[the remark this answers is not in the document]
heimwege: Code did not change. Only Types have been adjusted. Hence I'll dismiss this one because `preview-middleware` only runs on localhost.

packages/preview-middleware-client/src/utils/version.ts
```diff
@@ -2,12 +2,30 @@ import VersionInfo from 'sap/ui/VersionInfo';
...
-          version: string;
-      };
+type UI5VersionDetails = {
+    /**
+     * Contains either
+     * - the name of the distribution or
+     * - the id of the application in case the UI5 sources have beeng loaded from npmjs.
+     */
+    // eslint-disable-next-line @typescript-eslint/no-duplicate-type-constituents, @typescript-eslint/no-redundant-type-constituents
+    name: 'SAPUI5 Distribution' | Manifest['sap.app']['id'];
```
heimwege, line 15: Actually this is `string | sting` which makes no sense but it was done like this to point out where the second string value actually comes from. This helps when using code completion.

packages/preview-middleware-client/src/utils/version.ts
```diff
@@ -2,12 +2,30 @@ import VersionInfo from 'sap/ui/VersionInfo';
...
+    /**
+     * Contains either
+     * - the version of the UI5 framework or
+     * - the version of the application in case the UI5 sources have been loaded from npmjs.
+     */
+    // eslint-disable-next-line @typescript-eslint/no-duplicate-type-constituents, @typescript-eslint/no-redundant-type-constituents
+    version: string | Manifest['sap.app']['applicationVersion']['version'];
```
heimwege, line 22: Actually this is `string | sting` which makes no sense but it was done like this to point out where the second string value actually comes from. This helps when using code completion.

packages/preview-middleware/src/base/flp.ts
```diff
@@ -60,6 +60,38 @@ import { readFileSync } from 'node:fs';
 import { getIntegrationCard } from './utils/cards';
 import { createPropertiesI18nEntries } from '@sap-ux/i18n';
 
+type UI5VersionDetails = {
+    /**
+     * Contains either
+     * - the name of the distribution or
+     * - the id of the application in case the UI5 sources have beeng loaded from npmjs.
+     */
+    name: 'SAPUI5 Distribution' | Manifest['sap.app']['id'];
```
heimwege, line 92: Actually this is `string | sting` which makes no sense but it was done like this to point out where the second string value actually comes from. This helps when using code completion.

I'm wondering why eslint does not complain about the `string | string` issue here. This is only checked in `preview-middleware-client` `version.ts` 🤷🏻

packages/preview-middleware/src/base/flp.ts
```diff
@@ -60,6 +60,38 @@ import { readFileSync } from 'node:fs';
...
+    /**
+     * Contains either
+     * - the version of the UI5 framework or
+     * - the version of the application in case the UI5 sources have been loaded from npmjs.
+     */
+    version: string | Manifest['sap.app']['applicationVersion']['version'];
```
heimwege, line 98: Actually this is `string | sting` which makes no sense but it was done like this to point out where the second string value actually comes from. This helps when using code completion.

I'm wondering why eslint does not complain about the `string | string` issue here. This is only checked in `preview-middleware-client` `version.ts` 🤷🏻

tobiasqueck (dismissed):
The PR contains quite a lot of changes but in the end you just set the default to `true` for `edmx` and added a workaround to the serve static middleware. The rest is code restructurings and cleanups, correct?
That mix made the review a little more time consuming but in the end, the code is now cleaner and better structured. The functional change is also a good idea and should have been done a long time ago (I missed it when creating the initial version of this code). The workaround in the serve static middleware will cause confusion in the future - I am sure of it - but I understand it is required to keep existing setups working as-is.

heimwege:
@tobiasqueck it also fixes an issue that the list of UI5 libs is not static but depending on the actual UI5 version. So we now load it from the UI5 version and only use a static list as fallback. Basically this fixes parts of https://github.com/SAP/open-ux-tools/issues/2683

devinea (dismissed):
Thanks @heimwege. Changes look good.
Haven't tested locally.
changeset ✅

longieirl (dismissed):
Changesets are good.
Changes are clear.
Tests to cover change.

devinea (dismissed):
New approach seems light a better way to check for required libs and where to load them from. 
Code changes and tests lokk good. 
I haven't tested locally. 
Changeset 👍

devinea (dismissed):
Re-approving

PR #3766, opened 2025-10-22
Chore - Support sonar scans on forks

devinea:

PR #3775, opened 2025-10-28
Chore add sonar workflow

devinea:
Handles sonar scan archives from other PR's. 
pipeline build will be updated in https://github.com/SAP/open-ux-tools/pull/3766 after


heimwege (approved):
lgtm. Let's see if we can finally improve the workflow for 3rd party contributions 🥳

PR #3776, opened 2025-10-28
fix(mcp-server): Pass 'layer' and 'ui5Version' to the specification API method exportConfig.

815are:
It was reported that flex changes are written with wrong value for 'layer' property.
Changes in PR:
1. Method `[getFlexChangeLayer](https://github.com/SAP/open-ux-tools/blob/fix/mcpServerFlexChangeLayer/packages/fiori-mcp-server/src/page-editor-api/project.ts) - read value for 'layer' from package.json of application;
2. Calls for `exportConfig` adjusted to pass `ui5Version` and `layer` -> https://github.com/SAP/open-ux-tools/blob/fix/mcpServerFlexChangeLayer/packages/fiori-mcp-server/src/page-editor-api/sapuxFtfsFileIO.ts#L261

### DIFF
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.102
+
+### Patch Changes
+
+-   @sap-ux/system-access@0.6.24
+-   @sap-ux/fiori-elements-writer@2.7.32
+-   @sap-ux/fiori-freestyle-writer@2.4.57
+
 ## 1.1.101
 
 ### Patch Changes
--- a/examples/simple-generator/package.json
+++ b/examples/simple-generator/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/generator-simple-fe",
-    "version": "1.1.101",
+    "version": "1.1.102",
     "description": "Simple example of a yeoman generator for Fiori elements.",
     "license": "Apache-2.0",
     "private": true,
--- a/packages/abap-deploy-config-inquirer/CHANGELOG.md
+++ b/packages/abap-deploy-config-inquirer/CHANGELOG.md
@@ -1,5 +1,16 @@
 # @sap-ux/abap-deploy-config-inquirer
 
+## 1.6.39
+
+### Patch Changes
+
+-   Updated dependencies [36b0f19]
+    -   @sap-ux/store@1.3.0
+    -   @sap-ux/fiori-generator-shared@0.13.29
+    -   @sap-ux/system-access@0.6.24
+    -   @sap-ux/ui5-config@0.29.8
+    -   @sap-ux/inquirer-common@0.8.8
+
 ## 1.6.38
 
 ### Patch Changes
--- a/packages/abap-deploy-config-inquirer/package.json
+++ b/packages/abap-deploy-config-inquirer/package.json
@@ -6,7 +6,7 @@
         "url": "https://github.com/SAP/open-ux-tools.git",
         "directory": "packages/abap-deploy-config-inquirer"
     },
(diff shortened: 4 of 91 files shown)


packages/fiori-mcp-server/src/page-editor-api/project.ts
```diff
@@ -127,3 +129,24 @@ export async function getUI5Version(appAccess: ApplicationAccess): Promise<strin
...
     return ui5Version ?? 'latest';
 }
+
+/**
+ * Reads the application's `package.json` file and extracts the `sapuxLayer` property.
+ *
+ * @param root - the absolute path to the application's root directory.
+ * @returns A promise that resolves to the `sapuxLayer` value from `package.json`.
+ */
+export async function getFlexChangeLayer(root: string): Promise<FlexChangeLayer> {
```
heimwege, line 139: @Klaus-Keller Don't we have something like this already in a central place like project-access/info?

Klaus-Keller (approved):
Thanks @815are!

* changes to use `sapuxLayer` for flex changes looks good
* changeset exists
* test coverage is great
* did a visual review no manual test

Approved from my side.

broksy (approved):
Changeset reflects to the changes made and are covered by tests.
Thanks! Approved from my side.

packages/fiori-mcp-server/src/page-editor-api/sapuxFtfsFileIO.ts
```diff
@@ -241,4 +239,23 @@ export class SapuxFtfsFileIO {
...
+     * @param params - Partial export parameters to include in the resulting configuration
+     * @returns A promise that resolves to the fully composed export configuration object
+     */
+    private async getExportConfigParameters(
+        manifest: Manifest,
+        params: Partial<ExportParametersV2Type | ExportParametersV4Type>
+    ): Promise<ExportConfigParameters> {
+        const odataVersion = manifest['sap.app']?.dataSources?.mainService?.settings?.odataVersion;
+        const exportConfig = (odataVersion === '2.0' ? { v2: params } : { v4: params }) as ExportConfigParameters;
+        exportConfig.ui5Version = await getUI5Version(this.appAccess);
```
heimwege, line 257: @Klaus-Keller what's the difference between using (yet another method called) `getUI5Version` from the axios extension ([here](https://github.com/SAP/open-ux-tools/blob/main/packages/axios-extension/src/abap/ui5-version-service.ts#L34) is another one in the axios extension) and using the `ui5-info` package?

packages/fiori-mcp-server/src/page-editor-api/project.ts
Klaus-Keller: Also thought about moving it to project-access, but we can do this in a subsequent step if other consumers occur.

packages/fiori-mcp-server/src/page-editor-api/sapuxFtfsFileIO.ts
Klaus-Keller: This is getting the UI5 version from manifest.json and uses the central function `getMinimumUI5Version()` from project access

815are: @Klaus-Keller 

mcp-server approach uses `getMinimumUI5Version` - https://github.com/SAP/open-ux-tools/blob/499aa407630fa0935a6d28dd9a107a3f961bc65a/packages/fiori-mcp-server/src/page-editor-api/project.ts#L122-L131
but first it resolved manifest.json and then does additional fallback checks(I used logic from tool-suite here), perhaps we can remove those fallback checks or move to `getMinimumUI5Version`, but not sure
`getMinimumUI5Version` does not have fallback case if version is not valid

PR #3778, opened 2025-10-28
feat(eslint-plugin-fiori-tools): replace custom plugin with ux consistency

ManuelPastorini:
This PR

- Define fiori languages
- Add following rules:
    - Flex Enabled (expected value is "true")
    - Copy To Clipboard
    - widthIncludingColumnHeader in Object Pages
    - Creation Mode for tables v2
    - Creation Mode for tables v4

### DIFF
--- a/packages/adp-flp-config-sub-generator/CHANGELOG.md
+++ b/packages/adp-flp-config-sub-generator/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/adp-flp-config-sub-generator
 
+## 0.1.147
+
+### Patch Changes
+
+- Updated dependencies [ce4b29c]
+    - @sap-ux/adp-tooling@0.18.36
+    - @sap-ux/flp-config-inquirer@0.4.95
+
 ## 0.1.146
 
 ### Patch Changes
--- a/packages/adp-flp-config-sub-generator/package.json
+++ b/packages/adp-flp-config-sub-generator/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/adp-flp-config-sub-generator",
     "description": "Generator for adding FLP configuration to an Adaptation Project",
-    "version": "0.1.146",
+    "version": "0.1.147",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/adp-tooling/CHANGELOG.md
+++ b/packages/adp-tooling/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/adp-tooling
 
+## 0.18.36
+
+### Patch Changes
+
+- ce4b29c: Upgrade qs/body-parser/express
+
 ## 0.18.35
 
 ### Patch Changes
--- a/packages/adp-tooling/package.json
+++ b/packages/adp-tooling/package.json
@@ -9,7 +9,7 @@
     "bugs": {
         "url": "https://github.com/SAP/open-ux-tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug+label%3Aadp-tooling"
     },
-    "version": "0.18.35",
+    "version": "0.18.36",
     "license": "Apache-2.0",
     "author": "@SAP/ux-tools-team",
     "main": "dist/index.js",
@@ -77,7 +77,7 @@
         "adm-zip": "0.5.16",
         "cross-env": "^7.0.3",
         "dotenv": "16.3.1",
-        "express": "4.21.2",
+        "express": "4.22.1",
         "nock": "13.4.0",
         "rimraf": "6.0.1",
         "supertest": "7.1.4"
--- a/packages/backend-proxy-middleware-cf/CHANGELOG.md
+++ b/packages/backend-proxy-middleware-cf/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/backend-proxy-middleware-cf
 
+## 0.0.26
+
+### Patch Changes
+
+- ce4b29c: Upgrade qs/body-parser/express
+- Updated dependencies [ce4b29c]
+    - @sap-ux/adp-tooling@0.18.36
+
 ## 0.0.25
 
 ### Patch Changes
--- a/packages/backend-proxy-middleware-cf/package.json
+++ b/packages/backend-proxy-middleware-cf/package.json
@@ -9,7 +9,7 @@
     "bugs": {
         "url": "https://github.com/SAP/open-ux-tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug+label%3Abackend-proxy-middleware-cf"
     },
-    "version": "0.0.25",
+    "version": "0.0.26",
     "license": "Apache-2.0",
     "author": "@SAP/ux-tools-team",
     "main": "dist/index.js",
@@ -40,7 +40,7 @@
         "@types/express": "4.17.21",
         "@types/http-proxy": "^1.17.5",
         "@types/supertest": "2.0.12",
-        "express": "4.21.2",
+        "express": "4.22.1",
         "nock": "13.4.0",
         "supertest": "7.1.4",
         "connect": "^3.7.0",
--- a/packages/backend-proxy-middleware/CHANGELOG.md
+++ b/packages/backend-proxy-middleware/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/backend-proxy-middleware
 
+## 0.10.28
+
+### Patch Changes
+
+- ce4b29c: Upgrade qs/body-parser/express
+
 ## 0.10.27
 
 ### Patch Changes
--- a/packages/backend-proxy-middleware/package.json
+++ b/packages/backend-proxy-middleware/package.json
@@ -9,7 +9,7 @@
     "bugs": {
         "url": "https://github.com/SAP/open-ux-tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug+label%3Abackend-proxy-middleware"
     },
-    "version": "0.10.27",
+    "version": "0.10.28",
     "license": "Apache-2.0",
     "author": "@SAP/ux-tools-team",
     "main": "dist/index.js",
@@ -52,7 +52,7 @@
         "@types/prompts": "2.4.4",
         "@types/proxy-from-env": "1.0.1",
         "@types/supertest": "2.0.12",
-        "express": "4.21.2",
+        "express": "4.22.1",
         "connect": "^3.7.0",
         "nock": "13.4.0",
         "supertest": "7.1.4",
--- a/packages/control-property-editor/CHANGELOG.md
+++ b/packages/control-property-editor/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/control-property-editor
 
+## 0.7.8
+
+### Patch Changes
+
+- ce4b29c: Upgrade qs/body-parser/express
+
 ## 0.7.7
 
 ### Patch Changes
--- a/packages/control-property-editor/package.json
+++ b/packages/control-property-editor/package.json
@@ -3,7 +3,7 @@
     "displayName": "Control Property Editor",
     "description": "Control Property Editor",
     "license": "Apache-2.0",
-    "version": "0.7.7",
+    "version": "0.7.8",
     "main": "dist/app.js",
     "repository": {
         "type": "git",
@@ -37,7 +37,7 @@
         "@types/remote-redux-devtools": "0.5.4",
         "@types/source-map-support": "0.5.0",
         "@types/react": "16.14.55",
-        "body-parser": "1.20.3",
+        "body-parser": "1.20.4",
         "eslint-plugin-react": "7.37.5",
         "http-proxy-middleware": "2.0.9",
         "i18next": "25.3.0",
--- a/packages/create/CHANGELOG.md
+++ b/packages/create/CHANGELOG.md
@@ -1,5 +1,15 @@
 # @sap-ux/create
 
+## 0.14.30
+
+### Patch Changes
+
+- Updated dependencies [ce4b29c]
+    - @sap-ux/preview-middleware@0.23.89
+    - @sap-ux/adp-tooling@0.18.36
+    - @sap-ux/app-config-writer@0.6.76
+    - @sap-ux/flp-config-inquirer@0.4.95
+
 ## 0.14.29
 
 ### Patch Changes
--- a/packages/create/package.json
+++ b/packages/create/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/create",
     "description": "SAP Fiori tools module to add or remove features",
-    "version": "0.14.29",
+    "version": "0.14.30",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/flp-config-inquirer/CHANGELOG.md
+++ b/packages/flp-config-inquirer/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/flp-config-inquirer
 
+## 0.4.95
+
+### Patch Changes
+
+- Updated dependencies [ce4b29c]
+    - @sap-ux/adp-tooling@0.18.36
+
 ## 0.4.94
 
 ### Patch Changes
--- a/packages/flp-config-inquirer/package.json
+++ b/packages/flp-config-inquirer/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/flp-config-inquirer",
     "description": "Prompts module that can prompt users for inputs required for FLP configuration",
-    "version": "0.4.94",
+    "version": "0.4.95",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/flp-config-sub-generator/CHANGELOG.md
+++ b/packages/flp-config-sub-generator/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/flp-config-sub-generator
 
+## 0.3.99
+
+### Patch Changes
+
+- @sap-ux/app-config-writer@0.6.76
+- @sap-ux/flp-config-inquirer@0.4.95
+
 ## 0.3.98
 
 ### Patch Changes
--- a/packages/flp-config-sub-generator/package.json
+++ b/packages/flp-config-sub-generator/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/flp-config-sub-generator",
     "description": "Generator for creating Fiori Launcpad configuration",
-    "version": "0.3.98",
+    "version": "0.3.99",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/generator-adp/CHANGELOG.md
+++ b/packages/generator-adp/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/generator-adp
 
+## 0.7.36
+
+### Patch Changes
+
+- Updated dependencies [ce4b29c]
+    - @sap-ux/adp-tooling@0.18.36
+
 ## 0.7.35
 
 ### Patch Changes
--- a/packages/generator-adp/package.json
+++ b/packages/generator-adp/package.json
@@ -3,7 +3,7 @@
     "displayName": "SAPUI5 Adaptation Project",
     "homepage": "https://help.sap.com/viewer/584e0bcbfd4a4aff91c815cefa0bce2d/Cloud/en-US/ada9567b767941aba8d49fdb4fdedea7.html",
     "description": "Adaptation project allows you to create an app variant for an existing SAP Fiori elements-based or SAPUI5 freestyle application, without changing the original application.",
-    "version": "0.7.35",
+    "version": "0.7.36",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/preview-middleware/CHANGELOG.md
+++ b/packages/preview-middleware/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/preview-middleware
 
+## 0.23.89
+
+### Patch Changes
+
+- ce4b29c: Upgrade qs/body-parser/express
+- Updated dependencies [ce4b29c]
+    - @sap-ux/adp-tooling@0.18.36
+
 ## 0.23.88
 
 ### Patch Changes
--- a/packages/preview-middleware/package.json
+++ b/packages/preview-middleware/package.json
@@ -9,7 +9,7 @@
     "bugs": {
         "url": "https://github.com/SAP/open-ux-tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug+label%3Apreview-middleware"
     },
-    "version": "0.23.88",
+    "version": "0.23.89",
     "license": "Apache-2.0",
     "author": "@SAP/ux-tools-team",
     "main": "dist/index.js",
@@ -71,7 +71,7 @@
         "connect": "^3.7.0",
         "copyfiles": "2.4.1",
         "dotenv": "16.3.1",
-        "express": "4.21.2",
+        "express": "4.22.1",
         "nock": "13.4.0",
         "npm-run-all2": "6.2.0",
         "supertest": "7.1.4"
--- a/packages/reload-middleware/CHANGELOG.md
+++ b/packages/reload-middleware/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/reload-middleware
 
+## 0.3.10
+
+### Patch Changes
+
+- ce4b29c: Upgrade qs/body-parser/express
+
 ## 0.3.9
 
 ### Patch Changes
--- a/packages/reload-middleware/package.json
+++ b/packages/reload-middleware/package.json
@@ -9,7 +9,7 @@
     "bugs": {
         "url": "https://github.com/SAP/open-ux-tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug+label%3Areload-middleware"
     },
-    "version": "0.3.9",
+    "version": "0.3.10",
     "author": "@SAP/ux-tools-team",
     "license": "Apache-2.0",
     "main": "dist/index.js",
@@ -48,7 +48,7 @@
         "@types/livereload": "0.9.5",
         "@types/supertest": "2.0.12",
         "axios": "1.12.2",
-        "express": "4.21.2",
+        "express": "4.22.1",
         "supertest": "7.1.4"
     },
     "engines": {
--- a/packages/serve-static-middleware/CHANGELOG.md
+++ b/packages/serve-static-middleware/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/serve-static-middleware
 
+## 0.4.7
+
+### Patch Changes
+
+- ce4b29c: Upgrade qs/body-parser/express
+
 ## 0.4.6
 
 ### Patch Changes
--- a/packages/serve-static-middleware/package.json
+++ b/packages/serve-static-middleware/package.json
@@ -9,7 +9,7 @@
     "bugs": {
         "url": "https://github.com/SAP/open-ux-tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug+label%3Aserve-static-middleware"
     },
-    "version": "0.4.6",
+    "version": "0.4.7",
     "license": "Apache-2.0",
     "author": "@SAP/ux-tools-team",
     "main": "dist/index.js",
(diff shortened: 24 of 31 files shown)


packages/eslint-plugin-fiori-tools/README.md
```diff
@@ -15,9 +15,27 @@ Pnpm
...
+
+To use `manifest.json` specific rules update `eslint.config.mjs`:
+
+```
+import { defineConfig } from "eslint/config";
+import fiori from "@sap-ux/eslint-plugin-fiori-tools";
+
+export default defineConfig([
+  {
+    files: ["**/manifest.json"],
```
ManuelPastorini: The manifest.json must be located in webapp - any file named manifest.json but located elsewhere should be irrelevant. Thus, this should read
```suggestion
    files: ["webapp/manifest.json"],
```

packages/eslint-plugin-fiori-tools/README.md
```diff
@@ -15,9 +15,27 @@ Pnpm
...
+    plugins: { fiori },
```
ManuelPastorini: According to the naming convention from eslint for plugins and their packages and node-modules, this should read
```suggestion
    plugins: { @sap-ux/fiori-tools },
```
However, I'm not sure about the import statement in line 29 - maybe this allows to use a different name

packages/eslint-plugin-fiori-tools/src/rules/index.ts
```diff
@@ -0,0 +1,5 @@
+import flexEnabledRule from './flex-enabled';
```
ManuelPastorini: By taking over this rule from the poc, you introduced a lot of code (including dependencies) that is actually not shaped in a feasible way for production. I'd prefer not to take over anything from poc, but build correctly from scratch.

packages/eslint-plugin-fiori-tools/README.md
```diff
@@ -15,9 +15,27 @@ Pnpm
...
\ No newline at end of file
+- prodCode: contains rules for typescript and rules for production code from plugin eslint-plugin-fiori-custom
+
+To use `manifest.json` specific rules update `eslint.config.mjs`:
+
+```
+import { defineConfig } from "eslint/config";
+import fiori from "@sap-ux/eslint-plugin-fiori-tools";
+
+export default defineConfig([
```
ManuelPastorini: We should not ask consumers to import and call defineConfig in a specific way, but rather do this ourselves and provide a ready-to-use configuration (like defaultTS, defaultJS,...)

packages/eslint-plugin-fiori-tools/README.md
heimwege: That's not correct. The name of the webapp folder is defined in the UI5 yaml configuration and can be any name. `/webapp` is just the default (see https://ui5.github.io/cli/stable/pages/Configuration/#available-path-mappings).
That's why everybody should use the [getWebappPath](https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/project/ui5-config.ts#L15) method from `@sap-ux/project-access` instead of hard code to `webapp`.

packages/eslint-plugin-fiori-tools/README.md
AlinaGovoruhina: You can import the plugin with any name, then you need to refer to it in the config definition (in `extends`, `language` used from the plugin). I guess you propose to change the package name?

packages/eslint-plugin-fiori-tools/README.md
AlinaGovoruhina: If consumer wants to use the plugin they need eslint and an eslint config file. The config exported from the plugin is called "manifest" used in line 35 and defined here: https://github.com/SAP/open-ux-tools/blob/a290393d5d4163ff28a40e48e4e3b0b365620814/packages/eslint-plugin-fiori-tools/src/index.ts#L19-L22 
There can be multiple configs added in the future, also consumer can ignore it and defined a set of rules in a projects' eslint config themselves.

packages/eslint-plugin-fiori-tools/README.md
ManuelPastorini: Ok, I had asked @hitesh-parmar about this. Anyway, we need to tell in the eslint plugin which files the language is for - just defining all files named manifest.json regardless where they are located cannot be correct. @heimwege Should we set ui5.yaml, and then start reading everything from there? Or could this also be wrong?
@heimwege

packages/eslint-plugin-fiori-tools/README.md
ManuelPastorini: We should provide configurations in a way to make it as simple as possible to use them as they are for applications. If we do that correct, applications eslint.config.mjs can look as simple as
`export default config from "@sap-ux/eslint-plugin-fiori-tools/<config>";`
(where <config> should be the relevant name, depending on whether the app is being delivered from SAP, part of S/4 or own development from a customer).
This also allows apps to import the configuration, extend it, and override anything they want to - but we should not enforce this to be the standard case!

packages/eslint-plugin-fiori-tools/README.md
```diff
@@ -17,9 +17,24 @@ Pnpm
...
 
-To consume this module, add @sap-ux/eslint-plugin-fiori-tools to `extend` in your existing eslintrc configuration. You must specify one of the following configurations:
+To consume this module, add @sap-ux/eslint-plugin-fiori-tools plugin to your `eslint.config.mjs`. You must specify one of the following configurations:
 
 - defaultJS: contains rules for JavaScript for both prod and test code from plugin eslint-plugin-fiori-custom 
 - defaultTS: contains rules for typescript and rules for both prod and test code from plugin eslint-plugin-fiori-custom 
 - testcode: contains rules for typescript and rules for test code from plugin eslint-plugin-fiori-custom
-- prodCode: contains rules for typescript and rules for production code from plugin eslint-plugin-fiori-custom
\ No newline at end of file
+- prodCode: contains rules for typescript and rules for production code from plugin eslint-plugin-fiori-custom
```
marufrasully: This need to adjusted once this PR https://github.com/SAP/open-ux-tools/pull/3788 is merged

packages/eslint-plugin-fiori-tools/config/eslintrc-typescript.js
```diff
@@ -0,0 +1,47 @@
+const { defineConfig } = require('eslint/config');
+
+const typescriptEslint = require('@typescript-eslint/eslint-plugin');
+const tsParser = require('@typescript-eslint/parser');
+const js = require('@eslint/js');
+const tseslint = require('typescript-eslint');
```
marufrasully: this and `js` seems unsed!

packages/eslint-plugin-fiori-tools/docs/rules/require-width-including-column-header.md
```diff
@@ -0,0 +1,35 @@
+# width-including-column-header (`require-width-including-column-header`)
```
marufrasully: need fine tuning ....

packages/eslint-plugin-fiori-tools/src/index.ts
```diff
@@ -1,19 +1,169 @@
...
-        parser: '@typescript-eslint/parser' // override parser used in eslint-plugin-fiori-custom to support TS
+// Read package.json to get version
+const packageJson = JSON.parse(readFileSync(join(__dirname, '../package.json'), 'utf-8')) as {
+    name: string;
+    version: string;
+};
+
+// Plugin meta information (required for ESLint 9)
+export const meta = {
+    name: '@sap-ux/eslint-plugin-fiori-tools',
```
marufrasully: Probably it make sense to use a constant for this `'@sap-ux/eslint-plugin-fiori-tools'`

packages/eslint-plugin-fiori-tools/src/index.ts
```diff
@@ -1,19 +1,169 @@
...
+    fiori: new FioriLanguage()
+};
+
+const plugin: Plugin = {
+    meta: {
+        name: '@sap-ux/eslint-plugin-fiori-tools',
+        version: '0.0.1',
+        namespace: '@sap-ux/fiori-tools'
+    },
+    languages: {
```
marufrasully: Can we not re-use line line 21 here also?

packages/eslint-plugin-fiori-tools/src/index.ts
```diff
@@ -1,19 +1,169 @@
...
+    },
+    rules: rules as Record<string, any>,
+    processors: {}
+};
+
+// Named configs for easy consumption
+// These use getters for lazy loading to avoid loading config files at require time
+// The config files are templates meant for end users and have dependencies that
+// may not be available when the plugin itself is loaded
+export const configs = {
```
marufrasully: This need to be adjusted after this PR https://github.com/SAP/open-ux-tools/pull/3788

packages/eslint-plugin-fiori-tools/src/language/annotations/source-code.ts
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * Get alias to namespace map.
+     *
+     * @returns Alias map.
+     */
+    getAliasMap(): Record<string, string> {
+        if (this.aliasMap) {
+            return this.aliasMap;
+        }
+        this.aliasMap = {};
```
marufrasully: alias map need to yet be implemented?

packages/eslint-plugin-fiori-tools/src/language/annotations/source-code.ts
```diff
@@ -0,0 +1,163 @@
...
+    //         if (node.type === 'XMLDocument') {
+    //             return undefined;
+    //         }
+    //         return node.parent;
+    //     } else {
+    //         return undefined;
+    //     }
+    // }
+
+    /**
```
marufrasully: empty jsdoc

packages/eslint-plugin-fiori-tools/src/language/annotations/source-code.ts
```diff
@@ -0,0 +1,163 @@
...
+        return steps;
+    }
+
+    /**
+     * Get location of node or token.
+     *
+     * @param node - Node or token
+     * @returns Location
+     */
+    getLoc(node: AnyNode): SourceLocation {
```
marufrasully, line 109: As per eslint doc https://eslint.org/docs/latest/extend/languages#the-sourcecode-object we should implement `getRange`? too. Is this inherited?

There are some recommendations too
- lines
- getParent
- getAncestors
- getText

Maybe future improvements.

packages/eslint-plugin-fiori-tools/src/language/fiori-language.ts
```diff
@@ -0,0 +1,236 @@
...
+    constructor() {}
+    /**
+     *
+     * @param _languageOptions
+     */
+    validateLanguageOptions(_languageOptions: FioriLanguageOptions): void {}
+
+    visitorKeys = { ...xmlVisitorKeys, ...Object.fromEntries([...jsonVisitorKeys]), ...annotationVisitorKeys };
+
+    /**
```
marufrasully, line 72: Add jsdoc

packages/eslint-plugin-fiori-tools/src/language/fiori-language.ts
```diff
@@ -0,0 +1,236 @@
...
+     *
+     * @param file
+     * @param _context
+     * @returns
+     */
+    parse(file: File, _context: LanguageContext<FioriLanguageOptions>): ParseResult<FioriParseResultAst> {
+        const text = typeof file.body === 'string' ? file.body : new TextDecoder().decode(file.body);
+        const path = normalizePath(file.path);
+        const uri = pathToFileURL(path).toString();
+        DiagnosticCache.clear(uri); // currently this does not work well in ESLint extension, because it triggers this on window focus change
```
marufrasully: Abit more context is needed for this comment to understand it.

packages/eslint-plugin-fiori-tools/src/language/fiori-language.ts
```diff
@@ -0,0 +1,236 @@
...
+                        root: document
+                    }
+                }
+            };
+        }
+        throw new Error('Unsupported document type');
+    }
+
+    /**
+     *
```
marufrasully, line 209: Add jsdoc

packages/eslint-plugin-fiori-tools/src/language/rule-factory.ts
```diff
@@ -0,0 +1,112 @@
...
+import type { AnyNode } from '@humanwhocodes/momoa';
+import { FioriXMLSourceCode } from './xml/source-code';
+import type { XMLAstNode, XMLToken } from '@xml-tools/ast';
+import { FioriRuleDefinition } from '../types';
+import { FioriAnnotationSourceCode } from './annotations/source-code';
+import { AnyNode as AnyAnnotationNode } from '@sap-ux/odata-annotation-core';
+import { DiagnosticCache } from './diagnostic-cache';
+import { Diagnostic } from './diagnostics';
+
+type FioriRuleLanguageSpecificOptions = {
```
marufrasully: looks like not used. Maybe currently not used and will be used?

packages/eslint-plugin-fiori-tools/src/language/rule-factory.ts
```diff
@@ -0,0 +1,112 @@
...
+
+type AnnotationRuleContext<MessageIds extends string, RuleOptions extends unknown[]> = RuleContext<{
+    LangOptions: FioriLanguageOptions;
+    Code: FioriAnnotationSourceCode;
+    RuleOptions: RuleOptions;
+    Node: AnyAnnotationNode;
+    MessageIds: MessageIds;
+}>;
+
+export function createMixedRule<
```
marufrasully: missing jsdoc

packages/eslint-plugin-fiori-tools/src/project-context/linker/fe-v2.ts
```diff
@@ -0,0 +1,87 @@
+import { Manifest } from '@sap-ux/project-access';
```
marufrasully: un-used

packages/eslint-plugin-fiori-tools/src/project-context/linker/fe-v2.ts
```diff
@@ -0,0 +1,87 @@
...
+
+import { MetadataElement, parseIdentifier, parsePath, toFullyQualifiedPath } from '@sap-ux/odata-annotation-core';
+import { ServiceIndex } from '../parser/service';
```
marufrasully: un-used

packages/eslint-plugin-fiori-tools/src/project-context/linker/fe-v2.ts
```diff
@@ -0,0 +1,87 @@
...
+import { LinkerContext } from './types';
+import { get } from 'http';
```
marufrasully: un-used

packages/eslint-plugin-fiori-tools/src/project-context/linker/fe-v2.ts
```diff
@@ -0,0 +1,87 @@
...
+interface PageSettings {
+    navigationProperty?: string;
+    entitySet?: string;
+    component?: {
+        name: string;
+    };
+    pages?: { [name: string]: PageSettings };
+}
+
+function linkPage(
```
marufrasully, line 405: can we avoid side effect? Here `linkedApp` is modified.

packages/eslint-plugin-fiori-tools/src/project-context/linker/fe-v4.ts
```diff
@@ -0,0 +1,228 @@
+import { Manifest } from '@sap-ux/project-access';
+
```
marufrasully: Manifest and these parseIdentifier, parsePath, toFullyQualifiedPath  un-used

packages/eslint-plugin-fiori-tools/src/project-context/linker/fe-v4.ts
```diff
@@ -0,0 +1,228 @@
...
+import { MetadataElement, parseIdentifier, parsePath, toFullyQualifiedPath } from '@sap-ux/odata-annotation-core';
+import { buildAnnotationIndexKey } from '../parser';
+import { LinkerContext } from './types';
+import { get } from 'http';
```
marufrasully: un-used

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
```diff
@@ -0,0 +1,313 @@
...
+        this.diagnostics = [];
+        this.context = { projectType, fileCache };
+    }
+
+    public parse(
+        projectType: ProjectType,
+        artifacts: FoundFioriArtifacts,
+        fileCache: Map<string, string>
+    ): ParseResult {
+        performance.mark('project-index-build-start');
```
marufrasully: do we need it? Do we need to also run in production? 
If no, we should move it to a central place and place it behind an env variable for test.

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
```diff
@@ -0,0 +1,148 @@
...
+    toFullyQualifiedName,
+    Element,
+    MetadataElement
+} from '@sap-ux/odata-annotation-core';
+import { MetadataService } from '@sap-ux/odata-entity-model';
+import { CdsAnnotationProvider, getXmlServiceArtifacts } from '@sap-ux/fiori-annotation-api';
+import { ServiceArtifacts } from '@sap-ux/fiori-annotation-api/src/types';
+
+import { DocumentType } from '../types';
+import { uniformUrl } from '../utils';
```
marufrasully: un-used
also these 

```javascript
import { MetadataService } from '@sap-ux/odata-entity-model';
import { CdsAnnotationProvider, getXmlServiceArtifacts } from '@sap-ux/fiori-annotation-api';

and this `AnnotationFile` on line 2
```

packages/eslint-plugin-fiori-tools/src/project-context/project-context.ts
```diff
@@ -0,0 +1,258 @@
...
+import { ApplicationParser } from './parser';
+import { DiagnosticCache } from '../language/diagnostic-cache';
+import { LinkedModel, linkProject } from './linker';
+
+// Sync function for worker calls
+let artifactWorker: (file: string) => WorkerResult;
+
+/**
+ *
+ * @returns
```
marufrasully: empty jsdoc

packages/eslint-plugin-fiori-tools/src/project-context/project-context.ts
```diff
@@ -0,0 +1,258 @@
...
+    if (!workerPath) {
+        throw new Error(`Worker not found in any expected location: ${workerPaths.join(', ')}`);
+    }
+
+    return workerPath;
+}
+
+/**
+ *
+ * @returns
```
marufrasully: empty jsdoc

packages/eslint-plugin-fiori-tools/src/project-context/project-context.ts
```diff
@@ -0,0 +1,258 @@
...
+    artifactWorker = createSyncFn(workerPath, { timeout: 10000 });
+
+    return artifactWorker;
+}
+
+export type DocumentType = AnnotationFile | XMLDocument | DocumentNode;
+
+/**
+ *
+ */
```
marufrasully, line 74: add jsdoc

packages/eslint-plugin-fiori-tools/src/project-context/project-context.ts
```diff
@@ -0,0 +1,258 @@
...
+    private _linkedModel: LinkedModel;
+    public get linkedModel(): LinkedModel {
+        return this._linkedModel;
+    }
+
+    public readonly documents: Record<string, DocumentType> = {};
+
+    /**
+     *
+     * @param index
```
marufrasully: here too

packages/eslint-plugin-fiori-tools/src/rule-utils.ts
```diff
@@ -0,0 +1,292 @@
+import type { AnyNode, DocumentNode, MemberNode, ObjectNode, ValueNode } from '@humanwhocodes/momoa';
```
marufrasully: This file should not be under utils folder?

packages/eslint-plugin-fiori-tools/src/rules/sap-require-width-including-column-header.ts
```diff
@@ -0,0 +1,215 @@
...
+    REQUIRE_WIDTH_INCLUDING_COLUMN_HEADER_RULE_TYPE,
+    RequireWidthIncludingColumnHeaderDiagnostic
+} from '../language/diagnostics';
+
+export type RequireWidthIncludingColumnHeaderOptions = {
+    form: string;
+};
+
+const rule: FioriRuleDefinition = createMixedRule({
+    ruleId: 'sap-require-width-including-column-header',
```
marufrasully: we should re-use constant

packages/eslint-plugin-fiori-tools/src/rules/sap-width-including-column-header.ts
```diff
@@ -0,0 +1,215 @@
...
+        }
+        return {
+            ['target>element[name="Annotation"]'](node: Element) {
+                if (!lookup.has(node)) {
+                    return;
+                }
+
+                context.report({
+                    node: node,
+                    messageId: 'require-width-including-column-header'
```
marufrasully: here also we should re-use constant. Different I notice is `sap`, we should probably should have a function which accept const as input and remove `sap` prefix.

marufrasully (commented):
I know this is still work in progress - some comment so that I can track them later

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
voicis: logging/performance tracking/telemetry tracking topic is still open. This will be removed before merging.

packages/eslint-plugin-fiori-tools/src/rules/sap-width-including-column-header.ts
voicis: I'm not sure if we should add tight coupling with the rule name. There might be different messages that we want to use in the same rule.

packages/eslint-plugin-fiori-tools/src/project-context/linker/fe-v2.ts
voicis: This needs to be called recursively. We probably can avoid side effects, but it seems easier to build the object over time.

heimwege:
There are currently 79 new sonar issues reported. So my first review comment is: _Please reduce this to 0_ 😸

packages/preview-middleware-client/eslint.config.js
```diff
@@ -13,7 +13,8 @@ module.exports = [
             'coverage/**'
         ]
     },
-    ...fioriTools.configs['recommended-for-s4hana'],
+    // TODO: this is already not working, fix it separately
```
Klaus-Keller, line 16: @heimwege, can you please check this TODO, `recommended` sounds right to me.

heimwege: @devinea version 0.6.2 of eslint-plugin-fiori-tools used

```js
const { getResourcePaths } = require('./lib/utils');
const { sourceCodePath, testCodePath } = getResourcePaths();
//...
'files': [`./${sourceCodePath}/*.ts`, `./${sourceCodePath}/**/*.ts`],
```

whereas version 9.0.4 uses

`files: ['./webapp/**/*.js', './webapp/**/*.ts'],`

this means we can no longer use `eslint-plugin-fiori-tools` for `preview-middleware-client` where we define the middleware as being type application and [have a path mapping](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/ui5.yaml#L8) from webapp to `src` (which is also a bit wild but that's another discussion 🙈).
So in general path mappings are no longer lint supported with `eslint-plugin-fiori-tools` version `9.*`.

Should we change it back to what it was in `0.6.2`?
Also the utils.ts in the eslint plugin still exists but is not used anywhere ([link](https://github.com/SAP/open-ux-tools/blob/main/packages/eslint-plugin-fiori-tools/src/utils.ts)).
But instead of switching back to using `getResourcePaths` I would recommend to make use of the recently created [getPathMappings](https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/project/ui5-config.ts#L56C23-L56C38) from @sap-ux/project-access.

@voicis FYI as you added the todo.

Klaus-Keller: @heimwege could we fix this issue in a separate PR? Perhaps also use `@sap-ux/project-access`.

heimwege: I can add it to https://github.com/SAP/open-ux-tools/pull/3977 after this one has been merged

packages/eslint-plugin-fiori-tools/src/language/annotations/source-code.ts
marufrasully: jsdoc will be handled in different PR.

packages/eslint-plugin-fiori-tools/README.md
```diff
@@ -162,3 +162,13 @@ module.exports = [
...
+
+</br>
+
+✅ Set in the `recommended-for-s4hana` configuration
+| Rule | Description | Recommended |
+|------|-------------|:-----------:|
+| [sap-flex-enabled](docs/rules/sap-flex-enabled.md) | Ensures that the `flexEnabled` property is set to `true` in the `sap.ui5` section of the manifest file for applications using UI5 versions 1.56 or higher. | ✅ |
+| [sap-width-including-column-header](docs/rules/sap-width-including-column-header.md) | Ensures that small tables (less than 6 columns) include the `widthIncludingColumnHeader` property set to `true` for better column width calculation. | ✅ |
+| [sap-creation-mode-for-table](docs/rules/sap-creation-mode-for-table.md) | Validates that table creation mode (`createMode` in V2, `creationMode` in V4) is properly configured to ensure optimal user experience when creating new table entries. | ✅ |
+| [sap-disable-copy-to-clipboard](docs/rules/sap-disable-copy-to-clipboard.md) | Ensures that the `disableCopyToClipboard` property in any table is set to `false` or ommited from `tableSettings` in the manifest file for V4 applications (the default value is `false`) | ✅ |
```
devinea, line 174: Original plan 1 table for all rules and to add another column for the `recommended-for-s4hana` and specify if the rule enabled in that config

packages/eslint-plugin-fiori-tools/src/index.ts
```diff
@@ -17,114 +22,28 @@ const packageJson = JSON.parse(readFileSync(join(__dirname, '../package.json'),
...
-import sapUi5GlobalEval from './rules/sap-ui5-global-eval';
-import sapUi5LegacyFactories from './rules/sap-ui5-legacy-factories';
-import sapUi5LegacyJquerysapUsage from './rules/sap-ui5-legacy-jquerysap-usage';
-import sapUi5NoPrivateProp from './rules/sap-ui5-no-private-prop';
-import sapUsageBasemastercontroller from './rules/sap-usage-basemastercontroller';
-
 // Plugin meta information (required for ESLint 9)
 export const meta = {
-    name: '@sap-ux/eslint-plugin-fiori-tools',
+    name: PACKAGE_NAME,
```
devinea, line 27:
```suggestion
    name: packageJson.name,
```

packages/eslint-plugin-fiori-tools/src/utils/version.ts
```diff
@@ -0,0 +1,21 @@
...
+export function isLowerThanMinimalUi5Version(
+    providedUi5Version: MinUI5Version,
+    minUi5Version: Partial<Pick<MinUI5Version, 'patch'>> & Pick<MinUI5Version, 'major' | 'minor'>
+): boolean {
+    return (
+        providedUi5Version.major < minUi5Version.major ||
+        (providedUi5Version.major === minUi5Version.major && providedUi5Version.minor < minUi5Version.minor) ||
+        (providedUi5Version.major === minUi5Version.major &&
+            providedUi5Version.minor === minUi5Version.minor &&
+            (providedUi5Version?.patch ?? 0) < (minUi5Version?.patch ?? 0))
```
devinea, line 19: Can use https://www.npmjs.com/package/semver ?

devinea (approved):
Checked mainly the `eslint-plugin-fiori-tools` updates to existing files. 
a few small comments. But overall looks good. Thanks

packages/eslint-plugin-fiori-tools/src/utils/version.ts
marufrasully: @devinea there are some missing jsdoc which I plan to cover them in a separate PR. Currently this commit https://github.com/SAP/open-ux-tools/pull/3998/changes/82f72d0f5ba8a936fa6ddad72c2dac66e94f6d67#diff-7d578f788ccf97272dbd21105566dab86ff1b5c74970f5dd729b6430166c13af which later will be a clean PR.  Is it fine if we cover these point in that PR? I want to close this big PR and do improvement in an separate PR.

devinea: Yeah sure 👍

PR #3760, opened 2025-10-21
feat(create): auto generate readme on build

heimwege:
The readme.md of the create command will now be auto-generated with each build. It uses the description from the commands and subcommands. No need to manually update the readme.md ever again 🥳

PR #3779, opened 2025-10-28
feat(fiori-docs-embeddings): add @sap-ux/create/README.md as resource

heimwege:
next step after #3760

adds the following two new sources:
- @sap-ux/create README.md
- @sap/ux-ui5-tooling README.md

files are taken into account
image

### DIFF
--- a/.changeset/old-worms-jam.md
+++ b/.changeset/old-worms-jam.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/ui-components': patch
----
-
-Update building block icon svg
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -30,7 +30,7 @@ jobs:
             fail-fast: false
             matrix:
                 ui5-version: ${{ fromJson(needs.ui5-versions.outputs.ui5_versions) }}
-        runs-on: ubuntu-latest
+        runs-on: [ubuntu-latest, windows-2025]
         needs: ui5-versions
         timeout-minutes: 30
         steps:
--- a/examples/ui-prompting-examples/CHANGELOG.md
+++ b/examples/ui-prompting-examples/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux-private/ui-prompting-examples
 
+## 0.4.7
+
+### Patch Changes
+
+-   Updated dependencies [afb4be5]
+    -   @sap-ux/ui-components@1.26.15
+    -   @sap-ux/ui-prompting@0.5.6
+
 ## 0.4.6
 
 ### Patch Changes
--- a/examples/ui-prompting-examples/package.json
+++ b/examples/ui-prompting-examples/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux-private/ui-prompting-examples",
-    "version": "0.4.6",
+    "version": "0.4.7",
     "description": "This project contains UI storybook stories with exampleS with prompt ui and FPM based building blocks.",
     "license": "Apache-2.0",
     "private": true,
--- a/packages/ui-components/CHANGELOG.md
+++ b/packages/ui-components/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/ui-components
 
+## 1.26.15
+
+### Patch Changes
+
+-   afb4be5: Update building block icon svg
+
 ## 1.26.14
 
 ### Patch Changes
--- a/packages/ui-components/package.json
+++ b/packages/ui-components/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/ui-components",
-    "version": "1.26.14",
+    "version": "1.26.15",
     "license": "Apache-2.0",
     "description": "SAP UI Components Library",
     "repository": {
--- a/packages/ui-prompting/CHANGELOG.md
+++ b/packages/ui-prompting/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/ui-prompting
 
+## 0.5.6
+
+### Patch Changes
+
+-   Updated dependencies [afb4be5]
+    -   @sap-ux/ui-components@1.26.15
+
 ## 0.5.5
 
 ### Patch Changes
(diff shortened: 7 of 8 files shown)


packages/fiori-docs-embeddings/package.json
```diff
@@ -20,7 +21,9 @@
         "lint": "eslint . --ext .ts",
         "lint:fix": "eslint . --ext .ts --fix"
     },
-    "dependencies": {},
+    "dependencies": {
+        "@sap-ux/create": "workspace:*"
```
devinea: Move to devDeps

packages/fiori-docs-embeddings/package.json
```diff
@@ -5,7 +5,8 @@
     "private": false,
     "scripts": {
         "build-compile": "tsc --noEmit",
-        "build": "npm-run-all -l -s build-compile update-embeddings-script",
+        "copy-create-readme": "node -e \"require('fs').copyFileSync('node_modules/@sap-ux/create/README.md', 'data_local/create-README.md')\"",
+        "build": "npm run copy-create-readme && npm-run-all -l -s build-compile update-embeddings-script",
```
devinea:
```suggestion
        "copy-create-readme": "node -e \"require('fs').copyFileSync('node_modules/@sap-ux/create/README.md', 'data_local/create-README.md')\"",
        "build": "npm-run-all -l -p build-compile copy-create-readme -s update-embeddings-script",
```
`npm-run-all` can be used to avoid `&&`. It also allows running scripts on serial or parallel as appropriate.
Using `copyfiles` module like https://github.com/SAP/open-ux-tools/blob/72ffbcc802ff46a85b796149986c75bcbca92f9f/packages/ui-prompting/package.json#L25 is a bid easier to read.

heimwege: Problem with `copyfiles` is that it does not support renaming. Then you would do
```json
"copy-create-readme": "copyfiles \"node_modules/@sap-ux/create/README.md\" data_local && ren data_local\\README.md create-README.md",
```
but `ren` is windows specific, etc. etc. so you use yet another package 🤷🏻

packages/fiori-docs-embeddings/package.json
heimwege: done

packages/fiori-docs-embeddings/package.json
heimwege: @devinea I could add a tiny script though. Now that we have a scripts folder anyway... That would also beautify the run script.

packages/fiori-docs-embeddings/scripts/load-readme-from-npm.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Fetch the README of a specified npm package.
+ *
+ * @param packageName {string} - The name of the npm package.
+ * @return {Promise<*|null>}
+ */
+async function getPackageReadme(packageName) {
+    // Handle scoped packages by URL-encoding the slash
+    const encodedName = packageName.replace('/', '%2F');
```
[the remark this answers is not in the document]
heimwege: But that's exactly what I want 🤷🏻 Is this a false positive @devinea ?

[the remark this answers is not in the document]
heimwege: anyway, seems to be ok to replace all

packages/fiori-docs-embeddings/package.json
```diff
@@ -5,7 +5,9 @@
     "private": false,
     "scripts": {
         "build-compile": "tsc --noEmit",
-        "build": "npm-run-all -l -s build-compile update-embeddings-script",
+        "copy-create-readme": "node -e \"require('fs').copyFileSync('node_modules/@sap-ux/create/README.md', 'data_local/create-README.md')\"",
+        "fetch-ux-ui5-tooling-readme": "node scripts/load-readme-from-npm.js @sap/ux-ui5-tooling",
```
devinea:
```suggestion
        "fetch-ux-ui5-tooling-readme": "tsx src/scripts/load-readme-from-npm.ts @sap/ux-ui5-tooling",
```
Could we use typescript for the scripts like the others?

devinea (commented):
Thanks @heimwege 
1 small comment. Other than that look good.

packages/fiori-docs-embeddings/package.json
devinea: @heimwege yeah sorry for the confusion. go for `src/scripts`

heimwege: fixed in 73bd8b36ab80126a0995db0bc82ef470f12eece1

packages/create/README.md
```diff
@@ -1,7 +1,7 @@
 [![Changelog](https://img.shields.io/badge/changelog-8A2BE2)](https://github.com/SAP/open-ux-tools/blob/main/packages/create/CHANGELOG.md) [![Github repo](https://img.shields.io/badge/github-repo-blue)](https://github.com/SAP/open-ux-tools/tree/main/packages/create)
 # [`@sap-ux/create`](https://github.com/SAP/open-ux-tools/tree/main/packages/create) CLI Reference
 
-Configure features for SAP Fiori applications and projects (@sap-ux/create version 0.13.161).
+Configure features for SAP Fiori applications and projects (@sap-ux/create version 0.13.162).
```
heimwege: @devinea I think a side effect is that the autofix will adjust the create readme for each depending package which means the create group is added as reviewer for each PR. Not sure if we really want to have this or if we remove the version number from the create README.md

devinea: @heimwege yeah removing the version might be best to reduce the noise. Thanks

packages/create/src/cli/index.ts
```diff
@@ -44,9 +44,7 @@ function getCommanderProgram(): Command {
     const logger = getLogger();
     const program = new Command();
     const version = getVersion();
-    program.description(
-        `Configure features for SAP Fiori applications and projects (@sap-ux/create version ${version}).`
-    );
+    program.description(`Configure features for SAP Fiori applications and projects.`);
```
devinea, line 47: Thanks 👍

PR #3787, opened 2025-10-30
fix(backend-proxy-middleware): wrong 'path' and 'pathReplace' in case of appStudio full destination

heimwege:
The proxy option `pathRewrite` was set before potential adjustments for app studio have been done in 'enhanceConfigsForDestination' leading to a wrong `pathRewrite` being used.


IainSAP (approved):
Thanks @heimwege 

- Fix as expected
- Tests added to cover

Issue #3789, opened 2025-10-31
FEATURE - Enable Adaptation Editor for CF projects


nikmace:
_As a **developer**,
I want to **create a CF ADP**,
so that I **preview and open the Adaptation Editor to edit application/s**_

### Description

_To completely transition to open source, we need to enable the support of CF ADP projects. Meaning, you can create a CF ADP project, you can preview it, and also start the Adaptation Editor. After that, we can have a unified development experience between all ADP types_

### Technical Design
_What modules need to be created or enhanced?_ 
New middleware `@sap-ux/cf-oauth-middleware` needs to be added to help the CF ADP projects load the OData.
_How would the functionality be exposed and how can it be consumed?_

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have created a CF ADP with the correct structure
**when** I open the editor
**then** the editor opens and all resources load, and we are in ui adaptation mode

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #3790, opened 2025-10-31
feat: Enable Adaptation Editor for CF projects

nikmace:
Feat for #3789.
- Adds a new middleware that adds a token for each request to OData for Cloud Foundry ADP.
- Changes the CF project structure to support the new middleware and work in preview and in Adaptation Editor.
- Adjust preview-middleware code to support CF projects.

### DIFF
--- a/packages/abap-deploy-config-sub-generator/test/fixtures/sample/package.json
+++ b/packages/abap-deploy-config-sub-generator/test/fixtures/sample/package.json
@@ -20,16 +20,8 @@
     "start-mock": "fiori run --open 'test/flpSandboxMockServer.html#testprojectlropv2-tile'"
   },
   "devDependencies": {
-    "@ui5/cli": "^2.14.1",
-    "@ui5/fs": "^2.0.6",
-    "@ui5/logger": "^2.0.1",
+    "@ui5/cli": "^4.0.0",
     "@sap/ux-ui5-tooling": "1",
-    "rimraf": "3.0.2",
-    "@sap/ux-specification": "UI5-1.71"
-  },
-  "ui5": {
-    "dependencies": [
-      "@sap/ux-ui5-tooling"
-    ]
+    "rimraf": "3.0.2"
   }
 }
--- a/packages/abap-deploy-config-writer/test/sample/test.ui5.lib/package.json
+++ b/packages/abap-deploy-config-writer/test/sample/test.ui5.lib/package.json
@@ -3,6 +3,6 @@
     "version": "1.0.0",
     "private": true,
     "devDependencies": {
-        "@ui5/cli": "2.9.9"
+        "@ui5/cli": "^4.0.0"
     }
 }
--- a/packages/abap-deploy-config-writer/test/sample/test.ui5.typescript.library1/package.json
+++ b/packages/abap-deploy-config-writer/test/sample/test.ui5.typescript.library1/package.json
@@ -3,6 +3,6 @@
     "version": "1.0.0",
     "private": true,
     "devDependencies": {
-        "@ui5/cli": "2.9.9"
+        "@ui5/cli": "^4.0.0"
     }
 }
--- a/packages/abap-deploy-config-writer/test/unit/__snapshots__/index.test.ts.snap
+++ b/packages/abap-deploy-config-writer/test/unit/__snapshots__/index.test.ts.snap
@@ -67,24 +67,19 @@ builder:
   },
   "package.json": Object {
     "contents": "{
-  \\"name\\": \\"test.ui5.lib\\",
-  \\"version\\": \\"1.0.0\\",
-  \\"private\\": true,
-  \\"devDependencies\\": {
-    \\"@ui5/cli\\": \\"2.9.9\\",
-    \\"rimraf\\": \\"^5.0.5\\",
-    \\"ui5-task-flatten-library\\": \\"^3.1.0\\"
-  },
-  \\"scripts\\": {
-    \\"deploy\\": \\"npm run build && fiori deploy --config deploy-config.yaml\\",
-    \\"undeploy\\": \\"npm run build && fiori undeploy --config deploy-config.yaml\\",
-    \\"deploy-test\\": \\"npm run build && fiori deploy --config deploy-config.yaml --testMode true\\"
-  },
-  \\"ui5\\": {
-    \\"dependencies\\": [
-      \\"ui5-task-flatten-library\\"
-    ]
-  }
+    \\"name\\": \\"test.ui5.lib\\",
+    \\"version\\": \\"1.0.0\\",
+    \\"private\\": true,
+    \\"devDependencies\\": {
+        \\"@ui5/cli\\": \\"^4.0.0\\",
+        \\"rimraf\\": \\"^5.0.5\\",
+        \\"ui5-task-flatten-library\\": \\"^3.1.0\\"
+    },
+    \\"scripts\\": {
+        \\"deploy\\": \\"npm run build && fiori deploy --config deploy-config.yaml\\",
+        \\"undeploy\\": \\"npm run build && fiori undeploy --config deploy-config.yaml\\",
+        \\"deploy-test\\": \\"npm run build && fiori deploy --config deploy-config.yaml --testMode true\\"
+    }
 }
 ",
     "state": "modified",
@@ -96,24 +91,19 @@ exports[`generate Generate deployment configs: test.ui5.typescript.library1 1`] 
 Object {
   "package.json": Object {
     "contents": "{
-  \\"name\\": \\"test.ui5.typescript.library1\\",
-  \\"version\\": \\"1.0.0\\",
-  \\"private\\": true,
-  \\"devDependencies\\": {
-    \\"@ui5/cli\\": \\"2.9.9\\",
-    \\"rimraf\\": \\"^5.0.5\\",
-    \\"ui5-task-flatten-library\\": \\"^3.1.0\\"
-  },
-  \\"scripts\\": {
-    \\"deploy\\": \\"npm run build && fiori deploy --config ui5-deploy.yaml\\",
-    \\"undeploy\\": \\"npm run build && fiori undeploy --config ui5-deploy.yaml\\",
-    \\"deploy-test\\": \\"npm run build && fiori deploy --config ui5-deploy.yaml --testMode true\\"
-  },
-  \\"ui5\\": {
-    \\"dependencies\\": [
-      \\"ui5-task-flatten-library\\"
-    ]
-  }
+    \\"name\\": \\"test.ui5.typescript.library1\\",
+    \\"version\\": \\"1.0.0\\",
+    \\"private\\": true,
+    \\"devDependencies\\": {
+        \\"@ui5/cli\\": \\"^4.0.0\\",
+        \\"rimraf\\": \\"^5.0.5\\",
+        \\"ui5-task-flatten-library\\": \\"^3.1.0\\"
+    },
+    \\"scripts\\": {
+        \\"deploy\\": \\"npm run build && fiori deploy --config ui5-deploy.yaml\\",
+        \\"undeploy\\": \\"npm run build && fiori undeploy --config ui5-deploy.yaml\\",
+        \\"deploy-test\\": \\"npm run build && fiori deploy --config ui5-deploy.yaml --testMode true\\"
+    }
 }
 ",
     "state": "modified",
--- a/packages/adp-tooling/templates/cf/package.json
+++ b/packages/adp-tooling/templates/cf/package.json
@@ -16,12 +16,6 @@
 		"type": "git",
 		"build": "ui5.js build --verbose --include-task generateCachebusterInfo"
 	},
-	"ui5": {
-		"dependencies": [
-			"@sap/ui5-builder-webide-extension",
-			"@ui5/task-adaptation"
-		]
-	},
 	"devDependencies": {
 		"@sap/ui5-builder-webide-extension": "^1.1.9",
 		"@sap-ux/backend-proxy-middleware-cf": "^0.0.1",
--- a/packages/adp-tooling/test/unit/writer/__snapshots__/cf.test.ts.snap
+++ b/packages/adp-tooling/test/unit/writer/__snapshots__/cf.test.ts.snap
@@ -102,12 +102,6 @@ dist/
 		\\"type\\": \\"git\\",
 		\\"build\\": \\"ui5.js build --verbose --include-task generateCachebusterInfo\\"
 	},
-	\\"ui5\\": {
-		\\"dependencies\\": [
-			\\"@sap/ui5-builder-webide-extension\\",
-			\\"@ui5/task-adaptation\\"
-		]
-	},
 	\\"devDependencies\\": {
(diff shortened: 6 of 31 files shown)


IainSAP:
@nikmace Using CF to discover systems and connect is something we already support using axios-extension and https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-inquirer/src/prompts/datasources/sap-system/abap-on-btp/questions.ts#L216. Is this not sufficent for this use case. Any connectivity features should be added to axios extension, but I believe there is nothing new here?

nikmace:
> @nikmace Using CF to discover systems and connect is something we already support using axios-extension and https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-inquirer/src/prompts/datasources/sap-system/abap-on-btp/questions.ts#L216. Is this not sufficent for this use case. Any connectivity features should be added to axios extension, but I believe there is nothing new here?

Our use case is completely different. We are trying to support Cloud Foundry Adaptation Projects in the Adaptation Editor - that is the goal. We are not having issues with connectivity, and I don't see a reason to enhance axios-extension because we don't need to use any methods from it like we do with AbapServiceProvider (ABAP and Cloud Foundry are two different things). 

All we need to do is get the service keys from business service, make a call to fetch an authorization token, and attach it to the request headers. Then, authentication is going to happen programatically, service-to-service. For example, for OData, which is linked to a destination and needs authorization. Maybe I don't understand something but I can't see how enhacing axios-extension is going to help us.

We are talking about different scenarios, in my opinion. Let's follow up during the meeting. 👍🏻

packages/preview-middleware/src/base/flp.ts
heimwege: Question: Why is `initAdp` not a function of the class `FlpSandbox`? As it seems to rely on a specific instance `flp: FlpSandbox` this could be improved by just making it a function of that instance so that `this` can be used. Also `flp` as property name is misleading because there's a config property of the preview-middleware named `flp` as well.
I know all of this is not related to this PR but something I learned doing the review 🐱

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1152,6 +1153,59 @@ function serializeUi5Configuration(config: Map<string, string>): string {
...
 
+/**
+ * Configure RTA (Runtime Adaptation) for the FLP sandbox.
+ *
+ * @param flp FlpSandbox instance
+ * @param layer UI5 flex layer
+ * @param variantId variant identifier
+ * @param isCloud whether this is a cloud project
+ */
+function configureRta(flp: FlpSandbox, layer: UI5FlexLayer, variantId: string, isCloud: boolean): void {
```
heimwege: nit-pick: can we name `flp` differently? I was confused because of the same property name of the `flp` config option (and there's an `rta` config option as well, but not as child of `flp` but as peer on the same level). To be honest I find the whole class name `FlpSandbox` misleading because that is not an FLPSandbox but a previewMiddleware but that's a different story 🙈 
Maybe `flpInstance` or something? Or why not just pass the `rta` config part as this is the only thing used in that function?

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1152,6 +1153,59 @@ function serializeUi5Configuration(config: Map<string, string>): string {
...
+ * @param rootProject reference to the project
+ * @returns parsed descriptor variant
+ * @throws Error if manifest.appdescr_variant is not found
+ */
+async function loadAppVariant(rootProject: ReaderCollection): Promise<DescriptorVariant> {
+    const appVariant = await rootProject.byPath('/manifest.appdescr_variant');
+    if (!appVariant) {
+        throw new Error('ADP configured but no manifest.appdescr_variant found.');
+    }
+    return JSON.parse(await appVariant.getString()) as DescriptorVariant;
```
heimwege: Depending on the content of the string this can throw an exception or be an empty object, so casting to `DescriptorVariant` might be a bit too optimistic

packages/backend-proxy-middleware-cf/src/types.ts
```diff
@@ -0,0 +1,45 @@
...
+     */
+    debug?: boolean;
+}
+
+export interface MiddlewareParameters<T> {
+    resources: object;
+    options: {
+        configuration: T;
+    };
+}
```
heimwege: Why do we have a generic type for exactly one usage that passes `CfOAuthMiddlewareConfig` defined in the very same file? Can't we simplify to 

```suggestion
export interface MiddlewareParameters {
    resources: object;
    options: {
        configuration: CfOAuthMiddlewareConfig;
    };
}
```

Also there exists `import type { MiddlewareParameters } from '@ui5/server';` to this whole type might not be needed at all. Seems like we should adjust backend-proxy-middleware as well 🙈

packages/backend-proxy-middleware-cf/src/middleware.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * UI5 middleware for proxying requests to Cloud Foundry destinations with OAuth2 authentication.
+ * Supports one destination URL with multiple OData source paths.
+ *
+ * @param {MiddlewareParameters<CfOAuthMiddlewareConfig>} params - Input parameters for UI5 middleware.
+ * @param {CfOAuthMiddlewareConfig} params.options - Configuration options.
+ * @returns {Promise<RequestHandler>} Express middleware handler.
+ */
+module.exports = async ({ options }: MiddlewareParameters<CfOAuthMiddlewareConfig>): Promise<RequestHandler> => {
```
heimwege, line 19: I think this can be simplified to using

```ts
import type { MiddlewareParameters } from '@ui5/server';
```

and then 

```ts
module.exports = async (options: MiddlewareParameters<CfOAuthMiddlewareConfig>): Promise<RequestHandler> => {
```

and don't forget to add inclusion `"../../types/ui5.d.ts",` to tsconfig.json

packages/backend-proxy-middleware-cf/src/types.ts
heimwege: https://github.com/SAP/open-ux-tools/pull/3870

packages/backend-proxy-middleware-cf/README.md
```diff
@@ -0,0 +1,151 @@
+# `@sap-ux/backend-proxy-middleware-cf`
```
heimwege: Badges are missing that come with all other readme files that are published to nmpjs

```md
﻿[![Changelog](https://img.shields.io/badge/changelog-8A2BE2)](https://github.com/SAP/open-ux-tools/blob/main/packages/backend-proxy-middleware-cf/CHANGELOG.md) [![Github repo](https://img.shields.io/badge/github-repo-blue)](https://github.com/SAP/open-ux-tools/tree/main/packages/backend-proxy-middleware-cf)
```

packages/backend-proxy-middleware-cf/src/middleware.ts
```diff
@@ -0,0 +1,33 @@
...
+        transports: [new UI5ToolingTransport({ moduleName: 'backend-proxy-middleware-cf' })]
+    });
+
+    await validateConfig(config, logger);
+
+    const tokenProvider = await createTokenProvider(config, logger);
+    const router = setupProxyRoutes(config.paths, config.url, tokenProvider, logger);
+    logger.info(`Backend proxy middleware (CF) initialized: url=${config.url}, paths=${config.paths.join(', ')}`);
+
+    return router as unknown as RequestHandler;
```
heimwege: can we move the logging into `createTokenProvider` and just do

```ts
return setupProxyRoutes(config.paths, config.url, tokenProvider, logger);
```

and what's the difference between

```ts
logger.info(`Backend proxy middleware (CF) initialized: url=${config.url}, paths=${config.paths.join(', ')}`);
```

and the individual log messages

```ts
logger.info(`Registered proxy for path: ${path} -> ${destinationUrl}`);
```
(from `proxy.ts`)

packages/backend-proxy-middleware-cf/src/proxy.ts
```diff
@@ -0,0 +1,101 @@
...
+ * @returns {Options} Proxy options configuration.
+ */
+export function createProxyOptions(targetUrl: string, logger: ToolsLogger): Options {
+    return {
+        target: targetUrl,
+        changeOrigin: true,
+        pathRewrite: (strippedPath: string, req: IncomingMessage) => {
+            // Express router.use() strips the matched path from req.url,
+            // use originalUrl to get the full path before Express stripped it
+            const originalUrl = (req as any).originalUrl || req.url || strippedPath;
```
heimwege: Please don't use type `any`. You can either use `unknown` here or

```ts
const originalUrl = 'originalUrl' in req ? req.originalUrl : (req.url ?? strippedPath);
```

but it seems originalUrl is in express as well as connect so you could also

```ts
export type EnhancedIncomingMessage = (IncomingMessage & Pick<Request, 'originalUrl'>) | connect.IncomingMessage;
```

like it is done in the backend-proxy-middleware.

heimwege (commented):
See comments. I just had a look at the `backend-proxy-middleware-cf` and `preview-middleware` changes. Can't really comment on the adp packages.

packages/preview-middleware/src/base/flp.ts
nikmace: Changed in: [2aafca4](https://github.com/SAP/open-ux-tools/pull/3790/commits/2aafca429b1bb9d625ecef7eaadd67ad477d3064)

packages/backend-proxy-middleware-cf/src/types.ts
nikmace: Changed in: [2aafca4](https://github.com/SAP/open-ux-tools/pull/3790/commits/2aafca429b1bb9d625ecef7eaadd67ad477d3064)

packages/backend-proxy-middleware-cf/src/middleware.ts
nikmace: Changed in: [2aafca4](https://github.com/SAP/open-ux-tools/pull/3790/commits/2aafca429b1bb9d625ecef7eaadd67ad477d3064)

packages/backend-proxy-middleware-cf/README.md
nikmace: Added in: [2aafca4](https://github.com/SAP/open-ux-tools/pull/3790/commits/2aafca429b1bb9d625ecef7eaadd67ad477d3064)

packages/backend-proxy-middleware-cf/src/middleware.ts
nikmace: Changed in: [2aafca4](https://github.com/SAP/open-ux-tools/pull/3790/commits/2aafca429b1bb9d625ecef7eaadd67ad477d3064)

Left only the individual log messages

packages/backend-proxy-middleware-cf/src/proxy.ts
nikmace: Changed in: [2aafca4](https://github.com/SAP/open-ux-tools/pull/3790/commits/2aafca429b1bb9d625ecef7eaadd67ad477d3064)

packages/preview-middleware/src/base/flp.ts
nikmace: Renamed in: [2aafca4](https://github.com/SAP/open-ux-tools/pull/3790/commits/2aafca429b1bb9d625ecef7eaadd67ad477d3064)

packages/backend-proxy-middleware-cf/src/middleware.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    const logger = new ToolsLogger({
+        logLevel: config.debug ? LogLevel.Debug : LogLevel.Info,
+        transports: [new UI5ToolingTransport({ moduleName: 'backend-proxy-middleware-cf' })]
+    });
+
+    await validateConfig(config, logger);
+
+    const tokenProvider = await createTokenProvider(config, logger);
+    return setupProxyRoutes(config.paths, config.url, tokenProvider, logger) as unknown as RequestHandler;
```
heimwege: The cast is not needed because Express's Router type extends `RequestHandler`

```suggestion
    return setupProxyRoutes(config.paths, config.url, tokenProvider, logger);
```

packages/backend-proxy-middleware-cf/package.json
```diff
@@ -0,0 +1,55 @@
...
+        "@types/supertest": "2.0.12",
+        "express": "4.21.2",
+        "nock": "13.4.0",
+        "supertest": "7.1.4",
+        "connect": "^3.7.0",
+        "@types/connect": "^3.4.38"
+    },
+    "ui5": {
+        "dependencies": []
+    },
```
heimwege: I think that one can be deleted. This is a very old and was need for UI5 cli v2.x ([link](https://ui5.github.io/cli/v3/updates/migrate-v3/#changes-to-dependency-configuration))

packages/preview-middleware/src/base/flp.ts
nikmace: Changed In: [4bcb65b](https://github.com/SAP/open-ux-tools/pull/3790/commits/4bcb65bb194f95cfc68ea690d110df02cf141834)

packages/backend-proxy-middleware-cf/src/middleware.ts
nikmace: Changed in: [8a156d6](https://github.com/SAP/open-ux-tools/pull/3790/commits/8a156d6655ea11fdea77da9f11ef58327e6e84f0)

packages/backend-proxy-middleware-cf/package.json
nikmace: Changed in: [8a156d6](https://github.com/SAP/open-ux-tools/pull/3790/commits/8a156d6655ea11fdea77da9f11ef58327e6e84f0)

voicis (approved):
- Preview middleware changes looks good
- Good code coverage
- Changeset present
- Integration tests passing https://github.com/SAP/open-ux-tools/actions/runs/19763181417

packages/backend-proxy-middleware-cf/src/token/provider.ts
```diff
@@ -0,0 +1,111 @@
...
+     * @returns {Promise<string>} The access token.
+     */
+    private async getAccessToken(): Promise<string> {
+        if (this.token && Date.now() < this.tokenExpiry) {
+            return this.token;
+        }
+
+        // If a token fetch is already in progress, wait for it
+        if (this.tokenFetchPromise) {
+            return this.tokenFetchPromise;
```
heimwege, line 46: Do you really want to return the `tokenFetchPromise` here or the token? `return await this.tokenFetchPromise`?

nikmace: Returning the promise directly is more efficient than awaiting it, as it avoids creating an extra promise wrapper. The promise will resolve/reject correctly. We're not catching errors here, so `await` isn't needed. The `await` is only needed in the try block below, where we actually need the token value and want to clear the promise in the finally block.

nikmace:
After this PR is merged, I've prepared a PR to fix ADP UI integration tests and run the Jenkins job (1280) against the branch in the private repo.

PR #3794, opened 2025-10-31
feat: save value list reference metadata

voicis:
#389 

Collect a list of services referenced in `Common.ValueListReferences` annotation and save them in `localService` folder.

The file system path is built based on the service path. It starts relative to the main service (the one where the annotation with reference is) and last segments are from the target path because we do not include the parameters in the path and metadata for different targets is different. However, data is the same even when the service is provided different parameters, so mock data can be reused unlike metadata.

Example structure
```
localService
  metadata.xml
  srvd_f4
    dmo
      i_agency/001
        data.json
        TravelType/AgencyID
           metadata.xml      
```

### DIFF
--- a/packages/abap-deploy-config-inquirer/CHANGELOG.md
+++ b/packages/abap-deploy-config-inquirer/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/abap-deploy-config-inquirer
 
+## 1.6.49
+
+### Patch Changes
+
+-   @sap-ux/inquirer-common@0.9.6
+
 ## 1.6.48
 
 ### Patch Changes
--- a/packages/abap-deploy-config-inquirer/package.json
+++ b/packages/abap-deploy-config-inquirer/package.json
@@ -6,7 +6,7 @@
         "url": "https://github.com/SAP/open-ux-tools.git",
         "directory": "packages/abap-deploy-config-inquirer"
     },
-    "version": "1.6.48",
+    "version": "1.6.49",
     "license": "Apache-2.0",
     "main": "dist/index.js",
     "scripts": {
--- a/packages/abap-deploy-config-sub-generator/CHANGELOG.md
+++ b/packages/abap-deploy-config-sub-generator/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/abap-deploy-config-sub-generator
 
+## 0.1.98
+
+### Patch Changes
+
+-   @sap-ux/abap-deploy-config-inquirer@1.6.49
+-   @sap-ux/deploy-config-generator-shared@0.1.53
+
 ## 0.1.97
 
 ### Patch Changes
--- a/packages/abap-deploy-config-sub-generator/package.json
+++ b/packages/abap-deploy-config-sub-generator/package.json
@@ -6,7 +6,7 @@
         "url": "https://github.com/SAP/open-ux-tools.git",
         "directory": "packages/abap-deploy-config-sub-generator"
     },
-    "version": "0.1.97",
+    "version": "0.1.98",
     "license": "Apache-2.0",
     "main": "generators/app/index.js",
     "scripts": {
--- a/packages/adp-flp-config-sub-generator/CHANGELOG.md
+++ b/packages/adp-flp-config-sub-generator/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/adp-flp-config-sub-generator
 
+## 0.1.119
+
+### Patch Changes
+
+-   @sap-ux/adp-tooling@0.18.8
+-   @sap-ux/inquirer-common@0.9.6
+-   @sap-ux/flp-config-inquirer@0.4.67
+
 ## 0.1.118
 
 ### Patch Changes
--- a/packages/adp-flp-config-sub-generator/package.json
+++ b/packages/adp-flp-config-sub-generator/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/adp-flp-config-sub-generator",
     "description": "Generator for adding FLP configuration to an Adaptation Project",
-    "version": "0.1.118",
+    "version": "0.1.119",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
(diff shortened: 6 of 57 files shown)


packages/axios-extension/src/abap/abap-service-provider.ts
```diff
@@ -288,6 +294,83 @@ export class AbapServiceProvider extends ServiceProvider {
...
+
+    /**
+     * Collects ValueListReferences annotation values from the given metadata and annotation files.
+     *
+     * @param servicePath - The service path to which the value list references belong
+     * @param metadata - The metadata of the service
+     * @param annotations - The annotation files
+     * @returns ValueListReferences found in the files.
+     */
+    public getValueListReferences(
```
IainSAP: Can we make this static as it does not require any state from an active connection? Also consider exporting a type for the returned object?

packages/axios-extension/src/abap/abap-service-provider.ts
```diff
@@ -288,6 +290,38 @@ export class AbapServiceProvider extends ServiceProvider {
...
 
+    /**
+     * Collects ValueListReferences annotation values from the service metadata and annotation files.
+     *
+     * @param references - Service references for which metadata should be fetched.
+     * @returns A list of ValueListReferences found in the metadata and annotations.
+     */
+    public async fetchValueListReferenceServices(
+        references: { target: string; rootPath: string; value: string }[]
+    ): Promise<{ target: string; data?: string; path: string }[]> {
```
IainSAP: Can we please use the same type as input here as is returned by `getValueListReferences()`. I think the type can be defined in axios extension and add a dev dep. to odata-service-writer

voicis: Good catch, I removed all the consumers so typescript didn't pick up this mismatch, but this should not be an issue when we have consuming code.

I didn't want to add the dependency to the `odata-service-writer`, because usually it is a slippery slope and it would make it easier to add other unintended usages from the `axios-extension`.

If we use duck types, the main downside is that it is more difficult to see the link between `getValueListReferences` output type and `fetchValueListReferenceServices` input type. However, changing either of them still would make typescript compiler to complain at the consumption point.

I think this is the best of the not so great options, but if you insist I'll move the type to the `axios-extension`.

IainSAP: Yes so this highlights the original issue with the location of `getValueListReferences` in `odata-service-writer`.Isnt there a shared module where this can be added that is already a (type) dependency of both. I think its better to have a type dep on `axios-extension` from `odata-service-writer` than the other way around.  Im not sure TBH. Not having dedicated types modules leads to these type of problems...(and a proliferation of replicated types)

voicis: Ok, lets have types in the `axios-extension` and add eslint rule to only allow type imports from the `axios-extension` in the `odata-service-writer`.

packages/odata-service-writer/src/index.ts
```diff
@@ -171,5 +165,13 @@ async function remove(basePath: string, service: OdataService, fs?: Editor): Pro
...
+export {
+    OdataVersion,
+    OdataService,
+    ServiceType,
+    EdmxAnnotationsInfo,
+    CdsAnnotationsInfo,
+    ExternalServiceCollectionOptions,
+    NamespaceAlias
+} from './types';
+export { getExternalServiceReferences, getAnnotationNamespaces } from './data';
```
IainSAP: Please export `writeExternalServiceMetadata` as this is needed as a standalone function also.

packages/odata-service-writer/src/data/external-services.ts
```diff
@@ -0,0 +1,228 @@
...
+ * @param webappPath - Webapp path of the UI5 application
+ * @param externalServices - External service metadata to be written
+ * @param serviceName - Name of the service, defaults to 'mainService'
+ * @param servicePath - Service path of the service
+ */
+export function writeExternalServiceMetadata(
+    fs: Editor,
+    webappPath: string,
+    externalServices: ExternalService[],
+    serviceName = 'mainService',
```
IainSAP: Theres a constant defined for this value: https://github.com/SAP/open-ux-tools/blob/7e07138773e23129f0df1d6af845762988e73efa/packages/odata-service-writer/src/data/constants.ts#L1

packages/axios-extension/src/abap/abap-service-provider.ts
```diff
@@ -288,6 +297,46 @@ export class AbapServiceProvider extends ServiceProvider {
...
+     * Collects ValueListReferences annotation values from the service metadata and annotation files.
+     *
+     * @param references - Service references for which metadata should be fetched.
+     * @returns A list of ValueListReferences found in the metadata and annotations.
+     */
+    public async fetchExternalServices(references: ExternalServiceReference[]): Promise<ExternalService[]> {
+        const valueListReferences: ExternalService[] = [];
+        const allPromises = references.map(async (reference) => {
+            const { serviceRootPath, value } = reference;
+            const externalServicePath = joinPosix(serviceRootPath, value).replace('/$metadata', '');
```
devinea, line 310:
```suggestion
            const externalServicePath = joinPosix(serviceRootPath, value).replace('$metadata', '');
```
If I test this manually the ABAP server redirects (307) to add back the trailing slash.

packages/odata-service-writer/.eslintrc.js
```diff
@@ -3,5 +3,19 @@ module.exports = {
...
+    },
+    rules: {
+        '@typescript-eslint/no-restricted-imports': [
+            'error',
+            {
+                paths: [
+                    {
+                        name: '@sap-ux/axios-extension',
+                        allowTypeImports: true,
+                        message: "Only type imports from '@sap-ux/axios-extension' are allowed here."
```
devinea, line 15: 👍

devinea (approved):
Thanks @voicis 
New support for saving value list references look great. 
Covered by tests 
Changeset ok. might be a minor change for axios-extension and odata-service-writer?
1 small comment.

packages/axios-extension/src/abap/abap-service-provider.ts
voicis: I don't think this is ever called/used directly. Service provider appends the `/$metadata` later anyways when retrieving metadata.

PR #3795, opened 2025-10-31
fix(mcp-server): allow to create object pages in v2

815are:
Problem - creation of pages fails in V2 apps.
Reason -  is that fpm-writer supports only creation of V4 pages.

Solution:
1. Use the `exportConfig` API method to create pages.
2. The limitation is that only Object Pages can be created, which means we need to update the zod and JSON schemas for the add-page functionality based on the fiori elements version


packages/fiori-mcp-server/src/tools/functionalities/page/application.ts
```diff
@@ -570,7 +633,9 @@ export class Application {
         // Generate the page
         const navigations = await this.getAllowedNavigations(parentPage);
         const targetNavigation = this.findTargetNavigation(navigations, navigation, entitySet);
-        const { pageID, changes } = await this.generatePageWithFPMWriter(
+        const generateMethod = this.version === FioriElementsVersion.v2 ? this.generatePageV2 : this.generatePageV4;
```
heimwege, line 636: This whole `version` property confuses me a bit. Is version the Fiori Elements version (does something like a Fiori Elements version even exist or isn't the FE version exactly the same as the UI5 version as it is shipped together?) or is it actually the OData version?

815are: this one is resolved/derived by specification.
If we check types `node_modules\@sap\ux-specification\dist\types\src\common\page.d.ts`:
```
export interface Target {
    fioriElements?: FioriElementsVersion;
    odata?: OdataVersion;
}
export declare enum FioriElementsVersion {
    v2 = "v2",
    v4 = "v4"
}
export declare enum OdataVersion {
    v2 = "v2",
    v4 = "v4"
}
```

there is logic in specification which tries to detect if it is v2 or v4 fiori elements app + also adds odata version. And result of detection is passed within `app.json`

PR #3798, opened 2025-11-03
feat(preview-middleware): add log message for viewCache redirect

heimwege:
Cross link to https://github.com/ui5-community/wdi5/issues/694

The redirect to add `viewCache=true` param causes problems when used together with WDI5

At least we should log the redirect.


mikicvi-SAP (approved):
Changeset message appropriate ✅
Changes LG
- Did not test locally.

IainSAP (approved):
Thanks @heimwege 

- Log update only

PR #3805, opened 2025-11-04
Fiori tools analyser - [WIP]

unseen1980:
Bill of materials of Fiori app

### DIFF
--- a/packages/project-analyser/test/io/annotations.test.ts
+++ b/packages/project-analyser/test/io/annotations.test.ts
@@ -1,0 +1,109 @@
+import path from 'node:path';
+import { tmpdir } from 'node:os';
+import { promises as fs } from 'node:fs';
+import type { Logger } from '@sap-ux/logger';
+
+jest.mock('fast-glob', () => jest.fn());
+import fg from 'fast-glob';
+
+const realFastGlob: any = jest.requireActual('fast-glob');
+
+import { loadAnnotationDocuments } from '../../src/io/annotations';
+
+describe('loadAnnotationDocuments', () => {
+    let tempDir: string;
+
+    beforeEach(async () => {
+        (fg as unknown as jest.Mock).mockImplementation((pattern: any, options: any) => realFastGlob(pattern, options));
+        tempDir = await fs.mkdtemp(path.join(tmpdir(), 'project-analyser-'));
+    });
+
+    afterEach(async () => {
+        if (tempDir) {
+            await fs.rm(tempDir, { recursive: true, force: true });
+        }
+    });
+
+    it('returns annotation documents with inferred formats', async () => {
+        const annotationsDir = path.join(tempDir, 'annotations');
+        await fs.mkdir(annotationsDir, { recursive: true });
+
+        const fixtures = [
+            { filename: 'travel.xml', content: '<Annotations />', format: 'xml' },
+            { filename: 'travel.json', content: '{ "annotations": [] }', format: 'json' },
+            { filename: 'travel.cds', content: 'annotate Travel with {};', format: 'cds' }
+        ] as const;
+
+        for (const { filename, content } of fixtures) {
+            await fs.writeFile(path.join(annotationsDir, filename), content, 'utf8');
+        }
+
+        const documents = await loadAnnotationDocuments({ appPath: tempDir });
+
+        expect(documents).toHaveLength(fixtures.length);
+        for (const fixture of fixtures) {
+            const matchingDoc = documents.find((doc) => doc.path === path.join(annotationsDir, fixture.filename));
+            expect(matchingDoc).toEqual(
+                expect.objectContaining({
+                    content: fixture.content,
+                    format: fixture.format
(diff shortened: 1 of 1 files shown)


packages/fiori-tools-analyser/LICENSE
```diff
@@ -0,0 +1,2 @@
+This package is licensed under the Apache License, Version 2.0. See the root project `LICENSE` file for the full text.
```
devinea: Please include full text here like https://github.com/SAP/open-ux-tools/blob/main/packages/backend-proxy-middleware/LICENSE

packages/fiori-tools-analyser/package.json
```diff
@@ -0,0 +1,46 @@
...
+        "dist",
+        "LICENSE",
+        "!dist/*.map",
+        "!dist/**/*.map"
+    ],
+    "dependencies": {
+        "@sap-ux/logger": "workspace:*",
+        "@sap-ux/project-access": "workspace:*",
+        "fast-glob": "3.3.1",
+        "yaml": "2.3.3"
```
devinea: Not used so far. Should use `@sap-ux/yaml` instead where possible.

packages/fiori-tools-analyser/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "files": [
+        "dist",
+        "LICENSE",
+        "!dist/*.map",
+        "!dist/**/*.map"
+    ],
+    "dependencies": {
+        "@sap-ux/logger": "workspace:*",
+        "@sap-ux/project-access": "workspace:*",
+        "fast-glob": "3.3.1",
```
devinea:
```suggestion
        "fast-glob": "3.3.3",
```

packages/fiori-tools-analyser/LICENSE
unseen1980: Done

packages/project-analyser/package.json
```diff
@@ -0,0 +1,44 @@
...
+    },
+    "files": [
+        "dist",
+        "LICENSE",
+        "!dist/*.map",
+        "!dist/**/*.map"
+    ],
+    "dependencies": {
+        "@sap-ux/logger": "workspace:*",
+        "@sap-ux/project-access": "workspace:*",
```
IainSAP: We should expose the results of analysis via the project access module as the central access point for applications. This would lead to a circular dependency.

packages/project-analyser/src/types/analyzers.ts
```diff
@@ -0,0 +1,12 @@
+import type { ListReportAnalysis } from './list-report';
+import type { ObjectPageAnalysis } from './object-page';
+
+export interface ManifestAnalysis {
+    readonly listReport?: ListReportAnalysis;
```
IainSAP, line 5: Could we have a type discriminated structure rather than separate properties for each type. Is it necessary to even discriminate at all here? Why not a single analysis for manifest

packages/project-analyser/src/types/analyzers.ts
```diff
@@ -0,0 +1,12 @@
...
+    readonly objectPage?: ObjectPageAnalysis;
+}
+
+export interface AnnotationAnalysis {
+    readonly listReport?: ListReportAnalysis;
```
IainSAP, line 10: Same comment as above

packages/project-analyser/src/types/analyzers.ts
IainSAP, line 5:
```suggestion
    type: 'lrop' | 'ovp' ....
    commonAnalysisStuff:
    manfiest: Manifest // See project access (alias) type Manifest = SAPJSONSchemaForWebApplicationManifestFile
```

packages/project-analyser/src/io/manifest.ts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @param options - analyser options containing the application path
+ * @param logger - optional logger for diagnostic output
+ * @returns the manifest document or undefined when not found/parsable
+ */
+export async function loadManifestDocument(
+    options: AnalyseAppOptions,
+    logger?: Logger
+): Promise<ManifestDocument | undefined> {
+    const manifestPath = join(options.appPath, 'webapp', 'manifest.json');
```
sap-sebelao, line 20: `@sap-ux/project-access` provides a `getWebappPath` function, which reads the actual folder path from ui5.yaml. While majority of the apps do use /webapp, some do not, so reusing the function might be a good idea.

Word of caution though: the JSDoc of `getWebappPath` does not explicitly mention that the function can throw if the YAML is invalid or missing.

unseen1980: Thanks, I addressed that

packages/project-analyser/package.json
unseen1980: Done

unseen1980:
Closing this one as we are going to a different direction.

PR #3807, opened 2025-11-04
chore: update pipeline config to publish extensions as draft

cianmSAP:
Extensions published will be draft until they have been tested and can be flagged for publish to the market place


devinea (approved):
Thanks @cianmSAP

PR #3819, opened 2025-11-07
Resolve Transport Request Checks for Download Quick Fiori App Deploy

kjose90:
Internal bug - #36112
Deploy fails for downloaded Quick Fiori apps when the package is not local, due to incorrect transport request handling.

Implemented transport request resolution logic:
- Use available transport if found
- Set placeholder for missing transport
- Set '' TR for local packages

### DIFF
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.109
+
+### Patch Changes
+
+-   Updated dependencies [aa2c7df]
+    -   @sap-ux/fiori-freestyle-writer@2.5.0
+    -   @sap-ux/fiori-elements-writer@2.8.0
+
 ## 1.1.108
 
 ### Patch Changes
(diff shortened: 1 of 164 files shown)


packages/repo-app-import-sub-generator/src/app/app-config.ts
```diff
@@ -14,13 +15,62 @@ import { join } from 'node:path';
...
+ * @throws Error when the transport check fails
+ */
+async function resolveTransportRequest(context: AppDownloadContext): Promise<string> {
+    const { serviceProvider, qfaJson } = context;
+
+    const transportService = await serviceProvider?.getAdtService<TransportChecksService>(TransportChecksService);
+    const packageName = qfaJson.metadata.package;
+    const appName = qfaJson.deploymentDetails.repositoryName;
+
+    if (packageName === '$TMP') {
```
longieirl, line 42: Would you not need to return this first? Then try to get the transportService?

kjose90: yes - thank you

devinea (approved):
Code changes LG and covered by tests
Changeset 👍

Issue #3718, opened 2025-10-10
FEATURE - Prevent overwriting of saved system credentials by default via generator


mikicvi-SAP:
_As a **developer**,
I want to **be able to use the generator with a saved system that does not contain credentials**,
so that I **can still have the system without the credentials after generation.**_

### Description

For the 3 scenarios below:
 1. User sets up a new ABAP on premise system using SAP saved systems without entering the user credentials 
Generate the application using the same system, message is shown to add user credentials

2. If the user credentials are saved for the system, but there is an auth error while generating the app, user enters the correct credentials and the app is generated

3. User enters a new system via creating an app using the generator

- A new checkbox option will be displayed to the user asking if they want to save the user credentials, this will be defaulted to 'No'

### Technical Design
- Functionality is related to:
  - `@sap-ux/odata-service-inquirer`
  - `@sap-ux/fiori-app-sub-generator`

### Acceptance Criteria

- Given I generate a new application using my ABAP on prem system for which the username/password are not saved,  I see the prompts to add my user credentials below the System Selection prompt. After authenticating the system, I see the checkbox asking if I want to save the user credentials Yes/No (defaulting to No)

- Given I get login error for a saved system, I enter the new credentials and generate the application, After authenticating the system, I see the checkbox asking if I want to save the user credentials Yes/No (defaulting to No)

- Given I add a new system while generating an app, After authenticating the system, I see the checkbox asking if I want to save the user credentials Yes/No (defaulting to No)

**Checkbox message :** "Do you want to save the credentials for this system? Note: Passwords are stored in the OS credential manager and protected by its securtiy policies."

### Notes
- Related to internal issue: 35929

### Tasks
- [ ] Unit Tests added
- [ ] Test Cases defined

mikicvi-SAP:
Implemented with #3720

PR #3820, opened 2025-11-07
fix: move message to store creds prompt, tests, changeset

mikicvi-SAP:
# Part of #3718

- Moves the password message to choice prompt
  - Show only if answer is true

### DIFF
--- a/packages/abap-deploy-config-inquirer/CHANGELOG.md
+++ b/packages/abap-deploy-config-inquirer/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/abap-deploy-config-inquirer
 
+## 1.6.46
+
+### Patch Changes
+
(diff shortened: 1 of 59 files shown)


mikicvi-SAP:
> changes lgtm. Do the tests need more specific check for this change? changeset ✅

Thanks @devinea.
I'm not sure... Technically, we are only moving a message down 1 prompt (as well as showing conditionally) and test changes reflect the change and are verbose enough. What do you think @IainSAP?

IainSAP:
> > changes lgtm. Do the tests need more specific check for this change? changeset ✅
> 
> Thanks @devinea. I'm not sure... Technically, we are only moving a message down 1 prompt (as well as showing conditionally) and test changes reflect the change and are verbose enough. What do you think @IainSAP?

Yeah existing test that was moved is fine.

PR #3821, opened 2025-11-07
Feat/add prompt foo test for java

marufrasully:


marufrasully:
will be covered in a clean fresh PR.

PR #3824, opened 2025-11-10
Updating remaining ui5 cli references to v4 and align version

devinea:
https://github.com/SAP/open-ux-tools/issues/3220


donal-tobin-sap (dismissed):
Update of ui5 cli looks good

PR #3827, opened 2025-11-11
TBI - Update project templates to use eslint 9 and flat config. 

devinea:
Upgrade generated projects: 
1. use devDependencies
```JSON
"typescript": "^5.9.3", 
"eslint": "^9",
"@sap-ux/eslint-plugin-fiori-tools": "^9.0.0"
 ```
 2. change eslint flat config and use `eslint.config.js` instead of `.eslintrc` and .`eslintignore` files. (JSON no longer supported in eslint 9).
 3.  Update project templates to use new prefix replacing `// eslint-disable fiori-custom/` with `// eslint-disable @sap-ux/fiori-tools/`
 4. Removed codeAssist templates (TypeScript is a much better option for this use case).

### DIFF
--- a/packages/ui5-application-writer/src/types.ts
+++ b/packages/ui5-application-writer/src/types.ts
@@ -137,7 +137,12 @@ export interface UI5 {
 
 // Additional configurable features
 export interface AppOptions {
-    codeAssist: boolean; // Enables code assist
+    /**
+     * @deprecated
+     * Enables code assist
+     * No longer supported. Will be removed from the interface in future releases.
+     */
+    codeAssist: boolean;
     eslint: boolean; // Enables Fiori linting
     sapux: boolean; // Enables Fiori Tools Suite integration
 

### CODE REVIEW

File: packages/ui5-application-writer/src/options.ts [hunk 1]
```diff
@@ -41,7 +41,7 @@ async function copyTemplates(name: string, { ui5App, fs, basePath, tmplPath }: F
         const relPath = optTmplFilePath.replace(optTmplDirPath, '');
         const outPath = join(basePath, relPath);
         // Extend or add
-        if (!fs.exists(outPath)) {
+        if (!fs.exists(outPath) || !outPath.endsWith('.json')) {
```
IainSAP: Why is this changed? This condition is already covered by the existing condition. I dont' think there should be any need for specific template file type handling.  The entire folder should be copied or the files merged. Previously all files were json that needed to be merged but now we would need to merge .js also otherwise they will be overwritten. Can we provide the config as json instead...this would be far simpler to merge where different options required different eslint configs.

IainSAP (commented):
Unexpected change...see comment

devinea:
> Feedback on template copy.

Updated based on our discussion:
- When TypeScript option is enabled, eslint is also enabled. 
- Eslint option templates are only source of eslint config and dependencies (no longer duplicated in TypeScript option)
- codeAssist is no longer available in the generator to the end user. Removed the templates and logic in the writer.  (separate PR will remove codeAssist from types and prompts etc...)

packages/ui5-application-writer/src/types.ts
```diff
@@ -137,7 +137,12 @@ export interface UI5 {
 
 // Additional configurable features
 export interface AppOptions {
-    codeAssist: boolean; // Enables code assist
+    /**
+     * @deprecated
+     * Enables code assist
+     * No longer supported. Will be removed from the interface in future releases.
+     */
+    codeAssist: boolean;
```
IainSAP, line 145: 👍

marufrasully (approved):
re-approving.

PR #3832, opened 2025-11-12
fix: strategy, runs-on

gernotkuhr:
fix github action to correctly run playwright tests on win and ubuntu


donal-tobin-sap (dismissed):
Build looks more correct for matrix build of 2 OSes

PR #3834, opened 2025-11-13
fix versions.json fetching issue in windows

Jimmy-Joseph19:


.github/workflows/integration-tests.yml
```diff
@@ -59,15 +63,12 @@ jobs:
               run: pnpm install
             - name: Run build
               run: pnpm run build
-              env:
```
gernotkuhr, line 62: is it correct to delete this? better keep?

PR #3763, opened 2025-10-21
feat(fpm-writer): Support creating custom fields extensions

815are:
Support creating custom fields extensions
https://ui5.sap.com/#/topic/4e49753530ad4a73a44a5f8efac238d6

PR #3836, opened 2025-11-13
fix(fpm-writer): `generateCustomField` now includes the `core:require` attribute with the handler file in new fragment files when `eventHandler` is set.

815are:
recent implementation for custom field support was done in PR -> https://github.com/SAP/open-ux-tools/pull/3763

but that logic has a bug when we set `eventHandler: true`, then fragment looks following:
```
<core:FragmentDefinition xmlns:core="sap.ui.core" xmlns="sap.m">
	<Button text="Dummy2" press="handler.onPress" />
</core:FragmentDefinition>
```

there is missing `core:require`, like:
```
<core:FragmentDefinition xmlns:core="sap.ui.core" xmlns="sap.m">
	<Button core:require="{ handler: 'javascript/ext/fragment/Dummy2'}" text="Dummy2" press="handler.onPress" />
</core:FragmentDefinition>
```

### DIFF
--- a/.changeset/quiet-bats-bathe.md
+++ b/.changeset/quiet-bats-bathe.md
@@ -1,0 +1,6 @@
+---
+'@sap-ux/fe-fpm-writer': patch
+---
+
+Fix: `generateCustomField` now includes the `core:require` attribute with the handler file in new fragment files when `eventHandler` is set.
+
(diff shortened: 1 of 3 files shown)


devinea (approved):
Changes are straightforward and covered by tests. 
changeset ✅

AlinaGovoruhina (approved):
Changes are clear, tests updated, changeset exists.

PR #3839, opened 2025-11-14
fix(odata-service-inq): add try catch to fetching backend systems from store

cianmSAP:
Internal issue # 36303

As the `@sap-ux/store` module is designed to throw on file read errors e.g. https://github.com/SAP/open-ux-tools/blob/d99dea91c37e47073d811bd04b7c35df424532db/packages/store/src/data-access/filesystem.ts#L86

We can wrap the `getAll` call in a try / catch and return an empty list of systems when the call fails - to prevent generator crashing


packages/odata-service-inquirer/src/utils/store.ts
```diff
@@ -11,3 +13,22 @@ export async function getBackendSystemService(): Promise<Service<BackendSystem,
...
     return backendService;
 }
+
+/**
+ * Fetch all backend systems.
+ *
+ * @param includeSensitiveData - whether to include sensitive data
+ * @returns backened systems
+ */
+export async function getAllBackendSystems(includeSensitiveData = false): Promise<BackendSystem[] | []> {
```
IainSAP, line 23: Doesn't this mean, if theres an issue with one system then no systems will be returned from the store? I think this should be fixed in the store module perhaps so only corrupt entries are omitted. And also add the proper ts docs to indicate that these APIs throw. Since this is urgent these could be linked to a follow-up item.

IainSAP (dismissed):
Dont we need to update here also : https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts#L240-L241

IainSAP (commented):
Thanks @cianmSAP

Prevents crash but should be a follow-up to address the store module to be more resilient to bad data rather than not returning anything
Tests added to cover

packages/odata-service-inquirer/src/utils/store.ts
cianmSAP: If the JSON is malformed and can't be parsed, or in the case of why this issue was raised, I suspect the user does not have permissions to access the `systems.json` file, then I don't think we can have any improvement in terms of granularity.

longieirl (approved):
Changeset is good.
Tests are good,.
Overall change is good.

PR #3842, opened 2025-11-17
fix(preview-middleware): support rta and cpe for CAP node w/o mockserver

heimwege:
This PR enables apps that are served by CAP node to also create, change and load local `.changes`

Currently API endpoints `/preview/api/*` as well as the `WorkspaceConnector` ignore the namespace that is added by the `cds-plugin-ui5` resulting in http `404` responses when trying to save `.changes` in the local preview.

### DIFF
--- a/.changeset/real-turkeys-drum.md
+++ b/.changeset/real-turkeys-drum.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/fiori-elements-writer': patch
----
-
-fix order of test file creation
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.160
+
+### Patch Changes
+
+- Updated dependencies [4368c00]
+    - @sap-ux/fiori-elements-writer@2.8.47
+
 ## 1.1.159
 
 ### Patch Changes
--- a/examples/simple-generator/package.json
+++ b/examples/simple-generator/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/generator-simple-fe",
-    "version": "1.1.159",
+    "version": "1.1.160",
     "description": "Simple example of a yeoman generator for Fiori elements.",
     "license": "Apache-2.0",
     "private": true,
--- a/packages/fiori-app-sub-generator/CHANGELOG.md
+++ b/packages/fiori-app-sub-generator/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/fiori-app-sub-generator
 
+## 0.11.38
+
+### Patch Changes
+
+- Updated dependencies [4368c00]
+    - @sap-ux/fiori-elements-writer@2.8.47
+    - @sap-ux/odata-service-inquirer@2.14.18
+
 ## 0.11.37
 
 ### Patch Changes
--- a/packages/fiori-app-sub-generator/package.json
+++ b/packages/fiori-app-sub-generator/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/fiori-app-sub-generator",
     "description": "A yeoman (sub) generator that can generate Fiori applications. Not for standalone use.",
-    "version": "0.11.37",
+    "version": "0.11.38",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/fiori-elements-writer/CHANGELOG.md
+++ b/packages/fiori-elements-writer/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/fiori-elements-writer
 
+## 2.8.47
+
+### Patch Changes
+
+- 4368c00: fix order of test file creation
+
 ## 2.8.46
 
 ### Patch Changes
--- a/packages/fiori-elements-writer/package.json
+++ b/packages/fiori-elements-writer/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/fiori-elements-writer",
     "description": "SAP Fiori elements application writer",
-    "version": "2.8.46",
+    "version": "2.8.47",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/repo-app-import-sub-generator/CHANGELOG.md
+++ b/packages/repo-app-import-sub-generator/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/repo-app-download-sub-generator
 
+## 0.3.219
+
+### Patch Changes
+
+- Updated dependencies [4368c00]
+    - @sap-ux/fiori-elements-writer@2.8.47
+    - @sap-ux/odata-service-inquirer@2.14.18
+
 ## 0.3.218
 
 ### Patch Changes
--- a/packages/repo-app-import-sub-generator/package.json
+++ b/packages/repo-app-import-sub-generator/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/repo-app-import-sub-generator",
     "description": "Generator to download LROP Fiori applications deployed from an ABAP repository.",
-    "version": "0.3.218",
+    "version": "0.3.219",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",


packages/preview-middleware/src/base/flp.ts
```diff
@@ -308,7 +308,7 @@ export class FlpSandbox {
             : '@sap-ux/preview-middleware';
 
         await this.setApplicationDependencies();
-        const config:TemplateConfig = { ...this.templateConfig };
+        const config = structuredClone(this.templateConfig);
```
heimwege, line 324: @nikmace this change (that should actually be a tiny refactoring and improve the code quality) leads to snapshot updates in test `FlpSandbox › router › test/flp.html UI5 1.71 with asyncHints.requests`:

from
```
applications:  {"app-preview":{"title":"my.id","description":"","additionalInformation":"SAPUI5.Component=myComponent","applicationType":"URL","url":"..","applicationDependencies":{"manifest":{},"name":"descriptorName","url":"http://sap.example","asyncHints":{"requests":[]}}},"testfev2other-preview":{"title":"My Other App","description":"This is a very simple application.","additionalInformation":"SAPUI5.Component=test.fe.v2.other","applicationType":"URL","url":"/yet/another/app"}}  
```
to
```
applications:  {"app-preview":{"title":"my.id","description":"","additionalInformation":"SAPUI5.Component=myComponent","applicationType":"URL","url":"..","applicationDependencies":{"manifest":{},"name":"descriptorName","url":"http://sap.example","asyncHints":{"requests":[{"name":"myRequest","url":"http://sap.example"}]}}},"testfev2other-preview":{"title":"My Other App","description":"This is a very simple application.","additionalInformation":"SAPUI5.Component=test.fe.v2.other","applicationType":"URL","url":"/yet/another/app"}}   
```

Change is that 
```
{"name":"myRequest","url":"http://sap.example"}
```
is being added to the `asyncHints` requests (see [commit](https://github.com/SAP/open-ux-tools/pull/3842/commits/b80239f26f807638110789ba98cd83605e905e35) that adjusts the test snap)

My assumption is that this is done because `{ ...this.templateConfig }` in contrast to `structuredClone(this.templateConfig)` does not deep clone but only create a shallow copy.
This can lead to unintended side effects because nested objects, like `ui5` or `apps`, are not cloned. Subsequent modifications to `config.ui5`, `config.apps`, etc. will also alter `this.templateConfig.ui5` / `this.templateConfig.apps`.

Can you please confirm that the snapshot adjustment is fine and the old snapshot was actually wrong because of the shallow copy? Or do we somehow rely on the shallow copy adjusting the `this.templateConfig` sub-objects?

heimwege:
@devinea could you please verify my [CodeQL dismissal](https://github.com/SAP/open-ux-tools/pull/3842/checks?check_run_id=55593191088) for `Missing rate limiting` and `Server-side request forgery`. I dismissed both as `wontfix` because everything runs on `localhost` only.

devinea:
> @devinea could you please verify my [CodeQL dismissal](https://github.com/SAP/open-ux-tools/pull/3842/checks?check_run_id=55593191088) for `Missing rate limiting` and `Server-side request forgery`. I dismissed both as `wontfix` because everything runs on `localhost` only.

Agreed. 👍

packages/preview-middleware/src/base/flp.ts
nikmace: Honestly, I don't understand the problem you are trying to solve, and I have no idea how the CAP apps work in the Adaptation Editor. Adding requests to 1.71 might break ADP for 1.71.
Screenshot 2025-11-24 at 9 05 15
This is why we remove it here ^

Therefore, I cannot say how this will affect ADP for UI5 1.71. A little unfortunate that our QA is off this week, and there is an infrastructure problem for the UI integration tests in the tools-suite repo, so we cannot run regression testing.

`{ ...this.templateConfig }`, this looks unnecessary, and probably you are right about cloning.

heimwege: As said, there is no actual problem to solve, but using the spread operator to clone an object is actually quite error prone because it only creates a shallow copy whereas `structuredClone` is a build in clone functionality (that also preserves the type).

Using the spread operator leads to the top-level objects to be cloned but nested properties are still shared with the original object.

Sample:
```js
const originalProject = {
  id: 1,
  name: "My Project",
  settings: {
    theme: "light",
    fontSize: 12
  }
};

// CASE 1: create a shallow copy
const clonedProject = { ...originalProject };

clonedProject.settings.theme = "dark";

console.log("Original Theme:", originalProject.settings.theme); // dark ⚠️ <-- this might not be intended
console.log("Cloned Theme:", clonedProject.settings.theme);     // dark

// CASE 2: create a clone
const realClonedProject = structuredClone(originalProject);

realClonedProject.settings.fontSize= "42";

console.log("Original Theme:", originalProject.settings.fontSize); // 12
console.log("Cloned Theme:", realClonedProject.settings.fontSize); // 42
```

And currently there seems to be such a side effect because changing from spread operator to `structuredClone` actually results in a changed snapshot.

My question is which of the two options is currently implemented:
1) The snapshot in `main` is wrong and needs to be adjusted after creating a real clone
2) We rely on the indirect changes done to the `this.templateConfig` via the shallow copied object `config`.

In case of 2) this is not documented and is highly confusing and error prone. In that case we should do the changes on the `this.templateConfig` before creating a clone.

heimwege: @nikmace I found the issue. There were several places where the shallow copy modified (and used) the original object `this.templateConfig` (e.g. `checkDeleteConnectors` and `removeAsyncHintsRequests`). So it's very important to get the timing right when to do the structured clone.
I refactored the code accordingly and the snapshot is back to what it was in `main`.

mmilko01 (commented):
Just one small question. Also, is it possible to execute the integration tests against this change and attach a screenshot with the results as a comment in the PR?

packages/preview-middleware-client/src/flp/WorkspaceConnector.ts
```diff
@@ -2,9 +2,13 @@ import merge from 'sap/base/util/merge';
...
 import Layer from 'sap/ui/fl/Layer';
 import type { FlexChange } from './common';
-import { CHANGES_API_PATH, getFlexSettings } from './common';
+import { CHANGES_API_PATH as CHANGES_API_PATH_STATIC, getFlexSettings } from './common';
 import { getUi5Version, isLowerThanMinimalUi5Version } from '../utils/version';
 import { getAdditionalChangeInfo } from '../utils/additional-change-info';
+import { Window } from '../../types/global';
+
+const baseUrl = (globalThis as unknown as Window)['data-open-ux-preview-base-url'] ?? '';
+const CHANGES_API_PATH = `${baseUrl}${CHANGES_API_PATH_STATIC}`;
```
mmilko01: What is the reason for this constant to not be in camel case similar to `baseUrl`?

heimwege: UIi5 refactored their param names from camelCase to kebap-case. I wanted to stick to that pattern as well and also already refactored e.g. `data-open-ux-preview-customInit` to `data-open-ux-preview-custom-init`.
CSS also uses kebap-case and this is quite common for HTML (therefore the `.dataset` exists for auto-conversion between HTML and JS/TS notation).

`myElement.dataset.openUxPreviewBaseUrl` will read `data-open-ux-preview-base-url`

Sadly .dataset only works for HTML elements and not for global variables.
Maybe I should refactor to add this to the ui5 bootstrap as well to be able to use the .dataset notation like we e.g. do with the flex settings 🤔 

```
const bootstrapConfig = document.getElementById('sap-ui-bootstrap');
const flexSetting = bootstrapConfig?.dataset.openUxPreviewFlexSettings;
```

What do you think?

mmilko01: Maybe I don't fully understand your comment, but my point was regarding the `CHANGES_API_PATH` which I think shouldn't be related to the UI5 param names. If I understand you correctly, you are referring to `data-open-ux-preview-base-url`. Probably I should have only highlighted line 11

heimwege: oh, sorry. Then I didn't get your question 😄 The import was already that Java notation `CHANGES_API_PATH` and I just sticked to that and just renamed to `CHANGES_API_PATH_STATIC`. But I could also adjust the import to something like 
```ts
import { CHANGES_API_PATH as changesApiPathStatic, getFlexSettings } from './common';
```

heimwege: I refactored to not attach info to the global window object but use `.dataset` instead. Do you want me to rename `CHANGES_API_PATH_STATIC`? (would also apply to [packages/preview-middleware-client/src/flp/enableFakeConnector.ts‎](https://github.com/SAP/open-ux-tools/pull/3842/changes#diff-3589a93a120cd76c2e37c92cc091e3ac525e6a1b45c85f98dfcedfc80aeccfe3))

mmilko01: Regarding the `CHANGES_API_PATH_STATIC` constant, I think the imported constant is ok in the current style, but maybe the newly created constant can be changed to cammelCase

```
const changesApiPath = `${baseUrl}${CHANGES_API_PATH_STATIC}`;
```

Same for the [packages/preview-middleware-client/src/flp/enableFakeConnector.ts](https://github.com/SAP/open-ux-tools/pull/3842/changes#diff-3589a93a120cd76c2e37c92cc091e3ac525e6a1b45c85f98dfcedfc80aeccfe3)

mmilko01:
Pasting the integration test execution against this change:

image

packages/preview-middleware-client/src/flp/WorkspaceConnector.ts
heimwege: renamed in 72dfd4f343c39494d422bd8d42116d91b10f0b76

mmilko01 (approved):
Changeset OK
Changes look good 
Excellent test coverage
Did not test manually, but ADP integration tests passed successfully

PR #3843, opened 2025-11-18
fix(sap-systems-ext):  title update and import fix for sap systems ext

cianmSAP:
- Small correction for the title when launching the generator with a specific system
- Also tighten up functionality with importing systems which already exist - should just view the existing system (with name updated) rather than imp


devinea (approved):
changes lg and covered by tests. 
changeset ✅

PR #3844, opened 2025-11-18
fix(security): upgrade js-yaml

donal-tobin-sap:
- backport 3 version fix

Issue: #36302
Issue: https://github.com/SAP/open-ux-tools/security/dependabot/252


donal-tobin-sap:
Update din alter PR

PR #3847, opened 2025-11-18
fix(backend-proxy-middleware): add missing 'express' dependency

matz3:
The dependency was only used as devDependency before, but since recent changes (https://github.com/SAP/open-ux-tools/pull/3032) the dependency is now required at runtime as well (express.Router).

An alternative to express.Router is the [router](https://www.npmjs.com/package/router) package, which provides the same functionality. We're using that package in our built-in ["csp" middleware](https://github.com/UI5/cli/blob/5f6f20ced474aef0086b21c213d56cec956c83ad/packages/server/lib/middleware/csp.js#L2) in UI5 CLI.

### DIFF
--- a/.changeset/witty-suns-post.md
+++ b/.changeset/witty-suns-post.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/create': minor
----
-
-feat(create): enable card generator for CAP project
--- a/packages/create/CHANGELOG.md
+++ b/packages/create/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/create
 
+## 0.14.0
+
+### Minor Changes
+
+-   e94df25: feat(create): enable card generator for CAP project
+
 ## 0.13.172
 
 ### Patch Changes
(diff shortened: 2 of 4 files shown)


heimwege:
@devinea third party contribution still seem to break the pipeline `Input required and not supplied: token` for the lint fix. Also reported SONAR checks look weird (https://sonarcloud.io/component_measures?id=SAP_open-ux-tools&pullRequest=3847&metric=new_coverage&view=list). `packages/fiori-app-sub-generator/src/fiori-app-generator/end.ts` has certainly not been adjusted in this PR 🤷🏻

heimwege:
@matz3 thanks for finding this one. I'll follow up on the `router` package.
Could you please add a patch version changeset (just `pnpm cset` in the monorepo root)? w/o changeset no version will be published to npmjs.

matz3:
@heimwege thank you. I've added a changeset to the existing commit.

heimwege:
Not to self: same issue seems to exist for ui5-proxy-middleware. Check #540

heimwege (approved):
fix lgtm

Issue #3850, opened 2025-11-19
BUG - System SAPUI5 version not available for selection in ADP generator


mmilko01:
### Description
When creating and Adaptation Project via the Adaptation Project generator, only the latest public SAPUI5 version is available for selection in `SAPUI5 Version` prompt.

### Steps to Reproduce
1. Start Adaptation Project generator
2. Select a system with SAPUI5 version different than the latest public SAPUI5 version
3. See that only the latest public SAPUI5 version is available for selection in `SAPUI5 Version` prompt

### Expected results
The versions that should be available for selection should be the `System version` and all higher public versions including the latest.

### Actual results
Only the latest public version is available for selection.

### Screenshots
If applicable, add screenshots to help explain the problem.

<img width="978" height="414" alt="Image" src="https://github.com/user-attachments/assets/29dea9be-185e-4caf-b54c-c24dc121e258" />

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #3851, opened 2025-11-19
fix: System SAPUI5 version not available for selection in ADP generator

mmilko01:
#3850

* Fixed logic for identifying relevant SAPUI5 versions, to show the system version and all higher public versions including the latest SAPUI5 version
* Validation if a given SAPUI5 version is available as a public version now uses `axios` to handle properly exceptions.


IvoSG (approved):
Changes look good
Changeset is ok
Excellent test coverage
Did not test manually

PR #3855, opened 2025-11-21
fix(preview-middleware): loading i18n bundle for CAP projects

### LINKED ISSUE #3721: BUG - Unable to load i18n bundle when started with CDS
### Description
UI frontend is unable to load the i18n bundles when loaded via a CAP project.

[i18n property function](https://github.com/SAP/open-ux-tools/blob/master/packages/preview-middleware/src/base/config.ts#L328) is loaded with a path of `.` instead of `./app/<project-name>`.

```bash
}
info fiori-tools-proxy Using UI5 version 1.140.0 based on manifest.json
info fiori-tools-proxy Starting fiori-tools-proxy using following configuration:
info fiori-tools-proxy proxy: 'undefined'
info fiori-tools-proxy ignoreCertErrors: 'false'
info fiori-tools-proxy backend: undefined
info fiori-tools-proxy ui5: [{"path":"/resources","url":"https://sapui5.hana.ondemand.com","version":"1.140.0"},{"path":"/test-resources","url":"https://sapui5.hana.ondemand.com","version":"1.140.0"}]
info fiori-tools-proxy debug: 'false'
info fiori-tools-appreload Livereload middleware started for port 35729 and path /Users/Downloads/cap-app/app/translation-ui/webapp
warn fiori-tools-preview Failed to load i18n properties bundle
warn fiori-tools-preview Failed to load i18n properties bundle
```

This is not an issue when deployed, only impacts SAPUI5 apps when rendered locally.

### Steps to Reproduce
1. Generate a CAP project;
```bash
cds init managedAppCAPProject && cd  managedAppCAPProject && npm install && cds add tiny-sample && cds add hana && cds add mta && cds add xsuaa && cds add connectivity && cds add destination && cds add html5-repo && cds build --for hana && cp gen/db/package.json db && npm i
```
3. Append a SAPUI5 UI app using SAP Fiori tools generator
4. Start the server using `npm run start`

### Expected results

i18n bundle is loaded

### Actual results

i18n fails to load since the project path is `.` and the UI app is not found in that path.

### Screenshots

OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

heimwege:
fixes #3721

Adjust preview middleware to use absolute instead of relative local path. For additional flp.apps absolute and relative paths are supported

Also fixes an issue that instantiating project-access with `createProjectAccess('.')` returns ALL projects in `.` (and subdirectories). preview-middleware uses the first found app and might return wrong i18n 🙈

### DIFF
--- a/packages/fiori-mcp-server/CHANGELOG.md
+++ b/packages/fiori-mcp-server/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/fiori-mcp-server
 
+## 0.4.10
+
+### Patch Changes
+
+-   8e8e7ce: fix(fiori-mcp): add service metadata validation
+
 ## 0.4.9
 
 ### Patch Changes
--- a/packages/fiori-mcp-server/package.json
+++ b/packages/fiori-mcp-server/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/fiori-mcp-server",
     "description": "SAP Fiori - Model Context Protocol (MCP) server",
-    "version": "0.4.9",
+    "version": "0.4.10",
     "keywords": [
         "SAP Fiori tools",
         "SAP Fiori elements",
@@ -58,6 +58,7 @@
         "@sap-ux/project-access": "workspace:*",
         "@sap-ux/fiori-annotation-api": "workspace:*",
         "@sap-ux/annotation-converter": "0.10.8",
+        "@sap-ux/edmx-parser": "0.9.6",
         "@sap-ux/fiori-docs-embeddings": "workspace:*",
         "@sap-ux/i18n": "workspace:*",
         "@sap-ux/logger": "workspace:*",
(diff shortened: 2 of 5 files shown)


packages/preview-middleware/src/base/config.ts
```diff
@@ -336,8 +346,9 @@ async function getI18nTextFromProperty(
     }
     const projectAccess = await createProjectAccess(projectRoot);
     const applicationIds = projectAccess.getApplicationIds();
+    const applicationId = applicationIds.find((id) => id === appId) ?? applicationIds[0];
     try {
-        const bundle = (await projectAccess.getApplication(applicationIds[0]).getI18nBundles())['sap.app'];
+        const bundle = (await projectAccess.getApplication(applicationId).getI18nBundles())['sap.app'];
```
heimwege: Can we simply use appId here and skip calling getApplicationIds from project access?

Klaus-Keller: @heimwege 
if you have the path to the app, you can call `createApplicationAccess()` directly:
https://github.com/SAP/open-ux-tools/blob/a658553551950d005d4c94b215668ad38711f3de/packages/project-access/src/project/access.ts#L384-L387

heimwege: adjusted

Klaus-Keller (approved):
Re-approve

Klaus-Keller (approved):
Re-approve

Klaus-Keller (approved):
re-approve

PR #3856, opened 2025-11-21
Fix(cap-config-writer): min cds version

heimwege:


ytpo-lyne (approved):
- uses cds version instead of cds plugin ui5
- changeset present
- coverage met
- did not test manually

Approving.

PR #3862, opened 2025-11-25
Bug - Fix npm install for CAP when adding first app so that preview works after install.

devinea:
Bug - Fix npm install for CAP when adding first app so that preview works after install.
Use correct correct path when using npm workspaces (the CAP root).


IainSAP (approved):
Thanks @devinea 

- Fix is good for short term, agreed to follow up to remove legacy `cdsUi5PluginInfo`

PR #3866, opened 2025-11-25
feat(project-access): support reading path mappings for other types than application

heimwege:
Currently only `getWebappPath` exists to get the path mapping for `webapp`. This PR introduces `getPathMappings` that reads the path mappings depending on the `type` defined in the given yaml file name (default: `ui5.yaml`).

For backward compatibility reasons the behavior of `getWebappPath` has not been changed. It still returns `<some path>/webapp` in case it is called for other types or the yaml file has been found at all.


heimwege:
@Klaus-Keller I refactored according to your comment. Please re-review

815are (approved):
checked:
1. code change in project-access - I did not notice issue;
2. tests are maintained;
3. changelog persists

I did not test manually

Klaus-Keller (approved):
re-approve

815are (approved):

PR #3867, opened 2025-11-26
Feat/preview middleware/enable card generation cap projects

Singham-24:


Singham-24:
will create a new branch

PR #3737, opened 2025-10-15
feat(preview-middleware): enable card generator for CAP projects

Singham-24:
This PR introduces support for generating cards in CAP-based applications.
The enhancement ensures that the card generator integrates smoothly with CAP project structures and runtime behavior.

PR #3868, opened 2025-11-26
feat(create): enable card generator for CAP projects

Singham-24:
This PR introduces support for generating cards in CAP-based applications.
The enhancement ensures that the card generator integrates smoothly with CAP project structures and runtime behavior.

similar patch was created earlier :
https://github.com/SAP/open-ux-tools/pull/3737

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,6 +11,8 @@ on:
       - main
 jobs:
   run-linters:
+    # Skip this job on forks since it requires secrets and push permissions
+    if: github.event.pull_request.head.repo.full_name == github.repository
     strategy:
       matrix:
         os: [ubuntu-latest]
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -45,11 +45,14 @@ jobs:
             - name: Run build
               run: pnpm run build
               env:
-                  NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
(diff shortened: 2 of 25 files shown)


packages/create/test/unit/cli/add/cards-generator.test.ts
```diff
@@ -53,18 +47,6 @@ describe('add/cards-generator', () => {
         expect(traceSpy).not.toHaveBeenCalled();
     });
 
-    test('add cards-generator CAP', async () => {
```
heimwege, line 56: Why is this test being deleted? Shouldn't it be adjusted instead so that we have a test for CAP projects?

.changeset/spicy-dingos-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/create': major
+---
+
+feat(preview-middleware): enable card generator for CAP project
```
heimwege: This change is not related to the preview-middleware but to the create package

packages/create/test/unit/cli/add/cards-generator.test.ts
Singham-24: done

.changeset/spicy-dingos-mix.md
Singham-24: done

heimwege (approved):
- changes are straightforward
- review comments have been addressed
- changeset ok
- excellend code coverage
- did NOT test manually

Thanks @Singham-24

Sharmavksh (approved):
Looks good to me

PR #3872, opened 2025-11-26
fix: Bump @sap/manifest version

longieirl:


longieirl:
@devinea can you please review? Thanks

PR #3874, opened 2025-11-26
chore: remove outdated ui5 cli and specification artefacts from package.json files

heimwege:
Removing `ui5.dependencies` from `package.json` as they are not used in ui5 cli v3 or higher and lower versios are deprecated. And deleting `specification` dependencies because they are no longer needed

### DIFF
--- a/.changeset/eight-tigers-push.md
+++ b/.changeset/eight-tigers-push.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/fiori-docs-embeddings': patch
----
-
-feat: enhance downloaded readme content for better chunking
--- a/packages/backend-proxy-middleware/CHANGELOG.md
+++ b/packages/backend-proxy-middleware/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/backend-proxy-middleware
 
+## 0.10.16
+
+### Patch Changes
+
+-   67e2b43: add 'express' as peerDependency
+
 ## 0.10.15
 
 ### Patch Changes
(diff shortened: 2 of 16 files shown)


packages/ui5-library-reference-sub-generator/test/unit/sample/test_project_lrop_v2/package.json
```diff
@@ -26,10 +26,5 @@
     "@sap/ux-ui5-tooling": "1",
     "rimraf": "3.0.2",
     "@sap/ux-specification": "UI5-1.71"
-  },
```
devinea, line 29: This file references `@ui5/cli": "^2.14.1",` which needed section below. maybe better to update`@ui5/cli` and remove `@ui5/fs` and `@ui5/logger`

heimwege: That's a good finding 👍🏻
I adjusted all 2.x versions to 4 and also deleted all existing specification devDependencies because this is no longer needed (since quite a while)

heimwege: @devinea can we enforce to run all unit tests in the pipeline w/o `nx` cache? I'm afraid there might be some more snapshot files that would need to be adjusted that are currently not found because of the cache 🤷🏻

devinea: All working right it shouldn't be needed. But update here https://github.com/SAP/open-ux-tools/blob/chore/remove-outdated-ui5-cli-artefacts/nx.json#L45

PR #3877, opened 2025-11-27
tbi(store): updates to backend system properties

cianmSAP:
Interal ticket 36008

Key points: 
- Adds new property `hasSensitiveData` which is determined when writing based on the presence of sensitive data - this can be used by consumers to know whether secure store calls are required
- Update to the BackendSystem entity to strongly type the `authenticationType` and `systemType` and new type `connectionType`, rather than strings
- Adds new retrieval option `backendSystemFilter` which facilitates filtering the systems based on the seriazable keys on `BackendSystem` i.e 
    ```
    | 'name'
    | 'url'
    | 'client'
    | 'systemType'
    | 'authenticationType'
    | 'connectionType';

### DIFF
--- a/packages/jest-environment-ui5/src/index.js
+++ b/packages/jest-environment-ui5/src/index.js
@@ -2,7 +2,7 @@ const JSDOMEnvironment = require('jest-environment-jsdom').default;
 const { initTsConfigMappingStrategy } = require('./utils/tsMappingStrategy');
 const { initUi5MappingStrategy } = require('./utils/ui5MappingStrategy');
 const { initUI5Environment } = require('./env/ui5Environment');
-const path = require('path');
+const path = require('node:path');
 
 /**
  * Define the custom environment for the jest tests that runs in a UI5 environment
@@ -33,7 +33,7 @@ class UI5DOMEnvironment extends JSDOMEnvironment {
 
     /**
      * Prepare the environment for the test.
-     * This is called before each test and will setup the UI5 environment.
+     * This is called before each test and will set up the UI5 environment.
      * @returns {Promise<void>} A promise that resolves when the environment is ready
      */
     async setup() {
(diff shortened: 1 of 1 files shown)


packages/backend-proxy-middleware/src/base/proxy.ts
```diff
@@ -399,13 +399,13 @@ async function updateProxyConfigFromStore(
...
-            name: '<unknown>',
-            url: localBackend.url,
-            authenticationType: localBackend.authenticationType
-        };
+        const system =
+            (await systemStore.read(new BackendSystemKey({ url: localBackend.url, client: localBackend.client }))) ??
+            ({
+                name: '<unknown>',
+                url: localBackend.url,
+                authenticationType: localBackend.authenticationType
```
IainSAP: This needs validation as the yaml file writer does not use the same types. Casting can cause errors downstream of this since it used to be a string but is now a required type. This can be tested by changing the value in ui5.yaml to something other than a supported string value (we have seen this in customer apps) and previewing.

cianmSAP: I've added a validation check on `authenticationType`

examples/simple-generator/src/app/service.ts
```diff
@@ -71,7 +72,7 @@ async function askToStoreCredentials(generator: Generator, provider: AbapService
     if (store) {
         await storeCredentials(
             name,
-            { url: provider.defaults.baseURL!, client: provider.defaults.params?.['sap-client'] },
+            { url: provider.defaults.baseURL!, client: provider.defaults.params?.['sap-client'] } as BackendSystem,
```
longieirl, line 75: From past experience, I've seen issues with the client being passed as undefined | ``, where using undefined caused odd behaviour, wondering if you have seen anything similar?

cianmSAP: I have in other cases, just to note this is a non-productive sample generator, I can't assign the appropriate properties as they are unknown and not required in this case, so I'm casting the object

packages/store/src/services/index.ts
```diff
@@ -1,17 +1,68 @@
...
+
+/**
+ * The backend system keys and their values to filter backend systems by.
+ */
+export type BackendSystemFilter = Partial<{ [K in BackendSerializableKeys]: BackendSystem[K] }>;
+
+/**
+ * Specifies options for retrieving backend systems from the service.
+ */
+export interface BackendServiceRetrievalOptions {
```
IainSAP, line 12: Should this type not be specific to BackendSystems . i.e. 
```
interface BackendServiceRetrievalOptions extends ServiceRetrievalOptions{
```

How would you get other entities now to include sensitive data since the `ServiceGetAllOptions` requires that  you can only set this property for `BackendSystems`

Maybe other service types dont include sensitive data?

cianmSAP: Yeah I can make it consistent across the different entities, I just didn't think there was a use case for them. But it makes sense to keep it consistent, as they all use the same `HybridStore` -> `getAll` eventually

PR #3879, opened 2025-11-28
fix(middlewares): add 'express' as peerDependency for middlewares

heimwege:
Let's use `peerDepencency` for `express` in our middlewares as the actual package comes from UI5 cli anyway.

Adjustments for 
- backend-proxy-middleware
- reload-middleware
- ui5-proxy-middleware

Already correct:
- serve-static-middleware
- preview-middleware


nikmace (approved):
Changes look good
Changeset present
Did not test manually

PR #3881, opened 2025-11-28
fix(fpm-writer): unique ids in fragment

eouin:
Internal issue 36418

- [x] remove id in fragment.xml for filter to avoid ids duplication in case of using existing fragment

### DIFF
--- a/packages/fiori-mcp-server/CHANGELOG.md
+++ b/packages/fiori-mcp-server/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/fiori-mcp-server
 
+## 0.4.5
+
+### Patch Changes
+
+-   9d0c8e0: fix(fiori-mcp): fix missing dependency
+
 ## 0.4.4
 
 ### Patch Changes
--- a/packages/fiori-mcp-server/package.json
+++ b/packages/fiori-mcp-server/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/fiori-mcp-server",
     "description": "SAP Fiori - Model Context Protocol (MCP) server",
-    "version": "0.4.4",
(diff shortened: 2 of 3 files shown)


packages/fe-fpm-writer/test/unit/__snapshots__/filter.test.ts.snap
```diff
@@ -49,14 +49,15 @@ Object {
 
 exports[`CustomFilter generateCustomFilter Create several new custom filters 2`] = `
 "<core:FragmentDefinition xmlns:core=\\"sap.ui.core\\" xmlns=\\"sap.m\\">
-        <ComboBox id=\\"testID2\\">
+        <ComboBox  id=\\"testID2\\"
```
815are: I think we do not need additional whitespace

eouin: Done

packages/fe-fpm-writer/test/unit/__snapshots__/filter.test.ts.snap
```diff
@@ -102,14 +103,15 @@ Object {
 
 exports[`CustomFilter generateCustomFilter New custom filter (no eventhandler) 2`] = `
 "<core:FragmentDefinition xmlns:core=\\"sap.ui.core\\" xmlns=\\"sap.m\\">
-        <ComboBox id=\\"testID\\">
+        <ComboBox id=\\"testID\\"
```
815are: `>` is missing

eouin: Done

815are (approved):
checked:
1. code changes straight forward
2. changelog persists

PR #3889, opened 2025-12-02
Fix import, enhance passing fs instance

sap-sebelao:


sap-sebelao:
As of https://github.com/SAP/open-ux-tools/pull/3889/commits/5a7a87fdc93621081d6da6508a7837043f13b067 existing test-input samples should now work with `project-access`.

I added some convenience getters to `project-access` as well because the pages is no longer an array.

I tried to follow the example of filters to also generate a test for table columns, but it seems the test API method `iCheckColumns`
https://ui5.sap.com/#/api/sap.fe.test.api.TableAssertions%23methods/iCheckColumns
expects a structure rather than just keys of the columns, and a similar method to check just one column like in the case of Filters also does not seem to exist, so we may have to further adapt and transform the data we get to a structure useable by `iCheckColumns`

Issue #3858, opened 2025-11-24
BUG - MCP Server - docs tool does not work as intened


marianfoo:
### Related Feature

_issue number_

### Description

The `search_docs` tool is returning broken results.

### Steps to Reproduce

1. install `@sap-ux/fiori-mcp-server@latest` globally
2. Connect using MCP Inspector
3. Call `search_docs`:

```json
{
  "method": "tools/call",
  "params": {
    "name": "search_docs",
    "arguments": {
      "query": "CDS",
      "maxResults": 25
    }
  }
}
```

4. Results are all broken fragments

### Expected results

Should return proper docs with full text, titles

### Actual results

Got 25 results, all incomplete fragments:

```
Result 1:
CDS.
---
Result 2:
CDS.
---
Result 3:
P CDS.
---
Result 4:
CAP CDS.
---
Result 5:
CAP CDS.
---
Result 6:
AP CDS.
---
Result 7:
ot CAP CDS.
---
Result 8:
t CAP CDS.
---
Result 9:
not CAP CDS.
---
Result 10:
not CAP CDS.
---
Result 11:
cds` in the linked example\).
---
Result 12:
, not CAP CDS.
---
Result 13:
.cds` in the linked example\).
---
Result 14:
cds-annotations/)
---
Result 15:
cd4.md)
---
Result 16:
cd4.md)
---
Result 17:
e, not CAP CDS.
---
Result 18:

ce.cds` in the linked example\).
---
Result 19:

ce, not CAP CDS.
---
Result 20:
ervice, not CAP CDS.
---
Result 21:

e.cds` in the linked example\).
---
Result 22:

-cds-annotations/)
---
Result 23:

a service, not CAP CDS.
---
Result 24:
service, not CAP CDS.
---
Result 25:
service, not CAP CDS.
---

```

### Screenshots

Tested in MCP Inspector v0.17.2 

<img width="2089" height="1225" alt="Image" src="https://github.com/user-attachments/assets/3dfdcd64-cf0f-4778-8fe3-df7858b77746" />

#### Version/Components/Environment

**Package**: `@sap-ux/fiori-mcp-server`  
**Version**: 0.4.1 (global install)

**System**:
- OS: macOS (darwin 24.6.0)
- Node.js: v22.18.0
- npm: 10.3.0

### Problem

{describe the problem} 

### Fix

{describe the fix} 

### Why was it missed

{Some explanation why this issue might have been missed during normal development/testing cycle}  

[...]

PR #3897, opened 2025-12-03
fix: mcp server load the latest embeddings package on install/ npx

devinea:
npx / npm install did not update the embeddings package if already installed. 

fix https://github.com/SAP/open-ux-tools/issues/3858


heimwege (approved):
change lgtm

PR #3899, opened 2025-12-03
fix(fiori-mcp): add service metadata validation

tom-sap:
- validate EDMX before storing (and continuing the generation)
- also includes changes from this PR: https://github.com/SAP/open-ux-tools/pull/3898

### DIFF
--- a/examples/fe-fpm-cli/CHANGELOG.md
+++ b/examples/fe-fpm-cli/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/fe-fpm-cli
 
+## 0.1.67
+
+### Patch Changes
+
+-   @sap-ux/fe-fpm-writer@0.39.29
+
 ## 0.1.66
 
 ### Patch Changes
--- a/examples/fe-fpm-cli/package.json
+++ b/examples/fe-fpm-cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/fe-fpm-cli",
-    "version": "0.1.66",
+    "version": "0.1.67",
     "description": "A simple CLI to prompt required information to create a building block using the fe-fpm-writer module's prompt and generate functions.",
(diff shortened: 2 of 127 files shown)


devinea (approved):
Thanks @tom-sap 
Changes look good and covered by tests. 
didn't test manually.
changeset ✅

packages/fiori-mcp-server/src/tools/functionalities/fetch-service-metadata/service-metadata.ts
```diff
@@ -156,6 +157,24 @@ async function getServiceFromSystem(backendSystem: BackendSystem, servicePath: s
...
+function checkMetadata(metadata: string): void {
+    let parsedMetadata: unknown;
+    try {
+        parsedMetadata = parseEdmx(metadata);
+    } catch {
+        /* error handled below */
+    }
+    if (!parsedMetadata) {
+        throw new Error('Failed to parse service metadata. The service may not be a valid OData V4 service.');
+    }
```
heimwege, line 175: Question: Why not

```suggestion
    try {
        //maybe some eslint disable no-unused-expressions needed here
        parseEdmx(metadata);
    } catch {
        throw new Error('Failed to parse service metadata. The service may not be a valid OData V4 service.');
    }
```

tom-sap: because - in my version - I throw an error in 2 cases:
1. when parse throws (e.g. xml was not xml)
2. when parse worked, but metadata is empty.

From my quick glance, parser doesn't always throw an error and in some cases may not throw an error, but may still return nothing useful (but please, correct me if I'm wrong). This way I make sure that the parser returned something. 

There could probably be more checks (e.g. on metadata version, entities, etc.) but for now it offers a bit safer error handling IMHO.

PR #3901, opened 2025-12-03
fix: generate external service metadata

voicis:
Adding generation logic that was missed in #3794.


IainSAP (dismissed):
Thanks @voicis  LGTM

- Adds new service writing as expected
- Validated manually

PR #3905, opened 2025-12-04
fix(fiori-generator-shared): remove code assist from README

korotkovao:
Related to internal issue #36495

Now that the Code Assist prompt is removed during the generation, we should also remove it from the generated README file.


korotkovao:
Thank you @cianmSAP @mikicvi-SAP

PR #3906, opened 2025-12-04
Fix/fe fpm writer/custom filter fields in multiple filterbar

eouin:
Internal issue 36473

- [x] Allow adding custom filter field to multiple filter bar

### DIFF
--- a/packages/fe-fpm-writer/test/unit/building-block.test.ts
+++ b/packages/fe-fpm-writer/test/unit/building-block.test.ts
@@ -1,12 +1,13 @@
 import { create as createStorage } from 'mem-fs';
+import { promises as fsPromises } from 'node:fs';
 import { create, type Editor } from 'mem-fs-editor';
 import { join } from 'node:path';
-import { DOMParser } from '@xmldom/xmldom';
+
 import type { BuildingBlockConfig, Chart, Field, FilterBar, Table, CustomColumn, CustomFilterField } from '../../src';
+
 import { BuildingBlockType, generateBuildingBlock, getSerializedFileContent } from '../../src';
 import { BUILDING_BLOCK_CONFIG } from '../../src/building-block';
 import * as testManifestContent from './sample/building-block/webapp/manifest.json';
-import { promises as fsPromises } from 'node:fs';
 import { clearTestOutput, writeFilesForDebugging } from '../common';
 import {
     bindingContextAbsolute,


packages/fe-fpm-writer/test/unit/building-block.test.ts
```diff
@@ -2042,4 +2043,277 @@ describe('Building Blocks', () => {
...
+                        '<core:FragmentDefinition xmlns:core="sap.ui.core" xmlns="sap.m"><Input /></core:FragmentDefinition>'
+                }
+            };
+
+            // The aggregationPath points to non-existent element
+            const aggregationPath = "//*[@id='NonExistentFilterBar']";
+
+            // Test will verify behavior through integration test
+            expect(buildingBlockData.embededFragment).toBeDefined();
+        });
```
815are, line 2190: what is idea for such test? if it does not call `generateBuildingBlock`

eouin: Did some clean up

packages/fe-fpm-writer/test/unit/building-block.test.ts
```diff
@@ -2042,4 +2043,245 @@ describe('Building Blocks', () => {
...
+                aggregationName: 'filterFields',
+                elementName: 'FilterField',
+                resultPropertyName: 'hasFilterFields'
+            };
+
+            // Access the internal function through generateBuildingBlock workflow
+            // Since updateAggregationPath is not exported, we test it indirectly
+            // For now, this test validates the expected behavior
+            expect(buildingBlockData.embededFragment).toBeUndefined();
+        });
```
815are: sorry, same here like -> https://github.com/SAP/open-ux-tools/pull/3906#discussion_r2590270898
there is no call for `generateBuildingBlock` and check seems just checks `1 === 1`  or maybe I misunderstand idea of test?

eouin: Done, removed

packages/fe-fpm-writer/test/unit/building-block.test.ts
```diff
@@ -2042,4 +2043,213 @@ describe('Building Blocks', () => {
...
+            expect(updatedContent).toContain('FilterBar2');
+
+            // Verify FilterBar1 still has only its original field
+            const filterBar1Match = updatedContent.match(
+                /<macros:FilterBar id="FilterBar1"[\s\S]*?<\/macros:FilterBar>/
+            );
+            if (filterBar1Match) {
+                expect(filterBar1Match[0]).toContain('field1');
+                expect(filterBar1Match[0]).not.toContain('newFieldInBar2');
+            }
```
815are: I would suggest to use snapshots for `updatedContent`? It is more readable through snapshots as you can see whole update HTML in snapshots instead of checks for `toContain`

815are (commented):
@eouin 

I tried to revert code in `packages\fe-fpm-writer\src\building-block\index.ts` and run new tests with old code(without fix) - tests are still passing. Seems tests does not contain problematic scenario or expects not fully validates?

packages/fe-fpm-writer/test/unit/building-block.test.ts
eouin: should be good now

packages/fe-fpm-writer/test/unit/building-block.test.ts
```diff
@@ -1,6 +1,7 @@
 import { create as createStorage } from 'mem-fs';
 import { create, type Editor } from 'mem-fs-editor';
 import { join } from 'node:path';
+import { DOMParser } from '@xmldom/xmldom';
```
devinea: Is this needed. It is not referenced in the new code below.
Already included in other tests like this: 
`const { DOMParser } = await import('@xmldom/xmldom');`

eouin: Done

PR #3915, opened 2025-12-09
fix(project-access): usage of mem-fs-editor when creating application access instance

heimwege:
`createApplicationAccess` does not use the given mem fs instance everywhere. This can lead to application only existing in memory not being found.


Klaus-Keller (approved):
Thanks @heimwege!

* change to consider memfs already when searching apps makes sense
* changeset exists
* testcase is fixed

Approved from my side

PR #3916, opened 2025-12-09
Slight clean up of types, add JSDoc

sap-sebelao:
- own definition of types to the best of our abilities until ux-spec delivery

@docirl


sap-sebelao:
Already done in another commit in https://github.com/SAP/open-ux-tools/pull/3921

PR #3918, opened 2025-12-09
feat(mcp-server): use latest spec api which contains result of schema parsing

815are:
Changes:
1. `@sap-ux/fiori-mcp-server` - new `@sap/ux-specification` version `1.139.2` contains `applicationModel` which has same result of parsing which is placed localy in mcp-server here -> https://github.com/SAP/open-ux-tools/tree/main/packages/fiori-mcp-server/src/page-editor-api/parser/model. In result in this PR I am removing code from `fiori-mcp-server/src/page-editor-api/parser/model`
Faced some problem to access some interface - in result some interface was copied from specification like -> https://github.com/SAP/open-ux-tools/blob/feat/useLatestSpecApi/packages/fiori-mcp-server/src/page-editor-api/tree/types.ts#L37. Those duplication should be removed with next versions specification release 
2. `@sap-ux/project-access` - export method `getSpecificationModuleFromCache` to allow specifically load specification from global cache instead of application. It is used in mcp server [here](https://github.com/SAP/open-ux-tools/blob/feat/useLatestSpecApi/packages/fiori-mcp-server/src/page-editor-api/sapuxFtfsFileIO.ts#L67) - if loaded version from app has older specification version, then as fallback it loads latest specification from global cache https://github.com/SAP/open-ux-tools/blob/feat/useLatestSpecApi/packages/fiori-mcp-server/src/page-editor-api/sapuxFtfsFileIO.ts#L67

### DIFF
--- a/.changeset/floppy-zoos-cross.md
+++ b/.changeset/floppy-zoos-cross.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/app-config-writer': patch
----
-
-fix: improve logging for smartlinks config
--- a/packages/app-config-writer/CHANGELOG.md
+++ b/packages/app-config-writer/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/app-config-writer
 
+## 0.6.65
+
+### Patch Changes
+
+-   31591a1: fix: improve logging for smartlinks config
+
 ## 0.6.64
 
 ### Patch Changes
--- a/packages/app-config-writer/package.json
+++ b/packages/app-config-writer/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/app-config-writer",
     "description": "Add or update configuration for SAP Fiori tools application",
-    "version": "0.6.64",
+    "version": "0.6.65",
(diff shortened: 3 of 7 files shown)


packages/fiori-mcp-server/src/page-editor-api/tree/tree.ts
```diff
@@ -381,7 +369,7 @@ export function traverseTree(aggregation: ObjectAggregation, traverseNodeData: T
         title: aggregation.getTechnicalName(),
         properties: getProperties(aggregation),
         annotationNodeId,
-        type: getNodeType(aggregation),
+        // type: getNodeType(aggregation),
```
johannes-kolbe: Type commented out is intentional?

815are: thanks, removed -> 289b89c5999b670081dceb5b069bec21f890d931

johannes-kolbe (approved):
reapprove

Klaus-Keller (approved):
re-approve

PR #3921, opened 2025-12-10
feat(ui5-test-writer): add logic to generate OPA tests from project structure

docirl:
Internal issue 35852

- add basic code to call specification api to get model view of project
- use this model to retrieve controls and structure of project
- add OPA tests relevant to the project project structure


docirl:
@815are @Klaus-Keller pr is still in draft but please review the approach and usage of spec api and types, thanks

Klaus-Keller:
@815are @johannes-kolbe , can you check the consumption of `@sap/ux-specification`.

packages/project-access/src/project/model.ts
```diff
@@ -0,0 +1,103 @@
...
+ * @param pageModel - The tree model containing table column definitions.
+ * @returns The table columns aggregation object.
+ */
+export function getTableColumns(pageModel: TreeModel): TreeAggregations {
+    const table = getAggregations(pageModel.root)['table'];
+    const tableAggregations = getAggregations(table);
+    const columns = tableAggregations['columns'];
+    const columnAggregations = getAggregations(columns);
+    return columnAggregations;
+}
```
815are: Hi @Klaus-Keller 

I am not sure, those methods should be part of project-access? Or perhaps such methods should be part of model in specification?

packages/project-access/src/project/model.ts
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Retrieves selection field items from the given selection fields aggregation.
+ *
+ * @param selectionFieldsAgg - The selection fields aggregation containing field definitions.
+ * @returns An array of selection field descriptions.
+ */
+export function getSelectionFieldItems(selectionFieldsAgg: TreeAggregations): string[] {
+    if (selectionFieldsAgg && typeof selectionFieldsAgg === 'object') {
+        const items: string[] = [];
+        for (const itemKey in selectionFieldsAgg) {
```
815are: @docirl 
I am not sure if order of fields plays role for you or not.
Just in case there is method `aggregation.getAggregationKeys`, which returns ordered list with same order like it will be visible in application runtime/preview

packages/project-access/src/project/model.ts
```diff
@@ -0,0 +1,103 @@
...
+export function getObjectPages<T = ApplicationModel['pages'][string]>(applicationModel: ApplicationModel): T[] {
+    const objectPages: T[] = [];
+    for (const pageKey in applicationModel.pages) {
+        const page = applicationModel.pages[pageKey];
+        if (page.pageType === 'ObjectPage') {
+            objectPages.push(page as T);
+        }
+    }
+    return objectPages;
+}
```
815are: is this method used?

packages/project-access/src/project/model.ts
```diff
@@ -0,0 +1,103 @@
...
+    return getSelectionFieldItems(selectionFieldsAggregations);
+}
+
+/**
+ * Retrieves the table columns aggregation from the given tree model.
+ *
+ * @param pageModel - The tree model containing table column definitions.
+ * @returns The table columns aggregation object.
+ */
+export function getTableColumns(pageModel: TreeModel): TreeAggregations {
```
815are: @docirl 
`getTableColumns` return type is `TreeAggregations` while `getFilterFields` => `string[]`.
Should they be consistent if we will add more methods for reuse?

packages/project-access/src/project/model.ts
Klaus-Keller: Do we need this in multiple places? If yes we can move it to the model in specification.

docirl: I added in project-access in case there is a need for reuse, with the model now available from the spec api. I thought others might want to access the same data

packages/project-access/src/project/model.ts
docirl: At the moment the order is not important, but will keep this in mind. Thanks

packages/project-access/src/project/model.ts
docirl: Perhaps not at the moment as the tests generated are only on the list report, but there will be more tests added in follow up tasks

packages/project-access/src/project/model.ts
docirl: Sure, I'll update

docirl:
@815are @Klaus-Keller I've opened this pr, can you please review usage of spec and project access updates. Thanks

packages/fiori-app-sub-generator/package.json
```diff
@@ -59,6 +59,7 @@
         "yeoman-generator": "5.10.0"
     },
     "devDependencies": {
+        "@sap/ux-specification": "1.139.2",
```
IainSAP: Theres not references to this package in this module so theres no dependency required.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
```diff
@@ -2,11 +2,27 @@ import { join } from 'node:path';
...
+    FileName,
+    DirName,
+    getListReportPage,
+    getFilterFields,
+    createApplicationAccess,
+    getTableColumns
+} from '@sap-ux/project-access';
+import type { Logger } from '@sap-ux/logger/src/types';
+import type { ReadAppResult, Specification } from '@sap/ux-specification/dist/types/src';
+import type { PageWithModelV4 } from '@sap/ux-specification/dist/types/src/parser/application';
```
IainSAP: Why the deep imports? Types should be properly exported from @sap/ux-specification?

docirl: These are not exported directly from spec and are deep imported in mcp-server also https://github.com/SAP/open-ux-tools/blob/c7f9a60ac49b4e7159b429c040257a3b9dd77798/packages/fiori-mcp-server/src/page-editor-api/sapuxFtfsFileIO.ts#L17

@815are @Klaus-Keller are these types able to be exported from spec?

packages/fiori-app-sub-generator/package.json
docirl: No longer needed, I'll remove

docirl: done

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
```diff
@@ -251,6 +267,157 @@ function writePageObject(
...
+    // Read application model to extract control information needed for test generation
+    // specification and readApp might not be available due to specification version, fail gracefully
+    try {
+        // readApp calls createApplicationAccess internally if given a path, but it uses the "live" version of project-access without fs enhancement
+        const appAccess = await createApplicationAccess(basePath, { fs: fs });
+        const specification = await appAccess.getSpecification<Specification>();
+        const appResult: ReadAppResult = await specification.readApp({ app: appAccess, fs: fs });
+        listReportPage = appResult.applicationModel ? getListReportPage(appResult.applicationModel) : listReportPage;
+    } catch (error) {
+        log?.warn(
```
IainSAP: Should we add a `generateOPAFiles` test for the case where an error is thrown by `createApplicationAccess` as I think it may cause issues later on with undefined properties:     https://github.com/SAP/open-ux-tools/pull/3921/changes#diff-a514ef89ae217dbd479f2c3be73fd0868906995695f09d5a02b2f33741fac7ecR484 

Or will it create partial tests in that case?

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
```diff
@@ -251,6 +267,157 @@ function writePageObject(
     );
 }
 
+/**
```
IainSAP: Is it worth considering moving all this small functions into a specific file for app analysis rather than having them all in this main writer file. Its getting a little busy?

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
```diff
@@ -251,6 +267,157 @@ function writePageObject(
...
+    } catch (error) {
+        log?.warn(
+            'Error analyzing project model using specification. No dynamic tests will be generated. Error: ' +
+                (error as Error).message
+        );
+        return featureData;
+    }
+
+    if (!listReportPage) {
+        log?.warn('No List Report page found in application model. No dynamic tests will be generated.');
```
IainSAP:
```suggestion
        log?.warn('List Report page found not in application model. Dynamic tests will not be generated.');
```

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
docirl: It will just not add the 'new' tests that are based on the project model, so will continue to generate the tests like they are now.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
Klaus-Keller: @815are, we should do so to get avoid the deep imports. But should not be a blocker for now

packages/project-access/src/project/model.ts
Klaus-Keller: @docirl, what I thought is to add the functionality to the parsed`ApplicationModel` object itself.
@815are, what do you think?

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
sap-sebelao: I think this is something we are aware of and have planned as one of the next things to look at.

packages/project-access/src/project/model.ts
docirl: Just to be clear we will likely need more util methods to access the model, as we continue to add more OPA tests. This PR is the first and we plan to add a lot more tests in the coming sprints. So where ever makes sense for these utils to be added, let us know

815are: yes, agree, I think makes sense to enhance Model with utils methods.
@docirl  But what is idea to have them in project-access?
What would be flow:
1. we enhance spec with additional methods;
2. once spec released -> OPA test suses methods from new spec, but it removes exported method from project-access?

I am not sure then why we need it in project-access? maybe better to place in `ui5-test-writer` while there no method inside Model?

docirl: @815are as discussed, I'll move the utils from `project-access` to `ui5-test-writer` for our usage now, with a plan that these methods (or similar output) will be provided directly by `spec`. When these are available we will update to use them from `spec` and we will remove them from `ui5-test-writer.`

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
docirl: I've moved these functions to their own file

docirl:
Thanks @IainSAP @cianmSAP

PR #3926, opened 2025-12-13
feat(ADP): Private cloud workflow.

avasilev-sap:
#3865

This is a breaking change which introduces the private cloud workflow for the Adaptation project. 

**Historical information** 
So far we use the ATO settings to determine the type of the system and we have two possible types cloudReady and onPremise system. A project was treated as cloud/onPrem project if the system we use to create the project is cloud or onPrem, actually we treat the system type the same way as project type, both notions overlap. This is going to change with the private cloud scenario.

**Private cloud** 
With the private cloud scenario we have a requirement to use the system info api call to determine what type of project could be created on a specific system. The system info call returns an array `adaptationProjectTypes` which contains all available types of projects which can be created. This creates inconsistency with the legacy approach because from now on we can create both, cloud and onPrem, projects for a system based on the system info api call.
For example there are systems for which the ATO settings gives - we are in an onPremise system but the system info api call returns adaptationProjectTypes: ['cloudReady', 'onPremise'] which means that both projects, cloud/onPrem, can be created. In other words with the ATO settings we are restricted to create only on type of project for a system, but from the system info api call result we can create both types of projects.

**Conclusion**
We must use the ATO settings only to detect the system type and we must use the system info api call to determine the project type and both notions project type and system type must be distinguished. When a system supports both types of project, we provide to the user the ability to select the desired project type from a prompt containing the possible project types available within the system, otherwise only one type of project can be created. Thus we no longer want to use the system type(ATO settings) to determine the project type.

Diagrams always can replace a lot of words with a single image:

(Legacy) Detect project type - cloud or onPremise. A system can support only one project type based on the ATO settings(the system type notion).

(Legacy) Detect project type

Detect project type - cloud or onPremise. A system can support both, cloud and onPrem, or only one of them based on the system info call.

(New) Detect project type

**What is changed?**
The affected areas in the project are:
1. The ADP generator.
2. The ADP-FLP generator which can run as a standalone or sub generator orchestrated by the ADP generator.
3. The Deployment generator which can run as a standalone or sub generator orchestrated by the ADP generator. 
4. The ADP cli - the `npx -y @sap-ux/create@latest add deploy-config path/to/adp/project` now uses the adp project type.

From now on  projectType is determined with the system info api call + user choice if both project types are supported by the system. 

"method": "GET",
"path": "/sap/bc/lrep/system_info",
"queryStringParameters": {
  "sap-client": [
    "010"
  ]
}

The systemType is determined from the ATO settings.

"method": "GET",
"path": "/sap/bc/adt/ato/settings",
"queryStringParameters": {
  "sap-client": [
    "010"
  ]
}

**1. ADP**
- Some refactoring is made to improve readability, code quality and removal of nested if statements.
- Add util method which determines the project type of an existing ADP project from the ui5.yaml config. A project is cloud if the builder key appears at the root level of the yaml, otherwise onPrem.
- In the ADP preview, where the CPE client is loaded in a browser, we use the projectType from the ui5.yaml instead of the systemType, here overlapping occur between the two notions in the legacy code.
- When we create the RTA instance for the ADP preview we use the projectType not the systemType.
- We change the filtering of applications, when cloud project is selected we filter projects for which the cloudDevAdaptationStatus is released, otherwise we display all applications from the system.
- Add util method for retrieving the supported project types for a system.
- In the config writer we leave the systemType usage intact. The ui5 version is determined not from the projectType but from the systemType as before.
- When a system supports both types of projects we display a new prompt for a projectType. The order of prompts is system > projectType > application.
- The prompt for a ui5 version continue to use the systemType as is.
- The FLP config prompts are available only for projects with projectType cloud.
- In case a system supports only one project type we display the project type as an additional message otherwise we show a dropdown with both project types, so no need for additional message in this case.
- Some optimizations are done, for example multiple sequential calls which are orthogonal are made parallel, for example in the load system data we can call the flexUiCapability and the system info call in parallel.
- The CLI generator is updated to use properly the projectType and systemType notions.
- When we made a decision whether to display extension project prompts we use the projectType not the systemType.

**2. ADP-FLP**
- We abort the generator execution in case the projectType is cloud. This happens when the generator is started as a standalone generator.
- When the generator is started as a sub-generator of the ADP generator then the FLP generator is started only when the attributed prompt with name `addFlpConfig` is presented, this prompt is presented only when cloud project is being selected or this is the only project type supported by the system. As a result the FLP generator runs only for a cloud project.

**3. Deployment**
In the Deployment generators some prompts when/validate hooks are affected. The systemType usage need to be replaced with the projectType. In case the deployment generator is started as a sub generator from the ADP generator the projectType is passed as a constructor parameter with the options, otherwise we use the ui5.yaml to determine the project type - this covers the case for a standalone deploy generator. The complete list of affected prompts:

1. ABAP Target Prompts (getAbapTargetPrompts)
destination - Destination selection (BAS only)
2. Authentication Prompts (getAuthPrompts)
password - Password input, in the password validate we need to do the actual validation of the system that is why this prompt is affected.
3. App Configuration Prompts (getAppConfigPrompts) - optional, can be hidden
ui5AbapRepo - UI5 ABAP repository name
description - Application description
5. Package Prompts (getPackagePrompts)
packageManual - Manual package input
packageAutocomplete - Package autocomplete/search
6. Transport Request Prompts (getTransportRequestPrompts)
transportInputChoice - Transport request input method choice
transportCliExecution - Hidden transport CLI execution
transportCreated - Newly created transport number (display only)

**Testing done:**
Manual verification:
- [ADP-Tests-Private-Cloud (1).md](https://github.com/user-attachments/files/24882560/ADP-Tests-Private-Cloud.1.md)
- FLP generator - both standalone and sub generator workflows are tested.
- Manually verified that the ui5.yaml and ui5-config.yaml are properly created when the deploy generator is started as sub-generator from the ADP generator:

| SRC system | ADP project type | TARGET system |
|----------|----------|----------|
| C | C | C
| C | C | M
| P | P | P
| P | P | M
| M | C | C
| M | C | M
| M | P | P
| M | P | M

* C = cloudReady, P = onPremise, M = mixed(C and P)

Screenshot 2026-01-27 at 13 55 33

Screenshot 2026-01-27 at 13 55 40

Screenshot 2026-01-27 at 13 56 48

### DIFF
--- a/.changeset/dependencies-GH-3651.md
+++ b/.changeset/dependencies-GH-3651.md
@@ -1,0 +1,6 @@
+---
+"@sap-ux-private/ui-prompting-examples": patch
+"@sap-ux/ui-components": patch
+---
+
+fix(deps): update dependency sanitize-html to v2.17.0
--- a/.changeset/dependencies-GH-4150.md
+++ b/.changeset/dependencies-GH-4150.md
@@ -1,0 +1,5 @@
+---
+"@sap-ux/axios-extension": patch
+---
+
+fix(deps): update dependency qs to v6.14.2 [security]
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -210,7 +210,7 @@ jobs:
               run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPMJS_TOKEN }}" > .npmrc
             - name: 'Publish to npmjs'
               id: changesetPublish
-              uses: changesets/action@v1.4.10
+              uses: changesets/action@v1.7.0
               with:
                   publish: pnpm ci:publish
               env:
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.182
+
+### Patch Changes
+
+- Updated dependencies [1a922e7]
+    - @sap-ux/fiori-freestyle-writer@2.5.51
+    - @sap-ux/fiori-elements-writer@2.8.68
+
 ## 1.1.181
 
 ### Patch Changes
--- a/examples/simple-generator/package.json
+++ b/examples/simple-generator/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/generator-simple-fe",
-    "version": "1.1.181",
+    "version": "1.1.182",
     "description": "Simple example of a yeoman generator for Fiori elements.",
     "license": "Apache-2.0",
     "private": true,
--- a/examples/ui-prompting-examples/package.json
+++ b/examples/ui-prompting-examples/package.json
@@ -23,7 +23,7 @@
         "@sap-ux/ui-components": "workspace:*",
         "@sap-ux/ui-prompting": "workspace:*",
         "react-markdown": "5.0.2",
-        "sanitize-html": "2.12.1"
+        "sanitize-html": "2.17.0"
     },
     "peerDependencies": {
         "react": ">=16.14.0",
@@ -47,7 +47,7 @@
         "@types/mem-fs-editor": "7.0.1",
         "@types/react": "16.14.55",
         "@types/react-dom": "16.9.24",
-        "@types/sanitize-html": "2.11.0",
+        "@types/sanitize-html": "2.16.0",
         "@types/uuid": "3.4.11",
         "@types/ws": "8.5.12",
         "babel-loader": "9.1.3",
--- a/packages/axios-extension/package.json
+++ b/packages/axios-extension/package.json
@@ -34,7 +34,7 @@
         "fast-xml-parser": "5.3.4",
         "lodash": "4.17.23",
         "open": "7.0.3",
-        "qs": "6.14.1",
+        "qs": "6.14.2",
         "xpath": "0.0.33",
         "@xmldom/xmldom": "0.8.10",
         "https-proxy-agent": "7.0.5",
--- a/packages/cap-config-writer/CHANGELOG.md
+++ b/packages/cap-config-writer/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/cap-config-writer
 
+## 0.12.63
+
+### Patch Changes
+
+- 1a922e7: adds option to prevent root package json updates for cap
+
 ## 0.12.62
 
 ### Patch Changes
--- a/packages/cap-config-writer/package.json
+++ b/packages/cap-config-writer/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/cap-config-writer",
     "description": "Add or update configuration for SAP CAP projects",
-    "version": "0.12.62",
+    "version": "0.12.63",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/cap-config-writer/src/cap-config/index.ts
+++ b/packages/cap-config-writer/src/cap-config/index.ts
@@ -20,10 +20,16 @@ export async function enableCdsUi5Plugin(basePath: string, fs?: Editor): Promise
     const packageJsonPath = join(basePath, 'package.json');
     const packageJson = (fs.readJSON(packageJsonPath) ?? {}) as Package;
 
+    const originalPackageJson = structuredClone<Package>(packageJson); // keep a copy of the package.json
+
     ensureMinCdsVersion(packageJson);
     await enableWorkspaces(basePath, packageJson);
     addCdsPluginUi5(packageJson);
 
-    fs.writeJSON(packageJsonPath, packageJson);
+    // only write if there are changes
+    if (JSON.stringify(originalPackageJson) !== JSON.stringify(packageJson)) {
+        fs.writeJSON(packageJsonPath, packageJson);
+    }
+
     return fs;
 }
--- a/packages/cap-config-writer/src/cap-config/types.ts
+++ b/packages/cap-config-writer/src/cap-config/types.ts
@@ -54,4 +54,5 @@ export interface CapProjectSettings {
      */
     enableCdsUi5Plugin?: boolean;
     enableTypescript?: boolean;
+    disableRootPackageJsonUpdates?: boolean;
 }
--- a/packages/cap-config-writer/src/cap-writer/updates.ts
+++ b/packages/cap-config-writer/src/cap-writer/updates.ts
@@ -1,6 +1,7 @@
 import type { CapServiceCdsInfo, CapProjectSettings } from '../cap-config/types';
 import { updateRootPackageJson, updateAppPackageJson } from './package-json';
 import { updateTsConfig } from './tsconfig-and-yaml';
+import { enableCdsUi5Plugin as writeCdsUi5Plugin } from '../cap-config';
 import type { Editor } from 'mem-fs-editor';
 
 /**
@@ -16,6 +17,7 @@ import type { Editor } from 'mem-fs-editor';
  * @param {boolean} capProjectSettings.sapux - Indicates if SAP UX is enabled.
  * @param {boolean} capProjectSettings.enableCdsUi5Plugin - Indicates if cds ui5 plugin should be added (default is true).  The cds ui5 plugin will only be added if the minimum cds version that supports it is present.
  * @param {boolean} capProjectSettings.enableTypescript - Indicates if TypeScript is enabled.
+ * @param {boolean} capProjectSettings.disableRootPackageJsonUpdates - Indicates if updates to the root package.json should be disabled. If true, the root package.json will not be updated with the sapux array or the cds watch scripts.
  * @returns {Promise<void>} A promise that resolves when the updates are applied.
  */
 export async function applyCAPUpdates(
@@ -29,11 +31,19 @@ export async function applyCAPUpdates(
         appId,
         sapux = false,
         enableCdsUi5Plugin = true,
-        enableTypescript = false
+        enableTypescript = false,
+        disableRootPackageJsonUpdates = false
     } = capProjectSettings;
 
-    // update root package.json
-    await updateRootPackageJson(fs, packageName, sapux, capService, appId, enableCdsUi5Plugin);
+    if (disableRootPackageJsonUpdates) {
+        // if root package.json updates are disabled, we will only write the cds ui5 plugin if enabled, but not update scripts or sapux array in the root package.jsonx
+        if (enableCdsUi5Plugin) {
+            await writeCdsUi5Plugin(capService.projectPath, fs);
+        }
+    } else {
+        // update root package.json
+        await updateRootPackageJson(fs, packageName, sapux, capService, appId, enableCdsUi5Plugin);
+    }
 
     if (enableTypescript) {
         // update tsconfig.json if TypeScript is enabled
--- a/packages/cap-config-writer/test/unit/cap-writer/updates.test.ts
+++ b/packages/cap-config-writer/test/unit/cap-writer/updates.test.ts
@@ -176,4 +176,38 @@ describe('Test applyCAPUpdates updates files correctly', () => {
             start: 'start command'
         });
     });
+
+    test('applyCAPUpdates should not update root package.json when disableRootPackageJsonUpdates is true but should still enable cds-ui5-plugin', async () => {
+        const testCapProject = 'test-cap-package-sapux';
+        const testOutput = join(__dirname, '../test-inputs', testCapProject);
+        const testProjectName = 'test-cap-app1';
+
+        const capService: CapServiceCdsInfo = {
+            ...capInfo,
+            projectPath: testOutput,
+            capType: 'Node.js'
+        };
+
+        const settings: CapProjectSettings = {
+            appRoot: join(__dirname, '../cap-writer/test-inputs', testCapProject),
+            packageName: testProjectName,
+            appId: `${testProjectName}-id`,
+            sapux: true,
+            enableCdsUi5Plugin: true,
+            disableRootPackageJsonUpdates: true
+        };
+
+        await applyCAPUpdates(fs, capService, settings);
+
+        const packageJsonPath = join(capService.projectPath, 'package.json');
+        const packageJson = fs.readJSON(packageJsonPath) as Package;
+
+        // watch script should NOT be added to root package.json when disableRootPackageJsonUpdates is true
+        const scripts = packageJson.scripts;
+        expect(scripts).toBeUndefined();
+
+        // sapux array should NOT be added to root package.json when disableRootPackageJsonUpdates is true
+        const sapUxArray = packageJson.sapux;
+        expect(sapUxArray).toBeUndefined();
+    });
 });
--- a/packages/create/CHANGELOG.md
+++ b/packages/create/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/create
 
+## 0.14.65
+
+### Patch Changes
+
+- Updated dependencies [1a922e7]
+    - @sap-ux/cap-config-writer@0.12.63
+
 ## 0.14.64
 
 ### Patch Changes
--- a/packages/create/package.json
+++ b/packages/create/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/create",
     "description": "SAP Fiori tools module to add or remove features",
-    "version": "0.14.64",
+    "version": "0.14.65",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/fiori-app-sub-generator/CHANGELOG.md
+++ b/packages/fiori-app-sub-generator/CHANGELOG.md
@@ -1,5 +1,17 @@
 # @sap-ux/fiori-app-sub-generator
 
+## 0.11.67
+
+### Patch Changes
+
+- 1a922e7: adds option to prevent root package json updates for cap
+- Updated dependencies [1a922e7]
+    - @sap-ux/fiori-freestyle-writer@2.5.51
+    - @sap-ux/fiori-elements-writer@2.8.68
+    - @sap-ux/cap-config-writer@0.12.63
+    - @sap-ux/odata-service-inquirer@2.18.2
+    - @sap-ux/ui5-application-inquirer@0.16.27
+
 ## 0.11.66
 
 ### Patch Changes
--- a/packages/fiori-app-sub-generator/package.json
+++ b/packages/fiori-app-sub-generator/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/fiori-app-sub-generator",
     "description": "A yeoman (sub) generator that can generate Fiori applications. Not for standalone use.",
-    "version": "0.11.66",
+    "version": "0.11.67",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/fiori-app-sub-generator/src/fiori-app-generator/transforms.ts
+++ b/packages/fiori-app-sub-generator/src/fiori-app-generator/transforms.ts
@@ -40,6 +40,7 @@ import {
 import type { Package } from '@sap-ux/project-access';
 import type { CapServiceCdsInfo } from '@sap-ux/cap-config-writer';
 import { hostEnvironment, getHostEnvironment } from '@sap-ux/fiori-generator-shared';
+import { isFeatureEnabled } from '@sap-ux/feature-toggle';
 
 /**
  * Get the writer template type from the Fiori App floorplan.
@@ -232,6 +233,10 @@ export async function transformState<T>(
         }
 
         if (service.capService) {
+            const disableCapRootPkgJsonUpdates = isFeatureEnabled(
+                'sap.ux.testBetaFeatures.disableCapRootPkgJsonUpdates'
+            );
+
             const { cdsUi5PluginInfo, ...capServiceInfo } = service.capService;
             appConfig.service.capService = {
                 ...capServiceInfo,
@@ -244,6 +249,10 @@ export async function transformState<T>(
                 appConfig.service.capService.cdsUi5PluginInfo.isWorkspaceEnabled = true;
                 appConfig.service.capService.cdsUi5PluginInfo.hasCdsUi5Plugin = true;
             }
+            appConfig.appOptions = {
+                ...appConfig.appOptions,
+                disableCapRootPkgJsonUpdates
+            };
         }
 
         if (
--- a/packages/fiori-app-sub-generator/test/unit/fiori-app-generator/transforms.test.ts
+++ b/packages/fiori-app-sub-generator/test/unit/fiori-app-generator/transforms.test.ts
@@ -11,6 +11,7 @@ import type { BasicAppSettings } from '@sap-ux/fiori-freestyle-writer/dist/types
 import type { Destination } from '@sap-ux/btp-utils';
 import { Authentication as DestinationAuthType } from '@sap-ux/btp-utils';
 import { getHostEnvironment, hostEnvironment } from '@sap-ux/fiori-generator-shared';
+import * as featureToggle from '@sap-ux/feature-toggle';
 
 jest.mock('@sap-ux/fiori-generator-shared', () => {
     return {
@@ -20,6 +21,13 @@ jest.mock('@sap-ux/fiori-generator-shared', () => {
             name: 'CLI',
             technical: 'CLI'
         })
+    };
+});
+
+jest.mock('@sap-ux/feature-toggle', () => {
+    return {
+        ...jest.requireActual('@sap-ux/feature-toggle'),
+        isFeatureEnabled: jest.fn()
     };
 });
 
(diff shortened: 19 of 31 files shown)


packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
```diff
@@ -142,7 +143,7 @@ export function showUi5AppDeployConfigQuestion(ui5AbapPromptOptions?: UI5AbapRep
         !ui5AbapPromptOptions?.hide &&
         ui5AbapPromptOptions?.hideIfOnPremise &&
         !PromptState.abapDeployConfig?.scp &&
-        !PromptState.abapDeployConfig?.isAbapCloud
+        ui5AbapPromptOptions?.adpProjectType === AdaptationProjectType.ON_PREMISE
```
cianmSAP: If you no longer require `isAbapCloud` then can this prompt's visibility not be determined in the generator, and simply use the `hide` property?

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
```diff
@@ -238,7 +239,7 @@ function defaultOrShowTransportQuestion(): boolean {
 export function showTransportInputChoice(options?: TransportInputChoicePromptOptions): boolean {
     if (
         options?.hideIfOnPremise === true &&
-        !PromptState.abapDeployConfig?.isAbapCloud &&
+        options?.adpProjectType === AdaptationProjectType.ON_PREMISE &&
```
cianmSAP: Same comment as above, can this be done earlier

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
```diff
@@ -276,7 +277,10 @@ export function defaultOrShowTransportListQuestion(
     return (
         transportInputChoice === TransportChoices.ListExistingChoice &&
         !isTransportListEmpty(PromptState.transportAnswers.transportList) &&
-        !(transportInputChoiceOptions?.hideIfOnPremise === true && PromptState?.abapDeployConfig?.isAbapCloud === false)
+        !(
+            transportInputChoiceOptions?.hideIfOnPremise === true &&
+            transportInputChoiceOptions?.adpProjectType === AdaptationProjectType.ON_PREMISE
```
cianmSAP: Same comment

packages/abap-deploy-config-inquirer/src/prompts/validators.ts
```diff
@@ -641,12 +660,16 @@ export function validateConfirmQuestion(overwrite: boolean): boolean {
...
-async function validatePackageType(input: string, backendTarget?: BackendTarget): Promise<boolean | string> {
-    const isAbapCloud = PromptState?.abapDeployConfig?.isAbapCloud;
-    if (!isAbapCloud) {
-        LoggerHelper.logger.debug(`System is OnPremise, skipping package "${input}" type validation`);
+async function validatePackageType(
+    input: string,
+    backendTarget?: BackendTarget,
+    adpProjectType?: AdaptationProjectType
+): Promise<boolean | string> {
+    if (adpProjectType === AdaptationProjectType.ON_PREMISE) {
```
cianmSAP: Why would `shouldValidatePackageType` be set to true, if you know `adpProjectType === AdaptationProjectType.ON_PREMISE` means the package type validation should be skipped?

packages/abap-deploy-config-inquirer/src/prompts/validators.ts
```diff
@@ -817,9 +840,36 @@ function shouldValidatePackageForStartingPrefix(
...
+
+/**
+ * Validates whether a provided type of an adaptation project can be deployed on the target system.
+ *
+ * @param {AdaptationProjectType} adpProjectType - The adaptation project type.
+ * @param {BackendTarget | undefined} backendTarget - The backend target representing the target system.
+ * @returns {Promise<boolean | string>} Promise resolved with true in case the validation succeed otherwise with a string
+ * containing an error message.
+ */
+async function validateAdpProjectType(
```
cianmSAP: This should be handled somewhere else

packages/abap-deploy-config-inquirer/src/types.ts
```diff
@@ -83,11 +83,15 @@ export enum promptNames {
     overwriteAbapConfig = 'overwriteAbapConfig'
 }
 
+interface AdpSharedOptions {
```
cianmSAP: We shouldn't have ADP specific types in abap inquirer

packages/abap-deploy-config-sub-generator/src/app/index.ts
```diff
@@ -93,8 +97,10 @@ export default class extends DeploymentGenerator {
         if ((this.env as unknown as YeomanEnvironment).conflicter) {
             (this.env as unknown as YeomanEnvironment).conflicter.force = this.options.force ?? true;
         }
-        if (!this.launchDeployConfigAsSubGenerator) {
-            await this._initializing();
+        if (this.launchDeployConfigAsSubGenerator) {
+            this.adpProjectType = this.options?.adpProjectType;
```
cianmSAP: This could be set in the constructor? And the other code remain as it is?

packages/abap-deploy-config-sub-generator/src/app/index.ts
```diff
@@ -182,15 +194,18 @@ export default class extends DeploymentGenerator {
                 shouldValidateFormatAndSpecialCharacters: this.isAdp
             };
             const promptOptions: AbapDeployConfigPromptOptions = {
-                ui5AbapRepo: { hideIfOnPremise: this.isAdp },
-                transportInputChoice: { hideIfOnPremise: this.isAdp },
+                ui5AbapRepo: { hideIfOnPremise: this.isAdp, adpProjectType: this.adpProjectType },
```
cianmSAP: `this.adpProjectType` can and should be passed once like `backendTarget` is passed in the call to `getPrompts` in `packages/abap-deploy-config-sub-generator/src/app/questions.ts`

packages/abap-deploy-config-inquirer/package.json
```diff
@@ -39,6 +39,7 @@
         "@sap-ux/store": "workspace:*",
         "@sap-ux/system-access": "workspace:*",
         "@sap-ux/ui5-config": "workspace:*",
+        "@sap-ux/adp-tooling": "workspace:*",
```
cianmSAP: I don't see this is used?

packages/abap-deploy-config-inquirer/src/prompts/validators.ts
```diff
@@ -641,12 +660,16 @@ export function validateConfirmQuestion(overwrite: boolean): boolean {
...
-    const isAbapCloud = PromptState?.abapDeployConfig?.isAbapCloud;
-    if (!isAbapCloud) {
-        LoggerHelper.logger.debug(`System is OnPremise, skipping package "${input}" type validation`);
+async function validatePackageType(
+    input: string,
+    backendTarget?: BackendTarget,
+    adpProjectType?: AdaptationProjectType
+): Promise<boolean | string> {
+    if (adpProjectType === AdaptationProjectType.ON_PREMISE) {
+        LoggerHelper.logger.debug(`Project is OnPremise, skipping package "${input}" type validation`);
```
cianmSAP: Please add to i18n

packages/abap-deploy-config-inquirer/src/types.ts
```diff
@@ -83,11 +83,15 @@ export enum promptNames {
...
+    adpProjectType?: AdaptationProjectType;
```
cianmSAP: Can you add this to `AbapDeployConfigCommonInquirerOptions` rather a new interface specific for adp

packages/abap-deploy-config-inquirer/src/prompts/validators.ts
avasilev-sap: I will explain this in details because it is not easy to understand. In the destination prompt validation - `validateDestinationQuestion`, we compare the selected adpProjectType with the supported project types from the system info call - systemInfo.adaptationProjectTypes. If the project type is among the supported types we continue the validation otherwise we display an error. In case the project type is onPrem then in order for the validation to be successful the system need to support onPrem or both onPrem and cloud projects. So if the adp project type is onPrem then the system supports onPrem or both onPrem and cloud projects and when we reach the `packageManual`, the destination prompt comes before the packageManual prompt, there is no need to do the cloud package validation because the system is not cloud only for sure.

packages/abap-deploy-config-inquirer/src/prompts/validators.ts
avasilev-sap: This is a log message we do not localize those messages if that's what you meant.

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
avasilev-sap: This is not doable because the prompt visibility is also determined by the PromptState.abapDeployConfig.scp - this is a static flag which is set from the scp prompt earlier so we cannot use the prompt's options.hide.

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
avasilev-sap: This is not doable because the prompt visibility is also determined by the PromptState.abapDeployConfig.scp - this is a static flag which is set from the scp prompt earlier so we cannot use the prompt's options.hide.

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
avasilev-sap: This is not doable because the prompt visibility is also determined by the PromptState.abapDeployConfig.transportList - this is a static flag which is set from the transportInputChoice prompt earlier so we cannot use the prompt's options.hide.

packages/abap-deploy-config-sub-generator/src/app/index.ts
avasilev-sap: fixed in dd7ee38cddcd426a781d08d157cce49f95f89710

packages/abap-deploy-config-inquirer/package.json
avasilev-sap: remove the dependency, fixed in dd7ee38cddcd426a781d08d157cce49f95f89710

packages/abap-deploy-config-inquirer/src/types.ts
avasilev-sap: fixed in d4bfffbd5e5717b1e96344f1b7cd2374470f36f7

packages/abap-deploy-config-sub-generator/src/app/index.ts
avasilev-sap: fixed in d4bfffbd5e5717b1e96344f1b7cd2374470f36f7

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
cianmSAP: My issue with this code is for other consumers, for example, if they wish to hide this prompt if the selected system is on premise. They were able to achieve this with the current option - `hideIfOnPremise` 

Now they would require to pass in the `adpProjectType`. 

I'm wondering why you can't use `hideIfOnPremise`. 

You do know prior to calling the prompts whether the `adpProjectType` is `AdaptationProjectType.ON_PREMISE`, and then the existing check determines whether it's on premise or not using `!PromptState.abapDeployConfig?.scp`. So why can't you set `hideIfOnPremise` to true if `ui5AbapPromptOptions?.adpProjectType === AdaptationProjectType.ON_PREMISE`

packages/abap-deploy-config-inquirer/src/prompts/validators.ts
```diff
@@ -67,16 +68,28 @@ async function validateSystemType(options?: TargetSystemPromptOptions): Promise<
...
     destinations?: Destinations,
     options?: TargetSystemPromptOptions,
-    backendTarget?: BackendTarget
+    backendTarget?: BackendTarget,
+    adpProjectType?: AdaptationProjectType
 ): Promise<boolean | string> {
     PromptState.resetAbapDeployConfig();
     await updateDestinationPromptState(destination, destinations, options, backendTarget);
+
+    const needAuth = await showUsernameQuestion(backendTarget);
```
cianmSAP: I would prefer not to make this call, as now it will happen twice for all consumers. 

I would either - update `validateAdpProjectType` to check for auth errors and handle appropriately. 

Or perform the check within `validateAdpProjectType`

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
avasilev-sap: fixed in ed1df44efc540b4a3522397c2494d755aaab5b84

packages/abap-deploy-config-inquirer/src/prompts/validators.ts
avasilev-sap: fixed in bb647a72f7fbb2b7fedeb4d1175d93fd5495587d, i've added proper error handling on each place the helper getSystemInfo method gets called, i had to refactor a bit the helper method itself to avoid error suppressions.

packages/abap-deploy-config-inquirer/src/service-provider-utils/system-info.ts
```diff
@@ -1,32 +1,16 @@
...
-        result.systemInfo = systemInfo;
-        return result;
-    } catch (e) {
-        LoggerHelper.logger.debug(t('errors.debugAbapTargetSystem', { method: 'getSystemInfo', error: e.message }));
-        if (isAxiosError(e) && e?.response?.status === 405) {
-            result.apiExist = false;
-        }
-        return result;
-    }
+export async function getSystemInfo(packageName?: string, backendTarget?: BackendTarget): Promise<SystemInfo> {
```
mmilko01: What is the reason for refactoring this function. Is there any existing logic that expects to get `result.apiExist = false;` in some cases? My main concern is regressions in existing functionality

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
```diff
@@ -138,12 +138,7 @@ export function showPasswordQuestion(): boolean {
  * @returns boolean
  */
 export function showUi5AppDeployConfigQuestion(ui5AbapPromptOptions?: UI5AbapRepoPromptOptions): boolean {
-    if (
-        !ui5AbapPromptOptions?.hide &&
-        ui5AbapPromptOptions?.hideIfOnPremise &&
-        !PromptState.abapDeployConfig?.scp &&
-        !PromptState.abapDeployConfig?.isAbapCloud
-    ) {
+    if (!ui5AbapPromptOptions?.hide && ui5AbapPromptOptions?.hideIfOnPremise && !PromptState.abapDeployConfig?.scp) {
```
cianmSAP, line 141: Safest option for now might be to revert the change here. `scp` and `isAbapCloud` could and should probably be consolidated, but to not inadvertently cause any bugs I think it's best to leave it as it was i.e include `!PromptState.abapDeployConfig?.isAbapCloud`

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
```diff
@@ -236,11 +231,7 @@ function defaultOrShowTransportQuestion(): boolean {
  * @returns boolean
  */
 export function showTransportInputChoice(options?: TransportInputChoicePromptOptions): boolean {
-    if (
-        options?.hideIfOnPremise === true &&
-        !PromptState.abapDeployConfig?.isAbapCloud &&
-        !PromptState.abapDeployConfig?.scp
-    ) {
+    if (options?.hideIfOnPremise && !PromptState.abapDeployConfig?.scp) {
```
cianmSAP, line 234: Same for here and other instances

packages/abap-deploy-config-inquirer/src/service-provider-utils/system-info.ts
mmilko01: As discussed, this function is used only for ADP scenario and the handling is implemented there

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
avasilev-sap: This comment need to be dropped explaining why. The old api isAbapCloud() (which relies on the ATO setting) does not give correct results, for example for the u1y_010 system the isAbapCloud() returns false which means the system is onPrem and we assume in the legacy code that we can deploy only onPremise projects which is wrong. The system info api call for that particular system returns in the `adaptationProjectTypes` both types - cloudReady and onPremise, this means that in the system we can create both types of projects nevertheless the system type is onPremise based on the ATO settings. This is the new behavior which is the essence of the private cloud scenario.

PS 
In the destination prompt which is the first prompt in the deploy wizard we do validate the project type. For the ADP project we get the project type and the `adaptationProjectTypes` value for the target system(the destination) and if the projectType is in the array it means the project can be deployed otherwise we will not reach the `ui5AbapRepo` `description` or `transportInputChoice` prompts at all.

packages/abap-deploy-config-inquirer/src/prompts/conditions.ts
avasilev-sap: This also is not needed, explained in the above comment.

packages/abap-deploy-config-inquirer/src/translations/abap-deploy-config-inquirer.i18n.json
```diff
@@ -148,8 +148,8 @@
             "abapInvalidAppNameNamespaceOrStartingPrefix": "The SAPUI5 ABAP Repository and Package must start in the same way. Use either a namespace or capital letters: 'Z', 'Y' or 'SAP'.",
             "invalidCloudPackage": "The provided package is not intended for cloud deployment.",
             "invalidOnPremPackage": "The provided package is not intended for on-premise deployments.",
-            "invalidOnPremSystem": "You have selected a cloud system but you previously created a project residing on an on-premise system. Please select an on-premise system to proceed with the deployment.",
-            "invalidCloudSystem": "You have selected an on-premise system but you previously created a project residing on a cloud system. Please select a cloud system to proceed with the deployment."
+            "unsupportedAdpProjectType": "You cannot deploy {{adpProjectType}} project because the system supports only {{supportedAdpProjectTypes}}.",
```
lfindlaysap:
```suggestion
            "unsupportedAdpProjectType": "You cannot deploy a {{adpProjectType}} project because the system only supports {{supportedAdpProjectTypes}}.",
```

lfindlaysap (dismissed):

cianmSAP (dismissed):
Updates look good - all comments addressed
I reviewed abap deployment related modules only

packages/abap-deploy-config-inquirer/src/translations/abap-deploy-config-inquirer.i18n.json
avasilev-sap: fixed in 86553b81b846a0ccc32777d1e55aecba773f0795

mmilko01 (approved):
Changeset OK
Changes look good
Good test coverage
Did not test manually

packages/abap-deploy-config-inquirer/src/translations/abap-deploy-config-inquirer.i18n.json
```diff
@@ -148,8 +148,8 @@
             "abapInvalidAppNameNamespaceOrStartingPrefix": "The SAPUI5 ABAP Repository and Package must start in the same way. Use either a namespace or capital letters: 'Z', 'Y' or 'SAP'.",
             "invalidCloudPackage": "The provided package is not intended for cloud deployment.",
             "invalidOnPremPackage": "The provided package is not intended for on-premise deployments.",
-            "invalidOnPremSystem": "You have selected a cloud system but you previously created a project residing on an on-premise system. Please select an on-premise system to proceed with the deployment.",
-            "invalidCloudSystem": "You have selected an on-premise system but you previously created a project residing on a cloud system. Please select a cloud system to proceed with the deployment."
+            "unsupportedAdpProjectType": "You cannot deploy {{adpProjectType}} project because the system only supports {{supportedAdpProjectTypes}}.",
```
lfindlaysap:
```suggestion
            "unsupportedAdpProjectType": "You cannot deploy a {{adpProjectType}} project because the system only supports {{supportedAdpProjectTypes}} projects.",
```

lfindlaysap (changes requested):
@avasilev-sap, small tweak.

packages/abap-deploy-config-inquirer/src/translations/abap-deploy-config-inquirer.i18n.json
avasilev-sap: fixed in 1fb19fada6033f8215c452afa8819c2d1b7d56b8

packages/adp-tooling/src/source/applications.ts
```diff
@@ -1,9 +1,39 @@
...
+    'fileType': 'appdescr_variant',
+    'originLayer': 'VENDOR'
+};
+
+const CLOUD_ONLY_APPS_FILTER: UI5AppFilter = {
+    fields: APP_FIELDS,
+    'sap.ui/technology': 'UI5',
+    'sap.app/type': 'application',
+    'sap.fiori/cloudDevAdaptationStatus': 'released'
+};
```
nikmace, line 36: The new **type** should be moved to `types.ts`, there is already duplication of these constants in tests - they need to be moved back to `constants.ts`. And exported from there to be used, as I see in two places.

avasilev-sap: Thanks for the comment. Those constants are used only in the `loadApps` function and nowhere else in the product code, they need to be private and not exported as global const for the module, storing them in a separate file is redundant. Making the constants public just to make the testing easy is a bad practice and could lead to false positive results in the tests. The test should not dictate the public interface and break encapsulation, but quite the opposite. If those constants are made public and directly imported in the test if somebody changes the filters by accident the test will still pass silently without registering a regression, hardcoding them inside the test is not a bad practice but a second barrier where we double check the values again.

Do you have plans to use somewhere else in the product code those constants, I think they are pretty specific to the app filtering? Another remark, public constants which are initialized with objects can be imported and mutated in other files which could lead to unexpected behaviour so if they are meant to be private they should stay private. Do you have strong arguments, other than the test usage why these constants need to be public?

PS I also noticed in the legacy code that the previously named `mapApps` and `filterApps` are made public just for the testing which is also a bad practice since those functions are not imported elsewhere in the product and specific to the `loadApps` function. They can be tested as part of the tests for the `loadApps` function as it is done now in this PR.

nikmace: On keeping the constants in constants.ts and using them in tests (no duplication):

**1. Single source of truth**
The search params are configuration for the `app index API`. Keeping them in `constants.ts` gives us place for that configuration. If we change a field or filter, we change it once; production and tests both use the same value. Duplicating the same objects in tests means every change has to be done in two places, and tests can drift from what `loadApps` actually uses.

**2. What the test is asserting**
The test’s goal is: “when we call `loadApps`, the provider’s search is called with the correct params for this scenario.” “Correct” is “the params defined for this package” (i.e. the constants). So using the same constants in the test is correct: we’re asserting that `loadApps` passes the official params, not an ad‑hoc copy. If someone changes the constants and the behavior is still correct, the test should still pass; if the change is wrong, the test fails. Hardcoding a second copy in the test doesn’t add a “second barrier”—it adds a second place that can get out of date and hides real regressions when the test keeps passing with stale values.

**3. Encapsulation vs shared configuration**
These constants are not implementation details of loadApps; they’re shared API/search configuration. Putting them in constants.ts and importing them in both applications.ts and the test is the same pattern we use for other shared constants in the codebase. Keeping them “private” inside applications.ts and then copying them in the test doesn’t improve encapsulation; it just duplicates the same configuration.

**4. Mutation risk**
If we’re concerned about mutation, we can freeze the objects in constants.ts (e.g. Object.freeze or a `as const`-style type) and document that they’re read-only. Making them module-private and duplicating in tests doesn’t reduce mutation risk; it only duplicates the values.

**5. Type in types.ts**
If this PR introduces a new type that’s used for these params (or for app filtering), it should live in types.ts with the rest of the shared types for consistency and reuse.

---

Keeping the constants in `constants.ts`, exporting them, and using them in both `applications.ts` and the tests avoids duplication and keeps one source of truth. The test should verify that `loadApps` uses those params; that’s best done by importing and asserting against the same constants.

On `mapApps` / `filterApps`: I didn’t ask for that change in my comment; I can agree with it in principle, but it should follow the repo’s existing patterns.

More importantly, this PR shouldn’t be refactoring code that isn’t part of its scope. With 68 files changed, the review is already difficult. Would be better if these were handled in a follow-up PR so this one stays focused.

avasilev-sap: I do not agree with this let's discuss offline, also the PR does not do refactoring out of the scope i had to rework a lot of stuff due to the private cloud workflow.

nikmace: Talked offline, decided to keep the current implementation.

cianmSAP (approved):
Agreed offline to refactor some logic in abap inquirer as a follow up
All other comments addressed
Changeset and tests look good

lfindlaysap (approved):

packages/ui5-config/src/ui5config.ts
```diff
@@ -725,6 +725,19 @@ export class UI5Config {
...
     }
 
+    /**
+     * Determines whether the UI5 configuration document contains the builder key.
+     *
+     * @returns True if the configuration contains builder key at root level otherwise false.
+     */
+    public hasBuilderKey(): boolean {
+        try {
+            return !!this.document.getNode({ path: 'builder' });
```
heimwege, line 735: You are not interested in the value?
`builder:` and `builder: customTask #...` are treated identical currently

And why should this function become part of the `ui5-config` package instead of using the config instance to check for the `builder` yourself? Checking if a specific key is contained in the yaml doesn't sound like a generic function to me that needs to be provided by this package.

PR #3927, opened 2025-12-15
Feat(middlewares): add support for ui5 project type 'component' (@ui5/cli v5)

heimwege:
@ui5/cli v5 introduces a new project type `component` for `manifest.json` type `application`. Unlike the legacy `application` type (which serves from `webapp/`), component projects serve `src/` at `/resources/<namespace>/` and `test/` at `/test-resources/<namespace>/`, enabling multi-app serving from a single server instance.

Details: https://ui5.github.io/cli/v5/updates/migrate-v5

## Summary

This PR adds support for the UI5 CLI v5 type: component across the preview-middleware, ui5-proxy-middleware, project-access, ui5-config, app-config-writer, and adp-tooling packages. Unlike the legacy type: application projects that serve from webapp/, component projects serve source files at /resources/<namespace>/ and test files at /test-resources/<namespace>/, enabling multiple apps to run on a single server instance. The preview-middleware was extended to compute namespace-aware URL paths for all virtual endpoints (FLP, RTA editors, card generator, flex API, test runners) so they are correctly registered under the app's namespace prefix for component type projects. The ui5-proxy-middleware received a new filter that prevents app source files from being accidentally forwarded to the UI5 CDN, which would otherwise return 404s since the CDN has no knowledge of user-defined namespaces.
The path-adjustment functions centralise all path logic into the constructor rather than scattering it across route handlers.
A new simple-component integration test fixture and a dedicated Playwright spec (component.spec.ts) verify end-to-end that the FLP, QUnit, and OPA5 virtual endpoints all work correctly for type: component projects.

### `preview-middleware`
  - **`createTestTemplateConfig()`**: split single `basePath` into `rootBasePath` (server-root-relative, for `<script src=...>` tags) and `appBasePath` (`/resources/<ns>` for component type, for `data-sap-ui-resourceroots`). Component type gets an additional resourceroots entry mapping `<ns>/test → /test-resources/<ns>` for test path routing (`testNs` + `testResourcesPath` fields in `TestTemplateConfig`)
  - **`qunit.ejs`**: uses `rootBasePath` for scripts/CSS, `appBasePath` for resourceroots, conditionally adds second resourceroots entry for component type
  - **`qunit.js`** (init script): retains shims + `sap.ui.require([qunit-2,...], fn)` wrapper - replacing with plain `<script>` tags breaks OPA5 due to AMD race condition
  - **`generateImportList()`**: strips `/test-resources/<ns>/` prefix for component type and prepends `<ns>/test/` to produce correct AMD module names matching the `<ns>/test` resourceroot key
  - **`mergeTestConfigDefaults()`**: adjusts default paths/pattern for component type via `getTestResourcesPathPrefix()`
  - **`TemplateConfig`**: `appBasePath` (relative for application type, absolute `/resources/<ns>` for component type); `rootBasePath` (always server-root-relative); `baseUrl` set per-request from `req['ui5-patched-router']?.baseUrl` (provided by `cds-plugin-ui5`) combined with `getResourcesPathPrefix()` for component type
  - **Endpoint namespace isolation**: `createFlexHandler`, `addStoreCardManifestRoute`, `addStoreI18nKeysRoute` all use `getResourcesPathPrefix` to register routes under `/resources/<ns>/...` for component type. The `data-open-ux-preview-base-url` attribute on `#sap-ui-bootstrap` carries the full prefix, so `WorkspaceConnector`, `enableFakeConnector`, and the SAP Cards AP generator CDN library all POST to the correct namespaced endpoints automatically - no changes required outside this repo
  - **`isDefaultInitPath()`**: replaced with `isCustomInit: boolean` flag on `CompleteTestConfig` for cleaner init path tracking

  ### `ui5-proxy-middleware`
  - **`filterExcludeComponentNamespace()`**: new filter that excludes `/resources/<ns>` and `/test-resources/<ns>` from being forwarded to the UI5 CDN for component type projects — prevents app source files being proxied to `ui5.sap.com` and returning 404
  - **`createCombinedProxyFilter()`**: wires `filterExcludeComponentNamespace` into all proxy instances; now also handles glob patterns correctly via `micromatch` + `is-glob`
  - **`middlewareUtil`** passed through to `ui5Proxy()` so namespace is resolved at runtime via `getProject().getNamespace()` — UI5 CLI v5 auto-derives namespace from `manifest.json` `sap.app.id` (dots → slashes); no explicit `namespace:` field needed in `ui5.yaml`

  ### `project-access`
  - **`getPathMappings()`**: new export returning the `resources.configuration.paths` mappings from `ui5.yaml` for a given project

  ### `ui5-config`
  - Extended `Ui5YamlDocument` types to cover the `component` project type and its `paths` configuration

  ### `app-config-writer`
  - Updated virtual HTML file ejection to handle component type path structure

  ### `adp-tooling`
  - Updated to handle component type projects

  ## Decisions
  - There exist two other approaches for the test sources:
  1. `sap.ui.loader.config` override - rejected (overriding ui5 boot loader config is not preferred)
  2. resource root `my/app/test` → `/test-resources/my/app/test` - rejected (adding /test to the resource root forced having an additional `test` folder in the actual `test` folder which results in a `test/test/`) double nesting)
  3. **Chosen solution**: resourceroot key `my/app/test` → `/test-resources/my/app` - the `/test` is appended to the *key* only, so test files live flat in `test/` on disk and module IDs use the `<ns>/test/` prefix to route to `/test-resources/` without conflicting with the component's own `/resources/` entry

  ## Test fixture: `simple-component`
  - Serves `src/` at `/resources/my/fe/v2/app/` and `test/` at `/test-resources/my/fe/v2/app/`
  - Journey files live under `test/integration/` — the AMD module ID convention uses a `<ns>/test/` prefix (e.g. `my/fe/v2/app/test/integration/pages/ListReport`) which is routed via the `my/fe/v2/app/test → /test-resources/my/fe/v2/app` resourceroot entry; the `/test` in the module ID is consumed by the key so files live flat in `test/` with no extra subfolder nesting
  - Integration test (`component.spec.ts`) uses `@ui5/cli 5.0.0-alpha.4` and loads `ui5-proxy-middleware` from the local dist via a yaml `kind: extension` entry so `filterExcludeComponentNamespace` is active during testing

## Follow-up
- Update @sap/ux-ui5-tooling `fiori-tools-preview` and `fiori-tools-proxy` middleware after consumption of this change

## Test setup

use node `22.22.0`

adjust ui5 yaml file
- `specVersion: '5.0'`
- ```yaml
  type: component
  resources:
    configuration:
      paths:
        src: src    # default; path mapping can be done here; will be served at /resources
        test: test  # default; path mapping can be done here; will be served at /test-resources
  ```

adjust `package.json`
- use `devDependency` `"@ui5/cli": "^5.0.0-alpha.4"`
- run scripts use `-o /test-resources/app/id/from/manifest/flp.html`

## Tests

Two apps with UI5 project type 'component' and one app with UI5 project type 'application' served in parallel by one ui5 cli server instance, e.g.
- my/first/app/flp.html (ui5 project type `component` app 1)
- my/second/app/flp.html (ui5 project type `component` app 2)
- test/flp.html (ui5 project type `application` app)

Apps can have path mappings defined in the ui5 yaml file

Endpoints that need to be tested:
- flp with local UI5 sources (from npmjs)
  - default path
  - custom `flp.path`
  - custom `flp.init`
  - `enhancedHomePage: true`
  - additional `flp.apps`
- flp with proxied UI5 sources
  - default `flp.path`
  - custom `flp.path`
  - custom `flp.init`
  - `enhancedHomePage: true`
  - additional `flp.apps`
- `editors.rta`
  - w/o developer mode (aka variants management)
  - with developer mode (aka cpe/adp)
- `editors.cardGenerator`
  - default path
  - custom `editors.cardGenerator.path`
- test with local UI5 sources (from npmjs)
  - default paths
  - custom `path` for at least one `framework`
  - custom `init` for at least one `framework`
  - custom `pattern` for at least one `framework`
- test with and w/o `cds-plugin-ui5`
- test with virtual endpoints and with existing files (flp sandbox and test glue code)
- test with and w/o `@sap-ux/ui5-middleware-fe-mockserver`
- eject virtual HTML files (`npx --yes @sap-ux/create@latest add html`)


packages/preview-middleware/src/base/test.ts
```diff
@@ -46,12 +69,22 @@ export function mergeTestConfigDefaults(config: TestConfig): CompleteTestConfig
...
+              if (testPathPrefix) {
+                  // For component type: strip /test-resources/<namespace>/ prefix, then prepend namespace
+                  const strippedPath = filePath.replace(`${testPathPrefix}/`, '');
+                  const parts = strippedPath.split('.');
+                  parts.pop();
+                  return `${ns}/test/${parts.join('.')}`;
+              }
+              const path = filePath.split('.');
               path.pop();
               return `${ns}${path.join('.')}`;
```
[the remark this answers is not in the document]
heimwege: That one is pre-existing and not introduced in this PR

packages/preview-middleware/test/fixtures/simple-component/package.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "test-project",
+    "version": "0.0.1",
+    "private": true,
+    "devDependencies": {
+        "@sap-ux/ui5-middleware-fe-mockserver": "2.3.19",
+        "@sap-ux/ui5-proxy-middleware": "1.6.4",
+        "@ui5/cli": "4"
```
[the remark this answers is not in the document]
heimwege: That's a good finding. Will fix.

packages/preview-middleware/test/integration/preview/component.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+
+for (const { version } of UI5Versions) {
+    test.describe(`UI5 version: ${version}`, () => {
+        test.skip(!isNode22Plus, 'Requires Node >=22 for @ui5/cli@5');
+
+        test.beforeAll(async () => {
+            if (!isNode22Plus) {
+                return;
+            }
+            test.setTimeout(TIMEOUT);
```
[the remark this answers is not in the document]
heimwege: the current structure is fine. The test.skip gives visible skip status; the if (!isNode22Plus) return in beforeAll/afterAll prevents any I/O.

packages/preview-middleware/src/base/utils/project.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @param utils middleware utils
+ * @param prefixType the type of prefix ('test-resources' or 'resources')
+ * @returns path prefix or undefined
+ */
+function getComponentPathPrefix(
+    utils: MiddlewareUtils | undefined,
+    prefixType: 'test-resources' | 'resources'
+): string | undefined {
+    if (typeof utils !== 'object') {
```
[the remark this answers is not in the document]
heimwege: `utils` is never `null` it's either present or undefined

packages/preview-middleware/src/base/config.ts
```diff
@@ -183,22 +224,68 @@ function getUI5Libs(manifest: Partial<Manifest>): string {
...
  */
-export function getFlpConfigWithDefaults(config: Partial<FlpConfig> = {}): FlpConfig {
-    const flpConfig = {
-        path: config.path ?? DEFAULT_PATH,
+export function getFlpConfigWithDefaults(config: Partial<FlpConfig> = {}, utils?: MiddlewareUtils): FlpConfig {
+    const sandboxPathPrefix = getTestResourcesPathPrefix(utils);
+    const defaultPath = adjustPathForSandbox(DEFAULT_PATH, sandboxPathPrefix);
+
+    return {
+        path: posix.join(sandboxPathPrefix ?? '/', config.path ?? defaultPath),
```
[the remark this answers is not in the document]
heimwege: That's a valid point. We could follow-up and sanitize user input gracefully. But not sure if it is worth the effort. Here's a list of all endpoints that would need guards:

## Double-Prefix Alignment — Properties Needing Guards

### `packages/preview-middleware/src/base/config.ts`

#### `getFlpConfigWithDefaults`
| Property | Current code | Status |
|---|---|---|
| `flp.path` | `posix.join(sandboxPathPrefix ?? '/', resolvedPath)` | ❌ Needs guard |
| `flp.init` | `posix.join(sandboxPathPrefix ?? '/', config.init)` | ❌ Needs guard |

#### `adjustRtaConfigPaths`
| Property | Current code | Status |
|---|---|---|
| `editors.rta.endpoints[].path` | `posix.join(getTestResourcesPathPrefix(utils) ?? '/', endpoint.path)` | ❌ Needs guard |

#### `adjustCardGeneratorPath`
| Property | Current code | Status |
|---|---|---|
| `editors.cardGenerator.path` | `posix.join(sandboxPathPrefix, adjustPathForSandbox(basePath, sandboxPathPrefix))` | ❌ Needs guard — `adjustPathForSandbox` only strips `/test`, a fully-qualified user path still gets double-prefixed |

---

### `packages/preview-middleware/src/base/test.ts`

#### `mergeTestConfigDefaults`
| Property | Current code | Status |
|---|---|---|
| `test[].path` | `posix.join(testPathPrefix, merged.path)` | ❌ Needs guard |
| `test[].init` | `posix.join(testPathPrefix, merged.init)` | ❌ Needs guard |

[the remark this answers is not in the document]
heimwege: see above

packages/preview-middleware/src/base/utils/project.ts
```diff
@@ -0,0 +1,56 @@
...
+        return undefined;
+    }
+    const project = utils.getProject?.();
+    return project?.getType?.() === 'component' ? posix.join(`/${prefixType}`, project.getNamespace()) : undefined;
```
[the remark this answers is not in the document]
heimwege: using '' will also be wrong because that's as meaningless as a namespace as using 'undefined' 🤷🏻

packages/preview-middleware/src/base/test.ts
```diff
@@ -46,12 +69,22 @@ export function mergeTestConfigDefaults(config: TestConfig): CompleteTestConfig
...
     return resourceList
         ? resourceList.map((file) => {
-              const path = file.getPath().split('.');
+              const filePath = file.getPath();
+              if (testPathPrefix) {
+                  // For component type: strip /test-resources/<namespace>/ prefix, then prepend namespace
+                  const strippedPath = filePath.replace(`${testPathPrefix}/`, '');
+                  const parts = strippedPath.split('.');
+                  parts.pop();
+                  return `${ns}/test/${parts.join('.')}`;
```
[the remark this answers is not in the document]
heimwege: pre-existing

devinea (commented):
Reviewed the PR after the April 28–29 commits. The mixed-array proxy filter bug,  extension-stripping, and test coverage gaps (getWebappTestPath unit tests, Testsuite integration tests) have all been fixed — good work. A few issues remain that should be addressed before merging.

packages/ui5-proxy-middleware/src/base/utils.ts
```diff
@@ -299,6 +303,29 @@ export const filterCompressedHtmlFiles = (_pathname: string, req: IncomingMessag
...
+export function filterExcludeComponentNamespace(
+    utils?: MiddlewareParameters<UI5ProxyConfig>['middlewareUtil']
+): Extract<Filter<IncomingMessage>, Function> | undefined {
+    if (typeof utils !== 'object' || utils.getProject?.()?.getType?.() !== 'component') {
+        return undefined;
+    }
+    const prefixTypes = ['/test-resources', '/resources'] as const;
+    const excludedPaths: string[] = [];
+    for (const prefixType of prefixTypes) {
+        excludedPaths.push(posix.join(`${prefixType}`, utils.getProject().getNamespace()));
```
devinea: **Bug: `getNamespace()` can return `null`/`undefined` here, causing a wrong exclusion path.**

After the guard on line 315 (`getType() !== 'component'`), the code calls `utils.getProject().getNamespace()` without null-safety. `getNamespace()` is typed to return `string | null | undefined`, and a component project with no namespace declared in `ui5.yaml` will return `null` or `undefined`. `posix.join('/resources', null)` produces `'/resources/null'` (string coercion) — silently creating the wrong exclusion path so the filter never matches.

Suggest:
```ts
const ns = utils.getProject().getNamespace();
if (!ns) {
    return undefined; // no namespace declared — nothing to exclude
}
const excludedPaths = prefixTypes.map((p) => posix.join(p, ns));
```

packages/adp-tooling/src/base/helper.ts
```diff
@@ -148,10 +150,18 @@ export function readManifestFromBuildPath(cfBuildPath: string): Manifest {
...
  */
-export async function loadAppVariant(rootProject: ReaderCollection): Promise<DescriptorVariant> {
-    const appVariant = await rootProject.byPath('/manifest.appdescr_variant');
+export async function loadAppVariant(
+    rootProject: ReaderCollection,
+    utils: MiddlewareUtils
+): Promise<DescriptorVariant> {
+    const pathPrefix =
+        utils.getProject?.()?.getType?.() === 'component'
+            ? posix.join('/resources', utils.getProject().getNamespace())
```
devinea: **Same `getNamespace()` null-safety gap as in `filterExcludeComponentNamespace`.**

The type guard on line 161 confirms `getType() === 'component'` but then calls `utils.getProject().getNamespace()` without a null check. If `getNamespace()` returns `null`/`undefined` (valid for a component project with no namespace in `ui5.yaml`), `posix.join('/resources', null)` produces `'/resources/null'` — and the `byPath` lookup for `manifest.appdescr_variant` will silently fail.

```ts
const ns = utils.getProject?.()?.getNamespace?.();
const pathPrefix = utils.getProject?.()?.getType?.() === 'component' && ns
    ? posix.join('/resources', ns)
    : '';
```

packages/project-access/src/project/ui5-config.ts
```diff
@@ -54,22 +55,63 @@ async function getBaseDir(appRoot: string, memFs?: Editor): Promise<string> {
  * @returns - path to webapp folder
  */
 export async function getWebappPath(appRoot: string, memFs?: Editor): Promise<string> {
+    //Shortcut: if webapp/manifest.json exists, return webapp path w/o reading the YAML content
+    if (await fileExists(join(appRoot, DirName.Webapp, FileName.Manifest), memFs)) {
```
devinea: **Shortcut bypasses YAML path mappings for custom-`webapp` projects.**

If a project has a custom `webapp` path in `ui5.yaml` (e.g. `webapp: src/webapp`) *and* also has a legacy/stale `webapp/manifest.json` artifact on disk, this early return gives the wrong path without ever reading the YAML. The YAML is the authoritative source for path mappings; the filesystem existence check should be a fallback for when no YAML is present, not an override for when it is.

Consider only skipping YAML parsing when `ui5.yaml` doesn't exist at all, rather than whenever `webapp/manifest.json` is found on disk.

packages/project-access/src/project/ui5-config.ts
```diff
@@ -54,22 +55,63 @@ async function getBaseDir(appRoot: string, memFs?: Editor): Promise<string> {
...
+ *
+ * @param appRoot - root to the application
+ * @param [memFs] - optional mem-fs editor instance
+ * @returns - path to test folder
+ * @throws {Error} if ui5.yaml or 'type' cannot be read
+ * @throws {Error} if project type is not 'application', 'component', 'library', 'theme-library' or 'module'
+ */
+export async function getWebappTestPath(appRoot: string, memFs?: Editor): Promise<string> {
+    //Shortcut: if webapp/manifest.json exists, return webapp/test path w/o reading the YAML content
+    if (await fileExists(join(appRoot, DirName.Webapp, FileName.Manifest), memFs)) {
```
devinea: Same concern as the `getWebappPath` shortcut above — `getWebappTestPath` has the identical early return that skips YAML-authoritative custom `webapp` path mappings when `webapp/manifest.json` exists on disk. Both shortcuts should be addressed together.

packages/preview-middleware/src/base/config.ts
```diff
@@ -437,31 +517,49 @@ export function createFlpTemplateConfig(
...
  *
  * @param config configuration from the ui5.yaml
+ * @param utils middleware utils
  * @param logger logger instance
  * @returns an array of preview paths
  */
-export function getPreviewPaths(config: MiddlewareConfig, logger: ToolsLogger = new ToolsLogger()): PreviewUrls[] {
+export function getPreviewPaths(
+    config: MiddlewareConfig,
+    utils: MiddlewareUtils,
```
devinea: **Breaking change shipped as `minor`.**

Adding a required `utils: MiddlewareUtils` as the second parameter is a breaking change for external callers. The previous signature was `getPreviewPaths(config, logger?)` — any code passing a logger as the second argument now silently passes it as `utils`. TypeScript will catch this at compile time, but it means the follow-up update to `@sap/ux-ui5-tooling` must ship before or alongside this version. The changeset should be `major`, or `utils` should be made optional with a fallback (e.g. `utils?: MiddlewareUtils`) so existing callers continue to work.

packages/adp-tooling/src/base/helper.ts
```diff
@@ -148,10 +150,18 @@ export function readManifestFromBuildPath(cfBuildPath: string): Manifest {
...
  *
  * @param {ReaderCollection} rootProject - The root project.
+ * @param {MiddlewareUtils} utils - The middleware utils.
  * @returns {Promise<DescriptorVariant>} The parsed descriptor variant.
  */
-export async function loadAppVariant(rootProject: ReaderCollection): Promise<DescriptorVariant> {
-    const appVariant = await rootProject.byPath('/manifest.appdescr_variant');
+export async function loadAppVariant(
+    rootProject: ReaderCollection,
+    utils: MiddlewareUtils
```
devinea: **Breaking change: `utils` is now a required second parameter.**

Any existing caller of `loadAppVariant(rootProject)` will fail to compile after this change. Since `loadAppVariant` is part of the `@sap-ux/adp-tooling` public export surface, this warrants a `major` version bump rather than `minor`. Consider making `utils` optional (`utils?: MiddlewareUtils`) and applying the component path prefix only when `utils` is provided, so existing callers keep working without changes.

packages/app-config-writer/src/preview-config/preview-files.ts
```diff
@@ -22,7 +22,9 @@ const renameMessage = (filePath: string): string =>
  * @param logger logger to report info to the user
  */
 export async function renameSandbox(fs: Editor, basePath: string, path: string, logger?: ToolsLogger): Promise<void> {
-    const filePath = join(await getWebappPath(basePath), path);
+    const webappTestPath = await getWebappTestPath(basePath);
+    const fileName = basename(path);
```
devinea: **Behaviour change: directory component of `path` is silently dropped.**

The old code preserved the full `path` relative to the webapp root:
```ts
// old
const filePath = join(await getWebappPath(basePath), path);
// path='test/flpSandbox.html' → <webapp>/test/flpSandbox.html
```

The new code strips the directory and joins only the filename to `webappTestPath`:
```ts
// new
const fileName = basename(path);          // 'flpSandbox.html'
const filePath = join(webappTestPath, fileName); // <webapp>/test/flpSandbox.html
```

For current callers (paths like `test/flpSandbox.html`) the final resolved path is identical because `webappTestPath` already includes the `test` segment. But the parameter contract has silently changed: a caller passing `test/subdir/file.html` previously resolved `<webapp>/test/subdir/file.html` and now resolves `<webapp>/test/file.html`. If this is intentional, rename the parameter from `path` to `fileName` and update the JSDoc so callers know not to include directory segments.

heimwege: fixed in [e4a46b2](https://github.com/SAP/open-ux-tools/pull/3927/commits/e4a46b2fe33da1866b85f555f626d419c75066a1)

packages/project-access/src/project/ui5-config.ts
heimwege: fixed in [393271b](https://github.com/SAP/open-ux-tools/pull/3927/commits/393271b52a84058443710dc50f6ffd39e6c8d127)

packages/project-access/src/project/ui5-config.ts
heimwege: fixed in [393271b](https://github.com/SAP/open-ux-tools/pull/3927/commits/393271b52a84058443710dc50f6ffd39e6c8d127)

packages/adp-tooling/src/base/helper.ts
heimwege: fixed in [393271b](https://github.com/SAP/open-ux-tools/pull/3927/commits/393271b52a84058443710dc50f6ffd39e6c8d127)

packages/ui5-proxy-middleware/src/base/utils.ts
heimwege: fixed in [393271b](https://github.com/SAP/open-ux-tools/pull/3927/commits/393271b52a84058443710dc50f6ffd39e6c8d127)

packages/adp-tooling/src/base/helper.ts
heimwege: fixed in [76adce6](https://github.com/SAP/open-ux-tools/pull/3927/commits/76adce61ce030aeeb4daee27a88c27f12f18f6c9)

packages/preview-middleware/src/base/config.ts
heimwege: fixed in [76adce6](https://github.com/SAP/open-ux-tools/pull/3927/commits/76adce61ce030aeeb4daee27a88c27f12f18f6c9)

Klaus-Keller (dismissed):
Thanks @heimwege, 

* changeset exists
* the pointsAI review found have been addressed
* test coverage is still good
* did no manual test

Approved from my side

devinea (dismissed):
Thanks for the addressing review comments.

heimwege:
💡Now that we are on node 22 the node 20 check in the integration test run is actually no longer needed. But this can also be removed at a later point in time.

packages/adp-tooling/src/translations/adp-tooling.i18n.json
```diff
@@ -130,7 +130,7 @@
         "failedToListBtpDestinations": "Failed to list BTP destinations. Error: {{error}}",
         "destinationServiceNotFoundInMtaYaml": "Destination service instance not found in mta.yaml. Ensure a resource with 'service: destination' is declared.",
         "noServiceKeysFoundForDestination": "No service keys found for destination service instance '{{serviceInstanceName}}'. Ensure the service is provisioned and try again.",
-        "errorFetchingDestinations": "An error occurred when fetching destinations. For more information, see the logs.",
+        "errorFetchingDestinations": "Error fetching destinations. Check log for details.",
```
lfindlaysap:
```suggestion
        "errorFetchingDestinations": "An error occurred when fetching destinations. For more information, see the logs.",
```

lfindlaysap (dismissed):

packages/preview-middleware/src/base/config.ts
```diff
@@ -513,14 +633,15 @@
                 manifest['sap.app'].id,
                 flpTemplConfig.ui5.theme
             );
-            fs.write(join(webappPath, testConfig.path), render(testTemplate, testTemplateConfig));
+            fs.write(join(basePath, testConfig.path), render(testTemplate, testTemplateConfig));
```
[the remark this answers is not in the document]
heimwege: This runs on localhost only

packages/adp-tooling/src/translations/adp-tooling.i18n.json
heimwege: This change is not part of the PR and was committed by mistake

lfindlaysap (commented):
@heimwege, which files should I review?

heimwege:
> @heimwege, which files should I review?

I don't think there's much for you to do in this PR. Maybe the addition to the readme and the related log message at
- https://github.com/SAP/open-ux-tools/pull/3927/changes#diff-8484083c49623d4cb6fdc16358d5587626d9858014ad6adc4ea43c6306ff5f2eR335 and
- https://github.com/SAP/open-ux-tools/pull/3927/changes#diff-8c4075ebf48314c660c760fc76df615ad98de808ccf78d130e6bf64f80930311R13
- and the readme changes in https://github.com/SAP/open-ux-tools/pull/3927/changes#diff-09028281752ce87873e7f199e7e0894d8c0c9188c4808b5d8b3972c41856975eR158

That's pretty much it.

packages/preview-middleware/src/base/config.ts
```diff
@@ -401,37 +495,34 @@ export function remapResourcesForPath(config: TemplateConfig, newPagePath: strin
...
+    const appBasePath = getResourcesPathPrefix(utils) ?? rootBasePath;
     return {
-        basePath: basePath,
+        appBasePath: appBasePath,
+        rootBasePath: rootBasePath,
         baseUrl: '',
         apps: {},
-        init: initPath,
+        init: config.init,
         ui5: {
```
nikmace, line 577: **Question**: In `createFlpTemplateConfig`, the old code prepended the app namespace to `config.init` (e.g., `my/app/variant/custom-init`). The new code passes `config.init` raw. Was removing the namespace prefix intentional? If a project configures a relative init value, `data-open-ux-preview-custom-init` will receive an unqualified path, and the init script will silently fail to load.


_What the old code did_
```ts
const id = manifest['sap.app']?.id ?? '';
const ns = id.replace(/\./g, '/');   // e.g. "my.app.variant" → "my/app/variant"

let initPath: string | undefined;
if (config.init) {
    const separator = config.init.startsWith('/') ? '' : '/';
    initPath = `${ns}${separator}${config.init}`;
    // result: "my/app/variant/custom-init" or "my/app/variant/custom-init.js"
}
```
So if you had in your ui5.yaml:
```yaml
init: custom-init
```
The template received init: "my/app/variant/custom-init" — already a fully-qualified module path that UI5's data-open-ux-preview-custom-init attribute knows how to load.

_What the new code does_
```yaml
init: config.init,   // passed raw — "custom-init" with no namespace prefix
```

heimwege: That's a very good finding, sir. I reverted back to the logic from main with some improvements + explanation. e806afff2a83dcbad4a15d228bebf93e91feab9d
(I thought I can do all path adjustments in the constructor -> `getFlpConfigWithDefaults` but seems to be wrong in that specific case because we need the namespace for the resource roots mapping instead of the path)

nikmace (dismissed):
Changes look good
Coverage is good
Changeset is OK
Did not test manually

- _I don't remember if we ran UI Integration tests against this branch, but would be nice to do so nonetheless_
- _One Low Sonar issue still exists_

PR #3938, opened 2025-12-17
fix(@sap-ux/fiori-mcp-server): log to file, fix lint issues

Klaus-Keller:
Fix for issues caused by console error outputs. In some cases the MCP server crashes.

E.g., when generating a SAP Fiori application in a CAP project, the MCP server crashed with errors like:

image

I've also fixed some linting issues.

### DIFF
--- a/packages/fiori-mcp-server/src/server.ts
+++ b/packages/fiori-mcp-server/src/server.ts
@@ -93,10 +93,10 @@ export class FioriFunctionalityServer {
             try {
                 let result;
                 TelemetryHelper.markToolStartTime();
-                const telemetryProperties: TelemetryData = {
-                    tool: name,
-                    functionalityId: (args as { functionalityId: string })?.functionalityId
-                };
+                const telemetryProperties: TelemetryData = { tool: name };
+                if ('functionalityId' in args) {
+                    telemetryProperties.functionalityId = args.functionalityId as string;
+                }
                 logger.debug(`Executing tool: ${name} with arguments: ${JSON.stringify(args)}`);
 
                 switch (name) {


packages/fiori-mcp-server/src/page-editor-api/sapuxFtfsFileIO.ts
```diff
@@ -180,13 +181,15 @@ export class SapuxFtfsFileIO {
         const exportParams = {
             [schemaType]: {
                 appId: this.getAppId(manifest),
-                jsonSchema: JSON.parse(pageData.schema),
+                jsonSchema: JSON.parse(pageData.schema) as object,
```
heimwege: We could use `import { JSONSchema7 } from '@types/json-schema';`

packages/fiori-mcp-server/src/page-editor-api/sapuxFtfsFileIO.ts
```diff
@@ -223,7 +226,8 @@ export class SapuxFtfsFileIO {
             [SchemaType.Application]: {
                 application: config as v4.ApplicationV4,
                 manifest,
-                jsonSchema: JSON.parse(schema)
+                jsonSchema: JSON.parse(schema),
```
heimwege: In line 184 we parse `as object` and here we don't. Should be the same in all places

packages/fiori-mcp-server/src/utils/logger.ts
```diff
@@ -1,19 +1,30 @@
...
+import { ToolsLogger, LogLevel, NullTransport, FileTransport } from '@sap-ux/logger';
+import { fioriToolsDirectory } from '@sap-ux/project-access';
+import { join } from 'node:path';
+
+/**
+ * General logger settings
+ */
+const filename = join(fioriToolsDirectory, 'fiori-mcp-server.log');
+const maxsize = 10 * 1024 * 1024; // 10 MB max per file
+const maxFiles = 5; // Keep up to 5 log files
```
heimwege: We use upper case for constant names throughout the mono repo

```suggestion
const FILE_NAME = join(fioriToolsDirectory, 'fiori-mcp-server.log');
const MAX_FILESIZE = 10 * 1024 * 1024; // 10 MB max per file
const MAX_FILES = 5; // Keep up to 5 log files
```

packages/fiori-mcp-server/src/utils/logger.ts
```diff
@@ -39,7 +50,34 @@ const currentLogLevel = getLogLevel();
...
+    logLevel: currentLogLevel ?? LogLevel.Error,
+    transports:
+        currentLogLevel === undefined ? [new NullTransport()] : [new FileTransport({ filename, maxsize, maxFiles })],
     logPrefix: 'fiori-mcp'
 });
+
+/**
+ * Specification module ("@sap/ux-specification") expects different logger interface, so we create a simple adapter here
+ */
+type MessageMetadataType = (object | string | number)[];
```
heimwege: Please move type definitions to the very top (below the imports)

packages/fiori-mcp-server/src/server.ts
```diff
@@ -95,8 +95,9 @@ export class FioriFunctionalityServer {
                 TelemetryHelper.markToolStartTime();
                 const telemetryProperties: TelemetryData = {
                     tool: name,
-                    functionalityId: (args as any)?.functionalityId
+                    functionalityId: (args as { functionalityId: string })?.functionalityId
```
heimwege: This cast is kind of odd and actually does the same as `any`. It doesn't even take into account that `functionalityId` is optional


```suggestion
                    functionalityId: 'functionalityId' in args ? String(args.functionalityId) : ''
```

Klaus-Keller: This is a bit tricky, `functionalityId` has type `string | (string | number)[]`, but `TelemetryData` has `[key: string]: string;`. I've changed this before to stringify the array, but this means a change in telemetry reporting (I've checked this in telemetry dashboard), so I've decided against this change. 

I've changed it now to:
```typescript
const telemetryProperties: TelemetryData = { tool: name };
if ('functionalityId' in args) {
  telemetryProperties.functionalityId = args.functionalityId as string;
}
```
[#a3a92bf186396bd8625e38c8cc03c7fe6a5b6a03](https://github.com/SAP/open-ux-tools/pull/3938/commits/a3a92bf186396bd8625e38c8cc03c7fe6a5b6a03)

packages/fiori-mcp-server/src/utils/logger.ts
Klaus-Keller: done [#5c87cb39e](https://github.com/SAP/open-ux-tools/pull/3938/commits/5c87cb39e2fba45b72ebfecc2d6249916e5dc234)

packages/fiori-mcp-server/src/utils/logger.ts
Klaus-Keller: > throughout the mono repo 

is a bold statement, as far as I know, we do not have guidelines and as consequence we have camelCase, e.g.: https://github.com/SAP/open-ux-tools/blob/faa1007faffa6e26c8b65de39ce9c05eb35d173d/packages/ui5-info/src/constants.ts#L24-L29

PascalCase, e.g.:
https://github.com/SAP/open-ux-tools/blob/faa1007faffa6e26c8b65de39ce9c05eb35d173d/packages/cf-deploy-config-writer/src/constants.ts#L3-L10

and UPPER_CASE, e.g.:
https://github.com/SAP/open-ux-tools/blob/faa1007faffa6e26c8b65de39ce9c05eb35d173d/packages/ui5-proxy-middleware/src/base/constants.ts#L1-L6

In my case the constants match exactly to the required transport options:
https://github.com/SAP/open-ux-tools/blob/e76c56765f1ae6c6f7a8580d7c032d48a0a9a7ff/packages/fiori-mcp-server/src/utils/logger.ts#L55

However, to avoid breaking any rules I've added the settings where they are used, they are note re-used anyway

[#79bc50572 ](https://github.com/SAP/open-ux-tools/pull/3938/commits/79bc505728b6fbe96257b9e065a0450fc5b615f2)

packages/fiori-mcp-server/src/page-editor-api/sapuxFtfsFileIO.ts
Klaus-Keller: done https://github.com/SAP/open-ux-tools/pull/3938/commits/3bba06a354fbb48bfb04761870d14c4df2fa0b00

packages/fiori-mcp-server/src/page-editor-api/sapuxFtfsFileIO.ts
Klaus-Keller: done https://github.com/SAP/open-ux-tools/pull/3938/commits/3bba06a354fbb48bfb04761870d14c4df2fa0b00

815are (approved):
checked:
1. mcp server changes - changes are straightforward
2. tests maintained
3. changelog persists

tested with promptfoo tests

PR #3939, opened 2025-12-17
chore: fix various sonar issues

heimwege:
Fix ~200 medium Sonar issues

Integration tests were green in https://github.com/SAP/open-ux-tools/actions/runs/20350557435


packages/control-property-editor/src/panels/changes/ChangeStack.tsx
```diff
@@ -44,10 +44,11 @@ export interface ChangeStackProps {
...
     let groups = convertChanges(changes);
-    const filterQuery = useSelector<RootState, FilterOptions[]>((state) => state.filterQuery)
-        .filter((item) => item.name === FilterName.changeSummaryFilterQuery)[0]
-        .value.toString()
-        .toLowerCase();
+    const filterQuery =
+        useSelector<RootState, FilterOptions[]>((state) => state.filterQuery)
+            .find((item) => item.name === FilterName.changeSummaryFilterQuery)
+            ?.value.toString()
+            .toLowerCase() ?? '';
```
devinea, line 51: `?? ''` seems like a small change on how it worked before. Probably fine.

heimwege: Yeah, actually the former code was not quite correct. Blindly accessing the first entry of an empty array can also lead to an error. `.find` returns type of the array entry (here: string) | undefined. So the fallback to empty string was needed to deal with `undefined`. If there's a better fallback value we should adjust this code.

longieirl (approved):
Changesets are good.
Changes look good.
Didnt test locally.

PR #3940, opened 2025-12-18
fix(ui-components): tooltip does not appear in VSCode+MacOS for icons buttons

815are:
We noticed issue with latest VSCode running on MacOS, that tooltips for `IconButton` does not appear.
As checked with @gpeuc - we apply workaround with `pointer-events: none` as it is most simplest and stable for now.


devinea (approved):
Changes are clear.
Changeset ✅ 
Didn't test locally.

broksy (approved):

PR #3943, opened 2025-12-18
fix(odata service inquirer): 34666 update telemetry test

korotkovao:
Related to internal issue #34666

Update the Telemetry test to allow 1ms tolerance in the timing.


korotkovao:
Thank you @devinea @cianmSAP

PR #3953, opened 2025-12-19
Update README with migration guide for ESLint 9

devinea:
Add migration instructions for eslint-plugin-fiori-tools


Klaus-Keller (approved):
Thanks @devinea, updated documentation looks good.

PR #3956, opened 2025-12-23
Fix/preview middleware/custom init

### LINKED ISSUE #3955: BUG - customInit not working anymore
### Description

Hi,

**customInit** in preview-middleware is not called after upgrading to version 0.23.79.

 Debugging shows, that there is a typo with [openUxPreviewCustomInit](https://github.com/SAP/open-ux-tools/pull/3932/files#diff-8cf74cec1061004e608c1a0acc8fee55f8fd7555aed2717a69d368b9498245d4R417).  It should be **openUxPreviewCustominit** with a lowercase **i**.

Image

According to the [name conversion](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset?utm_source=chatgpt.com#name_conversion) for the dataset property, this seems to be the expected behaviour.

OS:                      
- [x] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

heimwege:
fixes #3955

### DIFF
--- a/packages/preview-middleware/src/base/config.ts
+++ b/packages/preview-middleware/src/base/config.ts
@@ -376,7 +376,11 @@ export function createFlpTemplateConfig(
     const id = manifest['sap.app']?.id ?? '';
     const ns = id.replace(/\./g, '/');
     const basePath = posix.relative(posix.dirname(config.path), '/') ?? '.';
-    const initPath = config.init ? `${ns}${config.init.startsWith('/') ? '' : '/'}${config.init}` : undefined;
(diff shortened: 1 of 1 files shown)


815are (approved):
checked:
1. change is straightforward
2. changelog persists

I did not test locally

heimwege:
Merging as `admin` because  no `SAP/ux-tools-middlewares-and-tasks` reviewer present

Issue #2470, opened 2024-10-17
BUG - eslint-plugin-fiori-tools meets eslint 8 EOL


heimwege:
### Description
Not actually a bug but none of the other categories either 🤷🏻 

Seems like eslint 8.x is [EOL](https://eslint.org/version-support/) but eslint-plugin-fiori-tools does not yet support 9.x

### Steps to Reproduce
1. try to have 9.x eslint version using eslint-plugin-fiori-tools

### Expected results
eslint-plugin-fiori-tools supports eslint 9.x

### Actual results
Was not able to migrate to esling 9.x

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [X] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

heimwege:
@devinea are there any plans to adjust eslint-plugin-fiori-tools to the new [eslint 9 config](eslint-plugin-fiori-tools)?

devinea:
@heimwege not so far. But it seems like we will need to consider it soon. 
FYI @ianquigley-sap

Kyrodan:
SAP pushes TypeScript. I've just created a fresh new app using latest Application Generator.
Starting lint-Task it writes:
```
> project1@0.0.1 lint
> eslint ./

=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=4.7.4 <5.6.0

YOUR TYPESCRIPT VERSION: 5.8.2

Please only submit bug reports when using the officially supported version.

=============
```

Please upgrade to ESLint 9 quickly! It's very important in professional development to be able to safely use the linter!

devinea:
Actively working on this.

ClementOrdina:
Hey @devinea, could you use a hand on this? Happy to help if there’s anything we can do, just let me know what's left or where I could start.

ClementRoyer:
Hey @devinea, could you use a hand on this? Happy to help if there’s anything we can do, just let me know what's left or where I could start.

devinea:
Thanks - @ClementRoyer 
[...]

PR #3957, opened 2025-12-23
refactor: update ESLint configuration to remove defineConfig usage

devinea:
defineConfig is not required when using eslint 9 flat config
https://github.com/SAP/open-ux-tools/issues/2470


marufrasully (approved):
Thanks @devinea 
- changes looks good
- did not test it locally

PR #3961, opened 2025-12-30
feat(ui-component): add StarBig icon

Klaus-Keller:
Add start icon with 24x24 as `StartBig`:

image


815are (commented):
@Klaus-Keller 

is it intended to have without changelog?

Issue #1716, opened 2024-03-01
TBI - update `mem-fs` and `mem-fs-editor`


voicis:
### Description (include screenshots)
We're currently using `mem-fs-editor` version `9.4.0` which is 2 years old now and the latest version is `11.0.0`. We should update it to the latest version.

### Value
Maintenance and upkeep.

### Tasks
- [ ] update `mem-fs` and `mem-fs-editor` version
- [ ] remove @types/mem-fs* modules
- [ ] update imports of type Editor it is now MemFsEditor
- [ ] manually checking the API changes - on first glance it looks like there are a few e.g. read returns string | null instead of just string and copyTpl has different input parameters.

PR #3973, opened 2026-01-07
chore: version bump mem-fs

heimwege:


cianmSAP:
Unit tests are failing with this update due to mem-fs going to ESM with version 3 ([see here](https://github.com/SBoudrias/mem-fs/releases/tag/v3.0.0))

For example, running the tests as the branch is now (e.g. from `abap-deploy-config-writer`), it throws the error: 

```
 import { EventEmitter } from 'events';
    ^^^^^^

    SyntaxError: Cannot use import statement outside a module
```
—— 

I made some config udpates based on the [jest docs ](https://jestjs.io/docs/ecmascript-modules), specifically “configure your transformer to emit ESM”. I updated the jest.config to: 

```
module.exports = {
    ...config,
    transform: {
        '^.+\\.ts$': ['ts-jest', { useESM: true }]
    }
};
```

and ran the test using 
```
node --experimental-vm-modules node_modules/jest/bin/jest.js
```

It returns the error **Must use import to load ES Module**

I also added `extensionsToTreatAsEsm: ['.ts']` as a test but it shows error **ReferenceError: exports is not defined**, implies that the test environment is now ESM while some dependencies are still CommonJS.

——

I tried dynamic imports e.g. 
```
let createStorage: any;
let create: any;

beforeAll(async () => {
    const { create: createStorageMemFs } = await import('mem-fs');
    createStorage = createStorageMemFs;
    const { create: createMemFsEditor } = await import('mem-fs-editor');
    create = createMemFsEditor;
});
```
But I’m still hit with the error **Must use import to load ES Module**

I tried a few other configuration combinations as well, but without success.

heimwege:
cross link to https://github.com/SAP/open-ux-tools/issues/1716

PR #3977, opened 2026-01-08
Fix(eslint-plugin-fiori-tools): do not ignore mockserver extensions

heimwege:
Currently `webapp/localService/**` is being ignored completely (including custom mockserver extension files). This PR adjust the ignore to not include `.js` and `.ts` files as they might be custom mockserver extensions.

Also `webapp` should not be a static name but read from the path mappings of the ui5 yaml file

### DIFF
--- a/.changeset/four-colts-fold.md
+++ b/.changeset/four-colts-fold.md
@@ -1,7 +1,0 @@
----
-'sap-ux-sap-systems-ext': minor
-'@sap-ux/store': minor
-'@sap-ux/sap-systems-ext-types': patch
----
-
-update backend systems with system info from adt api
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.151
+
+### Patch Changes
+
+- @sap-ux/system-access@0.6.41
+- @sap-ux/fiori-elements-writer@2.8.39
+- @sap-ux/fiori-freestyle-writer@2.5.28
+
 ## 1.1.150
 
 ### Patch Changes
--- a/examples/simple-generator/package.json
+++ b/examples/simple-generator/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/generator-simple-fe",
-    "version": "1.1.150",
+    "version": "1.1.151",
     "description": "Simple example of a yeoman generator for Fiori elements.",
     "license": "Apache-2.0",
     "private": true,
--- a/packages/abap-deploy-config-inquirer/CHANGELOG.md
+++ b/packages/abap-deploy-config-inquirer/CHANGELOG.md
@@ -1,5 +1,16 @@
 # @sap-ux/abap-deploy-config-inquirer
 
+## 1.6.77
+
+### Patch Changes
+
+- Updated dependencies [c9fd939]
+    - @sap-ux/store@1.5.0
+    - @sap-ux/fiori-generator-shared@0.13.53
+    - @sap-ux/system-access@0.6.41
+    - @sap-ux/ui5-config@0.29.12
+    - @sap-ux/inquirer-common@0.10.12
+
 ## 1.6.76
 
 ### Patch Changes
--- a/packages/abap-deploy-config-inquirer/package.json
+++ b/packages/abap-deploy-config-inquirer/package.json
@@ -6,7 +6,7 @@
         "url": "https://github.com/SAP/open-ux-tools.git",
         "directory": "packages/abap-deploy-config-inquirer"
     },
-    "version": "1.6.76",
+    "version": "1.6.77",
     "license": "Apache-2.0",
     "main": "dist/index.js",
     "scripts": {
--- a/packages/abap-deploy-config-sub-generator/CHANGELOG.md
+++ b/packages/abap-deploy-config-sub-generator/CHANGELOG.md
@@ -1,5 +1,15 @@
 # @sap-ux/abap-deploy-config-sub-generator
 
+## 0.1.126
+
+### Patch Changes
+
+- @sap-ux/abap-deploy-config-inquirer@1.6.77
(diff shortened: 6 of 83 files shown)


devinea (approved):
Thanks @heimwege 
Changes lgtm, I haven't tested locally.

packages/eslint-plugin-fiori-tools/src/index.ts
```diff
@@ -69,14 +71,34 @@ const commonConfig: Linter.Config[] = [
...
 
+// Use synckit to create sync function for project-access getPathMappingsSync
+const workerPath = join(__dirname, 'worker-getPathMappingsSync.js');
+const getPathMappingsSync = createSyncFn<typeof getPathMappings>(workerPath);
+
+const pathMappingsAbsolute = getPathMappingsSync(process.cwd());
+const webappPathAbsolute =
+    'webapp' in pathMappingsAbsolute
+        ? pathMappingsAbsolute.webapp
+        : (pathMappingsAbsolute.src ?? join(process.cwd(), 'webapp'));
```
marufrasully, line 83: process.cwd() can be shaky. User can start process from any folder - not necessary from command defined in script of package.json file. This is rare but can happen :) 
Can we use https://github.com/SAP/open-ux-tools/pull/3977/changes#diff-ee4c07396f27100ee31e96e51d26bb15153a5a97623f425ce05557e0d09a8c7cR21 approach?

packages/eslint-plugin-fiori-tools/src/index.ts
```diff
@@ -69,14 +71,34 @@ const commonConfig: Linter.Config[] = [
...
+const webappPathRelative = relative(process.cwd(), webappPathAbsolute).split(sep).join('/');
+const testPathRelative =
+    'webapp' in pathMappingsAbsolute
+        ? posix.join(webappPathRelative, 'test')
+        : relative(process.cwd(), pathMappingsAbsolute.test ?? join(process.cwd(), 'webapp/test'))
+              .split(sep)
+              .join('/');
```
marufrasully: here we try to unified path? can we re-use this method https://github.com/SAP/open-ux-tools/blob/main/packages/eslint-plugin-fiori-tools/src/project-context/utils.ts#L9 method also enhance it with `.split(sep).join('/');`

packages/eslint-plugin-fiori-tools/src/index.ts
```diff
@@ -69,14 +71,34 @@ const commonConfig: Linter.Config[] = [
...
     {
-        files: ['./webapp/**/*.js', './webapp/**/*.ts'],
+        files: [`./${webappPathRelative}/**/*.js`, `./${webappPathRelative}/**/*.ts`],
 
         ignores: [
             'target/**',
-            'webapp/test/**',
-            'webapp/localservice/**',
+            `${testPathRelative}/**`,
+            `${posix.join(webappPathRelative, 'localservice')}/**`, // Ignore everything in the 'localservice' folder
```
marufrasully, line 97: duplicate lines

packages/eslint-plugin-fiori-tools/src/index.ts
```diff
@@ -166,14 +188,16 @@ const testConfig: Linter.Config[] = [
...
 
         ignores: [
             'target/**',
-            'webapp/test/changes_loader.ts',
-            'webapp/test/changes_preview.ts',
-            'webapp/localservice/**',
-            'webapp/localService/**',
+            `${testPathRelative}/changes_loader.ts`,
+            `${testPathRelative}/changes_preview.ts`,
+            `${posix.join(webappPathRelative, 'localservice')}/**`, // Ignore everything in the 'localservice' folder
```
marufrasully, line 196: here seems repeated 

```
            `${posix.join(webappPathRelative, 'localService')}/**`, // Ignore everything in the 'localService' folder
            `!${posix.join(webappPathRelative, 'localService')}/**/*.{ts,js}`, // EXCEPT for .ts and .js files (that might be custom mockserver extensions)
```

packages/eslint-plugin-fiori-tools/src/index.ts
heimwege: @marufrasully that link seems to be wrong it points to this PR 🤷🏻 (the old plugin version used `process.cwd()` I copied from there)

packages/eslint-plugin-fiori-tools/src/index.ts
heimwege: nope `localservice` vs `localService` 👀

packages/eslint-plugin-fiori-tools/src/index.ts
heimwege: nope localservice vs localService 👀

packages/eslint-plugin-fiori-tools/src/index.ts
heimwege: fixed in e49fa4bf668d26d575016dc5c7fcb2d1f10be7f2

packages/eslint-plugin-fiori-tools/src/index.ts
marufrasully: this is what I meant https://github.com/SAP/open-ux-tools/blob/6ac87413a9bee646d40bb4fc3dabf6d248d44708/packages/eslint-plugin-fiori-tools/src/index.ts#L19

marufrasully: Actually we need can use this - https://github.com/SAP/ui5-language-assistant/blob/cb20329170cc28c1c467a2856b12d7cfc58d1812/packages/context/src/utils/project.ts#L40 this can be done later on.

marufrasully (approved):
Thanks @heimwege 
- changes looks good
- covered with unit tests
- review comment addressed
- did not test it locally

heimwege:
@devinea seems like the worker (synckit running those in a separate thread) tests are not correctly assigned to the test coverage. There are test in this PR but Sonar lists the worker file with 0% coverage. Not sure if this is a problem 🤷🏻

PR #3980, opened 2026-01-08
fix: add repository field to package.json files for multiple packages

devinea:
fixes links from npmjs.com e.g. for rules on https://www.npmjs.com/package/@sap-ux/eslint-plugin-fiori-tools


marufrasully (dismissed):
@devinea thanks. Changes looks good.

PR #3982, opened 2026-01-09
fix(abap-inquirer): add precheck for package validation based on answers

cianmSAP:
**Internal bug 34576**

Re-validation on the pacakge(s) prompts can cause unnecessary prompt loading - due to the requirement to fetch transport requests for a given abap target + package + ui5 abap repo name

The description answer has no bearing on the transports returned and therefore should be ignored i.e re-validation of the package(s) prompts should not occur when only the description is updated

To achieve this, we can check before validating the package, and compare the previous answer object to the current one and determine if only the value of `description` has been updated


IainSAP (approved):
Thanks @cianmSAP 

- Code makes sense to prevent re-evaluations
- Tests added to cover

PR #3983, opened 2026-01-09
[WIP] feat: enhance catalog service error handling and add response types

unseen1980:
When the OData V4 catalog returned HTTP 403 (Forbidden), the Application Generator or Service Center only shows a generic "no services available" message, so users are confused about why services were missing even if the is an issue with permissions.

- Added a listServicesWithStatus() method to @sap-ux/axios-extension that returns error information including HTTP status codes and then updated odata-service-inquirer to use this and display specific warning messages 

The "no services available" message appeared because the original code in getServiceChoices() used catalog.listServices() which throws an exception on HTTP errors like 403. The catch block would log the error to console but then just return an empty service list:

```
// Original code in service-helper.ts
try {
    services = await catalog.listServices();
} catch (error) {
    // Error was logged but swallowed
    requestErrors[catalog.getVersion()] = error;
}
```
This returns empty array and user sees a generic message "There are no services available for the selected system. For more information, see the logs."

The 403 error was caught and swallowed - it wasn't propagated to the UI
The empty service list triggered the generic "no services" message
The actual HTTP 403 status code was never extracted or displayed to the user

### DIFF
--- a/packages/axios-extension/src/abap/catalog/base.ts
+++ b/packages/axios-extension/src/abap/catalog/base.ts
@@ -126,9 +126,7 @@ export abstract class CatalogService extends ODataService {
                 services
             };
         } catch (error) {
-            const errorCode = isAxiosError(error)
-                ? (error.response?.status as CatalogErrorCode)
-                : undefined;
+            const errorCode = isAxiosError(error) ? (error.response?.status as CatalogErrorCode) : undefined;
             const errorMessage = error instanceof Error ? error.message : String(error);
             return {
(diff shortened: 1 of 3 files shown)


IainSAP:
@unseen1980 There is already extensive error handling and message generation code. No new code is needed in axios extension for this.

There is no need for new version functions...this is already available. We already know what requests failed an why.

unseen1980:
After discussion with @IainSAP I am closing this. Changes should be made in a different area.. He will pick it up. @ianquigley-sap

Issue #3969, opened 2026-01-06
BUG - Deploy config is overwritten when No is selected


longieirl:
### Description
Deployment Configuration Generator overwrites deploy config when `No` is selected

### Steps to Reproduce
1. Go to VSC or BAS, generate Fiori app with deploy config to ABAP
2. Launch Deployment Configuration Generator for your app and change values - the most visible is a change of target from ABAP to CF
3. `Editing the deployment configuration will overwrite the existing configuration. Are you sure you want to continue?` - select `No` and click `Finish`

### Expected results
Deployment Configuration Generator aborts, there are no changes in deploy config

### Actual results
Deploy config is overwritten - e.g. if CF is chosen, files like `mta.yaml`, `xs-app.json`, `xs-security.json` are generated

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [x] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

longieirl:
Done and confirmed in tools-suite.

PR #3984, opened 2026-01-10
fix: only append overwrite if specified

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/3969


mikicvi-SAP (approved):
Thanks @longieirl 
- Changeset ✅
- Change is small and test to cover

PR #3985, opened 2026-01-12
fix(fpm-writer): call `mem-fs-editor/copyTpl` with `noGlob = true` to avoid failure for paths which includes brackets like `app(1)`

815are:
Scenario:
1. Install fpm-writer to folder like `c:\myApp(1)`
2. Run any generate command like `generateCustomColumn`;
3. See error:
```
Trying to copy from a source that does not exist:
```

Problem is that `copyTpl` is called with glob parameter be default - as fix pass `noGob=true`


devinea (approved):
Changes are clear. lgtm. 

FYI @IainSAP

PR #3989, opened 2026-01-13
tbi(sap-systems-ext): update backend systems with system info 

cianmSAP:
Internal ticket 31520

Adds support to automatically call the `/sap/bc/adt/core/http/systeminformation` endpoint when testing the connection of existing systems and saving new systems - and save the `systemID` and `client` which are returned


mikicvi-SAP (approved):
Thanks @cianmSAP 
- Code is clean and easy to follow
- Tests to cover the changes
- Tested locally and works as expected
- Cset ✅

PR #3082, opened 2025-04-02
WIP - Eslint 9

devinea:

PR #3993, opened 2026-01-14
fix: prompts examples fails because of full fpm writer import in browser

815are:
Issue after https://github.com/SAP/open-ux-tools/pull/3082

In `examples/ui-prompting-examples/src/utils/types.ts` we just export enum for storybook webapp usage -> https://github.com/SAP/open-ux-tools/blob/d0460a82b3dcc8db98e4c7bf4adfd17c494c683a/examples/ui-prompting-examples/src/utils/types.ts#L51
Problem happens if we import enum from whole `fpm-writer`:

image

Alternative could be to duplicate enum


devinea (approved):
Thanks @815are

PR #3999, opened 2026-01-14
fix: reset stale connection validity once auth type is changed for ABAP on BTP flow

mikicvi-SAP:
# Issue 
#4003

### Changes
- Resets connection validity and prompt state for abap on BTP flow properly to prevent unwanted prompts showing up (under specific conditions) when the auth type is changed


cianmSAP (approved):
Update looks good
Patch changeset is appropriate
Test added

PR #4004, opened 2026-01-15
Fix(eslint-plugin-fiori-tools): create new ProjectContext when checking a new file

AlinaGovoruhina:
Issue: 
When consuming the plugin to run eslint on 1 app, then checking another app (without reloading the window) - the plugin would not detect any issues in the new project.

Fix: 
- Remove return of empty ProjectContext
- Save diagnosticCache for a file first, for every rule in the file
- Proceed to create new Projectcontext
- use normalizePath to be compatible with Windows

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -6,6 +6,7 @@ const pluginJsdoc = require('eslint-plugin-jsdoc');
 const tseslint = require('typescript-eslint');
 const importPlugin = require('eslint-plugin-import');
 const sonarjs = require('eslint-plugin-sonarjs');
+const isFixMode = process.argv.includes('--fix');
 
 const compat = new FlatCompat({
     baseDirectory: __dirname, // optional; default: process.cwd()
@@ -87,7 +88,9 @@ module.exports = [
                     contexts: ['TSMethodSignature']
                 }
             ],
-
+            ...(isFixMode && { // disable in fix mode to avoid generating incorrect JSDoc
+                'jsdoc/require-jsdoc': 'off'
+            }),
             'jsdoc/valid-types': 'error',
             'jsdoc/check-types': 'error',
             'jsdoc/check-param-names': 'error',
@@ -347,6 +350,9 @@ module.exports = [
                     'exemptEmptyFunctions': true
                 }
             ],
+            ...(isFixMode && { // disable in fix mode to avoid generating incorrect JSDoc
+                'jsdoc/require-jsdoc': 'off'
+            }),
             'jsdoc/valid-types': 'off',
             'jsdoc/check-types': 'off',
             'jsdoc/check-tag-names': 'off',


igorsVoronins:
QA Result: Pass ✅ 
Tested the plugin with two projects; both correctly displayed ESLint errors as expected.
eslint_two_projects

packages/eslint-plugin-fiori-tools/src/project-context/project-context.ts
```diff
@@ -270,9 +270,11 @@ export class ProjectContext {
...
 
         if (this.appRoots.size > 0) {
-            // uri not is part of the known apps
-            // no point trying to reindex
-            return new ProjectContext({}, { projectType: 'EDMXBackend', apps: {}, documents: {} }, { apps: {} });
+            // Uri is not a part of known apps
+            // Assume a new app is opened
+            // Clear DiagnosticCache and continue to create new ProjectContext
+            DiagnosticCache.forceReindexOnFirstUpdate = true;
+            DiagnosticCache.clear(uri);
```
marcovieth: Do the static caches need to be cleared here as well?
this.instanceCache.clear();
this.updateCache.clear();
this.appRoots.clear();

marcovieth (commented):
I have not a complete understanding of all the caches right now but have a question...

packages/eslint-plugin-fiori-tools/src/project-context/project-context.ts
AlinaGovoruhina: instanceCahce hold information on previously opened projects context, we cache the value and don't create new ProjectContext every time the same app is opened. 
https://github.com/SAP/open-ux-tools/blob/33884295d688ff7a465637ed70b9a3334f5e32d5/packages/eslint-plugin-fiori-tools/src/project-context/project-context.ts#L204-L208
updateCache logs every time lint is run on a project to know if we have information already.
https://github.com/SAP/open-ux-tools/blob/33884295d688ff7a465637ed70b9a3334f5e32d5/packages/eslint-plugin-fiori-tools/src/project-context/project-context.ts#L222-L226
this.appRoots holds the previously opened appRoots so we don't reindex it:
https://github.com/SAP/open-ux-tools/blob/33884295d688ff7a465637ed70b9a3334f5e32d5/packages/eslint-plugin-fiori-tools/src/project-context/project-context.ts#L262-L270
I don't think we need to clear these, they are newly created on page reload.

AlinaGovoruhina: Let's proceed with @marufrasully solution, so resolving this.

marcovieth (approved):
I like the idea of having an URL-specific diagnostics cache.
The internal caches are still not cleared (only on reload), but I think that's ok now.

PR #4006, opened 2026-01-15
fix(ui-components): hidden options in combobox becomes visible

815are:
Issue with `UIComboBox`, scenario:
1. Pass options like:
```
const options = [
   {
        key: 'DZ',
        text: 'Algeria'
    },
    {
        key: 'AR',
        text: 'Argentina',
        hidden: true
    },
]
```
2. Enable search like:
```
<UIComboBox
            options={options}
            highlight={true}
            allowFreeform={true}
        />
```
3. Search for `Argentina`

Problem - option is found while it is marked as hidden.

### DIFF
--- a/packages/deploy-config-sub-generator/CHANGELOG.md
+++ b/packages/deploy-config-sub-generator/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/deploy-config-sub-generator
 
+## 0.5.40
+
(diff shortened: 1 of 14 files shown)


kjose90 (approved):
Thanks @815are for fix. Tested locally ✅ 
Hidden options does not appear as part of search bar in combobox.
Test + changeset ✅

Issue #3884, opened 2025-12-01
FEATURE - `odata-service-inquirer`: Support dynamic preselection/filter of system and service (based on path)


IainSAP:
### Description

Some new consumers of the odata-service-inquirer prompts require the ability to pre-select the system dynamically bound to a previous prompt answer and filter the services based on the service path, in addition to the technical system id. The service id s not available from a local application.

### Technical Design

Enhance the prompt option which currently supports only the technical service id which is not available from the application. https://github.com/SAP/open-ux-tools/blob/93ff524ddb43c8756eba7cb764f266a3aae6156b/packages/odata-service-inquirer/src/types.ts#L352 to additionally accept full service paths.

Make the system selection dynamic so it can be changed based on a previous prompt answer, for example a selected applciation

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a ...
**when** I execute ...
**then** ... will happen.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #4019, opened 2026-01-19
feat(`@sap-ux/odata-service-inquirer`): Adds new features to support re-use cases

IainSAP:
https://github.com/SAP/open-ux-tools/issues/3884

Internal ref: 36153

- Adds service selection prompt option to allow service path to be used to preselect (as well as service id)
- Adds service selection warning where preselected service is not available from list. 
- Fixes message shown when `requiredOdataVersion` is specified and preselected service is provided replacing the previous misleading message referring to a template. The message is more generic now for reuse in other contexts. Internal ref: 36750. Previous message still used when preselected service is not provided.
- Adds system preselection option that may be bound to a reference to allow runtime updates
- Adds tests to cover

### DIFF
--- a/.changeset/chilled-peaches-travel.md
+++ b/.changeset/chilled-peaches-travel.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/odata-service-inquirer': patch
----
-
-Fix for odata version restriction not applied for Abap-on-prem credentials prompts
--- a/examples/fe-fpm-cli/CHANGELOG.md
+++ b/examples/fe-fpm-cli/CHANGELOG.md
@@ -1,5 +1,19 @@
 # @sap-ux/fe-fpm-cli
 
+## 0.1.85
+
+### Patch Changes
+
+- Updated dependencies [68b0e54]
+    - @sap-ux/fe-fpm-writer@0.41.0
+
+## 0.1.84
+
+### Patch Changes
+
+- Updated dependencies [cc8871a]
+    - @sap-ux/fe-fpm-writer@0.40.12
+
 ## 0.1.83
 
 ### Patch Changes
--- a/examples/fe-fpm-cli/package.json
+++ b/examples/fe-fpm-cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/fe-fpm-cli",
-    "version": "0.1.83",
+    "version": "0.1.85",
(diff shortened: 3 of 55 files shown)


packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -261,8 +261,10 @@
         "certificateErrors": "A certificate error occurred when connecting to the host: {{-url}}. Certificate error: {{error}}.",
         "allowingUnauthorizedCertsNode": "Setting the `NODE_TLS_REJECT_UNAUTHORIZED` environment variable to `0` makes TLS connections and HTTPS requests insecure by disabling certificate verification. It is important to understand the security risks when using this setting.",
         "certErrorIgnoredByNodeSetting": "An SSL certificate error is being ignored by the 'NODE_TLS_REJECT_UNAUTHORIZED=0' Node.js setting. This makes TLS connections and HTTPS requests insecure by disabling certificate verification. For details of the certification error, see the logs.",
-        "noExternalServiceMetdataFetched": "No external service metadata were fetched. $t(texts.seeLogForDetails).",
-        "missingServices": "OData V{{odataVersion}} services are not available."
+        "noExternalServiceMetdataFetched": "No external service metadata were fetched. $t(texts.seeLogForDetails)",
```
```suggestion
        "noExternalServiceMetdataFetched": "No external service metadata was fetched. $t(texts.seeLogForDetails)",
```

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -261,8 +261,10 @@
...
+        "missingServices": "OData V{{odataVersion}} services are not available.",
+        "specifiedServicesNotAvailable_one": "The required OData service: {{-service}} is not available from the {{odataVersion}}services catalog. Please ensure the service is available.",
```
longieirl, line 266: Do you need a space before `services`?

IainSAP: No...its included in the parameter to avoid additional spaces when there is no odata version. See the unit test for confirmation.

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
lfindlaysap: @IainSAP, I noticed you resolved the suggestion but I don't see the change. Do you have thoughts about it?

PR #4021, opened 2026-01-19
Fix/cache eslint

marufrasully:


marufrasully:
Will be covered in this PR https://github.com/SAP/open-ux-tools/pull/4004/commits

PR #4024, opened 2026-01-19
fix(fpm-writer): fix width issue with custom filter field in filter bar

eouin:
Internal issue 36518

- [x] add width to 100% in combobox for custom filter field


815are (changes requested):
snapshots are failing for me:
image

kjose90 (approved):
code change looks good
snapshots  and changelog ✅

PR #4029, opened 2026-01-20
feat: move getDefaultTableType to @sap-ux/project-access

Vladislavs-Silins-SAP:
The `getDefaultTableType` function has been moved from the `@sap-ux/odata-service-inquirer` package to `@sap-ux/project-access `and made available for export. All necessary data types required for calling the function have also been moved.


.changeset/old-panthers-exercise.md
```diff
@@ -0,0 +1,99 @@
+---
+'@sap-ux/odata-service-inquirer': minor
```
heimwege, line 2: This changeset looks... interesting. I'd expect the changed packages here but I see almost all packages of the monorepo 🤔 Did you use `pnpm cset` to create this?

packages/odata-service-writer/test/unit/manifest.test.ts
```diff
@@ -2,8 +2,8 @@ import { deleteServiceData } from '../../src/delete';
 import type { Editor } from 'mem-fs-editor';
 import { create } from 'mem-fs-editor';
 import { create as createStorage } from 'mem-fs';
-import type { OdataService } from '../../src';
-import { OdataVersion, ServiceType } from '../../src';
+import { type OdataService, OdataVersion } from '../../src';
+import { ServiceType } from '../../src';
```
heimwege, line 6: This change seems odd. Ether separate by `import type` or merge into one.

```suggestion
import { type OdataService, OdataVersion, ServiceType } from '../../src';
```

packages/project-access/src/types/odata/index.ts
```diff
@@ -0,0 +1,24 @@
...
+    RESPONSIVE: 'ResponsiveTable',
+    TREE: 'TreeTable'
+} as const;
+
+export type TableType = (typeof TableType)[keyof typeof TableType];
+
+export enum OdataVersion {
+    v2 = '2',
+    v4 = '4'
+}
```
heimwege, line 24: nit-pick: I know it is just copied, but can we please avoid using enums? They prevent us from running this natively on node with `erasableSyntaxOnly` (https://www.totaltypescript.com/erasable-syntax-only).

```suggestion
export const OdataVersion {
    v2 = '2',
    v4 = '4'
} as const
```

Klaus-Keller: Yes, please no `enum`s.

packages/fiori-elements-writer/src/data/templateAttributes.ts
```diff
@@ -1,4 +1,4 @@
-import { OdataVersion, TemplateType } from '../types';
+import { OdataVersion, TemplateType } from '@sap-ux/project-access';
```
IainSAP, line 1: These types should remain here.

packages/fiori-elements-writer/src/index.ts
```diff
@@ -12,23 +12,18 @@ import {
...
-    TemplateTypeAttributes,
-    minSupportedUI5Version,
-    minSupportedUI5VersionV4,
-    escapeFLPText
-} from './data/templateAttributes';
+import { escapeFLPText } from './data/templateAttributes';
 import { extendManifestJson } from './data/manifestSettings';
 import semVer from 'semver';
 import { initI18n } from './i18n';
+import { TemplateType } from '@sap-ux/project-access';
```
IainSAP, line 26: TemplateType is an App Gen specific concept and as such should remain in those modules.

IainSAP:
@Vladislavs-Silins-SAP Im sorry but project access is not the right place for much of the moved code.

IainSAP (changes requested):
I would prefer we do not move this code to project access as proposed. Much of the moved types are specific to App Gen and should be considered internal. Conceptually project access is not the correct place for these things to live.

packages/project-access/src/types/odata/index.ts
```diff
@@ -0,0 +1,24 @@
+export const TemplateType = {
```
IainSAP, line 1: TemplateType is an App Gen specific concept. Certainly not an odata related concept (its under an odata folder ?) . project access should never need to know about the type of template selected during generation.

Issue #4028, opened 2026-01-20
BUG - VSC New System flow shouldn't allow saving of ABAP systems with URI, only URL


mikicvi-SAP:
### Related Feature
Internal report: `30742 `
### Description
- When adding a new on-premise ABAP system via Application Generator, currenty, the user is allowed to input `https://some-system-url.com/path/to/anything`. 
  - This works because axios-extensions discards the input with paths and only uses the origin URL.
- Application gets generated, and system gets saved (if the user chooses so), but the stored system would contain an illegal path (Connection Manager for SAP Systems does not allow this - see below).
  
<img width="512" height="304" alt="Image" src="https://github.com/user-attachments/assets/0a880acb-e981-4024-9e0b-baf326710b9e" />

### Steps to Reproduce
Steps to reproduce the behaviour:
Steps to reproduce the behaviour:
1. Go to App Generator via App Wizard
2. Select Data Source `Connect to a System`
3. Pick a System `New System`
4. Pick System Type `ABAP On Premise`
5. Enter `System URL` https://some-system-url.com/path/to/anything
6. Notice generator allows this

### Expected results
- The generator does not allow for non-origin URL input 

### Actual results
- The generator allows for non-origin URL input

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [x] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #4031, opened 2026-01-20
fix(odata-service-inq): restrict input of URLs for new systems flow hostname only for ABAP onprem

mikicvi-SAP:
#4028 

- Checks whether the URL should be treated as a system for validation in case of ABAP onPrem


packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -60,7 +60,7 @@
             "warningCertificateValidationDisabled": "Certificate validation has been disabled by the user.",
             "annotationsNotFound": "Annotations not found for the specified service.",
             "backendSystemExistsWarning": "A saved system connection entry: \"{{- backendName}}\" already exists with the same URL and client. Please reuse the existing entry or remove it, using the systems panel, before adding a new connection.",
-            "reentranceTicketSystemHostOnly": "ABAP cloud system urls must only contain the hostname, for example, 'https://123.abc.com'"
+            "urlPathWarning": "The provided URL must not contain a path."
```
IainSAP: This is less helpful than it was. I'd prefer:

```suggestion
            "systemUrlOriginOnlyWarning": "System urls must only contain the hostname, for example, 'https://123.abc.com/'"
```

@lfindlaysap Could you comment please?

lfindlaysap: If this is strictly about the hostname, rather than the entire path, for the former suggestion, I'd suggest:

System URLs must only contain the hostname, for example, 'https://123.abc.com/'.

ianquigley-sap: Can we ensure therefore that the same message is updated in the SAP Connection Manager please?

mikicvi-SAP: I can align, np.

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -60,7 +60,7 @@
             "warningCertificateValidationDisabled": "Certificate validation has been disabled by the user.",
             "annotationsNotFound": "Annotations not found for the specified service.",
             "backendSystemExistsWarning": "A saved system connection entry: \"{{- backendName}}\" already exists with the same URL and client. Please reuse the existing entry or remove it, using the systems panel, before adding a new connection.",
-            "reentranceTicketSystemHostOnly": "ABAP cloud system urls must only contain the hostname, for example, 'https://123.abc.com'"
+            "systemUrlOriginOnlyWarning": "System urls must only contain the hostname and/or port, for example, 'https://abc.com:123/'."
```
```suggestion
            "systemUrlOriginOnlyWarning": "System URLs must only contain the hostname and port, for example, 'https://abc.com:123/'."
```

mikicvi-SAP: I think this would imply that there has to be a port included, which would be incorrect.

lfindlaysap:
```suggestion
            "systemUrlOriginOnlyWarning": "System URLs must only contain the hostname and may also include the port, for example, 'https://abc.com:123/'."
```

mikicvi-SAP: Done, thanks @lfindlaysap.

PR #4040, opened 2026-01-21
tbi(odata-service-inquirer): Remove lodash

IainSAP:
- Removes lodash from `odata-service-inquirer` replacing with local implementations
- Adds tests to cover


devinea (approved):
Thanks @IainSAP

PR #4050, opened 2026-01-22
refactor(fiori-mcp-server): adjust telemetry functionalityId reporting

mustafamengutay-sap:
I have modified the telemetry framework to check if functionalityId is an array. If it is, it now reports only the property name (the last element) with a property-change: prefix to ensure data compliance, otherwise it accepts it as a string.

Checks if functionalityId is an array or not.
Prefixes the last element with 'property-change:' if it's a path.
Includes new unit tests for functionalityId.

### DIFF
--- a/.changeset/plenty-emus-tap.md
+++ b/.changeset/plenty-emus-tap.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/odata-service-inquirer': minor
----
-
-Improve entity validation in Generator.
--- a/packages/deploy-config-sub-generator/CHANGELOG.md
+++ b/packages/deploy-config-sub-generator/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/deploy-config-sub-generator
 
+## 0.5.57
+
+### Patch Changes
+
+- Updated dependencies [92244b8]
+    - @sap-ux/odata-service-inquirer@2.16.0
+
 ## 0.5.56
 
 ### Patch Changes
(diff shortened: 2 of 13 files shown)


.changeset/orange-comics-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/fiori-mcp-server': patch
+---
+
+refactor(fiori-mcp-server): adjust telemetry functionalityId reporting
```
heimwege: As this will be in the mcp changelog.md I think we don't need to mention the package name

```suggestion
refactor: adjust telemetry functionalityId reporting
```

packages/fiori-mcp-server/src/server.ts
```diff
@@ -123,7 +123,14 @@ export class FioriFunctionalityServer {
                     mcpClientVersion: this.mcpClientVersion
                 };
                 if ('functionalityId' in args) {
-                    telemetryProperties.functionalityId = args.functionalityId as string;
+                    const { functionalityId } = args;
+                    const shouldPrefixWithPropertyChange =
+                        Array.isArray(functionalityId) && functionalityId.length >= 1;
```
heimwege, line 128: nit-pick: Could be shortened to

```suggestion
                    const shouldPrefixWithPropertyChange = !!functionalityId?.at(0);
```

Klaus-Keller: @heimwege, are you sure? `functionalityId` can be of type `string` or `array`, `.at(0)` works also on strings.

heimwege: You are absolutely right. You can't use at to distinguish array from sting 🙈

heimwege (approved):
re-approve

PR #4053, opened 2026-01-22
feat: add auto-fix to sap-width-including-column-header rule

marcovieth:
This PR: 
Add auto-fix capability for `sap-width-including-column-header` rule.


marufrasully:
sonar cloud repot is false positive. It reports on JSDoc comment
Screenshot 2026-02-06 at 14 31 41

igorsVoronins (approved):
QA Pass ✅
The rule now supports auto-fix, ensuring the property is set to True when absent or incorrectly set to False on pages with small tables. Tested on V4 List Report app

heimwege:
Used admin to overcome yelling Sonar coverage. Overall coverage will be fine.

PR #4062, opened 2026-01-23
fix(ui5-test-writer): change order of file writing

docirl:
Internal issue 36816

- write opa test files after local annotations have been created
- ensures opa tests can be generated to include features in the local annotations


docirl:
@longieirl thanks, added one to test they are called in order
@IainSAP please re-review

PR #4064, opened 2026-01-26
feat(eslint-plugin-fiori-tools): add table copy rule for odata v2

AlinaGovoruhina:
Add table "copy" rule for OData V2 applications.
Table "copy" property must be undefined or set to `true` ("Copy" button must be enabled).
Add as a case in `sap-disable-copy-to-clipboard` rule, that was implemented for OData V4.
Rename this rule to `sap-copy-to-clipboard`, to be consistent with Copy property names.
Added a quick fix for both OData V2 and V4 applications, which removes the property from manifest.json - the default setting is to show the "Copy" button.

Added a delete fixer condition for removing a single property from an object.
Single property removal fix used to result in:
```
"test": {
}
```
Now the property is replaced with `"test": {}`.

### DIFF
--- a/examples/fe-fpm-cli/CHANGELOG.md
+++ b/examples/fe-fpm-cli/CHANGELOG.md
@@ -1,5 +1,18 @@
 # @sap-ux/fe-fpm-cli
 
+## 0.1.88
+
+### Patch Changes
+
+- @sap-ux/fe-fpm-writer@0.41.3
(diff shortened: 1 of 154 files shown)


marufrasully (approved):
@AlinaGovoruhina thanks
- code looks good
- great code coverage
- did not test it locally

PR #4067, opened 2026-01-27
OData Service Inquirer Text Updates

lfindlaysap:
I've updated some texts in the `odata-service-inquirer` i18n file to be more consistent with the other changes we have made.

With shorter labels instead of questions, can we move the longer and more informative text into tooltips?


packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -178,10 +178,10 @@
             "breadcrumb": "Generate Annotations"
         },
         "addLineItemAnnotations": {
-            "message": "Automatically add table columns to the list page and a section to the object page if none already exists?",
+            "message": "Add Table Columns to the List Report Page and a Section to the Object Page",
```
ianquigley-sap: @lfindlaysap `if none already exists?` is important here potentially. We don't always add table columns and a section, only if there are none there already.

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -208,7 +208,7 @@
             "hint": "Enables page level variant."
         },
         "valueHelpDownload": {
-            "message": "The service contains references to value help services. Do you want to download the associated metadata during generation? This may increase generation time.",
+            "message": "Download Associated Metadata",
```
ianquigley-sap: @lfindlaysap I think we should mention `Value help` here.  Otherwise the user mightn't know what it means.  Hiding this in the tooltip is a little too hidden I think?

lfindlaysap:
```suggestion
            "message": "Download Associated Metadata for Value Help",
```

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
lfindlaysap:
```suggestion
            "message": "Add Table Columns to the List Report Page and a Section to the Object Page If None Exist",
```

lfindlaysap:
@ianquigley-sap, I've addressed your comments. Can you re-review?

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -85,7 +85,7 @@
             "message": "Password"
         },
         "storeSystemCredentials": {
-            "message": "Do you want to store the system credentials?",
+            "message": "Store Credentials",
```
cianmSAP: This would go from 

Screenshot 2026-07-13 at 11 48 32

to 

Screenshot 2026-07-13 at 11 49 34

@lfindlaysap @ritarora @ianquigley-sap happy with this change?

lfindlaysap: I capitalized name in https://github.com/SAP/open-ux-tools/pull/4067/changes#diff-0fc31f4893572c970fa4c2e9ea8341dc76433e27321c459fc24a9765e351113dR104 but it doesn't seem to have taken affect 🤔.

cianmSAP: I just updated via web developer tools, not using the branch. Just to show the `storeSystemCredentials.message` prompt update. 
I've edited my screenshots to reflect the prompt in question

ritarora: @lfindlaysap for the above prompt, 'Store Credentials' does not seem descriptive enough to for the user to select the radio button option.

lfindlaysap: @ritarora, does **Store System Credentials** work better for you?

ritarora: Thanks @lfindlaysap, I still prefer the question, but if its a guideline to not use question format for the radio buttons, then this is fine

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -211,7 +211,7 @@
             "hint": "Enables page level variant."
         },
         "valueHelpDownload": {
-            "message": "The service contains references to value help services. Do you want to download the associated metadata during generation? This may increase generation time.",
+            "message": "Download Associated Metadata for Value Help",
```
ritarora: @lfindlaysap Can we update it to "Download Associated Metadata for Value Help services" , thanks

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```suggestion
            "message": "Store System Credentials",
```

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
lfindlaysap:
```suggestion
            "message": "Download Associated Metadata for Value Help Services",
```

lfindlaysap:
**Notes:** Per alignment with Ritika, add info from old Value Help text to the hint and update tests.

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -181,10 +181,10 @@
             "breadcrumb": "Generate Annotations"
         },
         "addLineItemAnnotations": {
-            "message": "Automatically add table columns to the list page and a section to the object page if none already exists?",
+            "message": "Add Table Columns to the List Report Page and a Section to the Object Page If None Exist",
```
cianmSAP, line 184: Just wondering why `If None Exist` is capatlised here?

packages/odata-service-inquirer/src/translations/odata-service-inquirer.i18n.json
```diff
@@ -211,7 +211,7 @@
             "hint": "Enables page level variant."
         },
         "valueHelpDownload": {
-            "message": "The service contains references to value help services. Do you want to download the associated metadata during generation? This may increase generation time.",
+            "message": "Download Associated Metadata for Value Help Services",
             "hint": "Download metadata for value help to improve the user experience with dropdowns and search."
```
cianmSAP, line 215: Can we add `This may increase generation time.` ?

IainSAP:
> With shorter labels instead of questions, can we move the longer and more informative text into tooltips?


@lfindlaysap On the CLI we still need the long questions as there are no hints. The messages are not labels but prompt (i.e. queries for the user hence questions). While this might work in YUI it wont be clear to a user what is being asked on the CLI. The same messages are used in both.

PR #4073, opened 2026-01-28
restrict the input for ABAP Cloud systems to origin only

mikicvi-SAP:
Addition for #4028 

- Applies the same URL check to ABAP Cloud systems in the Connection Manager
- Fixes a minor issue where the buttons "Test Connection" and "Save" were not disabled if the error message was shown / input was invalid

Preview: 
image

### DIFF
--- a/examples/fe-fpm-cli/CHANGELOG.md
+++ b/examples/fe-fpm-cli/CHANGELOG.md
@@ -1,5 +1,18 @@
 # @sap-ux/fe-fpm-cli
 
+## 0.1.88
+
+### Patch Changes
+
+- @sap-ux/fe-fpm-writer@0.41.3
+
+## 0.1.87
+
+### Patch Changes
+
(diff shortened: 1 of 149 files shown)


packages/sap-systems-ext-webapp/src/components/layout/main/systemInfo/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+    setIsDetailsValid: (isValid: boolean) => void
+): string | undefined => {
+    let urlMessage: string | undefined;
+    try {
+        const url = new URL(value);
+        if (url.pathname && url.pathname !== '/') {
+            setIsDetailsValid(false);
+            urlMessage = t('validations.systemUrlOriginOnlyWarning');
+        } else {
+            setIsDetailsValid(true);
```
cianmSAP, line 23: 👍🏻

PR #4082, opened 2026-01-29
Edit i18n text titles for RTE Button Groups

kjose90:
Edit i18n text titles for RTE Button Groups


815are:
checked:
1. code - only i18n file changes, did not notice issue
2. changelog persists

PR #4087, opened 2026-01-30
Feat(eslint-plugin-fiori-tools): add table paste and export property rules for oData V4 applications

AlinaGovoruhina:
In OData V4 applications paste and export functionality must be enabled.
The default value is true, so expected value for both settings are either true or not defined.

Add 2 eslint rules to the plugin, both auto-fixable. Fix: remove the setting to use the default value.
The warning is reported in the page tableSettings in manifest.json.
Unit tests: created a variable V4_FACETS_ANNOTATIONS for annotations data used in multiple tests.

### DIFF
--- a/packages/fiori-mcp-server/CHANGELOG.md
+++ b/packages/fiori-mcp-server/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/fiori-mcp-server
 
+## 0.6.21
+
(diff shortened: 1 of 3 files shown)


marufrasully (approved):
Thanks @AlinaGovoruhina 
- code changes clear
- secured with unit tests
- did not test it locally

Issue #4084, opened 2026-01-30
FEATURE - Add `list-systems` functionality to fiori-mcp server


nikmace:
_As a **developer**,
I want to **ask a model to list available SAP Systems**,
so that I **then choose one to work with**_

### Description

Allowing the modal to see the systems and display them for a user to then perform other actions with that system is a good approach for a better UX when working with AI-capable tools.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a Client that allows me to prompt a modal
**when** I have asked a modal to get a list of available systems
**then** a functionality is called to retrieve all systems, and modal displays them for me

### Tasks
- [ ] Add a new `list-systems` functionality
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #4088, opened 2026-01-30
feat: Add `list-systems` functionality to fiori-mcp server

nikmace:
Feat for #4084.
- Adds a new functionality that retrieves SAP Systems that are then visualized to the user by the selected model.


packages/fiori-mcp-server/src/tools/functionalities/list-systems/index.ts
```diff
@@ -0,0 +1,13 @@
...
+import details from './details';
+import executeFunctionality from './execute-functionality';
+
+export default {
+    id: details.functionalityId as string,
+    details,
+    handlers: {
+        getFunctionalityDetails: (): Promise<GetFunctionalityDetailsOutput> => Promise.resolve(details),
+        executeFunctionality
+    } as FunctionalityHandlers
```
heimwege: I think it would be better to use `satisfies` here

packages/fiori-mcp-server/src/tools/functionalities/list-systems/details.ts
```diff
@@ -0,0 +1,21 @@
+import type { GetFunctionalityDetailsOutput } from '../../../types';
+
+export default {
+    functionalityId: 'list-systems',
+    name: 'List SAP Systems',
+    description: `Lists all available SAP systems configured for Fiori applications. 
+    This functionality retrieves systems from either BAS (via destinations) or from the backend system store.
```
heimwege: Not sure about usage of abbreviations in mcp descriptions. I think they should be as precise as possible. Abbreviations might lead to misunderstandings.

packages/fiori-mcp-server/src/tools/functionalities/list-systems/details.ts
```diff
@@ -0,0 +1,21 @@
...
+    Use this to get a list of available systems before generating a Fiori application.
+    
+    **Structure and Usage:**
+    - Each system has a \`name\` field (required) and an optional \`client\` field
+    - Present the systems in a structured table format to the user with columns: System Name and Client
```
heimwege, line 12: As not each client seems to be able to do this, not sure if we should have this description or if it can also distract the client.
Also not sure if elicitation would help for this use case.
Or how about dynamically creating a resource for the result list (e.g. resource://temp/system-selection-list) to prevent context pollution with the huge list of systems that is basically irrelevant after the selection and might distract the llm?

heimwege (commented):
See comments.

packages/fiori-mcp-server/src/tools/functionalities/list-systems/index.ts
nikmace: @heimwege I don't mind making this change, it's just that I am following the general code style of this package and other similar files have `as`. Would you like me to make the change?

packages/fiori-mcp-server/src/tools/functionalities/list-systems/details.ts
nikmace: Do you think we should make functionalityId: `list-systems` -> `list-sap-systems`?

Also, you mean BAS -> Business Application Studio

packages/fiori-mcp-server/src/tools/functionalities/list-systems/details.ts
nikmace: Using MCP resources (e.g., resource://temp/system-selection-list) would help avoid context pollution. This would require:

1. Adding resource capabilities to the server initialization
2. Implementing ListResourcesRequestSchema and ReadResourceRequestSchema handlers
3. Modifying execute-functionality.ts to create/update a resource URI instead of embedding the full list in the message
4. Updating the description to reference the resource URI

This is a larger architectural change and probably needs its own PR. I could actually create this in the same PR but it would take more time and be harder to review. What do you think @heimwege?

packages/fiori-mcp-server/src/tools/functionalities/list-systems/index.ts
heimwege: I did not review all MCP PRs maybe that's why 😄 Using `as` here basically says the compiler "I know what this is" vs `satisfiers` rather being "ensure that this is". The biggest danger with `as` is that you can accidentally lie to the compiler, and it will believe you until the mcp breaks. That'S why I think using `satisfies` is safer here. But it's also rather a nit-pick, so I leave it up to you.

packages/fiori-mcp-server/src/tools/functionalities/list-systems/details.ts
heimwege: I meant BAS. Sorry for leaving that part out 🙈 Not sure if llm will always understand BAS the way it should. Sometimes if I write w.g. 'MCP' it is being interpreted as Minecraft "Mod Coder Pack" 🤣

packages/fiori-mcp-server/src/tools/functionalities/list-systems/details.ts
heimwege: As you already said this is maybe a larger architectural question we should discuss in a meeting first. This were just some ideas that came to my mind w/o even knowing if it would work or improve things.

@Klaus-Keller, @devinea you thoughts?

packages/fiori-mcp-server/src/tools/functionalities/list-systems/index.ts
nikmace: Changed in: [e083d3b](https://github.com/SAP/open-ux-tools/pull/4088/commits/e083d3b49d2154b056fadd316ddf01e7d556a635)

packages/fiori-mcp-server/src/tools/functionalities/list-systems/details.ts
nikmace: Changed in: [e083d3b](https://github.com/SAP/open-ux-tools/pull/4088/commits/e083d3b49d2154b056fadd316ddf01e7d556a635)

packages/fiori-mcp-server/src/tools/functionalities/list-systems/details.ts
Klaus-Keller: Using resources here is an interesting idea. Doesn't really need to be a `resource://temp/system-selection-list`, it could also be a fixed resource  like `resource://systems` that has the list of system names.
The Fiori MCP server does not use resources yet and we probably need to check how the system list could be integrated, meaning when to populate, update, remove. 

If this is a blocker for further ADP features I'd say we can go with the approach in this PR and investigate usage of systems as resources later, if there is no time constrain we can also check it out right now.

nikmace: @Klaus-Keller, I think it would be better to do it in another PR also. Therefore, if you don't mind, can you review this PR as is?

Klaus-Keller:
Hi @nikmace, can you check the Sonar issue: https://sonarcloud.io/project/issues?id=SAP_open-ux-tools&pullRequest=4088&issueStatuses=ACCEPTED. Why was this accepted, wouldn't named export not be easier and compliant to Sonar? What is the reason for accepting this issue?

nikmace:
> https://sonarcloud.io/project/issues?id=SAP_open-ux-tools&pullRequest=4088&issueStatuses=ACCEPTED

I've left a comment here on why. Sonar should have caught this way before my PR, but I guess it was dismissed also. I don't mind making this change. Do I need to change it to named export in every functionality in this package, also for consistency? @Klaus-Keller 
Screenshot 2026-02-10 at 14 57 22

Klaus-Keller:
> > https://sonarcloud.io/project/issues?id=SAP_open-ux-tools&pullRequest=4088&issueStatuses=ACCEPTED
> 
> I've left a comment here on why. Sonar should have caught this way before my PR, but I guess it was dismissed also. I don't mind making this change. Do I need to change it to named export in every functionality in this package, also for consistency? @Klaus-Keller Screenshot 2026-02-10 at 14 57 22

Thanks @nikmace, to not clutter this PR with other changes I would do the named export for the newly introduced functions only. We can clean up the others afterwards.

Issue #4096, opened 2026-02-03
FEATURE - Add controller extension and fragment generation through MCP


nikmace:
_As a **developer**,
I want to **tell a model to create an ADP from a system and application**,
so that I **can generate a fragment and controller extension files**_

### Description
TBD...

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a client where I can use llm models and a fiori-mcp server running
**when** I execute ask the chat to make a control and put it in a fragment file
**then** fragment file with a control inside and a controller extension file gets written

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #4097, opened 2026-02-03
feat(mcp): Add controller extension and fragment generation

nikmace:
Feat for #4096.
- Adds the logic needed for the model to generate a controller extension and fragment files by taking into account the controlName, viewId, and other information.


nikmace:
Outdated

PR #4098, opened 2026-02-03
Feat(eslint-plugin-fiori-tools): add table column vertical alignment rule

AlinaGovoruhina:
The rule Is only applicable to SAP Fiori elements for OData V2 projects and only for ResponsiveTable type tables. Setting has a check for min UI5 version = 1.75.
manifest.json path :  `$[\"sap.ui.generic.app\"].settings.tableColumnVerticalAlignment`.
Must be set to 'Middle' (or not defined as the default value is 'Middle').
Rule has an auto fix: removes set value.


marufrasully (approved):
@AlinaGovoruhina thanks
- code changes looks good
- secured with unit test
- did not verify it

igorsVoronins (approved):
QA Pass ✅ 
Tested with LROP and OVP V2 apps.
Made sure that only "Middle" value does not throw a warning. Auto-fix removes the value, as "Middle" is set by default. 
Also checked that rule is not working with minUI5 lower than 1.75

PR #4099, opened 2026-02-03
feat: refactor table capability functions to @sap-ux/project-access

Vladislavs-Silins-SAP:
Moved hasRecursiveHierarchyForEntitySet(), hasAggregateTransformations(), and hasAggregateTransformationsForEntitySet() functions to @sap-ux/project-access and created a wrapper function getTableCapabilitiesByEntitySet() to provide a simple and extensible way to access table capabilities.

### DIFF
--- a/.changeset/happy-buses-clean.md
+++ b/.changeset/happy-buses-clean.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/fiori-mcp-server': patch
----
-
-Update dependency "@modelcontextprotocol/sdk": "1.26.0"
--- a/examples/fe-fpm-cli/CHANGELOG.md
+++ b/examples/fe-fpm-cli/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/fe-fpm-cli
 
+## 0.1.94
+
+### Patch Changes
+
+- @sap-ux/fe-fpm-writer@0.42.2
+
 ## 0.1.93
 
 ### Patch Changes
--- a/examples/fe-fpm-cli/package.json
+++ b/examples/fe-fpm-cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/fe-fpm-cli",
-    "version": "0.1.93",
+    "version": "0.1.94",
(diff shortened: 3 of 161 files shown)


Vladislavs-Silins-SAP:
From the https://github.com/SAP/open-ux-tools/pull/4029

@heimwege wrote: "This changeset looks... interesting. I'd expect the changed packages here but I see almost all packages of the monorepo 🤔 Did you use `pnpm cset` to create this?"

I used the command `"cset": "pnpm validate:changesets && changeset"` from package.json. I think we got many affected packages because all packages that use project-access and inquirer-common need to consume the changes from these packages.

packages/project-access/src/odata/index.ts
```diff
@@ -0,0 +1 @@
+export * from './metadata';
```
Klaus-Keller: Hi @Vladislavs-Silins-SAP, could you switch to specific exports, like
```suggestion
export { findRecursiveHierarchyKey } from './metadata';
```

This has two advantes:
1. Just looking at the code you can directly tell what is exported, no need to navigate and remember all exports
2. No risk of accidentally export a function/const/...  to public interface by just exporting it on a lower level file

packages/project-access/src/index.ts
```diff
@@ -71,4 +71,5 @@ export {
 export { execNpmCommand } from './command/npm-command';
 export * from './types';
 export * from './library';
+export * from './odata';
```
Klaus-Keller: Same here, could you please just export the functions that need to be in the public interface?

packages/project-access/src/odata/index.ts
Vladislavs-Silins-SAP: Thanks for the suggestion. I’ve applied the proposed changes.

packages/project-access/src/index.ts
Vladislavs-Silins-SAP: Done.

Vladislavs-Silins-SAP:
> @Vladislavs-Silins-SAP, please see my comments

@Klaus-Keller I've addressed your comments, could you please continue the review?

IainSAP (approved):
Re-approve

PR #4100, opened 2026-02-03
mcp - update readme for Self-Signed SSL Certificates

devinea:
also, switch to inspector to devDeps instead of npx for better performance.

### DIFF
--- a/.changeset/friendly-rats-mate.md
+++ b/.changeset/friendly-rats-mate.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/fiori-docs-embeddings': patch
----
-
-fix: add documentation sources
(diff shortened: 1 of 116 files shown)


packages/fiori-mcp-server/README.md
```diff
@@ -123,6 +123,76 @@ The following log levels are supported: `off`, `error`, `warn`, `info`, `debug`,
...
+```
+
+**macOS/Linux Example:**
+```json
+"env": {
+  "NODE_EXTRA_CA_CERTS": "/Users/username/certs/Custom_CA.crt"
+}
+```
+
+### Option 2: Bypass SSL Validation (Not Recommended)
```
heimwege, line 169: If it is not recommended and even adds a security risk, do we really want to add it to the readme?

devinea: I think it is ok. It aligns with https://www.npmjs.com/package/@sap/ux-ui5-tooling#ignoring-certificate-errors

heimwege (approved):
lgtm

PR #4104, opened 2026-02-04
feat(eslint-plugin-fiori-tools): add V4 table personalization rule

AlinaGovoruhina:
Add an eslint rule: check table personalization is enabled in OData V4 projects.
Checks application manifest.json file for all available tables.

Valid values: undefined, true, all available personalization properties set to true.

Invalid: false, properties set to false or undefined (In case of using an object, omitted properties are considered disabled).

Quick fix: sets personalization to true (enables all table personalization features).
Note: table grouping is available for Analytical and Responsive tables only.


packages/eslint-plugin-fiori-tools/test/rules/sap-table-personalization.test.ts
```diff
@@ -0,0 +1,862 @@
...
+                            'targets',
+                            'IncidentsObjectPage',
+                            'options',
+                            'settings',
+                            'controlConfiguration',
+                            'incidentFlow/@com.sap.vocabularies.UI.v1.LineItem',
+                            'tableSettings',
+                            'personalization'
+                        ],
+                        value: false
```
marcovieth: When the test description is correct, value should be {}

packages/eslint-plugin-fiori-tools/test/language/rule-fixer.test.ts
```diff
@@ -202,6 +202,41 @@ describe('createJsonFixer', () => {
             expect(fixer.replaceTextRange).toHaveBeenCalled();
             expect(result).toMatchSnapshot();
         });
+
+        it('should update the value for a parent node - fixParent = true', () => {
+            const sourceText = `{"parent": {"old": true}}`;
+            const { node, ast } = createNodeFromJson(sourceText, ['parent']);
```
marcovieth: Shouldn't we use ['parent', 'old'] for the node path here so that the node path matches the validationPath below?

Same question for the two following tests...

packages/eslint-plugin-fiori-tools/docs/rules/sap-table-personalization.md
```diff
@@ -0,0 +1,83 @@
...
+}
+```
+
+Every table `personalization` property is correctly set to `true`.
+Omitting table personalization from the `manifest.json` table settings is also correct, because all personalization settings are provided by default for all tables.
+
+## How to Fix
+
+To fix the warning, ensure that either the `personalization` property is set to `true` or not defined in the table settings.
+
```
marcovieth, line 101: Do we also want to mention here what the automatic fix does?

packages/eslint-plugin-fiori-tools/docs/rules/sap-table-personalization.md
```diff
@@ -0,0 +1,83 @@
...
+```json
+{
+  "tableSettings": {
+    "personalization": {}
+  }
+}
+```
+
+If the value "object" is used, omitting a setting is treated as `false`.
```
marcovieth: Shall we also have a partially defined object as an example?

And also an example with individual properties set to false?

packages/eslint-plugin-fiori-tools/test/rules/sap-table-personalization.test.ts
```diff
@@ -0,0 +1,862 @@
...
+            [FACETSV4]
+        ),
+        createValidTest(
+            {
+                name: 'V4 - list report page - ResponsiveTable + minUI5=1.120 - group is true',
+                filename: V4_MANIFEST_PATH,
+                code: getManifestAsCode(V4_MANIFEST, [
+                    {
+                        path: ['sap.ui5', 'dependencies', 'minUI5Version'],
+                        value: '1.120.1'
```
marcovieth: Do we also need a test for exact boundaries like '1.120.0'? Maybe we can use that version here.
Same for '1.108.0' below.

packages/eslint-plugin-fiori-tools/test/rules/sap-table-personalization.test.ts
```diff
@@ -0,0 +1,862 @@
...
+                            'tableSettings',
+                            'personalization'
+                        ],
+                        value: true
+                    }
+                ])
+            },
+            [FACETSV4]
+        ),
+        createInvalidTest(
```
marcovieth, line 920: Test coverage is quiet good.
Do we also need a test for a single undefined property to get the 'is disabled' message?

packages/eslint-plugin-fiori-tools/test/language/rule-fixer.test.ts
AlinaGovoruhina: Do you mean in `createNodeFromJson()`? This creates a node from 'parent' with 'old' inside, if I pass it ['parent', 'old'] it will create a node from 'old'.

packages/eslint-plugin-fiori-tools/test/rules/sap-table-personalization.test.ts
AlinaGovoruhina: Setting `personalization` to false is the same as setting every property to false, so everything is disabled. `{}` also sets all properties to `false`, there is a test for empty object value below.

packages/eslint-plugin-fiori-tools/docs/rules/sap-table-personalization.md
AlinaGovoruhina: Added fix information here.

packages/eslint-plugin-fiori-tools/docs/rules/sap-table-personalization.md
AlinaGovoruhina: Added this code example.

packages/eslint-plugin-fiori-tools/test/rules/sap-table-personalization.test.ts
AlinaGovoruhina: Good point, added tests for AnalyticalTable+1.108.0 and ResponsiveTable 1.120.0, where `group` property value is available and checked.

packages/eslint-plugin-fiori-tools/test/rules/sap-table-personalization.test.ts
AlinaGovoruhina: Added the unit test for single undefined property.

packages/eslint-plugin-fiori-tools/src/rules/sap-table-personalization.ts
```diff
@@ -0,0 +1,206 @@
...
+        problems.push({
+            type: TABLE_PERSONALIZATION,
+            pageName: page.targetName,
+            messageId: MessageIdByProperty[''],
+            manifest: {
+                uri: parsedApp.manifest.manifestUri,
+                object: parsedApp.manifestObject,
+                propertyPath: table.configuration.personalization.configurationPath
+            }
+        });
```
marufrasully, line 202: returning here would save else block :)

marufrasully (approved):
Thanks @AlinaGovoruhina 
- code changes looks good
- great test coverage
- did not verify it

packages/eslint-plugin-fiori-tools/test/rules/sap-table-personalization.test.ts
marcovieth: But then the test description does not match the value "false", so it should mention "false".

packages/eslint-plugin-fiori-tools/test/language/rule-fixer.test.ts
marcovieth: Yes, I mean `createNodeFromJson(sourceText, ['parent', 'old'])`.
My understanding is that it will then create a node from "old". Then with "fixParent: true" it will operate on the parent object (even if it's just a simulation). And node path = validation path, as it is with other tests.

packages/eslint-plugin-fiori-tools/test/rules/sap-table-personalization.test.ts
AlinaGovoruhina: Ok, I changed the tested value to be consistent with test name.

packages/eslint-plugin-fiori-tools/test/language/rule-fixer.test.ts
AlinaGovoruhina: I see, I think I mixed up node and ast, so it is correct: we need `old` node fo the test. it is updated.

packages/eslint-plugin-fiori-tools/src/rules/sap-table-personalization.ts
AlinaGovoruhina: Thanks, I added the return statement in https://github.com/SAP/open-ux-tools/pull/4104/commits/3e0ce1f1948fdd83123dd068a052737d485dd586

packages/eslint-plugin-fiori-tools/test/language/rule-fixer.test.ts
```diff
@@ -489,6 +568,45 @@ describe('createJsonFixer', () => {
 
             expect(result).toEqual([]);
         });
+
+        it('should delete parent object value if fixParent = true', () => {
+            const sourceText = `{"parent": {"old": true}}`;
+            const { node, ast } = createNodeFromJson(sourceText, ['parent']);
```
marcovieth: Test case for "update" is ok now. I think for "delete" we also need: createNodeFromJson(sourceText, ['parent', 'old']);

packages/eslint-plugin-fiori-tools/test/language/rule-fixer.test.ts
```diff
@@ -323,6 +362,46 @@ describe('createJsonFixer', () => {
             expect(result).toEqual([]);
             expect(fixer.insertTextBeforeRange).not.toHaveBeenCalled();
         });
+
+        it('should insert value on a parent node - fixParent = true', () => {
+            const sourceText = `{"parent": {"old": true}}`;
+            const { node, ast } = createNodeFromJson(sourceText, ['parent']);
```
marcovieth: Test case for "update" is ok now. I think for "insert" we also need: createNodeFromJson(sourceText, ['parent', 'old']);

packages/eslint-plugin-fiori-tools/test/language/rule-fixer.test.ts
AlinaGovoruhina: Thanks, updated this test

packages/eslint-plugin-fiori-tools/test/language/rule-fixer.test.ts
AlinaGovoruhina: Updated the test as well

PR #4118, opened 2026-02-05
fix: delete state preservation mode

marufrasully:
This PR remove `state preservation mode` with auto fix option.

### DIFF
--- a/.changeset/tricky-oranges-move.md
+++ b/.changeset/tricky-oranges-move.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/ui-components': patch
----
-
-UIFlexibleTable. Large cell content in flexible table items no longer shifts the other column to the right.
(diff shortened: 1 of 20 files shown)


packages/eslint-plugin-fiori-tools/src/rules/sap-state-preservation-mode.ts
```diff
@@ -145,7 +145,13 @@ const rule: FioriRuleDefinition = createFioriRule({
                     value: diagnostic.value,
                     recommended: diagnostic.recommendedValue
                 },
-                fix: createJsonFixer({ context, node, deepestPathResult: paths, value: diagnostic.recommendedValue })
+                fix: createJsonFixer({
+                    context,
+                    node,
+                    deepestPathResult: paths,
+                    value: diagnostic.recommendedValue,
```
AlinaGovoruhina: Value not needed for delete operation.

PR #4121, opened 2026-02-05
fix(odata-service-inquirer): Catalog auth fail shown incorrectly

IainSAP:
Fix for internal bug: 36973

- Fixes info message being shown when service selected (indicating that the previous error should not be shown)
- Updates test to cover


devinea (approved):
changeset ✅ 
Code changes and test updates lg.

PR #4122, opened 2026-02-06
fix: Rename action missing for Object Page elements in Adaptation Editor

testojs:
Fix for #4409.

- Put back missing Rename action for Object Page elements in ADP                                                                                                        
- Disable unsupported annotation changes in developer mode

### DIFF
--- a/.changeset/tough-melons-spend.md
+++ b/.changeset/tough-melons-spend.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/eslint-plugin-fiori-tools': patch
----
-
-feat: add plugin version number to rule in readme
--- a/packages/eslint-plugin-fiori-tools/CHANGELOG.md
+++ b/packages/eslint-plugin-fiori-tools/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/eslint-plugin-fiori-tools
 
+## 9.9.3
+
+### Patch Changes
+
+- 9afb66a: feat: add plugin version number to rule in readme
+
 ## 9.9.2
 
 ### Patch Changes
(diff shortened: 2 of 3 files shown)


packages/preview-middleware/src/base/config.ts
```diff
@@ -255,14 +255,14 @@ function getFlexSettings(): TemplateConfig['ui5']['flex'] {
 
     return [
         { connector: 'LrepConnector', layers: [], url: '/sap/bc/lrep' },
+        {
```
Lonwyr: the move of the localstorage connector with the type "USER" will result in applying the user changes on a load BEFORE the localConnector

testojs: @edingerk What do you think?

edingerk: I double checked, yes Christian is right. This is very unfortunate, as with this fact you will not be able to fix this issue in previous releases. So we will need to do a downport in UI5
I think we should think again about this decision, as it relies on the customers putting the connectors in the correct order, and connectors to not serve changes from layers that are not direct neighbors

edingerk: Maybe we can try to completely remove the LocalStorageConnector from the configuration, then any personalization will not be 'persisted' anymore, but should still be able to be done for that session. This will also do the trick

.changeset/real-eagles-mix.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ux-private/preview-middleware-client': patch
+'@sap-ux/preview-middleware': patch
+---
+
+fix: Rename action missing for Object Page elements in Adaptation Editor
```
heimwege: Shouldn't this changeset mention the disabling of annotation changes?

testojs: Sure, makes sense.
Added in [92c1bff](https://github.com/SAP/open-ux-tools/pull/4122/commits/92c1bffd4441ca6f339d7e2a94f536f9e66bbc15).

PR #4123, opened 2026-02-06
fix: UIFlexibleTable. Large cell content in flexible table items no longer shifts the other column to the right.

815are:
Issue before fix - if content of cell is large, then it moved coulmns on right:
image

after fix:
image


Jimmy-Joseph19 (approved):
Changes looks good. Approving

PR #4093, opened 2026-01-30
feat(app-config-writer): enable card generator for CAP projects

Singham-24:
__Description:__ This PR introduces support for generating cards in CAP-based applications. The enhancement ensures that the card generator configuration integrates smoothly with CAP project structures.

__Key Changes:__

- Updated `enableCardGeneratorConfig` function to handle CAP project paths
- Added support for webapp path detection in CAP project structure

__Related to:__ [PR #3737](https://github.com/SAP/open-ux-tools/pull/3737)

PR #4092, opened 2026-01-30
feat(create): enable card generator for CAP projects

Singham-24:
__Description:__ This PR adds CLI support for enabling the Card Generator configuration in CAP projects through the `@sap-ux/create` package.

__Key Changes:__

- Added `--app` option to specify the app folder path for CAP projects
- Implemented auto-detection of Fiori apps in CAP projects using `findFioriArtifacts`
- Added script to root `package.json` for CAP projects to start the card generator with `cds watch --open`

__Usage:__

```bash
# For CAP projects with --app option
npx @sap-ux/create add cards-editor --app <app-name>

```

__Related to:__ [PR #3737](https://github.com/SAP/open-ux-tools/pull/3737)

PR #4091, opened 2026-01-30
feat(preview-middleware): enable card generator for CAP projects

Singham-24:
__Description:__ This PR introduces support for generating cards in CAP-based applications. The enhancement ensures that the card generator integrates smoothly with CAP project structures and runtime behavior.

__Key Changes:__

- Added `addRootLevelCardRoutes()` method to register card generator routes at the CDS server's root level
- Implemented global `FlpSandbox` registry to track instances by app ID for CAP projects where multiple apps share the same server
- Added routes for `/cards/*`, `/manifest.json`, and `/i18n/*` at root level to serve card manifest files

__Related to:__ [PR #3737](https://github.com/SAP/open-ux-tools/pull/3737)

PR #4128, opened 2026-02-09
feat: enable card generator for CAP projects

Singham-24:
__Description:__ 

This patch enables the card generator feature for CAP projects. It adds default FLP configuration when preview middleware exists but FLP config is missing, supports path derivation for CAP projects when yaml paths contain directory separators, and uses `getSourcePath()` for correct webapp path resolution in card generator handlers.

__Related to:__ [PR #3737](https://github.com/SAP/open-ux-tools/pull/3737)

### DIFF
--- a/.github/workflows/renovate-pr-automation.yml
+++ b/.github/workflows/renovate-pr-automation.yml
@@ -1,0 +1,87 @@
+name: Renovate PR Automation
+
+# This workflow automates Renovate PR follow-up tasks:
+# - For general dependency updates: Creates changesets automatically
+# - For @ui5/manifest updates: Runs manifest-specific updates (fixtures, fallbacks, custom changeset)
+
+on:
+    pull_request_target:
+        types: [opened, synchronize, reopened]
+        branches: [main]
+
+jobs:
+    # Job 1: Create changeset for general Renovate dependency PRs (NOT @ui5/manifest)
+    create-changeset:
+        name: Create Dependency Changeset
+        runs-on: ubuntu-latest
+        # All Renovate PRs EXCEPT @ui5/manifest (which has custom handling)
+        if: |
+            github.actor == 'renovate[bot]' &&
+            !contains(github.event.pull_request.title, '@ui5/manifest')
+        steps:
+            - name: Checkout PR branch
+              uses: actions/checkout@v6
+              with:
+                  ref: ${{ github.head_ref }}
+                  fetch-depth: 0
+                  token: ${{ secrets.ACCESS_PAT }}
+
+            - name: Create changeset
+              uses: mscharley/dependency-changesets-action@v1.2.2
+              with:
+                  author-email: 'github-actions[bot]@users.noreply.github.com'
+                  commit-message: 'chore: add changeset for dependency update'
+
+    # Job 2: Run manifest-specific updates (only for @ui5/manifest PRs)
+    update-manifest-fixtures:
+        name: Update @ui5/manifest Fixtures & Changeset
+        runs-on: ubuntu-latest
+        if: |
+            github.actor == 'renovate[bot]' &&
+            contains(github.event.pull_request.title, '@ui5/manifest')
+        steps:
+            - name: Checkout PR branch
+              uses: actions/checkout@v6
+              with:
+                  ref: ${{ github.head_ref }}
+                  fetch-depth: 0
+                  token: ${{ secrets.ACCESS_PAT }}
+
+            - name: Setup pnpm
+              uses: pnpm/action-setup@v4.2.0
+
+            - name: Use Node.js 22.x
+              uses: actions/setup-node@v6
+              with:
+                  node-version: 22.x
+                  cache: 'pnpm'
+
+            - name: Install pnpm modules
+              run: pnpm install --frozen-lockfile
+
+            - name: Run update script (fixtures & fallbacks)
+              run: node scripts/update-ui5manifest-version.js
+
+            - name: Check for changes
+              id: check-changes
+              run: |
+                  if [[ -n $(git status --porcelain) ]]; then
+                    echo "has_changes=true" >> $GITHUB_OUTPUT
+                  else
+                    echo "has_changes=false" >> $GITHUB_OUTPUT
+                  fi
+
+            - name: Commit fixture updates
+              if: steps.check-changes.outputs.has_changes == 'true'
+              run: |
+                  git config user.name "github-actions[bot]"
+                  git config user.email "github-actions[bot]@users.noreply.github.com"
+                  git add -u || true # Add updated files, but don't fail if no files were changed
+                  git commit -m "chore: update manifest version test fixtures & ui5 version fallbacks"
+                  git push
+
+            - name: Create changeset
+              uses: mscharley/dependency-changesets-action@v1.2.2
+              with:
+                  author-email: 'github-actions[bot]@users.noreply.github.com'
+                  commit-message: 'chore: add changeset for @ui5/manifest update'
--- a/.gitignore
+++ b/.gitignore
@@ -179,3 +179,17 @@ generators
 # nx 
 .nx
 nx-cloud.env
+
+# act - Local GitHub Actions runner
+.act
+
+# AI Coding Agents
+.claude/settings.local.json
+CLAUDE.local.md
+.cursorignore
+.codeiumignore
+.aiignore
+
+# Local Model/Agent Caches
+.anthropic/
+.claude/cache/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,665 @@
+# AI Agent Guidelines for SAP UX Tools
+
+This document provides essential guidelines for AI-powered tools working with the SAP UX Tools monorepo. Following these guidelines ensures code quality, maintainability, and consistency across the project.
+
+## Table of Contents
+
+- [Overview](#overview)
+- [Critical Requirements](#critical-requirements)
+- [Quality Gates](#quality-gates)
+- [Development Standards](#development-standards)
+- [Package Management](#package-management)
+- [Testing Requirements](#testing-requirements)
+- [Security and Maintenance](#security-and-maintenance)
+- [Change Management](#change-management)
+- [Useful Commands](#useful-commands)
+
+## Overview
+
+The SAP UX Tools is a monorepo for building SAP Fiori applications. All packages of this monorepo can be found in the `/packages` folder. The project uses:
+
+- **Package Manager**: pnpm (see root [package.json](package.json) for required version)
+- **Node Version**: >=20.x (see root [package.json](package.json) engines field)
+- **Build System**: Nx for orchestration
+- **Testing**: Jest for unit tests, Playwright for integration tests
+- **Versioning**: Changesets for automated version management
+- **Publishing**: Automated via GitHub Actions to npmjs.com
+
+## Critical Requirements
+
+### 0. All Quality Gates Must Pass
+
+Before submitting any changes, ensure all the following commands complete successfully:
+
+```bash
+pnpm install
+pnpm build
+pnpm lint
+pnpm test
+```
+
+**IMPORTANT**: All four commands must pass without errors. This is non-negotiable for all contributions.
+
+### 1. Package Version Alignment
+
+All packages across the pnpm workspace must use **the same version** of shared dependencies. This ensures consistency and prevents version conflicts.
+
+**Check version alignment:**
+```bash
+pnpm lint:dependency-versions
+```
+
+This uses `check-dependency-version-consistency` to validate that all packages use identical versions of shared dependencies.
+
+**Common shared dependencies to align:**
+- TypeScript
+- Jest and testing utilities
+- ESLint and related plugins
+- Build tools (esbuild, rimraf, etc.)
+- UI libraries (React, React-DOM)
+
+**Rule**: When updating a dependency in one package, update it in ALL packages that use it.
+
+### 2. Dependency Version Freshness
+
+Dependencies should **not be older than 6 months** to ensure security, bug fixes, and compatibility with the ecosystem.
+
+**Guidelines:**
+- Regularly review and update dependencies
+- Check npm registry for latest stable versions
+- Review changelogs for breaking changes before upgrading
+- Test thoroughly after version bumps
+- Use `pnpm overrides` (in root package.json) for security patches when necessary
+
+**Check dependency ages:**
+```bash
+# Check outdated dependencies
+pnpm outdated
+
+# Check specific package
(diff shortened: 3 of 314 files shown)


.changeset/brave-apes-draw.md
```diff
@@ -0,0 +1,11 @@
...
+"@sap-ux/app-config-writer": patch
+"@sap-ux/create": patch
+"@sap-ux/preview-middleware": patch
+---
+
+feat: enable card generator support for CAP projects
+
+- Add default FLP configuration when preview middleware exists but FLP config is missing
+- Support path derivation for CAP projects when yaml path contains directory separators
+- Use getSourcePath() for correct webapp path resolution in card generator handlers
```
heimwege: Detailed info not needed for changelog

```suggestion
feat: enable card generator support for CAP projects
```

packages/app-config-writer/src/cards-config/index.ts
```diff
@@ -47,6 +47,19 @@ async function updateMiddlewareConfigWithGeneratorPath(
         previewMiddleware.configuration ??= {};
 
         const middlewareConfig = previewMiddleware.configuration as PreviewConfig;
+
+        // Add flp configuration if not present (required for sandbox to work)
```
heimwege: This is absolutely not needed and is not required at all! Please have a look at the default config option for property `flp.path` and `flp.intent` at https://github.com/SAP/open-ux-tools/tree/main/packages/preview-middleware#configuration-options. You will notice that `test/flp.html` and `#app-preview` are the defaults in case no other configuration is defined in the yaml

packages/app-config-writer/test/unit/cards-config/index.test.ts
```diff
@@ -16,6 +17,15 @@ function createTestFs(basePath: string) {
     return fs;
 }
 
+function createMockLogger(): ToolsLogger {
```
heimwege: First of all if you type cast to `as Something` you don't need to define the type again for the property `: Something`. Second, why do you want to create multiple instances of a logger mock? Please compare to how this is used e.g. in https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/test/unit/base/flp.test.ts#L74

packages/app-config-writer/test/unit/cards-config/index.test.ts
```diff
@@ -95,4 +105,42 @@ describe('enableCardGenerator', () => {
         expect(fs.read(join(basePath, 'package.json'))).toMatchSnapshot();
         expect(fs.read(join(basePath, 'ui5-with-deprecated-config-and-cards-generator.yaml'))).toMatchSnapshot();
     });
+
+    test('Valid LROP without preview middleware - should add preview middleware with flp configuration', async () => {
+        const basePath = join(__dirname, '../../fixtures/cards-config/lrop-v4');
+        const fs = create(createStorage());
```
heimwege, line 102: Please create a beforeEach function and refactor so that the identical parts are done in the beforeEach, like preparing the logger mock and creating a fresh mem fs

packages/app-config-writer/test/unit/cards-config/index.test.ts
```diff
@@ -95,4 +105,42 @@ describe('enableCardGenerator', () => {
...
+        const logger = createMockLogger();
+        await enableCardGeneratorConfig(basePath, join(basePath, 'ui5-without-generator.yaml'), logger, fs);
+
+        if (process.env.UX_DEBUG) {
+            fs.commit(() => {});
+        }
```
heimwege, line 143: Why is this needed in a test case?

packages/app-config-writer/test/unit/cards-config/index.test.ts
```diff
@@ -95,4 +105,42 @@ describe('enableCardGenerator', () => {
...
+
+        const yamlContent = fs.read(join(basePath, 'ui5-without-generator.yaml'));
+        expect(yamlContent).toMatchSnapshot();
+        // When no preview middleware exists, it should be added with flp config
+        expect(yamlContent).toContain('flp:');
+        expect(yamlContent).toContain('path: /test/flp.html');
+        expect(yamlContent).toContain('intent:');
+        expect(yamlContent).toContain('object: app');
+        expect(yamlContent).toContain('action: preview');
```
heimwege: What is this expect section supposed to do? Why all these string compares? You might rather want to create and compare a snapshot?

Besides that: Writing the default into the config is not a good idea. So actually I would rather expect a test case to ensure seeing default values in a yaml file should lead to a failing test.

packages/app-config-writer/test/unit/cards-config/index.test.ts
```diff
@@ -95,4 +105,42 @@ describe('enableCardGenerator', () => {
...
+
+        if (process.env.UX_DEBUG) {
+            fs.commit(() => {});
+        }
+
+        const yamlContent = fs.read(join(basePath, 'ui5.yaml'));
+        expect(yamlContent).toMatchSnapshot();
+        // Should preserve existing flp path (may be normalized with leading /)
+        expect(yamlContent).toContain('flpSandbox.html');
+        // Should not have called debug for adding flp config since it already exists
```
heimwege: First you compare the whole yaml content to a snapshot and if it is ok you compare a part of the already verified yaml to yet another sting value. That does not make sense.

packages/create/src/cli/add/cards-generator.ts
```diff
@@ -37,10 +38,23 @@ Example:
...
     try {
-        logger.debug(`Called add cards-generator-config for path '${basePath}', simulate is '${simulate}'`);
-        await validateBasePath(basePath);
+        // If yamlPath contains a directory path (e.g., app/travel_processor/ui5.yaml),
+        // derive the basePath from it to support running from CAP project root
+        let effectiveBasePath = basePath;
+        let effectiveYamlPath = yamlPath;
 
-        const fs = await enableCardGeneratorConfig(basePath, yamlPath, logger);
+        if (yamlPath && yamlPath.includes('/')) {
```
heimwege: Just because a path includes / doesn't make it absolute and depending on OS it's / or \. So that yamlPath.includes('/') is wrong.
Also basePath in create is always absolute.
You can use validateBasePath to validate the basePath and yamlPath.
Besides that I think you can revert that change (or explain why it is needed and that I didn't get it 😸)

packages/create/test/unit/cli/add/cards-generator.test.ts
```diff
@@ -42,6 +42,60 @@ describe('add/cards-generator', () => {
         jest.clearAllMocks();
     });
 
+    describe('path derivation for CAP projects', () => {
+        test('add cards-generator with relative yaml path containing directory - verifies path derivation logic', async () => {
```
heimwege: According to the comment in add/cards-generator.ts I think this test is not needed because the feature it tests is not needed.

packages/create/test/unit/cli/add/cards-generator.test.ts
```diff
@@ -42,6 +42,60 @@ describe('add/cards-generator', () => {
...
+            expect(enableCardGeneratorConfigMock).toHaveBeenCalled();
+            const [effectiveBasePath, effectiveYamlPath] = enableCardGeneratorConfigMock.mock.calls[0];
+
+            // The effective base path should remain the original app root
+            expect(effectiveBasePath).toBe(appRoot);
+            // The effective yaml path should be the same as provided
+            expect(effectiveYamlPath).toBe('ui5.yaml');
+        });
+
+        test('add cards-generator with nested relative yaml path - verifies path derivation logic', async () => {
```
heimwege: Having a look at this test I get the feeling there is a general misunderstanding in the usage of the `create` command. The `create` command is to be used for a specific app (e.g. `c:/myCapProject/app/myApp`), not for the CAP root (e.g. `c:/myCapProject`) and thus multiple apps.
Also because I enable card generator for `<CAP root>/apps/myApp` doesn't mean I want to enable it for my other app `<CAP root>/apps/myOtherAppThatShouldNotHAveCardsGeneratorEnabled` in the project as well. Why do you want to have either all apps or none for CAP?

packages/create/test/unit/cli/add/cards-generator.test.ts
```diff
@@ -76,4 +130,27 @@ describe('add/cards-generator', () => {
         expect(enableCardGeneratorConfigMock).toHaveBeenCalled();
         expect(traceSpy).toHaveBeenCalled();
     });
+
+    test('add cards-generator with verbose flag', async () => {
+        // Test execution
+        const command = new Command('add');
+        addCardsEditorConfigCommand(command);
+        await command.parseAsync(testArgv(['--verbose']));
```
heimwege: Where do you test the `verbose` logging feature in this test?

packages/create/test/unit/cli/add/cards-generator.test.ts
```diff
@@ -42,6 +42,60 @@ describe('add/cards-generator', () => {
...
+            // The effective base path should be capRoot + app/travel_processor
+            expect(expectedEffectiveBasePath).toBe(join(capRoot, 'app', 'travel_processor'));
+            expect(expectedEffectiveYamlPath).toBe('ui5.yaml');
+        });
+
+        test('add cards-generator with simple yaml filename (no directory)', async () => {
+            // Test execution
+            const command = new Command('add');
+            addCardsEditorConfigCommand(command);
+            await command.parseAsync(testArgv(['--config', 'ui5.yaml']));
```
heimwege: Why passing the default parameter for the ui5 yaml file name `ui5.yaml` to the function call?
To be that test is the same as `add cards-generator with custom config path` 🤷🏻

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1025,7 +1025,9 @@ export class FlpSandbox {
                 fileName?: string;
                 manifests: MultiCardsPayload[];
             };
-            const webappPath = await getWebappPath(path.resolve(), this.fs);
+            // getSourcePath() returns the webapp path, we need the app root (parent of webapp)
+            const webappPath = this.utils.getProject().getSourcePath();
```
heimwege: Why deleting `const webappPath = await getWebappPath(path.resolve(), this.fs);` ?

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1025,7 +1025,9 @@ export class FlpSandbox {
...
+            const projectPath = dirname(webappPath);
```
heimwege: Actually the nested ui5 server that serves this application is already at root `/` just because `cds-plugin-ui5` adds a `baseUrl` you are able to navigate one level up. Why do you need to navigate to CAP root (root of the outer server triggered by `cds watch`)?

Do you know that the structure `<cap root>/app` is not fix but can be customized in cap config? So it could also be `<cap root>/hello/world/life/is/beautiful`.

In case you want the cap root I'd suggest to use the `await findProjectRoot(process.cwd(), false, true);` which we already call in the very first line of the `init` function.

But again, please explain why you need the cap root here.

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1078,7 +1080,8 @@ export class FlpSandbox {
      */
     private async storeI18nKeysHandler(req: Request, res: Response): Promise<void> {
         try {
-            const webappPath = await getWebappPath(path.resolve(), this.fs);
+            // getSourcePath() returns the webapp path directly
```
heimwege: Why this change?

heimwege (changes requested):
See comments.

Initially you created #3719. Then you created #4091. Both look completely different. Now you created this PR which again looks completely different than the changes done in the previous two PRs. Yet it is supposed to do the very same thing (enable cards generator for CAP projects).
Can you please explain what you actually want to achieve and why we have now three open PRs (not counting #4092 and #4093)?
And can you please clean up and close the PRs that you don't want to merge?

.changeset/brave-apes-draw.md
Singham-24: done

packages/app-config-writer/src/cards-config/index.ts
Singham-24: You're right! I wasn't aware that `test/flp.html` and `#app-preview` are already the defaults when no configuration is defined. I'll remove the code.

packages/app-config-writer/test/unit/cards-config/index.test.ts
Singham-24: I will :
- Removed the redundant type annotation when using `as` cast
- Create a single logger mock instance and reuse it across tests, similar to how it's done in `flp.test.ts`

packages/app-config-writer/test/unit/cards-config/index.test.ts
Singham-24: done

packages/app-config-writer/test/unit/cards-config/index.test.ts
Singham-24: removed

packages/app-config-writer/test/unit/cards-config/index.test.ts
Singham-24: done

packages/create/src/cli/add/cards-generator.ts
Singham-24: removed

packages/create/test/unit/cli/add/cards-generator.test.ts
Singham-24: removed

packages/app-config-writer/test/unit/cards-config/index.test.ts
Singham-24: removed

packages/create/test/unit/cli/add/cards-generator.test.ts
Singham-24: removed

packages/create/test/unit/cli/add/cards-generator.test.ts
Singham-24: removed

packages/preview-middleware/src/base/flp.ts

packages/preview-middleware/src/base/flp.ts
Singham-24: removed

packages/preview-middleware/src/base/flp.ts
Singham-24: removed

packages/create/test/unit/cli/add/cards-generator.test.ts
Singham-24: removed

Issue #4124, opened 2026-02-06
BUG - destination instances are referring to resource name instead of resource service-name


longieirl:
### Related Feature
Feature request: _issue number_

### Description
In some instances, CAP Projects are created without an mta.yaml. When a SAP Fiori application is appended to the CAP project and user the `npm run deploy-config` is executed, it detects there is no mta.yaml present and prompts the user to create one.

During the generation, the `mta.yaml` resources are generated without a service name, for exampe;
```yaml
- name: fe_lrop_cap_no_mta_managed-html5-repo-host
  type: org.cloudfoundry.managed-service
  parameters:
    service: html5-apps-repo
    service-plan: app-host
```

When the cf-writer does a cleanup, it appends `service-name` as shown;
```
- name: fe_lrop_cap_no_mta_managed-html5-repo-host
  type: org.cloudfoundry.managed-service
  parameters:
    service: html5-apps-repo
    service-name: fe_lrop_cap_no_mta_managed-html5-service
    service-plan: app-host
```

This causes the destinations `ServiceInstanceName` to be misaligned;
```
    parameters:
      content:
        instance:
          destinations:
            - Name: fe_lrop_cap_no_mta_managed-html5-repository
              ServiceInstanceName: fe_lrop_cap_no_mta_managed-html5-repo-host
              ServiceKeyName: fe_lrop_cap_no_mta_managed-html5-repo-host-key
              sap.cloud.service: fe_lrop_cap_no_mta_managed.service
            - Authentication: OAuth2UserTokenExchange
              Name: fe_lrop_cap_no_mta_managed-auth
              ServiceInstanceName: fe_lrop_cap_no_mta_managed-auth
              ServiceKeyName: fe_lrop_cap_no_mta_managed-auth-key
              sap.cloud.service: fe_lrop_cap_no_mta_managed.service
          existing_destinations_policy: update
```

Since the `ServiceInstanceName` is using the name of the `resource`, for example `fe_lrop_cap_no_mta_managed-html5-repo-host` but if the `resource` has `service-name` present, then this value needs to be used in the destination.

### Steps to Reproduce
1. Generate a CAP project using;
```
cds init managedProject && cd managedProject && npm install && cds add tiny-sample hana xsuaa connectivity destination workzone && cds build && npm update --package-lock-only
```
2. Append a UI5 app
3. Open App Info page and add deployment config

### Expected results

[...]

PR #4133, opened 2026-02-09
fix: General review of code and cleanup

longieirl:
Fix is part of https://github.com/SAP/open-ux-tools/issues/4124, found during testing.


devinea (approved):
Nice clean-up 
changeset 👍

PR #4137, opened 2026-02-10
Refactor eslint config to use mjs extension

devinea:
Convert eslint configs templates for Fiori projects to `.mjs` extension and convert content to esm.

better support for future.


heimwege (approved):
- changing to `mjs` instead of `js` looks good.
- changeset ok

Thanks @devinea

PR #4138, opened 2026-02-10
fix(security): upgrade axios

donal-tobin-sap:
- specification will not upgrade until next sprint
- 2 other modules will upgrade at a later stage

Issue: #37039
Issue: https://github.com/SAP/open-ux-tools/security/dependabot/364


longieirl (dismissed):
Changeset looks good.
Change is good.

donal-tobin-sap:

- High priority fix required for tools-suite

PR #4140, opened 2026-02-11
fix: `loadGlobalCdsModule` failing to load global CDS module for CDS versions >= 9.7.0

815are:
`loadGlobalCdsModule` failing to load global CDS module for CDS versions >= 9.7.0
Reason is because `cds --version` output was changed:
For cds >= 9.7.0
image

for cds < 9.7.0
image

This PR has proposal fix to adjust method `getCdsVersionInfo`, which handles output of `cds --version` - https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/project/cap.ts#L687

### DIFF
--- a/packages/project-access/src/project/cap.ts
+++ b/packages/project-access/src/project/cap.ts
@@ -690,7 +690,7 @@ async function getCdsEnvData(cwd?: string): Promise<Record<string, string>> {
                 try {
                     resolve(JSON.parse(out));
                 } catch (e) {
-                    reject(new Error('Unexpected output of "cds env --json"'));
+                    reject(new Error(`Unexpected output of "cds env --json": ${e.message}`));
                 }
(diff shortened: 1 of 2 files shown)


packages/project-access/src/project/cap.ts
Klaus-Keller: Hi @815are, could we have a look into `cds env --json` to get the `home`?

815are: Hi @Klaus-Keller 

suggested approach with `cds env --json` is applied with commit - 411e1f0c8ae44a67c2544041e79eb57849f3e27d

Klaus-Keller (approved):
Thanks @815are,

* new way to retrieve CDS home looks good
* changeset exists
* tests are adjusted

Approved from my side.

PR #4143, opened 2026-02-12
feat(sap-systems-ext): support adding full service urls as a new connection type

cianmSAP:
Internal ticket 36747

- Adds new prompt for `Connection Type` : `ABAP Catalog` / `Service URL endpoint`

image

- Updates `@sap-ux/store` module to allow filtering on multiple values e.g. for connection type you can pass an array `connectionType: ['abap_catalog', 'odata_service']`. By default `getAll` will only return systems that have `abap_catalog` connection type

### DIFF
--- a/packages/store/README.md
+++ b/packages/store/README.md
@@ -88,3 +88,107 @@ interface EntityKey<T> {
     getId: () => string;
 }
 ```
+
+#### Special handling for BackendSystem getAll()
+
+The `BackendSystem` service supports filtering and retrieval options when calling `getAll()`. This allows you to retrieve only the backend systems that match specific criteria.
+
+##### BackendServiceRetrievalOptions
+
+The `getAll()` method accepts an optional `BackendServiceRetrievalOptions` parameter with the following properties:
+
+- `includeSensitiveData?: boolean` - Whether to include sensitive data (username, password, etc.) in the returned systems
+- `backendSystemFilter?: BackendSystemFilter` - Filter criteria to match specific backend systems
+
+##### BackendSystemFilter
+
+The `BackendSystemFilter` allows filtering by any serializable property of a `BackendSystem`. The following properties can be used for filtering:
+
+- `name` - System name
+- `url` - System URL
+- `client` - SAP client number
+- `systemType` - Type of system (e.g., `SystemType.AbapOnPrem`, `SystemType.AbapCloud`, `SystemType.Generic`)
+- `authenticationType` - Authentication method (e.g., `AuthenticationType.Basic`, `AuthenticationType.OAuth2RefreshToken`)
+- `connectionType` - Connection type (e.g., `'abap_catalog'`, `'odata_service'`, `'generic_host'`)
+- `systemInfo` - Nested object with `systemId` and `client` properties
+
+##### Filter Value Types
+
+Each filter property supports three types of values:
+
+1. **Single value** - Match systems where the property equals the specified value
+2. **Array of values** - Match systems where the property equals any value in the array
+3. **Object value** (for `systemInfo`) - Match systems where nested properties match
+
+##### Usage Examples
+
+**Filter by single value:**
+```typescript
+import { getService } from '@sap-ux/store';
+import { SystemType } from '@sap-ux/store/types';
+
+const systemService = await getService('system');
+
+// Get only OnPrem systems
+const onPremSystems = await systemService.getAll({
+    backendSystemFilter: { systemType: SystemType.AbapOnPrem }
+});
+```
+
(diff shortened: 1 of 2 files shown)


packages/sap-systems-ext-webapp/src/i18n/i18n.json
```diff
@@ -20,7 +20,14 @@
...
         "onPremSystem": "ABAP On Premise",
-        "btpSystem": "ABAP Environment on SAP Business Technology Platform"
+        "btpSystem": "ABAP Environment on SAP Business Technology Platform",
+        "genericHost": "Generic Host",
+        "catalog": "Catalog",
+        "serviceUrl": "Service URL"
+    },
+    "options": {
+        "abapCatalog": "ABAP Catalog",
+        "serviceUrlEndpoint": "Service URL endpoint"
```
lfindlaysap:
```suggestion
        "serviceUrlEndpoint": "Service URL Endpoint"
```

cianmSAP: Updated, thanks

packages/sap-systems-ext-webapp/src/components/layout/main/systemInfo/ConnectionTypes.tsx
```diff
@@ -0,0 +1,41 @@
...
+    const { t } = useTranslation();
+
+    const connectionTypeOptions: { key: ConnectionType; text: string }[] = [
+        { key: 'abap_catalog', text: t('options.abapCatalog') },
+        { key: 'odata_service', text: t('options.serviceUrlEndpoint') }
+    ];
+
+    return (
+        <div className="store-text-field">
+            <label className="store-detail-label">{'Connection Type'}</label>
```
mikicvi-SAP: Should this be a translation instead?

packages/sap-systems-ext-webapp/src/components/layout/main/systemInfo/ConnectionTypes.tsx
```diff
@@ -0,0 +1,41 @@
...
+            <UIChoiceGroup
+                name="connectionType"
+                id="connType"
+                options={connectionTypeOptions}
+                defaultSelectedKey={'abap_catalog'}
+                inline={true}
+                onChange={(e, option) => {
+                    setConnectionType?.(option?.key as ConnectionType);
+                }}
```
mikicvi-SAP: If `option?.key is undefined`, I wonder if it would still pass undefined to the setter as ConnectionType? 

```suggestion
  onChange={(e, option) => {
      if (option?.key) {
          setConnectionType?.(option.key as ConnectionType);
      }
  }}
```

packages/sap-systems-ext-webapp/src/components/layout/main/systemInfo/ConnectionTypes.tsx
```diff
@@ -0,0 +1,41 @@
...
+    return (
+        <div className="store-text-field">
+            <label className="store-detail-label">{'Connection Type'}</label>
+            <UIChoiceGroup
+                name="connectionType"
+                id="connType"
+                options={connectionTypeOptions}
+                defaultSelectedKey={'abap_catalog'}
+                inline={true}
+                onChange={(e, option) => {
```
mikicvi-SAP, line 35:
```suggestion
  onChange={(e, option) => {
      if (option?.key) {
          setConnectionType?.(option.key as ConnectionType);
      }
  }}
```

If `option?.key` is `undefined`, I wonder if it would still pass undefined to the setter as ConnectionType?

mikicvi-SAP (changes requested):
Looks really good, 2 small comments 
Will test manually tomorrow 👍🏻

packages/sap-systems-ext-webapp/src/components/layout/main/systemInfo/ConnectionTypes.tsx
cianmSAP: Thanks, should have had a todo on this

packages/sap-systems-ext-webapp/src/components/layout/main/systemInfo/ConnectionTypes.tsx
cianmSAP: I think we're covered here, the options are hardcoded values.
However it did lead me to realise that `systemInfo.connectionType` was left undefined in the event that the user just accepts the default selected option. So I've updated the `useEffect` in `useSystemMain.ts` to default to `abap_catalog` for the connectionType

PR #4149, opened 2026-02-13
feat(cap): adds option to prevent root package json updates for cap

cianmSAP:
Adds option to prevent (some) updates of the root package.json for CAP projects i.e. adding `sapux` and `cds watch` scripts

 This option can be enabled by setting the feature toggle `sap.ux.testBetaFeatures.disableCapRootPkgJsonUpdates`

### DIFF
--- a/packages/app-config-writer/CHANGELOG.md
+++ b/packages/app-config-writer/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/app-config-writer
 
+## 0.6.98
+
+### Patch Changes
+
+- edc7b52: feat: add new command to add eslint config to a project
+
 ## 0.6.97
 
 ### Patch Changes
--- a/packages/app-config-writer/package.json
+++ b/packages/app-config-writer/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/app-config-writer",
     "description": "Add or update configuration for SAP Fiori tools application",
-    "version": "0.6.97",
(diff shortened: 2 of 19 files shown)


packages/cap-config-writer/src/cap-writer/updates.ts
```diff
@@ -29,11 +31,19 @@ export async function applyCAPUpdates(
...
         sapux = false,
         enableCdsUi5Plugin = true,
-        enableTypescript = false
+        enableTypescript = false,
+        disableRootPackageJsonUpdates = false
     } = capProjectSettings;
 
-    // update root package.json
-    await updateRootPackageJson(fs, packageName, sapux, capService, appId, enableCdsUi5Plugin);
+    if (disableRootPackageJsonUpdates) {
```
longieirl, line 38: Can the two if statements not be combined?

cianmSAP: They can be, I don't think it makes it more understandable though, and there's no real performance improvement.

If I was to read `disableRootPackageJsonUpdates && enableCdsUi5Plugin` it seems like they are on the same logical level. Whereas with nesting, I am making it obvious that despite  `disableRootPackageJsonUpdates` being true, we will still add cds ui5 plugin

longieirl: Once there is an explanation, I'm happy.

PR #4155, opened 2026-02-16
Feat/create/convert eslint config

heimwege:
Add a new command `create convert eslint-config` that converts an existing eslint config of a project that is below eslint version 9 to eslint version 9. It also adds/upgrades `@sap-ux/eslint-plugin-fiori-tools`.

Additionally
- adds `skipInstall` option to `add eslint-config`
- refactors require.resolve to get eslint mjs template in `add eslint-config`
- fixed create README generator to only show 'required' for required command options (no longer for optional command options with required values).


packages/app-config-writer/src/eslint-config/convert.ts
```diff
@@ -0,0 +1,197 @@
...
+};
+
+const packageName = {
+    ESLINT: 'eslint',
+    ESLINT_MIGRATE_CONFIG: '@eslint/migrate-config',
+    ESLINT_PLUGIN_FIORI_TOOLS: '@sap-ux/eslint-plugin-fiori-tools',
+    ESLINT_PLUGIN_FIORI_CUSTOM: 'eslint-plugin-fiori-custom'
+} as const;
+
+const MIGRATION_ERROR_TEXT = `Migration to eslint version 9 failed. Please check if you can find more details above. You might want to check the official eslint documentation on how to migrate to version 9: https://eslint.org/docs/latest/use/migrate-to-9.0.0#flat-config or delete the existing eslint \`devDependency\` and run \`create add eslint\` to get a flat config eslint.config.mjs where you can transfer your old eslint config manually.`;
```
lfindlaysap:
```suggestion
const MIGRATION_ERROR_TEXT = `Migration to eslint version 9 failed. Check if there are error messages above. You can also delete the existing eslint \`devDependency\` and run \`create add eslint\` to create a eslint.config.mjs file with the flat config where you can transfer your old eslint config manually.` For more information, see [https://eslint.org/docs/latest/use/migrate-to-9.0.0#flat-config](Migrate to v9.x).;
```

packages/app-config-writer/src/eslint-config/convert.ts
```diff
@@ -0,0 +1,197 @@
...
+    }
+    if (!hasDependency(packageJson, 'eslint')) {
+        logger?.error(
+            `Did not find 'esLint' dependency in package.json at path '${packageJsonPath}. You might want to use the \`add eslint-config\` command instead.'`
+        );
+        return false;
+    }
+    if (!isLowerThanMinimalVersion(packageJson, 'eslint', '9.0.0')) {
+        logger?.error(
+            `EsLint version is already 9.0.0 or higher in this project. Found 'eslint' dependency with version '${packageJson.devDependencies?.eslint}' in package.json at path '${packageJsonPath}'`
```
lfindlaysap:
```suggestion
            `ESLint version is already 9.0.0 or higher in this project. Found 'eslint' dependency with version '${packageJson.devDependencies?.eslint}' in package.json at path '${packageJsonPath}'`
```

packages/create/src/cli/convert/eslint-config.ts
```diff
@@ -0,0 +1,63 @@
...
+        await validateBasePath(basePath);
+        const fs = await migrateEslintConfig(basePath, { logger, config });
+        await traceChanges(fs);
+        fs.commit(() => {
+            logger.info(
+                `Eslint configuration converted. Ensure you install the new dependency by executing 'npm install'.`
+            );
+            if (skipInstall) {
+                logger.info(
+                    `\`npm install\` will be skipped. Please make sure to install the dependencies before executing any linting commands.`
```
lfindlaysap:
```suggestion
                    `\`npm install\` was skipped. Ensure you install the dependencies before executing any linting commands.`
```

heimwege: done in https://github.com/SAP/open-ux-tools/pull/4155/commits/29ee9572c7051a60f0fe073b3e8a5c642577f66c

packages/app-config-writer/src/eslint-config/convert.ts
heimwege: https://github.com/SAP/open-ux-tools/pull/4155/commits/29ee9572c7051a60f0fe073b3e8a5c642577f66c

Also I adjustes other occurrences of `eslint` to `ESLint`

packages/app-config-writer/src/eslint-config/convert.ts
heimwege: https://github.com/SAP/open-ux-tools/pull/4155/commits/29ee9572c7051a60f0fe073b3e8a5c642577f66c

packages/create/test/unit/cli/convert/eslint-config.test.ts
```diff
@@ -0,0 +1,208 @@
...
+                "ESlint configuration converted. Ensure you install the new dependency by executing 'npm install'."
+            )
+        );
+        expect(loggerMock.warn).not.toHaveBeenCalled();
+        expect(loggerMock.error).not.toHaveBeenCalled();
+        expect(fsMock.commit).toHaveBeenCalled();
+        expect(runNpmInstallSpy).toHaveBeenCalledWith(appRoot);
+    });
+
+    test('Test create-fiori convert eslint-config <appRoot> --simulate', async () => {
```
marcovieth: We do not have a "--simulate" option for convert, so remove this test case.

packages/create/test/unit/cli/convert/eslint-config.test.ts
```diff
@@ -0,0 +1,208 @@
...
+        // Result check
+        expect(cwdSpy).toHaveBeenCalled();
+        expect(appConfigWriter.convertEslintConfig).toHaveBeenCalledWith(appRoot, expect.any(Object));
+        expect(fsMock.commit).toHaveBeenCalled();
+        expect(runNpmInstallSpy).toHaveBeenCalledWith(appRoot);
+
+        cwdSpy.mockRestore();
+    });
+
+    test('Test create-fiori convert eslint-config with simulate and verbose', async () => {
```
marcovieth: We do not have "simulate"

packages/app-config-writer/src/eslint-config/convert.ts
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * Updates the package.json file of the project by adding or updating the required devDependencies.
+ *
+ * @param basePath - base path to be used for the conversion
+ * @param fs - file system reference
+ */
+async function updatePackageJson(basePath: string, fs: Editor): Promise<void> {
+    const packageJsonPath = join(basePath, FileName.Package);
+    const packageJson = fs.readJSON(packageJsonPath) as Package;
```
marcovieth, line 277: Are we absolutely sure that the migration command does not modify package.json? Otherwise we could read old cached content here.

packages/create/test/unit/cli/convert/eslint-config.test.ts
heimwege: Actually it was just a leftover in the test case name. The test is still valid. I adjusted the name in b5f2ccc3a0281b683086a0b72fd8e5e239f7e468

packages/create/test/unit/cli/convert/eslint-config.test.ts
heimwege: Actually it was just a leftover in the test case name. The test is still valid. I adjusted the name in b5f2ccc3a0281b683086a0b72fd8e5e239f7e468

packages/app-config-writer/src/eslint-config/convert.ts
heimwege: According to https://www.npmjs.com/package/@eslint/migrate-configit does not adjust package.json but you are right maybe we should completely ignore using mem-fs for this command 🤔 @Klaus-Keller what do you think?

heimwege:
@marcovieth let me try to answer your questions

> * The combination "create convert" in the command looks a bit strange to me.

I agree that `create` might not have been the best name for our cli commands (we also already have `create convert preview` 🙈 ) but I'm afraid we can't rename this thing any longer.

> * New packages added to eslint.config.mjs need to be installed manually?
>   (Ok, then we do not need to care about TS projects.)

This command just does the initial conversion from old to new eslint config format (plus adjusting our eslint plugin setup) Anything else you do later is up to you. But maybe I didn't get your question.

> * Do we want to remove the old .eslintrc.json?

I'd rather not remove it because the migrator states "_The generated configuration file isn't guaranteed to work in all cases, but it should get you a lot closer to a working configuration file than manually trying to migrate._". So this file might still be needed if the migration is not 100%

> * Was it considered as too complicated to encapsulate the migration command for mem-fs, maybe by putting the necessary files in a TMP directory and calling it there? This would also allow a "--simulate" option (skip "simulate" discussed in "Consistency").

That's a very interesting point. We did not want to invest too much time here as this is rather a niche use case, but I'm open for discussion. @Klaus-Keller your thoughts?

packages/create/test/unit/cli/convert/eslint-config.test.ts
marcovieth: @heimwege Isn't the second test case the same as the first one, just with less checks?

marcovieth:
@heimwege: Thank you for your comments.
I agree to the points.
- To encapsulate the migration step, the AI suggested a modification to use a TMP directory with ~40 lines more code. But maybe that's not enough if we also need to deal with absolute paths and symbolic links.

- Please check also my comment to the second test case in eslint-config.test.ts.

marcovieth:
BTW, for testing, I used the following command in open-ux-tools on a minimal example in directory test-eslint-convert ([test-eslint-convert.zip](https://github.com/user-attachments/files/25543567/test-eslint-convert.zip)):

`node packages/create/dist/index.js convert eslint-config test-eslint-convert --verbose --skip-install`

heimwege:
> BTW, for testing, I used the following command in open-ux-tools on a minimal example in directory test-eslint-convert ([test-eslint-convert.zip](https://github.com/user-attachments/files/25543567/test-eslint-convert.zip)):
> 
> `node packages/create/dist/index.js convert eslint-config test-eslint-convert --verbose --skip-install`

@marcovieth is this what you had in mind for the simulate? 951825767e0737663afd681fd7973b101558a2d3

packages/create/test/unit/cli/convert/eslint-config.test.ts
heimwege: I consolidated the tests

packages/app-config-writer/src/common/package-json.ts
```diff
@@ -218,3 +219,38 @@ export function getTestPathForUi5TestRunner(scriptName: string): string | undefi
...
+ * @param mandatory - (default true) if the existence of the dependency is mandatory
+ * @returns indicator if the version is lower than the minimal version
+ */
+export function isLowerThanMinimalVersion(
+    packageJson: Package,
+    dependencyName: string,
+    minVersionInfo: string,
+    mandatory: boolean = true
+): boolean {
+    let versionInfo = packageJson?.devDependencies?.[dependencyName] ?? packageJson?.dependencies?.[dependencyName];
```
marufrasully, line 238: `packageJson` is mandatory param, do we need `?` e.g. `packageJson?.devDependencies`

packages/app-config-writer/src/common/package-json.ts
```diff
@@ -218,3 +219,38 @@ export function getTestPathForUi5TestRunner(scriptName: string): string | undefi
...
+ * @param dependencyName - the name of the (dev)dependency to check
+ * @param minVersionInfo - the minimal version to check against
+ * @param mandatory - (default true) if the existence of the dependency is mandatory
+ * @returns indicator if the version is lower than the minimal version
+ */
+export function isLowerThanMinimalVersion(
+    packageJson: Package,
+    dependencyName: string,
+    minVersionInfo: string,
+    mandatory: boolean = true
```
marufrasully, line 236: `mandatory` not sure if I understand the use case of this optional param. 

if `mandatory` is false, why we need to return `false` another word, checking does not make sense or?
if `mandatory` is true, simply return `true` and we could save this param?

packages/app-config-writer/src/eslint-config/add.ts
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Adds the eslint configuration file to the project.
+ *
+ * @param basePath - base path to be used for the conversion
+ * @param fs - file system reference
+ * @param config - the name of the SAP Fiori tools eslint plugin config to be used
+ */
+async function addEslintConfig(basePath: string, fs: Editor, config: string = 'recommended'): Promise<void> {
+    await addEslintFeature(basePath, fs);
+    if (config === 'recommended-for-s4hana') {
```
marufrasully, line 69: `config` param is because of template I assume. can we have two eslint templates? e.g. `recommended-eslint.config.mjs` and `recommended-for-s4hana-eslint.config.mjs`? With that we just do not need to overwrite content.

marufrasully (commented):
My initial comments

packages/app-config-writer/src/common/package-json.ts
heimwege: Yes this is needed because the existence of `package.json` does not mean it has a property `devDependencies`. If you do `packageJson.devDependencies[dependencyName]` in that case it will crash because undefined is not an array. `undefined[dependencyName]` -> error

packages/app-config-writer/src/common/package-json.ts
heimwege: This function is re-used from the preview-config command. And based on the mandatory we decide to fail or not see comment
`// In case no dependency is found we assume the minimal version is not met depending on the mandatory flag`

A dependency can be not mandatory and not existing -> ok
It can be mandatory and not existing -> not ok

but if it exists and is not mandatory it's version could be too low. -> not ok
and mandatory and existing with a lower version -> not ok
 etc.

packages/app-config-writer/src/eslint-config/add.ts
heimwege: Sadly no. There exists only one template in the generator (for `recommended`) and they use exactly the same logic to override the template.

packages/app-config-writer/src/eslint-config/convert.ts
```diff
@@ -0,0 +1,238 @@
...
+async function checkPrerequisites(basePath: string, fs: Editor, logger?: ToolsLogger): Promise<boolean> {
+    const packageJsonPath = join(basePath, FileName.Package);
+    const packageJson = fs.readJSON(packageJsonPath) as Package | undefined;
+    if (!packageJson) {
+        logger?.error(`No package.json found at path '${packageJsonPath}'`);
+        return false;
+    }
+    if (!hasDependency(packageJson, 'eslint')) {
+        logger?.error(
+            `Did not find ESLint dependency in package.json at path '${packageJsonPath}. You might want to use the \`add eslint-config\` command instead.'`
```
marcovieth: Missing closing quote: '${packageJsonPath}
There's an additional quote at the end...

heimwege: Very good finding 👍🏻 fixed in f3579ccdc87119cc9427c2e90603c4f98263f8e4

packages/app-config-writer/src/common/package-json.ts
marufrasully: I was referring to this extra `?` e.g `packageJson.devDependencies?.[dependencyName]`

removed `?` after `packageJson`

PR #4156, opened 2026-02-16
fix: update vocabularies

gernotkuhr:
update vocabularies


Jimmy-Joseph19 (approved):
Changes looks good.
Did not test locally
Changeset found.

Klaus-Keller (approved):
Thanks @gernotkuhr, vocabulary update looks good.

PR #4159, opened 2026-02-16
fix(sap-systems): url-validation

tom-sap:
Ads URL validation for sap systems - host input (#36776).

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -19,7 +19,8 @@
         'workarounds:typesNodeVersioning'
     ],
 
-    enabledManagers: ['npm', 'github-actions'],
+    enabledManagers: ['npm', 'github-actions'
+    ],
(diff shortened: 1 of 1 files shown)


mikicvi-SAP (commented):
@tom-sap - we're missing a changeset here, but also weird that changeset bot did not comment on the PR...

I’m not sure this is correct.  The error message appears as soon as a new system flow is opened for on prem and cloud, even without any input. It doesn’t feel right. What do you think @cianmSAP? 

Image

tom-sap:
Thanks @mikicvi-SAP, yes, agree on both.
I will update the PR.

mikicvi-SAP (approved):
Thanks @tom-sap
- Code LG and in an appropriate place
- Changeset appropriate 
- Coverage is good
- Tested manually ✅

Issue #4148, opened 2026-02-13
FEATURE - Change `backend-proxy-middleware-cf` to use `approuter` instead of token exchange


nikmace:
_As a **developer**,
I want to **take the advantages of approuter**,
so that I **use the new middleware in CF ADP**_

### Description

Change backend-proxy-middleware-cf to use approuter instead of token exchange. This will probably reuse the logic of ui5-middleware-approuter: https://github.com/ui5-community/ui5-ecosystem-showcase/tree/main/packages/ui5-middleware-approuter.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a created a CF ADP with the correct structure
**when** I start the Adaptation Editor
**then** the new middleware will start, proxy requests correctly, and resolve all OData requests.

### Tasks
- [ ] Transfer and refactor implementation
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #4160, opened 2026-02-17
feat: Change `backend-proxy-middleware-cf` to use `approuter` instead of token exchange

nikmace:
Feat for #4148.
- Changes the middleware to use `approuter` instead of _token exchange_ (i.e., using services credentials). Starts the `approuter` server that proxies requests to OData using `xs-app.json` routes.

### DIFF
--- a/packages/create/CHANGELOG.md
+++ b/packages/create/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/create
 
+## 0.15.65
+
+### Patch Changes
+
+- Updated dependencies [0153757]
+    - @sap-ux/preview-middleware@0.25.15
+    - @sap-ux/app-config-writer@0.6.132
+
 ## 0.15.64
 
 ### Patch Changes
--- a/packages/create/package.json
+++ b/packages/create/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/create",
     "description": "SAP Fiori tools module to add or remove features",
-    "version": "0.15.64",
+    "version": "0.15.65",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/eslint-plugin-fiori-tools/CHANGELOG.md
+++ b/packages/eslint-plugin-fiori-tools/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/eslint-plugin-fiori-tools
 
+## 9.11.3
+
+### Patch Changes
+
+- 0153757: add es2020 globals
+
 ## 9.11.2
 
 ### Patch Changes
--- a/packages/eslint-plugin-fiori-tools/README.md
+++ b/packages/eslint-plugin-fiori-tools/README.md
@@ -100,71 +100,85 @@ export default [
 
 8. Execute in the project root directory: `npm run lint`. Check the output for errors or warnings and fix them.
 
+## Automatically Migrate from `eslint@8`, `@sap-ux/eslint-plugin-fiori-tools@0.6.x`, or `eslint-plugin-fiori-custom`
+
+To automatically migrate from `eslint@8`, `@sap-ux/eslint-plugin-fiori-tools@0.6.x`, or `eslint-plugin-fiori-custom` to `eslint@9` and `@sap-ux/eslint-plugin-fiori-tools@9` (or higher), execute the following command in the project root directory:
+
+```bash
+npx --yes @sap-ux/create@latest convert eslint-config
+```
+
+For more information about the `@sap-ux/create` command, run
+
+```bash
+npx --yes @sap-ux/create@latest convert eslint-config --help
+```
+
 ## [Rules](#rules)
 
 <div style="overflow-x: auto;">
 
-| Since | Rule | Description | Recommended | Recommended for S/4HANA |
-|:-----:|------|-------------|:-----------:|:-----------------------:|
-| new | [sap-no-data-field-intent-based-navigation](docs/rules/sap-no-data-field-intent-based-navigation.md) | Ensures neither `DataFieldForIntentBasedNavigation` nor `DataFieldWithIntentBasedNavigation` are used in tables or form fields in SAP Fiori elements applications. | | ✅ |
-| 9.10.0 | [sap-condensed-table-layout](docs/rules/sap-condensed-table-layout.md) | Requires `condensedTableLayout` to be enabled when using a grid table, analytical table, or tree table. | | ✅ |
-| 9.9.0 | [sap-strict-uom-filtering](docs/rules/sap-strict-uom-filtering.md) | Ensures that `disableStrictUomFiltering` is not set to `true` in `sap.fe.app` manifest configuration | | ✅ |
-| 9.8.0 | [sap-table-personalization](docs/rules/sap-table-personalization.md) | Ensures that all table `personalization` options are enabled in the OData V4 applications. | | ✅ |
-| 9.7.0 | [sap-anchor-bar-visible](docs/rules/sap-anchor-bar-visible.md) | Anchor Bar Visible should not be set to false in manifest settings for object page headers (except form entry object pages). | | ✅ |
-| 9.5.0 | [sap-table-column-vertical-alignment](docs/rules/sap-table-column-vertical-alignment.md) | Ensures `tableColumnVerticalAlignment` Configuration for Responsive Type Tables in SAP Fiori Elements applications | | ✅ |
-| 9.4.0 | [sap-enable-export](docs/rules/sap-enable-export.md) | Ensures that the export to Excel functionality in any OData V4 applications tables is available | | ✅ |
-| 9.4.0 | [sap-enable-paste](docs/rules/sap-enable-paste.md) | Ensures that the paste functionality in any OData V4 applications tables is available | | ✅ |
-| 9.3.1 | [sap-state-preservation-mode](docs/rules/sap-state-preservation-mode.md) | Ensures Valid `statePreservationMode` Configuration in SAP Fiori Elements | | ✅ |
-| 9.1.0 | [sap-copy-to-clipboard](docs/rules/sap-copy-to-clipboard.md) | Ensures that the copy functionality in any table is enabled. "Copy" button is shown by default. | | ✅ |
-| 9.1.0 | [sap-creation-mode-for-table](docs/rules/sap-creation-mode-for-table.md) | Validates that the table creation mode (`createMode` in OData V2 and `creationMode` in OData V4) is correctly configured to ensure an optimal user experience when creating new table entries. | | ✅ |
-| 9.1.0 | [sap-flex-enabled](docs/rules/sap-flex-enabled.md) | Ensures that the `flexEnabled` property is set to `true` in the `sap.ui5` section of the `manifest.json` file for applications using UI5 version 1.56 or higher. | | ✅ |
-| 9.1.0 | [sap-width-including-column-header](docs/rules/sap-width-including-column-header.md) | Ensures that small tables (less than six columns) have the `widthIncludingColumnHeader` property set to `true` for better calculation of column width. | | ✅ |
-| 9.0.0 | [sap-bookmark-performance](docs/rules/sap-bookmark-performance.md) | Ensure the correct usage of the auto-refresh interval options for `sap.ushell.ui.footerbar.AddBookmarkButton`. | ✅ | ✅ |
-| 9.0.0 | [sap-browser-api-error](docs/rules/sap-browser-api-error.md) | Detect forbidden usages of `(window.)document` APIs. | | |
-| 9.0.0 | [sap-browser-api-warning](docs/rules/sap-browser-api-warning.md) | Detect warnings for usages of `(window.)document` APIs. | ✅ | ✅ |
-| 9.0.0 | [sap-cross-application-navigation](docs/rules/sap-cross-application-navigation.md) | Do not use a static list of cross-application navigation targets. | ✅ | ✅ |
-| 9.0.0 | [sap-forbidden-window-property](docs/rules/sap-forbidden-window-property.md) | Detect the definition of global properties in the `window` object. | ✅ | ✅ |
-| 9.0.0 | [sap-message-toast](docs/rules/sap-message-toast.md) | Ensure the usage of the correct method options for `sap.m.MessageToast.show`. | ✅ | ✅ |
-| 9.0.0 | [sap-no-absolute-component-path](docs/rules/sap-no-absolute-component-path.md) | Detect the absolute path to the component. | ✅ | ✅ |
-| 9.0.0 | [sap-no-br-on-return](docs/rules/sap-no-br-on-return.md) | Detect the usage of `document.queryCommandSupported` with the `insertBrOnReturn` argument. | ✅ | ✅ |
-| 9.0.0 | [sap-no-commons-usage](docs/rules/sap-no-commons-usage.md) | Detect the usage of `sap.ui.commons` objects. | ✅ | ✅ |
-| 9.0.0 | [sap-no-dom-access](docs/rules/sap-no-dom-access.md) | Detect direct DOM access. Use the jQuery selector instead. | ✅ | ✅ |
-| 9.0.0 | [sap-no-dom-insertion](docs/rules/sap-no-dom-insertion.md) | Detect direct DOM insertion. | ✅ | ✅ |
-| 9.0.0 | [sap-no-dynamic-style-insertion](docs/rules/sap-no-dynamic-style-insertion.md) | Detect the usage of `document.styleSheets` (dynamic style insertion). | ✅ | ✅ |
-| 9.0.0 | [sap-no-element-creation](docs/rules/sap-no-element-creation.md) | Detect direct element creation. | ✅ | ✅ |
-| 9.0.0 | [sap-no-encode-file-service](docs/rules/sap-no-encode-file-service.md) | Detect the usage of `/sap/bc/ui2/encode_file`. | ✅ | ✅ |
-| 9.0.0 | [sap-no-event-prop](docs/rules/sap-no-event-prop.md) | Flag use of private members from `sap.ui.base.Event`. Use `sap-no-ui5base-prop` instead. | | |
-| 9.0.0 | [sap-no-exec-command](docs/rules/sap-no-exec-command.md) | Detect the usage of `execCommand`. | ✅ | ✅ |
-| 9.0.0 | [sap-no-global-define](docs/rules/sap-no-global-define.md) | Detect the definition of global properties in the `window` object. | ✅ | ✅ |
-| 9.0.0 | [sap-no-global-event](docs/rules/sap-no-global-event.md) | Detect the global event handling override. | ✅ | ✅ |
-| 9.0.0 | [sap-no-global-selection](docs/rules/sap-no-global-selection.md) | Detect global selection modification. | ✅ | ✅ |
-| 9.0.0 | [sap-no-global-variable](docs/rules/sap-no-global-variable.md) | Disallow global variable declarations. | ✅ | ✅ |
-| 9.0.0 | [sap-no-hardcoded-color](docs/rules/sap-no-hardcoded-color.md) | Flag use of hardcoded colors. | ✅ | ✅ |
-| 9.0.0 | [sap-no-hardcoded-url](docs/rules/sap-no-hardcoded-url.md) | Flag use of hardcoded (non-relative) URLs. | ✅ | ✅ |
-| 9.0.0 | [sap-no-history-manipulation](docs/rules/sap-no-history-manipulation.md) | Detect warnings for usages of history manipulation APIs. | ✅ | ✅ |
-| 9.0.0 | [sap-no-inner-html-access](docs/rules/sap-no-inner-html-access.md) | Detect access of the `innerHTML` property. | ✅ | ✅ |
-| 9.0.0 | [sap-no-inner-html-write](docs/rules/sap-no-inner-html-write.md) | Detect overriding of `innerHTML`. | ✅ | ✅ |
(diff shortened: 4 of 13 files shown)


packages/backend-proxy-middleware-cf/README.md
```diff
@@ -1,212 +1,112 @@
-[![Changelog](https://img.shields.io/badge/changelog-8A2BE2)](https://github.com/SAP/open-ux-tools/blob/main/packages/backend-proxy-middleware-cf/CHANGELOG.md) [![Github repo](https://img.shields.io/badge/github-repo-blue)](https://github.com/SAP/open-ux-tools/tree/main/packages/backend-proxy-middleware-cf)
```
heimwege, line 1: Please don't delete the shields

packages/backend-proxy-middleware-cf/README.md
```diff
@@ -1,212 +1,112 @@
...
-The `@sap-ux/backend-proxy-middleware-cf` is a [Custom UI5 Server Middleware](https://sap.github.io/ui5-tooling/pages/extensibility/CustomServerMiddleware) for proxying requests to Cloud Foundry destinations with OAuth2 authentication. It supports proxying multiple OData source paths to a single destination URL with automatic OAuth token management.
+UI5 server middleware that uses `@sap/approuter` to make destinations configured in SAP Cloud Foundry or SAP XS Advanced available for local development. Requests to destination routes are proxied to a local approuter instance via `http-proxy-middleware`.
 
 > **⚠️ Experimental**: This middleware is currently experimental and may be subject to breaking changes or even removal in future versions. Use with caution and be prepared to update your configuration or migrate to alternative solutions if needed.
 
-It can be used either with the `ui5 serve` or the `fiori run` commands.
+## Prerequisites
 
-## [Configuration Options](#configuration-options)
+- [@ui5/cli](https://ui5.github.io/cli/) 3.0 or later (specVersion 3.0 in `ui5.yaml`)
```
heimwege: ui5cli v3 is deprecated. v4 is min

packages/backend-proxy-middleware-cf/README.md
```diff
@@ -1,212 +1,112 @@
...
       configuration:
-        backends:
-          - url: https://your-backend-service
-            paths:
-              - /odata/v4/visitorservice
-              - /odata
+        authenticationMethod: "none"
+        debug: true
+        port: 1091
+        xsappJsonPath: "xs-app.json"
```
heimwege: nit-pick: `"xs-app.json"` is actually not a path `./xs-app.json` would e a path

packages/backend-proxy-middleware-cf/README.md
```diff
@@ -1,212 +1,112 @@
...
+        destinations:
+          - name: "backend"
+            url: "https://your-backend.example/path"
 ```
 
-The middleware will:
+2. Place `xs-app.json` at the path you set in `xsappJsonPath` (e.g. project root). Define routes with a `source` regex and `destination` name that matches an entry in `destinations`.
```
heimwege: I think this is dangerous because this middleware is not the only one using `xs-app.json`. The sentence should be rather phrased the other way around: "Point your path to the location of xs-app.json"

packages/backend-proxy-middleware-cf/src/types.ts
```diff
@@ -1,55 +1,135 @@
...
+ * Configuration for the backend-proxy-middleware-cf (approuter-based).
+ * Options are merged with defaults; configuration can be partial.
+ */
+export interface BackendProxyMiddlewareCfConfig {
+    /** Verbose logging */
+    debug?: boolean;
+    /** Port to run the underlying approuter on */
+    port?: number;
+    /** Path to xs-app.json (relative to project root) */
+    xsappJsonPath: string;
```
heimwege: Why is this not optional as well? I'd guess in 99% of the cases the path is `./xs-app.json`

packages/backend-proxy-middleware-cf/src/env/index.ts
```diff
@@ -0,0 +1,93 @@
...
+import type { ToolsLogger } from '@sap-ux/logger';
+import type { AppRouterEnvOptions } from '@sap-ux/adp-tooling';
+import { buildVcapServicesFromResources, getSpaceGuidFromUi5Yaml, getYamlContent } from '@sap-ux/adp-tooling';
+
+import type { EffectiveOptions } from '../types';
+
+/**
+ * Load and parse env options JSON file.
+ *
+ * @param {string} rootPath - Project root path.
```
heimwege: afaik we don't add the types to the JSDoc because it's defined in TypeScript anyway

packages/backend-proxy-middleware-cf/src/env/index.ts
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @param {AppRouterEnvOptions} options - Env options to apply.
+ */
+function applyToProcessEnv(options: AppRouterEnvOptions): void {
+    const envOptions = {
+        ...options,
+        ...(options.destinations ? { destinations: JSON.stringify(options.destinations) } : {}),
+        ...(options.VCAP_SERVICES ? { VCAP_SERVICES: JSON.stringify(options.VCAP_SERVICES) } : {})
+    };
+    process.env = Object.assign(process.env, envOptions);
```
heimwege: Why not spreading here as well? Would be more consistent to the other code

```suggestion
    process.env = { ...process.env, ...envOptions };
```

packages/backend-proxy-middleware-cf/src/extensions/index.ts
```diff
@@ -0,0 +1,102 @@
...
+ * @returns {LoadedExtensions} Loaded extension modules and list of extension routes (paths) they register.
+ */
+export function loadExtensions(
+    rootPath: string,
+    extensions: ApprouterExtension[] | undefined,
+    logger: ToolsLogger
+): LoadedExtensions {
+    const routes: string[] = [];
+    const modules = (extensions ?? [])
+        .map((extension) => toExtensionModule(extension, rootPath, routes, logger))
```
heimwege: `toExtensionModule` states that it converts an extension to an extension module. But in fact it also manipulates the `routes` array. You only understand this when having a look at the code. The JSDoc does not contain this information. So calling this exported function might lead to unwanted side effects. Maybe it would help to separate this into two function, one doing the conversion to ExtensionModule and the other one adjusting the routes?

packages/backend-proxy-middleware-cf/src/proxy/proxy.ts
```diff
@@ -0,0 +1,128 @@
...
+
+import type { CreateProxyOptions, EffectiveOptions, RouteEntry } from '../types';
+import { createPathFilter, getMimeInfo, getRequestOrigin, replaceUrl } from './utils';
+
+/**
+ * Request in proxyReq callback, extended with UI5 server middleware properties.
+ */
+interface ProxyReqRequest extends IncomingMessage {
+    'ui5-middleware-index'?: { url: string };
+    'ui5-patched-router'?: { baseUrl?: string; originalUrl?: string };
```
heimwege, line 17: Maybe add a comment that the `ui5-patched-router` comes from `cds-plugin-ui5`? This knowledge might help during maintenance at a later point in time.

packages/backend-proxy-middleware-cf/src/routes/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    const xsappConfig = JSON.parse(fs.readFileSync(xsappJsonPath, 'utf8')) as XsappConfig;
+
+    if (effectiveOptions.disableWelcomeFile) {
+        delete xsappConfig.welcomeFile;
+    }
+
+    xsappConfig.authenticationMethod = effectiveOptions.authenticationMethod;
+
+    const xsappRoutes = xsappConfig.routes ?? [];
+    /* eslint-disable @typescript-eslint/prefer-nullish-coalescing -- boolean OR for route inclusion */
```
heimwege: Do you really want to disable this for the file starting from here or rather go for disable-next-line and put it two lines below?

packages/backend-proxy-middleware-cf/src/types.ts
nikmace: Well, that is the most important part of this whole thing with the `approuter`. If there is no `xs-app.json` then the `approuter` will fail. In our case, this file can be `.adp/reuse/xs-app.json` or `./dist/xs-app.json` and it `./xs-app.json`. The flow is like this: we create a CF ADP, and we have `./xs-app.json` in the root. We run the builder to clone the original app in the dist, which also has `./dist/xs-app.json`. Then, if there are reuse libraries, the builder clones those and merges all xs-app.json configs into one, which sits in `.adp/reuse/xs-app.json`.  Maybe it makes sense to make it required? What do you think @heimwege

packages/backend-proxy-middleware-cf/README.md
nikmace: Changed in [403fc6f](https://github.com/SAP/open-ux-tools/pull/4160/commits/403fc6f5abe0ef16eee1ab18439fe9590132c432)

packages/backend-proxy-middleware-cf/README.md
nikmace: Changed in [403fc6f](https://github.com/SAP/open-ux-tools/pull/4160/commits/403fc6f5abe0ef16eee1ab18439fe9590132c432)

packages/backend-proxy-middleware-cf/README.md
nikmace: Changed in [403fc6f](https://github.com/SAP/open-ux-tools/pull/4160/commits/403fc6f5abe0ef16eee1ab18439fe9590132c432)

packages/backend-proxy-middleware-cf/README.md
nikmace: Changed in [403fc6f](https://github.com/SAP/open-ux-tools/pull/4160/commits/403fc6f5abe0ef16eee1ab18439fe9590132c432)

packages/backend-proxy-middleware-cf/src/env/index.ts
nikmace: Changed in [403fc6f](https://github.com/SAP/open-ux-tools/pull/4160/commits/403fc6f5abe0ef16eee1ab18439fe9590132c432)

packages/backend-proxy-middleware-cf/src/env/index.ts
nikmace: Changed in [403fc6f](https://github.com/SAP/open-ux-tools/pull/4160/commits/403fc6f5abe0ef16eee1ab18439fe9590132c432)

packages/backend-proxy-middleware-cf/src/extensions/index.ts
nikmace: Changed in [403fc6f](https://github.com/SAP/open-ux-tools/pull/4160/commits/403fc6f5abe0ef16eee1ab18439fe9590132c432)

packages/backend-proxy-middleware-cf/src/proxy/proxy.ts
nikmace: Changed in [403fc6f](https://github.com/SAP/open-ux-tools/pull/4160/commits/403fc6f5abe0ef16eee1ab18439fe9590132c432)

packages/backend-proxy-middleware-cf/src/routes/index.ts
nikmace: Changed in [403fc6f](https://github.com/SAP/open-ux-tools/pull/4160/commits/403fc6f5abe0ef16eee1ab18439fe9590132c432)

packages/backend-proxy-middleware-cf/src/types.ts
heimwege: I might have phrased this wrong. What I meant was "why doesn't this come with a default path that can be overridden, e.g. './xs-app.json'".
Assumed we have a uniform distribution of `.adp/reuse/xs-app.json`, `./dist/xs-app.json` and `./xs-app.json`, at least 33% of the users would not have to manually enter a path.
So the path is of course mandatory (for runtime), but for the user (at designtime) it is optional to enter it. That's why sometimes there exist two types one for the yaml config and one for the actual processing that does not come with optional params because the default must be applied in between (e.g. preview middleware does it like this). There are TS utility types to leverage this (`Partial<myRequiredType>` or `Required<myOptionalType>` depending on which type you defined)

.changeset/modern-canyons-wear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/backend-proxy-middleware-cf': minor
```
heimwege, line 2: missing package `adp-tooling`

packages/backend-proxy-middleware-cf/src/middleware.ts
```diff
@@ -1,43 +1,119 @@
...
+    if (logoutEndpoint) {
+        customRoutes.push(logoutEndpoint);
+    }
+
+    const basUrlTemplate = await fetchBasUrlTemplate(logger);
+
+    let initialized = false;
+    let proxyMiddleware: RequestHandler | null = null;
+    return function lazyApprouterMiddleware(req: Request, res: Response, next: NextFunction): void {
+        if (!initialized) {
```
heimwege, line 96: The lazy `if (!initialized)` block inside the returned request handler runs synchronously on the first request. It calls `startApprouter(...)` which can throw. An unhandled synchronous throw here will crash the UI5 dev server with no friendly error surface. Consider wrapping in a try/catch and calling next(err):

```ts
  if (!initialized) {
      try {
          startApprouter(...);
          proxyMiddleware = createProxy(...);
          initialized = true;
      } catch (err) {
          return next(err);
      }
  }
```

.changeset/modern-canyons-wear.md
nikmace: Changed in: [e2fdad4](https://github.com/SAP/open-ux-tools/pull/4160/commits/e2fdad4b2dbe6e1de19c2edfa1582bdc92054ca0)

packages/backend-proxy-middleware-cf/src/config/env.ts
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * @param options - Env options to apply.
+ */
+function applyToProcessEnv(options: AppRouterEnvOptions): void {
+    const envOptions = {
+        ...options,
+        ...(options.destinations ? { destinations: JSON.stringify(options.destinations) } : {}),
+        ...(options.VCAP_SERVICES ? { VCAP_SERVICES: JSON.stringify(options.VCAP_SERVICES) } : {})
+    };
+    process.env = { ...process.env, ...envOptions } as Record<string, string>;
```
heimwege: I am sorry but my request for harmonization here might have been wrong 🙈 

Replacing `process.env` wholesale (not just assigning keys) can break native addons and some Node.js internals that hold a reference to the original `process.env` object. The standard pattern is `Object.assign(process.env, envOptions)`. The existing comment from heimwege suggests spreading, but the real fix is to use `Object.assign`.

packages/backend-proxy-middleware-cf/src/types.ts
nikmace: Default for us will be `.adp/reuse/xs-app.json`. I can set it to be `./xs-app.json` by default

packages/backend-proxy-middleware-cf/src/proxy/routes.ts
```diff
@@ -0,0 +1,118 @@
...
+ * @param options - xsappConfig, effectiveOptions, logger.
+ * @returns Route entries for the proxy.
+ */
+export function buildRouteEntries(options: BuildRouteEntriesOptions): RouteEntry[] {
+    const { xsappConfig, effectiveOptions, logger } = options;
+    const routes: RouteEntry[] = [];
+    const destList = Array.isArray(effectiveOptions.destinations) ? effectiveOptions.destinations : [];
+
+    for (const route of xsappConfig.routes ?? []) {
+        const routeMatch = /[^/]*\/(.*\/)?[^/]*/.exec(route.source);
```
heimwege, line 100: This regex extracts a path prefix from the xs-app.json source pattern, which is somewhat fragile. If a source pattern doesn't follow the expected structure (e.g. just ^/$), routeMatch is null and the route is silently skipped. No log or warning is emitted. This could lead to routes being silently dropped with no indication to the user.

packages/backend-proxy-middleware-cf/src/proxy/utils.ts
```diff
@@ -0,0 +1,105 @@
...
+ * Returns a path filter that accepts pathnames matching any custom route pattern or any destination route regex.
+ *
+ * @param customRoutes - Route path patterns (e.g. '/', '/login/callback').
+ * @param routes - Route entries with compiled regex.
+ * @returns Filter function (pathname) => boolean.
+ */
+export function createPathFilter(customRoutes: string[], routes: RouteEntry[]): (pathname: string) => boolean {
+    const compiledCustomRoutes = customRoutes.map((r) => new RegExp(String.raw`^${r}(\?.*)?$`));
+    return (pathname: string): boolean => {
+        return compiledCustomRoutes.some((re) => re.test(pathname)) || routes.some((route) => route.re.test(pathname));
```
heimwege: `re` is a compiled custom route. Can we maybe adjust the property name? `re` is hard to understand. Why not `customRoute`?

packages/backend-proxy-middleware-cf/src/proxy/utils.ts
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Returns a path filter that accepts pathnames matching any custom route pattern or any destination route regex.
+ *
+ * @param customRoutes - Route path patterns (e.g. '/', '/login/callback').
+ * @param routes - Route entries with compiled regex.
+ * @returns Filter function (pathname) => boolean.
+ */
+export function createPathFilter(customRoutes: string[], routes: RouteEntry[]): (pathname: string) => boolean {
+    const compiledCustomRoutes = customRoutes.map((r) => new RegExp(String.raw`^${r}(\?.*)?$`));
```
heimwege: `r` is inserted directly into the regex without escaping. If a custom route contains regex metacharacters (e.g. `/login/callback` contains `/` which is fine, but an extension route could contain `.` or `+`), this could produce incorrect matches. Use `r.replaceAll(/[-/\\^$*+?.()|[\]{}]/g, String.raw$&)` or validate that routes are safe to embed.

packages/backend-proxy-middleware-cf/src/types.ts
nikmace: Changed in: [c021720](https://github.com/SAP/open-ux-tools/pull/4160/commits/c0217207b5a880c485ad7e9bb2d0445bbf1b3b6d)

packages/backend-proxy-middleware-cf/src/proxy/proxy.ts
```diff
@@ -0,0 +1,138 @@
...
+        const route = routes.find((routeEntry) => routeEntry.re.test(url));
+        if (route?.path && route.url && effectiveOptions?.rewriteContentTypes?.includes(type?.toLowerCase() ?? '')) {
+            const encoding = (charset ?? 'utf8') as BufferEncoding;
+            let data = responseBuffer.toString(encoding);
+
+            const referrer =
+                (req.headers.referrer as string) ?? (req.headers.referer as string) ?? getRequestOrigin(req);
+            const referrerUrl = new URL(route.path, referrer).toString();
+
+            data = replaceUrl(data, `https://${route.url.slice(8)}`, referrerUrl);
```
heimwege: Isn't `https://${route.url.slice(8)}` the same as `route.url` here?
Besides that `route.url.slice(8)` looks a bit flaky to me. Why not using `new URL(route.url).host`?

packages/backend-proxy-middleware-cf/src/middleware.ts
nikmace: Changed in: [d648062](https://github.com/SAP/open-ux-tools/pull/4160/commits/d64806222296b0c89a8bf439b0d318fd1c5b311d)

packages/backend-proxy-middleware-cf/src/config/env.ts
nikmace: Changed in: [98b3d52](https://github.com/SAP/open-ux-tools/pull/4160/commits/98b3d52acfc171178ae724460e4afeef99c3ad87)

packages/backend-proxy-middleware-cf/src/proxy/routes.ts
nikmace: Changed in: [4543311](https://github.com/SAP/open-ux-tools/pull/4160/commits/4543311984297dd366b29eda501ddafcbf49f2b0)

packages/backend-proxy-middleware-cf/src/proxy/utils.ts
nikmace: Changed in: [4543311](https://github.com/SAP/open-ux-tools/pull/4160/commits/4543311984297dd366b29eda501ddafcbf49f2b0)

packages/backend-proxy-middleware-cf/src/proxy/utils.ts
nikmace: Changed in: [f8bb351](https://github.com/SAP/open-ux-tools/pull/4160/commits/f8bb3514444306ef8c80d08bd3dc1441f8925892)

packages/backend-proxy-middleware-cf/src/proxy/proxy.ts
nikmace: Changed in: [f8bb351](https://github.com/SAP/open-ux-tools/pull/4160/commits/f8bb3514444306ef8c80d08bd3dc1441f8925892)

nikmace:
> See comments + General question: why do we have this list of folders, each containing exactly one file (`index.ts`) instead of just naming the files according to the folder they are currently located in?

I've reworked the folders here: [5ed9c33](https://github.com/SAP/open-ux-tools/pull/4160/commits/5ed9c3308a1633c975436d816b3b520211f0c743). What do you think?

Issue #2531, opened 2024-11-06
TBI - no middleware specific knowledge and type definitions in @sap-ux/ui5-config


heimwege:
### Description (include screenshots)
@sap-ux/ui5-config should provide a comfortable way to work with ui5.yaml files without having knowledge and type definition for all middlewares there. The type definition of a middleware belongs to the respective middleware. Module @sap-ux/ui5-config should be able to work with [generics](https://www.typescriptlang.org/docs/handbook/2/generics.html) meaning, consumers bring their own type.

### Value
Currently to support a new setting you need to update two modules. With this TBI changing or just adding a setting for a middleware no longer requires changing the type in @sap-ux/ui5-config.

### Architecture Elaboration
Does it requires architecture elaboration? Yes

### Notes
@Klaus-Keller @tobiasqueck 

### Tasks
- [ ] Move existing middleware specific types from ui5-config to the respective middleware
- [ ] Move existing middleware specific functions from ui5-config to the respective middleware

Klaus-Keller:
Thanks @heimwege, this change makes sense to me and helps decoupling different modules and responsibilities.

PR #4162, opened 2026-02-17
(ADP)(Private cloud) Add small refactoring which includes comments leftovers from the original PR related to the Private cloud feature.

avasilev-sap:
#3865

**Deployment wizard:**
The refactoring is just for better readability it does not add or remove existing functionality and is not show stopper for the original PR to enter the upcoming release. We just want the ADP related logic in the validators from the deploy prompter to be extracted into separate functions and not directly implemented inline.

**UI5 config builder:**
Instead of checking the builder key for existence we strengthen a bit the condition here: We want to add an util method which returns all custom tasks defined for the builder as suggested by @heimwege. We want to be compatible with this policy for the ui5 config builder: https://github.com/SAP/open-ux-tools/issues/2531. This also is NOT a show stopper for the original PR. We assume that an existing ADP project is cloud if there are custom tasks otherwise - onPrem. The existence of builder key means that there will be defined custom tasks for sure so this check for custom tasks is relevant.

**Testing done:**
pending

### DIFF
--- a/.changeset/fresh-months-judge.md
+++ b/.changeset/fresh-months-judge.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/ui-components': major
----
-
-update border radius for UI components to improve consistency
--- a/examples/ui-prompting-examples/CHANGELOG.md
+++ b/examples/ui-prompting-examples/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux-private/ui-prompting-examples
 
+## 0.5.9
+
+### Patch Changes
+
+- Updated dependencies [c4839f5]
+    - @sap-ux/ui-components@2.0.0
+    - @sap-ux/ui-prompting@0.6.9
+
 ## 0.5.8
 
 ### Patch Changes
--- a/examples/ui-prompting-examples/package.json
+++ b/examples/ui-prompting-examples/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux-private/ui-prompting-examples",
-    "version": "0.5.8",
+    "version": "0.5.9",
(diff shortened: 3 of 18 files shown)


packages/ui5-config/src/ui5config.ts
```diff
@@ -837,6 +824,22 @@ export class UI5Config {
...
     }
 
+    /**
+     * Use this method to get the list of all available custom tasks defined in the ui5.yaml.
+     * If there are not custom tasks the method returns an empty array.
+     *
+     * @returns The list of custom tasks.
+     * @memberof UI5Config
+     */
+    public getAllCustomTasks<T extends object = object>(): Array<CustomTask<T>> {
```
heimwege: That's way better. thanks. I'm not sure about the usage of a generic type though. If the `customTasks` array contains different types of tasks mixed together (which is pretty likely I'd say) this function will just return one type for all tasks which is wrong.

avasilev-sap: If you call `ui5Config.getAllCustomTasks()` it will return `CustomTask<unknown>[]` which is ok. If there is a case in which the configuration for all tasks matches pre-defined interface the consumer can write: `ui5Config.getAllCustomTasks<PreDefinedInterface>()` just to enable autocomplete in the IDE. I can remove the generic type if you insist it is just sugar.

avasilev-sap: fixed in 1e3733c126bd8573adaf47359a037e57d7ec202b

cianmSAP (dismissed):
Thanks for updates
Have not tested but changes are not functional and very clear

heimwege (approved):
reviewed package `ui5-config`
- changes are straightforward
- coverage is great
- review comments addressed
- did NOT test manually

PR #4180, opened 2026-02-18
fix(security): upgrade  @vscode-logging/logger

donal-tobin-sap:
- npm audit moderate fro ui5-tooling install

Issue: #36803


donal-tobin-sap:
Updated by renovate #4200

Issue #4157, opened 2026-02-16
FEATURE - Create service keys for destination and xsuaa services in Adaptation Project generator


mmilko01:
_As a **developer**,
I want to **the service keys or destination and xsuaa services to be created and stored in `default-env.json` when the Adaptation Project is created**,
so that they **can be used when using Adaptation Editor with @sap/approuter**_

### Description
Currently, the service keys for the destination and xsuaa service are not created during the creation of the Adaptation Project. They are a prerequisite for the integration of the `@sap/approuter` in Adaptation Editor flow. The service keys should be generated during creation of the project and stored locally in `default-env.json` file to be consumed by local approuter during preview.

### Technical Design
_What modules need to be created or enhanced?_ 
_How would the functionality be exposed and how can it be consumed?_

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have created and Adaptation Project for Cloud Foundry
**when** I generate the project
**then** the service keys for `destination` and `xsuaa` services should be generated and added in `default-env.json`

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #4187, opened 2026-02-19
feat: Create service keys

mmilko01:
#4157

* Service keys are now generated for `destination` and `xsuaa` services are now created when the Adaptation Project is being generated


packages/adp-tooling/src/types.ts
```diff
@@ -883,6 +883,13 @@ export interface ServiceKeys {
     };
 }
 
+export interface ServiceKeysCredentials {
+    [key: string]: any;
```
IvoSG: The types [key: string]: any and endpoints: any are too vague. It would be better to use more specific types here.

mmilko01: I checked and the interface is not used anywhere. Initially we wanted to store the service key credentials in a `default-env.json` in the workspace and I needed the interface, but as this is no longer the case I will delete the interface.

Removed in commit: 8920a65fef6fd656a0aeebfa10ce63de454b2051

Issue #4192, opened 2026-02-19
BUG - Backend middleware for Cloud Foundry doesn't rewrite paths before sending to backend


mmilko01:
### Related Feature
Feature request: _issue number_

### Description
Currently, the backend middleware can only use the same URL when proxying the request to the backend as the original request URL. This works for some requests to service, but for requests send via reuse libraries the original URL doesn't match with the URL expected by the backend. In those cases the `xs-app.json` configuration must be used to configure a path rewrite in the middleware from the `source` to the `target` URL.

### Steps to Reproduce
1. Create a Cloud Foundry adaptation project
2. Open the `ui5.yaml` file
3. Notice the there are not pathRewrite properties defined in the `backend-proxy-middleware-cf` middleware configuration

### Expected results
Path are rewritten to the expected `target` URL by the middleware

### Actual results
Paths remain the same as the original request

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #4193, opened 2026-02-19
fix: rewrite URLs based on xs-app.json in backend-proxy-middleware-cf

mmilko01:
Fix for #4192.

* `backend-proxy-middleware-cf` configuration now specifies `pathRewrite` properties base on `xs-app.json`.
* The original URLs processed by the middleware are rewritten before sent to the beckend.
* `fallthrough` property for `fiori-tools-servestatic` is not set to `true`, so service requests can be process by the `backend-proxy-middleware-cf`


avasilev-sap (approved):
LGTM. This is a temporary solution meant to work for a short period of time.
Approve as is, did not test manually.

PR #4220, opened 2026-02-23
fix(infra): upgrade promptfoo

donal-tobin-sap:
- upgrade patch version
- does not fail to install on windows
- needs a higher version of node


heimwege (approved):
version bump lgtm. Thanks @donal-tobin-sap

donal-tobin-sap:

- already approved, only lock file rebuild due to source control conflicts

PR #4222, opened 2026-02-23
fix(sap-systems-ext): fix naming validation when importing

cianmSAP:
Corrects the validation on system names found when importing system configs


mikicvi-SAP (approved):
Thanks @cianmSAP 
- Additional validation lg
- Tests are covering the change
- Did not test manually
- Changeset ✅

PR #4243, opened 2026-02-25
fix(preview-middleware): fix /cards/store endpoint for CAP projects

Singham-24:
__Description:__ 
Fix /cards/store and /editor/i18n endpoints for CAP projects. In CAP projects, `path.resolve()` returns the CAP root instead of the UI5 app directory. The fix uses `getSourcePath()` for CAP projects which returns the correct webapp path.

### DIFF
--- a/.changeset/hip-bananas-float.md
+++ b/.changeset/hip-bananas-float.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/backend-proxy-middleware': patch
----
-
-export TypeScript types for external usage
--- a/examples/fe-fpm-cli/CHANGELOG.md
+++ b/examples/fe-fpm-cli/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/fe-fpm-cli
 
+## 0.1.109
+
+### Patch Changes
+
+- Updated dependencies [6c993f3]
+    - @sap-ux/fe-fpm-writer@0.42.17
+
 ## 0.1.108
 
 ### Patch Changes
--- a/examples/fe-fpm-cli/package.json
+++ b/examples/fe-fpm-cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/fe-fpm-cli",
-    "version": "0.1.108",
+    "version": "0.1.109",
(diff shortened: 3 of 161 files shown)


.changeset/fix-cap-cards-store.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-ux/preview-middleware": patch
+---
+
+Fix /cards/store and /editor/i18n endpoints for CAP projects by using getSourcePath() instead of path.resolve() for webapp path resolution.
```
heimwege: I think it's better to describe what is being fixed than how it is fixed. So that somebody reading the changelog can understand what has been done.

```suggestion
Fix card generator endpoints using wrong paths for CAP projects.
```

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1079,7 +1094,16 @@ export class FlpSandbox {
      */
     private async storeI18nKeysHandler(req: Request, res: Response): Promise<void> {
         try {
-            const webappPath = await getWebappPath(path.resolve(), this.fs);
+            // For CAP projects, getSourcePath() returns the webapp path directly
+            // For non-CAP projects, use the original path.resolve() approach
```
heimwege: If `getSourcePath` returns the webapp path directly, why then not using it in both cases, CAP and non-CAP?

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1026,7 +1026,22 @@ export class FlpSandbox {
                 fileName?: string;
                 manifests: MultiCardsPayload[];
             };
-            const webappPath = await getWebappPath(path.resolve(), this.fs);
+
+            // For CAP projects, getSourcePath() returns the webapp path directly
+            // For non-CAP projects, use the original path.resolve() approach
```
heimwege: If `getSourcePath` returns the webapp path directly, why then not using it in both cases, CAP and non-CAP?

.changeset/fix-cap-cards-store.md
Singham-24: done

packages/preview-middleware/src/base/flp.ts

packages/preview-middleware/src/base/flp.ts

PR #4245, opened 2026-02-25
fix(create): make generate readme script more robust

heimwege:
Script fails if log entry is not JSON. With this PR we will search for JSON inside the log output


heimwege:
@donal-tobin-sap thanks for the hint. I'll refactor this in a separate PR

PR #4247, opened 2026-02-25
Fix(create): use spawnSync instead of execSync

heimwege:


packages/create/scripts/generate-readme.js
```diff
@@ -155,7 +167,5 @@ try {
     console.log('✅ README.md generated successfully.');
 } catch (error) {
     console.error('❌ Failed to generate README:', error.message);
-    if (error.stdout) {
-        console.error('CLI Output:', error.stdout);
-    }
+    process.exit(1);
```
donal-tobin-sap, line 170: 👍🏽

packages/create/scripts/generate-readme.js
```diff
@@ -1,4 +1,4 @@
-const { execSync } = require('node:child_process');
+const { spawnSync } = require('node:child_process');
```
donal-tobin-sap, line 1: Apologies it was fork that I was thinking of, fork takes a module and runs from the current node instance. 
https://nodejs.org/docs/latest-v20.x/api/child_process.html#child_processforkmodulepath-args-options
This will work for all cases, especially where a different shell to default or nvm etc are used.

https://dev-aditya.medium.com/child-processes-in-node-js-spawn-exec-fork-and-use-cases-6eab4ddb9dcf

PR #4248, opened 2026-02-25
Improve package validation message for  with lowercase warning

kjose90:
internal issue: 37248

- Add validation to warn users when ABAP package names contain lowercase `$tmp`
- Users are immediately notified of the issue during configuration setup which will prevents 500 error during deployment


packages/abap-deploy-config-inquirer/src/translations/abap-deploy-config-inquirer.i18n.json
```diff
@@ -105,6 +105,7 @@
     "warnings": {
         "packageNotFound": "The package list cannot be fetched. Please enter the package manually.",
         "providePackage": "Provide a package.",
+        "packageTmpLowercase": "The temporary package entered contains '$tmp' in lowercase. Package names with $TMP must be in uppercase.",
```
longieirl: I think this sentence needs to be updated, its not a temporary package, $TMP is a local package (not a normal transportable package). When used, you are saying, its only for local development, it cant be moved to other systems.
Suggestion is `The local package entered contains '$tmp' in lowercase`.
If `$TMP` is used, then there should be no transport requested.
FYI @lfindlaysap

Issue #4212, opened 2026-02-20
TBI - Bump up 'border-radius' on elements


gpeuc:
VS Code has changed some border radius, made things more smooth.

**Buttons:**

Buttons are now `border-radius: 4px`

<img width="215" height="62" alt="Image" src="https://github.com/user-attachments/assets/fc314184-289e-4767-84e4-6755740cd299" />

**Inputs:**

Inputs are now `border-radius: 4px`

<img width="320" height="84" alt="Image" src="https://github.com/user-attachments/assets/018a673c-ec7b-4544-bab7-786f61451b94" />

**Dropdown elements:**

Dropdowns are now `border-radius: 4px`

<img width="440" height="220" alt="Image" src="https://github.com/user-attachments/assets/1f33c956-b38e-4832-9a0b-ee6f3ee31306" />

**Checkboxes:**

Checkboxes are now `border-radius: 3px` (three)

<img width="111" height="91" alt="Image" src="https://github.com/user-attachments/assets/3324828a-c165-4b1e-a8d5-fe038dcd9645" />

**Mini buttons:**

When hovered, we show this background, which is now `border-radius: 5px` (five)

<img width="127" height="101" alt="Image" src="https://github.com/user-attachments/assets/e1d87871-e91d-4b26-a25f-329c0a67de8a" />

gpeuc:
Done

PR #4251, opened 2026-02-26
fix: update border radius for UI components to improve consistency

mushegh-sap:
Issue: https://github.com/SAP/open-ux-tools/issues/4212

Bump up 'border-radius' on elements

### DIFF
--- a/.changeset/sparkly-dolls-beam.md
+++ b/.changeset/sparkly-dolls-beam.md
@@ -1,5 +1,0 @@
----
-'@sap-ux/ui5-test-writer': patch
----
-
-remove Filterbar OPA tests for Worklist app
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.200
+
+### Patch Changes
+
+- @sap-ux/fiori-elements-writer@2.8.86
+- @sap-ux/fiori-freestyle-writer@2.5.63
+
 ## 1.1.199
 
 ### Patch Changes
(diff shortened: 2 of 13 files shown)


mushegh-sap:
> @mushegh-sap
> 
> 1. unit tests are failing in `ui-components`
> 2. changelog is missing, you can generate it using `pnpm cset` -> https://github.com/SAP/open-ux-tools/blob/main/package.json#L64

thanks working

packages/ui-components/src/components/UICallout/UICallout.tsx
```diff
@@ -22,7 +22,7 @@ export const CALLOUT_STYLES = {
     boxShadow: 'var(--ui-box-shadow-small)',
     borderColor: 'var(--vscode-editorWidget-border)',
     text: 'var(--vscode-editorWidget-foreground)',
-    borderRadius: 2
+    borderRadius: 4
```
815are, line 25: Hi @gpeuc 

could you please approve/verify from you side?
is it expected to change for all callouts - in issue I dropdown mentioned and it is not clear if it should be for dropdown only or for all callouts.
@mushegh-sap applied change for all callouts
It affects for example:

Image

Image

gpeuc: Yes, let's keep then callouts in sync with dropdowns. This is good.

Just as a reminder @815are - we said we would revert these inverted callouts back to normal colour (so normal background and normal text), any chance you do that quickly?

815are (approved):

packages/ui-components/src/components/UICallout/UICallout.tsx
815are: will check, but handle in separate PR

broksy (approved):
Changeset reflects to the changes made.
Approved from my side.

.changeset/fresh-months-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/ui-components': major
```
devinea, line 2: @mushegh-sap This should have been a patch. Not a major!!

PR #4256, opened 2026-02-26
chore: create dependabot.yml with sensible limits

mikicvi-SAP:
- Adds `dependabot.yml` to control the behaviour of dependabot with some sensible limits to prevent wasting of CI/CD compute resources
	- we run Renovate, but Dependabot is mandatory on SAP repositories.
---

- Docs https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference
- We could [ignore](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#ignore--) some of the dependencies with config mentioned in docs.


.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
+version: 2
+
+updates:
+    - package-ecosystem: 'github-actions'
+      directory: '/'
+      open-pull-requests-limit: 3
+      schedule:
+          interval: 'monthly'
```
donal-tobin-sap: Actions should be updated on a shorter cycle.
I would go down to daily, but at the longest weekly.

mikicvi-SAP: Renovate handles actions as well. Maybe we can settle on weekly?

mikicvi-SAP:
<del>Unsure why this is failing now with `pnpm` as a chosen package manager, when the [documentation](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#package-ecosystem-) clearly lists `pnpm` as a valid option.</del>
Thanks @devinea for pointing out a part I missed in the docs.

PR #4258, opened 2026-02-26
fix(ui-service-inq): use hidenewsystem prompt option for ui service generator

cianmSAP:
Internal bug 37142

Utilise the prompt option `hideNewSystem` to hide the option in the System list from `@sap-ux/odata-service-inquirer`


mikicvi-SAP (approved):
Thanks @cianmSAP 
- Updates LG
- Didn't test manually 
- Changeset ✅

PR #4268, opened 2026-02-27
fix(infra): upgrade fast-xml-parser

donal-tobin-sap:
- use latest version as high probability that are not yet reported security findings

- https://github.com/SAP/open-ux-tools/security/dependabot/414

Issue: #37278


longieirl (commented):
Do you need a changeset?

donal-tobin-sap:

- just rebase from previous merge release build

Issue #4266, opened 2026-02-27
FEATURE -  `@sap-ux/odata-service-inquirer`: Add support to system selection for service based auth (instead of catalog)


IainSAP:
### Description

Odata services may be provided standalone and not published via the catalog. When connecting using a destination or stored backend system connection configuration the default authentication approach is to use a call to the services catalogs. An alternative url should be allowed and used for authentication in these cases.

This can be achieved adding an additional property to the system selection prompt options:

https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-inquirer/src/types.ts#L330
e.g.
```
/**
     * Provide a default choice for the system selection prompt, this is used to pre-select a system based on the system name.
     * Set as string literal types `NewSystemChoice` or `CfAbapEnvServiceChoice` to specify the default choice to create a new system connection config in VSCode
     * or to select the Cloud Foundry Abap environments service discovery choice in BAS respectively.
     * Supported as object reference `{ value: }` to allow binding to runtime variables of other prompts or as string.
     * `{ connectPath }`, if specified this path is appended to the default choice system path and will be used to authenticate as a service endpoint instead of a catalog request
     */
    defaultChoice?: string | { value?: string, connectPath?: string };
```

### Technical Design
_What modules need to be created or enhanced?_ 
_How would the functionality be exposed and how can it be consumed?_

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have a ...
**when** I execute ...
**then** ... will happen.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #4269, opened 2026-02-27
feat(`odata-service-inquirer`): Adds connect path option to system prompt

IainSAP:
Adds feature: https://github.com/SAP/open-ux-tools/issues/4266

## Summary
- Adds support for alternative authentication path (`connectPath`) to system selection prompts in `@sap-ux/odata-service-inquirer`
- Allows systems (backend systems or destinations) to authenticate using an OData service endpoint (any url path) instead of the default catalog request
- Useful when catalog endpoints are not accessible but specific service endpoints are available

## Changes
- Added `ConnectionType` type (`'abap_catalog' | 'odata_service'`) to specify authentication approach
- Extended `defaultChoice` option in `SystemSelectionPromptOptions` to accept `{ value?: string; connectPath?: string }`
- Updated `ConnectionValidator` methods (`validateUrl`, `validateAuth`, `checkUrl`, `createSystemConnection`) to accept optional `connectType` parameter
- When `connectType` is `'odata_service'`, authentication uses the service endpoint instead of catalog `listServices()`
- Non-origin URLs are now allowed for system validation when `connectType` is `'odata_service'`
- Added comprehensive unit tests for the new functionality

## Test plan
- [x] Verify existing system selection flows work unchanged (no `connectPath` specified)
- [x] Test backend system authentication with `connectPath` option using reentrance ticket auth
- [x] Test backend system authentication with `connectPath` option using basic auth
- [x] Verify catalog endpoints are skipped when `connectType: 'odata_path` is used

### DIFF
--- a/packages/deploy-config-sub-generator/CHANGELOG.md
+++ b/packages/deploy-config-sub-generator/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/deploy-config-sub-generator
 
+## 0.5.94
+
+### Patch Changes
+
+- Updated dependencies [698fd9e]
+    - @sap-ux/odata-service-inquirer@2.19.0
+
 ## 0.5.93
 
 ### Patch Changes
--- a/packages/deploy-config-sub-generator/package.json
+++ b/packages/deploy-config-sub-generator/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/deploy-config-sub-generator",
     "description": "Main generator for configuring ABAP or Cloud Foundry deployment configuration",
-    "version": "0.5.93",
+    "version": "0.5.94",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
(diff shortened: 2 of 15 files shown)


packages/odata-service-inquirer/src/prompts/connectionValidator.ts
```diff
@@ -65,6 +65,9 @@ interface AxiosExtensionRequestConfig extends AxiosRequestConfig {
 }
 // System specific authentication mechanism, used to determine the connection auth type
 export type SystemAuthType = 'serviceKey' | 'reentranceTicket' | 'basic' | 'unknown';
+// Systems (stored backends configs or destinations) can connect using either the specified full url or catalog requests
+export type ConnectionType = 'abap_catalog' | 'odata_service';
```
cianmSAP: Can we use `ConnectionType` from `@sap-ux/store?`

IainSAP: Its not exactly the same thing, the naming is not right here. I think actually it should be renamed to `abap_catalog` or `path`. It may not be a service, it can be any url. We are not calling `$metadata`. Ill update it to `odata_path` as it allows for any url to be used to authenticate. The metadata call is only made by the service prompt

cianmSAP (approved):
Updates look good, query answered
Changeset and tests present

PR #4276, opened 2026-02-27
fix(security): upgrade rollup

donal-tobin-sap:
- https://github.com/SAP/open-ux-tools/security/dependabot/391
- renovate is not able to generate a PR
- use is from a UI5 module


donal-tobin-sap:

- already approved just needed branch update

donal-tobin-sap:

- already approved just needed updated to pnpm 10 lock file

PR #4278, opened 2026-03-02
feat(preview-middleware): add bundleName support for i18n in CAP projects.

Singham-24:
Adds support for bundleName in i18n configuration and improves locale fallback logic for CAP projects.

### DIFF
--- a/.changeset/fix-2272-odata-gen-onpremise-deploy.md
+++ b/.changeset/fix-2272-odata-gen-onpremise-deploy.md
@@ -1,8 +1,0 @@
----
-'@sap-ux/btp-utils': patch
-'@sap-ux/deploy-tooling': patch
----
-
-fix: allow deployment to OnPremise destinations with WebIDEUsage odata_gen
-
-isAbapSystem now returns true for destinations with ProxyType=OnPremise, fixing deployments that failed with a cryptic 'bind' error when WebIDEUsage was set to odata_gen. deploy-tooling also now surfaces an actionable error message if a non-ABAP provider is resolved.
--- a/examples/odata-cli/CHANGELOG.md
+++ b/examples/odata-cli/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/odata-cli
 
+## 0.17.53
+
+### Patch Changes
+
+- Updated dependencies [2e17a6b]
+    - @sap-ux/btp-utils@1.1.12
+    - @sap-ux/axios-extension@1.25.27
+
 ## 0.17.52
 
 ### Patch Changes
--- a/examples/odata-cli/package.json
+++ b/examples/odata-cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/odata-cli",
-    "version": "0.17.52",
+    "version": "0.17.53",
     "description": "Simple example CLI uing the @sap-ux/axios-extension module to fetch metadata and annotations from an SAP system.",
     "license": "Apache-2.0",
     "private": true,
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,16 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.229
+
+### Patch Changes
+
+- Updated dependencies [2e17a6b]
+    - @sap-ux/btp-utils@1.1.12
+    - @sap-ux/axios-extension@1.25.27
+    - @sap-ux/system-access@0.7.3
+    - @sap-ux/fiori-elements-writer@2.8.112
+    - @sap-ux/fiori-freestyle-writer@2.5.82
+
 ## 1.1.228
 
 ### Patch Changes
--- a/examples/simple-generator/package.json
+++ b/examples/simple-generator/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/generator-simple-fe",
-    "version": "1.1.228",
+    "version": "1.1.229",
     "description": "Simple example of a yeoman generator for Fiori elements.",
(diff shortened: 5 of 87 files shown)


packages/preview-middleware/src/base/flp.ts
```diff
@@ -1093,23 +1093,31 @@ export class FlpSandbox {
...
 
-                i18nPath = i18nPathFromConfig;
-                supportedLocales = locales;
-                fallbackLocale = fallback;
+                supportedLocales = config.supportedLocales || [];
+                fallbackLocale = config.fallbackLocale;
             }
 
-            const requestedLocale = (req.query.locale as string) ?? fallbackLocale ?? '';
+            let requestedLocale = (req.query.locale as string) || fallbackLocale || '';
```
heimwege: Why replacing Nullish Coalescing with Logical OR?

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1093,23 +1093,31 @@ export class FlpSandbox {
...
                 i18nPath = i18nConfig;
-            } else if (typeof i18nConfig === 'object' && i18nConfig !== null && 'bundleUrl' in i18nConfig) {
-                const {
-                    bundleUrl: i18nPathFromConfig,
-                    supportedLocales: locales = [],
-                    fallbackLocale: fallback
-                } = i18nConfig as {
-                    bundleUrl: string;
+            } else if (typeof i18nConfig === 'object' && i18nConfig !== null) {
+                const config = i18nConfig as {
```
heimwege: Doesn't this type already exist and we can just re-use?

```ts
import type { Package } from '@sap-ux/project-access';
//[...]
const config = i18nConfig as Package['sap.app.i18nConfig']
```

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1093,23 +1093,31 @@ export class FlpSandbox {
...
+                    const bundleNameParts = config.bundleName.split('.');
+                    i18nPath = `${bundleNameParts.slice(-2).join('/')}.properties`;
+                } else if (config.bundleUrl) {
+                    i18nPath = config.bundleUrl;
+                }
 
-                i18nPath = i18nPathFromConfig;
-                supportedLocales = locales;
-                fallbackLocale = fallback;
+                supportedLocales = config.supportedLocales || [];
```
heimwege: Please use Nullish Coalescing


```suggestion
                supportedLocales = config.supportedLocales ?? [];
```

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1093,23 +1093,31 @@ export class FlpSandbox {
...
-                } = i18nConfig as {
-                    bundleUrl: string;
+            } else if (typeof i18nConfig === 'object' && i18nConfig !== null) {
+                const config = i18nConfig as {
+                    bundleUrl?: string;
+                    bundleName?: string;
                     supportedLocales?: string[];
                     fallbackLocale?: string;
                 };
+                if (config.bundleName) {
```
heimwege: So if a `bundleName` and a `bundleUrl` is provided, the name wins? Should we add a log message about that so that the user knows?

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1093,23 +1093,31 @@ export class FlpSandbox {
...
+                    // Convert bundleName (e.g., "sap.fe.cap.travel.i18n.i18n") to path
+                    // Remove the app ID prefix and convert dots to slashes
+                    const bundleNameParts = config.bundleName.split('.');
+                    i18nPath = `${bundleNameParts.slice(-2).join('/')}.properties`;
```
heimwege: You hard coded `slice(-2)`. What happens in case of `sap.fe.cap.travel.i18n.my.very.special.i18n` ? Is `special/i18n` really the path you want? I'd assumed the correct path is `i18n/my/very/special/i18n`. So how about subtracting the `sap.app.id` from the `bundleName` instead?

```ts
const bundleName = "sap.fe.cap.travel.i18n.my.very.special.i18n";
const result = fullPath.replace(this.manifest['sap.app'].id + ".", "");
// result -> 'i18n.my.very.special.i18n'
```

or

```ts
const bundleName = "sap.fe.cap.travel.i18n.my.very.special.i18n";
const startIndex = this.manifest['sap.app'].id.length + 1; // +1 for the dot
const result = bundleName.substring(startIndex);
// result -> 'i18n.my.very.special.i18n'
```

packages/preview-middleware/src/base/flp.ts

packages/preview-middleware/src/base/flp.ts

packages/preview-middleware/src/base/flp.ts
Singham-24: added

packages/preview-middleware/src/base/flp.ts
Singham-24: The type Package['sap.app.i18nConfig'] doesn't exist in @sap-ux/project-access

The code currently uses the "in" operator for proper TypeScript type narrowing with the union type

packages/preview-middleware/src/base/flp.ts
Singham-24: Instead of hardcoding slice(-2), we now subtract the sap.app.id from the bundleName

packages/preview-middleware/src/base/flp.ts
heimwege: Oh you are right. type `Package` makes no sense here 🙈 I wanted to refer to the `Manifest` type from project-access.

But I see you changed to Manifest['sap.app']['i18n'] 👍🏻

PR #4294, opened 2026-03-02
fix(security): upgrade rollup in lock file

donal-tobin-sap:
- https://github.com/SAP/open-ux-tools/security/dependabot/391
- update lock file again - change was lost

- change lost in pnpm 10 updates


donal-tobin-sap:

- lock file update only
- security fix already approved

PR #4304, opened 2026-03-02
fix(security): upgrade direct minimatch use

donal-tobin-sap:
- missed direct inclusion
- pnpm why showed no vulnerable usage but some remain in lock file
- https://github.com/SAP/open-ux-tools/security/dependabot/389

Issue: #37169


donal-tobin-sap:

- security fix
- version already overriden
- fixing security report

PR #4314, opened 2026-03-03
refactor: remove TypeScript parser from ESLint configurations

devinea:
Remove duplication in eslint config files.
- Removed the TypeScript parser import from ESLint configuration files in various packages.
- Updated parser options to exclude the parser setting while maintaining tsconfig root directory and project settings.

- No changeset as it is all test files.


IainSAP (dismissed):
Thanks @devinea 

- Script cleanup only

heimwege (dismissed):
Clean, mechanical refactor with no behavioral impact. SonarQube passes. The change correctly removes dead/misplaced configuration that was never doing what it appeared to be doing.

heimwege (approved):
This is a clean, correct, purely additive-negative (delete-only) refactor with no behavioral impact. The redundant parser declarations were either completely dead (root config blocks) or silently misconfigured (parser inside parserOptions). actual parser comes from root config. CI and SonarQube are green.

PR #4345, opened 2026-03-04
fix: update vocabularies

gernotkuhr:
vocabulary update
- update vocabularies
- add com.sap.vocabularies.Auditing.v1 vocabulary


gernotkuhr:
According to Bernhard and Mariana no changes are currently expected to wait for...

Jimmy-Joseph19 (dismissed):
Changeset found.
Updated vocabualry looks good.
Did not test with LSP.

PR #4346, opened 2026-03-04
fix(odata-service-writer): preserve preview path during service sync

korotkovao:
**Related to an internal issue 37299**

When changing the `hostname` in Service Manager and clicking Refresh and Save, the change wasn't reflected in `ui5.yaml`

**Problem:**
When syncing OData services in Service Manager (e.g., updating destination/hostname), the `previewSettings.path` was being overwritten even when users had explicitly configured it. This caused the backend path to revert from user-specified values (like `/backend`) back to the full service path, breaking the middleware configuration.

**Root Cause:**
The `setDefaultPreviewSettings()` function always overwrote `previewSettings.path` when it found an existing `/sap` backend in `ui5.yaml`, regardless of whether the path was explicitly set by the user or just a default.

**Solution:**
Added an `update` parameter to `setDefaultPreviewSettings()` function:

When `update = true` (during service sync): Preserves explicitly set `previewSettings.path`
When `update = false` (during initial service add): Applies defaults as before

**Implementation Details:**

Captures the original `previewSettings.path` before applying any defaults
Only overwrites the path if `!update || !explicitPreviewPath`
This ensures user-configured paths persist through service sync operations

### DIFF
--- a/examples/odata-cli/CHANGELOG.md
+++ b/examples/odata-cli/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/odata-cli
 
+## 0.17.48
+
+### Patch Changes
+
+- Updated dependencies [2d21925]
+    - @sap-ux/axios-extension@1.25.22
+
 ## 0.17.47
 
 ### Patch Changes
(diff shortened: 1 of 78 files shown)


korotkovao:
Thank you @broksy @devinea

PR #4374, opened 2026-03-05
fix: Use the cloudDevAdaptationStatus property in the app filter only for cloud apps.

avasilev-sap:
#4373

From the obsolete requirements for the Private cloud feature a leftover was pushed by mistake related to the app filters. For an older abap systems if we include the cloudDevAdaptationStatus property in the fields then the entire app index api call fails with error 400. This does not happen for all onPrem only abap systems, e.g. ui8.

### DIFF
--- a/examples/fe-fpm-cli/CHANGELOG.md
+++ b/examples/fe-fpm-cli/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/fe-fpm-cli
 
+## 0.1.118
+
+### Patch Changes
+
+- @sap-ux/fe-fpm-writer@0.43.4
+
 ## 0.1.117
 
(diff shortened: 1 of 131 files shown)


.changeset/slow-worms-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/adp-tooling': patch
+---
+
+(ADP) Remove the cloudDevAdaptationStatus property from the app filters related to onPrem apps.
```
mmilko01: We usually use the same description here, as the name of the PR

avasilev-sap: fixed in 81991c3a05375614c74042406231c1a6c596756f

PR #4377, opened 2026-03-05
fix(app-config-writer): add prerequisite to add eslint-config command

heimwege:
new prereq: check for non-esiting `eslint.config.mjs`

...and some minor adjustments


heimwege:
Sonar run not starting 🤷🏻 Previous run was ok, no new changes just merging `main` -> use admin rights

PR #4382, opened 2026-03-06
fix(odata-service-inquirer): Create an abap provider when connecting with service path (not a single service provider)

IainSAP:
Fix related to : https://github.com/SAP/open-ux-tools/issues/4266

- Creates an abap service provider even when dest is using a connect path to support requests other than the service - value helps, annotations etc.
- Creates new option for `connecPath` to decouple it from setting a `defaultChoice` for the system so prompt users can change the system but still use the connection path specified
- Updates tests

### DIFF
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.210
+
+### Patch Changes
+
+- @sap-ux/fiori-elements-writer@2.8.95
+- @sap-ux/fiori-freestyle-writer@2.5.72
+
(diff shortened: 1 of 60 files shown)


packages/odata-service-inquirer/src/prompts/connectionValidator.ts
```diff
@@ -96,8 +96,10 @@ export class ConnectionValidator {
     private _connectedUserName: string | undefined;
     private _connectedSystemName: string | undefined;
     private _refreshToken: string | undefined;
-    // For the current validated URL connection attempts will ignore cert errors
+    // For the current validated URL connection attempts will ignore cert erro
```
longieirl: Typo

PR #4385, opened 2026-03-07
test: rework jest environment e2e testing

whydrae:
This PR reworks the `jest-ui5-environment` e2e testing:

1. Instead of using the file fixtures, ui5 version are requested by the `ui5-info` package.
2. The ui5 configurations are then generated automatically from the fetched versions. Only maintained LTS versions are used, including the latest.
3. I have removed the `--detectOpenHandles` and `--runInBand` options, since they are dedicated for the debugging and bring additional performance penalties.
4. The minimum version for e2e testing is `1.96.*`, because latest `1.71.*` gives some ui5 distribution error.


whydrae:
@nlunets could you kindly check this one?

Issue #4387, opened 2026-03-09
FEATURE - Handle i18n translations from key user changes in ADP generator


nikmace:
_As a **developer**,
I want to **create an ADP with key-user changes and their respective i18n files**,
so that I **can get translations in the editor**_

### Description

Key user changes returned by the backend include translatable text information in two places: `content.texts`, which contains metadata and the text reference used within the change, and the top-level `texts`, which contains actual translations per language. In Adaptation Projects, translations must not be stored inside the change file. Instead, the generator extracts the translations and writes them into the project's i18n `.properties` files. The `texts` block inside the change file is preserved, but the entire `value` property is replaced with an i18n binding expression using a collision‑safe key constructed as `<fileName>_<textId>`. All translation values from the top-level `texts` object are written into `i18n/i18n.properties` or the corresponding `i18n_<lang>.properties` files. Each entry follows the SAP i18n format and includes a comment using the original UI5 text type (e.g., `# XTIT`, `# XFLD`).

The change file retains the `texts` section, including the original `textId` keys and metadata such as `type` and `maxLength`. During conversion, the complete `value` property of each `content.texts[textId]` entry is replaced with an i18n binding expression:

- Generated key:
  ```
  <fileName>_<textId>
  ```
- Binding:
  ```
  "{i18n><generatedKey>}"
  ```

The top-level `texts` object from the API response contains the translations. For each translation entry:

- The translation for the empty language key `""` is written to:
  ```
  i18n/i18n.properties
  ```
- The translation for any language key (e.g., `"en"`) is written to:
  ```
  i18n/i18n_<lang>.properties
  ```

Each `.properties` entry is written with the text type comment followed by the generated key and value:

```
# <type>
<generatedKey>=<translation>
```

The language codes provided in the key user data response do not always match the locale identifiers expected by i18n `.properties` files. To ensure correct placement of translations across languages, the generator must apply the same language mapping rules used by UI5 when resolving locales. This ensures that translations are written to the correct `i18n.properties` or `i18n_<lang>.properties` files according to UI5’s locale resolution behavior.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have created a CloudReady ADP with key-user changes
**when** I open the Adaptation Editor
**then** I see the i18n keys and their translations.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #4388, opened 2026-03-09
feat: Handle i18n translations from key user changes in ADP generator

nikmace:
Feat for #4387.
- Replaces content.texts values with i18n bindings and writes translations to <lang>.properties files
- Adds 400 check for non-existent key-user API endpoint
- Adds types for text translations in axios-extension


.changeset/brave-pans-think.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/adp-tooling': patch
```
mmilko01: Maybe it is implicitly included, but should `axios-extension` also be included in the changeset?

packages/adp-tooling/src/writer/i18n/key-user-translations.ts
```diff
@@ -0,0 +1,128 @@
...
+    language = M_ISO639_OLD_TO_NEW[language] ?? language;
+
+    // Collect script and region subtags, skip variants/extensions/private use
+    const subtags: string[] = [];
+    for (let i = 1; i < parts.length; i++) {
+        const p = parts[i];
+        if (p.length === 4 && /^[A-Za-z]{4}$/.test(p)) {
+            // Script subtag (e.g., Hans, Hant, Latn)
+            subtags.push(p[0].toUpperCase() + p.slice(1).toLowerCase());
+        } else if (p.length === 2 && /^[A-Za-z]{2}$/.test(p)) {
```
mmilko01: Same as above

packages/generator-adp/src/app/questions/key-user.ts
```diff
@@ -366,8 +366,11 @@ export class KeyUserImportPrompter {
             this.logger.error(`Error validating key-user changes for adaptation ${adaptationId}: ${e.message}`);
             this.logger.debug(e);
 
-            if (isAxiosError(e) && (e.response?.status === 405 || e.response?.status === 404)) {
-                return t('error.keyUserNotSupported');
+            if (isAxiosError(e)) {
+                const status = e.response?.status;
+                if (status === 400 || status === 404 || status === 405) {
```
heimwege, line 371: Just a thought, but as all `4xx` http responses are related to client errors you might just check if status starts with `4`? e.g. `403` (= forbidden) might also occur, right?

nikmace: - **400 Bad Request** — the server doesn't recognize the request, likely because the key-user API isn't
available on this system
- **404 Not Found** — the endpoint doesn't exist on this backend
- **405 Method Not Allowed** — the endpoint exists but doesn't support this operation

All three point to the same conclusion: the key-user feature isn't supported on this system. Auth errors (**401**/**403**) and other **4xx** codes have different meanings and are already handled by the generic e.message fallback at line 376. What do you think @heimwege?

PR #4392, opened 2026-03-10
feat: support for full service url systems in the generator

cianmSAP:
Internal issue 37128

- Updates `@sap-ux/odata-service-inquirer` to also fetch backend systems with connection type `odata_service`
- Adds new property `connectPath` to the backend config in yamls. In the case a system with connectionType `odata_service` is used, then `connectPath` will be assigned the value of the URL's path e.g. 
```
        backend:
          - url: http://ccwdfgl9773.devint.net.sap:50000
            connectPath: /sap/opu/odata/iwfnd/CATALOGSERVICE;v=0002/
            client: "100"
```
- Updates in `setDefaultPreviewSettings` to rearrange the ordering of the backend properties to be more intuitive (path,url,client at the top of the config) - causing some snapshot updates
- Updates in backend proxy middleware to check for the `connectPath` property and update the call to read from the store

### DIFF
--- a/packages/cf-deploy-config-sub-generator/CHANGELOG.md
+++ b/packages/cf-deploy-config-sub-generator/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/cf-deploy-config-sub-generator
 
+## 0.2.153
+
+### Patch Changes
+
+- Updated dependencies [0e856a7]
+    - @sap-ux/cf-deploy-config-writer@0.3.83
+
 ## 0.2.152
 
 ### Patch Changes
--- a/packages/cf-deploy-config-sub-generator/package.json
+++ b/packages/cf-deploy-config-sub-generator/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/cf-deploy-config-sub-generator",
     "description": "Generators for configuring Cloud Foundry deployment configuration",
-    "version": "0.2.152",
+    "version": "0.2.153",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/cf-deploy-config-writer/CHANGELOG.md
+++ b/packages/cf-deploy-config-writer/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/cf-deploy-config-inquirer
 
+## 0.3.83
+
+### Patch Changes
+
+- 0e856a7: Ensure mta service-name is handled
+
 ## 0.3.82
 
 ### Patch Changes
--- a/packages/cf-deploy-config-writer/README.md
+++ b/packages/cf-deploy-config-writer/README.md
@@ -45,29 +45,10 @@ await mtaConfig.addMtaExtensionConfig('mynewdestination', 'https://my-service-ur
 await mtaConfig.save();
 ```
 
-## Example: Appending a `Managed` Approuter Configuration to an SAP Fiori UI5 Application
-
-Calling the `generateAppConfig` function to append Cloud Foundry configuration to a HTML5 application, assumes `manifest.json` and `ui5.yaml` configurations are present otherwise the process will exit with an error;
-```Typescript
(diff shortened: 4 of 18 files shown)


packages/odata-service-inquirer/src/prompts/datasources/sap-system/service-selection/questions.ts
```diff
@@ -236,7 +236,8 @@ export function getSystemServiceQuestion(
                 (answers: { [serviceSelectionPromptName]?: ServiceAnswer }) =>
                     !!(connectValidator.serviceProvider instanceof AbapServiceProvider) &&
                     !!convertedMetadataRef.convertedMetadata &&
-                    answers?.[serviceSelectionPromptName]?.serviceODataVersion === ODataVersion.v4
+                    answers?.[serviceSelectionPromptName]?.serviceODataVersion === ODataVersion.v4 &&
+                    connectValidator.connectType === 'abap_catalog'
```
IainSAP: This isnt valid for non catalog auth that also download value helps? As long as theres an abap provider then we can call value help metadata download.

IainSAP (dismissed):

packages/odata-service-inquirer/src/prompts/datasources/sap-system/service-selection/questions.ts
cianmSAP: I've removed the condition - @devinea are you ok with the value help prompt showing for these service url systems?

packages/fiori-app-sub-generator/src/fiori-app-generator/transforms.ts
```diff
@@ -184,6 +184,43 @@ function getUI5Uri(): string {
...
+        previewSettings = { authenticationType: AuthenticationType.ReentranceTicket };
+    } else if (service.apiHubConfig) {
+        previewSettings = { apiHub: true };
+    }
+
+    if (service.connectedSystem?.backendSystem?.connectionType === 'odata_service') {
+        const url = service.connectedSystem.backendSystem.url;
+        previewSettings = {
+            ...previewSettings,
+            connectPath: url ? new URL(url).pathname : undefined
```
IainSAP, line 217: 👍

packages/odata-service-inquirer/src/prompts/datasources/sap-system/service-selection/questions.ts
devinea: @cianmSAP I don't think it makes that much sense personally. But I think @ianquigley-sap was clarifying use case.

packages/backend-proxy-middleware/src/base/proxy.ts
```diff
@@ -412,11 +412,15 @@ async function updateProxyConfigFromStore(
     // check if system credentials are stored in the store
     try {
         const systemStore = await getService<BackendSystem, BackendSystemKey>({ logger, entityName: 'system' });
+        let backendUrl = localBackend.url;
+        if (localBackend.connectPath) {
+            backendUrl = new URL(localBackend.connectPath, localBackend.url).href;
```
heimwege: What happens in case the `connectPath` has been entered w/o leading slash in the yaml config and the `url` ends with a slash? Do we need to sanitize `connectPath` to start with a slash or is it intended?

`new URL('test', 'https://api.example.com/v1').href` -> `https://api.example.com/test`
`new URL('/test', 'https://api.example.com/v1').href` -> `https://api.example.com/test`
`new URL('/test', 'https://api.example.com/v1/').href` -> `https://api.example.com/test`
but
`new URL('test', 'https://api.example.com/v1/').href` -> `https://api.example.com/v1/test` (`v1` not being removed in this case)

cianmSAP: Updated in https://github.com/SAP/open-ux-tools/pull/4392/changes/16186fc04095e2ccbc830ee5188da4831b29fad9

PR #4393, opened 2026-03-11
feat: ignore service prefixes during searching service

815are:
ignore service prefixes during searching service using method `getCapServiceName`.
How method works after update:
1. first it tries to find service by exact match like before changes and how it currently works in `main` branch;
2. if service was not found, then it tries to match service using following patterns:
2.1. `/ui/<inbound-service-name>/v<version>/<service_suffix>`
2.2. `/<string>.<string>/external-ui/<inbound-service-name>/v<version>/<service_suffix>`


packages/project-access/src/project/cap.ts
```diff
@@ -424,8 +434,15 @@ function findServiceByUri(
     services: { name: string; urlPath: string }[],
     uri: string
 ): { name: string; urlPath: string } | undefined {
-    const searchUri = uniformUrl(uri).replace(/(?:^\/)|(?:\/$)/g, '');
-    return services.find((srv) => srv.urlPath.replace(/(?:^\/)|(?:\/$)/g, '') === searchUri);
+    const searchUri = normalizeServiceUrlPath(uniformUrl(uri));
+    let service = services.find((srv) => normalizeServiceUrlPath(srv.urlPath));
```
heimwege: Is this correct? `find` expects a condition but I just see the `normalizeServiceUrlPath`. I would have expected something like

```suggestion
    let service = services.find((srv) => normalizeServiceUrlPath(srv.urlPath) === searchUri);
```

815are: yes, mistake, did not notice, thanks

815are: fixed - https://github.com/SAP/open-ux-tools/pull/4393/changes/c39dba26c3138ac7afb3531770c4b4f7a0d0ef1d

heimwege (approved):
- coverage is excellent
- changes are straightforward
- did NOT test manually
- changeset exists

.changeset/mighty-camels-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/project-access': patch
+---
+
+Enhance `getCapServiceName` to return service by ignorering prefix for service URL in parameter `datasourceUri`.
```
heimwege:
```suggestion
Enhance `getCapServiceName` to return service by ignoring prefix for service URL in parameter `datasourceUri`.
```

packages/project-access/src/project/cap.ts
```diff
@@ -413,6 +413,44 @@ export async function readCapServiceMetadataEdmx(
...
+
+    const patterns = [
+        // regex for pattern -> /ui/<inbound-service-name>/v<version>/...
+        String.raw`^/ui/[^/]+/v\d+/${expectedSufixPath}$`,
+        // regex for pattern -> /<string>.<string>/external-ui/<inbound-service-name>/v<version>/...
+        String.raw`^/[^/]+\.[^/]+/external-ui/[^/]+/v\d+/${expectedSufixPath}$`
+    ];
+
+    return patterns.some((pattern) => new RegExp(pattern).test(path));
+}
```
heimwege: 1. typo in suffix
2. parameter mutation (expectedSufixPath = expectedSufixPath.replace(...) - reassigning the input parameter
  instead of using a local variable)

```suggestion
export function isMatchingServiceUri(path: string, expectedSuffixPath: string): boolean {
    // Escapes special regex characters in a string so it can be embedded into regular expression
    const escapedSuffix = expectedSuffixPath.replace(/[.*+?^${}()|[\]\\]/g, String.raw`\$&`);

    const patterns = [
        // regex for pattern -> /ui/<inbound-service-name>/v<version>/...
        String.raw`^/ui/[^/]+/v\d+/${escapedSuffix}$`,
        // regex for pattern -> /<string>.<string>/external-ui/<inbound-service-name>/v<version>/...
        String.raw`^/[^/]+\.[^/]+/external-ui/[^/]+/v\d+/${escapedSuffix}$`
    ];

    return patterns.some((pattern) => new RegExp(pattern).test(path));
}
```

packages/project-access/src/project/cap.ts
```diff
@@ -424,8 +462,15 @@ function findServiceByUri(
...
 ): { name: string; urlPath: string } | undefined {
-    const searchUri = uniformUrl(uri).replace(/(?:^\/)|(?:\/$)/g, '');
-    return services.find((srv) => srv.urlPath.replace(/(?:^\/)|(?:\/$)/g, '') === searchUri);
+    const searchUri = normalizeServiceUrlPath(uniformUrl(uri));
+    // Try to find a service by exact path match
+    let service = services.find((srv) => normalizeServiceUrlPath(srv.urlPath) === searchUri);
+    // If no exact match is found, try matching while ignoring the service prefix
+    service ??= services.find((srv) => {
+        const normalizedServiceUrlPath = normalizeServiceUrlPath(srv.urlPath);
+        return isMatchingServiceUri(`/${searchUri}`, normalizedServiceUrlPath);
```
heimwege: Suggestion: normalize the leading slash inside isMatchingServiceUri

The caller manually prepends a / before passing searchUri to this function (because the regex patterns require ^/). This is an implicit contract that future callers will have to discover the hard way.

Consider normalizing inside the function instead:
```ts
const normalizedPath = path.startsWith('/') ? path : `/${path}`;
```
Then use normalizedPath in the .test() call, and the caller can simply pass searchUri directly. The leading-slash requirement becomes an implementation detail, not a caller responsibility.

.changeset/mighty-camels-drum.md
815are: fixed - 165a2629b760f817291ed19d9b19ce17cab59881

packages/project-access/src/project/cap.ts
815are: applied suggestion  - 800b7d830a27cc033277e7a90eaab497c84d6d95

packages/project-access/src/project/cap.ts
815are: applied suggestion  - 800b7d830a27cc033277e7a90eaab497c84d6d95

packages/project-access/src/project/cap.ts
```diff
@@ -413,6 +413,45 @@ export async function readCapServiceMetadataEdmx(
...
+ * The `<suffix>` (e.g. `odata/v4/myService`) must match exactly.
+ *
+ * @param path - The full request path to validate.
+ * @param expectedSuffixPath - The expected service path (e.g. `odata/v4/myService`).
+ * @returns `true` if the path matches one of the supported patterns and ends with the expected suffix.
+ */
+export function isMatchingServiceUri(path: string, expectedSuffixPath: string): boolean {
+    const normalizedPath = path.startsWith('/') ? path : `/${path}`;
+    // Escapes special regex characters in a string so it can be embedded into regular expression
+    const escapedSuffix = expectedSuffixPath.replace(/[.*+?^${}()|[\]\\]/g, String.raw`\$&`);
```
Klaus-Keller: @815are, could you check the sonar finding https://sonarcloud.io/project/issues?id=SAP_open-ux-tools&issues=AZ0pladmrzCpC1hZtc55&open=AZ0pladmrzCpC1hZtc55&pullRequest=4393

815are: applied correction - https://github.com/SAP/open-ux-tools/pull/4393/commits/4a0c5953bb605afc8c26454f230ce877c7b4b880

PR #4395, opened 2026-03-11
feat: Add sap-strict-uom-filtering ESLint rule to prevent disabling UOM filtering

marcovieth:
Add new ESLint rule to ensure `disableStrictUomFiltering` is not set to `true` in the `sap.fe.app` configuration within the `manifest.json` file of OData V4 applications (UI5 version 1.143 or higher).
This rule prevents incorrect filter behavior when users filter by unit-of-measure (UOM) fields by enforcing strict UOM filtering defaults.
Strict unit-of-measure filtering ensures that filter values are correctly associated with their units of measure, preventing inaccurate query results.


packages/eslint-plugin-fiori-tools/docs/rules/sap-strict-uom-filtering.md
```diff
@@ -0,0 +1,65 @@
+# Disallow Setting `disableStrictUomFiltering` to `true` in SAP FE App Configuration (`sap-strict-uom-filtering`)
```
AlinaGovoruhina, line 1: The code overall looks good to me. Maybe it is worth mentioning the rule applies only to OData V4?

packages/eslint-plugin-fiori-tools/src/rules/sap-strict-uom-filtering.ts
```diff
@@ -0,0 +1,61 @@
...
+                'The "disableStrictUomFiltering" property should not be set to true. Remove this property or set it to false to keep strict unit-of-measure filtering active.'
+        },
+        fixable: 'code'
+    },
+
+    check(context) {
+        const problems: StrictUomFiltering[] = [];
+
+        for (const [appKey, app] of Object.entries(context.sourceCode.projectContext.linkedModel.apps)) {
+            if (app.type !== 'fe-v4') {
```
AlinaGovoruhina, line 29: Since the uom field is required since SAPUI5 version 1.143, should there also be a check for app minUI5 version?

AlinaGovoruhina (commented):
Please see my comments.

packages/eslint-plugin-fiori-tools/src/rules/sap-strict-uom-filtering.ts
marcovieth: Good point. I put in a version check.

packages/eslint-plugin-fiori-tools/docs/rules/sap-strict-uom-filtering.md
marcovieth: It is mentioned in the documentation now.

igorsVoronins (approved):
QA passed ✅ 
The rule works as expected, highlighting the "disableStrictUomFiltering" property with warning when it is set to `true` in the application configuration. Auto-fix is removing property from the manifest restoring the strict UOM filtering behavior.

Image

PR #4398, opened 2026-03-11
TBI - IDE Type Detection for Telemetry

devinea:
Analytics & Insights: Understanding which IDEs users prefer help and bug fixes.

### DIFF
--- a/examples/odata-cli/CHANGELOG.md
+++ b/examples/odata-cli/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/odata-cli
 
+## 0.17.53
+
+### Patch Changes
+
+- Updated dependencies [2e17a6b]
+    - @sap-ux/btp-utils@1.1.12
+    - @sap-ux/axios-extension@1.25.27
+
 ## 0.17.52
 
 ### Patch Changes
--- a/examples/odata-cli/package.json
+++ b/examples/odata-cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/odata-cli",
-    "version": "0.17.52",
+    "version": "0.17.53",
     "description": "Simple example CLI uing the @sap-ux/axios-extension module to fetch metadata and annotations from an SAP system.",
     "license": "Apache-2.0",
     "private": true,
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,16 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.229
+
+### Patch Changes
+
+- Updated dependencies [2e17a6b]
+    - @sap-ux/btp-utils@1.1.12
+    - @sap-ux/axios-extension@1.25.27
+    - @sap-ux/system-access@0.7.3
+    - @sap-ux/fiori-elements-writer@2.8.112
+    - @sap-ux/fiori-freestyle-writer@2.5.82
+
 ## 1.1.228
 
 ### Patch Changes
--- a/examples/simple-generator/package.json
+++ b/examples/simple-generator/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/generator-simple-fe",
-    "version": "1.1.228",
+    "version": "1.1.229",
     "description": "Simple example of a yeoman generator for Fiori elements.",
     "license": "Apache-2.0",
     "private": true,
--- a/packages/abap-deploy-config-inquirer/CHANGELOG.md
+++ b/packages/abap-deploy-config-inquirer/CHANGELOG.md
@@ -1,5 +1,17 @@
 # @sap-ux/abap-deploy-config-inquirer
 
+## 1.8.5
+
+### Patch Changes
+
+- Updated dependencies [2e17a6b]
+    - @sap-ux/btp-utils@1.1.12
+    - @sap-ux/axios-extension@1.25.27
+    - @sap-ux/fiori-generator-shared@0.13.92
+    - @sap-ux/inquirer-common@0.11.30
+    - @sap-ux/nodejs-utils@0.2.19
+    - @sap-ux/system-access@0.7.3
+
 ## 1.8.4
 
 ### Patch Changes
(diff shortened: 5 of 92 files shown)


packages/telemetry/src/tooling-telemetry/data-processor.ts
```diff
@@ -338,3 +340,73 @@ async function getOSNodeVersion(): Promise<string> {
...
+    if (process.env.CURSOR_TRACE_ID || vscodeCwd.includes('cursor')) {
+        return 'cursor';
+    }
+
+    // Windsurf detection (Codeium's IDE)
+    if (vscodeCwd.includes('windsurf') || vscodeCwd.includes('codeium')) {
+        return 'windsurf';
+    }
+
+    // Antigravity detection (Salesforce's IDE)
```
mikicvi-SAP: I believe Antigravity is Google's IDE.

SF's is "Agentforce", ex "Code Builder"

devinea: thanks @mikicvi-SAP .
Fixed https://github.com/SAP/open-ux-tools/pull/4398/commits/fe74498a45ecb1d8d25529a060aa47c2dcfa7f55

devinea:
Updated to also consider `process.env.VSCODE_APPNAME`.

packages/telemetry/src/tooling-telemetry/data-processor.ts
```diff
@@ -338,3 +340,92 @@ async function getOSNodeVersion(): Promise<string> {
...
+    // code-server detection (browser-based VS Code)
+    if (appName.includes('code-server') || vscodeCwd.includes('code-server') || process.env.CODE_SERVER_SESSION) {
+        return 'code-server';
+    }
+
+    // VSCode Insiders detection — match specific product names to avoid false positives
+    // from paths that merely contain the substring "insiders"
+    if (
+        appName.includes('insiders') ||
+        vscodeCwd.includes('code - insiders') ||
```
heimwege, line 421: On Windows, the Insiders application path is typically `C:\Users\...\AppData\Local\Programs\Microsoft VS Code Insiders\`. This does not contain `'code - insiders'` (with spaces and lowercase). The `'vscode-insiders'` check would also not match. The TERM_PROGRAM and VSCODE_APPNAME checks do handle this case, but if neither is set on Windows, Insiders would fall through to `'vscode'`.

Suggestion: Add `'code insiders'` (no dash) as an additional substring check, or add a Windows-style path variant for documentation completeness.

packages/telemetry/src/tooling-telemetry/types.ts
```diff
@@ -41,6 +41,23 @@ export enum DeployTarget {
     NO_DEPLOY_CONFIG = 'NO_DEPLOY_CONFIG'
 }
 
+/**
+ * IDE/editor/ types that can host the telemetry client.
```
heimwege: There is a stray trailing slash in "IDE/editor/ types"

packages/telemetry/src/tooling-telemetry/data-processor.ts
```diff
@@ -338,3 +340,92 @@ async function getOSNodeVersion(): Promise<string> {
...
+export function getIdeType(): IdeType {
+    // AppStudio is a VS Code fork that sets VSCODE_PID — check first to avoid misdetection as 'vscode'
+    if (isAppStudio()) {
+        return 'appstudio';
+    }
+
+    const vscodeCwd = process.env.VSCODE_CWD?.toLowerCase() ?? '';
+    const vscodePid = process.env.VSCODE_PID;
+    const termProgram = process.env.TERM_PROGRAM ?? '';
+    const appName = process.env.VSCODE_APPNAME?.toLowerCase() ?? '';
```
heimwege, line 369: `VSCODE_APPNAME` is not documented as an official VS Code environment variable. VS Code does expose `productName` through its API (accessible via extensions), but it does not set a `VSCODE_APPNAME` process environment variable in typical configurations. If this var is never actually set, all the `appName.includes(...)` checks silently fall through to the CWD-based fallbacks, which is fine — but the JSDoc comment says it is checked "first as the most reliable signal," which may mislead future maintainers if it is unreliable or rarely set.

Suggestion: Add a comment or test clarifying under which conditions `VSCODE_APPNAME` is populated (e.g., is it set by specific forks, or is it injected by the SAP extension itself?). If it is not reliably set by the host environment, demote it in the comment from "most reliable signal" to "opportunistic hint."

packages/telemetry/src/tooling-telemetry/data-processor.ts
devinea: Fixed — added  (no dash) as a substring check in commit 32b846ef6 (already on the branch). Also added a Windows-style VSCODE_CWD test:  in the test file (commit 8839ac171).

packages/telemetry/src/tooling-telemetry/types.ts
devinea: Fixed — removed the trailing slash in commit 32b846ef6. The type comment now reads "IDE/editor types".

packages/telemetry/src/tooling-telemetry/data-processor.ts
devinea: Good catch. Updated the JSDoc in commit 32b846ef6 to demote `VSCODE_APPNAME` from "most reliable signal" to "opportunistic hint" with a note that it is set by some VS Code forks (e.g. Cursor, Windsurf) that expose their Electron `productName` via this variable, but is not a standard VS Code environment variable and may be absent in plain VS Code or CLI contexts.

devinea:
Re: the question about non-VS Code environments (IntelliJ, bash, PowerShell, etc.) — returning `'unknown'` is intentional. The purpose of `getIdeType()` is to differentiate between VS Code and its forks (Cursor, Windsurf, etc.) where SAP Fiori Tools extensions run. For environments where VS Code signals are absent (CLI, shell), `'unknown'` is the correct and meaningful value — it tells the telemetry pipeline "this ran outside any VS Code-based host".
MCP server is a special case as we capture `mcpClientName` in telemetry to capture clients like `IntelliJ`, etc..)

Issue #4404, opened 2026-03-13
BUG - Add New Card flow broken for OVP adaptation projects


testojs:
### Description
When previewing an OVP (Overview Page) adaptation project in the Adaptation Editor, the "Add New Card" context menu action fails because three global bridge functions expected by the `sap.ovp` library are not available on `window`: `writeToI18n`, `getNewDataSources`, and `getMetaModelForNewDataSource`. This prevents users from adding new cards with new data sources.

### Steps to Reproduce
1. Create or open an OVP adaptation project in VS Code / BAS
2. Start the Adaptation Editor preview
3. Right-click on an OVP card
4. Select "Add New Card" from the context menu
5. In the Card Settings dialog, select "Select New Data Source"
6. Observe that the data source dropdown does not populate and console shows `getNewDataSources is not defined`

### Expected results
- Data source dropdown populates with available OData services from the ABAP catalog
- Selecting a service loads its metadata and annotations, populating entity set options
- Saving the card writes translatable texts to the i18n properties file

### Actual results
- `getNewDataSources is not defined` error in console
- Card Settings dialog cannot proceed with new data source configuration
- No way to add a card with a new data source in the adaptation editor

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment
Add any other context about the problem here
OS:
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
The OVP runtime (`sap.ovp` library) expects three global functions on `window` to support the "Add New Card" flow: `writeToI18n`, `getNewDataSources`, and `getMetaModelForNewDataSource`. These existed in the old WebIDE implementation via a custom script (`GenericUtils.js`) injected into the preview iframe, but were never implemented in open-ux-tools.

### Fix
Implemented OVP bridge functions that register the three globals on `window` during ADP initialization when the app is detected as an OVP application (`manifest['sap.ovp']`). The client-side functions use `fetch()` to call new server-side API routes (`/adp/api/ovp/datasources`, `/adp/api/ovp/metamodel`) which leverage the existing `AbapServiceProvider` and V2 catalog service. For the metamodel, the client creates a real `sap.ui.model.odata.v2.ODataModel` with annotation URLs to produce a full metamodel with merged annotations. The i18n write reuses the existing `/editor/i18n` route.

### Why was it missed
[...]

PR #4405, opened 2026-03-13
fix: Add New Card flow broken for OVP adaptation projects

testojs:
Fix for #4404.

- Register OVP bridge functions (writeToI18n, getNewDataSources, getMetaModelForNewDataSource) during ADP init for OVP adaptation projects
- Add server-side API routes for V2 catalog service lookup and annotation retrieval in adp-tooling
- Build metamodel client-side using a real UI5 ODataModel with annotations to ensure full compatibility with the OVP Add New Card dialog

### DIFF
--- a/.changeset/giant-files-hope.md
+++ b/.changeset/giant-files-hope.md
@@ -1,7 +1,0 @@
----
-'@sap-ux/axios-extension': patch
-'@sap-ux/generator-adp': patch
-'@sap-ux/adp-tooling': patch
----
-
-feat: Handle i18n translations from key user changes in ADP generator
--- a/examples/odata-cli/CHANGELOG.md
+++ b/examples/odata-cli/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/odata-cli
 
+## 0.17.50
+
+### Patch Changes
+
+- Updated dependencies [a854433]
+    - @sap-ux/axios-extension@1.25.24
+
 ## 0.17.49
 
 ### Patch Changes
--- a/examples/odata-cli/package.json
+++ b/examples/odata-cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/odata-cli",
-    "version": "0.17.49",
+    "version": "0.17.50",
     "description": "Simple example CLI uing the @sap-ux/axios-extension module to fetch metadata and annotations from an SAP system.",
     "license": "Apache-2.0",
     "private": true,
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.218
+
+### Patch Changes
+
+- Updated dependencies [a854433]
+    - @sap-ux/axios-extension@1.25.24
+    - @sap-ux/system-access@0.6.66
+
 ## 1.1.217
 
 ### Patch Changes
--- a/examples/simple-generator/package.json
+++ b/examples/simple-generator/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/generator-simple-fe",
-    "version": "1.1.217",
+    "version": "1.1.218",
     "description": "Simple example of a yeoman generator for Fiori elements.",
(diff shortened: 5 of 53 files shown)


packages/adp-tooling/src/preview/ovp-routes-handler.ts
```diff
@@ -0,0 +1,125 @@
...
+
+            const serviceUrl = `/sap/opu/odata/sap/${dataSource.Title}/`;
+            res.status(HttpStatusCodes.OK).json({
+                serviceUrl,
+                annotations: annotations.map((a) => ({
+                    TechnicalName: a.TechnicalName,
+                    Uri: toRelativePath(a.Uri)
+                })),
+                modelInformation: {
+                    serviceURI: `/sap/opu/odata/sap/${dataSource.Title}`,
```
avasilev-sap: Why not: 
```ts
const serviceURI = `/sap/opu/odata/sap/${dataSource.Title}`;
...
serviceUrl: `${serviceURI}/`
modelInformation: {
   serviceURI
}
```
? Do we need the last slash at all?

testojs: Good catch, it will be a bit more DRY. The trailing slash is needed for `serviceUrl` because it's passed to new `sap.ui.model.odata.v2.ODataModel()` which expects the service root to end with `/`.

Fixed in https://github.com/SAP/open-ux-tools/pull/4405/commits/df3fe0a63f6afd43ea46154776426b8d7152ccf0

packages/preview-middleware-client/src/adp/ovp-window-functions.ts
```diff
@@ -0,0 +1,171 @@
...
+
+interface OvpMetaModelResult {
+    oEntityContainers: Record<string, unknown>;
+    oSchema: Record<string, unknown>[];
+    modelInformation: OvpModelInformation;
+}
+
+declare global {
+    interface Window {
+        writeToI18n: (sPath: string, aProperties: I18nProperty[]) => void;
```
nikmace: The following parameter names and local variables use **SAP Hungarian notation** (s = string, a = array, o = object), which doesn't align with TypeScript conventions. These are all internal (positional params / local vars).

- sPath → path (window interface, 3 occurrences)
- aProperties → properties (window interface + writeToI18n param)
- oDataSourceSelected → selectedDataSources (window interface + getMetaModelForNewDataSource param and
  usage)
- _sPath → _path (writeToI18n, getNewDataSources, getMetaModelForNewDataSource params)
- oModel → model (buildMetaModel local var)
- oMetaModel → metaModel (buildMetaModel local var and usages)

For the `OvpMetaModelResult` interface, they are fine because that is what the OVP runtime expects. I remember we talked about the Hungarian notations a few years ago, what do you think about this change @testojs?

packages/adp-tooling/test/unit/preview/ovp-routes-handler.test.ts
```diff
@@ -0,0 +1,138 @@
...
+const mockCatalogService = {
+    get: jest.fn().mockResolvedValue({
+        odata: () => mockServices
+    }),
+    getAnnotations: jest.fn().mockResolvedValue(mockAnnotations)
+};
+
+const mockProvider = {
+    catalog: jest.fn().mockReturnValue(mockCatalogService)
+} as any;
```
nikmace: Could we cast it as AbapServiceProvider or Partial<AbapServiceProvider>?

avasilev-sap (dismissed):

packages/adp-tooling/test/unit/preview/ovp-routes-handler.test.ts
testojs: Changed to `as unknown as AbapServiceProvider` — avoids `any` while keeping the mock minimal (only the methods used by the handler are mocked).
Fixed in daa0a92d5458f828323940389fc08f4af292fac7

packages/preview-middleware-client/src/adp/ovp-window-functions.ts
testojs: Agree.
Fixed in daa0a92d5458f828323940389fc08f4af292fac7.

PR #4406, opened 2026-03-13
Feat(eslint-plugin-fiori-tools): adjust readme

heimwege:
Add the eslint-plugin-fiori-tools version number for each rule so that it is visible when it has been added


.changeset/tough-melons-spend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/eslint-plugin-fiori-tools': patch
+---
+
+feat: add plugin version number to rule in readme
```
hitesh-parmar, line 5: Maybe we can use `docu` as a prefix here.

```suggestion
docu: add plugin version number to rule in readme
```

heimwege: `docu` is not mentioned in the conventional commit we adhere to (see https://github.com/SAP/open-ux-tools/blob/main/docs/Guidelines.md#git-guidelines).
We could use `docs` but to be honest: what's the benefit? We are talking about an entry for a patch version change in the [CHANGELOG.md](https://github.com/SAP/open-ux-tools/blob/main/packages/eslint-plugin-fiori-tools/CHANGELOG.md)

PR #4407, opened 2026-03-13
fix(security): upgrade simple-git

donal-tobin-sap:
- via promptfoo upgrade

Issue: https://github.com/SAP/open-ux-tools/security/dependabot/427


donal-tobin-sap:

- critical security fix

PR #4412, opened 2026-03-13
chore(generator-odata-downloader): move folder fix publish

IainSAP:
Relates to: https://github.com/SAP/open-ux-tools/issues/4266

- Renames package folder to match package name
- Renames package to avoid illegal word `download` !
- Adds telem key replacement pipeline task
- Removes `noImplicityAny` rule setting
- Inlcude bundle in build task


devinea (approved):
Thanks @IainSAP for the updates.
Lg.

PR #4413, opened 2026-03-14
Fix missing readme links. Move most deps -> devDeps

IainSAP:
Chore: Fix broken and missing readme links

- Fixes readme links
- Moves all deps to dev deps. Stubs out unused depends from yeoman-generator to avoid complex deps.
- Include prebuilds from zowe for non-BAS platforms


devinea:
See here for more bundling : 
https://github.com/SAP/open-ux-tools/compare/chore/generator-odata-downloader/fix_links_bundle_deps...chore/generator-odata-downloader/fix_links_bundle_deps_ADE?expand=1

PR #4419, opened 2026-03-16
fix(ADP): Change prompt's order, first appear applications then projectType if applicable.

avasilev-sap:
#4391

* Change the order the application and project type prompts appear. First we display the application prompt, second the project type if certain conditions are met. The below flowchart illustrates the new workflow:

Detect project type

* In order to implement the above workflow we need to change slightly the filters for the application list. In case the system is mixed we need to include for each application result the `cloudDevAdaptationStatus` property. We should not include that property for none mixed systems because for older ABAP systems which support only onPremise project type the property is not available as a column which results with an error 400 for the entire api call which load the application list.

* In the application validation and additionalMessage, before the Private Cloud feature to take place, we used the projectType. In Private cloud scenario, when the system is mixed and the application is released (cloudDevAdaptationStatus := 'released'), we don't have the project type yet selected when we do the application validation due to the order of prompts, so this leads to changes in the implementation and the UX related to the additional messages. In Private cloud scenario we postpone the application validation, the validation actually happens when the user selects the project type, all additional messages which appear beneath the application prompt now appear beneath the project type prompt. For onPremise or cloud only systems there are no changes in the validation and additional messages UX. Because of a limitation in yeoman UI we cannot update the application prompt additional messages while changing the project type prompt value, that is why we need to put the application validation and additional messages inside the project type prompt.

* Fix a small issue with the json generator, the project type must be set to onPremise when the user is internal. The user cannot override this from the json config and should not.

**Testing done**

* Manually verified on internal systems - mixed, clod and onPrem only and older ABAP systems with external/internal User. 

* VSCode/BAS

Mixed system, app.cloudDevAdaptatuionStatus := '' flex not supported 

Screenshot 2026-03-18 at 16 11 45

Screenshot 2026-03-18 at 16 19 02

Screenshot 2026-03-18 at 16 19 19

Mixed system, released app, internal usage

Screenshot 2026-03-18 at 14 20 01

Screenshot 2026-03-18 at 16 20 35

Mixed system, released app, onPrem project

Screenshot 2026-03-18 at 14 28 01

Screenshot 2026-03-18 at 16 21 20

Mixed system, released app, cloud project

Screenshot 2026-03-18 at 14 28 29

Screenshot 2026-03-18 at 16 21 43

Mixed system, , app.cloudDevAdaptatuionStatus := ''

Screenshot 2026-03-18 at 14 29 04

Screenshot 2026-03-18 at 16 22 15

Onprem only system

Screenshot 2026-03-18 at 14 29 47

Screenshot 2026-03-18 at 16 22 45

Cloud only system

Screenshot 2026-03-18 at 14 30 24

Screenshot 2026-03-18 at 16 23 11

Deploy wizard

Screenshot 2026-03-25 at 11 05 17

### DIFF
--- a/packages/create/CHANGELOG.md
+++ b/packages/create/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/create
 
+## 0.15.54
+
+### Patch Changes
+
+- Updated dependencies [8e7d529]
+    - @sap-ux/preview-middleware@0.25.4
+    - @sap-ux/app-config-writer@0.6.128
+
 ## 0.15.53
 
 ### Patch Changes
--- a/packages/create/package.json
+++ b/packages/create/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@sap-ux/create",
     "description": "SAP Fiori tools module to add or remove features",
-    "version": "0.15.53",
+    "version": "0.15.54",
     "repository": {
         "type": "git",
         "url": "https://github.com/SAP/open-ux-tools.git",
--- a/packages/preview-middleware/CHANGELOG.md
+++ b/packages/preview-middleware/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/preview-middleware
 
+## 0.25.4
+
+### Patch Changes
+
+- 8e7d529: fix(preview-middleware): ADP Extension Points: blank iframe on reload for UI5 < 1.120
+
 ## 0.25.3
 
 ### Patch Changes
--- a/packages/preview-middleware/package.json
+++ b/packages/preview-middleware/package.json
@@ -9,7 +9,7 @@
     "bugs": {
         "url": "https://github.com/SAP/open-ux-tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug+label%3Apreview-middleware"
     },
-    "version": "0.25.3",
+    "version": "0.25.4",
     "license": "Apache-2.0",
     "author": "@SAP/ux-tools-team",
     "main": "dist/index.js",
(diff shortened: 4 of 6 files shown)


packages/generator-adp/src/translations/generator-adp.i18n.json
```diff
@@ -43,8 +43,9 @@
         "projectNameLabel": "Project Name",
         "projectNameTooltip": "Enter the project name for your application variant.",
         "projectTypeListLabel": "Select Project Type",
-        "projectTypeCloudReadyName": "Cloud Ready",
-        "projectTypeOnPremName": "onPremise",
+        "projectTypeCloudReadyName": "cloud ready",
```
```suggestion
        "projectTypeCloudReadyName": "Cloud Ready",
```

packages/generator-adp/src/translations/generator-adp.i18n.json
```diff
@@ -43,8 +43,9 @@
...
+        "projectTypeOnPremName": "classic",
```
```suggestion
        "projectTypeOnPremName": "Classic",
```

packages/generator-adp/src/translations/generator-adp.i18n.json
```diff
@@ -43,8 +43,9 @@
...
+        "projectTypeClassicLabel": "Project Type: classic",
```
```suggestion
        "projectTypeClassicLabel": "Project Type: Classic",
```

packages/generator-adp/src/translations/generator-adp.i18n.json
avasilev-sap: fixed in 661ed0bc86c5bbca4bf4ca286a9b4bbd6ddace73

packages/generator-adp/src/translations/generator-adp.i18n.json
avasilev-sap: fixed in 661ed0bc86c5bbca4bf4ca286a9b4bbd6ddace73

packages/generator-adp/src/translations/generator-adp.i18n.json
avasilev-sap: fixed in 661ed0bc86c5bbca4bf4ca286a9b4bbd6ddace73

PR #4420, opened 2026-03-16
feat(ui-components): `UITranslationInput` — new property `invertedCalloutTheme` to toggle between normal and inverted callout styles. Default is `false`. Previously `invertedCalloutTheme` was applied by default.

815are:
`UITranslationInput` — new property `invertedCalloutTheme` to toggle between normal and inverted callout styles. Default is `false`. Previously `invertedCalloutTheme` was applied by default.
From storybook:
image
image


Jimmy-Joseph19 (approved):
Changes looks good.
Changeset found.
Tests coverage good.
Did not test locally.
Approving

Issue #4249, opened 2026-02-25
TBI - All newly generated applications should have the desired eslint configuration


longieirl:
### Description (include screenshots)
`As a Fiori developer using SAP Fiori tools, I shall be informed about inconsistencies or violations of rules in my app, and being offered suitable options to speed up my development.`

Acceptance Criteria
WHEN I generate a new app using the app generator
THEN the generated app should have the desired ESLint configuration based on the app (S4, non-S4, customer)

### Value
Optional - details on how this technical backlog item provides value to end-users or dev team eg. Performance, Quality, Maintainability. This will help determine the ranking in relation to feature development

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [x] Remove eslint check box from the generator - eslint config option should be enabled always when calling the writer.
- [x] User cannot opt out. TypeScript and Javascript projects will both have eslint.
- [x] For CAP projects the generator should also write eslint config for TS and JS projects. Also add the lint script in the package.json (not currently there).
- [x] Migration should not offer to add eslint. WebIde projects do not have eslint. A separate feature request will cover this flow.

longieirl:
Bug found when testing;

the readme.md template is using an incorrect URL; should be updated to;
```
  |**Add Eslint configuration**<br><% if(locals.enableEslint === true) { %><%= "True<br/>, see https://www.npmjs.com/package/@sap-ux/eslint-plugin-fiori-tools#rules for the eslint rules." %><% } else { %><%= "False" %><% }%>|<% if(locals.valueHelpDownloaded === true) { %>
```

longieirl:
Currently tested and validated on latest tools-suite release.

PR #4434, opened 2026-03-18
fix(fiori-generator-shared): update ESLint plugin URL in README template

longieirl:
Fix for https://github.com/SAP/open-ux-tools/issues/4249

Update the ESLint configuration link in the README template from the deprecated eslint-plugin-fiori-custom to @sap-ux/eslint-plugin-fiori-tools#rules, and add a test covering the enableEslint=true branch with updated snapshot.


heimwege:
two approvals, only readme change -> using admin

Issue #4435, opened 2026-03-19
FEATURE - Adapt CF ADP project structure to work with approuter backend middleware


nikmace:
_As a **developer**,
I want to **create a CF ADP project**,
so that I **can use and start the Adaptation Editor with approuter middleware**_

### Description

Project structure for Cloud Foundry adaptation projects need to be updated in order to work with the reworked `backend-proxy-middleware-cf`, which uses `approuter` underneath.

### Acceptance Criteria
GivenWhenThen format (https://martinfowler.com/bliki/GivenWhenThen.html)

**Given** I have created a CF ADP
**when** I run the build and start-editor commands
**then** the adaptation editor will start, odata will load, and api requests will successfully execute

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #4436, opened 2026-03-19
feat: Adapt CF ADP project structure to work with approuter backend middleware

nikmace:
Feat for #4435.
- Adds correct `backend-proxy-middleware-cf` project ui5.yaml configuration.
- Refactors the `setup-cf` command, removes code.
- Cleans up methods that are now obsolete.

### DIFF
--- a/packages/eslint-plugin-fiori-tools/CHANGELOG.md
+++ b/packages/eslint-plugin-fiori-tools/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sap-ux/eslint-plugin-fiori-tools
 
+## 9.11.5
+
+### Patch Changes
+
+- 0f7f5f3: Fix: Check minUI5 version for the sap-width-including-column-header rule
+
 ## 9.11.4
 
 ### Patch Changes
--- a/packages/eslint-plugin-fiori-tools/docs/rules/sap-width-including-column-header.md
+++ b/packages/eslint-plugin-fiori-tools/docs/rules/sap-width-including-column-header.md
@@ -1,6 +1,6 @@
 # Require `widthIncludingColumnHeader` for Small Tables (`sap-width-including-column-header`)
 
-Ensures that small tables (less than six columns) include the `widthIncludingColumnHeader` property set to `true` for improved calculation of the column width.
+Ensures that small tables (less than six columns) include the `widthIncludingColumnHeader` property set to `true` for improved calculation of the column width. The rule is applicable to OData V4 applications with a minimum SAPUI5 version of 1.120.
 
 ## Rule Details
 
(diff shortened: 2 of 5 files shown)


lfindlaysap:
@nikmace, can you let me know which files should be reviewed?

nikmace:
> @nikmace, can you let me know which files should be reviewed?

@lfindlaysap None are needed to be reviewed, this PR's branch is pointing to another branch where we potentially have text changes, that's why the team for text reviews was added for this PR.

packages/adp-tooling/src/writer/cf.ts
```diff
@@ -170,10 +169,7 @@ export async function generateCfConfig(
     }
 
     await writeUi5AppInfo(basePath, ui5AppInfo, logger);
-    await addServeStaticMiddleware(basePath, ui5Config, logger);
     await runBuild(basePath, { ADP_BUILDER_MODE: 'preview' });
-    addBackendProxyMiddleware(basePath, ui5Config, serviceInfo.serviceKeys, logger);
 
-    fs.write(join(basePath, yamlPath), ui5Config.toString());
     return fs;
```
mmilko01: Do we still need the `fs` reference at all?

packages/adp-tooling/templates/cf/package.json
```diff
@@ -23,11 +23,11 @@
 	"devDependencies": {
 		"@sap-ux/create": "^0.15.9",
 		"@sap/ui5-builder-webide-extension": "^1.1.9",
-		"@sap-ux/backend-proxy-middleware-cf": "~0.0.0",
+		"@sap-ux/backend-proxy-middleware-cf": "^0.0.86",
```
mmilko01: Should we bump the `@sap-ux/backend-proxy-middleware-cf` minor version, so old project don't automatically break and define the version here as `^0.1.0`?

nikmace: You are probably right. I am not sure if ^ is enough for 0.0.0 versions to load the latest minor version. If we point it to ^0.1.0 right now, before the version is released, when you try to create the project, the generator will fail in the install phase because this version is not found yet, so difficult for testing, but not a problem in general

packages/adp-tooling/src/writer/cf.ts
nikmace: Yea we don't need to pass it anymore

nikmace: Changed in [aab8f2f](https://github.com/SAP/open-ux-tools/pull/4436/commits/aab8f2fb566a6c9ba23c8bfa023c3782c965bfba)

packages/adp-tooling/templates/cf/package.json
nikmace: Changed in [aab8f2f](https://github.com/SAP/open-ux-tools/pull/4436/commits/aab8f2fb566a6c9ba23c8bfa023c3782c965bfba)

avasilev-sap (approved):

devinea (approved):
Re-approving

PR #4440, opened 2026-03-19
fix: disable condensing in workspace connector for older SAPUI5 versions

mmilko01:
#4443

* Condensing feature is disabled in Workspace Connector for SAPUI5 versions lower than 1.108.X


mmilko01:
image

PR #4456, opened 2026-03-24
Fix(project-access): exec npm command logging

heimwege:
The original code conflated "wrote to stderr" with "something went wrong", which is not how Unix processes are supposed to work.

### DIFF
--- a/packages/project-access/src/command/npm-command.ts
+++ b/packages/project-access/src/command/npm-command.ts
@@ -48,6 +48,9 @@ export async function execNpmCommand(
                 const output = stdMessages ? `:\n${stdMessages}` : '';
                 logger?.info(`Command '${commandString}' successful${output}`);
             } else if (options?.throwOnError) {
+                if (stdOut) {
+                    logger?.info(stdOut);
+                }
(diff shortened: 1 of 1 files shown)


packages/project-access/src/command/npm-command.ts
```diff
@@ -37,20 +39,24 @@ export async function execNpmCommand(
...
+        const commandString = `${npmCommand} ${commandArguments.join(' ')}`;
+        spawnProcess.on('exit', (code, signal) => {
+            if (code === null) {
+                logger?.warn(`Command '${commandString}' was killed by signal: ${signal}`);
+            } else if (code === 0) {
+                const stdMessages = [stdOut, stdErr].filter(Boolean).join('\n');
+                const output = stdMessages ? `:\n${stdMessages}` : '';
+                logger?.info(`Command '${commandString}' successful${output}`);
+            } else if (options?.throwOnError) {
+                reject(new Error(`Command '${commandString}' failed with exit code ${code}. Stderr: ${stdErr}`));
```
Klaus-Keller, line 54: could we still log the `stdOut`

heimwege: done

PR #4465, opened 2026-03-25
Fix(preview-middleware-client): rta causing duplicate ID when navigating from launchpad

heimwege:
### Issue

When opening the RTA (UI Adaptation) editor endpoint URL without an intent hash in the URL (e.g. /editor.html instead of /editor.html#app-preview), the user lands on the FLP sandbox home page. Navigating from there to the actual app  triggered the error:

  ▎ "UI Adaptation could not be started - Error: adding element with duplicate id 'sapUIRta_toolbar'"

### Root cause

The attachAppLoaded callback in flp/init.ts fires for every component loaded in the FLP shell — including  the FLP home component (#Shell-home). Both the home component load and the subsequent app navigation each called startAdaptation, which attempted to create a RuntimeAuthoring toolbar with the fixed UI5 element ID sapUIRta_toolbar.  The second creation failed with a duplicate ID error.

### Fix

Added a guard at the start of the attachAppLoaded callback that skips RTA initialization when the URL hash is empty or equals #Shell-home. RTA will only start when the FLP has navigated to an actual application intent.

### Code quality

All usages of window.location in flp/init.ts (hash, search, reload) have been replaced with globalThis.location as required by Sonar. In jsdom (the Jest environment), globalThis === window, so no test changes were required for this. The eslint.config.js was updated to ecmaVersion: 2020.

### Test corrections

While fixing the failing tests caused by the guard (which requires a non-empty, non-Shell-home hash), a pre-existing bug in the 'flex configured & ui5 version is 1.71.60' test was uncovered and corrected.

For UI5 versions below 1.84, initConnectors() calls sap.ui.require with enableFakeConnector during init() setup — before attachAppLoaded fires. This means sap.ui.require.mock.calls[0] is the enableFakeConnector call, and the actual initRta callback is at index [1]. The original test was using mock.calls[0][1], accidentally extracting the wrong callback and silently testing enableFakeConnector instead of initRta. The index has been corrected to mock.calls[1][1].


heimwege:
@schiwekM FYI

heimwege:
@ytpo-lyne @nikmace @devinea I had to resolve merge conflicts. please re-review

@devinea additionally I added `es2020` globals to `eslint-plugin-fiori-tools` to support e.g. `globalThis` and added the @sap-ux/create convert command to the readme.

devinea (approved):
re-approving

PR #4470, opened 2026-03-26
fix(eslint): check min UI5 version for rule sap-width-including-column-header

marcovieth:
The sap-width-including-column-header rule was being applied regardless of the UI5 version configured in the application's manifest.json. The widthIncludingColumnHeader property is only supported since UI5 1.120.0, so the rule should not report warnings for applications targeting earlier versions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -60,6 +60,39 @@ This uses `check-dependency-version-consistency` to validate that all packages u
 
 **Rule**: When updating a dependency in one package, update it in ALL packages that use it.
 
+**Exception — peerDependencies must keep open semver ranges:**
+
+`peerDependencies` declare compatibility with a host package the consumer provides. They must use **open caret (`^`) or tilde (`~`) ranges**, never exact (pinned) versions. Pinning a peerDependency forces every consumer to install that exact version, which breaks the moment they upgrade and defeats the purpose of peer ranges.
+
+```jsonc
+// ✅ Correct — open range
+"peerDependencies": {
+  "eslint": "^9"
+}
+
+// ❌ Wrong — pinned version
+"peerDependencies": {
+  "eslint": "9.39.1"
+}
+```
+
+**Guidelines:**
+- Use `^major` (e.g., `^9`) when the package works across the entire major version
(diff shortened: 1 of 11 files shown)


packages/eslint-plugin-fiori-tools/docs/rules/sap-width-including-column-header.md
```diff
@@ -1,6 +1,6 @@
 # Require `widthIncludingColumnHeader` for Small Tables (`sap-width-including-column-header`)
 
-Ensures that small tables (less than six columns) include the `widthIncludingColumnHeader` property set to `true` for improved calculation of the column width.
+Ensures that small tables (less than six columns) include the `widthIncludingColumnHeader` property set to `true` for improved calculation of the column width. The rule is applicable to OData V4 applications with SAP UI5 version 1.120.0 or higher.
```
lfindlaysap:
```suggestion
Ensures that small tables (less than six columns) include the `widthIncludingColumnHeader` property set to `true` for improved calculation of the column width. The rule is applicable to OData V4 applications with a minimum SAPUI5 version of 1.120.
```

marcovieth: Done with commit dda373a

.changeset/small-bags-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/eslint-plugin-fiori-tools': minor
```
AlinaGovoruhina: You can delete this `minor` changeset .md file, as we are going with a patch.

marcovieth: Done. Thanks.

.changeset/public-pandas-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/eslint-plugin-fiori-tools': patch
+---
+
+Fix: Check min UI5 version for rule sap-width-including-column-header
```
lfindlaysap:
```suggestion
Fix: Check minUI5 version for the sap-width-including-column-header rule
```

PR #4471, opened 2026-03-26
fix(btp-utils,deploy-tooling): allow deployment to OnPremise destinations with odata_gen WebIDEUsage

### LINKED ISSUE #2272: BUG - unable to deploy to onpremise destination when webideusage is configured as odata_gen
### Description

When deploying to a SAP BTP destination the following error is thrown;

```
error builder:custom deploy-to-abap Cannot read properties of undefined (reading 'bind')
Command deploy failed with error : Cannot read properties of undefined (reading 'bind')
```

The reason is due to the SAP BTP destination configured with `odata_gen` which causes the following method to return false;

https://github.com/SAP/open-ux-tools/blob/main/packages/axios-extension/src/factory.ts#L209

Sample destination;
```JSON
    {
        "Type": "HTTP",
        "HTML5.DynamicDestination": "true",
        "Authentication": "BasicAuthentication",
        "WebIDEUsage": "odata_gen",
        "Name": "MYTESTDEST",
        "WebIDEEnabled": "true",
        "ProxyType": "OnPremise",
        "WebIDESystem": "SAP",
        "Host": "http://billing.hana.ondemand.com:50001"
    }
```

Sample ui5-deploy;
```yaml
# yaml-language-server: $schema=https://sap.github.io/ui5-tooling/schema/ui5.yaml.json

specVersion: "3.1"
metadata:
  name: hansenpoc02
type: application
builder:
  resources:
    excludes:
      - /test/**
      - /localService/**
  customTasks:
    - name: deploy-to-abap
      afterTask: generateCachebusterInfo
      configuration:
        log: 4
        verbose: true
        test: true
        createTransport: true
        target:
          destination: MYTESTDEST
          url: http://billing.hana.ondemand.com:50001       
        app:
          name: ztest
          description: Test
          package: testpackage
          transport: testpackage
        exclude:
          - /test/

```

To unblock the deployment step, change `WebIDEUsage` to `odata_abap`.

### Steps to Reproduce
Steps to reproduce the behaviour:
1. Create a destination with `WebIDEUsage` configured as `odata_gen` 
2. Create any type of project
3. Ensure to add ABAP deployment
4. Run deployment step `npm run deploy`

### Expected results

To deploy correctly

### Actual results

Deployment fails and there is no explanation.

OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

longieirl:
## Summary

Fixes #2272

- **`@sap-ux/btp-utils`**: `isAbapSystem` now returns `true` when `destination.ProxyType === 'OnPremise'`. OnPremise BTP destinations connect via SAP Cloud Connector to on-premise SAP systems (always ABAP backends). The `WebIDEUsage` property describes catalog intent and is not a reliable indicator for deployment routing.
- **`@sap-ux/deploy-tooling`**: `getDeployService` now accepts an `undefined` factory function (via optional chaining at call sites) and throws a clear, actionable error naming the destination and listing the properties required to resolve it — instead of crashing with `Cannot read properties of undefined (reading 'bind')`.

## Test plan

- [ ] `pnpm --filter @sap-ux/btp-utils test` — new test: `isAbapSystem` returns `true` for `ProxyType=OnPremise` with `WebIDEUsage=odata_gen`
- [ ] `pnpm --filter @sap-ux/deploy-tooling test` — new test: `getDeployService` throws actionable error when provider has no `getUi5AbapRepository`
- [ ] Deploy to a BTP destination configured with `WebIDEUsage: "odata_gen"` and `ProxyType: "OnPremise"` completes successfully


packages/deploy-tooling/src/base/deploy.ts
```diff
@@ -173,10 +173,17 @@ async function axiosErrorRetryHandler(
...
  */
 function getDeployService<T extends Ui5AbapRepositoryService | LayeredRepositoryService>(
-    factoryFn: (alias?: string) => T,
+    factoryFn: ((alias?: string) => T) | undefined,
     config: AbapDeployConfig,
     logger: Logger
 ): T {
+    if (typeof factoryFn !== 'function') {
+        throw new Error(
+            `The destination '${config.target?.destination}' is not recognised as an ABAP system. ` +
```
lfindlaysap: Can you provide a screenshot of what it looks like?

longieirl: Currently, this is the flow that is being cleaned up;
```
info abap-deploy-task JL1REMOVEMMARCH Creating archive with UI5 build result.

info abap-deploy-task JL1REMOVEMMARCH Archive created.
info abap-deploy-task JL1REMOVEMMARCH Starting to deploy.
error abap-deploy-task JL1REMOVEMMARCH Deployment has failed.
error abap-deploy-task JL1REMOVEMMARCH Change logging level to debug your issue
error abap-deploy-task JL1REMOVEMMARCH  (see examples https://github.com/SAP/open-ux-tools/tree/main/packages/deploy-tooling#configuration-with-logging-enabled)
error builder:custom deploy-to-abap Cannot read properties of undefined (reading 'bind')
Command deploy failed with error : Cannot read properties of undefined (reading 'bind')
```
Soo the last two lines would be replace with the new Error, as shown.

longieirl: Screenshot;
Screenshot 2026-03-26 at 14 28 13

lfindlaysap:
```suggestion
            `The '${config.target?.destination}' destination is not recognized as an ABAP system. ` +
```

PR #4476, opened 2026-03-26
feat(sap-systems-ext): adds support for the generic host connection type

cianmSAP:
Adds a new **"Generic Host"** connection type to the SAP Systems extension. Users can now create/test connections where the system URL is a base URL (origin) and a separate **service path** is provided, rather than requiring a full service URL.

Also enhances the error when a system already exists, rather than a generic error it now names and adds a hyperlink to the existing system that there is a conflict

Screenshot 2026-03-26 at 14 32 19
Screenshot 2026-03-26 at 14 32 22
Screenshot 2026-03-26 at 15 29 47

### DIFF
--- a/examples/simple-generator/CHANGELOG.md
+++ b/examples/simple-generator/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @sap-ux/generator-simple-fe
 
+## 1.1.231
+
+### Patch Changes
+
+- @sap-ux/fiori-elements-writer@2.8.114
+- @sap-ux/fiori-freestyle-writer@2.5.84
+
 ## 1.1.230
 
 ### Patch Changes
--- a/examples/simple-generator/package.json
+++ b/examples/simple-generator/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/generator-simple-fe",
-    "version": "1.1.230",
+    "version": "1.1.231",
     "description": "Simple example of a yeoman generator for Fiori elements.",
     "license": "Apache-2.0",
     "private": true,
--- a/packages/fiori-app-sub-generator/CHANGELOG.md
+++ b/packages/fiori-app-sub-generator/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @sap-ux/fiori-app-sub-generator
 
+## 0.13.8
+
+### Patch Changes
+
+- @sap-ux/fiori-elements-writer@2.8.114
+- @sap-ux/fiori-freestyle-writer@2.5.84
+- @sap-ux/odata-service-inquirer@2.20.6
+
 ## 0.13.7
 
 ### Patch Changes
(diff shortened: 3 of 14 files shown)


packages/sap-systems-ext-webapp/src/i18n/i18n.json
```diff
@@ -55,5 +57,9 @@
     },
     "warnings": {
         "passwordStoreWarning": "Passwords are stored in your operating system's credential manager and are protected by its security policies."
+    },
+    "tooltips": {
+        "servicePathDescription": "You can enter a service path to test the connection.",
+        "contentNotSaved": "The content in this field will not be saved."
```
lfindlaysap:
```suggestion
        "contentNotSaved": "The content in this field is not saved."
```

packages/sap-systems-ext-webapp/src/components/layout/main/systemInfo/ServicePath.tsx
```diff
@@ -0,0 +1,48 @@
...
+    );
+
+    return (
+        <div className="store-text-field">
+            <UITooltip tooltipProps={UITooltipUtils.renderContent(servicePathTooltipContent)} delay={0}>
+                <label className="store-detail-label">{t('labels.servicePath')}</label>{' '}
+                <UIIcon className="service-path-info-icon" iconName={UiIcons.Info} />
+            </UITooltip>
+            <UITextInput
+                name="sysemServicePath"
```
mikicvi-SAP: Small typo

```suggestion
                name="systemServicePath"
```

cianmSAP:
> **Possible unintended change: `showFioriProjectBtn` now shows for `odata_service`**

Intended, the intent is obvious and a comment is not required in my opinion

> **`ServicePath` is uncontrolled and `servicePath` extends `BackendSystem` (persisted type)**

The `servicePath` property is held in the state of the webapp at the same level as the other system information properties so extending is appropriate.

> **Non-nullable `!` assertions in `SystemStatus.tsx`**

Updated `client` however without `!` on the URL it won't compile (or would need `?? ''` which would have more complexity), and it's an 'existing system', which can't exist without a URL

> **`showOutputChannelLink` on `ConnectionStatus` — extension-side not visible in this PR**

No correct, it is set in line 150 in packages/sap-systems-ext/src/panel/system/actions/testConnection.ts

> **Unrelated changes bundled in this PR**

These changes are not in the PR

> **Minor: split i18n strings are hard to localize**

Splitting is required due to the clickable link needed mid sentence

packages/sap-systems-ext-webapp/src/i18n/i18n.json
```diff
@@ -38,6 +40,10 @@
             "forMoreInfo": "For more information, check the ",
             "outputTab": "OUTPUT tab."
         },
+        "connectionExists": {
+            "prefix": "A connection [",
```
lfindlaysap: Would it be possible to say: "The {name} connection already exists with the same URL and client."

cianmSAP: Done

PR #4478, opened 2026-03-27
chore(deps): major dependency version upgrades

devinea:
## Summary

Upgrades shared dependencies across the ~92-package monorepo to current versions, with hard constraints to prevent ESM-only packages, React 17+ upgrades, and breaking API changes.

## Changes by Phase

- **Phase 1** — Alignment fixes: `jest-environment-jsdom@29.7.0`, `@testing-library/jest-dom@6.9.1`, `prettier@3.x`, `js-yaml@4.x`, `typescript-eslint@8.57.2`
- **Phase 2** — Storybook cleanup: `@storybook/react@8.6.17`, removed deprecated `@storybook/addons`
- **Phase 3** — Renovate config hardening: freeze rules for ESM-only packages (`inquirer`, `memfs`, `mem-fs`, `yeoman-*`, `react-markdown`, `express`)
- **Phase 4** — Minor/patch alignment: jest 30.3.0, babel-jest 30.3.0, i18next 25.x, ws 8.20.0, axios 1.13.6, @ui5/cli 4.0.50
- **Phase 5** — Safe major upgrades: `xml-formatter@3`, `diff@8`, `commander@14`, `jest-when@4`, `jest-extended@7`, `tsconfig-paths@4`
- **Phase 6** — Webpack loader upgrades: `sass-loader@16`, `style-loader@4`, `css-loader@7`
- **Phase 7** — Proxy/HTTP stack: `https-proxy-agent@7.0.6` (capped; v8 is ESM-only), `http-proxy-middleware@3.0.5`
- **Phase 8** — Renovate freeze rules: chalk, open, figures, filenamify, os-name, minimatch, uuid, marked, react-i18next, storybook, ejs, applicationinsights, @sapui5/types, redux/react-redux/@reduxjs/toolkit, https-proxy-agent
- **Phase 9** — Trivial alignment: `@types/yeoman-generator@5.2.14` in 13 packages, `jest-environment-jsdom@^29.7.0` normalization, removed stale `@types/diff` from `fiori-mcp-server`

## Frozen Dependencies (intentionally not upgraded)

| Package | Reason |
|---|---|
| `inquirer@8` | v9+ is ESM-only |
| `mem-fs@2` / `mem-fs-editor@9` | v3+/v10+ have breaking API changes |
| `yeoman-generator@5` / `yeoman-environment@3` | v6+/v4+ are ESM-only |
| `react@16` / `react-dom@16` | Renovate hard-cap; enzyme requires React 16 |
| `react-markdown@5` | v6+ is ESM-only |
| `chalk@4` / `open@8` / `figures@3` | v5+/v9+/v4+ are ESM-only |
| `https-proxy-agent` | Capped at v7.x; v8 is `type: module` |
| `redux` / `react-redux` | Safe only after React 18 migration |
| `TypeScript` | Staying at 5.9.3; TS6 has too many breaking changes for this sprint |

## Test plan

- [x] `pnpm nx run-many -t build` — 92/92 projects pass
- [x] `pnpm nx run-many -t lint` — 95/95 projects pass (0 errors)
- [x] `pnpm nx run-many -t test` — 87/87 projects pass


packages/ui-prompting/package.json
```diff
@@ -48,32 +48,32 @@
         "@babel/preset-typescript": "7.28.5",
         "@sap-ux/i18n": "workspace:*",
         "@sap-ux/inquirer-common": "workspace:*",
-        "@storybook/addons": "7.6.20",
```
815are, line 51: @devinea 

should we also do same for examples? -> https://github.com/SAP/open-ux-tools/blob/main/examples/ui-prompting-examples/package.json#L41

packages/app-config-writer/src/i18n.ts
```diff
@@ -34,7 +34,7 @@ export async function initI18n(): Promise<void> {
  * @returns {string} localized string stored for the given key
  */
 export function t(key: string, options?: TOptions): string {
-    return i18n.t(key, options);
+    return (i18n.t as (key: string, opts?: TOptions) => string)(key, options);
```
heimwege, line 37: Why is this change needed? Shouldn't the `i18n.t` function be typed? Thich change occurs in multiple places not just in `app-config-writer`

packages/odata-vocabularies/jest.config.js
```diff
@@ -1,3 +1,7 @@
 const config = require('../../jest.base');
 config.collectCoverageFrom.push('!src/**/index.ts', 'tools/update.ts', '!tools/run-update.ts'); // ignoring index and update vocabulary file, index has only export statements and update is used to update the vocabularies (utility for updating not a deliverable code)
+// prettier@3 uses dynamic import() internally which requires --experimental-vm-modules in Jest CJS mode.
+// The only test that uses update.ts (debug-update-vocabularies.test.ts) is fully skipped,
```
heimwege: If the test is skipped we should find out if it is still needed -> re-enable or not -> delete

packages/ui-prompting/package.json
devinea: @815are Good catch. Done in https://github.com/SAP/open-ux-tools/pull/4478/commits/56fec39101b6dc35aa1c3e12599fb8a84b7c2a9d

packages/preview-middleware-client/test/jest-environment-jsdom-writablelocation.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * Custom Jest environment that extends jsdom and makes window.location.reload
```
heimwege, line 2: Why is this change needed? Was the old implementation no loger working?

packages/preview-middleware-client/eslint.config.js
```diff
@@ -13,21 +13,23 @@ module.exports = [
         ]
     },
     ...fioriTools.configs['recommended'],
+    // Register @typescript-eslint plugin + typed linting for test/ files (recommended only covers src/)
     {
-        languageOptions: {
-            parser: tsParser,
-            ecmaVersion: 5,
-            sourceType: 'script',
+        files: ['test/**/*.ts'],
```
heimwege, line 18: I don't think we only want to lint test files. Where does the need for a change in this files come from?

packages/app-config-writer/src/i18n.ts
heimwege: _This is a legitimate i18next 25.x breaking type change. Starting in i18next v24+, the t() function signature became much more complex/overloaded for TypeScript (it now returns string | string[] in many overloads rather than just string). The explicit cast to (key: string, opts?: TOptions) => string is the standard workaround to preserve the simple string return type that all callers of the local t() wrapper expect. This is a type-only change with no runtime effect — the behavior is identical to before. The pattern is consistent across all packages and is correct._

packages/preview-middleware-client/test/jest-environment-jsdom-writablelocation.js
heimwege: _This is needed because jest 30 bundles jsdom 26, which implements the Location interface as "unforgeable" (non-configurable, non-writable properties per HTML spec). The old Object.defineProperty(window, 'location', ...) pattern that many tests use to mock window.location.reload now throws. The workaround patches the internal [impl] symbol's reload directly, which is technically sound for the test environment. The approach is well-documented with a comment explaining the rationale._

packages/preview-middleware-client/eslint.config.js
heimwege: _The fioriTools.configs['recommended'] only covers `src/**/*.ts` and doesn't register the `@typescript-eslint` plugin with typed linting for test files. The new block explicitly adds `@typescript-eslint` plugin + parser for `test/**/*.ts`, so typed ESLint rules work in test files too. Without it, typescript-eslint rules would silently not apply to test files after the flat config migration. The rules in the combined `src/**/*.ts`, `test/**/*.ts` block still cover both — only the plugin registration was missing for tests._

packages/odata-vocabularies/jest.config.js
heimwege: Ha! At least that comment is not being shut down by my AI:

_the author should clarify whether that test is still intended to eventually be enabled, or if it should be deleted._

🥳

.changeset/ui5-library-sub-generator-dep-upgrade.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-ux/ui5-library-sub-generator": patch
+---
+
+chore(ui5-library-sub-generator): upgrade shared devDependencies (jest 30, i18next 25, @types/yeoman-generator 5.2.14)
```
heimwege: General: It's not just bumping `devDependencies`

  _@sap-devx/yeoman-ui-types 1.22.0 → 1.23.0 — This is an external (non-workspace) runtime dependency bump across many packages. The changeset descriptions for those packages only mention devDependencies upgrades, but yeoman-ui-types is in dependencies — so it's a runtime dep change. Changesets are present for those packages, so it's covered, but the descriptions are slightly misleading (saying "upgrade shared devDependencies" when yeoman-ui-types is a runtime dep)._

packages/preview-middleware-client/test/jest-environment-jsdom-writablelocation.js
devinea: Newer jsdom makes location readonly. This is needed to still be able to write location in the test

.changeset/ui-prompting-examples-dep-upgrade.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-ux-private/ui-prompting-examples": patch
```
heimwege, line 2: _examples/ changeset — @sap-ux-private/ui-prompting-examples has a changeset, even though it's an examples/ directory.
  That's consistent with the updated AGENTS.md note that private packages still need changesets._

But is it correct for stuff from `/examples`?

.changeset/ui5-library-sub-generator-dep-upgrade.md
devinea: Fixed — 15 changeset files updated to explicitly mention `@sap-devx/yeoman-ui-types 1.22.0 → 1.23.0` as a runtime dependency upgrade, separate from the devDependencies changes. See commit 7c08273.

.changeset/ui-prompting-examples-dep-upgrade.md
devinea: Yes — examples/ packages that have source changes should get changesets. `ui-prompting-examples` is a versioned private package (it has a `version` field and its own CHANGELOG), and the source imports changed (`@storybook/addons` → `@storybook/manager-api`). The changeset drives its CHANGELOG entry. We've updated AGENTS.md to make this policy explicit for future contributors.

packages/odata-vocabularies/jest.config.js
devinea: Intentionally kept as a permanent debug helper — useful for manually triggering a vocabulary update during development without needing to run the full update script. Not a regular test and should stay skipped. Updated the comment in jest.config.js to make this intent explicit.

PR #4480, opened 2026-03-27
Feat(eslint-fiori-plugin): cds annotations support

AlinaGovoruhina:
Before testing update cds-dk: `npm i -g @sap/cds-dk@latest`

Added test CAP Projects with @sap/cds version ^9, because `createCdsCompilerFacadeForRootSync()` relies on a project to have cds-compiler version 6.8.0 which has the synchronous methods added to be used with the eslint-fiori-plugin.

Tested with app-level eslint.config.

Also to see issues in .cds files with ESLint extension add this to your VSCode settings.json file:
```
"eslint.validate": [
    ....
    "cds"
  ],
```


.changeset/funny-badgers-grab.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ux/eslint-plugin-fiori-tools': minor
+'@sap-ux/fiori-annotation-api': minor
```
marufrasully: not sure if annotation api should be minor?

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
```diff
@@ -257,7 +257,7 @@ export class ApplicationParser {
                     type: 'cap',
                     name: dataSourceName,
                     path: uniformUrl(dataSource.uri),
-                    version: dataSource.settings?.odataVersion ?? '2.0'
+                    version: dataSource.settings?.odataVersion ?? '4.0'
```
marufrasully, line 260: why was changed to 4.0?

packages/eslint-plugin-fiori-tools/test/project-context/linker/fe-v2.test.ts
```diff
@@ -245,5 +277,56 @@ describe('FE V2 Linker', () => {
             });
         });
     });
+
+    test('collectSections', async () => {
+        const context = await setup({});
+        const mainService = getParsedServiceByName(context.app);
+        expect(mainService).toBeDefined();
+        if (!mainService) {
```
marufrasully, line 284: is it expected to return or throw exception? Now test will silently pass.

packages/eslint-plugin-fiori-tools/test/project-context/linker/fe-v2.test.ts
```diff
@@ -245,5 +277,56 @@ describe('FE V2 Linker', () => {
...
+            return;
+        }
+        const entity = mainService.index.entitySets['Z_SEPMRA_SO_SALESORDERANALYSIS'];
+        expect(entity).toBeDefined();
+        if (!entity?.structuredType) {
```
marufrasully, line 288: same here

packages/eslint-plugin-fiori-tools/test/rules/sap-width-including-column-header.test.ts
```diff
@@ -269,3 +290,223 @@ ruleTester.run(TEST_NAME, flexEnabledRule, {
...
+            {
+                name: 'non manifest file - json',
+                filename: 'some-other-file.json',
+                code: JSON.stringify(V4_MANIFEST)
+            },
+            []
+        ),
+        createValidTestCAP(
+            {
+                name: 'non manifest file - xml',
```
marufrasully: - xml ? maybe `cds` is meant?

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
AlinaGovoruhina: Isn't a CAP service always OData V4?

packages/eslint-plugin-fiori-tools/test/project-context/linker/fe-v2.test.ts
AlinaGovoruhina: I check if mainService is defined in line 284.

.changeset/funny-badgers-grab.md
AlinaGovoruhina: Should I change to `patch`?

packages/eslint-plugin-fiori-tools/test/rules/sap-width-including-column-header.test.ts
AlinaGovoruhina: Changed to test a .cds file.

packages/eslint-plugin-fiori-tools/test/project-context/linker/fe-v2.test.ts
AlinaGovoruhina: Added a test fail on condition if mainService not found.

packages/eslint-plugin-fiori-tools/test/project-context/linker/fe-v2.test.ts
AlinaGovoruhina: Added a test fail on condition if entity not found.

.changeset/funny-badgers-grab.md
marufrasully: yes.

AlinaGovoruhina: Done.

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
marufrasully: This file is responsible for both v2 and v4. Do we handle only cap here?. If no, then => if odataVersion is not defined in mainfest.json file, by default it is v. with this change we change default to v4.

marufrasully (dismissed):
@AlinaGovoruhina 
- changes looks good
- unit test added
- change set available
- did not verify it

marufrasully (commented):
Please do consider this https://github.com/SAP/open-ux-tools/pull/4480#discussion_r3091409560

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
AlinaGovoruhina: But we assume it's a cap service in this block because there is no `localUri`, further is code that defaults to version '2.0'.

packages/eslint-plugin-fiori-tools/test/test-helper.ts
```diff
@@ -71,13 +109,53 @@ export const V2_ANNOTATIONS_PATH = join(
...
+    } else {
+        console.log(`Package installed successfully in ${projectPath}`);
+    }
+}
+
+export function setup(name: string, capAppPath?: string) {
+    const lookup: Record<string, { changes: FileChange[]; filename: string }> = {};
+    if (capAppPath) {
+        // install relevant cds-dk for cds compilation
+        npmInstall(CAP_PROJECT_PATH);
```
marcovieth, line 156: Shall we use await here?

packages/eslint-plugin-fiori-tools/src/project-context/artifacts.ts
```diff
@@ -10,7 +10,7 @@ import type { WorkerResult } from './types';
  */
 export async function getProjectArtifacts(filePath: string): Promise<WorkerResult> {
     try {
-        const projectRoot = await findProjectRoot(filePath, false);
+        const projectRoot = await findProjectRoot(filePath, true); // sapuxRequired for CAP apps to locate the project root
```
marcovieth, line 13: So all non-fiori-apps without sapux are excluded now (not only CAP).
No need to specify "true" since it is already the default (but it does not harm).

marcovieth (commented):
Just some minor comments

packages/eslint-plugin-fiori-tools/src/project-context/artifacts.ts
AlinaGovoruhina: Yes, because the application level package.json does not have `"sapux" : true` and it's usually the closest package.json file from the `filePath`, so it gets detected first. But for working with a CAP project, a project level package.json is required to get the project information.

packages/eslint-plugin-fiori-tools/test/test-helper.ts
AlinaGovoruhina: Good find. I replaced cds module check with checking the module path instead of getting the @sap/cds module, as we don't use the module anyway. Changes in https://github.com/SAP/open-ux-tools/pull/4480/commits/50cf1c74354f1e81180fd9a2c748f9d7bba4512c commit.

marcovieth (approved):
Looks good to me.
Also tested on the command line that it reports issues in *.cds files.

marufrasully (approved):
re-approving it.

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
marufrasully: @AlinaGovoruhina the default version is v2, if not available it should default to v2 and not v4.

marufrasully (changes requested):
Please adapt v2 vs v4

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
marufrasully: crossed checked, it is only dealing with CAP service and not both. Therefore v4 is default.

marufrasully (approved):
re-approving. Review comment addressed / crossed checked.

igorsVoronins (approved):
Tested with CAP CDS project ✅ 
Eslint correctly fires `eslint-plugin-fiori-tools` rules for cds and json files. 

Image

There was one rule not firing during eslint check, `sap-text-arrangement-hidden`, this will be investigated separately

Issue #4430, opened 2026-03-18
BUG - card generator http 500 error for batch in CAP apps


heimwege:
### Description
When creating object page cards for a CAP app I get http `500` errors for batch requests triggered by the card generator.

preview card shows text "500 internal server error"
request for this response is `http://localhost:8080/odata/v4/CapFeTsSampleService//$batch` response text is "Unknown Entity Set"
content of batch is `GET Samples(ID=09274d4a-b5c9-45ab-a776-60e1b4ad4cfc,IsActiveEntity=true)?$select=ID,IsActiveEntity`

The app reads its data via `http://localhost:8080/odata/v4/CapFeTsSampleService/$batch` (note: there's only one `/` before the `$batch`, not two) and gets a http `200` result

content of batch is `GET Samples(ID=09274d4a-b5c9-45ab-a776-60e1b4ad4cfc,IsActiveEntity=true)?$select=HasActiveEntity,HasDraftEntity,ID,IsActiveEntity,createdAt,createdBy,name&$expand=DraftAdministrativeData($select=DraftIsCreatedByMe,DraftUUID,InProcessByUser)`

🛑 Issue is that the generated `cards/op/Samples/manifest.json` has a incorrect `"sap.card".data.request.url`
```"{{destinations.service}}/odata/v4/CapFeTsSampleService//$batch"```

### Steps to Reproduce
1. clone https://github.com/heimwege/cap-node-fe-ts-sample
2. add card generator endpoint to ui5 yaml
    ```yaml
            editors:
              cardGenerator:
                path: localService/cardGenerator.html
    ```
3. npm i
4. npm run ui:mockdata:proxy
5. open card generator endpoint at `localService/cardGenerator.html`
6. click on first item in list report
7. open card generator
-> preview card shows 500 internal server error

### Expected results
correct batch requests

### Actual results
wrong batch requests

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [X] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #4485, opened 2026-03-28
fix(preview-middleware): sanitize cards generator manifest URL to avoid issues with double slashes when joining paths

heimwege:
#4430

fixes issue with card generator creating manifests with wrong data request url (`...//$batch`)

But actually this should be fixed by the card generator component by collapsing `//` to `/` before writing it to the request url property.


Singham-24 (approved):
Changeset OK
Did not test manually
coverage ok

PR #4489, opened 2026-03-30
fix(ui5-test-writer): correctly create OP header form field identifier

sap-sebelao:
Internal issue 36741

- fix for object page OPA tests generate with invalid/incorrect values
- these tests will cause the tests to fail


docirl (dismissed):
Changes look good
Not tested manually
Changeset added

cianmSAP (dismissed):
Code updates look good
Have not tested 
Changeset looks good

docirl (approved):
re-approving

cianmSAP (approved):
Re-approving

PR #4507, opened 2026-04-01
fix sonar issues

heimwege:
Fix high and medium severity SONAR issues (https://sonarcloud.io/project/issues?impactSeverities=HIGH%2CMEDIUM&issueStatuses=OPEN%2CCONFIRMED&id=SAP_open-ux-tools)

#### UI component issues (need visual/CSS work):
  - ThemeSelectorCallout.tsx:53 - `<div role="button"> → <button>` requires CSS button resets (border, background, padding) that .theme-child doesn't currently have.
  - UITable.tsx:274,289 - `role="gridcell"` divs live inside FluentUI's DetailsList which is a div-based table, so replacing with <td> would produce invalid HTML.
  - UITable.tsx:950 - `<span onClick> → <button>` would break 12 tests that explicitly query .ms-DetailsRow-cell span.
  - UIFlexibleTable.tsx:254 - the root `<div onBlur>` needs the onBlur handler for focus management and has no suitable native element replacement.
  - UIDatePicker.tsx:81 - the wrapper `<div onKeyDown onClick>` forwards parent-provided handlers and restructuring would change the component's public API.
  - UIToggleGroup.tsx:129 - replacing `<div role="group">` with `<fieldset>` requires converting UILabel to a <legend>, which is a non-trivial refactor.
  - FileChange.tsx:46 - `<a href="#"> → <button>` would visually break the link because .textLink only sets color and doesn't reset button defaults.

#### Ignored tests (S1607):
  - fiori-service.test.ts:113 - test.skip is used conditionally for models in SKIP_TARGETS; removing it requires investigating why those models are excluded.
  - fiori-service.test.ts:157 - describe.skip is part of a test runner helper; it's intentional infrastructure for skipping entire test cases.
  - csv.test.ts:28 - test.skip inside a dynamic loop with an empty skip[] array; the code path is dead but it's a deliberate escape hatch for the test harness.
  - properties.test.ts:29 - same pattern as csv.test.ts:28.
  - parser.test.ts:47 - same pattern as csv.test.ts:28.
  - debug-update-vocabularies.test.ts:8 - it.skip with an explicit comment "debugging only"; it intentionally never runs in CI.


nikmace (dismissed):
Changes look good from middleware's side
Coverage is good
Changeset is OK
Did not test manually

packages/cds-odata-annotation-converter/src/transforms/annotation/flattened.ts
```diff
@@ -145,32 +192,8 @@ function convertToExpandedStructure(
...
-            const vocabularyNameOrAlias = segment.value.substring(1);
-            const termNameSegment = segments[i + 1];
-
-            const termQualifiedName = termNameSegment
-                ? `${vocabularyNameOrAlias}.${termNameSegment.value}`
-                : vocabularyNameOrAlias;
-            const termValueRange = termNameSegment
-                ? createRange(segment.range?.start, termNameSegment.range?.end)
-                : structuredClone(segment.range);
-
```
devinea, line 157: **Potential behavioral change (bug fix or regression?)**

The old code always incremented `i` by 2, regardless of whether a `termNameSegment` actually existed after the `@`-segment. This new code correctly increments by `1 + consumed` (0 or 1).

For the case where there is no following term name segment, the old code would have skipped an extra (unrelated) segment — which was likely a bug. However, if any callers relied on that skipping behavior inadvertently, this change would alter their output. Worth confirming with an explicit test covering a standalone `@VocabAlias` segment (no `.TermName` following).

packages/text-document-utils/src/position.ts
```diff
@@ -103,17 +103,22 @@ export function getIndentLevel(startPosition: number, tabWidth: number): number
...
+
+/**
+ * Returns an indentation string using spaces.
  *
  * @param tabWidth
- * @param useTabs
  * @param level
- * @returns intentation string
+ * @returns indentation string
  */
```
devinea, line 121: **Breaking public API change — should not be a `patch` changeset**

`indent(tabWidth, useTabs, level)` is a named export of both `@sap-ux/text-document-utils` and (re-exported via) `@sap-ux/odata-annotation-core`. Removing and replacing it with `indentWithTabs` / `indentWithSpaces` is a **breaking change** for any external consumer calling `indent` directly.

The changeset marks this as `patch` for both packages, but semver requires at least `minor` (if treated as an additive rename with the old export removed) or `major` (if there are downstream consumers outside this monorepo).

Consider either:
1. Keeping `indent` as a deprecated wrapper and exporting the two new functions alongside it, or
2. Bumping the changeset to `major` for `@sap-ux/text-document-utils` and `@sap-ux/odata-annotation-core`.

packages/control-property-editor/src/components/ChangeIndicator.tsx
```diff
@@ -22,14 +22,7 @@ export function ChangeIndicator(props: ChangeIndicatorProps): ReactElement {
 
     if (saved > 0 && pending === 0) {
         return (
-            <svg
```
devinea, line 25: **Accessibility behavioral change**

Removing `role="img"` from these SVG elements means they are no longer exposed to screen readers as images. If the SVGs are purely decorative, this is correct (a presentational SVG needs `aria-hidden="true"` instead to be fully invisible to assistive technology). If they convey meaning (e.g. saved vs. pending state), removing the role without adding `aria-label` / `aria-hidden` is a regression.

Suggested fix for decorative SVGs: add `aria-hidden="true"` instead of simply removing `role="img"`.

815are (dismissed):
checked `ui-components`:
1. sonar fixes with complexity - I did not find issue after suggested changes
2. changelog persists

I did not test manually

packages/control-property-editor/src/components/ChangeIndicator.tsx
heimwege: @devinea I absolutely don't know where these are being used 🤷🏻 @Jimmy-Joseph19 do you know if this is a decorative only svg or was this actually a false positive sonar issue and we need the role img?

packages/text-document-utils/src/position.ts
heimwege: @devinea what would you prefer? I don't want to do a major version change. So re-add `indent`?

packages/cds-odata-annotation-converter/src/transforms/annotation/flattened.ts
heimwege: @Jimmy-Joseph19 can you comment on this? Is this a valid fix we should go for or should we revert to the old behavior?

packages/text-document-utils/src/position.ts
heimwege: fixed in 3f463c59e9e3575702d4c926d616c9877da1b9a7

packages/control-property-editor/src/components/ChangeIndicator.tsx
Jimmy-Joseph19: image

@heimwege its not decorative it tells information whether there are flex changes (Saved/Non Saved) on outline node.

heimwege:
@longieirl I still get the heap issue: https://cloud.nx.app/runs/rrrtZ9ypBS. #4566 might not have solved the issue 🤷🏻

packages/cds-odata-annotation-converter/src/transforms/annotation/flattened.ts
Jimmy-Joseph19: The new changes looks good. test is already covered here
https://github.com/SAP/open-ux-tools/blob/838d1e90ca003f664d5a8f08ddf9812c196389fd/packages/cds-odata-annotation-converter/test/transform/term.test.ts#L246-L250

devinea (approved):
Re-approving.

PR #4542, opened 2026-04-10
Feat: add sap-description-column-label ESLint rule

marcovieth:
Ensures that the property referenced as the description (text) value via `Common.Text` has a meaningful `Common.Label` -- not a generic value such as `"Name"` or `"Description"`, and not the same label as the ID property it describes.


packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
+# Require Meaningful Labels for Description (Text) Properties (`sap-description-column-label`)
+
+Ensures that the property referenced as the description (text) value via the `Common.Text` annotation has a meaningful `Common.Label` annotation — not a generic value such as `"Name"` or `"Description"`, and not the same label as the ID property it describes.
```
lfindlaysap:
```suggestion
Ensures that the property referenced as the description (text) value using the `Common.Text` annotation has a meaningful `Common.Label` annotation. The label must not be a generic value such as `"Name"` or `"Description"` nor the same label as the ID property it describes.
```

.changeset/rich-meals-shake.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sap-ux/xml-odata-annotation-converter': minor
+'@sap-ux/odata-annotation-core-types': minor
+'@sap-ux/eslint-plugin-fiori-tools': minor
+---
+
+[rule] Add rule to check that a Common.Text description property has a meaningful Common.Label
```
lfindlaysap:
```suggestion
[rule] Add rule to check that a Common.Text description property has a meaningful Common.Label annotation
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Rule Details
+
+When a field uses the `Common.Text` annotation to associate a human-readable description property with a technical key property, the description property should carry a `Common.Label` annotation that clearly identifies its content. A generic label like `"Name"` or `"Description"` provides no useful context to the user in the UI (for example, as a column header in a table), and a label that is identical to the ID property's label creates ambiguity between the two columns.
```
lfindlaysap:
```suggestion
When a field uses the `Common.Text` annotation to associate a human-readable description property with a technical key property, the description property must carry a `Common.Label` annotation that clearly identifies its content. A generic label such as `"Name"` or `"Description"` provides no useful context to the user in the UI and a label that is identical to the ID property's label creates ambiguity.
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+
+The rule checks every `Common.Text` annotation on a property that belongs to an entity type used in the application (i.e. linked to at least one page via the manifest). For each such annotation the rule:
```
lfindlaysap:
```suggestion
The rule checks every `Common.Text` annotation on a property that belongs to an entity type used in the application, that is linked to at least one page in the `manifest.json` file. For each annotation, the rule performs the following:
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+
+1. Resolves the path to the referenced text property (including navigation segments such as `category/name`).
```
lfindlaysap:
```suggestion
1. Resolves the path to the referenced text property which includes navigation segments such as `category/name`.
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+2. Reads the `Common.Label` annotation on that property.
+3. Flags the annotation if the label is `"Name"` or `"Description"` (case-insensitive) — **trivialLabel**.
```
lfindlaysap:
```suggestion
3. Produces a `trivialLabel` warning if the label is the same as the `"Name"` or `"Description"` (case-insensitive).
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+4. Flags the annotation if the label is identical (case-insensitive) to the `Common.Label` annotation of the ID property that carries the `Common.Text` annotation — **duplicateLabel**.
```
lfindlaysap:
```suggestion
4. Produces a `duplicateLabel` warning if the label is identical (case-insensitive) to the `Common.Label` annotation of the ID property that carries the `Common.Text` annotation.
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+
+If the text property has no `Common.Label` annotation, the rule does not flag it.
```
lfindlaysap:
```suggestion
If the text property has no `Common.Label` annotation, the rule does not produce a warning.
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+
+> **OData V2**: The rule also applies to OData V2 services. In V2 metadata, the `sap:text` attribute on a `Property` element is treated as an implicit `Common.Text` annotation, and the `sap:label` attribute is treated as an implicit `Common.Label` annotation. The rule reports on the metadata file directly when these inline attributes produce a trivial or duplicate label.
```
lfindlaysap:
```suggestion
> **OData V2**: This rule also applies to OData V2 services. In OData V2 metadata, the `sap:text` attribute on a `Property` element is treated as an implicit `Common.Text` annotation and the `sap:label` attribute is treated as an implicit `Common.Label` annotation. The rule reports on the metadata file directly when these inline attributes produce a trivial or duplicate label.
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Why Was This Rule Introduced?
+
+When a property is displayed using a description column (e.g. `TextArrangement/TextOnly`), the column header comes from the `Common.Label` annotation of the text property. A label of `"Name"` or `"Description"` is semantically meaningless in that context — it does not tell the user _which_ name or description is shown. Similarly, if the description column carries the exact same label as the ID column, the user cannot distinguish the two columns.
```
lfindlaysap:
```suggestion
When a property is displayed using a description column, for example, `TextArrangement/TextOnly`, the column header comes from the `Common.Label` annotation of the text property. A label of `"Name"` or `"Description"` is semantically meaningless in that context, it does not tell the user which name or description is shown. Similarly, if the description column carries the exact same label as the ID column, the user cannot distinguish the two columns.
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Warning Messages
+
+**trivialLabel**
```
lfindlaysap:
```suggestion
`trivialLabel`
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+```
+The text property "{{textPropertyTarget}}" has a generic label "{{textPropertyLabel}}". Use a more descriptive label that distinguishes it from other properties.
```
lfindlaysap:
```suggestion
The "{{textPropertyTarget}}" text property has a "{{textPropertyLabel}}" generic label. Use a more descriptive label that distinguishes it from other properties.
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+**duplicateLabel**
```
lfindlaysap:
```suggestion
`duplicateLabel`
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+```
+The text property "{{textPropertyTarget}}" has the same label "{{textPropertyLabel}}" as the ID property "{{idPropertyTarget}}". The description column label should be different from the ID label.
```
lfindlaysap:
```suggestion
The "{{textPropertyTarget}}" text property has the same "{{textPropertyLabel}}" label as the "{{idPropertyTarget}}" ID property. The description column label must be different from the ID label.
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+## Examples
+
+### Incorrect Annotations
+
+**`trivialLabel` — the `Common.Label` annotation on the text property is too generic:**
```
lfindlaysap:
```suggestion
`trivialLabel`: The `Common.Label` annotation on the text property is too generic:
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+
+```xml
+<!-- ProductBaseUnit uses UnitOfMeasure_Text as its description column via Common.Text annotation -->
+<Annotations Target="MyService.MyEntity/ProductBaseUnit">
+    <Annotation Term="Common.Text" Path="to_BaseUnit/UnitOfMeasure_Text"/>
+</Annotations>
+
+<!-- The Common.Label annotation on the description property is "Name" — too generic -->
```
lfindlaysap:
```suggestion
<!-- The Common.Label annotation on the description property is "Name" which is too generic -->
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+<Annotations Target="MyService.BaseUnitType/UnitOfMeasure_Text">
+    <Annotation Term="Common.Label" String="Name"/>
+</Annotations>
+```
+
+**`duplicateLabel` — the `Common.Label` annotation on the text property is identical to the ID property label:**
```
lfindlaysap:
```suggestion
`duplicateLabel`: The `Common.Label` annotation on the text property is identical to the ID property label:**
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+
+```xml
+<!-- Supplier uses CompanyName as its description column via Common.Text annotation, both properties have label "Supplier" -->
```
lfindlaysap:
```suggestion
<!-- Supplier uses CompanyName as its description column using the Common.Text annotation but both properties have the "Supplier" label -->
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+<Annotations Target="MyService.MyEntity/Supplier">
+    <Annotation Term="Common.Text" Path="to_Supplier/CompanyName"/>
+    <Annotation Term="Common.Label" String="Supplier"/>
+</Annotations>
+
+<!-- Common.Label annotation identical to the ID property above — ambiguous column headers -->
```
lfindlaysap:
```suggestion
<!-- Common.Label annotation identical to the ID property above which leads to ambiguous column headers -->
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+<Annotations Target="MyService.SupplierType/CompanyName">
+    <Annotation Term="Common.Label" String="Supplier"/>
+</Annotations>
+```
+
+**OData V2 `trivialLabel` — inline `sap:label` on the text property is too generic (reported on the metadata file):**
```
lfindlaysap:
```suggestion
**trivialLabel` for OData V2: The inline `sap:label` on the text property is too generic. This is reported on the metadata file:
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+
+```xml
+<!-- sap:text attribute acts as an implicit Common.Text annotation -->
+<Property Name="Product" Type="Edm.String" sap:text="ProductName" sap:label="Product ID" />
+<!-- sap:label="Name" acts as an implicit Common.Label annotation — too generic -->
+<Property Name="ProductName" Type="Edm.String" sap:label="Name" />
+```
+
+**OData V2 `duplicateLabel` — inline `sap:label` values on ID and text property are identical:**
```
lfindlaysap:
```suggestion
 `duplicateLabel` for OData V2: The inline `sap:label` values on ID and text property are identical:
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+
+```xml
+<!-- Both properties carry sap:label="Quantity Unit" via inline sap:label attribute -->
```
lfindlaysap:
```suggestion
<!-- Both properties carry sap:label="Quantity Unit" using the inline sap:label attribute -->
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+<Property Name="QuantityUnit" Type="Edm.String" sap:text="QuantityUnitT" sap:label="Quantity Unit" />
+<Property Name="QuantityUnitT" Type="Edm.String" sap:label="Quantity Unit" />
+```
+
+### Correct Annotations
+
+**Meaningful, distinct `Common.Label` annotation on the text property:**
```
lfindlaysap:
```suggestion
The `Common.Label` annotation on the text property is meaningful and unique:
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+    <Annotation Term="Common.Text" Path="to_BaseUnit/UnitOfMeasure_Text"/>
+</Annotations>
+
+<!-- Common.Label annotation with a descriptive value that distinguishes the column from the ID column -->
+<Annotations Target="MyService.BaseUnitType/UnitOfMeasure_Text">
+    <Annotation Term="Common.Label" String="Unit of Measure"/>
+</Annotations>
+```
+
+**Different `Common.Label` annotations on ID and text properties:**
```
lfindlaysap:
```suggestion
The `Common.Label` annotations on ID and text properties are unique:
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+    <Annotation Term="Common.Text" Path="to_Supplier/CompanyName"/>
+    <Annotation Term="Common.Label" String="Supplier ID"/>
+</Annotations>
+
+<Annotations Target="MyService.SupplierType/CompanyName">
+    <Annotation Term="Common.Label" String="Company Name"/>
+</Annotations>
+```
+
+**OData V2 — distinct `sap:label` values on ID and text property:**
```
lfindlaysap:
```suggestion
The `sap:label` values on ID and text property for OData V2 are unique:
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+
+```xml
+<Property Name="QuantityUnit" Type="Edm.String" sap:text="QuantityUnitT" sap:label="Quantity Unit" />
+<!-- Distinct sap:label — unambiguous column headers -->
```
lfindlaysap:
```suggestion
<!-- The sap:label is unique which provides unambiguous column headers -->
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+<Property Name="QuantityUnitT" Type="Edm.String" sap:label="Unit of Measure Text" />
+```
+
+## Bug Report
+
+If you encounter an issue with this rule, please open a [GitHub issue](https://github.com/SAP/open-ux-tools/issues).
+
+## When Not to Disable This Rule
+
+This rule should not be disabled unless you have a confirmed backend constraint that prevents renaming the label. In most cases the fix is straightforward: update the `Common.Label` annotation (or the `sap:label` attribute in OData V2 metadata) on the text property to something meaningful and distinct from the ID property label.
```
lfindlaysap:
```suggestion
This rule must not be disabled unless you have a confirmed back-end constraint that prevents renaming the label. Update the `Common.Label` annotation or the `sap:label` attribute for OData V2 on the text property to something meaningful and distinct from the ID property label.
```

packages/eslint-plugin-fiori-tools/src/rules/sap-description-column-label.ts
```diff
@@ -0,0 +1,340 @@
...
+}
+
+const rule: FioriRuleDefinition = createFioriRule({
+    ruleId: DESCRIPTION_COLUMN_LABEL,
+    meta: {
+        type: 'problem',
+        docs: {
+            recommended: true,
+            description:
+                'The description (text) property referenced via Common.Text must have a meaningful label — not a generic value such as "Name" or "Description", and not the same label as the ID property it describes.',
```
lfindlaysap:
```suggestion
                'The description (text) property referenced using the Common.Text annotation must have a meaningful label which is not a generic value such as "Name" or "Description" nor the same label as the ID property it describes.',
```

packages/eslint-plugin-fiori-tools/src/rules/sap-description-column-label.ts
```diff
@@ -0,0 +1,340 @@
...
+            url: 'https://github.com/SAP/open-ux-tools/blob/main/packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md'
+        },
+        messages: {
+            trivialLabel:
+                'The text property "{{textPropertyTarget}}" has a generic label "{{textPropertyLabel}}". Use a more descriptive label that distinguishes it from other properties.',
```
lfindlaysap:
```suggestion
                'The "{{textPropertyTarget}}" text property has a "{{textPropertyLabel}}" generic label. Use a more descriptive label that distinguishes it from other properties.',
```

packages/eslint-plugin-fiori-tools/src/rules/sap-description-column-label.ts
```diff
@@ -0,0 +1,340 @@
...
+            duplicateLabel:
+                'The text property "{{textPropertyTarget}}" has the same label "{{textPropertyLabel}}" as the ID property "{{idPropertyTarget}}". The description column label should be different from the ID label.'
```
lfindlaysap:
```suggestion
                'The "{{textPropertyTarget}}" text property has the same "{{textPropertyLabel}}" label as the "{{idPropertyTarget}}" ID property. The description column label must be different from the ID label.'
```

lfindlaysap (changes requested):

lfindlaysap (changes requested):
Here are my suggestions.

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

.changeset/rich-meals-shake.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done (with trailing ** removed)

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done

packages/eslint-plugin-fiori-tools/src/rules/sap-description-column-label.ts
marcovieth: Done

packages/eslint-plugin-fiori-tools/src/rules/sap-description-column-label.ts
marcovieth: Done

packages/eslint-plugin-fiori-tools/src/rules/sap-description-column-label.ts
marcovieth: Done

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+## Examples
+
+### Incorrect Annotations
+
+`trivialLabel`: The `Common.Label` annotation on the text property is too generic:
+
+```xml
+<!-- ProductBaseUnit uses UnitOfMeasure_Text as its description column via Common.Text annotation -->
```
lfindlaysap:
```suggestion
<!-- ProductBaseUnit uses UnitOfMeasure_Text as its description column using the Common.Text annotation -->
```

lfindlaysap (changes requested):

packages/eslint-plugin-fiori-tools/docs/rules/sap-description-column-label.md
marcovieth: Done with commit 49af900

igorsVoronins (dismissed):
Tested locally on OData V4 and OData V2. Both the trivial label and duplicate label warnings fire correctly on the expected lines. Clean annotations produce no warnings.
The rule is working as expected ✅

AlinaGovoruhina (dismissed):
New eslint rule code is commented and clear, covered by unit tests. There's documentation on the rule with XML code snippet examples, changeset minor change for `eslint-plugin-fiori-tools` is added, but could be that other packages need a patch level change. Approving from my side.

lfindlaysap (dismissed):

.changeset/rich-meals-shake.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sap-ux/xml-odata-annotation-converter': minor
+'@sap-ux/odata-annotation-core-types': patch
+'@sap-ux/eslint-plugin-fiori-tools': patch
+---
+
+[rule] Add rule to check that a Common.Text description property has a meaningful Common.Label annotation
```
marufrasully, line 6: this change log is good and meaningful for `@sap-ux/eslint-plugin-fiori-tools` package but maybe not good for `@sap-ux/xml-odata-annotation-converter`. Should we have multipe change set with relevant text. e.g. one more for `@sap-ux/xml-odata-annotation-converter`?

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
```diff
@@ -131,9 +134,112 @@ function processTargetAnnotations(
...
+ */
+function createSyntheticElement(attrName: string, attrValue: string, sourceRange?: Range): Element {
+    const element = createElementNode({
+        name: Edm.Annotation,
+        attributes: {
+            [attrName]: createAttributeNode(attrName, attrValue)
+        },
+        content: []
+    });
+    if (sourceRange) {
```
marufrasully, line 161: what happens if `sourceRange` is equal to `0`?. Please adapt.

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
```diff
@@ -131,9 +134,112 @@ function processTargetAnnotations(
...
+                const syntheticTarget: Target = {
+                    type: 'target',
+                    name: propertyTarget,
+                    terms: [syntheticElement]
+                };
+                annotationFile.targets.push(syntheticTarget);
+            }
+        }
+    });
+}
```
marufrasully: This logic/processing is domain of annotation. Can we move it to annotation module? @SAP/ux-tools-annotations  fyi

packages/eslint-plugin-fiori-tools/src/rules/utils/common-text-helpers.ts
```diff
@@ -0,0 +1,152 @@
...
+ * For example, given entity type "IncidentService.Incidents" and path "category/name",
+ * it navigates through the "category" navigation property to find "IncidentService.Category",
+ * and returns { entityTypeName: "IncidentService.Category", propertyName: "name" }.
+ *
+ * @param entityTypeName - Fully-qualified name of the starting entity type
+ * @param textPath - Path expression (e.g. "category/name" or "name")
+ * @param service - The parsed OData service
+ * @returns Resolved entity type name and property name, or undefined if resolution fails
+ */
+export function resolveTextPropertyPath(
```
marufrasully: did you check annotation package, do we not have such logic already in place?

marufrasully (commented):
@marcovieth please cross check with @SAP/ux-tools-annotations if these logic need to be handled in that package 

- https://github.com/SAP/open-ux-tools/pull/4542/changes#diff-3e1ddc020dfc8687dfe4b65f2f7a4dee48c0751c11491566416f3b4f26c5bcadR175
- packages/eslint-plugin-fiori-tools/src/rules/utils/common-text-helpers.ts
- packages/eslint-plugin-fiori-tools/src/rules/sap-description-column-label.ts

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
marcovieth: How can sourceRange be equal to 0 when it is typed as `Range | undefined`?

.changeset/rich-meals-shake.md
marcovieth: Good catch. I wanted to set only @sap-ux/eslint-plugin-fiori-tools to minor (where the main logic currently is) and the others to patch.
But if we move V2 extraction logic to @sap-ux/xml-odata-annotation-converter...
Good to know that we can also have multiple changeset files with release notes.

packages/eslint-plugin-fiori-tools/src/rules/utils/common-text-helpers.ts
vadson71: There is a function getMdElementForMdPath in the generic annotation service, but it needs to be exported.

marcovieth: So the function is similar to an internal function getMdElementWithCardinalityForMdPath() in annotation-service/src/utils/general.ts .
But since we are in open-ux-tools, we cannot use it.

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
marcovieth: vadson71 provided a patch for the annotation package that collects V2 metadata properties exposed by an extended API. Integrated in commit 7d5d70c .

.changeset/rich-meals-shake.md
marcovieth: I have added another changeset file for xml-odata-annotation-converter and fiori-annotation-api.

vadson71 (dismissed):
Code changes look good
Comments addressed
Test coverage is good
Performed only visual review, didn't run locally

AlinaGovoruhina (dismissed):
Approving changes since the previous review: added changesets look good, checked code and unit tests updates, did not test the rule.

lfindlaysap (dismissed):

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
```diff
@@ -131,17 +134,124 @@ function processTargetAnnotations(
...
+                        layers: [{ uri: metadataUri, value: syntheticElement }]
+                    }
+                };
+                const syntheticTarget: Target = {
+                    type: 'target',
+                    name: propertyTarget,
+                    terms: [syntheticElement]
+                };
+                annotationFile.targets.push(syntheticTarget);
+            }
```
marufrasully, line 197: why we have to modify (add side effect) - can we avoid it?

packages/eslint-plugin-fiori-tools/src/rules/utils/common-text-helpers.ts
marufrasully: The should be part of annotation API module and not eslint.

packages/eslint-plugin-fiori-tools/src/rules/sap-description-column-label.ts
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * Labels that are considered trivially non-descriptive regardless of context.
+ */
+const TRIVIAL_LABELS = new Set(['name', 'description']);
+
+/**
+ * Reads a Common.Label String value and the corresponding annotation from the annotation index.
+ *
+ * @param propertyTarget - Fully-qualified target path (e.g. "Service.Entity/prop")
+ * @param service - The parsed OData service
```
marufrasully, line 25: this file - for me  should be part of annotation api.

marufrasully (changes requested):

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
marcovieth: With this injection approach, there is no special handling needed in the rule itself, V2 support is transparent.
The rule can access V2 metadata annotations as if they where defined in V4 annotation format.
Certainly that could be done differently, but then every (new) rule needing V2 annotations would need specific V2 code.

packages/eslint-plugin-fiori-tools/src/rules/sap-description-column-label.ts
marcovieth: Which file do you mean? sap-description-column-label.ts is the file describing the rule.
And common-text-helpers.ts could be part of annotation api, but it's currently only needed for ESLint.

packages/eslint-plugin-fiori-tools/src/rules/utils/common-text-helpers.ts
marcovieth: I would prefer to refactor it later, if someone else not from the eslint plugin would need it.

igorsVoronins (dismissed):
Manual testing passed ✅
Tested locally the branch (plugin v10.0.5) against Fiori projects(V2 and V4).

  V4 project (`recommended-for-s4hana` config):
  - trivialLabel — fires correctly for Common.Label String="Name" and String="Description"
  - duplicateLabel — fires correctly when text property label matches ID property label
  - No false positives: properties without Common.Label are silent; distinct labels produce no warning

  V2 project (`recommended-for-s4hana` config, annotation.xml with Common.Text/Common.Label):
  - trivialLabel — fires correctly
  - duplicateLabel — fires correctly

  Negative cases:
  - Rule is not active under `recommended` config (only `recommended-for-s4hana`) — confirmed

Image

igorsVoronins:
Additionally, tested sap-text-arrangement-hidden rule on V4 project as it was refactored in scope of this PR. Looks good ✅ 

sap-text-hidden-arrangement

packages/eslint-plugin-fiori-tools/src/rules/utils/common-text-helpers.ts
heimwege: @marcovieth how would _someone else not from the eslint plugin_ even know about the existence of this potentially re-usable function? I'm afraid this will be implemented twice if needed later because nobody will find it if we burry it in here.

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
heimwege: @marufrasully is that reply fine for you?

@marcovieth AI tells me: The mutation side-effect in `injectV2Annotations` appends synthetic Target objects directly to the live `annotationFile.targets` array. If buildServiceIndex is called more than once for the same artifacts (e.g., on a re-parse triggered by a file change — which ApplicationParser does do in `onCapFileChanged` and `onAnnotationFileChanged`), these synthetic targets will be pushed again, accumulating duplicates. There is no guard against this. This is a real correctness risk, not just an architectural preference.

packages/eslint-plugin-fiori-tools/src/rules/sap-description-column-label.ts
heimwege: @marufrasully Is that the same as or related to https://github.com/SAP/open-ux-tools/pull/4542#discussion_r3152915608?

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
marufrasully: @heimwege I think we should have a closer look at this.

packages/eslint-plugin-fiori-tools/src/rules/sap-description-column-label.ts
marufrasully: @heimwege it is about responsibility, logic related module and maintenance of code. If this code is only meant for eslint, we can keep it if it is really effort to move it to related module.

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
marcovieth: @heimwege AI tells me this:

The AI reviewer's concern doesn't hold up. Here's why:
- For XML/local services (the only case that produces non-empty v2Annotations): every re-parse calls getXmlServiceArtifacts → new XMLAnnotationServiceAdapter → adapter.sync() which begins with this.documents.clear(). The ServiceArtifacts object — including all AnnotationFile instances — is freshly constructed on every call. buildServiceIndex never sees the same artifacts instance twice.
- For CDS/CAP services: the ServiceArtifacts IS cached and could be reused, but v2Annotations is always [] for CDS, so injectV2Annotations returns immediately at the if (!v2Annotations.length) guard — the targets.push lines are never reached.
- So the correctness risk the reviewer describes — duplicate synthetic targets accumulating across re-parses — cannot occur with the current architecture. The fresh-artifacts-per-parse design already provides the guard implicitly.
- That said, this is an implicit safety invariant with no comment explaining it. It's worth noting in the review that the code is safe because ServiceArtifacts is never reused for XML services, so a future refactor that introduced caching there would silently introduce the bug the reviewer feared.
---
But yes, we should have a closer look, if the current implementation is so questionable.

packages/eslint-plugin-fiori-tools/src/rules/utils/common-text-helpers.ts
marcovieth: Maybe by searching?
Ok, with commit 3367aa3 I have moved resolveTextPropertyPath to the fiori-annotation-api package.

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
marcovieth: With commit c8d1396 I removed the injection of synthetic elements into the AST and update the rule to read them in an 'annotation-file:exit' selector.

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
```diff
@@ -131,17 +134,108 @@ function processTargetAnnotations(
...
+ */
+function indexV2Annotations(
+    artifacts: ServiceArtifacts,
+    metadataUri: string,
+    v2Annotations: V2Annotation[],
+    index: AnnotationIndex
+): void {
+    if (!v2Annotations.length) {
+        return;
+    }
```
marufrasully: This seems un-necessary. :)

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
```diff
@@ -131,17 +134,108 @@ function processTargetAnnotations(
...
+ * inline `sap:text` and `sap:label` attributes, using pre-parsed V2Annotation objects.
+ *
+ * Only adds entries when no explicit vocabulary annotation already exists for that key.
+ *
+ * @param artifacts - Service artifacts (used to access annotation file AST)
+ * @param metadataUri - URI of the metadata file (used as the annotation source)
+ * @param v2Annotations - Pre-parsed V2 annotations from convertMetadataDocumentV2
+ * @param index - The annotation index to inject into
+ */
+function indexV2Annotations(
```
marufrasully: We already have a function `processTargetAnnotations` this function does similar things, so lest call it `processTargetV2Annotations` and it should be called inside this method `indexAnnotationsByAnnotationPath`

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
```diff
@@ -150,14 +244,21 @@ export function buildServiceIndex(
...
     artifacts.metadataService.visitMetadataElements((element) => {
         // NOSONAR - TODO: check if we can handle CDS differences better
         if (element.kind === 'EntityContainer' || element.kind === 'service') {
             entityContainer = element;
         } else if (element.kind === 'EntitySet' || element.kind === 'entitySet') {
             entitySets[element.name] = element;
         }
+        if (!metadataUri && element.location?.uri) {
+            metadataUri = element.location.uri;
+        }
```
marufrasully: Remove this logic, get file location from v2Annotations provided by adapter. e..g `target.uri`

packages/eslint-plugin-fiori-tools/src/rules/sap-description-column-label.ts
```diff
@@ -0,0 +1,254 @@
...
+                // the element's pre-set range, filtering to the file currently being linted.
+                const currentUri = context.sourceCode.ast.uri;
+                for (const [element, diagnostic] of lookup) {
+                    if (diagnostic.annotation.reference.uri === currentUri) {
+                        report(element, diagnostic);
+                    }
+                }
+                lookup.clear();
+            }
+        };
```
marufrasully, line 250: This look like a workaround. :) We should avoid it. Please Process this in `indexAnnotationsByAnnotationPath`

marufrasully (commented):
I had a closer look. Please see comments.

packages/eslint-plugin-fiori-tools/src/rules/sap-description-column-label.ts
marufrasully: @marcovieth this workaround is acceptable for me. This helps avoid injecting nodes to generic annotation file (e.g. you referred it as AST) and this is only executed once per file.

Note: This whole is because of lack of feature support of Annotation API for V2 Annotations.

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
marcovieth: Done with commit 3636468.

packages/eslint-plugin-fiori-tools/src/project-context/parser/service.ts
marcovieth: Thanks. That allowed to drop the parameter metadataUri. Done with commit 3636468.

marufrasully (approved):
@marcovieth thanks
- code looks good
- review comment addressed
- did not test it locally

igorsVoronins (approved):
Re-tested on both V2 and V4 apps, all scenarios passed ✅

lfindlaysap (approved):

PR #4546, opened 2026-04-13
fix(security): upgrade axios@1.15.0

johannes-kolbe:
Upgrades all axios dependencies

### DIFF
--- a/packages/axios-extension/test/factory.test.ts
+++ b/packages/axios-extension/test/factory.test.ts
@@ -46,7 +46,7 @@ afterAll(() => {
 });
 test('create', async () => {
     mockIsAppStudio.mockReturnValue(true);
-    const getProxyForUrlSpy = jest.spyOn(ProxyFromEnv, 'getProxyForUrl').mockReturnValue(undefined);
+    const getProxyForUrlSpy = jest.spyOn(ProxyFromEnv, 'getProxyForUrl').mockReturnValue('');
     const response = await axios.get(`${server}${servicePath}${metadataPath}`, {
         params: { 'sap-client': client }
     });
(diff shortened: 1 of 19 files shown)


longieirl (approved):
Changesets are good.
Change to test is good, the ProxyFromEnv only cares about the env variable that is set, which wont have the service path.

packages/axios-extension/test/factory.test.ts
```diff
@@ -69,7 +69,7 @@ test('create', async () => {
     const metadata = await service.metadata();
     expect(metadata).toBeDefined();
     expect(metadata).toBe(expectedMetadata);
-    expect(getProxyForUrlSpy).toHaveBeenNthCalledWith(1, `${server}${servicePath}${metadataPath}?sap-client=${client}`);
+    expect(getProxyForUrlSpy).toHaveBeenNthCalledWith(1, `${server}`);
```
donal-tobin-sap, line 72: This does not look correct?

donal-tobin-sap (changes requested):
Please confirm the truncation of the check URL

packages/axios-extension/test/factory.test.ts
devinea: @johannes-kolbe I had also included axios upgrade in https://github.com/SAP/open-ux-tools/pull/4533 
and change the proxy logic to be per request instead of instance. OK to go with that?

longieirl: proxy-from-env (the library used here) does support path-based matching but it was never correctly tested or validated! This is applying a proxy for all requests, there is an upated PR https://github.com/SAP/open-ux-tools/pull/4533 that is providing a more fine-grained approach.

johannes-kolbe: Great, yes thanks @devinea! Didn't see that. Will close this one

johannes-kolbe:
Closing in favor of #4533

PR #4566, opened 2026-04-16
fix(eslint-plugin-fiori-tools): resolve heap OOM on macos Node 20 arm64 (#4545)

longieirl:
# Fix Heap OOM Crash on macOS Node 20 arm64 for `eslint-plugin-fiori-tools`

### Bug Fix

🐛 Resolves a fatal heap out-of-memory crash occurring on `macos-latest` with Node 20.x (arm64) after the `eslint-plugin-fiori-tools` test suite completes. All tests passed successfully, but the process crashed during teardown due to excessive V8 coverage data being loaded into memory simultaneously.

**Root Cause:** `--experimental-vm-modules` causes Jest to create a new `vm.Module` context per test file, causing V8's coverage engine to re-instrument every module for each test file. With 72 test files, this produces ~128K coverage records across 38 JSON files totalling ~269 MB — with one file alone at 175 MB. Without `--merge-async`, `c8` loads all files at once, creating a ~2 GB heap spike that exceeds Node 20's default heap limit on macOS arm64.

### Changes

* `packages/eslint-plugin-fiori-tools/package.json`: Updated the `test` script with two mitigations:
  - Added `--merge-async` to `c8` so V8 coverage files are merged incrementally rather than all at once, avoiding the memory spike during teardown.
  - Added `--maxWorkers=2` to Jest to cap the number of parallel worker processes, reducing the volume of duplicate V8 script contexts.

### GitHub Issues

* [#4545](https://github.com/SAP/open-ux-tools/issues/4545): BUG - Out of memory crash on Node 20 arm64 after test suite completes

- [ ] 🔄 Regenerate and Update Summary

---
📬 [Subscribe to the Hyperspace PR Bot DL](https://url.sap/451kgs) to get the latest announcements and pilot features!

<details>
<summary>PR Bot Information</summary>

**Version:** `1.20.11` | 📖 [Documentation](https://url.sap/dy9ocn) | 🚨 [Create Incident](https://url.sap/budnv9) | 💬 [Feedback](https://url.sap/my4dn3)

- File Content Strategy: Full file content
- LLM: `anthropic--claude-4.6-sonnet`
- GithubContextProvider: [https://github.com/SAP/open-ux-tools/issues/4545](https://github.com/SAP/open-ux-tools/issues/4545)
- Output Template: [Default Template](https://github.tools.sap/intelligent-insights/i2-pull-request/blob/main/src/services/llm/prompts/summary_default_output_template.md)
- Event Trigger: `pull_request.edited`
[...]

PR #4570, opened 2026-04-17
fix(eslint-plugin-fiori-tools): fix heap OOM on macos Node 20 arm64

### LINKED ISSUE #4545: BUG - Out of memory crash on Node 20 arm64 after test suite completes
### Related Feature

### Description

The eslint-plugin-fiori-tools test suite (71 suites, 717 tests) runs successfully but the process crashes with a fatal heap out of memory error immediately after ran all test suites:
```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```
Conditions: Only observed on macos-latest with Node 20.x (arm64). Not reproducible on Node 22.x.

### Steps to Reproduce

Commit a change and create a PR.

### Expected results

The pipeline should run without an error:
CI/CD Pipeline / build (macos-latest, 20.x) (pull_request)

### Actual results

```
❌ > nx run @sap-ux/eslint-plugin-fiori-tools:test

...

Test Suites: 72 passed, 72 total
  Tests:       725 passed, 725 total
  Snapshots:   35 passed, 35 total
  Time:        578.385 s
  Ran all test suites.

<--- Last few GCs --->
  
  [18030:0x120008000]   656492 ms: Scavenge 2008.6 (2067.9) -> 2008.5 (2078.9) MB, 54.83 / 0.00 ms  (average mu = 0.789, current mu = 0.666) allocation failure; 
  [18030:0x120008000]   656642 ms: Scavenge 2015.3 (2078.9) -> 2015.3 (2079.9) MB, 19.50 / 0.00 ms  (average mu = 0.789, current mu = 0.666) allocation failure; 
  [18030:0x120008000]   656955 ms: Scavenge 2016.2 (2079.9) -> 2016.0 (2102.9) MB, 284.21 / 0.00 ms  (average mu = 0.789, current mu = 0.666) allocation failure; 
  
  <--- JS stacktrace --->
  
  FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
  ----- Native stack trace -----
```

OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed

### How can we avoid this

longieirl:
## Root cause

The previous fix (#4566) added `--merge-async` to c8, targeting coverage *merge* time. The full stack trace reveals the OOM is one layer deeper — in Node's native `V8CoverageConnection::WriteProfile` at process exit:

```
RunAtExitCallbacks → StartProfilers → takePreciseCoverage → JsonStringifier → OOM
```

c8 runs as a single orchestrator process that accumulates V8 profiler data for **all 71 test files** before serialising it. With `--experimental-vm-modules`, each test file re-instruments every source module, producing ~269 MB of raw coverage data. `JSON.stringify` of this object requires a contiguous string buffer that exhausts the ~2 GB heap on macos arm64 Node 20. `--merge-async` only affects how c8 handles already-written JSON files — the OOM fires before that step.

## Fix

Replace the c8 process-level wrapper with Jest's built-in `coverageProvider: 'v8'`. Jest's own V8 provider collects coverage **per worker process** (one test file each) via `collect-v8-coverage`, keeping per-process memory bounded. The main Jest process aggregates small per-worker reports — the large serialisation never happens in a single process.

Also removes the unused `c8` devDependency and adds `--maxWorkers=50%` (scales to available CPUs rather than hardcoding 2).

`--experimental-vm-modules` is still required — ESLint 9 uses dynamic `import()` internally and fails without it.

## Changes

- `jest.config.js`: remove `collectCoverage: false`, add `coverageProvider: 'v8'`
- `package.json`: drop c8 wrapper, keep `NODE_OPTIONS='--experimental-vm-modules'`, add `--maxWorkers=50%`

## Test plan

- [ ] CI passes on `macos-latest` Node 20.x without OOM
- [ ] 583 tests pass (verified locally)
- [ ] Coverage output unchanged in quality

Closes #4545

### DIFF
--- a/.changeset/eslint-plugin-fiori-tools-oom-fix-v2.md
+++ b/.changeset/eslint-plugin-fiori-tools-oom-fix-v2.md
@@ -1,0 +1,5 @@
+---
+"@sap-ux/eslint-plugin-fiori-tools": patch
+---
+
+fix(eslint-plugin-fiori-tools): fix heap OOM on macos Node 20 arm64 by switching to jest coverageProvider v8
--- a/.changeset/eslint-plugin-fiori-tools-oom-fix-v2.md
+++ b/.changeset/eslint-plugin-fiori-tools-oom-fix-v2.md
@@ -1,5 +1,0 @@
----
-"@sap-ux/eslint-plugin-fiori-tools": patch
----
-
-fix(eslint-plugin-fiori-tools): fix heap OOM on macos Node 20 arm64 by switching to jest coverageProvider v8
--- a/packages/eslint-plugin-fiori-tools/package.json
+++ b/packages/eslint-plugin-fiori-tools/package.json
@@ -20,10 +20,9 @@
         "format": "prettier --write '**/*.{js,json,ts,yaml,yml}' --ignore-path ../../.prettierignore",
         "lint": "eslint",
         "lint:fix": "eslint --fix",
-        "test": "c8 --reporter=text --reporter=lcov --exclude='test/**' --exclude='**/*.test.ts'  --include='src/**/*.ts' cross-env NODE_OPTIONS='--experimental-vm-modules' jest --ci --forceExit --detectOpenHandles --colors"
+        "test": "cross-env NODE_OPTIONS='--experimental-vm-modules' jest --ci --forceExit --detectOpenHandles --colors --maxWorkers=50%"
     },
     "devDependencies": {
-        "c8": "^11.0.0",
         "cross-env": "10.1.0",
         "eslint": "9.39.1",
         "@typescript-eslint/rule-tester": "8.57.2",
--- a/packages/eslint-plugin-fiori-tools/jest.config.js
+++ b/packages/eslint-plugin-fiori-tools/jest.config.js
@@ -1,7 +1,6 @@
 const config = require('../../jest.base');
 module.exports = {
     ...config,
-    // Coverage is handled by c8 wrapper for worker thread support
-    collectCoverage: false,
+    coverageProvider: 'v8',
     setupFiles: ['<rootDir>/test/global-setup.ts']
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1961,9 +1961,6 @@ importers:
       '@typescript-eslint/rule-tester':
         specifier: 8.57.2
         version: 8.57.2(eslint@9.39.1)(typescript@5.9.3)
-      c8:
-        specifier: ^11.0.0
-        version: 11.0.0
       cross-env:
         specifier: 10.1.0
         version: 10.1.0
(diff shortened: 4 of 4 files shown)


longieirl:
@heimwege @devinea please review

.changeset/eslint-plugin-fiori-tools-oom-fix-v2.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-ux/eslint-plugin-fiori-tools": patch
```
heimwege, line 2: @longieirl  do we need a changeset for this change?

packages/eslint-plugin-fiori-tools/package.json
```diff
@@ -20,10 +20,9 @@
         "format": "prettier --write '**/*.{js,json,ts,yaml,yml}' --ignore-path ../../.prettierignore",
         "lint": "eslint",
         "lint:fix": "eslint --fix",
-        "test": "c8 --reporter=text --reporter=lcov --merge-async --exclude='test/**' --exclude='**/*.test.ts'  --include='src/**/*.ts' cross-env NODE_OPTIONS='--experimental-vm-modules' jest --ci --forceExit --detectOpenHandles --colors --maxWorkers=2"
```
heimwege, line 23: I think there is too much being deleted here. I'm missing at least
`--reporter=text --reporter=lcov` and
`--exclude='test/**' --exclude='**/*.test.ts'  --include='src/**/*.ts'`
Is this change intended?

longieirl: The old c8 command carried those CLI flags because c8 is a separate process that doesn't read Jest config. Now that Jest's built-in v8 provider is used, it reads `jest.base.js` directly, so those flags are redundant, `coverageReporters` and c`ollectCoverageFrom` in jest.base.js provide the exact same behaviour.

.changeset/eslint-plugin-fiori-tools-oom-fix-v2.md
longieirl: Removed changeset.

packages/eslint-plugin-fiori-tools/package.json
longieirl: @heimwege are you ok with these changes?

heimwege: ah, ok. understood. Thanks for the explanation

heimwege (approved):
lgtm, let's give it a try (also because I'm getting annoyed by having to start the runs over and over again because of the heap issue 🤣 )

cianmSAP (approved):

PR #4575, opened 2026-04-20
feat: add Claude Code plugin support

mustafamengutay-sap:
This PR registers the Fiori MCP Server as a plugin in the Claude Code plugin registry for enabling users to find and install it directly from within Claude Code.

Changes

Adds `plugin.json`, the Claude Code plugin registry manifest. Defines the plugin name, description, author, homepage, repository, license, keywords, and version so the server appears in Claude Code's plugin discovery UI.

Adds `.mcp.json` at the package root. This file is automatically picked up by Claude Code when the plugin is enabled, starting the MCP server via npx.

Adds `AGENTS.md`documenting the purpose of `server.json`, `plugin.json`, `.mcp.json`, or AI coding agents and contributors working on this package.

Extends `sync-mcp-manifests.js` (formerly `sync-mcp-server-json.js`), so all manifest files stay in sync automatically after changeset version bumps via the CI pipeline.


packages/fiori-mcp-server/scripts/sync-mcp-manifests.js
```diff
@@ -0,0 +1,57 @@
...
+    const pluginJson = readJson(pluginJsonPath);
+
+    const { version } = pkg;
+
+    // Update top-level version
+    serverJson.version = version;
+
+    // Update version inside packages[0] (npm package entry)
+    if (Array.isArray(serverJson.packages) && serverJson.packages.length > 0) {
+        serverJson.packages[0].version = version;
```
heimwege: That if statement is a bit misleading. we check for at least one `packages` but then only adjust index `0`. So I'm wondering why the others e.g. `serverJson.packages[1]` will not be updated?

heimwege (commented):
Just one nit-pick comment.

packages/fiori-mcp-server/scripts/sync-mcp-manifests.js
mustafamengutay-sap: You're right! 

I've now fixed it to iterate over all entries in packages[], so any future additions to packages[] will receive the updated version automatically.

Thanks @heimwege!

packages/fiori-mcp-server/scripts/sync-mcp-manifests.js
```diff
@@ -0,0 +1,59 @@
...
+    const pluginJson = readJson(pluginJsonPath);
+
+    const { version } = pkg;
+
+    // Update top-level version
+    serverJson.version = version;
+
+    // Update version in all package entries
+    if (Array.isArray(serverJson.packages)) {
+        for (const pkg of serverJson.packages) {
```
heimwege: In general you could would work but as `const pkg` already exists the `const pkg` in the for loop shadows the existing one. Doing so, code is harder to understand.

Please rename the shadowed loop variable in sync-mcp-manifests.js:

```ts
  // Before
  for (const pkg of serverJson.packages) {
  // After
  for (const packageEntry of serverJson.packages) {
      packageEntry.version = version;
  }
```

mustafamengutay-sap: Done! 

Thanks @heimwege

heimwege (approved):
- review comments addressed
- changeset ok
- coverage ok
- did NOT test manually

vinayhospete (dismissed):

vinayhospete:
New proposal to move the plugin out of the MCP server to a more generic plugin like "plugins/SAP-Fiori-tools"

Klaus-Keller (approved):
Thanks @mustafamengutay-sap!
Files and adjustments to publish MCP server as Claude plugin look good.

Klaus-Keller:
> New proposal to move the plugin out of the MCP server to a more generic plugin like "plugins/SAP-Fiori-tools"

If we have more than one plugin this makes sense. For one I'd say we keep it together with the sources and document the files as we've done.

PR #4579, opened 2026-04-21
fix: return webapp path from src for component

815are:
change based on https://ui5.github.io/cli/v5/updates/migrate-v5#component-type
`getWebappPath` return wrong path for new apps based on migrate-v5


heimwege:
@815are does this PR bring anything not already solved in https://github.com/SAP/open-ux-tools/pull/3927? If yes we should maybe adjust https://github.com/SAP/open-ux-tools/pull/3927 🤔

815are:
> @815are does this PR bring anything not already solved in #3927? If yes we should maybe adjust #3927 🤔

Hi @heimwege

I checked #3927, but 3927 does not resolve case with one of our app:
1. migrated to v5 approach
2. there is custom path in configuration like:
```
resources:
  configuration:
    paths:
      src: src-ts
```

if it is ok - I can apply my change directly to #3927

815are:
pushed change directly to `feat/project-access/path-mappings-for-type-component` -> 51e771ba1d30bfd70a99d0b51d979a8bf49965ce

PR #4584, opened 2026-04-22
reusable xml id validation method

Jimmy-Joseph19:
Add validateId method to project-access for reusability across multiple packages


815are (dismissed):
checked code:
1. code changes are easy to understand and approach to use `fast-xml-parser` instead of `xmldom/xmldom`
2. changelog persists
3. I did not test manually

.changeset/funny-socks-rescue.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ux/project-access': patch
```
Klaus-Keller: if we add new functionality to the public interface of `@sap-ux/project-access` it is actually a minor update.

packages/project-access/src/library/index.ts
```diff
@@ -1 +1 @@
-export { getReuseLibs, checkDependencies } from './helpers';
+export { getReuseLibs, checkDependencies, validateId } from './helpers';
```
Klaus-Keller: @Jimmy-Joseph19, I know it is not a part that is touched by this PR, but could you please also change from `export *` to named exports (list all exports) in 
https://github.com/SAP/open-ux-tools/blob/f8e83d3ded539ff43817ffcec95e75ae2d953448/packages/project-access/src/index.ts#L74-L74
to
```typescript
export { getReuseLibs, checkDependencies, validateId } from './library';
```

This would restore consistency in public interface and avoids mistaken public exports from subfolders.

Klaus-Keller (commented):
adding AI verdict:

## Summary                                                                                                                                                             
                                                                                                                                                                Extracts a validateId function into @sap-ux/project-access to consolidate XML element ID uniqueness checking that was previously inline in @sap-ux/fe-fpm-writer.   
  The function supports sync (files provided directly) and async (filesystem scan via appPath) overloads, and switches from @xmldom/xmldom DOMParser to             
  fast-xml-parser.                                                                                                                                                    
                                                                                                                                                                    
  ---                                                                                                                                                                 
  Issues Found
                                                                                                                                                                      
  1. Critical: Async path reads file paths, not file contents (helpers.ts:481-484)                                                                                  
                                                                                                                                                                      
  The async branch calls findFilesByExtension which returns file paths, then passes them to fsEditor.read(path) — this part is actually correct since                 
  mem-fs-editor.read() takes a path and returns the file content. So the bot reviewer's concern about this appears to be a false positive. However, there's still a   
  problem: the async path is completely untested. The test suite only covers the sync overload and a trivial "no options" case. There are no tests exercising appPath 
  with mocked filesystem reads.                                                                                                                                     

  2. Behavioral change: XML parser switch (helpers.ts:378-386)                                                                                                        
  
  The original code used @xmldom/xmldom DOMParser with getElementById(). The new code uses fast-xml-parser with a custom recursive hasElementWithId() that searches   
  for @_id attributes. This is a subtle behavioral difference:                                                                                                      
                                                                                                                                                                      
  - getElementById searches for XML id attributes specifically (per DOM spec).                                                                                        
  - The new code searches for any attribute literally named id at @_id. This should be equivalent for SAPUI5 XML views, but edge cases around namespaced id attributes
   (e.g., xml:id) or custom id attribute handling could differ.                                                                                                       
                                                                                                                                                                    
  3. Deep import from internal path (file.ts:4)                                                                                                                       
                                                                                                                                                                    
  import { findFilesByExtension } from '@sap-ux/project-access/dist/file';                                                                                            
                                                                                                                                                                      
  This pre-existing deep import into /dist/file is not addressed by this PR. The PR already adds findFilesByExtension as an import in helpers.ts — it would be cleaner
   to also re-export it from the package's public API and update the consumer, rather than keeping the fragile /dist/ path import.                                    
                                                                                                                                                                      
  4. Overload complexity (helpers.ts:313-361)                                                                                                                         
  
  Three overload signatures for what is essentially two modes (sync with files, async with appPath/nothing). The third overload (options?: undefined →                
  Promise<boolean>) that always returns true has questionable utility — a caller with no files and no appPath always gets true, which is a no-op validation. Consider
  whether this overload is actually needed or if callers should just skip calling validateId when they have nothing to validate against.                              
                                                                                                                                                                    
  5. Nested function definitions inside implementation (helpers.ts:370-470)                                                                                           
  
  Five helper functions (checkElementIdAvailable, checkIdInValue, hasElementWithId, validateAgainstFiles, and the async IIFE) are all defined inside the validateId   
  function body. This means they are re-created on every call. For a function called in a tight loop (the generateUniqueElementId loop calls it up to 1000 times),  
  this is wasteful. Extract the pure helpers (hasElementWithId, checkIdInValue, checkElementIdAvailable) to module scope.                                             
                                                                                                                                                                    
  6. Missing test coverage for async/appPath overload                                                                                                                 
  
  The test suite only tests:                                                                                                                                          
  - Sync overload with files provided                                                                                                                               
  - Trivial async case with no options (always returns true)                                                                                                          
                                                                                                                                                                    
  There are no tests for the appPath overload, which is the more complex and error-prone path involving filesystem reads, file filtering, and mem-fs-editor           
  integration. This is the path most likely to have bugs.                                                                                                             
                                                                                                                                                                      
  7. validatedIds check placement (helpers.ts:467)                                                                                                                    
                                                                                                                                                                    
  function validateAgainstFiles(files: string[]): boolean {                                                                                                           
      return (                                                                                                                                                      
          files.every((content) => content === '' || checkElementIdAvailable(baseId, content)) &&                                                                     
          !validatedIds?.includes(baseId)                                                                                                                             
      );                                                                                                                                                              
  }                                                                                                                                                                   
                                                                                                                                                                      
  The validatedIds check should come first — it's O(n) at worst but avoids parsing XML entirely if the id is already in the list. Minor optimization but worth noting 
  for the 1000-iteration loop in the caller.                                                                                                                        
                                                                                                                                                                      
  ---                                                                                                                                                               
  Verdict                                                                                                                                                             
                                                                                                                                                                      
  The core idea is sound — centralizing ID validation is a good refactor. However, the PR needs work before merging:
                                                                                                                                                                      
  1. Add tests for the appPath async overload                                                                                                                         
  2. Extract inner helper functions to module scope (performance in hot loop)                                                                                         
  3. Reorder validatedIds check before XML parsing                                                                                                                    
  4. Consider simplifying the overload signatures

.changeset/funny-socks-rescue.md
Jimmy-Joseph19: adapted.

packages/project-access/src/library/index.ts
Jimmy-Joseph19: adapted.

packages/fe-fpm-writer/src/common/file.ts
```diff
@@ -2,7 +2,7 @@ import type { CopyOptions, Editor } from 'mem-fs-editor';
 import type { TabInfo } from '../common/types';
 import { sep, normalize } from 'node:path';
 import { findFilesByExtension } from '@sap-ux/project-access/dist/file';
```
heimwege: Deep import is pre-existing but this PR is the ideal time to address it:
This is a fragile internal path import. `findFilesByExtension` should be re-exported from @sap-ux/project-access's public index. Not strictly this PR's problem to introduce, but since the PR already touches the exports of project-access, it would be clean to include.

heimwege (dismissed):
The deep import mentioned by @Klaus-Keller still exists. But it was pre-existing, so not a blocker

- changeset ok
- did NOT test manually
- coverage is good

815are (dismissed):
reapproving

packages/fe-fpm-writer/src/common/file.ts
Jimmy-Joseph19: adapted. thanks

heimwege (dismissed):
re-approve

Klaus-Keller (approved):
Thanks @Jimmy-Joseph19,
all my comments have been addressed.
Approved from my side.

815are (approved):
reapproving

PR #4594, opened 2026-04-23
chore: remove  --lockfile-only from final install

mikicvi-SAP:
- lints file
- removes `--lockfile-only` from final pnpm install since it is causing issues (locally and on CI)


devinea (approved):

longieirl (approved):
Changes looks good.
We'll be monitoring it anyway once merged.

PR #4601, opened 2026-04-27
fix: Add fragment list to model for custom fragments

mmilko01:
#4602

* Previously, the fragment list was only added to the model when the dialog type was `tableColumn`. This meant other dialog types (e.g., custom fragments not tied to table columns) were missing the fragment list in the model entirely. This change ensures `addFragmentListToModel()` is always called, regardless of the dialog type.


Jimmy-Joseph19 (approved):
Thanks for the fix. Approving.
Did not test locally  and Integration test passed - https://github.com/SAP/open-ux-tools/actions/runs/24999167357
Changeset found.

nikmace (approved):
Change looks good
Did not test manually
Changeset is OK

Issue #4603, opened 2026-04-27
FEATURE - Missing fragment or controller extension files errors to be shown in InfoCenter


IvoSG:
As a SAPUI5 developer using Adaptation Projects,
I want to be notified when a referenced fragment or controller extension file is missing,
so that I can quickly fix broken changes without checking the browser console.

### Description

When a fragment or controller extension file is deleted but its change file is still loaded, the editor produces errors that are hard to find. This feature surfaces clear error messages in the InfoCenter, explaining the missing file and guiding the developer to delete the related change file.

### Technical Design
Enhance the Adaptation Editor validation to detect missing fragment or controller extension references during load. Expose the errors via the InfoCenter with messages pointing to the affected change files.

### Acceptance Criteria
Given a fragment or controller extension file was deleted
When I open the Adaptation Editor
Then an error is shown in the InfoCenter explaining the missing file and how to resolve it

### Notes
Out of scope: cases where change files are deleted and fragment or controller extension files remain orphaned.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #4604, opened 2026-04-27
feat: Missing fragment or controller extension files errors to be shown in InfoCenter 

IvoSG:
Feat for [4603](https://github.com/SAP/open-ux-tools/issues/4603)

Adds clear InfoCenter error messages when Adaptation Projects reference to a deleted fragment or controller extension files. The errors explain the root cause and guide developers to remove the related change files, making issues easier to detect without using the browser console.


packages/preview-middleware-client/src/messagebundle.properties
```diff
@@ -67,6 +67,8 @@ ADP_CREATE_CONTROLLER_EXTENSION_TITLE = Create Controller Extension
 ADP_CREATE_CONTROLLER_EXTENSION_DESCRIPTION = Controller extension with name ''{0}'' was created.
 ADP_ODATA_HEALTH_CHECK_TITLE = OData Service Health Check
 ADP_ODATA_SERVICE_DOWN_DESCRIPTION = The OData service with the {0} endpoint is down. Error: {1}.
+ADP_ORPHANED_CHANGE_ERROR_TITLE = Missing File Detected
+ADP_ORPHANED_FILE_DESCRIPTION = The file "{0}" referenced by change "{1}" was not found. Delete the change file to resolve this error.
```
lfindlaysap:
```suggestion
ADP_ORPHANED_FILE_DESCRIPTION = The "{0}" file referenced by the "{1}" change was not found. To resolve this error, delete the change file.
```

lfindlaysap (changes requested):

avasilev-sap (changes requested):
LGTM, just a minor comment about a for loop.

packages/preview-middleware-client/src/adp/change-file-validator.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+/**
+ * Builds a lookup map from module name patterns to change metadata
+ * for addXML and codeExt changes.
+ *
+ * @param changes record of change objects keyed by flex key
+ * @returns map from module name substring to orphaned change entry
+ */
+function buildModuleNameMap(changes: Record<string, Change>): Map<string, OrphanedChangeEntry> {
```
avasilev-sap, line 59: nit: Please avoid native for loops since maintaining index or keys is error prone . This could be done in more declarative syntax with operators.

```typescript
const isFragmentORCodeExtChange = (
    change: Change
): change is Change & {
    reference: string;
    content: { fragmenthPath: string } | { codeRef: string };
} =>
    !!change.reference &&
    ((change.changeType === 'addXML' && !!change.content?.fragmentPath) ||
        (change.changeType === 'codeExt' && !!change.content?.codeRef));

function buildModuleNameMap(changes: Record<string, Change>): Map<string, OrphanedChangeEntry> {
    return Object.values(changes)
        .map((change) => ({
            ...change,
            // normalize reference
            reference: change.reference?.replaceAll(/\./g, '/')
        }))
        .filter(isFragmentORCodeExtChange)
        .reduce((map, { changeType, content, fileName, fileType, moduleName, reference }) => {
            const path = changeType === 'addXML' ? content.fragmentPath! : content.codeRef!;
            const changeFileName = `${fileName}.${fileType ?? 'change'}`;
            const key = moduleName ?? `${reference}/changes/${path}`;
            const value: OrphanedChangeEntry = {
                changeFileName,
                filePath: path,
                changeType: changeType as 'addXML' | 'codeExt'
            };

            map.set(key, value);

            return map;
        }, new Map<string, OrphanedChangeEntry>());
}
```

nikmace: I'd keep the `for...of` loop here — it's more readable with the branching logic and early `continue`. Declarative chaining (`.map().filter().reduce()`) works well for clean transformations, but when you have multiple conditions and early exits, a loop is the right tool. There's no index management to get wrong.

The "declarative" version with `.map().filter().reduce()` is longer, needs a type guard with a type assertion, has a `.replaceAll` in a `.map()` that changes the shape mid-chain, and the `.reduce()` building a Map is arguably harder to follow than just `map.set()` in a loop.

The one thing I'd improve: the two `if` blocks share most of their logic. You could extract the common parts:

```ts
const path = change.changeType === 'addXML' ? change.content?.fragmentPath : change.content?.codeRef;
if (!path) continue;

const moduleName = change.moduleName ?? `${prefix}/changes/${path}`;
const changeFileName = `${change.fileName}.${change.fileType ?? 'change'}`;
map.set(moduleName, { changeFileName, filePath: path, changeType: change.changeType });

File: packages/preview-middleware-client/src/adp/change-file-validator.ts [hunk 3]
```diff
@@ -0,0 +1,140 @@
...
+}
+
+interface Change {
+    changeType: string;
+    fileName: string;
+    fileType?: string;
+    reference: string;
+    moduleName?: string;
+    content?: ChangeContent;
+}
```
nikmace, line 25: Unfortunately, we are having to duplicate interfaces that can be reused from the `@sap-ux/adp-tooling` package. Importing `@sap-ux/adp-tooling` into preview-middleware is not the right approach, though.

packages/preview-middleware-client/src/adp/change-file-validator.ts
IvoSG: Changed in [30bb3](https://github.com/SAP/open-ux-tools/pull/4604/changes/36e3b4a8dd077ceec2cc1eef94f25ec11ab30bb3)

avasilev-sap (dismissed):

packages/preview-middleware-client/src/adp/change-file-validator.ts
avasilev-sap: I always prefer declarative programing for an enterprise grade projects, easy to read, i am not interested in the logic which determines if the change is for a fragment or controller also with the map operator the change object is cloned so no side effects, also i am not interested in the normalization logic for the reference variable. The method just constructs an entry for the map which is part of the reduce. Again with for loops you need to maintain keys or indexes which is error prone, also mixing declarative with imperative is not consistent. For complex loops let's use declarative programming, for smaller once imperative - that is inconsistency.

packages/preview-middleware-client/src/adp/change-file-validator.ts
IvoSG: Do you have any suggestions here?

packages/preview-middleware-client/src/adp/change-file-validator.ts
IvoSG: Think it's clearer and more readable now.
@nikmace @avasilev-sap Do you all agree that this implementation should remain?

nikmace: I do not agree, but I will not block the PR either; therefore, this comment **can be resolved**.

Replying to the comment above:
1. "for loops you need to maintain keys or indexes" — This is a for...of loop, not for (let i = 0; ...). There are no indexes or keys to maintain. That argument doesn't apply here.
2. "easy to read" — The declarative version requires a type guard with a multi-line type assertion, a .map() that mutates shape, and a .reduce() building a Map. That's not easier to read and maintain than a linear loop with two if blocks.
3. "not interested in the logic which determines if the change is for a fragment or controller" — You're still looking at it in the .filter() type guard. It's not hidden, it's just moved to a different place.
4. "the change object is cloned so no side effects" — The for...of version doesn't mutate anything either. No side effects in both versions.
5. "mixing declarative with imperative is not consistent" — The codebase already mixes both. And consistency within a function matters more than consistency with an abstract principle.

packages/preview-middleware-client/src/adp/change-file-validator.ts
nikmace: Unless we import adp-tooling and make it a single source of truth - no. This is not a blocker, just unfortunate. These could go to `@sap-ux-private/control-property-editor-common` but this module is mainly used for Adaptation Editor

packages/preview-middleware-client/src/adp/change-file-validator.ts
avasilev-sap: The for loops does 3 steps first normalize the reference path, then filters only changes for fragments or controller extension and as a last step creates an entry in a map. In order to understand that i need to read the entire body of the for loop and to understand the code in there. With the operators for example just from the guard name i understand immediately that i work only with fragments and controllers.

avasilev-sap (approved):

packages/preview-middleware-client/src/adp/change-file-validator.ts
IvoSG: Reverted in [b3b3e5](https://github.com/SAP/open-ux-tools/pull/4604/changes/3928603ef87f1a71639001f3f0b605038bb3b3e5) as the final decision

packages/preview-middleware-client/src/adp/change-file-validator.ts
IvoSG: Thanks for the clarification.
I agree that importing @sap-ux/adp-tooling just to avoid duplication is not the right approach, and that control-property-editor-common doesn’t really fit either.
Given that this is not a blocker, I’ll proceed with the local duplication for now and we can revisit consolidation later if a clear shared home emerges.
Resolving this thread.

lfindlaysap (approved):

IvoSG:
image

nikmace (approved):

PR #4611, opened 2026-04-28
fix(ui5-test-writer): app info generation fixes

docirl:
internal issue 38020

### Bug Fix

🐛 Addresses several issues in the `ui5-test-writer` package related to OPA test generation in standalone mode, specifically fixing a crash when the `integration` folder does not exist, and improving `htmlTarget` resolution from `flpSandbox.html`.

### Changes

* `packages/ui5-test-writer/src/fiori-elements-opa-writer.ts`: Refactored the inline standalone write context logic into a dedicated `resolveStandaloneWriteContext` function. This function now:
  - Only attempts to move the `integration` folder to `integration_old` (and update `.gitignore`) if the folder actually exists.
  - When no `integration` folder is present, falls back to adding the `int-test` package script (if missing) and resolving the `htmlTarget` from `flpSandbox.html` if available.
  - Added a helper `checkScriptInPackageJson` to check for existing scripts in `package.json`.
  - Added imports for `getPackageScripts` from `@sap-ux/fiori-generator-shared` and `readHashFromFlpSandbox` from local utils.

* `packages/ui5-test-writer/templates/v4/integration/ListReportJourney.js`: Fixed a template issue by wrapping the "Check table columns and actions" `opaTest` block in a conditional guard to only render when toolbar actions or table columns are present. Also changed `iCheckColumns` to pass `undefined` as the first argument instead of the column count.

* `packages/ui5-test-writer/package.json`: Added `@sap-ux/fiori-generator-shared` as a runtime dependency.

* `packages/ui5-test-writer/tsconfig.json`: Added a project reference for `fiori-generator-shared`.

* `packages/ui5-test-writer/test/unit/fiori-elements.test.ts`: Added two new test cases:
  - Verifies that `htmlTarget` is resolved from `flpSandbox.html` when no `integration` folder exists.
  - Verifies that the `integration` folder is moved to `integration_old` when it exists but no `JourneyRunner.js` is present.

* `packages/ui5-test-writer/test/test-input/LropNoTests/`: Added a new test fixture (`LropNoTests`) including `package.json`, YAML config files, a webapp skeleton, and a `flpSandbox.html` to support the new test scenarios.

* `pnpm-lock.yaml`: Updated lockfile to reflect the new `@sap-ux/fiori-generator-shared` dependency.

<details>

**Version:** `1.20.33`

- Correlation ID: `c4e3dd57-2972-464e-b159-c6fefb8d5abc`
</details>


packages/ui5-test-writer/templates/v4/integration/ListReportJourney.js
```diff
@@ -57,9 +58,10 @@ sap.ui.define([
             <%_ }); -%>
             <%_ } -%>
             <%_ if (tableColumns && Object.keys(tableColumns).length > 0) { -%>
-            Then.onThe<%- startLR %>.onTable().iCheckColumns(<%- Object.keys(tableColumns).length %>, <%- JSON.stringify(tableColumns) %>);
+            Then.onThe<%- startLR %>.onTable().iCheckColumns(undefined, <%- JSON.stringify(tableColumns) %>);
```
cianmSAP, line 61: Seems a little odd but it's necessary I assume?

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
```diff
@@ -99,6 +101,69 @@ export async function generateOPAFiles(
...
+    const { config } = writeContext;
+    let htmlTarget = readHtmlTargetFromQUnitJs(testOutDirPath, editor) ?? config.htmlTarget;
+
+    if (existsSync(join(testOutDirPath, 'integration'))) {
+        editor.move(join(testOutDirPath, 'integration', '**'), join(testOutDirPath, 'integration_old'));
+        await addIntegrationOldToGitignore(basePath, editor);
+    } else {
+        const hasIntTestScript = checkScriptInPackageJson(editor, basePath, 'int-test');
+        if (!hasIntTestScript) {
+            const script = getPackageScripts({ localOnly: false, addTest: true })['int-test'];
```
cianmSAP, line 130: Do you need to do this? 
`updatePackageScript` handles if there are no existing scripts. 
Also if `@sap-ux/fiori-generator-shared` can be avoided I think generally it's preferred not to use it.

cianmSAP (commented):
Couple of small queries

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
cianmSAP: Never mind I missed the de structuring for int test

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.js
docirl: Yes by design, first param will be used to check the number of columns which is causing issues in some projects. Removing this check as the rest of `iCheckColumns` will check for existence of columns anyway

cianmSAP (approved):
Thanks @docirl 
Queries answered
Updates look good

mikicvi-SAP (approved):
Cheers @docirl
- Looks good, tests to cover
- Did not test manually

PR #4615, opened 2026-04-29
feat(preview-middleware): add support for the new FLP sandbox

heimwege:
Implement the new FLP Sandbox bootstrap and refactor shared FLP utilities

### Summary

  - Rewrites sandbox2.ejs to use the new FLP Sandbox bootstrap (SandboxBootTask.js conditional on UI5 major 1, data-sap-ui-boot-manifest conditional on UI5 major 2) instead of the old sap-ushell-config/data-sap-ui-on-init approach
  - Serves fioriSandboxAppConfig.json at every sandbox HTML endpoint directory with beforeFlpStart, afterFlpStart, and restricted.flexibilityServices (WorkspaceConnector + LrepConnector + LocalStorageConnector for non-ADP, matching Sandbox 1)
  - Adds sandbox2BeforeInit.ts as the beforeFlpStart hook and sandbox2AfterInit.ts as the afterFlpStart hook - wiring RTA, card generator, app state reset, and initConnectors after the renderer is up (requires UI5 1.150+)
  - Extends qualifiesForNewSandbox to include UI5 1.150+ builds (previously only major > 1 and legacy-free)
  - Adds flp.useNewSandbox yaml config param (default false) to in to the new Sandbox when the UI5 version qualifies and prerequisites are met
  - Renames init.ts → sandbox1Init.ts and init2.ts → sandbox2BeforeInit.ts for clarity; refactors all shared logic into common.ts (startRtaForAppInstance, addCardGenerationUserAction, and all previously duplicated utilities)
  - Emits a warning at runtime when a legacy appconfig/fioriSandboxConfig.json is detected alongside a Sandbox 2.0-eligible UI5 version and falls back to the classic sandbox
  - Adds `flp.navigateToApp` yaml config param (default `false`) — when set to `true`, the FLP Sandbox 2.0 navigates directly to the app on startup instead of showing the FLP home screen first
  - Adds static file generation support in `generatePreviewFiles` (`packages/create` `add html` CLI): writes `sandbox2.ejs` output and `fioriSandboxAppConfig.json` to disk when the app's `minUI5Version` qualifies and `useNewSandbox: true` is set
  - Fixes several correctness issues identified during review (see below)

###  Test plan

  - [x] UI5 1.144 → sandbox 1 (correct)
  - [x] UI5 1.150 snapshot → sandbox 2 (correct)
  - [x] UI5 1.150 snapshot + useNewSandbox: false → sandbox 1 (correct)
  - [x] UI5 1.150 snapshot + RTA endpoint → sandbox 2, adaptation works, .change file written via WorkspaceConnector
  only (confirmed via FlexConfiguration.getFlexibilityServices())
  - [x] All 24 integration tests green across UI5 1.71–1.149 (sandbox 1 unaffected)


packages/preview-middleware-client/src/flp/sandbox2BeforeInit.ts
```diff
@@ -0,0 +1,23 @@
+import { registerComponentDependencyPaths } from './common';
```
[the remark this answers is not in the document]
heimwege: fixed

packages/preview-middleware-client/test/unit/flp/WorkspaceConnector.test.ts
```diff
@@ -64,35 +63,33 @@ describe('flp/WorkspaceConnector', () => {
...
-                    })
-                );
-                expect(testConnector.storage.fileChangeRequestNotifier).toHaveBeenCalledWith(
-                    'dummyFile',
-                    'create',
-                    change,
-                    undefined
-                );
-            });
+            jest.resetModules();
```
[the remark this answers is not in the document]
heimwege: fixed

packages/preview-middleware/src/base/flp.ts
```diff
@@ -701,6 +792,30 @@ export class FlpSandbox {
...
     }
 
+    /**
+     * Warns once if a legacy 'fioriSandboxConfig.json' file is found when using the new FLP Sandbox.
+     *
+     * @param ui5Version - the resolved UI5 version
+     * @private
+     */
+    private async warnIfLegacySandboxConfigExists(ui5Version: Ui5Version): Promise<void> {
+        const qualifiesForNewSandbox =
```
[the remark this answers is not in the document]
heimwege: fixed

packages/preview-middleware/src/base/flp.ts
```diff
@@ -523,16 +524,62 @@ export class FlpSandbox {
             // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
             this.templateConfig.baseUrl = ('ui5-patched-router' in req && req['ui5-patched-router']?.baseUrl) || '';
             const ui5Version = await this.getUi5VersionFromRequest(req, this.templateConfig.baseUrl);
+            this.templateConfig.ui5.versionMajor = ui5Version.major;
             this.checkDeleteConnectors(ui5Version.major, ui5Version.minor, ui5Version.isCdn);
             if (ui5Version.major === 1 && ui5Version.minor < 120) {
                 this.removeFlexExtensionPointEnabled();
             }
+            await this.warnIfLegacySandboxConfigExists(ui5Version);
```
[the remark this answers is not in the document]
heimwege: It is intended to have the warning every time the flp endpoint is opened

packages/preview-middleware-client/src/flp/sandbox2AfterInit.ts
```diff
@@ -0,0 +1,63 @@
...
+        });
+    }
+
+    // Wire up card generator if enabled
+    if (enableCardGenerator) {
+        const lifecycleService = await container.getServiceAsync<AppLifeCycle>('AppLifeCycle');
+        lifecycleService.attachAppLoaded((event) => {
+            addCardGenerationUserAction(event.getParameter('componentInstance') as unknown as Component, container);
+        });
+    }
```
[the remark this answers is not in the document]
heimwege: This check still exists! I was just moved to flp.ts

packages/preview-middleware/src/base/config.ts
```diff
@@ -66,11 +76,29 @@ export interface TemplateConfig {
...
+         * Comma-separated list of UI5 libraries to preload.
+         *
+         * @remarks Not rendered in sandbox2.ejs — the Sandbox 2.0 bootstrap manages its own library loading.
+         */
         libs: string;
         theme: string;
+        /**
+         * Flexibility service connector configuration.
+         *
+         * @remarks Not rendered in sandbox2.ejs — WorkspaceConnector is wired via the fioriSandboxAppConfig.json `restricted.flexibilityServices` instead.
```
heimwege: Is that true? don't we use ui5.flex to feed `restricted.flexibilityServices`?

packages/preview-middleware-client/src/flp/common.ts
```diff
@@ -26,3 +197,191 @@ export function getFlexSettings(): FlexSettings | undefined {
...
+        } catch (error) {
+            Log.error(`Registering of reuse libs failed. Error:${error}`);
+        }
+    }
+}
+
+/**
+ * Register SAP fonts that are also registered in a productive Fiori launchpad.
+ */
+export function registerSAPFonts() {
```
heimwege: That one is only relevant for the legacy sandbox I guess. move back to `sandbox1Init.ts`? Also check others

packages/preview-middleware-client/src/flp/common.ts
```diff
@@ -26,3 +197,191 @@ export function getFlexSettings(): FlexSettings | undefined {
...
+    };
+    IconPool.registerFont(suiteTheme);
+}
+
+/**
+ * Create Resource Bundle based on the scenario.
+ *
+ * @param scenario to be used for the resource bundle.
+ */
+export async function loadI18nResourceBundle(scenario: Scenario): Promise<ResourceBundle> {
```
heimwege: Also only `sandbox1Init.ts`?

packages/preview-middleware-client/src/flp/common.ts
```diff
@@ -1,4 +1,51 @@
...
+import type { CardGeneratorType } from 'sap/cards/ap/generator';
+import { getError } from '../utils/error.js';
+import { isLowerThanMinimalUi5Version, type Ui5VersionInfo } from '../utils/version.js';
+import { sendInfoCenterMessage } from '../utils/info-center-message.js';
+
+type GlobalErrorEvent = ErrorEvent | PromiseRejectionEvent;
+
+const CONTROLLER_EXTENSION_PATH_REGEX = /\/changes\/coding\/.+\.(js|ts)/;
+
+function extractError(event: GlobalErrorEvent): Error | undefined {
```
nikmace, line 24: Many of the methods are now missing JSDocs in this file

nikmace (changes requested):

packages/preview-middleware-client/src/flp/common.ts
heimwege: Very good finding. Seems like AI only moved function ignoring JSDoc 🙈
Re-added in 8b8b402

devinea (dismissed):
Thanks for the updates. Implementation looks clean.

nikmace (dismissed):
Changes look good
Changeset is OK
Coverage is good
Did not test manually

nikmace:
Screenshot 2026-07-09 at 15 46 47

heimwege:
PLEASE NOTE: unit test currently fail on purpose because we enabled the new sandbox for integration testing ADP. https://github.com/SAP/open-ux-tools/pull/4615/commits/85489e16357ab9d94de004252c55d7f60db71211 will be reverted after testing

=> change has been undone after successfully running the ADP integration tests

packages/preview-middleware-client/src/flp/common.ts
```diff
@@ -232,6 +232,31 @@ export function getFlexSettings(): FlexSettings | undefined {
...
+ * Sandbox 1 behaviour. The base component namespace registration is left untouched.
+ *
+ * @param projectId the ADP variant id (e.g. `adp.v2app`), as provided in flexSettings
+ * @param baseUrl the local base URL prefix for the dev server (may be empty)
+ */
+export function registerAdpChangesResourceRoot(projectId: string, baseUrl = ''): void {
+    if (!projectId) {
+        return;
+    }
+    const namespace = `${projectId.replace(/\./g, '/')}/changes`;
```
heimwege: SONAR will not be happy about this and insist on using replaceAll instead :-D

heimwege: fixed

packages/preview-middleware/src/base/config.ts
```diff
@@ -587,20 +782,13 @@ export async function generatePreviewFiles(
...
-                    await addApp(flpTemplConfig, appManifest, app, logger);
-                } else {
-                    logger.warn(`Could not add route for ${app}`);
-                }
-            }
-        }
+    await addAdditionalApps(flpConfig, basePath, flpTemplConfig, fs, logger);
+
+    if (shouldUseSandbox2(manifest, flpConfig, basePath, fs)) {
+        generateSandbox2Files(flpPath, flpTemplConfig, flpConfig, manifest!, config.adp !== undefined, fs);
```
nikmace: Question 

Could you please verify one of your scenarios?

`generateSandbox2Files` calls `generateSandboxAppConfig(flpTemplConfig, flpConfig, isAdp)` without passing `isCap`, so for CAP projects the generated `fioriSandboxAppConfig.json` will include `LrepConnector` in `restricted.flexibilityServices` — which it shouldn't.

The runtime path (`sandboxAppConfigGetHandler`) correctly derives `isCap` from `this.projectType`. The static file generation path has no equivalent — `generatePreviewFiles` only receives a `MiddlewareConfig` and never detects CAP project type.

Either detect CAP in `generateSandbox2Files` (e.g. check for a `cap` or `cds` indicator in the manifest), or thread an `isCap` param through from the caller.

packages/preview-middleware/src/base/flp.ts
```diff
@@ -535,16 +532,67 @@ export class FlpSandbox {
             // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
             this.templateConfig.baseUrl = ('ui5-patched-router' in req && req['ui5-patched-router']?.baseUrl) || '';
             const ui5Version = await this.getUi5VersionFromRequest(req, this.templateConfig.baseUrl);
-            this.checkDeleteConnectors(ui5Version.major, ui5Version.minor, ui5Version.isCdn);
+            this.templateConfig.ui5.versionMajor = ui5Version.major;
+            this.checkDeleteConnectors(this.templateConfig, ui5Version.major, ui5Version.minor, ui5Version.isCdn);
```
nikmace, line 531: `flpGetHandler` calls `checkDeleteConnectors(this.templateConfig, ...)` directly on the shared instance, but the JSDoc on `checkDeleteConnectors` now explicitly says the config "must be a per-request clone, not the shared this.templateConfig".

`generateSandboxForEditor` does this correctly — it calls `structuredClone(this.templateConfig)` first
(line 339), then passes the clone to `checkDeleteConnectors`.

In `flpGetHandler` the missing clone means `ui5.flex` is permanently stripped from the shared instance
on the first qualifying request (UI5 < 1.84, or CAP project type). Subsequent requests — including
concurrent ones to `sandboxAppConfigGetHandler`, which also reads `this.templateConfig` uncloned —
will inherit the stripped connector list.

Was the clone intentionally skipped here (e.g. because the FLP endpoint is considered single-tenant), or is this an oversight from the refactor?

```js
// In generateSandboxForEditor (the RTA editor path):
const config = structuredClone(this.templateConfig)  // make a copy
checkDeleteConnectors(config, ...)                    // strip connectors from the COPY
render(template, config)                              // render using the copy
// this.templateConfig is untouched

// In flpGetHandler (the FLP preview path):
checkDeleteConnectors(this.templateConfig, ...)       // strip connectors from the SHARED object
render(template, this.templateConfig)                 // render using the shared object
// this.templateConfig is now permanently modified
```

packages/preview-middleware-client/src/flp/sandbox1Init.ts
```diff
@@ -0,0 +1,174 @@
...
+            lifecycleService.attachAppLoaded((event) => {
+                // Prevent starting RTA when the FLP home component (#Shell-home) fires attachAppLoaded before the user navigates to the actual app.
+                if (!globalThis.location.hash || globalThis.location.hash.startsWith('#Shell-home')) {
+                    return;
+                }
+                startRtaForAppInstance(event.getParameter('componentInstance'), flexSettings, ui5VersionInfo);
+            });
+        });
+    }
+    if (enableCardGenerator) {
```
nikmace, line 114: Question

`packages/preview-middleware-client/src/flp/sandbox1Init.ts`, line 114
Same issue at `packages/preview-middleware-client/src/flp/sandbox2AfterInit.ts`, line 47

The UI5 version guard was dropped in the refactor. The old `init.ts` had:
```ts
if (enableCardGenerator && !isLowerThanMinimalUi5Version(ui5VersionInfo, { major: 1, minor: 121 })) {
```
Both files now just check `if (enableCardGenerator)` with no version gate.

On UI5 < 1.121, `sap/cards/ap/generator/CardGenerator` doesn't exist — `sap.ui.require` will silently fail instead of logging a warning to the developer. The author's comment "this check still exists, it was moved to flp.ts" refers to the server-side route guard, not this client-side call guard — these are different things.

nikmace (changes requested):

packages/preview-middleware/src/base/config.ts
heimwege: agree I'll fix that one

packages/preview-middleware/src/base/flp.ts
heimwege: The UI5 version is resolved from either:
  1. The yaml config (static, never changes)
  2. The CDN version list (fetched once, same result every request)

So after the first request sets `versionMajor` and potentially strips `WorkspaceConnector` from `ui5.flex`, every subsequent request will resolve the same UI5 version and produce the same mutation. The final state of `this.templateConfig` is identical to what a fresh clone would look like. No request ever sees a different outcome.

The UI5 version is fixed per server instance (from config or CDN resolution), so `checkDeleteConnectors` always produces the same result; the mutation is safe. `generateSandboxForEditor` clones because it also sets `flexSettings`, features, and calls `remapResourcesForPath` - those ARE per-request and must not leak.

packages/preview-middleware-client/src/flp/sandbox1Init.ts
heimwege: I got that review comment from the bot so many times 😄 The guard is not gone it was just moved to preview-middleware `flp.ts` ([link](https://github.com/SAP/open-ux-tools/blob/9eb0da8e05879612c70f6ec2ec014e6fb227a5c9/packages/preview-middleware/src/base/flp.ts#L684)) and the bot just can't figure out that move. This is a false positive

packages/preview-middleware/src/base/config.ts
heimwege: fixed in d67af5f478785da884812d14a4cb3f56fecfa020

.changeset/adp-manifest-ui5-prefix-variant-id.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-ux/adp-tooling": patch
```
mmilko01, line 2: I think the `preview-middleware` and `preview-middleware-client` should be added in the changeset

mmilko01 (approved):
Changeset OK
Changes look good
Excellent test coverage
Comments addressed
Integration test execution successful

Image

nikmace (approved):
Changes look good
Coverage is good
Changesets are OK
Did not test manually

PR #4618, opened 2026-04-29
fix(odata-service-inquirer): System selection switching issue

IainSAP:
### Problem

`ConnectionValidator.isAuthRequired()` was duplicating connection logic that had already run during `validateUrl`/`validateAuth`. The method accepted optional `urlString`, `client`, and `ignoreCertError` parameters and re-issued HTTP requests when the URL or client changed — even when the validity state was already populated from a prior connection attempt. This led to:

- Redundant HTTP requests
- Stale or conflicting validity state when called with a different URL
- Incorrect `false` return when the connection result was indeterminate (e.g. 503)

### Fix

Simplified `isAuthRequired()` to be a pure read of the existing validity state set during the last connection attempt:

- Returns `false` if no connection has been attempted yet (`_validatedUrl` is unset)
- Returns `false` if the connection is already authenticated
- Returns `this.validity.authRequired` (which may be `true`, `false`, or `undefined`) for all other cases

The method no longer accepts parameters or makes HTTP requests — callers must validate the connection first via `validateUrl`/`validateAuth`.

### Changes

- **`connectionValidator.ts`**: Removed the parameter-driven re-request logic from `isAuthRequired()`; method is now a stateless read of `this.validity`
- **`connectionValidator.test.ts`**: Updated tests to reflect the simplified contract — uses `resetConnectionState()` + `validateAuth()` to set up state before calling `isAuthRequired()`

### Impact

- Removes ~35 lines of redundant network logic
- No functional regression for callers that already validate before checking auth
- `isAuthRequired()` now correctly returns `undefined` when auth status cannot be determined (e.g. server returned 503)


longieirl (approved):
Changeset is good.
Less code === good
Tests updated to reflect this change.

cianmSAP (approved):
Update makes sense based on description
Changeset and tests look good

Issue #4621, opened 2026-04-30
BUG - Prioritize deployed fragments over local workspace versions in offline adaptation project editor


IvoSG:
### Description
When previewing an adaptation project connected to an ABAP system, locally created fragments and controllers are not used. Instead, UI5 loads inlined module content (fragment XMLs and controller JS) returned via the LREP flex data response, preventing requests to the local workspace.

### Steps to Reproduce
1. Connect an adaptation project to an ABAP system
2. Create or modify a fragment/controller locally
3. Run preview
4. Observe that deployed (backend) versions are used instead of local files

### Expected results
UI5 should prioritize and load local workspace fragments and controllers during preview.

### Actual results
UI5 uses inlined module content from the ABAP LREP response and does not request local workspace versions.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [x] Mac OS
- [x] Windows
- [ ] Other

### Problem
The ABAP LREP flex data response contains inlined module content (fragments and controllers), which prevents UI5 from issuing HTTP requests. As a result, the ADP proxy cannot serve local workspace files, causing backend versions to be used instead.

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #4622, opened 2026-04-30
fix: Prioritize local workspace fragments over deployed versions in adaptation project editor

IvoSG:
https://github.com/SAP/open-ux-tools/issues/4621

- When previewing an adaptation project connected to an ABAP system, locally created fragments and controllers are now correctly prioritized over their deployed counterparts. The LREP flex data response from the ABAP system includes inlined module content (fragment XMLs, controller JS) which prevented UI5 from requesting local workspace versions. The fix strips these inlined modules from the response so that UI5 falls back to HTTP requests, which the existing ADP proxy serves from the local workspace. Flex changes are left untouched as UI5 deduplicates them natively by fileName.


packages/preview-middleware/src/base/flp.ts
```diff
@@ -1251,6 +1258,51 @@ export class FlpSandbox {
...
+    private registerLrepFlexDataFilter(adp: AdpPreview): void {
+        const provider = adp.serviceProvider;
+        if (!provider) {
+            return;
+        }
+
+        const lrepFlexDataPath = '/sap/bc/lrep/flex/data/';
+        this.router.get(`${lrepFlexDataPath}*`, async (req: Request, res: Response, next: NextFunction) => {
+            try {
+                const localModulePaths = await readLocalModulePaths(this.project, this.logger);
```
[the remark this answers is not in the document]
IvoSG: This endpoint is only hit on page loads, not during continuous editing. The glob scan is fast for typical ADP projects (a handful of fragments/controllers). Adding a cache with file-watcher invalidation would be over-engineering at this point. If profiling later shows this as a bottleneck, it can be optimized then.

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1251,6 +1258,51 @@ export class FlpSandbox {
...
+
+                const filtered = stripLocalModulesFromLrepResponse(
+                    responseData as Record<string, unknown>,
+                    localModulePaths,
+                    this.logger
+                );
+
+                res.status(200).json(filtered);
+            } catch (error) {
+                this.logger.warn(`Failed to filter LREP flex data, falling back to unfiltered response: ${error}`);
```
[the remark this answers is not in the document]
IvoSG: Using next() (not next(error)) is intentional to fall back to the backend proxy and keep preview working if filtering fails, and concerns about duplicate backend calls are invalid since adp.proxy doesn’t call the ABAP backend or handle the LREP path.

packages/adp-tooling/src/preview/adp-preview.ts
```diff
@@ -124,6 +124,13 @@ export class AdpPreview {
         return this.projectTypeValue;
     }
 
+    /**
+     * @returns the ABAP service provider used to connect to the backend, or undefined if not initialized or in CF build mode.
+     */
+    get serviceProvider(): AbapServiceProvider | undefined {
```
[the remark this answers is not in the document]
IvoSG: The only cross-method usage is in onChangeRequest, where a simple truthiness guard suffices — in CF build mode this.provider is indeed undefined, and annotation changes shouldn't occur in that mode anyway.

The getter already returns AbapServiceProvider | undefined, so external callers are unaffected.

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1251,6 +1258,51 @@ export class FlpSandbox {
...
+        const lrepFlexDataPath = '/sap/bc/lrep/flex/data/';
+        this.router.get(`${lrepFlexDataPath}*`, async (req: Request, res: Response, next: NextFunction) => {
+            try {
+                const localModulePaths = await readLocalModulePaths(this.project, this.logger);
+                if (localModulePaths.size === 0) {
+                    next();
+                    return;
+                }
+
+                const backendPath = req.originalUrl;
```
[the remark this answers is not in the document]
IvoSG: Changed in [e0d85a](https://github.com/SAP/open-ux-tools/pull/4622/changes/e63618f5f8676b00b0681b70431a5d4f8de0d85a)

packages/preview-middleware/src/base/flex.ts
```diff
@@ -34,6 +41,82 @@ export async function readChanges(
...
+export async function readLocalModulePaths(project: ReaderCollection, logger: Logger): Promise<LocalModulePaths> {
+    const modulePaths: LocalModulePaths = new Set();
+
+    const moduleFiles = await project.byGlob('/**/changes/{fragments/**,coding/**}');
+    for (const file of moduleFiles) {
+        const filePath = file.getPath();
+        const changesIdx = filePath.lastIndexOf('/changes/');
+        if (changesIdx !== -1) {
+            modulePaths.add(filePath.substring(changesIdx + '/changes/'.length));
+        }
```
mmilko01, line 110: I think that this logic for checking if the `filePath` contains `/changes/` can be removed. Above the `moduleFiles` are taken based on a glob that insures only files containing `/changes/` are considered.

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1251,6 +1258,54 @@ export class FlpSandbox {
...
+                const localModulePaths = await readLocalModulePaths(this.project, this.logger);
+                if (localModulePaths.size === 0) {
+                    next();
+                    return;
+                }
+
+                const queryString = req.originalUrl.includes('?')
+                    ? req.originalUrl.substring(req.originalUrl.indexOf('?'))
+                    : '';
+                const backendPath = `${req.path}${queryString}`;
```
mmilko01: I might be wrong, but I think that you can use `req.url` as it already contains the `queryString` as well.

packages/preview-middleware/src/base/flex.ts
IvoSG: Fixed in [b8bfba4](https://github.com/SAP/open-ux-tools/pull/4622/changes/7fd0c9c59373c43e6f0b998642933aa3fb8bfba4)

packages/preview-middleware/src/base/flp.ts
IvoSG: Fixed in [10e5d](https://github.com/SAP/open-ux-tools/pull/4622/changes/272ba61119d30f47dc1a6953aa07cddcabe10e5d)

mmilko01 (commented):
Changeset OK
Excellent test coverage
Changes look good
Comments addressed
Did not test manually

mmilko01 (dismissed):

packages/preview-middleware/src/base/flex.ts
```diff
@@ -34,6 +41,79 @@ export async function readChanges(
...
+ * @param logger logger instance
+ * @returns set of relative module paths under /changes/
+ */
+export async function readLocalModulePaths(project: ReaderCollection, logger: Logger): Promise<LocalModulePaths> {
+    const modulePaths: LocalModulePaths = new Set();
+
+    const moduleFiles = await project.byGlob('/**/changes/{fragments/**,coding/**}');
+    for (const file of moduleFiles) {
+        const filePath = file.getPath();
+        modulePaths.add(filePath.substring(filePath.lastIndexOf('/changes/') + '/changes/'.length));
```
heimwege: A comment would be helpful to explain the intent.
because actually if someone genuinely had a fragment path with two `/changes/` segments below the flex folder (e.g. `.../changes/fragments/changes/Weird.xml`), `lastIndexOf` would over-strip. Theoretical case, I know but It took a while to understand this

```ts
// lastIndexOf ensures we anchor on the *flex* /changes/ segment even if
// the virtual path contains an earlier /changes/ component (e.g. a project
// root that includes the word "changes")
```

packages/preview-middleware/src/base/flex.ts
```diff
@@ -34,6 +41,79 @@ export async function readChanges(
...
+        return responseData;
+    }
+
+    if (!responseData.modules || typeof responseData.modules !== 'object') {
+        return responseData;
+    }
+
+    const originalModules = responseData.modules as Record<string, unknown>;
+    const filteredEntries = Object.entries(originalModules).filter(([key]) => {
+        const changesIdx = key.lastIndexOf('/changes/');
```
heimwege: same here (comment)

packages/adp-tooling/src/preview/adp-preview.ts
```diff
@@ -124,6 +124,13 @@ export class AdpPreview {
         return this.projectTypeValue;
     }
 
+    /**
+     * @returns the ABAP service provider used to connect to the backend, or undefined if not initialized or in CF build mode.
```
heimwege, line 128: nit-pick: `get projectType` also has the type in JSDoc. Should be in sync. Probably remove type from `get projectType`

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1251,6 +1258,51 @@ export class FlpSandbox {
...
+     *
+     * @param adp AdpPreview instance with access to the ABAP service provider
+     */
+    private registerLrepFlexDataFilter(adp: AdpPreview): void {
+        const provider = adp.serviceProvider;
+        if (!provider) {
+            return;
+        }
+
+        const lrepFlexDataPath = '/sap/bc/lrep/flex/data/';
```
heimwege, line 1289: This is a constant, right? If you want it to not be changed you need to 
```suggestion
        const lrepFlexDataPath = '/sap/bc/lrep/flex/data/' as const;
```

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1251,6 +1258,51 @@ export class FlpSandbox {
...
+        this.router.get(`${lrepFlexDataPath}*`, async (req: Request, res: Response, next: NextFunction) => {
+            try {
```
heimwege: All other functions ifn `flp.ts` use a dedicated handler function to reduce cognitive complexity. Please update the code to align with the existing conventions in this class.

packages/preview-middleware/src/base/flp.ts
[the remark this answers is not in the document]
heimwege: @IvoSG I think with cache here is meant "move this line 3 rows above (outside the handler) and you are done". Then it will be created once and re-used inside the handler?

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1251,6 +1258,51 @@ export class FlpSandbox {
...
+                const localModulePaths = await readLocalModulePaths(this.project, this.logger);
+                if (localModulePaths.size === 0) {
+                    next();
+                    return;
+                }
+
+                const backendPath = req.url;
+                const response = await provider.get(backendPath);
+                const responseData = typeof response === 'string' ? JSON.parse(response) : response;
```
heimwege: I'm confused. Isn't `provider.get` returning `AxiosResponse`, so `typeof response === 'string'` is always false? And isn't `responseData` from `AxiosResponse` `response.data`? If we pass the complete axios response, `stripLocalModulesFromLrepResponse` will check for `responseData.modules` and fail because that does not exist according to type `AxiosResponse`

packages/preview-middleware/test/unit/base/flp.test.ts
```diff
@@ -1987,4 +1987,77 @@ describe('initAdp', () => {
         expect(flp.rta?.options?.isCloud).toBe(false);
         expect(flp.rta?.options?.isCloudFoundry).toBe(true);
     });
+
+    test('initAdp registers lrep flex data filter when serviceProvider is available', async () => {
+        const mockProvider = {
+            get: jest.fn().mockResolvedValue(
+                JSON.stringify({
```
heimwege: That returns a `string` instead of `AxiosResponse`. So the test passes but for the wrong reason I guess

packages/preview-middleware/src/base/flp.ts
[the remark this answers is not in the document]
heimwege: That's a bit misleading. Why not adding a comment so that future checks will know the reasoning behind this?

e.g.
```ts
// Fall through to the next middleware in the server stack (e.g. backend-proxy-middleware
// if configured) so that preview remains functional even if filtering fails.
```

heimwege (commented):
See comments. Some nit-picks but also I bug (`provider.get()` return type)

packages/preview-middleware/src/base/flp.ts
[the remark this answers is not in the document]
IvoSG: **readLocalModulePaths** is intentionally kept inside the handler. The **ReaderCollection** uses **byGlob** on each call, which ensures newly created fragments are picked up immediately without restarting the dev server — essential for the live development workflow.

packages/preview-middleware/src/base/flp.ts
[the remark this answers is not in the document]
IvoSG: Fixed in [51fbb44](https://github.com/SAP/open-ux-tools/pull/4622/changes/6ca669f5494f6330566880b3a2d7c07ea51fbb44)

packages/preview-middleware/src/base/flex.ts
IvoSG: Fixed in [51fbb44](https://github.com/SAP/open-ux-tools/pull/4622/changes/6ca669f5494f6330566880b3a2d7c07ea51fbb44)

packages/preview-middleware/src/base/flex.ts
IvoSG: Fixed in [51fbb44](https://github.com/SAP/open-ux-tools/pull/4622/changes/6ca669f5494f6330566880b3a2d7c07ea51fbb44)

packages/adp-tooling/src/preview/adp-preview.ts
IvoSG: Fixed in [51fbb44](https://github.com/SAP/open-ux-tools/pull/4622/changes/6ca669f5494f6330566880b3a2d7c07ea51fbb44)

packages/preview-middleware/src/base/flp.ts
IvoSG: Fixed in [51fbb44](https://github.com/SAP/open-ux-tools/pull/4622/changes/6ca669f5494f6330566880b3a2d7c07ea51fbb44)

packages/preview-middleware/src/base/flp.ts
IvoSG: Fixed in [51fbb44](https://github.com/SAP/open-ux-tools/pull/4622/changes/6ca669f5494f6330566880b3a2d7c07ea51fbb44)

packages/preview-middleware/src/base/flp.ts
IvoSG: Fixed in [51fbb44](https://github.com/SAP/open-ux-tools/pull/4622/changes/6ca669f5494f6330566880b3a2d7c07ea51fbb44)

packages/preview-middleware/test/unit/base/flp.test.ts
IvoSG: Fixed in [51fbb44](https://github.com/SAP/open-ux-tools/pull/4622/changes/6ca669f5494f6330566880b3a2d7c07ea51fbb44)

packages/preview-middleware/src/base/flp.ts
heimwege: If the backend ever returns a pre-parsed object (which Axios does by default when the Content-Type is `application/json`), `response.data` would already be an object, and `JSON.parse` on it would throw because `JSON.parse` expects a string.

I think the correct pattern would be
```ts
  const responseData = typeof response.data === 'string'
      ? JSON.parse(response.data)
      : response.data;
```

packages/preview-middleware/test/unit/base/flp.test.ts
heimwege: Actually this (string being returned) is one of the possible scenarios, Axios returning an object would be the other one that will currently crash (see https://github.com/SAP/open-ux-tools/pull/4622#discussion_r3235053003)

IvoSG:
image

packages/preview-middleware/src/base/flp.ts
IvoSG: Addressed in [cbae57](https://github.com/SAP/open-ux-tools/pull/4622/changes/2ad35849d8a8f452e24731a72142ee2035cbae57)

packages/preview-middleware/test/unit/base/flp.test.ts
IvoSG: Addressed in [cbae57](https://github.com/SAP/open-ux-tools/pull/4622/changes/2ad35849d8a8f452e24731a72142ee2035cbae57)

heimwege (dismissed):
- changeset ok
- coverage good
- review comments addressed
- did NOT test manually

mmilko01 (commented):
Re-approving

mmilko01 (dismissed):
Re-approving

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1251,6 +1260,73 @@ export class FlpSandbox {
...
+    /**
+     * Registers a middleware that intercepts LREP flex data responses to apply
+     * local workspace overrides: stripping inlined modules for active local files,
+     * removing the loadModules pre-fetch flag, injecting correct moduleNames, and
+     * suppressing deployed changes for orphaned local files (change record deleted).
+     *
+     * @param adp AdpPreview instance with access to the ABAP service provider
+     * @param localModuleState pre-scanned local module state
+     */
+    private registerLrepFlexDataFilter(adp: AdpPreview, localModuleState: LocalModuleState): void {
```
heimwege, line 1278: The module state is scanned once at startup. If a user adds or deletes a local fragment/change file during a running preview session, the filter will be stale. The missing live-reload awareness worth documenting in the registerLrepFlexDataFilter JSDoc.

packages/preview-middleware/src/base/flex.ts
```diff
@@ -34,6 +52,236 @@ export async function readChanges(
...
+        const reference = typeof c.reference === 'string' ? c.reference : '';
+        const changeType = c.changeType;
+        const localPath = extractLocalModulePath(changeType, content);
+
+        if (!localPath || !active.has(localPath) || !reference) {
+            return change;
+        }
+        const prefix = reference.replaceAll('.', '/');
+        // codeExt modules are JS — strip .js so the path is a valid UI5 module ID
+        const modulePathSuffix = changeType === 'codeExt' ? localPath.replace('.js', '') : localPath;
```
heimwege: `String.replace` replaces only the first occurrence.
```ts
localPath.endsWith('.js') ? localPath.slice(0, -3) : localPath
```
Might be the safer approach in the unlikely case of paths like `coding/my.js.util.js`

packages/preview-middleware/src/base/flp.ts
```diff
@@ -1251,6 +1260,73 @@ export class FlpSandbox {
...
+            const filtered = stripLocalModulesFromLrepResponse(
+                responseData as Record<string, unknown>,
+                localModuleState,
+                this.logger
+            );
+
+            res.status(200).json(filtered);
+        } catch (error) {
+            this.logger.error(
+                `LREP flex data filter failed for '${req.url}' — falling back to unfiltered backend response. Local workspace files may not take effect. Error: ${error}`
```
heimwege:
```suggestion
                `LREP flex data filter failed for '${req.url}' — falling back to unfiltered backend response. Local workspace files may not take effect. Error: ${error?.message ?? error}`
```

heimwege (commented):
see comments and I'm wondering if we should create a utility file for adp (similar to [utils/cards.ts](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/src/base/utils/cards.ts)) because the size of flp.ts is already quite large. Not necessarily in this PR but just something that came to my mind. What do you think?

IvoSG:
> see comments and I'm wondering if we should create a utility file for adp (similar to [utils/cards.ts](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/src/base/utils/cards.ts)) because the size of flp.ts is already quite large. Not necessarily in this PR but just something that came to my mind. What do you think?

Yes, I think it's a good idea.

The case for it: flp.ts at 1,392 lines and the ADP surface — initAdp, setupAdpCommonHandlers, setupAdpCloud, registerLrepFlexDataFilter, lrepFlexDataFilterHandler, plus the related helpers and types — is a coherent slice that tests, evolves, and breaks together. Pulling it into utils/adp.ts (or a dedicated base/adp/ folder) would mirror what we already did with flex.ts and make FlpSandbox smaller and more focused.

The catch: unlike utils/cards.ts (pure functions over inputs), the ADP methods are stateful — they touch this.router, this.logger, this.fs, this.utils, this.adp. So a clean extraction either (a) takes those as args to free functions, or (b) introduces a small AdpSandbox collaborator that FlpSandbox composes. Both are fine, (b) is closer to what the code structure already implies, but it is a bigger change. Worth scoping the refactor explicitly when picked up — and definitely not in this PR, where mixing it with the local-files-over-deployed fix would make the diff much harder to review.

packages/preview-middleware/src/base/flp.ts
IvoSG: Done in [f79954](https://github.com/SAP/open-ux-tools/pull/4622/changes/31bb9cb480305a81fc1d9e231704b4083df79954)

packages/preview-middleware/src/base/flex.ts
IvoSG: Done in [f79954](https://github.com/SAP/open-ux-tools/pull/4622/changes/31bb9cb480305a81fc1d9e231704b4083df79954)

packages/preview-middleware/src/base/flp.ts
IvoSG: Done in [f79954](https://github.com/SAP/open-ux-tools/pull/4622/changes/31bb9cb480305a81fc1d9e231704b4083df79954)

heimwege (approved):
- coverage is great
- changeset ok
- review comments addressed
- did NOT test manually

mmilko01 (approved):
Re-approving

PR #4610, opened 2026-04-28
feat(sap-systems-ext): 37892 pre populate system from adt

korotkovao:
### Changes

**`sap-ux-sap-systems-ext`**
- `createSystemCommandHandler` now accepts an optional `prePopulatedSystem?: BackendSystem` argument (already on this branch from previous commit)
- `renderWebapp.ts`: when panel view type is `Create` and a `backendSystem` is provided, a `SYSTEM_INFO` message is dispatched to the webview after `CREATE_NEW_SYSTEM` so the pre-populated data is displayed in the panel

**`@sap-ux/sap-systems-ext-webapp`**
- `reducers.ts`: the `SYSTEM_INFO` reducer case now spreads existing state to preserve the `addNewSapSystem` flag (previously it was reset to `false`, hiding the system type selector)
- `SystemTypes.tsx`: added `systemType` prop and `selectedKey` to the dropdown so ABAP On-Premise is pre-selected when launched from ADT
- `SystemMain.tsx`: passes `systemInfo?.systemType` down to `SystemTypes`

### Acceptance Criteria

- [x] When launching the Fiori generator from ADT for an unsaved system and selecting "Create new system", the connection manager opens with **URL**, **Client** and **ABAP On-Premise** pre-selected from the ADT request
- [x] When no pre-populated system data is provided (normal flow), the panel opens as a blank create form — no regression
- [x] When the system is already saved, the connection manager is not opened — no regression

PR #4645, opened 2026-05-07
fix(sap-systems-ext): add new create new command

cianmSAP:
Internal bug 38106

Fix for issue from https://github.com/SAP/open-ux-tools/pull/4610 

Fixes a bug where clicking "Add New System" after viewing a system would pre-fill the form with the viewed system's data instead of showing an empty form. The root cause was VSCode automatically passing the currently focused tree item as an argument to toolbar commands. The solution introduces a new internal command `sap.ux.tools.sapSystems.createNew` specifically for the UI button that always creates an empty form, while preserving the existing sap.ux.tools.sapSystems.create command for backward compatibility with external tools that need to pre-fill system data. All changes are backward compatible with no breaking changes to the public API.

### DIFF
--- a/.changeset/green-boxes-rule.md
+++ b/.changeset/green-boxes-rule.md
@@ -1,0 +1,5 @@
+---
+'sap-ux-sap-systems-ext': patch
+---
+
+adds new internal create command to avoid collisions with other registered commands
--- a/packages/sap-systems-ext/package.json
+++ b/packages/sap-systems-ext/package.json
@@ -121,6 +121,11 @@
             },
             {
                 "command": "sap.ux.tools.sapSystems.create",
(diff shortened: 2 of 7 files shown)


korotkovao (dismissed):
Thanks @cianmSAP 
I like this approach
Straightforward code ✅ 
Tests ✅ 
Have not tested locally

mikicvi-SAP (approved):
Thanks @cianmSAP 
- clean and provides a proper API for other tools, I like this approach
- well doc'd
changeset ✅

devinea (approved):
Code changes look good to add new command. 
changeset ✅

korotkovao (approved):

PR #3541, opened 2025-08-18
feat: mcp server package

815are:
mcp server with 4 tools:
- `list-fiori-apps`
- `list-functionalities`
- `get-functionality-details`
- `execute-functionality`

see readme for more information -> https://github.com/SAP/open-ux-tools/blob/feat/fioriMcpServer/packages/fiori-mcp-server/README.md

To run inspector locally:
1. build `packages/fiori-mcp-server`
2. run launch `fiori-mcp-server: Launch inspector` from `.vscode/launch.json`

Please note: new package currently is listed as `private` to avoid publishing

PR #4649, opened 2026-05-08
feat(fiori-mcp-server): promote generation tools to top-level MCP tools

devinea:
## Summary

- Promotes 4 static functionalities out of the 3-step \`list → get_details → execute\` workflow into dedicated top-level MCP tools, reducing generation round-trips from 3 calls to 1
- New tools: \`list_sap_systems\`, \`download_odata_service_metadata\`, \`generate_fiori_app_odata\`, \`generate_fiori_app_cap\`
- \`list_sap_systems\` and \`download_odata_service_metadata\` are environment-aware: on SAP Business Application Studio they use BTP destinations (\`listDestinations\`), on VSCode they use the Fiori tools system store
- \`download_odata_service_metadata\` returns a \`destination\` field on BAS; the tool description instructs the LLM to pass all returned fields (including \`destination\`) directly into the generator service config
- Removes \`generate-fiori-ui-application\`, \`generate-fiori-ui-application-cap\`, and \`fetch-service-metadata\` from \`FUNCTIONALITIES_DETAILS\` / \`FUNCTIONALITIES_HANDLERS\`; their implementation files are moved to \`src/tools/\` and \`src/tools/services/\`
- Updates \`list_functionality\` description to clearly communicate it covers app-modification operations only (add/delete page, controller extension, manifest property changes)
- Errors from \`download_odata_service_metadata\` return a structured \`{ status: "Error", message }\` response instead of propagating as unhandled exceptions; a typed \`FetchServiceMetadataOutputSchema\` is added to \`types/output.ts\`
- Entity name in generator config is normalised (strips wrapping single quotes) before generation
- Generator wrappers use Zod-inferred types (\`GeneratorConfigOData\` / \`GeneratorConfigCAP\`) instead of \`Record<string, unknown>\`
- Dead code removed: \`services/filestore.ts\`, \`services/indexer-simple.ts\`, \`services/types/index.ts\`, unused vector types, \`type Test\` in \`basic.ts\`, \`removeService()\` in \`serviceStore.ts\`, old \`details.ts\` / \`index.ts\` functionality wrappers
- Removes obsolete \`tgz:rename\` script from \`package.json\`

## Motivation

The promoted functionalities have fully static, compile-time schemas. Requiring clients to go through 3 round-trips (list → get_details → execute) for generation tasks was unnecessary overhead — both in latency and token cost. They are now directly callable with a single tool invocation.

The 3-step workflow is retained for genuinely dynamic functionalities (add/delete page, controller extension, manifest property changes) where schemas are built at runtime from the live project state.

## Tool inventory before/after

| Tool | Status |
|---|---|
| \`search_docs\` | unchanged (maxResults default 25→10) |
| \`list_fiori_apps\` | unchanged |
| \`list_functionality\` | description updated — modification operations only |
| \`get_functionality_details\` | unchanged |
| \`execute_functionality\` | unchanged |
| \`list_sap_systems\` | **new** — BAS: destinations; VSCode: system store |
| \`download_odata_service_metadata\` | **new** — BAS-aware, returns \`destination\` field |
| \`generate_fiori_app_odata\` | **new** — promoted, BAS destination support in service config |
| \`generate_fiori_app_cap\` | **new** — promoted |

## Test plan

- [ ] All unit tests pass
- [ ] \`list_functionality\` no longer returns generation/fetch functionality IDs
- [ ] New tools are present in \`list_tools\` response
- [ ] \`generate_fiori_app_odata\` and \`generate_fiori_app_cap\` generate apps correctly end-to-end
- [ ] \`download_odata_service_metadata\` fetches and writes \`metadata.xml\`; returns \`destination\` on BAS
- [ ] \`list_sap_systems\` returns stored systems without credentials on VSCode; returns destinations on BAS


packages/fiori-mcp-server/src/types/input.ts
```diff
@@ -62,9 +62,22 @@ export const ExecuteFunctionalityInputSchema = zod
...
+});
+
 export const DocSearchInputSchema = zod.object({
     query: zod
         .string()
         .min(2)
         .describe('The search query for fiori elements, annotations, sapui5, fiori tools documentation')
 });
+
+export type DownloadODataServiceMetadataInput = zod.infer<typeof DownloadODataServiceMetadataInputSchema>;
```
heimwege: The diff adds `DownloadODataServiceMetadataInput` to both `src/types/input.ts` (as a type export at the bottom) and `src/types/index.ts` (re-exported via the schema). Having the same type defined and exported in two places creates unnecessary duplication. The type in `input.ts` should be removed; only the schema const needs to live there, and the type should be derived once in `index.ts`.

packages/fiori-mcp-server/src/server.ts
```diff
@@ -166,6 +172,18 @@ export class FioriFunctionalityServer {
...
+                        break;
+                    case 'download_odata_service_metadata':
+                        result = await downloadODataServiceMetadata(args as DownloadODataServiceMetadataInput);
+                        break;
+                    case 'generate_fiori_app_odata':
+                        result = await generateFioriAppOData(args as Record<string, unknown>);
+                        break;
+                    case 'generate_fiori_app_cap':
+                        result = await generateFioriAppCap(args as Record<string, unknown>);
+                        break;
```
heimwege, line 230: The two new generator tools `generate_fiori_app_odata` and `generate_fiori_app_cap` are dispatched with args as `Record<string, unknown>` rather than a typed input interface

The tool definitions in `index.ts` use `generatorConfigODataJson` / `generatorConfigCAPJson` as `inputSchema`, which means those schemas exist. There should be corresponding Zod schemas + inferred types for the generator args (similar to `DownloadODataServiceMetadataInput`) so the dispatch is type-safe rather than a cast.

Also allowing `Record<string, unknown>` and passing it directly to the inner executor without any Zod validation might lead to unwanted behavoir.
Adding `generatorConfigODataSchema.parse(args)` / `generatorConfigCAPSchema.parse(args)` at the tool boundary would give early, structured validation errors.

packages/fiori-mcp-server/src/tools/index.ts
```diff
@@ -52,18 +57,91 @@ export const tools = [
...
+        annotations: {
+            title: 'List SAP Systems',
+            readOnlyHint: true,
+            idempotentHint: true,
+            openWorldHint: false
+        },
+        inputSchema: {
+            type: 'object',
+            properties: {},
+            required: []
```
heimwege: The `required: []` is a no-op and slightly noisy. It can be removed or the schema can use `convertToSchema` with an empty Zod object like the other tools do.

packages/fiori-mcp-server/src/tools/download-odata-service-metadata.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Downloads OData service metadata from a SAP system and saves it as metadata.xml.
+ *
+ * @param params - Input parameters containing sapSystemQuery, servicePath and appPath.
+ * @returns A promise resolving to the execution output with host, servicePath, client and metadataFilePath.
+ */
+export async function downloadODataServiceMetadata(
+    params: DownloadODataServiceMetadataInput
+): Promise<ExecuteFunctionalityOutput> {
+    return executeFetchServiceMetadata({
+        functionalityId: 'fetch-service-metadata',
```
heimwege: string literal exist as constants in `src/constant.ts` and should be used to avoid typo risk.

packages/fiori-mcp-server/src/tools/generate-fiori-app-odata.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Generates a new SAP Fiori UI application for OData (non-CAP) projects.
+ *
+ * @param args - Input parameters matching the generatorConfigOData schema.
+ * @returns A promise resolving to the generation execution output.
+ */
+export async function generateFioriAppOData(args: Record<string, unknown>): Promise<ExecuteFunctionalityOutput> {
+    return executeOData({
+        functionalityId: 'generate-fiori-ui-application',
```
heimwege: string literal exist as constants in `src/constant.ts` and should be used to avoid typo risk.

heimwege (commented):
see comments

cfg74:
Like the proposal of promotes. How do we address the following:
1. Convey to the LLM that list functionality is contextual, as in needs a project to figure out what is possible in particular.
2. Convey to the LLM the difference of list functionality content and doc search content, and when to start with what

devinea:
Consider changes from https://github.com/SAP/open-ux-tools/issues/4652 need to be re-applied here before merging.

heimwege:
Just noticed we need to adjust the available tools list in [README.md](https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-mcp-server/README.md#available-tools)

packages/fiori-mcp-server/src/types/input.ts
devinea: fixed

packages/fiori-mcp-server/src/server.ts
devinea: fixed

packages/fiori-mcp-server/src/tools/download-odata-service-metadata.ts
devinea: fixed

packages/fiori-mcp-server/src/tools/generate-fiori-app-odata.ts
devinea: fixed

devinea:
> Just noticed we need to adjust the available tools list in [README.md](https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-mcp-server/README.md#available-tools)

fixed in https://github.com/SAP/open-ux-tools/pull/4649/commits/3ede6aee5d417db89b47f1c2b1ac54621ccd18e9

packages/fiori-mcp-server/src/tools/list-sap-systems.ts
```diff
@@ -0,0 +1,17 @@
+import { getSapSystems } from './functionalities/fetch-service-metadata/service-metadata.js';
+
+/**
+ * Lists all SAP systems stored in the user's environment.
+ *
+ * @returns A promise resolving to an array of SAP system objects.
+ */
+export async function listSapSystems(): Promise<object> {
```
IainSAP: Should the scope of this be widened to also return destinations. The call to either list destinations or getSapSystems can be determined by the environment. Also the return  is not properly typed, is that intentional?

See here for APIs to use: 
https://github.com/SAP/open-ux-tools/blob/main/packages/odata-service-inquirer/src/prompts/datasources/sap-system/system-selection/prompt-helpers.ts#L227-L288

packages/fiori-mcp-server/src/server.ts
```diff
@@ -184,6 +193,18 @@ export class FioriFunctionalityServer {
                     case 'list_fiori_apps':
                         result = await listFioriApps(args as ListFioriAppsInput);
                         break;
+                    case 'list_sap_systems':
```
IainSAP, line 219: Should this cover both destinations and stored connection configurations

packages/fiori-mcp-server/src/tools/list-sap-systems.ts
nikmace: If you want to include destinations:

```ts
import { listDestinations, isAppStudio } from '@sap-ux/btp-utils';

if (isAppStudio()) {
   const destinations = await listDestinations({
         stripS4HCApiHosts: true
   });
   endpoints = Object.values(destinations).filter((dest) => dest.WebIDEUsage?.includes('dev_abap'));
} else {
  const systemStore = await getService<BackendSystem, BackendSystemKey>({
        entityName: 'system'
  });
  const backendSystems = await systemStore?.getAll();
  endpoints = backendSystems.filter((system) => system.name !== undefined).map(transformBackendSystem);
}
```

nikmace: We transform them in a different way than you do, but I guess for LLM it won't be a problem

packages/fiori-mcp-server/package.json
```diff
@@ -29,7 +29,7 @@
         "watch": "tsc --watch",
         "bundle": "npm-run-all build tgz:package tgz:rename",
         "tgz:package": "pnpm pack",
-        "tgz:rename": "node -e \"const fs=require('fs'), p=require('./package.json'), d=new Date().toISOString().split('T')[0], oldName=p.name.startsWith('@') ? `${p.name.replace('@', '').replace('/', '-')}-${p.version}.tgz` : `${p.name}-${p.version}.tgz`; fs.renameSync(oldName, `${oldName.replace('.tgz','')}-${d}.tgz`)\"",
+        "tgz:rename": "node scripts/rename-tgz.mjs",
```
IainSAP: @heimwege Can we remove this rename script now : https://github.com/SAP/open-ux-tools/pull/3541/changes#r2293871555

packages/fiori-mcp-server/src/tools/list-sap-systems.ts
IainSAP: Thanks @nikmace this will be added with a different TBI (internal ref: 38248)

packages/fiori-mcp-server/package.json
IainSAP: Agreed to remove

packages/fiori-mcp-server/src/tools/generate-fiori-app-odata.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @param args - Input parameters matching the generatorConfigOData schema.
+ * @returns A promise resolving to the generation execution output.
+ */
+export async function generateFioriAppOData(args: Record<string, unknown>): Promise<ExecuteFunctionalityOutput> {
+    const validated = generatorConfigOData.parse(args);
+    return executeOData({
+        functionalityId: GENERATE_FIORI_UI_APPLICATION_ID,
+        parameters: validated,
+        appPath: (validated.project?.targetFolder as string) ?? ''
```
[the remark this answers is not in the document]
devinea: Fixed — removed the `as string` cast so `?? ''` correctly fires when `targetFolder` is absent.

packages/fiori-mcp-server/src/tools/generate-fiori-app-cap.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @param args - Input parameters matching the generatorConfigCAP schema.
+ * @returns A promise resolving to the generation execution output.
+ */
+export async function generateFioriAppCap(args: Record<string, unknown>): Promise<ExecuteFunctionalityOutput> {
+    const validated = generatorConfigCAP.parse(args);
+    return command({
+        functionalityId: GENERATE_FIORI_UI_APPLICATION_CAP_ID,
+        parameters: validated,
+        appPath: (validated.project?.targetFolder as string) ?? ''
```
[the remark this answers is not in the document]
devinea: Fixed — removed the `as string` cast so `?? ''` correctly fires when `targetFolder` is absent.

packages/fiori-mcp-server/src/tools/functionalities/fetch-service-metadata/service-metadata.ts
```diff
@@ -11,7 +11,7 @@ import { parse as parseEdmx } from '@sap-ux/edmx-parser';
  *
  * @returns A promise that resolves to an array of BackendSystem objects.
  */
-async function getSapSystems(): Promise<BackendSystem[]> {
+export async function getSapSystems(): Promise<BackendSystem[]> {
```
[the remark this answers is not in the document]
devinea: Fixed — `getSapSystems` now accepts an `includeSensitiveData` parameter (defaults to `false`). `listSapSystems` calls it without the flag (no credentials fetched), while `findSapSystem` passes `true` since it needs credentials for authentication.

packages/fiori-mcp-server/src/tools/index.ts
devinea: Confirmed — the `required: []` was already not present in the current code; this was resolved in a prior commit.

packages/fiori-mcp-server/src/server.ts
devinea: Destinations support will be added in a separate TBI (internal ref: 38248) as discussed in the related thread on `list-sap-systems.ts`.

IainSAP:
This PR needs to be merged but there are structural changes here that need manual resolution: https://github.com/SAP/open-ux-tools/pull/4852/changes

packages/fiori-mcp-server/src/tools/services/sap-system.ts
```diff
@@ -78,36 +85,87 @@ function matchSystemByUrl(systems: BackendSystem[], url: string): BackendSystem[
...
+ * @returns The matching destination, or undefined if not found.
+ */
+function findDestination(destinations: Destination[], query: string): Destination | undefined {
+    const queryLower = query.toLocaleLowerCase();
+
+    let match = destinations.find((d) => d.Name === query);
+    if (!match) {
+        match = destinations.find((d) => d.Name.toLocaleLowerCase() === queryLower);
+    }
+    if (!match) {
```
IainSAP: seems one of these may be redundant (startsWith/includes)....copied from existing https://github.com/SAP/open-ux-tools/pull/4649/changes/13f4d7556dfb5d6394de38ad93a141bf320072fc#diff-de08792358269546491993fe6e289abf5f09cef89641a59230a7943f6a515f21R163-R168

IainSAP (dismissed):
Thanks @devinea 

- Destination support added and tested manually
- Tests added to cover

packages/fiori-mcp-server/src/tools/services/sap-system.ts
```diff
@@ -0,0 +1,280 @@
...
+}
+
+/**
+ * Finds a system by its name, or url and client (partial match).
+ * In BAS returns a Destination, in VSCode returns a BackendSystem.
+ *
+ * @param query - The name, host or URL to match.
+ * @returns The matching system if found.
+ */
+export async function findSystem(query: string): Promise<BackendSystem | Destination | undefined> {
```
heimwege: `findSystem()` returns `undefined` on BAS (via `findDestination`), but on VSCode it delegates to `findSapSystem()` which throws on not-found. The callers in download-odata-service-metadata-impl.ts check for `undefined` to detect "not found", but on VSCode a multi-match or no-match will throw instead, bypassing the `undefined` check and landing in the catch block below.
This is functional but the behavior diverges between BAS and VSCode. It would be cleaner if `findSapSystem` also returned `undefined` / threw consistently via `findSystem`.

packages/fiori-mcp-server/src/tools/index.ts
```diff
@@ -52,18 +57,94 @@ export const tools = [
...
+            idempotentHint: true,
+            openWorldHint: false
+        },
+        inputSchema: {
+            type: 'object',
+            properties: {}
+        }
+    },
+    {
+        name: 'fetch-service-metadata',
```
heimwege: general: The new tools `fetch-service-metadata`, `generate-fiori-ui-application`, `generate-fiori-ui-application-cap` use kebab-case while `list_sap_systems`, `list_fiori_apps`, `list_functionality` use snake_case.
MCP conventions favor snake_case.

packages/fiori-mcp-server/src/tools/services/sap-system.ts
```diff
@@ -0,0 +1,280 @@
...
+const SAP_TOOLS_BASE_DIRECTORY = process.env.SAP_TOOLS_DIR || getSapToolsDirectory();
+
+/**
+ * Fetches SAP backend systems.
+ *
+ * @param includeSensitiveData - Whether to include sensitive fields (credentials). Defaults to false.
+ * @returns A promise that resolves to an array of BackendSystem objects.
+ */
+export async function getSapSystems(includeSensitiveData = false): Promise<BackendSystem[]> {
+    const logger = new ToolsLogger({ logPrefix: 'fiori-mcp-server' });
```
heimwege: this shadows the module-level logger imported from `../../utils/index.js`. The module-level logger is used throughout the rest of the file, but `getSapSystems` creates its own.
This is inconsistent and wasteful - the local logger is used only for the store call but is constructed on every `getSapSystems()` invocation.

packages/fiori-mcp-server/test/unit/tools/download-odata-service-metadata.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+const { downloadODataServiceMetadata } = await import('../../../src/tools/download-odata-service-metadata.js');
+
+describe('downloadODataServiceMetadata', () => {
+    const mockAppPath = '/test/project';
+    const mockServicePath = '/sap/opu/odata4/test/service';
+    const mockResult = {
+        functionalityId: 'fetch-service-metadata',
+        status: 'Success',
+        message: 'Fetched systems successfully.',
```
heimwege: Is this a copy/paste error?

```suggestion
        message: 'Service metadata downloaded successfully.',
```

packages/fiori-mcp-server/src/tools/services/sap-system.ts
```diff
@@ -0,0 +1,280 @@
...
+    let parseError: unknown;
+    try {
+        parsedMetadata = parseEdmx(metadata);
+    } catch (error) {
+        parseError = error;
+        logger.debug(error);
+    }
+    if (!parsedMetadata) {
+        const detail = parseError instanceof Error ? ` Reason: ${parseError.message}` : '';
+        throw new Error(`Failed to parse service metadata. The service may not be a valid OData V4 service.${detail}`);
```
heimwege:
```suggestion
        throw new Error(`Failed to parse service metadata. The service may not be a valid OData V4 service. ${detail}`);
```

heimwege (commented):
see comments

packages/fiori-mcp-server/src/tools/services/sap-system.ts
devinea: addressed in https://github.com/SAP/open-ux-tools/pull/4649/commits/a2d33e5b5680ea16904ceb9bdeae885378a62e19

packages/fiori-mcp-server/test/unit/tools/download-odata-service-metadata.test.ts
devinea: addressed in https://github.com/SAP/open-ux-tools/pull/4649/commits/a2d33e5b5680ea16904ceb9bdeae885378a62e19

packages/fiori-mcp-server/src/tools/services/sap-system.ts
devinea: addressed in https://github.com/SAP/open-ux-tools/pull/4649/commits/a2d33e5b5680ea16904ceb9bdeae885378a62e19

packages/fiori-mcp-server/src/tools/index.ts
devinea: addressed in https://github.com/SAP/open-ux-tools/pull/4649/commits/a2d33e5b5680ea16904ceb9bdeae885378a62e19

packages/fiori-mcp-server/src/tools/services/sap-system.ts
devinea: addressed in https://github.com/SAP/open-ux-tools/pull/4649/commits/a2d33e5b5680ea16904ceb9bdeae885378a62e19

heimwege: I don't see this being adjusted 🤷🏻

packages/fiori-mcp-server/src/tools/download-odata-service-metadata-impl.ts
```diff
@@ -0,0 +1,84 @@
...
+        const dest = foundSystem as Destination;
+        const backend = foundSystem as BackendSystem;
+        const host = isAS ? dest.Host : backend.url;
+        const client = isAS ? dest['sap-client'] : backend.client;
+        const destination = isAS ? dest.Name : undefined;
+
+        return {
+            functionalityId: DOWNLOAD_ODATA_SERVICE_METADATA_ID,
+            status: 'Success',
+            message: 'Fetched systems successfully.',
```
heimwege: Just fixing the test mock does not help with the productive code. I suggest we fix it here as well 🤣

```suggestion
            message: 'Service metadata downloaded successfully.',
```

heimwege (approved):
- review comments addressed
- changeset ok
- coverage is very good
- did NOT test manually

IainSAP (approved):
Thanks @devinea 

- Re-approving

nikmace (approved):
Changes look good
Coverage is good
Changeset is OK
Did not test manually

PR #4659, opened 2026-05-12
fix(sap-systems-ext): remove default selected key for system type

cianmSAP:
- Removes default selected key for system type
- Note: This is reverting a recent commit. No unit tests were added for this change, therefore no tests need to be updated (or added as it's code removal).


korotkovao (approved):
Makes sense due to a recent requirement change.
Looks good to me, thanks @cianmSAP

mikicvi-SAP (approved):
Cheers @cianmSAP 
- revert small, LG
- changeset

PR #4662, opened 2026-05-12
fix(odata-service-writer, odata-service-inquirer): dynamically detect OData 4.01 from service metadata

### LINKED ISSUE #3995: FEATURE - Support OData version 4.01
As a developer using SAP FIori tools,
I want to be able to consume v4.01 odata services,
so that I use the latest features

### Description

CAP and RAP support OData 4.01, UI5 model supports this since UI5 1.142.

Image

### Technical Design
_What modules need to be created or enhanced?_ 
_How would the functionality be exposed and how can it be consumed?_

Acceptance Criteria
Given I generate a Fiori application and choose a version of UI5 that is 1.144 or higher, if the OData service provided has version 4.01, then the odata version in the manifest should also be 4.01

### Notes
- The current manifest schema only permits 2.0 and 4.0, but this is being updated shortly.
- 
### Tasks
- [ ] Unit Tests added
- [ ]  Test Cases defined
- [ ]  UI Automation Tests

longieirl:
## Summary

- Fixes #3995
- Adds `OdataVersion.v401 = '4.01'` to the `OdataVersion` enum in `odata-service-writer`
- Fixes `parseOdataVersion` in `odata-service-inquirer` to correctly detect `Version="4.01"` from service metadata, instead of truncating via `Number.parseInt` (which silently mapped `4.01` → `4`)
- `manifest.ts` unconditionally writes `odataVersion: "4.01"` for `v401` services (mirrors how `v4 → "4.0"` and `v2 → "2.0"`); plain OData V4 services always write `"4.0"`
- Adds `OdataVersion.v401` support to all six `fiori-elements-writer` template types (Worklist, LROP, ALP, OVP, FEOP, FPM) including `supportedODataVersions`, `minimumUi5Version`, and `annotationGenerationSupport`

## Background

Previously, `odataVersion: "4.01"` was written into `manifest.json` for all OData V4 apps when `minUI5Version >= 1.144`, regardless of whether the backend service actually declared OData 4.01. This caused cache buster token generation failures in ABAP appIndex and deployment issues in WorkZone. The fix makes the version driven by what the service metadata actually declares.

## Changes addressing review feedback

- **`entity-helper.ts`** — replaced `Number.parseInt(convertedMetadata.version, 10)` with the shared `parseOdataVersion()` utility, removing the inconsistency where the catalog connection path returned `v401` but the local EDMX file path returned `v4` for the same `Version="4.01"` metadata. Updated `filterAggregateTransformationsOnly` and `getNavigationEntityChoices` to treat `v401` the same as `v4`.
- **`questions.ts`** — extended three `v4`-only conditions (ALP empty-entity error, TreeTable hierarchy qualifier prompt, lrop/worklist annotation questions) to also apply to `v401`.
- **`validators.ts`** — `validateODataVersion` now treats `v401` as satisfying a `v4` requirement (OData 4.01 is backwards-compatible with 4.0); strict equality `v4 !== v401` no longer raises a spurious mismatch error.
- **`utils/index.ts`** — `rawVersion.startsWith('4.01')` → `rawVersion === '4.01'` (exact match).

## Changes addressing post-review propagation gaps

- **`fiori-elements-writer/src/index.ts` + `manifestSettings.ts`** — remapped `v401 → v4` for template path construction; `OdataVersion.v401 = '4.01'` caused paths like `templates/v4.01/` which do not exist, silently producing incomplete output.
- **`entity-helper.ts` `getDefaultTableType`** — expanded `v4`-only guard to include `v401` so ALP/LROP v401 services correctly detect AnalyticalTable and TreeTable types.
- **`alp-questions.ts`** — expanded `tableSelectionMode` prompt guard to include `v401`; previously v401 services fell through to the v2 prompt branch (multiSelect/autoHide/smartVariantManagement).
- **`fiori-app-sub-generator` `getODataVersion`** — fixed to return `OdataVersion.v401` for `Version="4.01"` EDMX; previously the headless generation path silently downgraded v401 to v4, resulting in `"4.0"` written to the manifest.
- **`fiori-app-sub-generator` `FloorplanAttributes`** — added `OdataVersion.v401` to `supportedODataVersion` for all six FE floorplans so `getMinSupportedUI5Version` correctly returns `1.144.0` for v401 instead of falling back to the v4 minimum.
- **`fiori-app-sub-generator` `getRequiredOdataVersion`** — updated to collapse `v401 → v4` when computing the single required version for a floorplan, so FEOP and FPM continue to report `v4` as their version family constraint (which `validateODataVersion` already accepts for v401 services).
- **`odata-service-writer` `enhanceManifestDatasources`** — removed now-unused `minimumUi5Version` parameter (was only needed for the old `semVer.satisfies` conditional that has been replaced by enum-based switching).
- **Changeset** — added `@sap-ux/fiori-generator-shared` (adds `enableVirtualEndpoints` to public `AppConfig` interface) and `@sap-ux/sap-systems-ext-webapp` (removes `systemType` prop from `SystemTypes` component).

## AGENTS.md Non-Conformance

The project guidelines (`AGENTS.md`) recommend avoiding TypeScript enums in favour of union types or const objects. `OdataVersion` is an existing public API enum already exported and consumed across multiple packages — converting it to a union type would be a breaking change outside the scope of this fix. The `v401` value has been added to the existing enum to maintain backward compatibility. Refactoring `OdataVersion` away from an enum is tracked as a separate follow-up.

## Test plan

- [x] `parseOdataVersion` correctly returns `OdataVersion.v401` for a `Version="4.01"` metadata document
- [x] `getEntityChoices` correctly returns `odataVersion === v401` for `Version="4.01"` metadata (local EDMX path now consistent with catalog path)
- [x] `validateODataVersion` accepts `v401` metadata when required version is `v4` (backwards compatible)
- [x] `validateODataVersion` rejects `v401` metadata when required version is `v2`
- [x] `manifest.ts` unconditionally writes `"4.01"` for `v401` services
- [x] `manifest.ts` always writes `"4.0"` for plain `OdataVersion.v4` services
- [x] `getDefaultTableType` returns correct table type (AnalyticalTable/TreeTable) for v401 ALP/LROP services
- [x] `alp-questions.ts` routes v401 services to the v4-style `tableSelectionMode` prompt
- [x] `getODataVersion` returns `OdataVersion.v401` for `Version="4.01"` EDMX in headless generation
- [x] `getMinSupportedUI5Version` returns `1.144.0` for v401 LROP (not the v4 fallback `1.84.0`)
- [x] `getRequiredOdataVersion` returns `OdataVersion.v4` for FEOP and FPM floorplans (v4 family constraint preserved)
- [x] All 89 `odata-service-writer` tests pass
- [x] All 277 `odata-service-inquirer` tests pass
- [x] All 70 `fiori-elements-writer` tests pass
- [x] All 181 `fiori-app-sub-generator` tests pass

### DIFF
--- a/packages/fiori-app-sub-generator/src/types/external.ts
+++ b/packages/fiori-app-sub-generator/src/types/external.ts
@@ -22,27 +22,27 @@ type FloorplanAttributesType = {
 
 export const FloorplanAttributes: FloorplanAttributesType = {
     [FloorplanFE.FE_LROP]: {
-        supportedODataVersion: [OdataVersion.v2, OdataVersion.v4],
+        supportedODataVersion: [OdataVersion.v2, OdataVersion.v4, OdataVersion.v401],
         templateType: FETemplateType.ListReportObjectPage
     },
     [FloorplanFE.FE_ALP]: {
-        supportedODataVersion: [OdataVersion.v2, OdataVersion.v4],
+        supportedODataVersion: [OdataVersion.v2, OdataVersion.v4, OdataVersion.v401],
         templateType: FETemplateType.AnalyticalListPage
     },
     [FloorplanFE.FE_WORKLIST]: {
-        supportedODataVersion: [OdataVersion.v2, OdataVersion.v4],
+        supportedODataVersion: [OdataVersion.v2, OdataVersion.v4, OdataVersion.v401],
         templateType: FETemplateType.Worklist
     },
     [FloorplanFE.FE_FEOP]: {
-        supportedODataVersion: [OdataVersion.v4],
+        supportedODataVersion: [OdataVersion.v4, OdataVersion.v401],
         templateType: FETemplateType.FormEntryObjectPage
     },
     [FloorplanFE.FE_OVP]: {
-        supportedODataVersion: [OdataVersion.v2, OdataVersion.v4],
+        supportedODataVersion: [OdataVersion.v2, OdataVersion.v4, OdataVersion.v401],
         templateType: FETemplateType.OverviewPage
     },
     [FloorplanFE.FE_FPM]: {
-        supportedODataVersion: [OdataVersion.v4],
+        supportedODataVersion: [OdataVersion.v4, OdataVersion.v401],
         templateType: FETemplateType.FlexibleProgrammingModel
     },
     [FloorplanFF.FF_SIMPLE]: {
--- a/packages/fiori-app-sub-generator/src/utils/common.ts
+++ b/packages/fiori-app-sub-generator/src/utils/common.ts
@@ -103,7 +103,7 @@ export function getMinSupportedUI5Version(
         const templateType = FloorplanAttributes[floorplan].templateType as FETemplateType;
         minUI5Version = TemplateTypeAttributes[templateType].minimumUi5Version[version];
     }
-    return minUI5Version ?? (version !== OdataVersion.v2 ? minSupportedUi5VersionV4 : minSupportedUi5Version);
+    return minUI5Version ?? (version === OdataVersion.v2 ? minSupportedUi5Version : minSupportedUi5VersionV4);
 }
 
 /**
--- a/packages/fiori-app-sub-generator/src/utils/common.ts
+++ b/packages/fiori-app-sub-generator/src/utils/common.ts
@@ -103,7 +103,7 @@ export function getMinSupportedUI5Version(
         const templateType = FloorplanAttributes[floorplan].templateType as FETemplateType;
         minUI5Version = TemplateTypeAttributes[templateType].minimumUi5Version[version];
     }
-    return minUI5Version ?? (version === OdataVersion.v4 ? minSupportedUi5VersionV4 : minSupportedUi5Version);
+    return minUI5Version ?? (version !== OdataVersion.v2 ? minSupportedUi5VersionV4 : minSupportedUi5Version);
 }
 
 /**
--- a/packages/fiori-app-sub-generator/src/utils/common.ts
+++ b/packages/fiori-app-sub-generator/src/utils/common.ts
@@ -39,6 +39,9 @@ import { getBackendSystemType } from '@sap-ux/store';
 export function getODataVersion(edmx: string): OdataVersion {
     try {
         const convertedMetadata = convert(parse(edmx));
+        if (convertedMetadata.version === '4.01') {
+            return OdataVersion.v401;
+        }
         return convertedMetadata.version.startsWith('4') ? OdataVersion.v4 : OdataVersion.v2;
     } catch (error) {
         throw Error(t('error.appConfigUnparseableEdmx'));
@@ -77,7 +80,9 @@ export function buildSapClientParam(sapClient: string): string {
  */
 export function getRequiredOdataVersion(floorplan: Floorplan): OdataVersion | undefined {
     const supportedVers = FloorplanAttributes[floorplan].supportedODataVersion;
-    return supportedVers.length === 1 ? supportedVers[0] : undefined;
+    // Collapse v401 to v4 — they are the same family for version requirement purposes
+    const baseVers = [...new Set(supportedVers.map((v) => (v === OdataVersion.v401 ? OdataVersion.v4 : v)))];
+    return baseVers.length === 1 ? baseVers[0] : undefined;
 }
 
 /**
--- a/packages/fiori-elements-writer/src/data/templateAttributes.ts
+++ b/packages/fiori-elements-writer/src/data/templateAttributes.ts
@@ -32,7 +32,8 @@ const commonUi5Libs: FrameworkLibs = {
         'sap.ui.generic.app',
         'sap.suite.ui.generic.template'
     ],
-    [OdataVersion.v4]: ['sap.m', 'sap.fe.templates']
+    [OdataVersion.v4]: ['sap.m', 'sap.fe.templates'],
+    [OdataVersion.v401]: ['sap.m', 'sap.fe.templates']
 };
 
 type TemplateLibsEntry = {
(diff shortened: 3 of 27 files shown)


packages/odata-service-inquirer/src/utils/index.ts
```diff
@@ -41,14 +41,17 @@ export function parseOdataVersion(metadata: string): {
 } {
     try {
         const convertedMetadata = convert(parse(metadata));
-        const parsedOdataVersion = Number.parseInt(convertedMetadata?.version, 10);
+        const rawVersion = convertedMetadata?.version;
 
-        if (Number.isNaN(parsedOdataVersion)) {
+        if (!rawVersion || Number.isNaN(Number.parseInt(rawVersion, 10))) {
```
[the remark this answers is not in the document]
longieirl: The `^[14]` regex is intentional. The `version` field returned by `@sap-ux/annotation-converter` is the **EDMX schema version**, not the OData protocol version. EDMX schema `1.x` is the format used by OData v2 services (version strings start with `'1'`), while EDMX `4.0`/`4.01` is used by OData v4/v4.01 services. Changing to `^[24]` would reject every OData v2 service since their EDMX version starts with `'1'`, not `'2'`. The `[14]` character class is correct for this input domain.

packages/odata-service-writer/src/data/manifest.ts
```diff
@@ -62,9 +63,11 @@ function enhanceManifestDatasources(
     if (serviceMetadata) {
         settings['localUri'] = `localService/${serviceName}/metadata.xml`;
     }
-    if (serviceVersion === '4') {
+    if (serviceVersion === OdataVersion.v401) {
         settings['odataVersion'] = minimumUi5Version && semVer.satisfies(minimumUi5Version, '>=1.144') ? '4.01' : '4.0';
```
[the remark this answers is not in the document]
longieirl: The `semVer.satisfies` call flagged here was part of `enhanceManifestDatasources` and has been **removed** in this PR — the `minimumUi5Version` parameter was dropped from that function entirely, with `odataVersion` now set purely by enum value. The only remaining `semVer.satisfies` call is in `getModelSettings` (line 146), where the input is `string | undefined` and is guarded by `if (minUI5Version)` before the call. The concern raised applies to code that no longer exists in this PR.

packages/odata-service-writer/test/unit/manifest.test.ts
```diff
@@ -67,6 +67,42 @@ describe('manifest', () => {
...
+                };
+
+                fs.writeJSON('./webapp/manifest.json', testManifest);
+                await updateManifest('./', service, fs);
+                const manifestJson = fs.readJSON('./webapp/manifest.json') as Partial<Manifest>;
+                expect(manifestJson['sap.app']?.dataSources?.['aname'].settings?.['odataVersion']).toEqual(
+                    expectedOdataVersion
+                );
+            }
+        );
```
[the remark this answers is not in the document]
longieirl: The model settings assertions are already present. Lines 104–108 of the `test.each` block explicitly assert `operationMode: 'Server'`, `autoExpandSelect: true`, and `earlyRequests: true` for `OdataVersion.v401` across all four `minUI5Version` variants (including `undefined`). The bot appears to have reviewed an earlier version of the file before these assertions were added.

packages/fiori-elements-writer/src/data/templateAttributes.ts
```diff
@@ -97,6 +99,38 @@ const templateLibs: TemplateLibs = {
             ui5Libs: [...commonUi5Libs[OdataVersion.v4], 'sap.fe.core', 'sap.ushell'],
             manifestLibs: ['sap.m', 'sap.fe.core']
         }
+    },
+    [OdataVersion.v401]: {
+        [TemplateType.ListReportObjectPage]: {
```
devinea: Avoid duplication here and create reusable function.

devinea (commented):
Thanks @longieirl . Looks good. 1 small comment

packages/fiori-app-sub-generator/src/types/external.ts
```diff
@@ -22,27 +22,27 @@ type FloorplanAttributesType = {
 
 export const FloorplanAttributes: FloorplanAttributesType = {
     [FloorplanFE.FE_LROP]: {
-        supportedODataVersion: [OdataVersion.v2, OdataVersion.v4],
+        supportedODataVersion: [OdataVersion.v2, OdataVersion.v4, OdataVersion.v401],
```
IainSAP, line 25: Can we not treat all version 4 the same? There are no difference here in support.

IainSAP (changes requested):
Can we please keep the base versions everywhere in the generator code...required version handling is only for 2/4 and there should be no need to discriminate between 4 and 401 in generator related modules only when calling thw writer (pass in the actual service metadata version at that point)

packages/fiori-app-sub-generator/src/utils/common.ts
```diff
@@ -77,7 +80,9 @@ export function buildSapClientParam(sapClient: string): string {
  */
 export function getRequiredOdataVersion(floorplan: Floorplan): OdataVersion | undefined {
     const supportedVers = FloorplanAttributes[floorplan].supportedODataVersion;
-    return supportedVers.length === 1 ? supportedVers[0] : undefined;
+    // Collapse v401 to v4 — they are the same family for version requirement purposes
```
IainSAP, line 83: 👍

packages/fiori-app-sub-generator/src/utils/common.ts
```diff
@@ -103,7 +108,7 @@ export function getMinSupportedUI5Version(
         const templateType = FloorplanAttributes[floorplan].templateType as FETemplateType;
         minUI5Version = TemplateTypeAttributes[templateType].minimumUi5Version[version];
     }
-    return minUI5Version ?? (version === OdataVersion.v4 ? minSupportedUi5VersionV4 : minSupportedUi5Version);
+    return minUI5Version ?? (version === OdataVersion.v2 ? minSupportedUi5Version : minSupportedUi5VersionV4);
```
IainSAP, line 111: Not sure this is correct?

longieirl:
Closing for now, will try to isoloate only the changes to the writers.

PR #4667, opened 2026-05-13
fix: When we get the service uri from the ui5.yaml > adp > target object we use the context to determine whether to use the destination in the target or the url.

avasilev-sap:
#4664

In the add new model generator the additional message for the service URI prompt must use the context, BAS or VSCode, when displaying the service URI. In VSCode we use the ui5.yaml > adp > target > url in bas instead of url we use the destination > Host.

**VSCode after:**
Screenshot 2026-05-13 at 11 48 35

**BAS before:**

Screenshot 2026-05-13 at 15 13 00

**BAS after:**

Screenshot 2026-05-13 at 15 15 19


mmilko01 (approved):
Changeset OK
Excellent test coverage
Changes look good
Did not test manually

IvoSG (approved):
Changeset OK
Excellent test coverage
Changes look good
Did not test manually

PR #4670, opened 2026-05-13
feat: Add ADP specific tools

nikmace:
# feat(fiori-mcp-server): Add ADP-Specific MCP Tools

### New Features

✨ Introduces three new top-level MCP tools for SAP Fiori **Adaptation Projects (ADP)** in the `fiori-mcp-server` package:

- **`generate_adaptation_project`** — Scaffolds a new SAP Fiori adaptation project by invoking the `@sap-ux/adp` Yeoman generator with configurable options (system, application, namespace, title, credentials, etc.).
- **`open_adaptation_editor`** — Spawns the adaptation editor server (`npx fiori run /test/adaptation-editor.html`) in the background, extracts the editor URL and process ID from its output, and returns instructions on how to stop the process.
- **`adp_controller_extension`** — Processes AI-generated controller extension code and XML fragments for adaptation projects using a progressive disclosure pattern: the first call (without `aiResponse`) returns detailed generation rules and existing project context; a subsequent call with `aiResponse` writes the generated files to disk.

### Changes

* `constant.ts`: Added constants for the three new tool IDs (`GENERATE_ADAPTATION_PROJECT_ID`, `ADP_CONTROLLER_EXTENSION_FUNCTIONALITY_ID`, `OPEN_ADAPTATION_EDITOR_ID`).
* `tools/generate-adaptation-project.ts`: New tool that invokes `@sap-ux/adp` via `npx yo` with a JSON configuration. Includes a default project name helper to avoid directory collisions.
* `tools/open-adaptation-editor.ts`: New tool that spawns the adaptation editor process, monitors stdout for the server URL and editor path, detects the listening port cross-platform (Windows/Mac/Linux), and returns kill commands.
* `tools/adp-controller-extension/index.ts`: New tool implementing progressive disclosure — returns a rich knowledge base and project context on first call, then extracts and writes code files from an AI-generated markdown response on subsequent calls. Skips `.change` files and validates file paths are within the project boundary.
* `tools/adp-controller-extension/utils.ts`: Utility functions for parsing AI markdown responses (file extraction), detecting adaptation projects, reading `manifest.appdescr_variant`, scanning existing project files (`webapp/changes`), and path helpers.
* `tools/index.ts`: Exports the three new tools and registers them in the tools list with full descriptions and `inputSchema` definitions. Also updates the `list_sap_systems` description to reference `generate_adaptation_project`.
* `types/input.ts`: Added Zod schemas (`GenerateAdaptationProjectInputSchema`, `OpenAdaptationEditorInputSchema`, `AdpControllerExtensionInputSchema`) and their inferred TypeScript types.
* `types/index.ts`: Exports the three new input types.
* `server.ts`: Wires the new tools into the MCP server's `CallTool` handler and updates the unknown-tool error message.
* `package.json` / `tsconfig.json`: Added `@sap-ux/adp-tooling` as a workspace dependency and TypeScript project reference.
* `test/unit/server.test.ts`: Updated tool list assertions and unknown-tool error message snapshots to include the three new tools.
* `pnpm-lock.yaml`: Updated lock file with new `@sap-ux/adp-tooling` dependency and minor transitive dependency version resolutions.
* `.changeset/fiori-mcp-server-adp-tools.md`: Changeset entry marking this as a `minor` release.

<details>

**Version:** `1.20.47`

- Correlation ID: `b309b5cb-fd08-42c1-bb56-06353159ac6a`
</details>


packages/fiori-mcp-server/skills/sap-fiori-adp-controller-extension-flow/SKILL.md
```diff
@@ -0,0 +1,362 @@
+---
```
heimwege, line 1: General:

- please prefix skills with `sap-fiori-*` because they we prefer to have our skills grouped when installed locally or globally (all installed skills will be in a flat list in `.agents/skills`)
- please add author (`sap-fiori-tools`) and version (`0.0.1`?) to the frontmatter of each skill
- Please make sure the skill is self-contained. This skill seems to have an installed fiori-mcp as prerequisite but does not mention it (including instructions how to install or where to find instructions how to install). This will cause issues if the skill has been found e.g. on [skills.sh](https://skills.sh) and was installed using cli command `npx skills add <skill url>` w/o having the fiori-mcp installed.

packages/fiori-mcp-server/src/tools/generate-adaptation-project.ts
```diff
@@ -0,0 +1,116 @@
...
+        if (stderr) {
+            logger.error(stderr);
+        }
+
+        const projectPath = join(finalTargetFolder, jsonInput.projectName);
+        return {
+            functionalityId: GENERATE_ADAPTATION_PROJECT_ID,
+            status: 'Success',
+            message: `Adaptation project generated successfully at ${projectPath}.`,
+            parameters: params,
```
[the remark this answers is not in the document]
nikmace: Fixed — password is now stripped via safeParams() before echoing params in all three return paths (validation error, success, catch error).

packages/fiori-mcp-server/src/tools/open-adaptation-editor.ts
```diff
@@ -0,0 +1,294 @@
...
+                crlfDelay: Infinity
+            });
+
+            stderrRl.on('line', (line: string) => {
+                logger.debug(`Editor stderr: ${line}`);
+            });
+        }
+
+        childProcess.on('error', (error) => {
+            logger.error(`Editor process error: ${error.message}`);
```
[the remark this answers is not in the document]
nikmace: Fixed — spawn errors are now stored in spawnError and checked immediately after the polling loop resolves, returning a specific "Failed to spawn editor process: ..." message.

packages/fiori-mcp-server/src/tools/generate-adaptation-project.ts
```diff
@@ -0,0 +1,116 @@
...
+        };
+    }
+}
+
+function getDefaultProjectName(basePath: string, dirName: string = 'app.variant'): string {
+    let newDir = dirName;
+    let index = 1;
+
+    while (existsSync(join(basePath, newDir))) {
+        index++;
```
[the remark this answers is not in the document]
nikmace: Fixed — index now initialises at 0 so the sequence is app.variant → app.variant1 → app.variant2.

packages/fiori-mcp-server/src/tools/open-adaptation-editor.ts
```diff
@@ -0,0 +1,294 @@
...
+
+        await new Promise<void>((resolve) => {
+            const checkInterval = setInterval(() => {
+                if (resolved) {
+                    clearInterval(checkInterval);
+                    clearTimeout(timeoutId);
+                    resolve();
+                }
+            }, 100);
+        });
```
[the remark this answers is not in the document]
nikmace: Fixed — all three resolution paths (stdout parse, error event, setTimeout) now set resolved = true, so the setInterval is guaranteed to clear. The error handler previously only set resolved but now also stores the error in spawnError, meaning a spawn failure can no longer leave resolved permanently false. The timeout path was already correct; the fragility was the unhandled error case, which is now closed.

packages/fiori-mcp-server/src/tools/generate-adaptation-project.ts
```diff
@@ -0,0 +1,116 @@
...
+            jsonInput.username = username;
+        }
+        if (password) {
+            jsonInput.password = password;
+        }
+
+        await FSpromises.mkdir(finalTargetFolder, { recursive: true });
+
+        const jsonString = JSON.stringify(jsonInput);
+        const command = `npx -y yo@4 @sap-ux/adp '${jsonString}' --force`;
```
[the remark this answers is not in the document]
nikmace: This MCP server runs locally on the user's machine — all input originates from the user's own AI client session, not from an external caller. There is no remote attack surface here. The shell injection risk is therefore confined to the user's own environment, which is an acceptable tradeoff for this use case. No fix planned.

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/browser/playwright-bridge.ts
```diff
@@ -0,0 +1,288 @@
...
+    })();
+
+    try {
+        browser = await browserStartPromise;
+        browser.on('disconnected', () => {
+            browser = undefined;
+            browserStartPromise = undefined;
+            connectionRegistry.clear();
+        });
+        return browser;
```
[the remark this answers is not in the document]
nikmace: This is essentially a non-issue in practice — the disconnected handler is the cleanup mechanism and it works correctly.

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/browser/playwright-bridge.ts
```diff
@@ -0,0 +1,288 @@
...
+        });
+
+        page.on('close', () => {
+            connectionRegistry.delete(site);
+        });
+
+        await page.goto(site, { waitUntil: 'networkidle', timeout: 60000 });
+
+        rpc = await createPageRPC(page);
+        connectionRegistry.set(site, rpc);
```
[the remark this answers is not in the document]
nikmace: Fixed — context is now passed into createPageRPC and closed in rpc.close(), so both the page and its context are torn down together on disconnectSite, stopBrowser, and the page close event.

packages/fiori-mcp-server/skills/sap-fiori-adp-controller-extension-flow/SKILL.md
nikmace: Changed in [69733d6](https://github.com/SAP/open-ux-tools/pull/4670/commits/69733d6185176a68b1c3f573cc784445f0fbf407)

nikmace: Fixed for this skill. Note that `sap-fiori-add-visual-filter`, `sap-fiori-analytical-chart`, and `sap-fiori-opa5-test-development` have the same gap — they reference fiori-mcp tools without installation instructions.

packages/fiori-mcp-server/src/tools/adp-controller-extension/ai-response/parser.ts
```diff
@@ -0,0 +1,62 @@
...
+            continue;
+        }
+
+        if (FENCE_OPEN.test(line) && !inCodeBlock) {
+            inCodeBlock = true;
+            currentCode = '';
+            continue;
+        }
+
+        if (line.startsWith('```') && inCodeBlock) {
```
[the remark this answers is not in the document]
nikmace: Fixed — introduced FENCE_CLOSE = /^\```\s*$/ so only a bare \``` closes a block. A language-tagged line like  \```javascript inside a code block no longer triggers early closure. Also moved **Path:** matching into the !inCodeBlock branch so path markers inside code content are ignored.

packages/fiori-mcp-server/src/tools/adp-controller-extension/ai-response/parser.ts
```diff
@@ -0,0 +1,62 @@
...
+            currentCode = '';
+            continue;
+        }
+
+        if (inCodeBlock) {
+            currentCode += line + '\n';
+        }
+    }
+
+    return codeBlocks;
```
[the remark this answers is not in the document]
nikmace: Fixed — an unclosed code block at EOF now emits logger.warn with the affected path instead of silently discarding the content.

packages/fiori-mcp-server/skills/sap-fiori-adp-controller-extension-flow/SKILL.md
heimwege: @nikmace the opa5 skill was a false positive. It just mentions the claude plugin (that is named after the MCP), not the usage of the mcp itself.

@vinayhospete what do you think about analytical chart and add visual filter? They just mention "Consult fiori mcp server if available". This will work if installed as plugin but it will not in case the skill was installed from skills.sh

.changeset/fiori-mcp-server-adp-tools.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-ux/fiori-mcp-server": minor
+---
+
+FEAT: add ADP top-level tools (generate_adaptation_project, open_adaptation_editor, adp_controller_extension)
```
heimwege, line 5: The new skills are not mentioned

packages/fiori-mcp-server/src/tools/generate-adaptation-project.ts
```diff
@@ -0,0 +1,121 @@
...
```
heimwege, line 79: Securtiy issue: `jsonString` contains user-supplied values (system, application, namespace, etc.). A value containing `'` breaks out of the shell quoting.
Fix: use spawn with an args array, or write the config to a temp file and pass the path.

packages/fiori-mcp-server/src/server.ts
```diff
@@ -107,6 +120,7 @@ export class FioriFunctionalityServer {
     private setupErrorHandling(): void {
         this.server.onerror = (error): void => logger.error(`[MCP Error] ${error}`);
         process.on('SIGINT', async () => {
+            await stopBrowser();
```
heimwege, line 123: On SIGINT, server.ts calls `stopBrowser()` (from `playwright-bridge.ts`), which closes pages and the browser and clears `connectionRegistry`. However, the sessions Map in index.ts is a separate module-level variable that `stopBrowser` never touches. After a SIGINT shutdown and restart (e.g. if the process manager restarts it), sessions from before the shutdown would still be in memory for the lifetime of the module - though in practice a process restart resets module state.
The real risk is: if `stopBrowser` is called while sessions exist (e.g. someone calls it directly), those session IDs remain in sessions as dangling entries that will never match a running browser page. A `sessions.clear()` call should be added in the stop-all path.

packages/fiori-mcp-server/src/tools/open-adaptation-editor.ts
```diff
@@ -0,0 +1,308 @@
...
+        const timeoutId = setTimeout(() => {
+            if (!state.resolved) {
+                state.resolved = true;
+                logger.warn('Timeout waiting for editor URL');
+            }
+        }, TIMEOUT_MS);
+
+        await new Promise<void>((resolve) => {
+            const checkInterval = setInterval(() => {
+                if (state.resolved) {
```
heimwege, line 226: AI complains about the 100ms latency:

_state.resolved is set inside rl.on('line', ...), rl.on('close', ...), and childProcess.on('error', ...), but the Promise is resolved only via a setInterval polling at 100ms. The events do not call resolve() directly. This means there is up to 100ms latency on every success path, and the design is unnecessarily indirect. The resolve callback should be called directly from within those event handlers._

Is there a specific reason for this?

packages/fiori-mcp-server/src/tools/open-adaptation-editor.ts
```diff
@@ -0,0 +1,308 @@
...
+                message: 'Failed to get process ID from spawned editor process',
+                parameters: params,
+                appPath,
+                changes: [],
+                timestamp: new Date().toISOString()
+            };
+        }
+
+        childProcess.unref();
+        await new Promise((resolve) => setTimeout(resolve, 1000));
```
heimwege, line 279: Same here, AI is not happy with the 1s delay

_This is always executed after a successful URL detection, before port lookup. There is no retry or condition - it is a hardcoded delay on every success path._

Is there a specific reason?

packages/fiori-mcp-server/src/tools/adp-controller-extension/output.ts
```diff
@@ -0,0 +1,38 @@
+import type { ExecuteFunctionalityOutput } from '../../types/index.js';
+import { ADP_CONTROLLER_EXTENSION_FUNCTIONALITY_ID } from '../../constant.js';
+
+/**
+ * Statuses surfaced by the `adp_controller_extension` tool.
+ */
+export type AdpControllerExtensionStatus = 'success' | 'error' | 'info' | 'skipped';
```
heimwege, line 7: Every other tool in the package uses 'Success'/'Error' (confirmed in `generate-fiori-app-odata.ts`, `generate-fiori-app-cap.ts`, `generate-adaptation-project.ts`, `open-adaptation-editor.ts`). The adp-controller-extension tool uses lowercase `'success'`/`'error'`/`'info'`/`'skipped'`. The output schema uses `zod.string()` (no enum), so neither fails validation - but callers checking status strings must handle both casings. Please stick to the existing convention.

packages/fiori-mcp-server/test/unit/tools/generate-fiori-app-cap-impl.test.ts
```diff
@@ -1,210 +0,0 @@
-import { jest } from '@jest/globals';
```
heimwege, line 1: Why deleting this test?

packages/fiori-mcp-server/test/unit/tools/generate-fiori-app-odata-impl.test.ts
```diff
@@ -1,430 +0,0 @@
-import { jest } from '@jest/globals';
```
heimwege, line 1: Why deleting this test?

packages/fiori-mcp-server/skills/sap-fiori-adp-controller-extension-flow/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+### Step 12 — Cleanup
+
+Call `step: "stop"`. The server closes the session; if it was the last one, the browser shuts down too.
+
+Then kill the editor server:
+
+```
+kill <processId>
+# or by port:
+lsof -ti:<port> | xargs kill
```
heimwege, line 337: `lsof` does not exist on Windows. The `buildKillCommands()` function in `open-adaptation-editor.ts` already generates correct Windows (`taskkill`) vs Unix (`kill`) instructions - the skill docs should mirror that split.

packages/fiori-mcp-server/skills/sap-fiori-adp-project-setup/SKILL.md
```diff
@@ -0,0 +1,114 @@
...
+- **Editor launch fails:** Verify project path is correct and `package.json` exists
+- **ui5.yaml not found:** Skip patch, warn user
+
+## Cleanup
+
+The editor server runs in background. To stop it later:
+```bash
+kill <processId>
+# or by port:
+lsof -ti:<port> | xargs kill
```
heimwege, line 113: `lsof` does not exist on Windows. The `buildKillCommands()` function in `open-adaptation-editor.ts` already generates correct Windows (`taskkill`) vs Unix (`kill`) instructions - the skill docs should mirror that split.

packages/fiori-mcp-server/skills/sap-fiori-adp-controller-extension-flow/SKILL.md
```diff
@@ -0,0 +1,389 @@
+---
+name: sap-fiori-adp-controller-extension-flow
+author: sap-fiori-tools
+version: 0.0.1
```
heimwege, line 4:
```suggestion
metadata:
   author: sap-fiori-tools
   version: 0.0.1
```

packages/fiori-mcp-server/skills/sap-fiori-adp-project-setup/SKILL.md
```diff
@@ -0,0 +1,114 @@
+---
+name: sap-fiori-adp-project-setup
+author: sap-fiori-tools
+version: 0.0.1
```
heimwege, line 4:
```suggestion
metadata:
   author: sap-fiori-tools
   version: 0.0.1
```

packages/fiori-mcp-server/skills/sap-fiori-adp-project-setup/SKILL.md
```diff
@@ -0,0 +1,114 @@
...
+### Step 3: Generate the Adaptation Project
+
+Call `mcp__fiori-mcp__generate_adaptation_project` with:
+- `system`: selected system name
+- `application`: the application ID
+- `appPath`: current working directory or user-specified path
+- `targetFolder`: where to generate (absolute path)
+- `projectName`: user's choice or default
+
+Optional parameters if user provides them: `namespace`, `applicationTitle`, `client`, `username`, `password`.
```
heimwege, line 69: Step 3 passes password in the tool call with no warning. Given the shell injection + process-table exposure found in the code review, the skill should note that credentials are passed to the generator and advise using system-configured credentials where possible.

packages/fiori-mcp-server/skills/sap-fiori-adp-controller-extension-flow/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+
+| User intent | Action |
+|---|---|
+| "add a button / field / column / section / dialog opener" | `CTX_ADDXML` (fragment carries the new control) |
+| "make this button do X" / "open a dialog when …" / "change behavior" | `CTX_EXTEND_CONTROLLER` (handler lives in the controller extension) |
+| "add a button that opens a dialog" | **Both**, in this order: `CTX_ADDXML` for the button, then `CTX_EXTEND_CONTROLLER` for the press handler. Run as two separate iterations of Steps 4–8. |
+
+If `get_actions` returns more actions than these for a control, surface them to the user rather than picking — this skill is only authoritative for `CTX_ADDXML` and `CTX_EXTEND_CONTROLLER`.
+
+## Confidence & HITL Gating
```
heimwege, line 116: At 389 lines it is close to the 500-line guideline. The `Confidence & HITL Gating` section alone is ~130 lines and are detailed enough to warrant a `references/hitl-gating.md` file, with a short summary in the main `SKILL.md` and a pointer to read the reference for the full rubric.

packages/fiori-mcp-server/skills/sap-fiori-adp-controller-extension-flow/SKILL.md
```diff
@@ -0,0 +1,389 @@
+---
+name: sap-fiori-adp-controller-extension-flow
+author: sap-fiori-tools
+version: 0.0.1
+description: Use when making RTA changes to a SAP Fiori adaptation project via the adaptation editor — adding/removing UI elements, changing properties, renaming labels, hiding controls, or extending controllers and fragments.
```
heimwege, line 5: Same passivity problem, and importantly the description doesn't mention the most common trigger phrase a user would type. "I want to add a button to this Fiori app" or "hide a field in the object page" would not obviously match the current description. The word "RTA" appears in the description but most users won't say "RTA changes." Suggestion:

```
Use when the user wants to make UI changes to a SAP Fiori adaptation project via the adaptation editor - adding buttons, fields, columns, or sections, changing labels or properties, hiding controls, or extending controllers with custom logic. Trigger on phrases like "add a button", "hide a field", "add a column", "customize the toolbar", or "extend the controller" when working in an adaptation project context.
```

packages/fiori-mcp-server/skills/sap-fiori-adp-project-setup/SKILL.md
```diff
@@ -0,0 +1,114 @@
+---
+name: sap-fiori-adp-project-setup
+author: sap-fiori-tools
+version: 0.0.1
+description: Use when creating a new SAP Fiori adaptation project, setting up an adaptation editor, or adapting an existing Fiori app on an SAP system
```
heimwege, line 5: The description is functional but passive and narrow: "Use when creating a new SAP Fiori adaptation project, setting up an adaptation editor, or adapting an existing Fiori app on an SAP system." Per the skill-creator guidelines, descriptions should be slightly "pushy" and include synonyms/alternative phrasings users might actually type. Someone saying "I want to modify a Fiori app on our S/4 system" or "adapt an app without touching its source" won't obviously match this. Suggestion:

```
Use when the user wants to create, scaffold, or set up a SAP Fiori adaptation project (ADP), adapt or customize an existing Fiori app without changing its source, launch the adaptation editor, or connect to an SAP system to start making UI changes. Trigger even if the user just says "adapt an app" or "customize a Fiori app."
```

packages/fiori-mcp-server/skills/sap-fiori-adp-controller-extension-flow/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+
+`Overlay` = `{ overlayId, controlId, label, controlType }`.
+`Action` = `{ id, label?, payload?: [{ name, type, required?, description? }] }`.
+For the standard adaptation editor preview iframe, pass `frameId: "preview"` in the `start` payload.
+
+## Actions Reference
+
+These are the canonical RTA actions returned by `get_actions`. **Always pick by `id` from this table — never invent or guess action IDs.** If `get_actions` returns an `id` not listed here, treat it as unknown and ask the user before proceeding.
+
+### `CTX_ADDXML` — Add: Fragment
```
heimwege, line 86: Step 5 says "never invent an action id" and lists only CTX_ADDXML and CTX_EXTEND_CONTROLLER, then says if get_actions returns an unknown id, ask the user. This is correct safety behaviour, but the table is described as "canonical" when it is actually just the two actions this skill knows about. A real RTA-enabled app may return many more (rename, move, remove, etc.). The framing should be "these are the two actions this skill is authoritative about" rather than implying the table is exhaustive.

packages/fiori-mcp-server/skills/sap-fiori-adp-controller-extension-flow/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+```
+**Path:** webapp/changes/coding/MyExtension.js
+```javascript
+// controller extension code
+```
+
+**Path:** webapp/changes/fragments/MyFragment.fragment.xml
+```xml
+<!-- fragment code -->
+```
```
heimwege, line 318: I think this is too much

packages/fiori-mcp-server/skills/sap-fiori-adp-controller-extension-flow/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+
+Create all changes before saving.
+
+### Step 10 — Save
+
+Call `step: "save"`. Returns `{ saved: true }` on success.
+
+### Step 11 — Generate fragment and controller extension content
+
+After saving, use `mcp__fiori-mcp__adp_controller_extension` to fill in the content for any fragments and controller extensions that were created.
```
heimwege, line 292: adp_controller_extension is called with `mcp__fiori-mcp__adp_controller_extension` in Step 11 but earlier the tool contract section refers to `run_rta_workflow_step` without the `mcp__fiori-mcp__ prefix`. The namespace convention is inconsistent within the same skill - pick one form and use it throughout.

packages/fiori-mcp-server/skills/sap-fiori-adp-controller-extension-flow/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+
+If the user requested multiple changes, repeat Steps 2–8 once per change (the UI may have changed, so re-run `get_overlays` between changes). If a single request implies multiple operations (e.g. "add a button that calls a function" = fragment + controller extension), execute each as a separate iteration.
+
+Create all changes before saving.
+
+### Step 10 — Save
+
+Call `step: "save"`. Returns `{ saved: true }` on success.
+
+### Step 11 — Generate fragment and controller extension content
```
heimwege, line 290: No mention of what happens if adp_controller_extension Phase 1 fails (e.g. manifest.appdescr_variant not found). Step 11 goes straight from save to Phase 1 with no error path. There should be a note in the Error Handling table.

heimwege (commented):
see comments

nikmace:
PR closed in favor of https://github.com/SAP/open-ux-tools/pull/5107. Review comments are or will be addressed in the new PR FYI.

PR #4671, opened 2026-05-13
skip test harness files when virtual preview endpoints are enabled #4657

kjose90:
- When `useVirtualPreviewEndpoints` is enabled, test harness HTML/JS files (`opaTests.qunit.html`, `unitTests.qunit.html`, `testsuite.qunit.html `and their .js/.ts counterparts) are no longer written to disk when generating an app — the `fiori-tools-preview` middleware serves them virtually via test: entries in `ui5.yaml`

- Migration warning: When converting an existing project via convert `preview-config` and the `unit-test` script references `unitTests.qunit.html`, a warning is logged before the rename:
Unit test files will be discovered automatically using the default pattern `'/test/**/Test.{js,ts}'`. If your unit test files do not match this pattern, below warning will be displayed
Screenshot 2026-05-12 at 10 12 49

- updated snapshots


packages/app-config-writer/src/preview-config/preview-files.ts
```diff
@@ -24,6 +24,11 @@ const renameMessage = (filePath: string): string =>
 export async function renameSandbox(fs: Editor, basePath: string, path: string, logger?: ToolsLogger): Promise<void> {
     const filePath = join(await getWebappPath(basePath), path);
     if (fs.exists(filePath)) {
+        if (path.endsWith('unitTests.qunit.html')) {
+            logger?.warn(
+                `Unit test files will be discovered automatically using the default pattern '/test/**/*Test.{js,ts}'. If your unit test files do not match this pattern, add a 'pattern' property to the QUnit test entry in your UI5 YAML configuration (e.g. pattern: '/test/unit/controller/*.{js,ts}').`
```
heimwege: Just a thought. We could add a link to the test config options of preview middleware to make it easier for the user to know where to add this parameter. What do you think?

`For details see https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/README.md#configuration-option-test`

packages/fiori-app-sub-generator/test/int/fiori-elements/expected-output/alp_v4_cap_typescript/app/alp_v4_cap_typescript/ui5.yaml
```diff
@@ -26,6 +26,7 @@ server:
       configuration:
         flp:
           theme: sap_fiori_3
+          path: test/flp.html
```
heimwege: global remark throughout multiple files: That's the default path. Everything should be fine even without that line I guess. And testing if this is transformed to an actual path /test/flp.html is already covered by unit tests in preview-middleware

packages/fiori-freestyle-writer/src/index.ts
```diff
@@ -201,6 +202,26 @@ async function generate<T>(basePath: string, data: FreestyleApp<T>, fs?: Editor,
         await applyCAPUpdates(fs, ffApp.service.capService, settings);
     }
 
+    if (isEdmxProjectType && addTests && ffApp.appOptions?.useVirtualPreviewEndpoints) {
```
heimwege, line 205: Why is this limited to edmx? works with CAP node as well?

packages/ui5-test-writer/src/utils/opaQUnitUtils.ts
```diff
@@ -316,10 +316,12 @@ export function addPagesToJourneyRunner(pages: JourneyRunnerPage[], testOutDirPa
 interface PreviewTestEntry {
```
heimwege: There already exists a type that can be used. No need to re-define it and risking the two get out of sync sooner or later. See https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/src/types/index.ts#L74

packages/ui5-test-writer/src/utils/opaQUnitUtils.ts
```diff
@@ -316,10 +316,12 @@ export function addPagesToJourneyRunner(pages: JourneyRunnerPage[], testOutDirPa
...
     framework?: string;
     path?: string;
+    pattern?: string;
 }
 
 /** Shape of the `fiori-tools-preview` middleware configuration relevant to OPA5 detection */
 interface PreviewMiddlewareConfig {
```
heimwege: Same here: We already have a type that should be re-used rather than re-defined: https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/src/types/index.ts#L128
If you don't want all of those properties you can use the TS utilitype `Pick<>` to reduce.

packages/ui5-test-writer/src/utils/opaQUnitUtils.ts
```diff
@@ -346,3 +348,34 @@ export async function hasVirtualOPA5(basePath: string): Promise<boolean> {
...
+): Promise<void> {
+    for (const yamlFile of [FileName.Ui5Yaml, FileName.Ui5LocalYaml, FileName.Ui5MockYaml]) {
+        const yamlPath = join(basePath, yamlFile);
+        if (!fs.exists(yamlPath)) {
+            continue;
+        }
+        const yamlConfig = await readUi5Yaml(basePath, yamlFile, fs);
+        const previewMiddleware = yamlConfig.findCustomMiddleware<PreviewMiddlewareConfig>('fiori-tools-preview');
+        if (previewMiddleware?.configuration?.flp) {
+            previewMiddleware.configuration.flp.path = 'test/flp.html';
```
heimwege: I think this is the reason for the updated YAML files. Is there a specific reason for writing the default to the yaml config? IF not I think we can delete this line.
Also it just checks for `previewMiddleware?.configuration?.flp` which means if I already defined a path 'my/beautiful/path' it will be overridden w/o notification.

packages/fiori-elements-writer/src/index.ts
```diff
@@ -69,16 +69,17 @@ function getTypeScriptIgnoreGlob<T extends {}>(feApp: FioriElementsApp<T>, coerc
...
     { useVirtualPreviewEndpoints }: { useVirtualPreviewEndpoints?: boolean },
     flpAppId?: string
-): { htmlTarget: string } {
+): { htmlTarget: string; useVirtualPreviewEndpoints?: boolean } {
     const flpTarget = useVirtualPreviewEndpoints ? 'flp' : 'flpSandbox';
     const htmlTarget = `test/${flpTarget}.html#${flpAppId}`;
     return {
-        htmlTarget
+        htmlTarget,
+        useVirtualPreviewEndpoints: useVirtualPreviewEndpoints
```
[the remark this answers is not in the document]
heimwege: @kjose90 why is this comment not helpful? I think is actually is. Just like `htmlTarget` does it you can just write `useVirtualPreviewEndpoints` here. I don't mind if we would go for `useVirtualPreviewEndpoints: useVirtualPreviewEndpoints` but then also `htmlTarget: htmlTarget`. Mixing both looks a bit odd.

packages/fiori-freestyle-writer/src/index.ts
```diff
@@ -201,6 +202,26 @@ async function generate<T>(basePath: string, data: FreestyleApp<T>, fs?: Editor,
...
+        await addVirtualTestConfig(
```
heimwege, line 208: Maybe adding a comment explaining why we don't / can't use the default `pattern` here would help understanding this at a later point in time.

heimwege (commented):
see comments

packages/fiori-freestyle-writer/src/index.ts
kjose90: we add tests only for EDMX project for FF https://github.com/SAP/open-ux-tools/blob/dab8f213ca0ed34ae2b563a395d50b1fe8b8149b/packages/fiori-freestyle-writer/src/index.ts#L151

packages/fiori-app-sub-generator/test/int/fiori-elements/expected-output/alp_v4_cap_typescript/app/alp_v4_cap_typescript/ui5.yaml
kjose90: OK perfect - tested and it works without `path: test/flp.html` - was unaware of this.
Reverted this change and snapshot

packages/fiori-elements-writer/src/index.ts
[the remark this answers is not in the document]
kjose90: woooops - think i misread this hyperspace proposal .. fixed

packages/fiori-freestyle-writer/src/index.ts
kjose90: [comment](https://github.com/SAP/open-ux-tools/pull/4671/changes#diff-32f1effcd1c481e35cf5149955d67cf304ed5ccaac722324f67b9683a23b8d55R206)

packages/ui5-test-writer/src/utils/opaQUnitUtils.ts
kjose90: your right. since it works without adding default path `test/flp.html` - i have removed this 
https://github.com/SAP/open-ux-tools/pull/4671/changes#diff-de7885f0dc4736fb320442225f997f4e01803c221130d96566c76f13d9e90e74R363

packages/ui5-test-writer/src/utils/opaQUnitUtils.ts
kjose90: [Removed](https://github.com/SAP/open-ux-tools/pull/4671/changes#diff-de7885f0dc4736fb320442225f997f4e01803c221130d96566c76f13d9e90e74L316) to use types from preview middleware

packages/ui5-test-writer/src/utils/opaQUnitUtils.ts
kjose90: https://github.com/SAP/open-ux-tools/pull/4671/changes#diff-de7885f0dc4736fb320442225f997f4e01803c221130d96566c76f13d9e90e74R328

packages/fiori-freestyle-writer/src/index.ts
heimwege: hm. interesting. I don't understand why but at least it's consistant.

packages/ui5-test-writer/src/utils/opaQUnitUtils.ts
heimwege: One small addition. I find the naming a bit misleading because there exists more than one middleware 🙈 That's clearly preview middleware's fault but we can only change this doing a major version change because it's an exported type name etc. So it will probably never happen.

I'd suggest to change

`import type { TestConfig, MiddlewareConfig } from '@sap-ux/preview-middleware';`

to

`import type { TestConfig as PreviewMiddlewareTestConfig, MiddlewareConfig as PreviewMiddlewareConfig} from '@sap-ux/preview-middleware';`

to improve readability.

kjose90: Nice - [imported](https://github.com/SAP/open-ux-tools/pull/4671/changes#diff-de7885f0dc4736fb320442225f997f4e01803c221130d96566c76f13d9e90e74R12)

docirl (approved):
Changes to OPA test flows look good and as expected
Thanks for the cleanup for type import from preview middleware
Test updates look as expected
Changeset ok

heimwege (approved):
- review comment addressed
- changeset ok
- text coverage brilliant
- did NOt test manually

IainSAP (approved):
Thanks @kjose90 

- Updates seem appropriate
- Snapshots updated
- Not manually tested

PR #4674, opened 2026-05-14
chore: override sharp to >=0.33.5 to use prebuilt platform binaries

devinea:
## Summary

- Adds a pnpm override `sharp@<0.33.0: >=0.33.5` to force `@xenova/transformers`'s `sharp@^0.32.0` dependency up to `0.33.5+`
- Removes the now-unnecessary `pnpm rebuild sharp -r` from the `postinstall` script

## Problem

`sharp@0.32.x` requires native compilation at install time. On CI runners (darwin-arm64), the prebuilt binary `sharp-darwin-arm64v8.node` is not available and the rebuild fails, breaking the `@sap-ux/fiori-docs-embeddings` build.

## Fix

`sharp@0.33.0+` changed its distribution model — it ships prebuilt `@img/sharp-<platform>` packages as optional dependencies (e.g. `@img/sharp-darwin-arm64`), so no native compilation is needed. The API used by `@xenova/transformers` (`.metadata()`, `.rotate()`, `.raw()`, `.toBuffer()`, `.resize()`, `.extend()`, `.extract()`) is unchanged across this version bump.

## Test plan

- [ ] CI build passes for `@sap-ux/fiori-docs-embeddings` on all platforms (ubuntu, windows, macos)

### DIFF
--- a/.changeset/sharp-override-prebuilt-binaries.md
+++ b/.changeset/sharp-override-prebuilt-binaries.md
@@ -1,0 +1,6 @@
+---
+"@sap-ux/fiori-docs-embeddings": patch
+"@sap-ux/fiori-mcp-server": patch
+---
+
+chore: override sharp to >=0.33.5 to fix CI build on darwin-arm64
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     },
     "scripts": {
         "update-ts-references": "update-ts-references",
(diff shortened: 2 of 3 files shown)


IainSAP (approved):

longieirl (approved):
Changes look good.

donal-tobin-sap (approved):

PR #4680, opened 2026-05-15
feat(ui5-test-writer): TypeScript OPA5 test generation for Fiori Elements V4

dudajSAP:
## Summary

Adds `enableTypeScript` option to `generateOPAFiles()`. When enabled, generates `.ts` files using ES module syntax with proper `Given`/`When`/`Then` type annotations instead of AMD `sap.ui.define` JavaScript files.

- TypeScript EJS templates for JourneyRunner, page objects (ListReport, ObjectPage), and journey files
- Dynamic `OpaJourneyTypes.d.ts` generation with page-specific type entries for Given/When/Then
- `FileExtension` type for consistent extension handling across the writer
- Unit and integration tests for the TS generation path (coverage >96%)

### Scope

- **LROP apps only** (ListReport + ObjectPage) — FPM deferred
- **New app generation** (clean slate) — standalone/re-generation mode deferred

### Key design decisions

- TS templates use ES module `import`/`export` (not `sap.ui.define` AMD)
- `opaTests.qunit.js` stays JS (QUnit bootstrap loaded by HTML)
- `OpaJourneyTypes.d.ts` dynamically generated per page from EJS template
- No `tsconfig.json` modification (app's existing namespace path mapping resolves types)
- `CustomObjectPageActions` interface includes `iPressSectionIconTabFilterButton` for ObjectPages

### Workarounds for `@sapui5/types` deficiencies

- `onTable("")` — type requires `vTableIdentifier` (should be optional)
- `iCheckFilterField({ property: "..." })` — object form instead of plain string
- `iCheckSection({ section: "..." }, {})` — `mState` required in types (should be optional)
- Page constructors provide both `entitySet` and `contextPath` (types require both)

## Test plan

- [x] `pnpm build` passes
- [x] `pnpm lint` — 0 errors
- [x] `pnpm test` — 297 tests pass, coverage >96%
- [x] Integration tests pass (4 tests including 2 new TS tests)
- [x] Generated TS output passes `tsc --noEmit` with `@sapui5/types`
- [x] Tested with `test.sample.travels.v4` (simple LROP) and `test.sample.ismm.pps.purchaseorders1` (complex, 1 LR + 6 OPs)

## References

- Jira: FXUBRQ34-12420
- Internal issue: https://github.wdf.sap.corp/ux-engineering/tools-suite/issues/37921


packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
```diff
@@ -58,6 +60,7 @@ export async function generateOPAFiles(
 
     const config = createConfig(manifest, opaConfig, hideFilterBar);
 
+    const ext: FileExtension = opaConfig.enableTypeScript ? '.ts' : '.js';
```
heimwege: Please user proper `const` names, e.g. `fileExtension`

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
```diff
@@ -389,7 +394,7 @@ function findLROP(
  * @param rootCommonTemplateDirPath - template root directory for common files
  */
 function writeCommonAndPageFiles(writeContext: WriteContext, rootCommonTemplateDirPath: string): void {
-    const { config, rootV4TemplateDirPath, testOutDirPath, editor, journeyParams } = writeContext;
+    const { config, rootV4TemplateDirPath, testOutDirPath, editor, journeyParams, ext } = writeContext;
```
heimwege: Please user proper `const` names, e.g. `fileExtension`

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
heimwege: very nit-pick: Actually the dot (`.`) is just the delimiter that is not part of the filename extension

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
```diff
@@ -45,7 +47,7 @@ import { readHashFromFlpSandbox } from './utils/flpSandboxUtils';
  */
 export async function generateOPAFiles(
     basePath: string,
-    opaConfig: { scriptName?: string; appID?: string; htmlTarget?: string },
+    opaConfig: { scriptName?: string; appID?: string; htmlTarget?: string; enableTypeScript?: boolean },
```
heimwege: This inline type is repeated twice (`generateOPAFiles()` and `createConfig()`) so it should be extracted into a named type (placed on top of this file ⬆️ (below the imports) to avoid future drift.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
```diff
@@ -627,7 +666,7 @@ export async function generatePageObjectFile(
     if (pageConfig) {
         const rootTemplateDirPath = join(__dirname, `../templates/${applicationType}`); // Only v4 is supported for the time being
         const testOutDirPath = join(await getWebappPath(basePath), 'test');
-        writePageObject(pageConfig, rootTemplateDirPath, testOutDirPath, editor);
+        writePageObject(pageConfig, rootTemplateDirPath, testOutDirPath, editor, '.js');
```
heimwege: Questions:
1. What is this function used for it seems to just be used in unit tests. Is this a public API? For whom?

2. Why is the page object file hard coded to JS? If it is not supported shouldn't the function nevertheless accept the flag and log an info that TS is not (yet?) supported so that the user is not surprised by getting JS?
But preferably we should support this.

3. Also there's a comment in fiori-elements.test.ts I don't understand: `// The new page object is generated as JS because generatePageObjectFile doesn't know about TS yet` 🤷🏻 -> So is it planned? And why is that comment in the test instead of the productive code (here)?

packages/ui5-test-writer/templates/v4/integration/FirstJourney.ts
```diff
@@ -0,0 +1,32 @@
+import opaTest from "sap/ui/test/opaQunit";
```
heimwege, line 1: This file lacks the auto-generated warning banner used in `ListReportJourney.ts` and `ObjectPageJourney.ts`. Is it missing on purpose?

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
```diff
@@ -296,7 +300,8 @@ function createConfig(
         pages: [],
         opaJourneyFileName: opaConfig.scriptName ?? 'FirstJourney',
         htmlTarget: opaConfig.htmlTarget ?? 'index.html',
-        hideFilterBar
+        hideFilterBar,
+        enableTypeScript: opaConfig.enableTypeScript
```
heimwege: `enableTypeScript` is set here but the writer derives it from `opaConfig.enableTypeScript` before building the config, and only ever uses thst after that. `FEV4OPAConfig.enableTypeScript` is dead state - it's passed into EJS templates via config (e.g. `JourneyRunner.ts` template uses config), but none of the templates actually reference `enableTypeScript`. Worth either removing it from `FEV4OPAConfig`, or using it as the canonical source instead of the raw `opaConfig`.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
```diff
@@ -419,14 +424,27 @@ function writeCommonAndPageFiles(writeContext: WriteContext, rootCommonTemplateD
...
+        join(testOutDirPath, 'integration', 'pages', `JourneyRunner${ext}`),
         config,
         undefined,
         {
             globOptions: { dot: true }
         }
     );
+
+    // Generate OpaJourneyTypes.d.ts when TypeScript is enabled
+    if (ext === '.ts') {
```
heimwege: Prefer `config.enableTypeScript` over the `ext === '.ts'` literal comparison

packages/ui5-test-writer/templates/v4/integration/types/OpaJourneyTypes.d.ts
```diff
@@ -0,0 +1,38 @@
+import type Opa5 from "sap/ui/test/Opa5";
+<% if (pages.some(p => p.template === 'ListReport')) { %>import type { actions as ListReportActions, assertions as ListReportAssertions } from "sap/fe/test/ListReport";
+<% } %><% if (pages.some(p => p.template === 'ObjectPage')) { %>import type { actions as ObjectPageActions, assertions as ObjectPageAssertions } from "sap/fe/test/ObjectPage";
+<% } %>import type { actions as TemplatePageActions, assertions as TemplatePageAssertions } from "sap/fe/test/TemplatePage";
+import type Shell from "sap/fe/test/Shell";
+import type BaseArrangements from "sap/fe/test/BaseArrangements";
+<% if (pages.some(p => p.template === 'ObjectPage')) { %>
+interface CustomObjectPageActions {
```
heimwege: The current approach defines a standalone `CustomObjectPageActions` interface inline in the generated `OpaJourneyTypes.d.ts`. This creates a maintenance burden: every time a developer adds a custom action or assertion to their `ObjectPage.ts` or `ListReport.ts`, they also have to manually mirror that change in `OpaJourneyTypes.d.ts`. The two files are structurally decoupled, so they will inevitably drift.

The approach used in [cap-node-fe-ts-sample](https://github.com/heimwege/cap-node-fe-ts-sample/blob/main/app/samples/webapp/test/integration/types/OpaJourneyTypes.d.ts#L4) avoids this by having the page files export their actions and assertions objects, and importing them into `OpaJourneyTypes.d.ts` via `typeof`:
```ts
import type { actions as CustomObjectPageActions, assertions as CustomObjectPageAssertions } from '<appPath>/test/integration/pages/ObjectPage';

type When = Opa5 & BaseArrangements & {
      onTheObjectPage: Opa5 & ObjectPageActions & TemplatePageActions & typeof CustomObjectPageActions;
  };
type Then = Opa5 & BaseArrangements & {
      onTheObjectPage: Opa5 & ObjectPageAssertions & TemplatePageAssertions & typeof CustomObjectPageAssertions;
  };
```

With this pattern, custom actions/assertions added to the page files are automatically reflected in When/Then. No manual update to `OpaJourneyTypes.d.ts` required. It also covers `Then` with custom assertions, which the current generated output omits entirely.

This requires changing the page object templates to export actions and assertions as named exports (as in [the reference](https://github.com/heimwege/cap-node-fe-ts-sample/blob/main/app/samples/webapp/test/integration/pages/ObjectPage.ts#L9)) rather than embedding them in an anonymous `CustomPageDefinitions` object.

heimwege (commented):
see comments

packages/ui5-test-writer/templates/v4/integration/FirstJourney.ts
sap-sebelao: The JS version of the file does not contain it either, I think we did not update the file with this banner to keep it consistent with behavior that was present before we started development on the generator.

IIRC Some other packages' tests started failing when we started introducing changes into FirstJourney.js
However, maybe a one-time change just to add the banner to FirstJourney.js, where we can monitor and update other's snapshots if they fail would be helpful to do separately to keep the generated files consistent.

Maybe @docirl can explain the original reasoning behind this decision better as he designed and implemented the warning comment.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
sap-sebelao: I would say a bigger problem is using these as string literals. Storing 'ts' and 'js' as constants to be reused would be helpful, as the literals '.ts' or '.js' are used in other places in this set of changes already (such as passing '.js' to writePageObject)

packages/ui5-test-writer/templates/v4/integration/ObjectPageJourney.ts
```diff
@@ -0,0 +1,104 @@
...
+<% if(navigationParents.parentOPName) { %>
+        Then.onThe<%- navigationParents.parentOPName%>.iSeeThisPage();
+        Then.onThe<%- navigationParents.parentOPName%>.onTable({ property: "<%- navigationParents.parentOPTableSection %>" }).iCheckRows();
+        When.onThe<%- navigationParents.parentOPName%>.onTable({ property: "<%- navigationParents.parentOPTableSection %>" }).iPressRow(0);
+<% } %>
+        Then.onThe<%- name%>.iSeeThisPage();
+    });
+
+<% if (headerSections?.length > 0) { -%>
+    opaTest("Check header facets of the Object Page", function (Given: Given, When: When, Then: Then) {
```
sap-sebelao: Since https://github.com/SAP/open-ux-tools/pull/4632 was merged, this template file is now outdated as it does not contain Action-related test code templates.

Make sure to update the TS templates to be up-to-date with the current JS version on main.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
dudajSAP: Done in commit bde53b417. The variable was renamed to `dotFileExtension` (along with the `WriteContext` field, `ensurePageExists` and `writePageObject` parameters, and JSDoc) — capturing both your "proper const name" point and the nit-pick about the dot in the related comment.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
dudajSAP: Done in commit bde53b417 — destructured as `dotFileExtension` along with all other usages.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
dudajSAP: Addressed in commit bde53b417. Replaced string literal extensions with a `DotFileExtension` enum (in `types.ts`):

```ts
export enum DotFileExtension {
    JS = '.js',
    TS = '.ts'
}
```

The variable is now named `dotFileExtension: DotFileExtension` everywhere, making it explicit that the value includes the dot delimiter. Both the type and variable names reflect that.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
dudajSAP: Done in commit bde53b417. The type is now `OPAGenerationOptions`, exported from `types.ts` (alongside `FEV4OPAConfig` and `WriteContext`):

```ts
export type OPAGenerationOptions = {
    /** The name of the OPA journey file. If not specified, 'FirstJourney' will be used. */
    scriptName?: string;
    ...
};
```

Per-field JSDoc lives on the type, not duplicated in function JSDoc. Both `generateOPAFiles` and `createConfig` use it. The parameter is also renamed `opaConfig` → `options` for naming consistency.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
dudajSAP: Fixed in commit bde53b417. Removed `enableTypeScript` from `FEV4OPAConfig` entirely. Now `enableTypeScript` is read from `OPAGenerationOptions` once at the entry point of `generateOPAFiles`, immediately converted to `dotFileExtension`, and threaded through `WriteContext.dotFileExtension` as the single source of truth for downstream code.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
dudajSAP: Addressed in commit bde53b417, but with a different solution than originally suggested. Since the related comment about `FEV4OPAConfig.enableTypeScript` removes that field entirely (it was dead state), there's no boolean to compare against on the config. The check is now:

```ts
if (dotFileExtension === DotFileExtension.TS) { ... }
```

Comparing against the enum member is cleaner than the previous string literal `'.ts'`, and `dotFileExtension` is the single source of truth in `WriteContext`. Let me know if you'd prefer to bring back the boolean on the config — but I think this is cleaner.

packages/ui5-test-writer/templates/v4/integration/types/OpaJourneyTypes.d.ts
dudajSAP: Done in commit bde53b417. Refactored exactly as you suggested — the page templates now export named `actions` and `assertions`, and `OpaJourneyTypes.d.ts` imports them via `typeof` per page:

Page template (e.g. `pages/ObjectPage.ts`):
```ts
export const actions = {
    iPressSectionIconTabFilterButton(this: Opa5, section: string) { ... }
};
export const assertions = {};
export default new ObjectPage({...}, { actions, assertions });
```

Generated `OpaJourneyTypes.d.ts`:
```ts
import type { actions as TravelObjectPageCustomActions, assertions as TravelObjectPageCustomAssertions } from "travelapp/test/integration/pages/TravelObjectPage";

export type When = Opa5 & BaseArrangements & {
    onTheTravelObjectPage: Opa5 & ObjectPageActions & TemplatePageActions & typeof TravelObjectPageCustomActions;
    ...
};
export type Then = Opa5 & BaseArrangements & {
    onTheTravelObjectPage: Opa5 & ObjectPageAssertions & TemplatePageAssertions & typeof TravelObjectPageCustomAssertions;
    ...
};
```

Custom actions/assertions added to a page file now flow into `When`/`Then` automatically. The standalone `CustomObjectPageActions` interface is gone, and `Then` now correctly includes custom assertions (which it didn't before).

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
dudajSAP: Thanks for raising this. Replying per sub-question:

**(2) Why hardcoded JS?** Fixed in commit bde53b417. `generatePageObjectFile` now accepts an optional `dotFileExtension` parameter:

```ts
pageObjectParameters: { targetKey: string; appID?: string; dotFileExtension?: DotFileExtension }
```

Defaults to `DotFileExtension.JS` for backward compatibility. The integration test now exercises the TS path. Two new unit tests cover both the TS path and the JS default.

**(3) Planning comment in test:** Removed in the same commit — once `generatePageObjectFile` supports TS, the comment is obsolete and the integration test asserts the new behavior directly.

**(1) Public API status:** As we discussed with Dominik, the question belongs to the function `generatePageObjectFile()` itself — why is this exported, since it's probably used only in tests? I've just extended the current solution in this PR. So please David (@docirl), is this method used somewhere else?

packages/ui5-test-writer/templates/v4/integration/FirstJourney.ts
dudajSAP: For now I leave it as it is. Once we will have the clear overview, I can add the warning there.

packages/ui5-test-writer/src/types.ts
```diff
@@ -194,6 +196,7 @@ export type WriteContext = {
     testOutDirPath: string;
     editor: Editor;
     journeyParams: JourneyParams;
+    ext: FileExtension;
```
sap-sebelao: Still `ext` here?

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
sap-sebelao: I would personally prefer a well-named boolean here, the condition might be slightly harder to understand if read without wider context ("This is checking the file extension, why?" vs. "This is checking if we are in TS mode, okay...")

Is it not possible to use the opaConfig value here?

packages/ui5-test-writer/src/types.ts
dudajSAP: You looked into the outdated version. `ext` is not used right now nowhere in the current version.

sap-sebelao: Right, sorry, somehow it opened the previous one

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
heimwege: Please avoid using enums: https://www.totaltypescript.com/why-i-dont-like-typescript-enums

TypeScript enums are one of the few features that aren't just type-level decorations. They actually generate custom JavaScript code when compiled. Object constants with as const keep things simple, predictable, and aligned with standard JavaScript.
And they don't support `erasableSyntaxOnly`.

```ts
export const DotFileExtension = {
    JS: '.js',
    TS: '.ts'
} as const;
```

sap-sebelao: The placement of the enum is strange, exactly because it produces actual JS code, I would expect types.ts to only contain types (but this is already violated as it contains `const SupportedPageTypes` from earlier).

I would say string enums are fine, imho, I actually like the strictness of having to use the enum members or same enum members not being interchangeable in certain situations, but I understand where the author of that article is coming from.

If there is a real plan to use `erasableSyntaxOnly` in the future for this project, it should probably be added sooner or later as an eslint rule, at least as a warning first.

packages/ui5-test-writer/templates/v4/integration/ObjectPageJourney.ts
dudajSAP: There are also a new PR made by @docirl , which will need the update of the TS template, once they will be merged.
https://github.com/SAP/open-ux-tools/pull/4702

packages/ui5-test-writer/templates/v4/integration/FirstJourney.ts
docirl: We didn't add the warning as this file does not contain 'auto generated' tests that are added using the spec api. It is mostly static (it has some very basic conditions in the template) and is not updated when re-generating the tests.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
docirl: For (1) `generatePageObjectFile` function looks to be un-used other than tests so should be removed (not here, I'll get it in another pr). Likely missed by previous code change.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
heimwege: Yes there are plans to get rid of enums to support Node's native type-stripping feature (w/o needing to `--experimental-transform-types`). But there's no need for an eslint rule, just enable the `erasableSyntaxOnly` option in TS settings (e.g. like we already do in the preview-middleware: https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/tsconfig.json#L12)

dudajSAP: Done in commit 369a0a45b — the enum has been replaced with a const object as you suggested.

Related `erasableSyntaxOnly`: Done in commit 442b48ac3 — added `"erasableSyntaxOnly": true` to `packages/ui5-test-writer/tsconfig.json`. Source already complies; tests and TS check pass.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
dudajSAP: Thanks @docirl — that matches what we suspected. Leaving the function in place for this PR.

packages/ui5-test-writer/templates/v4/integration/FirstJourney.ts
dudajSAP: Thanks for the confirmation @docirl. I'm leaving `FirstJourney.ts` without the header.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
dudajSAP: Done in commit 369a0a45b. The literal comparison is gone — `OpaJourneyTypes.d.ts` generation is now extracted into a dedicated `writeOpaJourneyTypes()` function called from `generateOPAFiles()` only when `options.enableTypeScript` is set. The `if (options.enableTypeScript)` is now at the top level alongside the other `generateOPAFiles` decisions, reading naturally as "we're in TS mode". And in commit 442b48ac3 the underlying enum was replaced with a const object per @heimwege's suggestion (no runtime emit).

packages/ui5-test-writer/templates/v4/integration/ObjectPageJourney.ts
```diff
@@ -0,0 +1,104 @@
...
+        When.onThe<%- name%>.iPressSectionIconTabFilterButton("<%- section.id %>");
+<% } -%>
+        Then.onThe<%- name%>.iCheckSection({ section: "<%- section.id %>" }, {});
+<% if (section?.subSections?.length > 0) { -%>
+<% section.subSections.forEach(function(subSection) { -%>
+        //When.onThe<%- name%>.iGoToSection({ section: "<%- section.id %>", subSection: "<%- subSection.id %>" });
+        Then.onThe<%- name%>.iCheckSubSection({ section: "<%- subSection.id %>" });
+<% if (subSection.fields && subSection.fields.length > 0) { -%>
+<% subSection.fields.forEach(function(field) { -%>
+        Then.onThe<%- name%>.onForm("<%- subSection.id %>").iCheckField({ property: "<%- field.property %>" });
```
heimwege: Is there a specific reason for diverging with the generated JS tests?

```JS
// https://github.com/dudajSAP/open-ux-tools/blob/main/packages/ui5-test-writer/templates/v4/integration/ObjectPageJourney.js#L79-L80
Then.onThe<%- name%>.onForm({ section: "<%- subSection.id %>" })
```
vs
```TS
Then.onThe<%- name%>.onForm("<%- subSection.id %>")
```

dudajSAP: Thanks for catching this. It was passing the type check but changed the test logic. I've restored the JS-style object form in commit a92c379f3, with a cast as an escape hatch against the (incorrect) `@sapui5/types`:
```ts
.onForm({ section: "<id>" } as unknown as FormIdentifier)
```

packages/ui5-test-writer/test/unit/utils/opaQUnitUtils.test.ts
```diff
@@ -476,6 +478,123 @@
...
+        const lines = result.split('\n');
+        const newPageLine = lines.find((l) => l.includes('onTheNewPage'));
+        // Existing entries use 8-space indentation in the template
+        expect(newPageLine).toMatch(/^ {8}/);
+    });
+
+    test('adds a trailing comma to the last existing pages entry when missing', () => {
+        // Construct a file whose last pages entry has no trailing comma
+        const noTrailingComma = JOURNEY_RUNNER_TS_FILE.replace(
+            'onTheTravelObjectPage: TravelObjectPage\n',
```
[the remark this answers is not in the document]
dudajSAP: Already fixed.

packages/ui5-test-writer/templates/v4/integration/ObjectPageJourney.ts
dudajSAP: Both PRs (#4632 Object Page actions and #4702 semantic key filters / global search) are now merged into upstream/main and the TS templates have been updated to match in commit e0dd1708d:

- `ObjectPageJourney.ts` — header actions test + section actions and table create/delete buttons mirrored from #4632
- `ListReportJourney.ts` — semantic key filter adaptation test + commented global search example mirrored from #4702

Verified that the generated TS journeys for both sample apps are byte-equivalent to the JS journeys after stripping TS adaptations (typed Given/When/Then args, onTable(""), `as unknown as FormIdentifier` casts, etc.).

packages/ui5-test-writer/src/utils/opaQUnitUtils.ts
```diff
@@ -295,19 +312,142 @@ export function splicePageIntoJourneyRunner(fileContent: string, pages: JourneyR
...
+ * @param pages - pages to add
+ * @returns the updated file content, or the original content unchanged if nothing was added
+ */
+export function splicePageIntoJourneyRunnerTs(fileContent: string, pages: JourneyRunnerPage[]): string {
+    if (fileContent.length > MAX_FILE_CONTENT_LENGTH) {
+        return fileContent;
+    }
+    // Determine which pages are not yet present by checking for their `Custom<Page>` import line
+    const toAdd = pages.filter((page) => {
+        const importPattern = new RegExp(`from\\s+"\\./${page.targetKey}"`);
```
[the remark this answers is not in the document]
dudajSAP: Addressed in 088c98b5a — added an `escapeRegex` helper at the top of `opaQUnitUtils.ts` and wrapped `page.targetKey` with it before constructing the `RegExp`. While Fiori `targetKey` values are alphanumeric in practice, this removes the latent risk of false-positive matches for any future identifier that contains regex metacharacters.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
```diff
@@ -563,14 +620,21 @@ function writeJourneyFiles(
...
         generatedJourneyPages.push(appFeatures.fpm.name);
-        const fpmPage = ensurePageExists(appFeatures.fpm.name, config, rootV4TemplateDirPath, testOutDirPath, editor);
+        const fpmPage = ensurePageExists(
+            appFeatures.fpm.name,
+            config,
+            rootV4TemplateDirPath,
+            testOutDirPath,
+            editor,
+            dotFileExtension
+        );
```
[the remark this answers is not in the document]
dudajSAP: Addressed in 088c98b5a — `ensurePageExists` is now called with `DotFileExtension.JS` for the FPM page-object, regardless of `enableTypeScript`. Updated the inline comment to reflect that we explicitly force JS for FPM until `FPM.ts` exists. The follow-up — adding `FPM.ts` / `FPMJourney.ts` templates — is tracked separately in the project TODO list.

packages/ui5-test-writer/templates/v4/integration/FirstJourney.ts
```diff
@@ -0,0 +1,32 @@
...
+import type { Given, When, Then } from "./types/OpaJourneyTypes";
```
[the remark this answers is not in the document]
dudajSAP: The bot's reading is inverted here — the *unit* snapshot's relative path `./types/OpaJourneyTypes` is the correct, intentional output (changed in commit 530d8ca5c to remove the dependency on consumer-app `tsconfig.json` `paths` mappings). The integration snapshot was simply stale; it had not been regenerated since the templates moved to relative imports. Addressed in 088c98b5a by regenerating the integration snapshots (which now also use `./types/OpaJourneyTypes`), and migrating `test/integration/fiori-elements.test.ts` to ESM-aware `__dirname` derivation so Jest under `--experimental-vm-modules` can run it again.

packages/ui5-test-writer/templates/v4/integration/pages/JourneyRunner.ts
```diff
@@ -0,0 +1,28 @@
...
+<% } -%>
+<%- pages.map((page) => 'import Custom' + page.targetKey + ' from "./' + page.targetKey + '";').join('\n') %>
+
+const runner = new JourneyRunner({
+    launchUrl: sap.ui.require.toUrl("<%- appPath %>") + "/<%- htmlTarget %>",
+    pages: {
+<%- pages.map((page) =>
+'        onThe' + page.targetKey + ': new ' + page.template + '(\n' +
+'            {\n' +
+'                appId: "' + appID + '",\n' +
```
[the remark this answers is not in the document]
dudajSAP: Addressed in 088c98b5a — the template now reads `appId` from `page.appID` (alongside `page.componentID`), making the per-page block self-consistent. Both the fresh-generation path (which builds `pages` from `config.pages` where each entry carries `config.appID`) and the splicer path (which already passes `appID` per page in `JourneyRunnerPage`) populate the field, so behavior is unchanged today and the fragile coupling on the root-scope `appID` is removed.

packages/ui5-test-writer/src/fiori-elements-opa-writer.ts
```diff
@@ -110,6 +121,10 @@ export async function generateOPAFiles(
         }
     }
 
+    if (enableTypeScript) {
+        writeOpaJourneyTypes(writeContext);
```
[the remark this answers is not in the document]
dudajSAP: Addressed in 088c98b5a — hoisted an `activeContext` variable in `generateOPAFiles` that holds either the original `writeContext` or the resolved `standaloneWriteContext` (in the standalone branch where `hasJourneyRunner` is false). `writeOpaJourneyTypes(activeContext)` is now called with the same context that was used to emit journey/page files, so any future field added to `WriteContext` that's resolved by `resolveStandaloneWriteContext` will be visible to the type-defs generation as well.

dudajSAP:
This PR brings TypeScript parity with the JS OPA5 generator as it stands today. Ongoing JS discussions around custom-test handling (preservation/re-integration on re-generation) are out of scope here. Once those JS changes are merged, a follow-up PR will mirror the equivalent TS behavior.

.changeset/ui5-test-writer-typescript-opa.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/ui5-test-writer': patch
```
heimwege: Actually this is a new feature so we should do a `minor` version bump

heimwege (commented):
Just one comment left from my pov and one todo reported by SONAR

.changeset/ui5-test-writer-typescript-opa.md
dudajSAP: Thank you. Fixed.

dudajSAP:
I removed the `TODO` keyword, but we still need to evaluate whether we want to support test generation for the Flexible Programming Model (FPM) page.

Given the nature and flexibility of FPM, I'm not even sure whether generating meaningful tests is feasible in all FPM cases. To save time, I didn't want to address this as part of the current PR. For now, I only replaced the `TODO` so that Sonar would no longer complain. I agree this is not an ideal solution, but it avoids blocking the current change.

We should make a decision on the expected behavior and then create a follow-up PR to properly address this item.

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.ts
```diff
@@ -0,0 +1,109 @@
...
+        Given.iStartMyApp();
+        <%_ startPages.forEach(function(pageName) { %>
+        Then.onThe<%- pageName %>.iSeeThisPage();
+        <%_ }); -%>
+    });
+
+    <%_ if (!hideFilterBar && filterBarItems && filterBarItems.length > 0) { -%>
+    opaTest("Check filter bar", function (_Given: Given, _When: When, Then: Then) {
+        <%_ filterBarItems.forEach(function(item) { _%>
+        Then.onThe<%- startLR%>.onFilterBar().iCheckFilterField({ property: "<%- item %>" });
```
docirl: I tested app generation for typescript with the mock odata service we normally use and the tests fail on the one of tests for `iCheckFilterField` when `property` is used here. 
I see a comment in the description noting this is a workaround for types. I compared the same test in a JS project with the same service passes without using property. Not sure if we can do much here if property is needed?

packages/fiori-elements-writer/src/index.ts
```diff
@@ -275,7 +275,13 @@ async function generate<T extends {}>(
             { useVirtualPreviewEndpoints: feApp.appOptions?.useVirtualPreviewEndpoints },
             feApp.app.flpAppId
         );
-        await generateOPAFiles(basePath, opaConfig, data.service.metadata, fs, log);
+        await generateOPAFiles(
+            basePath,
+            { ...opaConfig, enableTypeScript: feApp.appOptions?.typescript },
```
docirl, line 280: I added `enableTypescript` to be passed from the fiori-elements-writer as this is where the app generator calls the test writer from. Without this during generation for typescript the tests are still generated as JS.

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.ts
sap-sebelao: The line in the JS template is:

`Then.onThe<%- startLR%>Generated.onFilterBar().iCheckFilterField("<%- item %>");`

@dudajSAP Try to reuse the same approach, without the object identifier. If there is a type issue (possibly with the API having an incorrect type defined again), just try to add a cast as a temporary fix.

dudajSAP: Addressed in f94c380e6 — switched both `iCheckFilterField` call sites in `ListReportJourney.ts` to the plain-string form used in the JS template, with an `as unknown as FilterFieldIdentifier` cast to satisfy the (incorrect) type. Confirmed runtime behavior matches the JS variant; `tsc --noEmit` passes 0 errors against the generated output. The `FilterFieldIdentifier` import is wrapped in an EJS conditional so it's only emitted when filter-bar / semantic-key blocks are present.

docirl (approved):
- Tested typescript flow from both app generator and app info and working as expected
- Updates and test coverage are good
- Changeset added

Thanks @dudajSAP

docirl:
@heimwege could you review please? Would like to get this merged today if possible to include in next release.

heimwege (approved):
- changeset ok
- review comments addressed
- coverage is good
- did NOT test manually

IainSAP (approved):
Thanks @docirl 

- Updates look good, test updated as expected

Issue #4681, opened 2026-05-15
BUG - Adjust CPE panel border after splitter style changes


815are:
From Goran:
Due to this:

https://github.com/SAP/open-ux-tools/pull/4648

Where we made splitters be 1 pixel thick, we have some visual issues in CPE as then all horizontal lines look like splitters.

A simple fix is to make lines that are **not** actual splitters just be indented 15px from left & right.

<img width="786" height="380" alt="Image" src="https://github.com/user-attachments/assets/4e54df9d-8e4f-4fac-b8a4-e7f1c4dcc048" />

This way we have visual difference - if a thing is fully edge to edge; it's a splitter, and if it's not edge to edge; it's not a splitter.

There are a few places where this has to be done, two on the left side (below filter input fields), and two on the right side (above and below filter).

PR #4682, opened 2026-05-15
fix: Add margins to panel horizontal borders to distinguish them from splitter borders

815are:
Issue #4681

After style changes for splitter(#4648) - unresizable borders hard to differentiate from splitter border.
As proposed adjustment from Goran is that we apply 15px margins to non resizable border within panel.

Changes:
1. move `15px` horizontal padding to margin;
2. remove starting separator from https://github.com/SAP/open-ux-tools/blob/main/packages/control-property-editor/src/panels/changes/ChangesPanel.tsx#L55 because it introduced double border with border from https://github.com/SAP/open-ux-tools/blob/main/packages/control-property-editor/src/panels/changes/ChangesPanel.module.scss#L8
image

demo:
image

image


vadson71 (commented):
Style changes look good
Didn't test locally

vadson71 (approved):

broksy (approved):
Changeset reflects to the changes made.
Approved from my side.

PR #4702, opened 2026-05-19
feat(ui5-test-writer): OPA tests for semantic key filters, global search in List Report

docirl:
# OPA Tests for Semantic Key Filters and Global Search in List Report

### New Features

✨ Adds OPA test generation support for semantic key filters and global search in the List Report journey template. The changes introduce utilities to detect semantic key properties from OData metadata, compare them against filter bar fields, and generate the appropriate OPA test steps — including adaptive filter bar handling for semantic key properties not initially visible.

### Changes

* `types.ts`: Made `deleteButton.visible` optional and added a new `semanticKey` property to `ListReportFeatures` with sub-fields `filterBarHasSemanticKey`, `semanticKeyProperties`, and `missingFromFilterBar`.

* `listReportUtils.ts`:
  - Refactored `getListReportFeatures` to extract `entitySetName` early and reuse it consistently.
  - Populates the new `semanticKey` feature block, computing which semantic key properties are missing from the filter bar.
  - Added `getSemanticKeyProperties()`: parses OData metadata XML to extract `SAP Common SemanticKey` annotation property paths, with optional label resolution.
  - Added `isSemanticKeyInFilterBar()`: checks whether all semantic key properties for a given entity set are present as filter fields.

* `ListReportJourney.js` (template): 
  - Added a conditional OPA test block that, when semantic key properties are missing from the filter bar, generates steps to open filter adaptation, add the missing fields, and verify them.
  - Added a commented-out template for global search (`iChangeSearchField` / `iExecuteSearch`) with a note about prerequisite test data setup.

* `fiori-elements.test.ts.snap`: Updated snapshots to include the new commented-out global search OPA test block in generated journey files.

* `listReportUtils.test.ts`: Added comprehensive unit tests for `getSemanticKeyProperties()` (single/multiple keys, label resolution, missing entity set, invalid XML) and `isSemanticKeyInFilterBar()` (full match, partial match, no match, no semantic key annotation).

<details>

**Version:** `1.20.47`

- Correlation ID: `2c26691b-3913-42f6-8439-18c1ac86192a`
</details>


devinea (commented):
## Review: feat(ui5-test-writer): OPA tests for semantic key filters, global search

The PR adds useful OPA test scaffolding for semantic key filter adaptation and global search in the List Report journey. The overall approach is sound and well-tested. A few issues to address:

---

### 🐛 Bug: `filterBarHasSemanticKey` is always `false` when all keys are present in the filter bar

In `getListReportFeatures` (listReportUtils.ts), `missingFromFilterBar` is set to `undefined` when the filter array is empty (all keys matched). Then:

```ts
filterBarHasSemanticKey: semanticKeyProperties?.length ? missingFromFilterBar?.length === 0 : undefined
//                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
// missingFromFilterBar is undefined here → undefined?.length === 0 → false, not true
```

The correct derivation should use the original array before it is collapsed to `undefined`:

```ts
const missingKeys = semanticKeyProperties?.length && filterBarItems.length
    ? semanticKeyProperties.filter((key) => !filterBarItems.includes(key))
    : undefined;

return {
    ...
    semanticKey: {
        semanticKeyProperties,
        missingFromFilterBar: missingKeys?.length ? missingKeys : undefined,
        filterBarHasSemanticKey: semanticKeyProperties?.length ? missingKeys?.length === 0 : undefined
        //                                                        ↑ missingKeys (the raw array) not missingFromFilterBar (already collapsed)
    }
};
```

`filterBarHasSemanticKey` is not currently used in any template, so this is latent rather than a live bug — but it should be fixed (or the field removed) before it causes confusion downstream.

---

### ⚠️ `filterBarHasSemanticKey` is dead — not consumed anywhere

The field is declared in `types.ts`, computed in `getListReportFeatures`, but **never read** by any template or downstream code. If it's intended for a future test, add a TODO comment. If it's not needed yet, remove it to avoid the dead-code lint warning and the logic bug above.

---

### ⚠️ Misleading OPA test name

In `ListReportJourney.js` template (line 42):

```js
opaTest("Check semantic key columns in table", ...
```

The test body opens filter *adaptation*, adds fields to the *filter bar*, and confirms — it never inspects table columns. Suggested name: `"Add semantic key properties to filter bar"` or `"Configure filter bar for semantic key fields"`.

---

### ⚠️ `parse` + `convert` called three times per `getListReportFeatures` invocation

When `metadata` is provided, a single `getListReportFeatures` call now triggers:

1. `safeCheckButtonVisibility` → `checkButtonVisibility` → `convert(parse(metadataXml))`
2. `safeCheckActionButtonStates` → `checkActionButtonStates` → `convert(parse(metadataXml))`
3. `safeGetSemanticKeyProperties` → `getSemanticKeyProperties` → `convert(parse(metadataXml))`

This PR introduces the third parse. Ideally the converted metadata should be computed once and passed down — but this is a pre-existing pattern, so it's acceptable to leave for a follow-up refactor. Worth noting in a TODO.

---

### ℹ️ The `@sap-ux/edmx-parser` question

The PR correctly uses `parse()` from `@sap-ux/edmx-parser` followed by `convert()` from `@sap-ux/annotation-converter`, which is the right approach. This is the same pattern already used in `actionUtils.ts` for button visibility. The `SemanticKey` annotation is accessed via the strongly-typed `ConvertedMetadata` object from `@sap-ux/vocabularies-types`.

The one rough edge is the `as unknown as { value: string }[]` cast on the `SemanticKey` annotation. The actual type from vocabularies-types is `AnnotationTerm<Edm.PropertyPath[]>` where each `PropertyPath` has a `.value: string` field — so the cast is directionally correct, but it bypasses the type system. A typed access would be:

```ts
import type { SemanticKey } from '@sap-ux/vocabularies-types/vocabularies/Common';
// SemanticKey is AnnotationTerm<PropertyPath[]>, each entry has .value: string
const sk = entitySet.entityType.annotations?.Common?.SemanticKey as SemanticKey | undefined;
const propertyNames = sk?.map((entry) => entry.value).filter((v): v is string => typeof v === 'string') ?? [];
```

---

### ✅ What's good

- Changeset present and correctly scoped as `patch`
- New fixture XML (`metadata_filter_bar_semantic_key.xml`) is realistic
- `safeGetSemanticKeyProperties` wrapper follows the existing safe-wrapper pattern from `actionUtils.ts`
- `resolveLabels` parameter correctly aligns property names with the label-based `getFilterFieldNames()` output
- Test coverage for `getSemanticKeyProperties` and `isSemanticKeyInFilterBar` is comprehensive (single/composite key, labels, missing entity set, invalid XML)
- `missingFromFilterBar` short-circuits to `undefined` when `filterBarItems` is empty (no filter bar configured), avoiding false positives

packages/ui5-test-writer/src/utils/listReportUtils.ts
```diff
@@ -116,23 +137,38 @@ export function getListReportFeatures(
...
+        toolBarActions = safeCheckActionButtonStates(metadata, entitySetName, toolbarActions, log);
+    }
+
+    const missingFromFilterBar =
+        semanticKeyProperties?.length && filterBarItems.length
+            ? semanticKeyProperties.filter((key) => !filterBarItems.includes(key))
+            : undefined;
 
     return {
         name: listReportPage.name,
```
devinea, line 170: **Bug: `filterBarHasSemanticKey` is always `false` when all keys are present in the filter bar.**

`missingFromFilterBar` is set to `undefined` when the filtered array is empty (all keys matched), so `undefined?.length === 0` evaluates to `false` instead of `true`.

Fix: compute `filterBarHasSemanticKey` from the raw diff array before collapsing it:
```ts
const missingKeys = semanticKeyProperties?.length && filterBarItems.length
    ? semanticKeyProperties.filter((key) => !filterBarItems.includes(key))
    : undefined;

return {
    semanticKey: {
        semanticKeyProperties,
        missingFromFilterBar: missingKeys?.length ? missingKeys : undefined,
        filterBarHasSemanticKey: semanticKeyProperties?.length ? missingKeys?.length === 0 : undefined
    }
};
```

packages/ui5-test-writer/src/types.ts
```diff
@@ -150,13 +150,18 @@ export type ListReportFeatures = {
     };
     deleteButton?: {
         enabled?: boolean | string;
-        visible: boolean;
+        visible?: boolean;
         dynamicPath?: string;
     };
     filterBarItems?: string[];
```
devinea, line 156: **Dead field: `filterBarHasSemanticKey` is not consumed anywhere.**

This field is declared here and computed in `getListReportFeatures`, but no template or downstream code reads it. Either remove it now, or add a `// TODO:` comment explaining when it will be used, so it's clear this is intentional.

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.js
```diff
@@ -38,6 +38,32 @@ sap.ui.define([
             <%_ }); -%>
         });
 <%_ } -%>
+<%_ if (semanticKey && semanticKey.missingFromFilterBar && semanticKey.missingFromFilterBar.length > 0) { %>
```
devinea, line 41: **Misleading test name.** This test opens filter *adaptation* and adds fields to the *filter bar* — it never inspects table columns. Suggested: `"Add semantic key properties to filter bar"`.

packages/ui5-test-writer/src/utils/listReportUtils.ts
```diff
@@ -319,3 +355,66 @@ function extractItemDescriptions(aggregations: TreeAggregations): string[] {
...
+    const convertedMetadata: ConvertedMetadata = convert(parse(metadataXml));
+    const entitySet = convertedMetadata.entitySets.find((es: EntitySet) => es.name === entitySetName);
+
+    if (!entitySet) {
+        throw new Error(`Entity set '${entitySetName}' not found in metadata`);
+    }
+
+    const semanticKey = entitySet.entityType.annotations?.Common?.SemanticKey;
+    if (!semanticKey) {
+        return [];
```
devinea, line 437: **Unsafe `as unknown as` cast bypasses the type system.** The actual type from `@sap-ux/vocabularies-types` is `AnnotationTerm<PropertyPath[]>`, where each `PropertyPath` already has a typed `.value: string` field. Use the proper type instead:
```ts
import type { SemanticKey } from '@sap-ux/vocabularies-types/vocabularies/Common';

const sk = entitySet.entityType.annotations?.Common?.SemanticKey as SemanticKey | undefined;
const propertyNames = sk?.map((entry) => entry.value).filter((v): v is string => typeof v === 'string') ?? [];
```

packages/ui5-test-writer/src/utils/listReportUtils.ts
```diff
@@ -116,23 +137,38 @@ export function getListReportFeatures(
...
+
+    let buttonVisibility: ReturnType<typeof safeCheckButtonVisibility> | undefined;
+    let semanticKeyProperties: string[] | undefined;
+    let toolBarActions: ReturnType<typeof safeCheckActionButtonStates> = [];
+
+    if (metadata && listReportPage.entitySet) {
+        const entitySetName = listReportPage.entitySet;
+        buttonVisibility = safeCheckButtonVisibility(metadata, entitySetName, log);
+        semanticKeyProperties = safeGetSemanticKeyProperties(metadata, entitySetName, log);
+        toolBarActions = safeCheckActionButtonStates(metadata, entitySetName, toolbarActions, log);
```
devinea: **`parse()` + `convert()` is now called three times per `getListReportFeatures` invocation** (once each for button visibility, action states, and semantic keys). This is a pre-existing pattern, so leaving it as a follow-up refactor is fine — but worth a `// TODO: consolidate parse+convert into a single call` here so it's not forgotten.

devinea (approved):
Thanks for the updates.

longieirl (approved):
Changeset is good.
Overall changes are good.
All comments addressed.

PR #4730, opened 2026-05-26
chore: improve changeset

heimwege:
## What this PR does

  Introduces a custom changelog formatter and a mandatory summary prefix convention for all changeset files, making
  package changelogs significantly easier to read.

  ### Problem

  The default Changesets output is dominated by noise:

  - Every transitive workspace dependency bump gets its own line with a raw git commit hash
  - Real features and fixes are buried and indistinguishable from routine bumps
  - The package name is redundantly repeated in the summary even though it is already in the frontmatter
  - No release date is recorded anywhere in the changelog

  Example of current output:

  ```md
  ## 0.25.42

  ### Patch Changes

  - be5476f: feat(preview-middleware): Display unhandled exceptions from the Controller extension inside the Info
  Center.
  - Updated dependencies [678a08e]
      - @sap-ux/adp-tooling@0.18.129
      - @sap-ux/btp-utils@1.1.15
      - @sap-ux/project-access@1.36.3
  ```

  ### Solution

  A local formatter (`.changeset/changelog-format.cjs`) replaces the default one. Changeset summaries must now start
  with `FEAT:`, `FIX:`, or `BUMP:` - enforced by `pnpm validate:changesets`, which runs as part of `pnpm build` and
  `pnpm cset`.

  Example of new output:

  ```md
  ## 0.25.42

  ### Patch Changes

  #### Release Date

  2026-05-27

  #### Features

  - Display unhandled exceptions from the Controller extension inside the Info Center.
  [[be5476f](https://github.com/SAP/open-ux-tools/commit/be5476f)]

  #### Workspace Updates

  - @sap-ux/adp-tooling 0.18.128 → 0.18.129
  - @sap-ux/btp-utils 1.1.14 → 1.1.15
  - @sap-ux/project-access 1.36.2 → 1.36.3
  ```

  ### Prefix convention

  | Prefix | Section heading | Use for |
  |--------|----------------|---------|
  | `FEAT:` | `#### Features` | New features or behaviour additions |
  | `FIX:` | `#### Bug Fixes` | Bug fixes, security patches, CVE upgrades |
  | `BUMP:` | `#### Dependency Updates` | Third-party dependency version bumps |

  The package name must **not** be included in the summary - it is already declared in the frontmatter. Summaries
  without a valid prefix fall back to `#### Changes` so the entry is never lost, but `pnpm validate:changesets` will
  reject it before it can be merged.

  ```md
  ---
  "@sap-ux/preview-middleware": patch
  ---

  FIX: Handle missing configuration without crashing the preview server
  ```

  A changeset with an invalid or missing prefix will fail validation:

  ```
  🚫 Changeset validation failed:

  ❌ Invalid changeset summary in my-change.md
     Summary: "chore(preview-middleware): upgrade something"
     Must start with FEAT:, FIX:, or BUMP:
  ```

### Renovate bot compatibility

  Renovate-generated changesets (named `dependencies-GH-<number>.md`) use scoped conventional commit style (e.g. `fix(deps): ...`) and are exempt from prefix validation. The formatter automatically normalises them to `#### Dependency Updates` so they render consistently alongside manually written `BUMP:` entries.

  ### Files changed

  | File | Change |
  |------|--------|
  | `.changeset/config.json` | Points `changelog` to the local formatter |
  | `.changeset/changelog-format.cjs` | New custom formatter |
  | `scripts/validate-changesets.js` | Enforces the summary prefix convention |
  | `AGENTS.md` | Updated changeset conventions and examples |


heimwege:
Possible other options to add: 

  - **BREAK**: breaking API changes. Currently a major bump signals this implicitly, but an explicit label in the changelog entry would make it immediately visible without having to check the version number. This is probably the most valuable addition.
  - **DEPRECATE**: marks something as deprecated without removing it. Useful for consumers planning migrations. Sample: preview-middleware `rta.editors` -> deprecate, please use `edirots.rta` instead

.changeset/config.json
```diff
@@ -1,6 +1,6 @@
 {
     "$schema": "https://unpkg.com/@changesets/config@1.6.1/schema.json",
-    "changelog": "@changesets/cli/changelog",
+    "changelog": "./changelog-format.cjs",
```
devinea, line 3: Are loosing anything removing @changesets/cli/changelog 
./changelog-format.cjs doesn't seem to import anything from @changesets.

devinea (commented):
Overall I like the approach. 1 query

.changeset/config.json
heimwege: `@changesets/cli/changelog` is the default formatter that ships with the CLI - it's a simple module that also exports `getReleaseLine` and `getDependencyReleaseLine`, just with the noisy default behaviour (hash prefix, one line per dep).
Our `changelog-format.cjs` reimplements those same two functions from scratch, so there's nothing to import from it.

The `@changesets/cli` package itself is still used - it's still what runs `pnpm changeset version`, `pnpm cset`, etc. We're only replacing its built-in changelog formatter, not the CLI itself.

.changeset/changelog-format.cjs
```diff
@@ -0,0 +1,98 @@
...
+ * @typedef {{ summary: string; commit?: string; releases?: ChangesetRelease[] }} Changeset
+ */
+
+const REPO_URL = 'https://github.com/SAP/open-ux-tools';
+
+/** @type {PrefixEntry[]} */
+const PREFIX_MAP = [
+    { regex: /^FEAT:/i, label: 'Features' },
+    { regex: /^FIX:/i, label: 'Bug Fixes' },
+    { regex: /^BUMP:/i, label: 'Dependency Upgrades' }
```
devinea: Can we add `Chore` and `test` as valid prefix also ?
and/or others from types  at https://www.conventionalcommits.org/en/v1.0.0/

heimwege: Of course we can do, but does CHORE PR's have a changeset at all? If it does changes it should be BUMP right? I only found BREAK and DEPRECATE might be worth adding ([link](https://github.com/SAP/open-ux-tools/pull/4730#issuecomment-4545828190)) but as said I'm open for further additions

devinea (dismissed):
@heimwege Thanks for the updates. 
Should make the change logs clearer in future.

devinea (dismissed):
Re-approving

devinea (dismissed):
re-approving

cianmSAP (dismissed):
Updates look good, nice improvement

Klaus-Keller (commented):
Hi @heimwege, 

I like the new format of the changes and I think to mark pure version bumps of dependencies (`BUMP`) is great and really helps. 

What I'm personally not in favor is the need to classify other changes like `FEAT` or `FIX`. [Semantic Versioning](https://semver.org/) has already a clear definition for changes:

MAJOR version when you make incompatible API changes
MINOR version when you add functionality in a backward compatible manner
PATCH version when you make backward compatible bug fixes

I really like this concise and easy to understand definition and would refrain from adding more classification to it, except marking version bumps as such in patch updates.
A combination `FEAT` in a PATCH version update would even violate semantic versioning and `FIX` in MAJOR version update also feels wrong.

These are my personal opinions and if others are fine with new guidelines to write  changesets I'm ok with it.

heimwege:
> Hi @heimwege, 
> 
> I like the new format of the changes and I think to mark pure version bumps of dependencies (`BUMP`) is great and really helps. 
> 
> What I'm personally not in favor is the need to classify other changes like `FEAT` or `FIX`. [Semantic Versioning](https://semver.org/) has already a clear definition for changes:
> 
> MAJOR version when you make incompatible API changes
> MINOR version when you add functionality in a backward compatible manner
> PATCH version when you make backward compatible bug fixes
> 
> I really like this concise and easy to understand definition and would refrain from adding more classification to it, except marking version bumps as such in patch updates.
> A combination `FEAT` in a PATCH version update would even violate semantic versioning and `FIX` in MAJOR version update also feels wrong.
> 
> These are my personal opinions and if others are fine with new guidelines to write  changesets I'm ok with it.
> 
> 
> 

Thanks @Klaus-Keller, the prefixes are used to define the respective chapter in the changeset, not how the package version is affected. e.g. a bugfix can be a breaking change and thus a major version bump or a minor. Same for features. There is no 1:1 mapping I guess. The developer neede to decide.
Regarding the feature... If you stick to the semantic versioning definition, I think at least 80% of the bumps in open-ux-tools are wrongly classified as patch instead of minor. But that's another story 🙈 
What is your suggestion to improve the changelog? What else to use If not fix, feat or bump to structure it?

Klaus-Keller:
> the prefixes are used to define the respective chapter in the changeset, not how the package version is affected. e.g. a bugfix can be a breaking change and thus a major version bump or a minor. Same for features. There is no 1:1 mapping I guess. The developer neede to decide.

All I wanted to say: I think the information semantic versioning gives a consumer is sufficient. A good enhancement would be the information about pure dependency version bumps, this would be nice to have I think. 

>Regarding the feature... If you stick to the semantic versioning definition, I think at least 80% of the bumps in open-ux-tools are wrongly classified as patch instead of minor. But that's another story 🙈

I hope this is related to the fact, most of the modules have been a `0.x.y` version for a long time. For this case the semantic versioning states:

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

Once major version is >0 we must check for correct semantic versioning.

mikicvi-SAP:
Hi @heimwege, this would break the existing dependency automation flow, where we are generating changesets there with [dependency-changesets-action](https://github.com/mscharley/dependency-changesets-action).

It is being used in [here](https://github.com/SAP/open-ux-tools/blob/main/.github/workflows/renovate-pr-automation.yml#L39-L44) and [here](https://github.com/SAP/open-ux-tools/blob/main/.github/workflows/renovate-pr-automation.yml#L103-L108).

Here is an example of this in action: 

[https://github.com/SAP/open-ux-tools/pull/4561/changes/71b2e005a8f3563cb8fe29e6075e526ce5be7eb7#diff-38411d2a20c5d98d513[…]069a4e7ed2738104ba92b6R39](https://github.com/SAP/open-ux-tools/pull/4561/changes/71b2e005a8f3563cb8fe29e6075e526ce5be7eb7#diff-38411d2a20c5d98d513f8bcc8ef35b26dbee849f8b069a4e7ed2738104ba92b6R39)


Once the renovate deps dashboard is cleaned up (we need to get to updating the dependencies), this will reduce the maintenance effort greatly and allow us to simply review + merge the renovate bumps without needing to pull the pr just to generate the changeset.

What do you think @Klaus-Keller @devinea @cianmSAP

heimwege:
> Hi @heimwege, this would break the existing dependency automation flow, where we are generating changesets there with [dependency-changesets-action](https://github.com/mscharley/dependency-changesets-action).
> 
> It is being used in [here](https://github.com/SAP/open-ux-tools/blob/main/.github/workflows/renovate-pr-automation.yml#L39-L44) and [here](https://github.com/SAP/open-ux-tools/blob/main/.github/workflows/renovate-pr-automation.yml#L103-L108).
> 
> Here is an example of this in action:
> 
> [https://github.com/SAP/open-ux-tools/pull/4561/changes/71b2e005a8f3563cb8fe29e6075e526ce5be7eb7#diff-38411d2a20c5d98d513[…]069a4e7ed2738104ba92b6R39](https://github.com/SAP/open-ux-tools/pull/4561/changes/71b2e005a8f3563cb8fe29e6075e526ce5be7eb7#diff-38411d2a20c5d98d513f8bcc8ef35b26dbee849f8b069a4e7ed2738104ba92b6R39)
> 
> Once the renovate deps dashboard is cleaned up (we need to get to updating the dependencies), this will reduce the maintenance effort greatly and allow us to simply review + merge the renovate bumps without needing to pull the pr just to generate the changeset.
> 
> What do you think @Klaus-Keller @devinea @cianmSAP

We can either change the renovate/dependabot PR title (if that is even possible) to `bump:` or add `fix(deps)`, `chore(deps)`, `chore(dev-deps)` or whatever renovate/depandabot uses as synonym for `bump:`. I think for the changelog all these are bumps. No need to differenciate.

scripts/validate-changesets.mjs
```diff
@@ -40,6 +41,17 @@ function validateChangesets() {
                 );
             }
         }
+
```
heimwege, line 49: Skip validation for changesets from Renovate. Currently no auto-generated changesets for Dependabot

```js
// skip prefix validation for bot-generated dependency changesets
  if (file.match(/^dependencies-GH-\d+\.md$/)) {
      continue;
  }
```

mikicvi-SAP: That is good enough imo, nice and simple. IIRC, we were going to let renovate handle all of the updates, and use dependabot only to create security fixes/bumps, etc.

.changeset/changelog-format.cjs
```diff
@@ -0,0 +1,104 @@
...
+ * @returns {ParsedSummary} Object containing the category label (or empty string if no prefix) and the message text with the prefix removed.
+ */
+function parseSummary(summary) {
+    const trimmed = summary.trim();
+    for (const { regex, label } of PREFIX_MAP) {
+        if (regex.test(trimmed)) {
+            return { label, text: trimmed.replace(regex, '').trim() };
+        }
+    }
+    return { label: 'Changes', text: trimmed };
```
heimwege: Map conventional commit style messages to `bump:`. Only bots can pass the check in validate-changesets.mjs so we can do this w/o checking here

```suggestion
    const trimmed = summary.trim();
    const normalized = trimmed.replace(/^[a-z]+(\([^)]+\))?: /i, 'BUMP: ');
    for (const { regex, label } of PREFIX_MAP) {
        if (regex.test(normalized)) {
            return { label, text: normalized.replace(regex, '').trim() };
        }
    }
    return { label: 'Changes', text: normalized };
```

heimwege: done

mikicvi-SAP (approved):
Cheers @heimwege 
- allowing bots changesets to function as before lg, rest is a nice upgrade and makes sense now that modules went to v1

devinea (approved):
Thanks @heimwege 
Seems like a useful addition to the changesets

PR #4735, opened 2026-05-27
fix(ui-components): fix flaky UITranslationInput callout dismiss test

longieirl:
## Problem

The `<UITranslationInput /> › Close callout › Click outside of callout` test was intermittently failing on CI with:

```
Expected: 0
Received: 1
```

The callout remained visible after clicking outside, but only on slower CI runners — making it a flaky, non-deterministic failure.

## Root Cause

Two separate bugs combined to cause the flake:

### 1. Component bug — `onDismiss` used toggle semantics (primary)

`UITranslationButton.tsx` wired the Fluent UI Callout `onDismiss` to `onToggleCallout`:

```tsx
// Before — wrong
onDismiss={() => onToggleCallout()}
```

`onToggleCallout` calls `setCalloutVisible(!calloutVisible)`. The `onDismiss` handler should always _close_ the callout, not toggle it. Using toggle here had two failure modes:
- A stale `calloutVisible` closure (where it captured `false`) would _re-open_ the callout on dismiss
- In the `SuggestValueType.Existing` branch, it would incorrectly call `onShowExistingEntry` when the user simply clicked outside

### 2. Test fragility — timing-based wait + synthetic event (secondary)

```tsx
// Before — fragile
await new Promise((resolve) => setTimeout(resolve, 1));
fireEvent.click(document.body);
```

Fluent UI v8 `Callout` defers registration of its document-level `click` dismiss listener via an internal `setTimeout(..., 0)` after mount. The 1ms wait was not reliably enough on slow CI runners to guarantee that listener was registered before the click fired.

Additionally, `fireEvent.click` dispatches a synthetic RTL event which does not reliably reach Fluent's native `addEventListener('click', ...)` dismiss handler — `document.body.click()` (native DOM dispatch) is the correct approach, matching the pattern already used in `CalloutCollisionTransform.test.tsx`.

## Fix

**`UITranslationButton.tsx`** — replace toggle with explicit close:
```tsx
// After — correct
onDismiss={onCancel}
```

**`UITranslationInput.test.tsx`** — replace timing-based wait with `waitFor` + native click:
```tsx
// After — deterministic
await waitFor(() => {
    document.body.click();
    expect(document.querySelectorAll(selectors.callout).length).toEqual(0);
});
```

`waitFor` polls until the Fluent dismiss listener is registered and the state update propagates — no timing bets.

## Test plan

- [ ] `pnpm --filter @sap-ux/ui-components test` passes (557 tests, 0 failures)
- [ ] Previously flaky test `Click outside of callout` now passes deterministically
- [ ] No regressions in other `UITranslationInput` test cases (Cancel callout, Accept callout, etc.)

### DIFF
--- a/packages/ui-components/src/components/UITranslationInput/UITranslationButton.tsx
+++ b/packages/ui-components/src/components/UITranslationInput/UITranslationButton.tsx
@@ -119,7 +119,7 @@ export const UITranslationButton = <T extends TranslationEntry>(props: UITransla
                     isBeakVisible={false}
                     setInitialFocus={true}
                     className={getCalloutClassNames(invertedCalloutTheme)}
-                    onDismiss={() => onToggleCallout()}
+                    onDismiss={onCancel}
                     contentPadding={UICalloutContentPadding.Standard}>
                     <div className="ui-translatable__message">
                         {suggestion.message}
--- a/.changeset/fix-translation-button-callout-dismiss.md
+++ b/.changeset/fix-translation-button-callout-dismiss.md
@@ -3,7 +3,3 @@
 ---
 
 fix(UITranslationButton): use onCancel for callout onDismiss instead of onToggleCallout
-
-Dismiss should close the callout, not toggle it. Using onToggleCallout on onDismiss
-could re-open the callout via stale closure and incorrectly trigger onShowExistingEntry
-when the user clicks outside.
--- a/.changeset/fix-translation-button-callout-dismiss.md
+++ b/.changeset/fix-translation-button-callout-dismiss.md
@@ -1,0 +1,9 @@
+---
+"@sap-ux/ui-components": patch
+---
+
+fix(UITranslationButton): use onCancel for callout onDismiss instead of onToggleCallout
+
+Dismiss should close the callout, not toggle it. Using onToggleCallout on onDismiss
+could re-open the callout via stale closure and incorrectly trigger onShowExistingEntry
+when the user clicks outside.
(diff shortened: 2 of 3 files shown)


devinea (approved):
change lgtm. 
@815are / @eouin Please review,

eouin (approved):
LGTM 👍

packages/ui-components/src/components/UITranslationInput/UITranslationButton.tsx
```diff
@@ -119,7 +119,7 @@ export const UITranslationButton = <T extends TranslationEntry>(props: UITransla
                     isBeakVisible={false}
                     setInitialFocus={true}
                     className={getCalloutClassNames(invertedCalloutTheme)}
-                    onDismiss={() => onToggleCallout()}
+                    onDismiss={onCancel}
```
815are, line 122: I think with change ` onShowExistingEntry?.(suggestion.suggest.entry);` will never be executed - https://github.com/SAP/open-ux-tools/blob/3cb0c68417ffcdb6f0ea2fdc98ce4c5b4b5f586f/packages/ui-components/src/components/UITranslationInput/UITranslationButton.tsx#L79

815are: seems ok - `onShowExistingEntry` supposed to be used when click on right button and should not be triggered when callout closed

815are (approved):

PR #1650, opened 2024-02-07
feat: add ExtensionLogger for convenient VSCode extension logging

Klaus-Keller:
## Description

Add `ExtensionLogger` to package `@sap-ux/logger` for convenient logging in VSCode extensions.
Also updated dependencies and fixed minor linting issues.

PR #4740, opened 2026-05-29
fix(sap-systems-ext): remove unsafe Logger/ExtensionLogger casts in SystemsLogger

longieirl:
## Summary

- `SystemsLogger` in `sap-systems-ext` cast between `ExtensionLogger` and `Logger` using `as` — TypeScript 5.9 rejects this with TS2352 because `ExtensionLogger` overrides `error/warn/info/debug` with `message: string` while `Logger` interface requires `message: string | object`, so the types don't sufficiently overlap, breaking `nx run sap-ux-sap-systems-ext:build`
- Fix is in the consumer, not the library: type `_logger` as `Logger` (eliminates cast on getter/setter), use `instanceof ExtensionLogger` guard in `show()` instead of unsafe cast
- Root defect introduced in PR #1650 (`2e0b1a6df`) ~2 years ago; surfaced by TypeScript 5.9 stricter TS2352 checking

## Test Plan

- [x] `pnpm --filter sap-ux-sap-systems-ext exec tsc --noEmit` — no type errors
- [x] `pnpm --filter @sap-ux/logger test` — 42 tests pass, no logger changes
- [ ] `nx run sap-ux-sap-systems-ext:build` — verify full build passes


longieirl:
FYI, this fixes a TS2352 build failure in `sap-systems-ext` caused by incompatible type casts between `ExtensionLogger` (which overrides `error/warn/info/debug` with `message: string`) and the `Logger` interface (`message: string | object`). TypeScript 5.9 surfaced this latent issue. Fix is in the consumer (`SystemsLogger`) rather than changing the logger library contract.

cianmSAP (approved):
Hopefully helps the tests

kjose90 (approved):
looks good 👍

PR #4565, opened 2026-04-16
ESM migration

cianmSAP:
## ESM Migration: Configuration Consolidation and CI Fixes                                                                                  
                                                                                            
### Overview                                                                                                                                 
                                                                                                                             
This PR consolidates TypeScript configuration files, fixes cross-platform build issues, and resolves CI test failures as part of the ESM migration.                                                                                                                 
                                                                                                                                           
#### Changes                                                                                                                                  
                                                                               
##### TypeScript Configuration Consolidation                                                                                                   
                                                                                                                             
  - Merged tsconfig-esm.json into tsconfig.json as the canonical configuration                                                             
  - Renamed old tsconfig.json to tsconfig-cjs.json for CJS-only packages                                                     
  - Updated all 95+ packages to reference the new tsconfig.json                                                                            
  - Added isolatedModules: false override for fe-fpm-writer (uses ambient const enums)                                                     
  - Updated sap-systems-ext to extend tsconfig-cjs.json with explicit CommonJS overrides                                                   
                                                                                                                                           
##### Jest Configuration Improvements                                                                                                          
                                                                                                                                           
  - Removed global @sap-devx/yeoman-ui-types moduleNameMapper from jest.base.mjs                                                           
  - Scoped yeoman-ui-types mapping to only the 22 packages that need it                                                      
  - Made setupFiles path dynamic to work at any package depth                                                                              
  - Applied spread operator pattern to all @sap-ux/store mocks (31 files)                                                                  
                                                                                                                                           
##### ESLint Configuration Fixes                                                                                                               
                                                                                                                                           
  - Fixed leading-space globs in ignore patterns (esbuild*.js, esbuild*.mjs)                                                               
  - Consolidated duplicate jsdoc/require-jsdoc keys using conditional logic                                                  
  - Fixed import/enforce-node-protocol-usage rule to use correct format                                                                    
                                                                                                                                           
 ##### Test Timeout Increases                                                                                                                   
                                                                                                                                           
Increased Jest timeout to 60 seconds (from 30s) for heavy yeoman-test integration tests in ESM mode:                                     
  - fiori-app-sub-generator/test/unit/fiori-app-generator/fioriAppGenerator-lifecycle1.test.ts                               
  - fiori-app-sub-generator/test/unit/fiori-app-generator/fioriAppGenerator-lifecycle2.test.ts                                             
  - cf-deploy-config-sub-generator/test/app.test.ts                                                                          
[...]

PR #4746, opened 2026-05-31
fix(fiori-mcp-server): fix MCP server crashing on startup when installed via npx

devinea:
## Summary
Fix regression introduced in [SAP/open-ux-tools#4565](https://github.com/SAP/open-ux-tools/pull/4565)

Two separate bugs prevented the MCP server from starting when installed via `npx @sap-ux/fiori-mcp-server`:

**Bug 1 — Wrong `--external` flags caused `ERR_MODULE_NOT_FOUND`**

`@sap-ux/odata-annotation-core`, `@sap-ux/odata-annotation-core-types`, and `@sap-ux/text-document-utils` were marked as `--external` in esbuild (left as unresolved runtime imports) but were not listed in `dependencies`, so they were missing from the npm install. These packages have no bundling obstacles — they are plain TypeScript ESM libraries — so the fix is to remove them from `--external` and let esbuild inline them.

Packages that legitimately stay external: `@lancedb/lancedb` (native addon), `@sap-ux/store` (native keytar bindings), `@xenova/transformers` (large ML library with dynamic model loading), `@sap-ux/fiori-docs-embeddings` (optional — server handles its absence gracefully).

**Bug 2 — Runtime `require(package.json)` resolved to wrong path after bundling**

`server.ts` and `src/tools/schemas/appgen-config-schema-props.ts` used `createRequire(import.meta.url)` with relative paths to load `package.json` at runtime. Those paths were correct relative to the source files, but after esbuild flattens everything into `dist/index.js`, `createRequire` resolves relative to `dist/index.js` — putting the paths outside the installed package and causing `Cannot find module '../../../package.json'`.

Fixed by injecting `PACKAGE_NAME` and `PACKAGE_VERSION` as string literals at build time via esbuild `--define`, eliminating the runtime `require` entirely.

**Refactor — extracted esbuild invocation into `esbuild.mjs`**

The long inline shell command in `package.json` is replaced by `esbuild.mjs`, a proper Node script with comments explaining `external`, `define`, and `banner`.

## Test plan

- [ ] `pnpm --filter @sap-ux/fiori-mcp-server build` succeeds
- [ ] `dist/index.js` contains no `import ... from "@sap-ux/text-document-utils"` or `odata-annotation-core` static imports
- [ ] `dist/index.js` contains no relative `require(...package.json)` calls
- [ ] `node dist/index.js` starts without throwing `ERR_MODULE_NOT_FOUND` or `Cannot find module`


devinea:
merging to fix broken mcp release.

PR #4747, opened 2026-06-01
fix: loadModuleFromProject fails on Window with error Only URLs with …

815are:
Scenario:
1. use Windows OS
2. Use app with `"@sap/ux-specification` devDependency
3. Call `loadModuleFromProject` and pass path to app and `"@sap/ux-specification`

`loadModuleFromProject` throws error on `Windows`: `Only URLs with a scheme in: file, data, node, and electron are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'`

### DIFF
--- a/.changeset/breezy-paws-follow.md
+++ b/.changeset/breezy-paws-follow.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/project-access': patch
+---
+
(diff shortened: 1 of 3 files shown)


devinea (approved):
Thanks @815are

heimwege (approved):
lgtm

PR #4753, opened 2026-06-02
fix(node-js-utils): use yo --version for Yeoman version checks

cianmSAP:
- Switched from `npm list -g yo` to `yo --version` to validate the actual Yeoman executable available on the user's PATH, rather than relying on npm's global package registry.
- Improves compatibility across package managers, allowing detection of yo installations from npm, Yarn, pnpm, etc.
- Reduces CI/environment-specific failures by removing the dependency on npm global package metadata and configuration.
- Simplifies version detection logic by avoiding npm JSON parsing and directly querying the installed yo version from the executable itself.


lfindlaysap (approved):

longieirl (approved):
Changeset is good.
Tests are good.
Change is required based on support issues reported.

PR #4759, opened 2026-06-02
fix(project-access): serialize concurrent module-cache installs and bust ESM import cache on retry

devinea:
## Summary

Fixes a race condition where parallel callers of `getModule()` (e.g. Jest workers across packages) clobbered each other's installs in the shared module cache (`~/.fioritools/module-cache/`). Combined with Node's ESM loader permanently caching failed `import()` URLs, this produced flaky CI runs that could not self-recover via the existing retry branch.

- `getModule()` now acquires a cross-process file lock (`proper-lockfile`) on the **parent** (`scope/module`) directory before checking for a completed install. The parent is stable across reinstalls — the version subdir is wiped + recreated during recovery, so locking it would orphan the lock artefact.
- The retry branch passes a cache-buster query string (`?v=<pid>-<timestamp>`) to `loadModuleFromProject()` so the freshly reinstalled module is imported with a new URL instead of the rejected one.
- Lockfile failures (e.g. unsupported filesystems) degrade gracefully to a no-op release rather than blocking — preserves the prior unsynchronized behaviour as a safety net.

## Root cause (ESM-specific)

Pre-ESM (CommonJS), a failed `require()` could be retried after eviction from `require.cache`. Under ESM, Node's loader caches the rejection of an `import(url)` keyed by URL, with **no public eviction API**. Any subsequent `import()` of the same URL re-throws the original failure even after the on-disk module is fixed. The retry path therefore needs both:
1. a *fresh* import URL (cache-buster query string), and
2. prevention of the partial-write that triggers the failure in the first place (file lock).

## Test plan

- [x] `pnpm --filter @sap-ux/project-access test` — 423/423 pass, including:
  - new test asserting the lock targets the parent directory
  - new test exercising the `getModule` retry branch end-to-end (lock acquire → install → load fails → reinstall → load with cacheBuster)
  - assertions that `proper-lockfile.lock` is called with the parent dir and `releaseLock` runs in `finally`
- [x] `pnpm --filter @sap-ux/project-access lint` — 0 errors
- [x] `pnpm --filter @sap-ux/project-access build`
- [x] `pnpm validate:changesets`
- [x] Coverage for `module-loader.ts` at 97% (uncovered line is the lockfile-failure fallback)

## Review feedback addressed

The initial revision was reviewed by `hyperspace-insights[bot]`; the substantive points were addressed in commit `e0d5d80`:

| Point | Resolution |
|---|---|
| Lock acquired on a directory that gets `rm -rf`'d during reinstall | Lock target moved to the stable parent (`scope/module`) directory |
| Tests never exercised the lock (silently fell through `acquireModuleLock`'s catch) | Added `proper-lockfile` mock; tests now assert lock is called with parent dir and released |
| New retry-path test didn't go through `getModule`'s retry branch | Test restructured to drive through `getModule` end-to-end with first load forced to fail |
| Bot also flagged a pre-existing `existsSync(modulePackageLockPath)` ordering issue | Out of scope — the bot itself notes this is pre-existing and unchanged by this PR |

The bot also asked the retry-path test to assert `?v=<token>` substring on the import URL. This is not feasible under Jest: Jest's resolver strips query strings before module resolution, so the only observable URL is the unqueried path. The test now drives through the retry branch end-to-end and verifies its observable side-effects (install command, `pathToFileURL` invocation count, lock release in `finally`); the `?v=` behaviour is what makes the retry *possible* under real Node ESM (which keys its module map by URL), and is documented inline in `module-loader.ts`.

## Notes

- `proper-lockfile@4.1.2` was already a transitive dep — promoted to a direct runtime dep with `@types/proper-lockfile@4.1.4`.
- Pre-commit `pnpm audit` hook was bypassed with `--no-verify` per maintainer approval; the 3 reported vulnerabilities are pre-existing transitive issues on `main` (yeoman-environment, tmp, protobufjs) unrelated to this change.


devinea:
Does seem to be needed. closing.

PR #4760, opened 2026-06-02
fix: esm follow-up unit test adjustments

heimwege:
Follow-up to #4565

## Summary

  This PR fixes TypeScript type errors in test files across the monorepo that were surfaced by _<your IDE name here>_'s type checker (which checks all `.ts` files regardless of `tsconfig include`). All changes are confined to `test/` directories and test infrastructure - no source code was modified.

  ### Fix 1–3: Static & dynamic import extensions
  NodeNext module resolution requires explicit `.js` extensions on all relative imports. Applied via Node scripts across all 86 ESM packages:
  - **Static imports** (`from '...'`) - 1062 files updated, including multi-line imports where `from` is not at the start of a line (203 files, 265 occurrences)
  - **Dynamic imports** (`await import('...')`) - 328 files updated
  - Excluded `expected-output/`, `fixtures/`, `__snapshots__/`, `test-output/` directories to avoid modifying generated
  snapshot fixtures

  ### Fix 4: `import { jest } from '@jest/globals'`
  `jest.unstable_mockModule` is only typed in `@jest/globals`, not in the global `@types/jest` - without the explicit import the type checker errors on every call. Added to 35 files across 6 packages. Added `@jest/globals` to `devDependencies` in `odata-service-writer` and `ui-components`.

  ### Fix 5: `jest.fn<ReturnType, [Args]>()` two-type-arg signature
  Jest 30 dropped the two-argument generic form. Migrated 3 occurrences in `preview-middleware-client` to the new single-arg form: `jest.fn<(handler: ActionHandler) => void>()`.

  ### Fix 6: `mockImplementation()` with no arguments
  Jest 30 requires an argument. Fixed 14 occurrences across 6 packages:
  - `void`-return spies: `mockImplementation(() => {})`
  - `never`-return spies (`process.exit`, `Command.prototype.error`, VSCode message APIs): `mockImplementation((() => {}) as unknown as () => never)`

  ### Fix 7: Untyped `jest.fn()` declarations
  In Jest 30, `jest.fn()` without a type annotation infers `jest.Mock<UnknownFunction>`, causing TS2345 errors on `.mockResolvedValueOnce()`, `.mockReturnValue()`, and `.mockImplementation()` with typed callbacks. Applied across 243 files in 55 packages:
  - **341 mocks** typed with `jest.fn<typeof realModule.exportedFn>()` by parsing `jest.unstable_mockModule` registration bodies to derive the real function type
  - **499 mocks** cast to `jest.fn() as jest.Mock` where the type could not be automatically resolved (inline objects, 3rd-party libs without pre-imports)
  - **228 `mockImplementation` callbacks** had redundant explicit type annotations removed (inferred from the mock's declared type)
  - **Overloaded Node/3rd-party APIs** (`ejs.renderFile`, `fs.readFileSync`, `child_process.exec`, etc.) kept as `jest.fn<typeof ...>()` but with `as unknown as typeof ...` cast on their `mockImplementation` callbacks to satisfy TypeScript's overload resolution

  ### Fix 8: `tsconfig.json` / `test/tsconfig.json` split
  Two packages had test files pulled into the build type-check:

  **`preview-middleware-client`**: Removed `test` from the root `tsconfig.json` `include` and created
  `test/tsconfig.json` extending the root (inheriting `paths` aliases for `mock/*` and `open/ux/preview/client/*`). Tests are now type-checked via `test/tsconfig.json` only - the build type-check no longer processes test files.

  **`ui-components`**: Created `test/tsconfig.json` with `allowImportingTsExtensions: true` and `noEmit: true` to allow
  `import('../../../src/index.tsx')` in test files. Added `include: ["../test", "../src"]` so IntelliJ associates test files with the correct tsconfig.


devinea (approved):
Thanks @heimwege  for follow-up.

cianmSAP (approved):
Updates look good
Thanks for the cleanup @heimwege

PR #4765, opened 2026-06-03
feat: update visual filter skill for non cap projects

vinayhospete:
internal issue: 36980


.changeset/dark-knives-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/fiori-mcp-server': patch
+---
+
+update visual filter skill
```
heimwege:
```suggestion
FEAT: update visual filter skill
```

heimwege (commented):
see comment and:

1. Chart type table contradicts the detail files.
SKILL.md marks #Column and #Donut as "❌ Not supported" for OData V4, but cap-implementation.md and rap-implementation.md both list #Column and #Donut as valid chart types in their respective ChartType option tables. This is a direct contradiction — if they're genuinely unsupported, remove them from the detail files; if they're partially supported, the table in SKILL.md needs nuance rather than a flat ❌.
2. RAP chart type table says Donut is "OData V2 only, not for draft" (rap-implementation.md:1289) — this is a different reason than "OData V4 not supported." These need to be reconciled and stated consistently.
3. #Column support claim needs a source.
The OData V4 visual filter limitation is asserted without explanation. If this is a known Fiori Elements constraint, a brief "why" or a link to SAP documentation would strengthen it. Currently an LLM acting on this skill might refuse a user asking for a column chart when it would actually work.
4. Hardcoded service namespace in complete XML example.
In rap-implementation.md the complete example uses `/sap/opu/odata4/sap/your_service/srvd/sap/your_service/0001/$metadata` as the reference URI — this is a URL the AI agent would need to replace, but it's buried in XML that's easy to copy verbatim. Consider making the placeholder more visually obvious (e.g., <YOUR_SERVICE_URL>) or adding a note directly above the XML block.
  5. provider contract transactional_query listed as REQUIRED for aggregation (rap-implementation.md:1159) but the known limitation section says aggregations are not supported on draft-enabled entities, which use transactional_query. This creates a circular dependency — the doc says you need transactional_query to get aggregation, then says aggregation doesn't work with draft/transactional entities. This needs clarification: does it work on non-draft transactional views, or is a read-only/analytical contract required?
  6. AggregationMethod casing: cap-implementation.md uses lowercase 'sum', 'average' etc. in the CDS annotation examples. The original SKILL.md also used lowercase. This appears correct for CDS, but it's worth verifying — some SAP docs show mixed casing.
  7. Duplicate manifest JSON blocks. The manifest configuration is reproduced identically in three places: SKILL.md, cap-implementation.md, and rap-implementation.md. Since it's identical for both backends, this means three places to update if the schema ever changes. The main SKILL.md already notes "(same for both CAP and RAP)" — consider removing the duplicate blocks from the reference files and just linking back to SKILL.md#manifest-configuration.
  8. SKILL.md has a dangling troubleshooting section.
 There are two separate troubleshooting sections: "Common Troubleshooting" (around line 418) and "Common Mistakes to Avoid" (later). The "Common Troubleshooting" section ends with "Backend-Specific Issues" links, then a second ### Backend-Specific Troubleshooting: heading appears inside the "Common Mistakes" section (after Mistake 4). This is a structural artifact — the second occurrence should be removed or merged.
  9. SKILL.md links to MISTAKE 4 anchor (#mistake-4-using-dynamicmeasures-in-rap-projects) inside the same document.
  This is reasonable for humans but MCP skill consumers read the raw Markdown — test that the anchor renders correctly when the skill is served to an LLM.
  10. argument-hint change from a single field name to 'Filter field name, dimension field, measure field, chart type (Bar/Line)' is more informative but may confuse an LLM that tries to pass all four as a single argument string. Consider whether this is a real invocation signature or just prose guidance for the user.

.changeset/dark-knives-tie.md
vinayhospete: done

packages/fiori-mcp-server/skills/sap-fiori-add-visual-filter/SKILL.md
```diff
@@ -1,402 +1,221 @@
...
-- `AggregatableProperties`: Numeric fields that can be summed, averaged, etc. (measures)
-- `GroupableProperties`: Fields to group by (dimensions)
-- Both properties are required for visual filters to work
+## CAP Implementation
 
-### Step 2: Define Aggregated Property in Annotations
-In your annotations file (e.g., `app/your-app/annotations.cds`), define the aggregated property:
+### Enable Aggregation (MANDATORY)
 
+### Enable Aggregation (MANDATORY)
```
heimwege: header is doubled

```suggestion
### Enable Aggregation (MANDATORY)
```

packages/fiori-mcp-server/skills/sap-fiori-add-visual-filter/SKILL.md
```diff
@@ -1,402 +1,221 @@
...
-        },
-        Common.Label : 'Field Label',
-    )
+Category @Common.ValueList #visualFilter: {
+  $Type: 'Common.ValueListType',
+  CollectionPath: 'EntityName',
+  Parameters: [
+    { $Type: 'Common.ValueListParameterInOut', LocalDataProperty: Category, ValueListProperty: 'Category' }
+  ],
+  PresentationVariantQualifier: 'visualFilter'
```
heimwege, line 79: Missing UI.PresentationVariant in CAP path

The old skill had a dedicated Step 4 for UI.PresentationVariant. The new CAP section defines UI.Chart and Common.ValueList but omits UI.PresentationVariant entirely. However the ValueList still references PresentationVariantQualifier: 'visualFilter' - which means the implementation will fail at runtime because the PresentationVariant that the qualifier points to is never defined.

The RAP section also has no PresentationVariant - it goes directly from Chart to ValueList referencing PresentationVariantQualifier: "visualFilter". This is the same gap.

.changeset/dark-knives-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/fiori-mcp-server': patch
+---
+
+FEAT: update visual filter skill
```
heimwege: The changeset should ideally describe what changed for consumers - e.g. FEAT: update visual filter skill with ABAP RAP support.
Also if this is a `feat` and not a fix we should adjust the `minor` version because we're now on major version 1 no more the "on major version 0 we can do what we want excuse"
@devinea your thoughts?

packages/fiori-mcp-server/skills/sap-fiori-add-visual-filter/SKILL.md
```diff
@@ -1,402 +1,221 @@
...
+```bash
+npm run start-mock # Needs metadata refresh
 
-### Chart Not Showing Data
-- Verify there is data in the entity
-- Check that the aggregated property name matches in both definition and usage
-- Ensure the dimension field has multiple distinct values
+npm start          # No refresh needed - fetches metadata from live backend at runtime
+```
+- Consult fiori mcp server if available on how to refresh metadata for backend systems in case of RAP
```
heimwege: Is `fiori mcp server` precise enough or do we need the correct package name, e.g. `fiori mcp server (@sap-ux/fiori-mcp)`?

packages/fiori-mcp-server/skills/sap-fiori-add-visual-filter/SKILL.md
```diff
@@ -1,402 +1,221 @@
 ---
 name: sap-fiori-add-visual-filter
-description: 'Add visual filters with charts to SAP Fiori Elements value help dialogs. Use for: displaying aggregated data in filter fields, adding bar/column/line/donut charts to value help, configuring Analytics.AggregatedProperty with sum/average/min/max, setting up @Aggregation.ApplySupported, configuring manifest.json for visual filters, implementing OData V4 aggregation in CAP projects, enhancing List Report filter bars with visual analytics.'
-argument-hint: 'field name to add visual filter to (e.g., "Category", "Status", "Region")'
+description: Add visual filters (chart-based) to SAP Fiori Elements filter bar/value help using CAP or ABAP RAP.
+argument-hint: field name (e.g., Category, Status)
 metadata:
   author: sap-fiori-tools
-  version: "0.0.3"
+  version: "0.0.5"
```
heimwege: why skipping `0.0.4`?

heimwege (commented):
see comments

packages/fiori-mcp-server/skills/sap-fiori-add-visual-filter/SKILL.md
vinayhospete: removed

packages/fiori-mcp-server/skills/sap-fiori-add-visual-filter/SKILL.md
vinayhospete: Yes. PV is missing, although copilot added presentation variant during implementaiton. I will add it now

packages/fiori-mcp-server/skills/sap-fiori-add-visual-filter/SKILL.md
vinayhospete: done

packages/fiori-mcp-server/skills/sap-fiori-add-visual-filter/SKILL.md
vinayhospete: Above sentence was not sufficient, so i modified to include sap/cloud system. then fiori mcp fetched expected result.
Screenshot 2026-06-10 at 7 48 22 PM

packages/fiori-mcp-server/skills/sap-fiori-add-visual-filter/SKILL.md
vinayhospete: added.

.changeset/dark-knives-tie.md
devinea: @heimwege I don't have a strong view on this. Strictly speaking yes a minor maybe more appropriate for a new feature.

vinayhospete: ok. updated to minor

heimwege (approved):
- changeset ok
- review comments addressed
- SONAR happy
- did NOT test manually

devinea (approved):

PR #4766, opened 2026-06-03
fix: (Adaptation Editor) FL Variant changes do not appear in the unsaved changes list (history panel).

avasilev-sap:
#4761

When we create variant changes they do not appear in the history panel as unsaved changes. There are two separate issues which caused the actual problem.
1. The [getPrepareChange()](https://github.com/UI5/openui5/blob/9811555f0228b74f37773ef7a0d1553bde8cbdba/src/sap.ui.rta/src/sap/ui/rta/command/FlexCommand.js#L133) function for a flex command can return a single change or an array of changes. In the entire codebase, this is not addressed, which leads to exceptions which are swallowed in try/catch statements and just logged.
2. When we create the list of pending(in the UI labeled as unsaved changes) changes, we ignore changes for which the command changeType is undefined which is wrong. We should use the flex change type, not the command change type, provided by the UI framework and we should not ignore changes for which the type is undefined, otherwise they would not appear in the history.

This change address the possibility of getPrepareChange() returning an array of changes in all places inside the codebase.
The filtering of pending changes is changed - we still display changes for which the changeType is undefined also we use the [change.getChangeType() ](https://github.com/UI5/openui5/blob/9811555f0228b74f37773ef7a0d1553bde8cbdba/src/sap.ui.fl/src/sap/ui/fl/apply/_internal/flexObjects/FlexObject.js#L324)method instead of the previous logic which uses the command changeType or assumes the command has only one change and returns the changeType for that change.

In addition this change removes the deprecated [getDefinition()](https://github.com/UI5/openui5/blob/9811555f0228b74f37773ef7a0d1553bde8cbdba/src/sap.ui.fl/src/sap/ui/fl/apply/_internal/flexObjects/FlexObject.js#L407) function. The getDefinition() is actually a wrapper of the [convertToFileContent()](https://github.com/UI5/openui5/blob/9811555f0228b74f37773ef7a0d1553bde8cbdba/src/sap.ui.fl/src/sap/ui/fl/apply/_internal/flexObjects/FlexObject.js#L467). The convertToFileContent() function was introduced when the FlexObject class was introduced:
https://github.com/UI5/openui5/commit/2ab31247a851b6dac9ac26596ff74ebd3dcdf3c5
it is save to use that function instead because it is preserved in version 2.x of the UI5 framework:
https://github.com/UI5/openui5/blob/main/src/sap.ui.fl/src/sap/ui/fl/apply/_internal/flexObjects/FlexObject.js#L450C23-L450C43

For a very old versions of the UI5 framework, changes are represented by the Change class which has its own implementation of the [getDefinition()](https://github.com/UI5/openui5/blob/rel-1.96/src/sap.ui.fl/src/sap/ui/fl/Change.js#L824) function so we still need to fallback to that function in this case. The Change class has no convertToFileContent() method defined.

**Testing done:**
* Manually verified with variant change containing addFilter, default variant, fragment, controller extensions changes.
* Run staging job with ADP/CPE e2e tests.

Before:
fl-variant-before

After:
fl-variant-after


packages/preview-middleware-client/types/sap.ui.rta.d.ts
```diff
@@ -60,7 +60,9 @@ declare module 'sap/ui/rta/command/FlexCommand' {
             };
             setModuleName(moduleName: string): void;
         };
-        getPreparedChange(): Change<ChangeContentType>;
+        // avasilev: This may return heterogeneous collection so it is wrong to 
+        // use the generic type when the method returns an array of changes.
```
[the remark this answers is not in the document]
avasilev-sap: fixed in [61706a6](https://github.com/SAP/open-ux-tools/pull/4766/commits/61706a6d26b7b929236e915d02feb7fe619c7f90)

packages/preview-middleware-client/src/adp/utils.ts
```diff
@@ -106,16 +107,11 @@ export function getNestedProperty(obj: object, path: string): unknown {
...
-    const change = command.getPreparedChange()?.getDefinition?.();
-    if (!change) {
-        return false;
-    }
-
-    const nestedProperty = getNestedProperty(change, propertyPath);
-    return typeof nestedProperty === 'string' ? nestedProperty.includes(propertyValue) : false;
+    return getFlexChangeList(command).some((change) => {
+        const changeDefinition = change.convertToFileContent?.();
+        const nestedProperty = getNestedProperty(changeDefinition, propertyPath);
```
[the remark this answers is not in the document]
avasilev-sap: Redundant comment, we introduce a helper function getChangeDefinition() which returns the definition object

packages/preview-middleware-client/src/adp/controllers/AddTableColumnFragments.controller.ts
```diff
@@ -295,8 +296,10 @@ export default class AddTableColumnFragments extends BaseDialog<AddTableColumnsF
 
             const templateName =
                 fragment.targetAggregation === COLUMNS_AGGREGATION ? `V2_SMART_TABLE_COLUMN` : 'V2_SMART_TABLE_CELL';
-            const preparedChange = command.getPreparedChange();
-            setAdditionalChangeInfoForChangeFile(preparedChange.getDefinition().fileName, { templateName });
+            const preparedChange =
+                command.getPreparedChange() as unknown as Change<AddTableCellFragmentChangeContentType>;
```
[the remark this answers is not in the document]
avasilev-sap: For a command with changeType: addXML, the  getPreparedChange() always return a change object, so directly cast to a change object.

packages/preview-middleware-client/src/cpe/changes/service.ts
```diff
@@ -522,39 +527,42 @@ export class ChangeService extends EventTarget {
...
+                const { fileName } = changeDefinition;
+
+                const handler = GENERIC_CHANGE_HANDLER[
+                    changeType as ChangeType
+                ] as unknown as ChangeHandler<GenericChange>;
+                const isActive = index >= inactiveCommandCount;
+
+                return handler
+                    ? this.buildGenericChange(changeDefinition, changeType, handler, isActive, textBundle)
+                    : this.buildFallbackChange(changeType, selectorId, isActive, fileName);
```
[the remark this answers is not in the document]
avasilev-sap: For a FLVariant the ui5 framework returns changeType undefined the command.getChangeType is undefined. We still want to treat the flvariant object as a change so we do not filter out the fl variant change. The PendingChange type in the editor defines the changeType as string not string | undefined.

Should we refactor PendingChange change types or leave as-is?

packages/preview-middleware-client/src/utils/additional-change-info.ts
```diff
@@ -71,3 +69,53 @@ export function getAdditionalChangeInfo(change: Change): AdditionalChangeInfo {
...
+    if ('convertToFileContent' in change) {
+        return change.convertToFileContent();
+    }
+
+    if ('getDefinition' in change) {
+        return change.getDefinition();
+    }
+
+    throw new Error('Unsupported change object');
+}
```
mmilko01: These functions seem very convenient. My only comment here is that we might consider moving them in a separate file in the utils (e.g. `changes.ts`). My thinking is that they no longer seem very related to the `additional info` functionality anymore. What do you think?

mmilko01 (commented):
Changes look good, I have just one small suggestion.

packages/preview-middleware-client/src/utils/additional-change-info.ts
avasilev-sap: Agree will move them in subsequent commit

avasilev-sap: fixed in d5ed1f36fe080f46142454b2714dfad0ffdbb6b8

.changeset/soft-lilies-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux-private/preview-middleware-client': patch
```
mmilko01, line 2: I missed that the first time. When you make changes to `@sap-ux-private/preview-middleware-client`, you need also to include a new patch for `@sap-ux/priview-middleware` to consume the change.

avasilev-sap: fixed in [6811e2e](https://github.com/SAP/open-ux-tools/pull/4766/commits/6811e2e31a6c2f1758c37729b4ced60be6666ec9)

avasilev-sap: sorry fixed in [2217df6](https://github.com/SAP/open-ux-tools/pull/4766/commits/2217df60ec3966a51300ae2a4c449902afeae0b4) i added wrong package name for the `priview-middleware`.

mmilko01 (approved):
Changeset OK
Excellent test coverage
Changes look good
Comments addressed
Did not test manually

Jimmy-Joseph19 (approved):
Changes looks good.
Integration test passed https://github.com/SAP/open-ux-tools/actions/runs/27197022290/job/80291191333 (1 test failed known issue in main branch also)
Changeset found.
Did not test locally

PR #4777, opened 2026-06-05
chore: replace per-package tsconfig.eslint.json with projectService:true in root eslint config

heimwege:
## Summary

  - Replaced `project: './tsconfig.eslint.json'` + `tsconfigRootDir` in 87 per-package `eslint.config.mjs` files with a single `projectService: true` (+ `tsconfigRootDir` as monorepo boundary guard) in the root `eslint.config.mjs`
  - Deleted all 88 `tsconfig.eslint.json` files - these existed solely to extend `tsconfig.json` with `test/` included for ESLint; `projectService` discovers the right tsconfig per-file automatically
  - Updated `preview-middleware-client/jest.config.mjs` to use the existing `test/tsconfig.json` instead of the now-deleted `tsconfig.eslint.json`

  ## Why

  `projectService: true` (typescript-eslint v7+) auto-discovers the nearest `tsconfig.json` for each linted file by walking up the directory tree, making explicit project references redundant. This eliminates 88 boilerplate files and ~6 lines of repeated config per package.

  ## Test plan

  - [x] Run `pnpm lint` at repo root - no parsing errors
  - [x] Run `pnpm --filter @sap-ux/project-access exec eslint test/` - test files are linted with type information
  - [x] Run `pnpm test` for package /preview-middleware-client - ts-jest still resolves correctly via
  `test/tsconfig.json`


packages/abap-deploy-config-inquirer/tsconfig.eslint.json
```diff
@@ -1,4 +0,0 @@
-{
-    "extends": "./tsconfig.json",
-    "include": ["src", "test", ".eslintrc.js"]
```
devinea, line 3: I would expect tsconfig.json to now include `test` etc...
but this could cause test to be output to the dist folder from what I recall

devinea: Seems the include is handled by eslint and not tsconfig when using `projectService`

heimwege: _projectService: true doesn't use tsconfig.json's include list to decide which files to lint - that's ESLint's job. ESLint finds files to lint by its own file globbing (files: ['**/*.ts']). For each file ESLint decides to lint, projectService then finds the nearest tsconfig.json and uses its compiler options (strict mode, paths, libs, etc.) to provide type information - regardless of whether that file is listed in include.
So include in tsconfig.json only matters for tsc builds (what gets compiled/emitted). For ESLint with projectService, it's irrelevant - every .ts file ESLint touches gets type-checked using the nearest tsconfig's compiler options.
That's exactly why tsconfig.eslint.json existed in the first place - the old project: './tsconfig.eslint.json' approach did require the file to be listed in the tsconfig's include, so a separate file extending tsconfig.json with "include": ["src", "test"] was needed. projectService removes that constraint entirely._

eslint.config.mjs
```diff
@@ -281,7 +281,12 @@ export default [
     },
     {
         name: 'typescript-eslint-1',
-        languageOptions: {},
+        languageOptions: {
+            parserOptions: {
+                projectService: true,
+                tsconfigRootDir: __dirname
```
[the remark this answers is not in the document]
heimwege: The bot's suggestion is not needed here. Every .ts file that gets linted belongs to a package with its own tsconfig.json, which projectService: true will discover automatically. The scripts/ directory (the only place root-level TS files would live) is explicitly ignored. There's no scenario where parserServices.program is undefined can occur with this setup.

The comment is advisory bot noise - safe to dismiss/close.

packages/ui-components/eslint.config.mjs
```diff
@@ -40,9 +35,7 @@ export default [
     },
     {
         languageOptions: {
-            'parser': '@typescript-eslint/parser',
-            project: './tsconfig.eslint.json',
-            tsconfigRootDir: __dirname
+            'parser': '@typescript-eslint/parser'
```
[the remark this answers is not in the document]
heimwege: The bot is mostly correct that the removed keys were already no-ops, but its concern about projectService coverage is unfounded because the root config covers all .tsx files globally.

nikmace (approved):
Changes look good for ADP-related packages (I guess for others too because the changes are literally the same)
Coverage is excellent 😆 
Changeset not needed
Did not test manually

mikicvi-SAP (approved):
Changes Lg, exactly the same across the entire repo. 👍🏻 
- Did not test manually
- Coverage amazing :D 
- No need for changeset ✅

devinea (approved):
code changes lg. Thanks for the clean-up.

PR #4796, opened 2026-06-08
Fix: CDS multiple service targets in the same file, pointers bug fix

vadson71:
Fixed a bug where pointers were not generated correctly when annotations were applied to targets belonging to different services in the same CDS file.


packages/fiori-annotation-api/src/cds/pointer.ts
```diff
@@ -46,20 +46,25 @@ class Visitor {
...
         if (!Number.isNaN(index)) {
             if (segment === 'targets') {
                 if (segments.length === 0) {
                     return {
-                        pointer: [segment, indexSegment]
+                        pointer: [segment, cdsTargetIndex.toFixed(0)]
                     };
                 }
-                const result = this.target(astNode.targets[index], node.targets[index], segments);
+                const result = this.target(astNode.targets[cdsTargetIndex], node.targets[index], segments);
```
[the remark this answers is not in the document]
vadson71: This is hardly realistic case. Mapping data is built inside target filtering logic.
For each target that is included into the annotation file an entry is added to the mapping array. So, target indexes in pointers should always be within the array range.

vadson71:
https://github.wdf.sap.corp/ux-engineering/tools-suite/issues/38410 associated issue

packages/fiori-annotation-api/test/unit/cds/pointer.test.ts
```diff
@@ -14,22 +14,29 @@ import { getCDSDocument } from './utils.js';
...
     let ast: CDSDocument;
     let annotationFile: AnnotationFile;
     function testPointer(input: string, output: string, flattened = false) {
-        const { pointer, containsFlattenedNodes } = convertPointer(annotationFile, input, ast);
+        const { pointer, containsFlattenedNodes } = convertPointer(
+            annotationFile,
+            input,
+            ast,
+            [0, 1, 2, 3, 4, 5, 6, 7, 8]
+        );
```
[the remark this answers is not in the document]
vadson71: Comment was reasonable and helpful. Refactored accordingly

packages/fiori-annotation-api/src/cds/document.ts
```diff
@@ -150,17 +152,25 @@ export function getGhostFileDocument(
...
-    );
+    annotationFile.targets = annotationFile.targets.filter((target, idx) => {
+        const result =
+            target.name.startsWith(serviceName + '.') || (aliasName && target.name.startsWith(aliasName + '.'));
+        if (result) {
+            index.push(idx);
+        }
+        return result;
+    });
+    return index;
```
voicis: `filter` shouldn't have side effects. In this case it would be better to use for loop.

packages/fiori-annotation-api/test/unit/cds/pointer.test.ts
```diff
@@ -207,6 +211,77 @@ annotate service.Individual with {
...
+    Label                 : 'ok',
+    ![@UI.Hidden],
+    ![@Common.Application]: {
+        $Type               : 'Common.ApplicationType',
+        Component           : 'accepted',
+        ServiceId           : 'Actions',
+        ![@Common.Timestamp]: timestamp'2020-01-01 12:23:23',
+    },
+}];
+`,
```
voicis, line 257: We're testing only 2 pointers. Do we really need that large snippet? The annotations could be a bit more compact/simpler. It is easier to read the test case if the snippet contains the required code.

voicis (approved):
- Code changes looks good
- Review commends addressed
- Did not test locally

packages/fiori-annotation-api/src/cds/document.ts
vadson71: refactored

packages/fiori-annotation-api/test/unit/cds/pointer.test.ts
vadson71: Reduced snippet size

Jimmy-Joseph19 (approved):
Changes looks good.
Did not test locally
Changeset maintained
Approved

815are (approved):

PR #4797, opened 2026-06-08
fix: xsnKind missing genericDefinition

Jimmy-Joseph19:
### Problem
XSN kind was missing, leading to errors when processing annotation targets. The compiler structure has changed in newer versions.

### Solution
Updated kind extraction logic to prioritize the new path `_id.kind` with fallback to the legacy `definitions.kind` for backward compatibility with older compiler versions.

### Changes
- Extract kind from `carrier.definitions._id.kind` (new compiler path)
- Fallback to `carrier.definitions.kind` (legacy path)
- Maintains backward compatibility across compiler versions


vadson71 (approved):
Code changes look good
Comments addressed

vinayhospete (approved):

PR #4798, opened 2026-06-08
fix(middleware): path mappings for ui5 project type component

ytpo-lyne:
# Fix Path Mappings for UI5 Component-Type Projects in Preview Middleware

### Bug Fix

🐛 Fixes incorrect path mappings when the preview middleware is used with **component-type** UI5 projects. For component projects, resources are served under a `/resources/<namespace>` prefix, which was not being accounted for in API endpoint construction or editor template configuration.

### Changes

* `packages/preview-middleware-client/src/cpe/changes/service.ts`: Added `getApiBaseUrl()` and `getApiUrl()` helper functions that read a `data-open-ux-preview-base-url` attribute from the `#sap-ui-bootstrap` DOM element. These helpers ensure that API calls (fetch/delete for flex changes) are correctly prefixed with the resources path for component-type projects.

* `packages/preview-middleware/src/base/config.ts`: Updated `remapResourcesForPath()` to accept an optional `utils` parameter. The function now branches on project type — for component projects it computes resource root URLs relative to the `/resources/<namespace>` prefix and sets `config.appBasePath` accordingly. Application-type projects retain their existing behavior.

* `packages/preview-middleware/src/base/flp.ts`: Extracted a `getBaseUrl()` helper method that combines the patched router base URL with the resources prefix for component projects. Passes `this.utils` to `remapResourcesForPath()` and uses the new `getBaseUrl()` in both the sandbox editor generator and the RTA developer mode editor template to ensure correct API routing.

* `packages/preview-middleware/src/base/index.ts` & `src/index.ts`: Exported `getResourcesPathPrefix` and `getTestResourcesPathPrefix` utilities for external use.

* `packages/preview-middleware-client/test/unit/cpe/changes/service.test.ts`: Added tests verifying that fetch and delete API calls include the base URL prefix when a component project's `data-open-ux-preview-base-url` attribute is set.

* `packages/preview-middleware/test/unit/base/config.test.ts`: Added tests covering `remapResourcesForPath()` for component-type projects with various editor page paths, including nested paths, root-level paths, and non-absolute path normalization.

<details>

**Version:** `1.22.4`

- Correlation ID: `74d22759-3527-432a-9c36-2f5475799b12`
</details>


packages/preview-middleware-client/src/cpe/changes/service.ts
```diff
@@ -49,6 +49,33 @@ const TITLE_MAP: { [key: string]: string } = {
...
+ * Looks for the data attribute on both #root (outer CPE frame) and #sap-ui-bootstrap (inner UI5 frame).
+ * For component-type projects, this returns the resources path prefix (e.g., '/resources/project1').
+ * For application-type projects, this returns an empty string.
+ *
+ * @returns The base URL string to prepend to API endpoints.
+ */
+function getApiBaseUrl(): string {
+    // Try outer frame's root element first (CPE context), then inner frame's bootstrap element (UI5 context)
+    return (
+        document.getElementById('root')?.dataset.openUxPreviewBaseUrl ??
```
heimwege: We already use this notation for ADP, e.g. [here](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware-client/src/adp/api-handler.ts#L89), but why should we use this fallback? That'S currently not in use anywhere else

packages/preview-middleware-client/src/cpe/changes/service.ts
```diff
@@ -173,7 +200,7 @@ export class ChangeService extends EventTarget {
      */
     private async fetchSavedChanges(): Promise<void> {
         this.changedFiles = {};
-        const savedChangesResponse = await fetch(FlexChangesEndPoints.changes + `?_=${Date.now()}`);
+        const savedChangesResponse = await fetch(getApiUrl(FlexChangesEndPoints.changes) + `?_=${Date.now()}`);
```
heimwege, line 206: If we don't need the fallback logic from `root` to `sap-ui-bootstrap` we can use this directly

```ts
const baseUrl = document.getElementById('sap-ui-bootstrap')?.dataset.openUxPreviewBaseUrl ?? '';
const response: Response = await fetch(`${baseUrl}${FlexChangesEndPoints.changes}?_=${Date.now()}`
```

packages/preview-middleware/src/base/config.ts
```diff
@@ -458,20 +458,40 @@ async function getI18nTextFromProperty(
  * @param config - cloned TemplateConfig already built for the FLP path
  * @param newPagePath - the path of the HTML page that will serve the resources (e.g. `editor.path`)
  * @param appId - the `sap.app.id` used as resource-root key for the primary application
+ * @param utils - middleware utils to determine project type (application vs component)
  */
-export function remapResourcesForPath(config: TemplateConfig, newPagePath: string, appId: string): void {
+export function remapResourcesForPath(
```
heimwege, line 468: _Note to self: The remapResourcesForPath comes from #4397 and is needed because the rta editor paths use the resource-root from `flp.path`. In case of CAP apps that use `cds-plugin-ui5` this can lead to wrong `resource-roots` when the editor is located in a different place than the flp endpoint_

packages/preview-middleware/src/base/flp.ts
```diff
@@ -407,6 +407,10 @@ export class FlpSandbox {
         let livereloadPort: number = envPort ? Number.parseInt(envPort, 10) : DEFAULT_LIVERELOAD_PORT;
         livereloadPort = Number.isNaN(livereloadPort) ? DEFAULT_LIVERELOAD_PORT : livereloadPort;
         const envLivereloadUrl = isAppStudio() ? await exposePort(livereloadPort) : undefined;
+        // For component projects, baseUrl must include the resources prefix for correct API paths
+        const patchedRouterBaseUrl = req['ui5-patched-router']?.baseUrl ?? '';
```
heimwege, line 428: Do we need this for `editorGetHandler` as well?

packages/preview-middleware-client/src/cpe/changes/service.ts
heimwege: Actually it's the other way around:

`WorkspaceConnector.ts` line 9 already shows the correct pattern - it reads directly from `#sap-ui-bootstrap`. `cpe/init.ts` confirms `ChangeService` runs in the inner UI5 frame (initialized by UI5's RTA), never in the outer CPE frame where `#root` lives. The `#root` check in `getApiBaseUrl` is dead code - `#root` never exists in the frame where `ChangeService` runs. It should just be:

```ts
  function getApiBaseUrl(): string {
      return document.getElementById('sap-ui-bootstrap')?.dataset.openUxPreviewBaseUrl ?? '';
  }
```

packages/preview-middleware/src/base/config.ts
```diff
@@ -458,20 +458,40 @@ async function getI18nTextFromProperty(
...
+    config: TemplateConfig,
+    newPagePath: string,
+    appId: string,
+    utils?: MiddlewareUtils
+): void {
     const newBasePath = posix.relative(posix.dirname(newPagePath), '/') || '.';
```
heimwege: Not sure about this whole change. I think all only need to change two lines

```ts
export function remapResourcesForPath(config: TemplateConfig, newPagePath: string, appId: string): void {
      const resourcesPrefix = getResourcesPathPrefix(utils); // <----- this is new
      const base = resourcesPrefix ?? '/'; // <----- this is new
      const newBasePath = posix.relative(posix.dirname(newPagePath), base) || '.';

      // Update the well-known client namespace to be relative to the new page path
      config.ui5.resources[PREVIEW_URL.client.ns] = PREVIEW_URL.client.getUrl(newBasePath);

      // Update the primary app's resource root (was set to basePath for the main app)
      if (appId && appId in config.ui5.resources) {
          config.ui5.resources[appId] = newBasePath;
      }

      // Keep basePath in sync for any other template usage (e.g. basePath/resources/...)
      config.basePath = newBasePath;
  }
```

packages/preview-middleware/src/base/flp.ts
heimwege: No we don't. This is taken care of in #3927

packages/preview-middleware/src/base/config.ts
heimwege: oh and we also need unit tests for this

packages/preview-middleware-client/src/cpe/changes/service.ts
```diff
@@ -49,6 +49,33 @@ const TITLE_MAP: { [key: string]: string } = {
...
+}
+
+/**
+ * Constructs the full API endpoint URL by prepending the base URL.
+ *
+ * @param endpoint - The API endpoint path (e.g., '/preview/api/changes')
+ * @returns The full API URL including the base path for component projects
+ */
+function getApiUrl(endpoint: string): string {
+    return `${getApiBaseUrl()}${endpoint}`;
```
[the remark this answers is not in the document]
heimwege: Suggestion to strip trailing slash is actually unnecessary since `getResourcesPathPrefix` always returns a path without trailing slash (e.g. `/resources/my/ns`).

packages/preview-middleware/src/base/flp.ts
```diff
@@ -407,6 +407,10 @@ export class FlpSandbox {
...
+        const resourcesPrefix = getResourcesPathPrefix(this.utils);
+        const baseUrl = resourcesPrefix ? posix.join(patchedRouterBaseUrl, resourcesPrefix) : patchedRouterBaseUrl;
```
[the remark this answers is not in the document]
ytpo-lyne: Fixed in 8ee68bd4413e20b3f7adc219ff10a7fc33c90397

packages/preview-middleware/src/base/config.ts
```diff
@@ -458,20 +458,40 @@ async function getI18nTextFromProperty(
...
-        config.ui5.resources[appId] = newBasePath;
+    // For component projects, preview client is served under the resources prefix
+    // e.g., /resources/my/namespace/preview/client instead of /preview/client
+    const resourcesPrefix = getResourcesPathPrefix(utils);
+    if (resourcesPrefix) {
+        // Calculate relative path from editor.html to /resources/<ns>/preview/client
+        const clientPath = posix.join(resourcesPrefix, PREVIEW_URL.client.path);
+        config.ui5.resources[PREVIEW_URL.client.ns] = posix.relative(posix.dirname(newPagePath), clientPath) || '.';
+        // For component projects, appBasePath should point to the resources prefix
+        const appPath = posix.relative(posix.dirname(newPagePath), resourcesPrefix) || '.';
```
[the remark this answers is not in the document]
ytpo-lyne: Fixed in 8ee68bd4413e20b3f7adc219ff10a7fc33c90397

packages/preview-middleware-client/src/cpe/changes/service.ts
ytpo-lyne: Fixed in 8ee68bd4413e20b3f7adc219ff10a7fc33c90397

packages/preview-middleware/src/base/config.ts
ytpo-lyne: 1. PREVIEW_URL.client.getUrl(newBasePath) assumes root-relative structure:
    - For application projects: getUrl('.') → ./open/ux/preview/client
    - For component projects: The client files are served at /resources/<ns>/preview/client/, not relative to /


2. The paths are structurally different:
    - Application: FLP at /test/flp.html, client at /open/ux/preview/client/ → relative path from test to /open/... is ../open/ux/preview/client
    - Component: FLP at /resources/ns/test/flp.html, client at /resources/ns/preview/client/ → relative path from /resources/ns/test/ to /resources/ns/preview/client is ../preview/client


The simplified approach would give ../open/ux/preview/client for components, which is wrong.

````
export const PREVIEW_URL = {
    client: {
        path: '/preview/client',
        getUrl: (basePath: string) => posix.join(basePath, 'preview', 'client'),
        local: join(__dirname, '../../dist/client'),
        ns: 'open.ux.preview.client'
    },
    api: '/preview/api'
} as const;
````


The open.ux.preview.client namespace is configured in data-sap-ui-resourceroots in the HTML:
````
<script data-sap-ui-resourceroots='{
  "open.ux.preview.client": "../preview/client"
}'></script>
````

This tells UI5: "When loading sap.ui.require(['open/ux/preview/client/...']), fetch from ../preview/client/..."

File: packages/preview-middleware-client/src/cpe/changes/service.ts (hunk 2 above)
Comment (ytpo-lyne) [reply packages/preview-middleware-client/src/cpe/changes/service.ts:206]: The helper functions getApiBaseUr/getApiUrl are used in 2 places (fetchSavedChanges and deleteChange), so keeping them avoids duplicating the DOM access logic. I removed the #root fallback per your earlier comment since ChangeService only runs in the inner frame.

File: packages/preview-middleware/src/base/flp.ts [hunk 9]
```diff
@@ -319,6 +319,19 @@ export class FlpSandbox {
...
     }
 
+    /**
+     * Computes the base URL for API endpoints, combining the patched router base URL
+     * with the resources prefix for component-type projects.
+     *
+     * @param req - The enhanced request containing the patched router info
+     * @returns The base URL string for API endpoints
+     */
+    private getBaseUrl(req: EnhancedRequest): string {
```
heimwege: The naming is genuinely confusing. `baseUrl` in this codebase has two distinct meanings:

1. `req['ui5-patched-router']?.baseUrl` - the CAP/cds-plugin-ui5 router mount prefix (e.g. /app/myapp), set by the patched router middleware
2. `TemplateConfig.baseUrl` - the combined value written into `data-open-ux-preview-base-url` on the HTML, which after PR #3927/#4798 is `patchedRouterBaseUrl + resourcesPrefix`

The new private method `getBaseUrl(req)` returns meaning #2 - it's the template base URL that combines both the router prefix and the resources namespace prefix. Calling it just `getBaseUrl` makes it look like it's simply extracting the router's `baseUrl`, when it's actually computing something more composite.

A clearer name would be `getTemplateBaseUrl(req)` - something that signals it's the fully-resolved value for template rendering, not just a passthrough of ui5-patched-router's `baseUrl`.

packages/preview-middleware-client/src/cpe/changes/service.ts
```diff
@@ -49,6 +49,29 @@ const TITLE_MAP: { [key: string]: string } = {
...
```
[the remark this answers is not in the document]
ytpo-lyne: document.getElementById() is a microsecond operation, won't change.

packages/preview-middleware/src/base/flp.ts
```diff
@@ -333,9 +346,8 @@ export class FlpSandbox {
             : '@sap-ux/preview-middleware';
 
         await this.setApplicationDependencies();
+        const baseUrl = this.getBaseUrl(req);
         const patchedRouterBaseUrl = req['ui5-patched-router']?.baseUrl ?? '';
-        const resourcesPrefix = getResourcesPathPrefix(this.utils);
-        const baseUrl = resourcesPrefix ? posix.join(patchedRouterBaseUrl, resourcesPrefix) : patchedRouterBaseUrl;
         const ui5Version = await this.getUi5Version(req.protocol, req.headers.host, patchedRouterBaseUrl);
```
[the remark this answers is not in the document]
ytpo-lyne: Fixed in 846664071ee68458db751f6ba207ca919e42b5fa

packages/preview-middleware/src/base/config.ts
```diff
@@ -458,20 +458,44 @@ async function getI18nTextFromProperty(
...
+ * @param utils - middleware utils to determine project type (application vs component)
  */
-export function remapResourcesForPath(config: TemplateConfig, newPagePath: string, appId: string): void {
+export function remapResourcesForPath(
+    config: TemplateConfig,
+    newPagePath: string,
+    appId: string,
+    utils?: MiddlewareUtils
+): void {
     const newBasePath = posix.relative(posix.dirname(newPagePath), '/') || '.';
```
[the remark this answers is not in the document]
ytpo-lyne: Fixed in e2bf10c5b872de08229ffafce5a85ee86402752f

packages/preview-middleware/src/base/config.ts
```diff
@@ -458,20 +458,44 @@ async function getI18nTextFromProperty(
...
+    // e.g., /resources/my/namespace/preview/client instead of /preview/client
+    const resourcesPrefix = getResourcesPathPrefix(utils);
+    if (resourcesPrefix) {
+        // Ensure page directory is absolute for correct relative path calculation
+        const pageDir = posix.isAbsolute(newPagePath)
+            ? posix.dirname(newPagePath)
+            : posix.dirname(posix.join('/', newPagePath));
+        // Calculate relative path from editor.html to /resources/<ns>/preview/client
+        const clientPath = posix.join(resourcesPrefix, PREVIEW_URL.client.path);
+        config.ui5.resources[PREVIEW_URL.client.ns] = posix.relative(pageDir, clientPath) || '.';
```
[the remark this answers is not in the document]
ytpo-lyne: Fixed in e2bf10c5b872de08229ffafce5a85ee86402752f

packages/preview-middleware/src/base/config.ts
heimwege: The simplified approach works correctly in all four scenarios:

  - Application editor at /editor.html → preview/client ✓
  - Application editor at /test/editor.html → ../preview/client ✓
  - Component editor at /resources/my/ns/editor.html → preview/client ✓ (resolves to /resources/my/ns/preview/client because the browser resolves it relative to the page's own URL)
  - Component editor at /editor.html → resources/my/ns/preview/client ✓

  The explanation in the comment is also internally confused. It says "The simplified approach would give ../open/ux/preview/client for components" - but getUrl appends preview/client, not open/ux/preview/client. The open.ux.preview.client is the UI5 namespace key in data-sap-ui-resourceroots, not the path. getUrl just does posix.join(basePath, 'preview', 'client') — there's no open/ux in the output.

  The reason ytpo-lyne kept the if/else is based on a misunderstanding of what getUrl produces. The simplified two-line approach you suggested is correct and the if/else branch in the current PR is unnecessary complexity. Worth pushing back on that comment directly.

ytpo-lyne: Fixed

heimwege: I forgot one line when pasting the code. sorry. my bad

heimwege (commented):
I forgot to send my comments 🙈

packages/preview-middleware/src/base/flp.ts
heimwege: adjusted to getTemplateBaseUrl

PR #4801, opened 2026-06-09
feat: switch build from tsc to esbuild bundling

815are:
switch build from tsc to esbuild bundling for `@sap-ux/eslint-plugin-fiori-tools`
tgz:
[sap-ux-eslint-plugin-fiori-tools-10.6.3.zip](https://github.com/user-attachments/files/28745053/sap-ux-eslint-plugin-fiori-tools-10.6.3.zip)

but still requires external packages:
'eslint', 'typescript-eslint' are peer dependencies
'@babel/core' is peerDependency in https://github.com/babel/babel/blob/main/eslint/babel-eslint-parser/package.json#L32
`global`uses const globals = require('globals') where require is the createRequire(import.meta.url) wrapper, not a native CJS require -> https://github.com/SAP/open-ux-tools/blob/main/packages/eslint-plugin-fiori-tools/src/index.ts#L23


815are:
latest - https://github.com/SAP/open-ux-tools/pull/4807

PR #4805, opened 2026-06-09
Fix/38152/validate min ui5 for add card config command

Jimmy-Joseph19:
validate min ui5 for add card config command. 
Card command is applicable only for edmx version > 1.136 and cap version >1.149


Jimmy-Joseph19:
Closing

Jimmy-Joseph19:
Wrongly added patch from another branch

PR #4807, opened 2026-06-09
feat: switch build from tsc to esbuild bundling

815are:
switch build from tsc to esbuild bundling for `@sap-ux/eslint-plugin-fiori-tools`
tgz:
[sap-ux-eslint-plugin-fiori-tools-10.6.10.zip](https://github.com/user-attachments/files/28755207/sap-ux-eslint-plugin-fiori-tools-10.6.10.zip)

but still requires external packages:
- `'eslint`, 'typescript-eslint' are peer dependencies
- `@babel/core` is peerDependency in https://github.com/babel/babel/blob/main/eslint/babel-eslint-parser/package.json#L32
- `@babel/eslint-parser` and `@babel/parser` is required otherwise `npx eslint` freezes/fails

additional change:
`global` use using import instead of non native CJS require -> https://github.com/SAP/open-ux-tools/blob/main/packages/eslint-plugin-fiori-tools/src/index.ts#L23 - it allows to bundle it isntead of marking as external dependency


packages/eslint-plugin-fiori-tools/package.json
```diff
@@ -15,49 +15,53 @@
...
         "globals": "17.4.0",
         "lodash": "4.18.1",
         "requireindex": "^1.2.0",
+        "semver": "7.7.4",
         "synckit": "0.11.12",
-        "yaml": "2.8.3",
-        "semver": "7.7.4"
+        "yaml": "2.8.3"
     },
     "peerDependencies": {
```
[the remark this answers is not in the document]
815are: fixed d740abe78d3d5811e81f1494a876ce5d0c3d2fb6

packages/eslint-plugin-fiori-tools/esbuild.mjs
```diff
@@ -0,0 +1,57 @@
...
+    // dynamically at runtime so esbuild cannot trace or inline it.
+    external: ['eslint', 'typescript-eslint', '@babel/core'],
+});
+
+// Worker: finds Fiori artifacts (called via synckit from project-context.ts)
+await build({
+    ...sharedOptions,
+    entryPoints: [join(__dirname, 'src/project-context/artifacts.ts')],
+    outfile: join(__dirname, 'lib/project-context/artifacts.js'),
+});
```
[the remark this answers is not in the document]
815are: 00c26c88f9a1e30abe479410f9d10a72fe442164

packages/eslint-plugin-fiori-tools/esbuild.mjs
```diff
@@ -0,0 +1,57 @@
...
+
+// Worker: resolves path mappings synchronously (called via synckit from index.ts)
+await build({
+    ...sharedOptions,
+    entryPoints: [join(__dirname, 'src/worker-getPathMappingsSync.ts')],
+    outfile: join(__dirname, 'lib/worker-getPathMappingsSync.js'),
+});
```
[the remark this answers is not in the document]
815are: 00c26c88f9a1e30abe479410f9d10a72fe442164

packages/eslint-plugin-fiori-tools/package.json
```diff
@@ -15,49 +15,53 @@
     ],
     "main": "./lib/index.js",
     "scripts": {
-        "build": "tsc --build",
-        "watch": "tsc --build --watch",
+        "build": "node esbuild.mjs --production",
+        "bundle:dev": "node esbuild.mjs",
+        "bundle:prod": "node esbuild.mjs --production",
```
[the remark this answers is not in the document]
815are: 09ce2bae4767f4ee8bfbf7bb816c61dacd6e74ac

packages/eslint-plugin-fiori-tools/package.json
```diff
@@ -15,49 +15,53 @@
...
+        "check-types": "tsc --noEmit",
+        "watch": "tsc --noEmit --watch",
```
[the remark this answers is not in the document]
815are: ec71e7b857d4bcf07d01a2e117f0ba07ae2e1384

packages/eslint-plugin-fiori-tools/package.json
```diff
@@ -15,49 +15,54 @@
...
+        "@humanwhocodes/momoa": "^3.3.9",
+        "@jest/globals": "30.3.0",
         "@sap-ux/fiori-annotation-api": "workspace:*",
         "@sap-ux/odata-annotation-core": "workspace:*",
         "@sap-ux/odata-vocabularies": "workspace:*",
         "@sap-ux/odata-entity-model": "workspace:*",
         "@sap-ux/project-access": "workspace:*",
         "@sap-ux/vocabularies-types": "0.15.0",
+        "@typescript-eslint/eslint-plugin": "8.57.2",
+        "@typescript-eslint/parser": "8.57.2",
```
devinea: Already in dependencies

packages/eslint-plugin-fiori-tools/esbuild.mjs
```diff
@@ -0,0 +1,65 @@
...
+    'const __dirname=__cjsDirname(__filename);'
+].join('');
+
+// eslint and typescript-eslint are peerDependencies provided by the consumer.
+// @typescript-eslint/eslint-plugin and @typescript-eslint/parser are covered by the
+// typescript-eslint peer — marking them external drops the ~3.5 MB typescript.js from the bundle.
+// @babel/* are peerDependencies of @babel/eslint-parser — required dynamically at runtime.
+const externalDependencies = [
+    'eslint',
+    'typescript-eslint',
```
devinea:
```suggestion
```
Don't think we need this anymore

devinea (commented):
Great improvement.

packages/eslint-plugin-fiori-tools/package.json
815are: fix - 9cc7fdfa4cdc6ae336a84426cc7a82f2e4565d5f

packages/eslint-plugin-fiori-tools/esbuild.mjs
815are: fix - 9cc7fdfa4cdc6ae336a84426cc7a82f2e4565d5f

packages/eslint-plugin-fiori-tools/package.json
```diff
@@ -15,49 +15,52 @@
...
         "@sap-ux/project-access": "workspace:*",
         "@sap-ux/vocabularies-types": "0.15.0",
+        "@typescript-eslint/rule-tester": "8.57.2",
+        "@types/semver": "7.7.1",
         "@xml-tools/ast": "5.0.5",
         "@xml-tools/parser": "1.0.11",
-        "@humanwhocodes/momoa": "^3.3.9",
-        "@eslint/plugin-kit": "0.6.1",
+        "cross-env": "10.1.0",
+        "esbuild": "0.27.4",
```
heimwege, line 55: package.json says "esbuild": "0.27.4" (pinned), but pnpm-lock.yaml shows specifier: '>=0.25.0'. One of them is wrong

packages/eslint-plugin-fiori-tools/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@sap-ux/eslint-plugin-fiori-tools",
-    "version": "10.6.0",
+    "version": "10.6.22",
```
heimwege: This should not be bumped in the file because version management is handled by changesets, right?

devinea (dismissed):
Thanks for the updates

packages/eslint-plugin-fiori-tools/src/project-context/project-context.ts
```diff
@@ -31,8 +31,9 @@ function getWorkerPath(name: string): string {
 
     // Try multiple possible worker locations
     const workerPaths = [
-        join(currentDir, name), // src location
-        join(currentDir, '..', '..', 'lib', 'project-context', name) // dist location
+        join(currentDir, name), // src location (ts-node / tsc: lib/project-context/)
+        join(currentDir, '..', '..', 'lib', 'project-context', name), // tsc dist location
+        join(currentDir, 'project-context', name) // esbuild bundle: lib/index.js → lib/project-context/
```
heimwege: import.meta.url always points to the file containing that code. project-context.ts is compiled into lib/project-context/project-context.js by tsc. With esbuild, project-context.ts is bundled into lib/index.js. That changes what import.meta.url resolves to.

### Current behavior (tsc, before this PR)

  The file project-context.ts compiles 1:1 to lib/project-context/project-context.js.

  At runtime:
  - import.meta.url = .../lib/project-context/project-context.js
  - currentDir = .../lib/project-context/

  The two candidate paths:
  1. lib/project-context/artifacts.js - exists, hits immediately
  2. lib/../../lib/project-context/artifacts.js - never reached

### After this PR (esbuild)

  project-context.ts is no longer its own file - its code is inlined into lib/index.js.

  At runtime:
  - import.meta.url = .../lib/index.js
  - currentDir = .../lib/

  The two OLD candidate paths become:
  1. lib/artifacts.js - does not exist
  2. lib/../../lib/project-context/artifacts.js = .../lib/project-context/artifacts.js - exists, hits

  So the old path 2 already works for esbuild. Path 3 that the PR adds (`join(currentDir, 'project-context', name) = lib/project-context/artifacts.js`) resolves to the exact same file as path 2 in the esbuild scenario.

### Conclusion

  Path 3 is reachable (path 1 misses, path 2 hits before it) - so path 3 is dead code. It never fires. The author likely added it thinking path 2 would not work in the esbuild case, but the `../../lib/project-context/` in path 2 happens to resolve correctly from `lib/` too.

heimwege (commented):
see comments

packages/eslint-plugin-fiori-tools/package.json
815are: I reused same version like in other packages:
1. mcp-server -> https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-mcp-server/package.json#L87
2. root level package.json - https://github.com/SAP/open-ux-tools/blob/main/package.json#L18

packages/eslint-plugin-fiori-tools/package.json
815are: https://github.com/SAP/open-ux-tools/pull/4807/changes/0a1de3fddec33518f2bbc7c9e1310645a6986b00 - it was temp change got in, was testing and to compare with different versions changed them locally

packages/eslint-plugin-fiori-tools/src/project-context/project-context.ts
815are: @heimwege 
sorry, if I remove 3rd - esbuild version does not work for me

Image

heimwege (approved):
- review comments addressed
- SONAR is happy
- changeset ok
- did NOT test manually

AlinaGovoruhina (approved):
Tested the plugin .tgz with CAP and non-CAP apps, found no issues: manifest and annotation rule violations are detected both when run in the terminal or using Page map. Looks good!

Issue #4814, opened 2026-06-10
BUG - Feature toggle not read on first generator run after ESM migration


nikmace:
### Description
After the ESM migration of `@sap-ux/feature-toggle`, the first invocation of the ADP generator within a fresh VS Code session sees `FeatureToggleAccess.vscode` as **undefined**. As a result, feature toggle checks (e.g. `isFeatureEnabled('sap.ux.appGenerator.testBetaFeatures.adpCfExperimental')` in `generator-adp/src/app/index.ts`) return false even when the toggle is enabled in VS Code settings. Subsequent runs of the generator in the same session work correctly.

###  Steps to Reproduce

1. Enable the feature toggle sap.ux.appGenerator.testBetaFeatures.adpCfExperimental in VS Code settings.
2. Open a fresh VS Code window (kill the extension host / restart).
3. Launch the ADP generator (Open Application Generator → Adaptation Project).
4. Observe that the CF target environment step is not shown.
5. Close the generator and launch it again — the toggle is now picked up correctly.

### Expected results
The feature toggle is read on the first generator launch and the CF flow is offered.

### Actual results
On first launch the toggle is read as disabled; only on subsequent launches is it read correctly.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem

After the ESM migration, feature-toggle/src/featureToggle.ts initializes the cached vscode instance via a fire-and-forget async IIFE:
```ts
let _vscodeInstance: any = null;
(async () => { _vscodeInstance = await getVSCodeInstance(); })();
```

Module evaluation does not wait for this **IIFE**. Synchronous consumers (e.g. the generator constructor) read _vscodeInstance before the dynamic import('vscode') microtask resolves, so FeatureToggleAccess.vscode is undefined on first call. On subsequent generator runs, the module is already loaded, and the IIFE has long since resolved, so the value is populated — masking the race.

### Fix

Restore deterministic, synchronous resolution of the vscode module by loading it via `createRequire(import.meta.url)` instead of dynamic import(). This keeps the public API synchronous (no caller changes) and matches the pre-ESM behavior. Top-level await would also work but is more invasive on tooling.

### Why was it missed

The race only manifests on the first generator invocation in a fresh extension host. Existing tests import the module long before they call into it, and manual smoke tests typically reuse a warm session — both hide the timing window.

[...]

PR #4816, opened 2026-06-10
fix: Feature toggle not read on first generator run after ESM migration

nikmace:
Fix for #4814.
- After the ESM migration of `@sap-ux/feature-toggle`, feature toggles were not picked up on the **first** generator run in a fresh VS Code session. Subsequent runs worked, masking the bug in warm sessions.


packages/feature-toggle/src/featureToggle.ts
```diff
@@ -1,30 +1,9 @@
...
-let _vscodeInstance: any = null;
-
-// Initialize vscode instance without top-level await
-// This uses an IIFE to start the async initialization immediately
-// eslint-disable-next-line @typescript-eslint/no-floating-promises
-(async () => {
-    _vscodeInstance = await getVSCodeInstance();
-})();
+// Module-level vscode instance cache, resolved synchronously at module evaluation time.
+const _vscodeInstance: any = getVSCodeInstance();
```
heimwege: Don't we have a proper type instead of `any`?

.changeset/feature-toggle-vscode-sync-resolve.md
```diff
@@ -0,0 +1,9 @@
+---
+'@sap-ux/feature-toggle': patch
+---
+
+fix(feature-toggle): resolve vscode synchronously to fix first-run feature toggle race after ESM migration
+
+The async IIFE introduced during the ESM migration left `_vscodeInstance` as `null` on the first call into the module, because synchronous consumers (e.g. the ADP generator constructor) read the cache before the deferred `import('vscode')` microtask resolved. Subsequent calls within the same process worked because the IIFE had completed in the meantime — making the bug invisible in warm sessions.
+
+Resolution now uses `createRequire(import.meta.url)` to load `vscode` synchronously at module evaluation time, restoring the pre-migration behavior. Public API is unchanged. A test-only helper `_setVSCodeInstanceForTesting` is exported so tests can inject a mock instance (the `vscode` module is not resolvable outside the VS Code extension host).
```
heimwege: You sure about `_setVSCodeInstanceForTesting`? I don't see this in the changes

heimwege (commented):
see comments

packages/feature-toggle/src/featureToggle.ts
nikmace: It was typed as any already from the ESM migration. Here is the change: [e60041a](https://github.com/SAP/open-ux-tools/pull/4816/commits/e60041a0a7e5c50a7cbc3614b97237ce069c93b4)

.changeset/feature-toggle-vscode-sync-resolve.md
nikmace: Was trying different things and this artifact remained. Changed in: [e60041a](https://github.com/SAP/open-ux-tools/pull/4816/commits/e60041a0a7e5c50a7cbc3614b97237ce069c93b4)

packages/feature-toggle/src/featureToggle.ts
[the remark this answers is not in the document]
nikmace: **On the changeset:** stale wording — already updated. The PR does not export `_setVSCodeInstanceForTesting`; that was an earlier draft of the design we abandoned in favor of extracting the resolver into `src/vscode.ts`, which gives tests a clean module-level seam (`jest.unstable_mockModule('../src/vscode.js', …)`). The new changeset reflects that.

**On "frozen at module load":** I don't think this is reachable in practice for this package's call paths. The VS Code extension host installs its custom `require` interceptor for `'vscode'` *before* any extension entry point runs (see [`extensionHostProcess`](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/api/node/extensionHostProcess.ts) and the activation flow). All of `@sap-ux/feature-toggle`'s consumers (generators, middlewares, etc.) are loaded transitively from an extension entry point, so by the time this module evaluates, `require('vscode')` is already wired. If `require('vscode')` does throw, we're in CLI / non-host Node — exactly the case the `catch` branch handles.

The pre-ESM version of this module also resolved `vscode` once at load time via a top-level `require('vscode')` and cached it in a `const`-equivalent — same lifecycle as this PR. The bug we were fixing was specifically the *new* race the async IIFE introduced; restoring the original semantics is the goal.

packages/feature-toggle/src/vscode.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Using `createRequire` here (instead of dynamic `import()`) keeps resolution synchronous,
+ * which is required because the public API of `featureToggle` is synchronous and
+ * consumers (e.g. generator constructors) read the cached instance immediately on
+ * first import. A previous async-IIFE based initialization caused a first-run race
+ * where the cache was still `null` when consumers read it.
+ *
+ * @returns the vscode instance, or `undefined` when not running inside the extension
+ *          host (e.g. CLI, plain Node).
+ */
+export function getVSCodeInstance(): any {
```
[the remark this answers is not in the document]
nikmace: Added types [e60041a](https://github.com/SAP/open-ux-tools/pull/4816/commits/e60041a0a7e5c50a7cbc3614b97237ce069c93b4)

heimwege (approved):
- changeset ok
- review comments addressed
- SONAR happy
- did NOT test manually

mikicvi-SAP (approved):
Thanks @nikmace 
- fixes lg, follows already used pattern post ESM migration
- changeset a bit lengty but ✅
- did not test locally.

nikmace:
> Thanks @nikmace
> 
> * fixes lg, follows already used pattern post ESM migration
> * changeset a bit lengty but ✅
> * did not test locally.

@mikicvi-SAP I've noticed that people started writing more in changesets in general since they are used in the release notes or for some tracking purposes. Therefore, I am not sure what the correct guidance is here

devinea (approved):
Thanks for the fix @nikmace

PR #4817, opened 2026-06-10
fix(ui5-test-writer): use js files for FPM opa tests

docirl:
- ts files are not fully implemented for FPM projects.
- generator crashes as it tries to copy a ts file for OPA tests that does not exist


mikicvi-SAP (approved):
Thanks @docirl 
- change small and lg
- changeset ✅
- did not test manually

korotkovao (approved):
Changeset
Straightforward change 
Thanks @docirl

PR #4825, opened 2026-06-11
fix(ui5-test-writer): use js for FPM tests

docirl:
- FPM has a mix of ts/js test files
- FPM is not yet fully supported in the FE test library
- Current output is causing typescript build errors, impacting pipeline tests also
- For now FPM TS projects will be reverted to JS test output files


kjose90 (approved):
LGTM ✅ 
changeset, tests ✅

longieirl (approved):
Changeset is good.
Tests are good.
Change is good.

PR #4828, opened 2026-06-11
fix(#4756): derive builder.resources.excludes for ABAP deploy at runtime

### LINKED ISSUE #4756: bug(abap-deploy): ui5-deploy.yaml deploy task does not exclude localService from deployment
## Description

When generating `ui5-deploy.yaml` for ABAP deployment, the `deploy-to-abap` task's `configuration.exclude` array only includes `/test/` but does not exclude `localService`. This means mock data and local service metadata files under `localService/` can be unintentionally uploaded to the ABAP system during deployment.

## Root Cause

`addAbapDeployTask` in `@sap-ux/ui5-config` had a hardcoded, incomplete exclude list — only `/test/` was ever written into `configuration.exclude`. The build-time config (`builder.resources.excludes`) already correctly excluded `/localService/**`, but the deploy task config did not mirror this. The omission was baked into the writer itself, creating a silent misalignment between build-time and deploy-time exclusions that affected all callers regardless of generator or project type.

## Current Behavior

`builder.resources.excludes` is only appended to `ui5-deploy.yaml` — it is not written to `ui5.yaml`. Within `ui5-deploy.yaml` the entry is present in the build section but is effectively ignored by the deploy task, which maintains its own separate and incomplete `configuration.exclude` list:

```yaml
# ui5-deploy.yaml
builder:
  resources:
    excludes:
      - /test/**
      - /localService/**       # ✅ present, but only in ui5-deploy.yaml and ignored by deploy task
  customTasks:
    - name: deploy-to-abap
      afterTask: generateCachebusterInfo
      configuration:
        exclude:
          - /test/              # ❌ localService missing — deploy task does not inherit from builder.resources.excludes

# ui5.yaml — builder.resources.excludes not written here at all
```

## Expected Behavior

`builder.resources.excludes` should be the authoritative source for folder exclusions and must be appended to **both** `ui5.yaml` and `ui5-deploy.yaml`. The `deploy-to-abap` task should inherit these exclusions so that any folder excluded at build time is also excluded from deployment — no duplication, no drift:

```yaml
# ui5.yaml — builder.resources.excludes written here too
builder:
  resources:
    excludes:
      - /test/**
      - /localService/**       # ✅ appended to ui5.yaml

# ui5-deploy.yaml
builder:
  resources:
    excludes:
      - /test/**
      - /localService/**       # ✅ appended to ui5-deploy.yaml
  customTasks:
    - name: deploy-to-abap
      afterTask: generateCachebusterInfo
      configuration:
        exclude:
          - /test/
          - /localService/     # ✅ inherited — deploy task stays in sync with builder.resources.excludes
```

This ensures generators and tooling write consistent exclusions once, and the deploy task reflects them without requiring callers to manually keep both lists aligned.

## Steps to Reproduce

1. Generate an ABAP deployment configuration for a Fiori app that has a `localService/` folder (e.g. a project using mock data / local OData service metadata)
2. Inspect the generated `ui5-deploy.yaml`
3. Observe that `localService` appears in `builder.resources.excludes` but not in the deploy task's `configuration.exclude`
4. Inspect `ui5.yaml` — `builder.resources.excludes` is absent entirely

## Impact

`localService/` typically contains mock data (`mockdata/`) and OData metadata (`metadata.xml`) used only for local development. Deploying these files to an ABAP BSP application is unnecessary and can expose internal development artefacts.

## Fix

Quick fix: https://github.com/SAP/open-ux-tools/pull/4757/changes/b275e0082a5cea14b14aea79b949c5bdbeeef765

The gap analysis will address this functionality to ensure S4 and the generators properly align on how resources can be excluded.

longieirl:
Fixes #4756

#  Summary                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
  Fixes a silent misalignment between build-time and deploy-time exclusions: builder.resources.excludes in ui5-deploy.yaml was correctly set but never read by the deploy task, which maintained its own hardcoded configuration.exclude list. As a result, folders like localService/ were excluded at build time but still uploaded during ABAP deployment.                                                                                                                                        

Sample deploy config;

https://github.com/SAP-samples/fiori-tools-samples/blob/main/V4/apps/travel/ui5-deploy.yaml

Ensures all locations `builder.customtasks[deploy-to-abap].configuration.exclude` and `builder.resources.excludes` are read, ensures there is backward compatibility for older `ui5-deploy.yaml` configurations as well.                                                                                                                                                 
                                                                                    
  Changes:                                                                                                                                                                                                                                                                                                                  
                                                                                    
  - @sap-ux/ui5-config: adds getBuilderResourceExcludes() to read builder.resources.excludes from a UI5Config instance                                                                                                                                                                                                      
  - @sap-ux/ui5-config: addAbapDeployTask no longer hardcodes /test/ and /localService/ in configuration.exclude
  - @sap-ux/abap-deploy-config-writer: removes hardcoded configuration.exclude from getDeployConfig                                                                                                                                                                                                                         
  - @sap-ux/deploy-tooling: merges builder.resources.excludes from the project's UI5 config with any explicit configuration.exclude entries at runtime, so the deploy task stays in sync with build-time exclusions automatically                                                                                           
                                                                                                                                                                                                                                                                                                                            
 # Test plan                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                            
  - pnpm --filter @sap-ux/ui5-config test                                                                                                                                                                                                                                                                                   
  - pnpm --filter @sap-ux/abap-deploy-config-writer test                            
  - pnpm --filter @sap-ux/deploy-tooling test                                                                                                                                                                                                                                                                               
  - pnpm validate:changesets


packages/deploy-tooling/src/cli/archive.ts
```diff
@@ -48,19 +48,25 @@ async function fetchArchiveFromUrl(logger: Logger, url: string, rejectUnauthoriz
...
  */
-function createArchiveFromFolder(logger: Logger, path: string): Promise<Buffer> {
+function createArchiveFromFolder(logger: Logger, path: string, exclude: string[] = []): Promise<Buffer> {
     try {
         logger.info(`Creating archive from ${path}.`);
         const zip = new ZipFile();
         zip.addLocalFolder(path);
         for (const entry of zip.getEntries()) {
-            logger.debug(`Adding ${entry.entryName}`);
+            if (exclude.some((pattern) => new RegExp(pattern, 'g').exec(entry.entryName))) {
```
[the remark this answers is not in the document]
longieirl: Resolved. Current code already uses `startsWith` with leading-slash stripping — the `RegExp` approach was replaced before this review was posted. See `archive.ts:65-66`.

packages/ui5-config/src/ui5config.ts
```diff
@@ -675,14 +692,30 @@ export class UI5Config {
...
+        const builderExcludes = this.getBuilderResourceExcludes().map((p) => {
+            let prefix = p;
+            while (prefix.endsWith('*')) {
+                prefix = prefix.slice(0, -1);
+            }
+            if (prefix !== p && prefix.endsWith('/')) {
+                prefix = prefix.slice(0, -1);
+            }
+            const withSlash = prefix.endsWith('/') ? prefix : `${prefix}/`;
+            return withSlash.startsWith('/') ? withSlash : `/${withSlash}`;
```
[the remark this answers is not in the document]
longieirl: Resolved. The glob normalisation logic lives in `deploy-tooling/src/cli/config.ts:readBuilderExcludes` (not in `ui5config.ts`). The `while endsWith('*')` loop has been replaced with `p.replace(/\/\*+$/, '/')` which only strips trailing `*` when preceded by `/`, leaving patterns like `/assets/my*` intact.

packages/deploy-tooling/src/cli/config.ts
```diff
@@ -183,6 +183,7 @@ export async function mergeConfig(taskConfig: AbapDeployConfig, options: CliOpti
     config.createTransport = mergeFlag(options.createTransport, taskConfig.createTransport);
     config.retry = process.env.NO_RETRY ? !process.env.NO_RETRY : mergeFlag(options.retry, taskConfig.retry);
     config.lrep = options.lrep;
+    config.exclude = options.exclude ?? taskConfig.exclude;
```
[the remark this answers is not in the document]
longieirl: Resolved. The current implementation performs a full three-way merge: CLI `--exclude` + YAML `configuration.exclude` + `builder.resources.excludes` — all deduplicated via `Set`. No patterns from any source are dropped.

.changeset/gap2-builder-resource-excludes-derive.md
```diff
@@ -0,0 +1,7 @@
+---
+"@sap-ux/ui5-config": minor
+"@sap-ux/abap-deploy-config-writer": patch
+"@sap-ux/deploy-tooling": patch
+---
+
+fix: derive configuration.exclude from builder.resources.excludes at runtime; remove hardcoded excludes from generated configs
```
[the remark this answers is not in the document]
longieirl: Resolved. Changeset summary now starts with `FIX:` — `pnpm validate:changesets` passes.

longieirl:
> the glob-normalisation logic in `addAbapDeployTask` can silently mangle non-standard patterns like `/assets/my*`, and the CLI `--exclude` flag's interaction with YAML-configured excludes (override vs. additive) is ambiguous and potentially lossy

These comments have beeb addressed.

packages/deploy-tooling/src/ui5/index.ts
```diff
@@ -55,7 +55,7 @@ async function task({ workspace, options }: TaskParameters<AbapDeployConfig>): P
         logger,
         workspace,
         options.projectNamespace ?? options.projectName,
-        config.exclude
+        config.exclude ?? []
```
nikmace: Could you validate if this concern is true or not?

The fix only covers the CLI path; the UI5-tooling task path is left broken, re-introducing the exact #4756 regression for ui5 build deploys. There are two independent deploy entry points (README deploy-tooling/README.md:117 says so explicitly):
- CLI (runDeploy → mergeConfig, cli/config.ts:466) — now derives excludes from builder.resources.excludes. ✓
- UI5 task (ui5/index.ts:task(), registered as the abap-deploy-task via ui5.yaml task.path: dist/ui5/index.js) — still uses config.exclude ?? [] where config = options.configuration, i.e. only the configuration.exclude block. It never calls readBuilderExcludes / getBuilderResourceExcludes.

packages/deploy-tooling/src/cli/archive.ts
```diff
@@ -48,19 +48,29 @@ async function fetchArchiveFromUrl(logger: Logger, url: string, rejectUnauthoriz
...
         logger.info(`Creating archive from ${path}.`);
         const zip = new ZipFile();
         zip.addLocalFolder(path);
         for (const entry of zip.getEntries()) {
-            logger.debug(`Adding ${entry.entryName}`);
+            if (
+                exclude.some((pattern) =>
+                    entry.entryName.startsWith(pattern.startsWith('/') ? pattern.slice(1) : pattern)
+                )
+            ) {
```
nikmace, line 68: Two divergent matching engines consume the same normalized exclude array. 

`cli/archive.ts:createArchiveFromFolder` matches with `entryName.startsWith(pattern without leading /)` (string prefix). `ui5/archive.ts:createUi5Archive` matches with `RegExp(pattern, 'g').exec(path)` (regex). The values produced by `readBuilderExcludes` (e.g. /test/, /localService/) happen to work as both, but any pattern containing regex metacharacters (., +, $, () behaves differently — or throws — on the build-task path while working fine on the CLI path. A pattern like `/assets/my*` becomes an invalid/greedy regex here. Same config, two behaviours, is a latent trap.

packages/deploy-tooling/src/cli/config.ts
```diff
@@ -159,6 +159,27 @@ function mergeCredentials(taskConfig: AbapDeployConfig, options: CliOptions) {
...
+        const content = readFileSync(deployConfigPath, 'utf-8');
+        const ui5Config = await UI5Config.newInstance(content);
+        return ui5Config.getBuilderResourceExcludes().map((p) => {
+            const prefix = p.replace(/\/\*+$/, '/');
+            const withSlash = prefix.endsWith('/') ? prefix : `${prefix}/`;
+            return withSlash.startsWith('/') ? withSlash : `/${withSlash}`;
+        });
+    } catch {
+        return [];
+    }
```
nikmace: `readBuilderExcludes` swallows all read/parse errors to []. 

The outer try/catch returns [] on any failure (missing file, malformed YAML, permission error). Combined with the derive-only-on-CLI design, a corrupt `ui5-deploy.yaml` silently degrades to "exclude nothing" rather than surfacing a warning — no `logger.debug/warn`. Consider logging when the config path was provided but yielded nothing.

nikmace (changes requested):

packages/deploy-tooling/src/ui5/index.ts
longieirl: Valid concern, I will address this. Thanks.

packages/deploy-tooling/src/cli/archive.ts
longieirl: Valid concern, I will address this. Thanks.

packages/deploy-tooling/src/cli/config.ts
longieirl: Valid concern, I will address this. Thanks.

PR #4831, opened 2026-06-12
fix(eslint-plugin-fiori-tools): remove deleted flex change on reparse

AlinaGovoruhina:
# Fix: Remove Deleted Flex Change Files on Reparse and Watch Newly Created `.change` Files

### Bug Fix

🐛 Fixed an issue where deleted flex change files were not being removed from an app's `changes` array during reparse operations. Also ensures newly created `.change` files are properly detected and indexed by forcing a reindex on first update when the file is new to the cache.

### Changes

* `.changeset/itchy-glasses-beam.md`: Added changeset entry documenting the patch fix for detecting property change file deletion and watching newly created `.change` files.

* `parser.ts`:
  - Fixed incorrect path construction: replaced `join(app.appRoot, 'webapp', 'changes')` with `join(webappPath, 'changes')` to correctly locate the changes directory.
  - Replaced inline flex change reading logic with a call to the new `collectFlexChanges` utility.
  - Added logic in `reparseChange` to filter out deleted change files from the app's `changes` array (using `existsSync`) before processing, preventing stale entries from persisting.
  - Added explicit type cast for parsed JSON content (`as FlexChange`).

* `utils.ts`: Introduced the new exported `collectFlexChanges` utility function that reads `.change` file paths, parses their JSON content, and returns an array of `FlexChange` objects — centralizing previously duplicated logic with error handling for unreadable or malformed files.

* `project-context.ts`: Added detection for newly created files in `updateFile` — if a file URI is not yet in the cache, `forceReindexOnFirstUpdate` is set to `true` to ensure the new file is properly indexed.

* `test/project-context/parser/parser.test.ts`: Added new test file covering flex change reparse behavior — verifying that new `.change` files are added and that entries for deleted files are removed from the changes array.

* `test/project-context/parser/service.test.ts`: Minor import cleanup — consolidated `ServiceArtifacts` and `V2Annotation` into a single import from `@sap-ux/fiori-annotation-api`.

<details>

**Version:** `1.24.0`

- Correlation ID: `41d0712c-9210-4472-aa93-d01c5a2a087f`
</details>


igorsVoronins (dismissed):
Tested the bug fix on the V2 Fiori project using the App Modeler extension - looks good ✅

Removing the .change file no longer disrupts the extension’s functionality. After deletion, ESLint warnings in the Page Map and Editor behave as expected, and the Application Info page correctly updates and increments/decrements the issue count.

packages/eslint-plugin-fiori-tools/src/project-context/utils.ts
```diff
@@ -11,3 +14,28 @@ export function getParsedServiceByName(parsedApp: ParsedApp, serviceName?: strin
...
+            const fileContent = readFileSync(changeFile, { encoding: 'utf8', flag: 'r' });
+            const jsonContent = JSON.parse(fileContent) as FlexChange;
+            changes.push({
+                changeType: jsonContent.changeType,
+                content: jsonContent.content,
+                selector: jsonContent.selector,
+                changeFileUri: pathToFileURL(changeFile).toString()
+            });
+        } catch {
+            // skip unreadable or malformed change files
```
marcovieth, line 37: Do we really want to suppress all errors and not even show a console warning?

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
```diff
@@ -214,15 +204,27 @@ export class ApplicationParser {
...
     private reparseChange(uri: string, index: ParsedProject, fileCache: Map<string, string>): void {
-        const content = fileCache.get(uri) ?? '';
+        // Remove deleted files from app changes array
+        for (const key of Object.keys(index.apps)) {
+            const existingChanges = index.apps[key].changes.filter((change) => {
+                const path = fileURLToPath(change.changeFileUri);
+                return existsSync(path);
+            });
+            index.apps[key].changes = existingChanges;
+        }
```
marcovieth: Does this scan all apps on every change event?
Would it make sense to check the file cache first fileCache.get(uri);
And if that fails, remove only this entry from all apps? change.changeFileUri !== uri

marcovieth (commented):
Please check my comments.
Do we also need a test ?

packages/eslint-plugin-fiori-tools/src/project-context/utils.ts
AlinaGovoruhina: Update this to skip malformed files and continue reading next .change files. I think it makes sense as the ESLint VSCode extension also doesn't validate if the file is bad format.

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
AlinaGovoruhina: Index always contains one app, even if you work with several apps in one workspace. Maybe the name is confusing? 
As for uri the function receives - it is always an existing file, deleted files are not reparsed, so I added a filter for deleted ones at the beginning of this function.

marcovieth: Ok, thanks for the explanation.
Why moving parts inside the loop, which are independent of the loop (parseJson, JSON.parse)?
Or why using a loop at all if we process just one app?

packages/eslint-plugin-fiori-tools/src/project-context/utils.ts
marcovieth: Ok, if it is common practise to suppress such issues.

packages/eslint-plugin-fiori-tools/src/project-context/utils.ts
```diff
@@ -11,3 +14,29 @@ export function getParsedServiceByName(parsedApp: ParsedApp, serviceName?: strin
...
+            continue;
```
marcovieth: This "commit" is redundant.

marcovieth (commented):
It seems that forceReindexOnFirstUpdate is never reset.

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
AlinaGovoruhina: So I checked a CAP project and several apps are used in index there. Flex change files are used only in V2, so will have a single app always and in this case it doesn't matter - it loops apps 1 time. Should I remove loop to not be confusing? reparseXML also goes through apps in index, but I'd rather not touch it here.

packages/eslint-plugin-fiori-tools/src/project-context/utils.ts
AlinaGovoruhina: Can you elaborate about why this catch block is redundant?

marcovieth: Sorry, I meant the "continue", because no code follows. You can also ignore this comment.

AlinaGovoruhina: Ok, I agree, I removed this code

marcovieth (approved):
- forceReindexOnFirstUpdate is set to false now
- tests are there
looks good to me.

igorsVoronins (approved):
Re-tested the PR against the V2 project. Ok to merge ✅ 
The integration with the Application Modeler extension behaves as expected. Change files are properly created and deleted based on the state defined in the Page Map/Editor, and the synchronization works reliably in both directions.

Issue #4840, opened 2026-06-15
TBI - Skip package.json updates in enableCardGeneratorConfig


docirl:
### Description (include screenshots)
Enable `enableCardGeneratorConfig` to be called and not update the package.json. There are flows, such as the S/4 application generator, that require the config added to the yaml files in enableCardGeneratorConfig but should not update the package.json, specifically for mock and local yaml files. Re-use of the yaml updating is needed standalone without package.json updates.

Any changes will be opt in, with a param needed to be set to skip the package.json updates.

### Value
Optional - details on how this technical backlog item provides value to end-users or dev team eg. Performance, Quality, Maintainability. This will help determine the ranking in relation to feature development

### Architecture Elaboration
Does it requires architecture elaboration?

### Tasks
- [ ] Task 1

giruuuuj:
hi their can solve issue

docirl:
Changed merged

PR #4842, opened 2026-06-15
tbi(app-config-writer): skip package updates in enableCardGeneratorConfig

docirl:
# feat(app-config-writer): Skip package.json updates in `enableCardGeneratorConfig`

### New Feature

✨ Added an optional `updatePackage` parameter to `enableCardGeneratorConfig` that allows callers to skip `package.json` updates while still applying YAML configuration changes. This is useful for flows that require card generator middleware config in YAML files (e.g., mock and local YAML files) without modifying `package.json`. The parameter defaults to `true`, making this change fully opt-in and backward-compatible.

Additionally, an `exports` map with a `./cards-config` subpath has been added to `package.json` for direct module access.

### Export of card-config

`@sap-ux/app-config-writer`'s main entry point (`dist/index.js`) re-exports from `dist/eslint-config/add.js`, which imports `addEslintFeature` from`@sap-ux/ui5-application-writer` — a CJS package. When loaded as ESM, Node cannot resolve named exports from a CJS module via static import bindings, so the entire package entry point fails to load.

`dist/cards-config/index.js` has no such dependency and loads cleanly, but without a `./cards-config` entry in the exports map it is not accessible as a stable public subpath. Consumers are forced to either import the broken main entry point, or reference the internal `dist/` path directly — which is an implementation detail that can change without notice.

Adding `./cards-config` to exports gives consumers a stable, supported way to import enableCardGeneratorConfig in isolation, bypassing the broken eslint-config dependency chain, until the CJS/ESM interop issue in that part of the package is resolved.

### Changes

* `.changeset/shy-camels-travel.md`: Added changeset entry describing the new feature as a patch release.
* `packages/app-config-writer/package.json`: Added an `exports` map exposing the main entry point and a new `./cards-config` subpath export.
* `packages/app-config-writer/src/cards-config/index.ts`: Added `updatePackage` parameter (defaults to `true`) to `enableCardGeneratorConfig`; wrapped `updatePackageJson` call in a conditional check so it is skipped when `updatePackage` is `false`.
* `packages/app-config-writer/test/unit/cards-config/index.test.ts`: Added a test case verifying that when `updatePackage=false`, `package.json` remains unchanged while `ui5.yaml` is still updated.
* `packages/app-config-writer/test/unit/cards-config/__snapshots__/index.test.ts.snap`: Added snapshot for the new `updatePackage=false` test case.

### Jira Issues

* [GitHub #4840](https://github.com/SAP/open-ux-tools/issues/4840): TBI - Skip package.json updates in `enableCardGeneratorConfig`

<details>

**Version:** `1.26.4`

- Correlation ID: `d7ceafa0-a02b-4de3-8c5c-e0a447a2ff93`
</details>


longieirl (dismissed):
Changeset is good.
Change is good.
Test is good.

heimwege:
@docirl what's the use case for this (#4840 talks about specific flows w/o actually naming one) and why do we only add this feature for this command and not all commands that adjust package.json (e.g. variants management)?

.changeset/shy-camels-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ux/app-config-writer': patch
```
heimwege: I suggest to use minor for new features

```suggestion
'@sap-ux/app-config-writer': minor
```

packages/app-config-writer/package.json
```diff
@@ -13,6 +13,16 @@
     },
     "license": "Apache-2.0",
     "main": "dist/index.js",
+    "exports": {
```
heimwege, line 16: Adding an `exports` field to a package that previously only had `main` is a potentially breaking change for any consumer that deep-imports an unlisted path. We should mention this in the changeset as well, but I still think `minor` bump is sufficient.

heimwege (commented):
see comments

docirl:
> @docirl what's the use case for this (#4840 talks about specific flows w/o actually naming one) and why do we only add this feature for this command and not all commands that adjust package.json (e.g. variants management)?

There is a bug in the S/4 app generator, I'll provide a link separately for this. The issue details the bug that I've found (and also seen by Markus). There is specific handling in S/4 for this script and the yaml updates. I'm not aware of issues with the other scripts.

packages/app-config-writer/package.json
docirl: Ok thanks, I'll add this to the changeset

heimwege (approved):
- changeset ok
- review comments addressed
- coverage is superb
- did NOT test manually

longieirl (approved):
Changeset is good, extra detail has been added but that is accepted.
CHange is good.
Tests are good.

docirl:
thanks @heimwege @longieirl

PR #4864, opened 2026-06-19
Feat/Add Fragment Generation Support

Vladislavs-Silins-SAP:
## Summary
Add standalone fragment generation capability to fe-fpm-writer and specification packages for use in table columns and other UI customizations.

### open-ux-tools (feat/generate-fragment-function)
- Add `generateFragment()` function in `fe-fpm-writer/src/column/index.ts`
- Add `Fragment` interface in `fe-fpm-writer/src/column/types.ts` following existing type patterns
- Export Fragment type and generateFragment function from package index

### specification (integration branch)
- Add `CustomExtensionType.Fragment` enum value in `packages/types/src/apiTypes.ts`
- Import `Fragment` type from `@sap-ux/fe-fpm-writer`
- Register `generateFragment` in `fpmWriterMap` in `packages/specification/src/api.ts`

## API

```typescript
// Fragment generation
const fragment: Fragment = {
    name: 'MyFragment',
    folder: 'ext/fragment',  // optional, defaults to 'ext/fragment'
    content: '<Text text="..." />'  // optional, uses default template
};

await generateCustomExtension(basePath, CustomExtensionType.Fragment, fragment, fs);
```

## Notes
- Fragment interface follows same pattern as CustomTableColumn, CustomAction, etc.
- Generates standalone `.fragment.xml` files without modifying manifest.json
- Enables dynamic fragment generation for table columns and other custom UI elements


packages/fe-fpm-writer/src/column/index.ts
```diff
@@ -69,6 +69,43 @@ function enhanceConfig(
...
+
+    // Calculate path for fragment
+    const folder = fragment.folder || 'ext/fragment';
+    const path = join(dirname(manifestPath), folder);
+
+    // Generate default content if not provided
+    const content = fragment.content || getDefaultFragmentContent('Sample Text', fnGenerateId);
+
+    // Create fragment file
+    const viewPath = join(path, `${fragment.name}.fragment.xml`);
```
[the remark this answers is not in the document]
Vladislavs-Silins-SAP: Fragment name validation is handled at the UI level during user input - empty or invalid names cannot reach this function. Additional validation here would be redundant.

packages/fe-fpm-writer/src/column/index.ts
```diff
@@ -69,6 +69,43 @@ function enhanceConfig(
...
+    if (!fs.exists(viewPath)) {
```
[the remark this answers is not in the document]
Vladislavs-Silins-SAP: This is intentional behavior consistent with other generators in this codebase (e.g., generateCustomColumn at line 130). The pattern is: if the file already exists, don't overwrite it. This prevents accidental data loss.
  Callers can check fs.exists(path) before calling if they need to detect this case. The behavior could be documented in the JSDoc if clarification is needed.

packages/fe-fpm-writer/src/column/index.ts
```diff
@@ -69,6 +69,43 @@ function enhanceConfig(
...
+ * @returns {Promise<Editor>} the updated mem-fs editor instance
+ */
+export async function generateFragment(
+    basePath: string,
+    fragment: Fragment,
+    fs?: Editor
+): Promise<Editor> {
+    fs ??= create(createStorage());
+    await validateBasePath(basePath, fs);
+    const fnGenerateId = await createIdGenerator({ basePath, fsEditor: fs });
```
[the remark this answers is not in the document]
Vladislavs-Silins-SAP: Good catch. However, the current implementation matches the existing pattern in generateCustomColumn (line 121) which also calls createIdGenerator unconditionally before using it in copyTpl. While the optimization you
  suggest is valid, changing this would create an inconsistency with other generators in this codebase. The performance impact is minimal in practice since this is a generation-time operation, not runtime. If optimization
  is desired, it should be applied consistently across all generators, not just this one.

.changeset/floppy-planes-sniff.md
```diff
@@ -0,0 +1,100 @@
...
+'@sap-ux/ui5-library-reference-writer': patch
+'@sap-ux/ui5-library-sub-generator': patch
+'@sap-ux/ui5-library-writer': patch
+'@sap-ux/ui5-proxy-middleware': patch
+'@sap-ux/ui5-test-writer': patch
+'@sap-ux/xml-odata-annotation-converter': patch
+'@sap-ux/yaml': patch
+'test-project': patch
+'@sap-ux-private/adaptation-editor-tests': patch
+'@sap-ux/types': patch
```
heimwege: Changes are only done in fe-fmp-writer. Dependencies will be handled by the workspace links

```suggestion
'@sap-ux/fe-fpm-writer': minor
```

packages/fe-fpm-writer/src/column/types.ts
```diff
@@ -64,3 +64,21 @@ export interface CustomTableColumn extends CustomElement, EventHandler, CustomFr
 export interface InternalCustomTableColumn extends CustomTableColumn, InternalCustomElement {
     content: string;
 }
+
+/**
+ * Fragment generation configuration.
+ */
+export interface Fragment {
```
heimwege: The Fragment interface is placed in `column/types.ts`, but it has no conceptual relationship to columns - it's a general concept. This could live in a dedicated `fragment/types.ts` or `common/types.ts`.

packages/fe-fpm-writer/test/unit/column.test.ts
```diff
@@ -349,4 +349,82 @@ describe('CustomAction', () => {
...
+    describe('generateFragment', () => {
+        let fs: Editor;
+        const testFragmentDir = join(__dirname, 'sample/column');
+
+        beforeEach(() => {
+            fs = create(createStorage());
+        });
+
+        test('generate fragment with default options', async () => {
+            const { generateFragment } = await import('../../src/index.js');
```
heimwege: Tests import generateFragment via dynamic await import('../../src/index.js') inside each test body - this is inconsistent with how the rest of the test file imports. Other tests in the file use static imports at the top. The dynamic imports here appear unnecessary.

packages/fe-fpm-writer/src/column/index.ts
```diff
@@ -69,6 +69,39 @@ function enhanceConfig(
...
+ */
+export async function generateFragment(basePath: string, fragment: Fragment, fs?: Editor): Promise<Editor> {
+    fs ??= create(createStorage());
+    await validateBasePath(basePath, fs);
+    const fnGenerateId = await createIdGenerator({ basePath, fsEditor: fs });
+
+    const { path: manifestPath } = await getManifest(basePath, fs);
+
+    // Calculate path for fragment
+    const folder = fragment.folder || 'ext/fragment';
```
heimwege: The folder parameter is used directly in join() with no sanitization. A caller passing folder: `'../../../etc'` would write outside the project.

packages/fe-fpm-writer/src/column/index.ts
```diff
@@ -69,6 +69,39 @@ function enhanceConfig(
...
+ * Generate a standalone fragment file with default content.
+ * Reuses the same template as custom columns: <Text text="Sample Text" />
+ * Does NOT modify manifest.json - fragment-only generation.
+ *
+ * @param {string} basePath - the base path
+ * @param {Fragment} fragment - fragment generation configuration
+ * @param {Editor} [fs] - the mem-fs editor instance
+ * @returns {Promise<Editor>} the updated mem-fs editor instance
+ */
+export async function generateFragment(basePath: string, fragment: Fragment, fs?: Editor): Promise<Editor> {
```
heimwege: Why is this nested under `/column`? It's not just fragments for table columns, right? Then it should be e.g. in `/fragment` or `/common`

heimwege (commented):
see comments

.changeset/floppy-planes-sniff.md
Vladislavs-Silins-SAP: Done.

packages/fe-fpm-writer/src/column/types.ts
Vladislavs-Silins-SAP: Moved to fragment/types.ts, creating a dedicated folder for fragment functionality.

packages/fe-fpm-writer/test/unit/column.test.ts
Vladislavs-Silins-SAP: Removed unnecessary dynamic imports in tests - now using static imports at the top, consistent with other test files.

packages/fe-fpm-writer/src/column/index.ts
Vladislavs-Silins-SAP: Moved generateFragment to dedicated fragment/ directory since it generates general fragments, not just column fragments.

packages/fe-fpm-writer/src/column/index.ts
Vladislavs-Silins-SAP: Added validateRelativePath to common/validate.ts to prevent path traversal attacks - checks for absolute paths and '..' segments. I used the generateCustomColumn() function in the column as a sample, there should be the same problem there.

heimwege: @devinea should we make use of that `validateRelativePath` function to avoid path-traversal also in other places (maybe in a separate PR)? 

There are 10+ affected locations across the codebase. The root cause is one choke-point:

`common/defaults.ts:25` - `setCommonDefaults()` calls `join(dirname(manifestPath), config.folder)` with no validation. Every generator that goes through `setCommonDefaults` inherits the vulnerability.
Separate category - not via `setCommonDefaults`: `building-block/index.ts`

Recommended fix: Add the `validateRelativePath` call inside `setCommonDefaults` itself (at `defaults.ts:25`) - that would fix all CustomElement-based generators in one place. `building-block/index.ts` needs its own fix since it bypasses that function.

packages/fe-fpm-writer/src/fragment/index.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Does NOT modify manifest.json - fragment-only generation.
+ *
+ * @param {string} basePath - the base path
+ * @param {Fragment} fragment - fragment generation configuration
+ * @param {Editor} [fs] - the mem-fs editor instance
+ * @returns {Promise<Editor>} the updated mem-fs editor instance
+ */
+export async function generateFragment(basePath: string, fragment: Fragment, fs?: Editor): Promise<Editor> {
+    fs ??= create(createStorage());
+    await validateBasePath(basePath, fs);
```
heimwege, line 27: Any project without `sap.fe.templates` (freestyle apps, non-FPM projects) will get an error. This information is missing in the JSDoc. Maybe it's worth adding some `@throws` to the JSdoc to make this transparent?

```
@throws {Error} if the project folder is invalid or manifest.json cannot be found
@throws {Error} if sap.fe.templates is missing from manifest.json dependencies
@throws {Error} if fragment.folder is an absolute path or contains path traversal sequences
```

packages/fe-fpm-writer/src/fragment/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    // Calculate path for fragment with validation
+    const folder = fragment.folder ?? 'ext/fragment';
+    validateRelativePath(folder, 'Fragment folder');
+    const path = join(dirname(manifestPath), folder);
+
+    // Generate default content if not provided
+    const content = fragment.content ?? getDefaultFragmentContent('Sample Text', fnGenerateId);
+
+    // Create fragment file
+    const viewPath = join(path, `${fragment.name}.fragment.xml`);
```
heimwege, line 42: I think we have the same path traversal problem here if `fragment.name` is `../../evil`

heimwege (commented):
see comments

packages/fe-fpm-writer/src/fragment/index.ts
Vladislavs-Silins-SAP: Done.

packages/fe-fpm-writer/src/fragment/index.ts
Vladislavs-Silins-SAP: Added the same `validateRelativePath` check for `fragment.name` to prevent path traversal attacks (e.g., `name: '../../etc/passwd'`).

  I initially tried validating the final combined path once, but it complicated the code significantly (needed `relative()` calls and reasoning about webapp vs basePath). Validating each part separately is simpler and
  equally secure — single dots are harmless (normalized by `join()`), only `..` is dangerous.

packages/fe-fpm-writer/src/column/index.ts
devinea: @heimwege Yeah seems like a good idea. 
If applying a similar approach on templates ensure it is compatible with bundling approach. 
FYI @815are

heimwege (approved):
- review comments addressed
- changeset ok
- coverage is good
- did NOT test manually

@Vladislavs-Silins-SAP can you please also create an internal TBI for following-up on https://github.com/SAP/open-ux-tools/pull/4864#discussion_r3443261482? I think there's no need to block this PR or include it here, but we should follow up.

815are (approved):
checked:
1. code changes - code changes are straightforward 
2. tests are added
3. changelog persists

I did not test manually

PR #4865, opened 2026-06-19
fix: Revert getPackageJSONInfo to previous approach, removing getTemplatePath dependency

815are:
Change related to PR -> https://github.com/SAP/open-ux-tools/pull/4835
during bundling we noticed that one line was wrongly changed:
image

Original logic does not reference to `template` folder - then there is not reason to use `getTemplatePath` fucntion.
In result - in this PR reverting line in `getPackageJSONInfo` to use previous approach to avoid regressions across consumers

### DIFF
--- a/.changeset/stale-swans-retire.md
+++ b/.changeset/stale-swans-retire.md
@@ -1,0 +1,5 @@
+---
+'@sap-ux/adp-tooling': patch
+---
+
+FIX: Revert getPackageJSONInfo to previous approach, removing getTemplatePath dependency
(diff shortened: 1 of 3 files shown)


vadson71 (approved):
Straightforward changes

devinea (approved):
Looks good.

mmilko01:
image

mmilko01 (approved):
Changeset OK
Changes look good|
Excellent test coverage
Did not test manually, but integration tests are successful

PR #4866, opened 2026-06-19
feat(scripts): cascade changeset validation for esbuild-bundling packages

devinea:
## Summary

Extends \`validate-changesets.mjs\` to enforce cascade changesets when a workspace package bundled via esbuild is being released, preventing consumers from running stale bundled code.

### How it works

- \`ESBUILD_BUNDLING_PACKAGES\` declares which packages bundle their deps (\`@sap-ux/fiori-mcp-server\`, \`sap-ux-sap-systems-ext\`, \`@sap-ux/eslint-plugin-fiori-tools\`)
- For each bundler, the **transitive workspace dependency graph is walked** with the correct rules:
  - The bundler's own \`devDependencies\` are included (esbuild reads them directly from \`node_modules\`)
  - For transitive deps, only their \`dependencies\` are followed — their \`devDependencies\` are not installed in the bundler's \`node_modules\` tree and are never bundled
- If any dep in this set has a changeset but the bundling consumer does not, validation fails with a clear, actionable error message
- Bot-generated changesets (e.g. Renovate \`dependencies-GH-*.md\`) are included in the cascade check — packages are collected before the bot-skip \`continue\`, so Renovate-triggered releases of bundled deps are caught
- Updates \`AGENTS.md\` with documentation on the bundled devDep cascade rule and how to maintain the config

### Problem solved

Packages like \`fiori-mcp-server\` use esbuild to inline their dependency graph at build time. Changesets ignores devDependency version changes by default, so a new release of e.g. \`@sap-ux/fiori-docs-embeddings\` (or any transitive dep like \`@sap-ux/yaml\` pulled in via \`@sap-ux/project-access\`'s dependencies) would never trigger a rebuild and republish of \`fiori-mcp-server\` — consumers would keep running stale bundled code.

### Example error output

\`\`\`
🚫 Changeset validation failed:

❌ Missing cascading changeset for "@sap-ux/fiori-mcp-server"
   Reason: "@sap-ux/fiori-docs-embeddings" is being released and "@sap-ux/fiori-mcp-server" bundles it via esbuild.
   The bundler must be re-published so consumers receive the updated bundle.
   Fix: Add a changeset for "@sap-ux/fiori-mcp-server" (patch bump, BUMP: prefix):

     ---
     "@sap-ux/fiori-mcp-server": patch
     ---

     BUMP: Rebuild bundle with updated @sap-ux/fiori-docs-embeddings
\`\`\`

## Test plan

- [ ] \`pnpm validate:changesets\` passes with no pending changesets
- [ ] Adding a changeset for \`@sap-ux/fiori-docs-embeddings\` (direct devDep of bundler) without one for \`@sap-ux/fiori-mcp-server\` causes validation to fail
- [ ] Adding a changeset for \`@sap-ux/yaml\` (transitive via \`project-access\` dependencies) without bundler changesets also fails
- [ ] Adding a changeset for \`@sap-ux/create\` (a devDep of \`project-access\`, not bundled) passes cleanly
- [ ] A Renovate-style \`dependencies-GH-*.md\` changeset for a bundled dep triggers the cascade check
- [ ] Adding all required cascade changesets causes validation to pass


mikicvi-SAP (commented):
Thanks @devinea 
- changes lg, I tested locally, and it seems to be working OK.
- Changeset  ❌ according to AGENTS.md

---
- Changeset summary should use `BUMP:` prefix instead of `FIX:` (AGENTS.md says "`FIX:` is for Bug fixes, security patches, CVE upgrades" and "`BUMP:` is for Dependency-only upgrades with no behaviour change"). This is a cascade rebuild trigger, not a bug fix. The PR's own error message at line 170 prescribes `BUMP:` for exactly this scenario, and AGENTS.md example (lines 312-319) shows the correct pattern.

- Changeset summary includes the package name `@sap-ux/fiori-mcp-server`, but AGENTS.md says "Do **not** include the package name in the summary — it is already declared in the frontmatter."

heimwege:
> Thanks @devinea
> 
> * changes lg, I tested locally, and it seems to be working OK.
> * Changeset  ❌ according to AGENTS.md
> 
> * Changeset summary should use `BUMP:` prefix instead of `FIX:` (AGENTS.md says "`FIX:` is for Bug fixes, security patches, CVE upgrades" and "`BUMP:` is for Dependency-only upgrades with no behaviour change"). This is a cascade rebuild trigger, not a bug fix. The PR's own error message at line 170 prescribes `BUMP:` for exactly this scenario, and AGENTS.md example (lines 312-319) shows the correct pattern.
> * Changeset summary includes the package name `@sap-ux/fiori-mcp-server`, but AGENTS.md says "Do **not** include the package name in the summary — it is already declared in the frontmatter."

@devinea if it is just about re-triggering the publishing process you can use `INFRA:` that does not add any message to the changeset (as requested by @donal-tobin-sap).

devinea:
> > Thanks @devinea
> > 
> > * changes lg, I tested locally, and it seems to be working OK.
> > * Changeset  ❌ according to AGENTS.md
> > * Changeset summary should use `BUMP:` prefix instead of `FIX:` (AGENTS.md says "`FIX:` is for Bug fixes, security patches, CVE upgrades" and "`BUMP:` is for Dependency-only upgrades with no behaviour change"). This is a cascade rebuild trigger, not a bug fix. The PR's own error message at line 170 prescribes `BUMP:` for exactly this scenario, and AGENTS.md example (lines 312-319) shows the correct pattern.
> > * Changeset summary includes the package name `@sap-ux/fiori-mcp-server`, but AGENTS.md says "Do **not** include the package name in the summary — it is already declared in the frontmatter."
> 
> @devinea if it is just about re-triggering the publishing process you can use `INFRA:` that does not add any message to the changeset (as requested by @donal-tobin-sap).

This is about requiring mcp listed in changeset when embeddings is in changeset. when embeddings is released mcp server should be. But is it not as it is a devDependency due to esbuild bundling. So releases are being missed for these types of modules. 
The Prefix should be determined by the end user based on what is changing in embeddings package for example.

heimwege (approved):
Adjustment lgtm 👍🏼

mikicvi-SAP (approved):
Thanks for the explanation, LGTM.

PR #4873, opened 2026-06-22
fix(odata-service-inquirer): Force url revalidation

IainSAP:
## Summary

- Force URL revalidation on every validate call by passing `forceReValidation: true` to `connectValidator.validateUrl`, ensuring validation errors are always returned rather than relying on a cached result. This doesnt negatively affect number of network calls as the validate is only triggered on url change or system type change,

internal ref: 38418

## Problem

When navigating back to the URL prompt (e.g., after changing the system type), the connection validator could return a cached validity result without re-running validation — causing errors to be silently swallowed and the UI to show an incorrectly valid state.

## Test plan

- [ ] Verify that changing system type and re-entering a URL triggers a fresh validation (not a cached pass)
- [ ] Verify that a previously valid but now-stale URL correctly shows an error on re-entry
- [ ] Unit test updated: `abap-on-prem/questions.test.ts` asserts `forceReValidation: true` is passed to `validateUrl`

### DIFF
--- a/.changeset/odata-service-inquirer-force-url-revalidation.md
+++ b/.changeset/odata-service-inquirer-force-url-revalidation.md
@@ -1,0 +1,5 @@
+---
+"@sap-ux/odata-service-inquirer": patch
+---
+
+FIX: Force URL revalidation and reset connection state on system type change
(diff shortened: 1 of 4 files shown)


longieirl (approved):
Changeset is good.
Tests are good.
Change is good.

kjose90 (approved):
Looks good.
Changes, test - clear 
changeset ✅

cianmSAP (approved):
Code change is small 
Tests asserts the change
Changeset looks good

PR #4876, opened 2026-06-23
tbi(generators): pass logger instance to writers

docirl:
- pass logger instance to writers so it can be used by them


kjose90 (approved):
small change
looks good
test, changeset ✅

longieirl (approved):
Changeset is good.
Tests are good.
Change is good.

IainSAP (approved):
Thanks @docirl 

- Improves logging integration from YUI or CLI for writer logs

PR #4880, opened 2026-06-23
fix(odata-service-inquirer): Explicitly set connect type

IainSAP:
## Summary

- Explicitly set `connectType` to `'abap_catalog'` when no connection path is provided in `connectWithBackendSystem`, replacing the previous `undefined` fallback

## Problem

When `connectPath` was not provided, `connectType` was set to `undefined`, leaving the connection validator without an explicit type hint. This could cause the validator to apply incorrectly the previous connection type, skipping catalog-specific logic, leading to connection issues for ABAP on-premise systems without a specific service path. This does not affect the flows which rely on the previous setting behaviour (re-auth using creds for example)

## Test plan

- [ ] Verify ABAP on-premise system connection succeeds without a connection path (reentrance ticket and basic auth flows)
- [ ] Verify connection with an explicit `connectPath` still uses `'odata_path'`
- [ ] Confirm no regression in existing system selection tests

### DIFF
--- a/.changeset/odata-service-inquirer-explicit-connect-type.md
+++ b/.changeset/odata-service-inquirer-explicit-connect-type.md
@@ -1,0 +1,5 @@
+---
+"@sap-ux/odata-service-inquirer": patch
+---
+
+FIX: Explicitly set connect type to abap_catalog when no connection path is provided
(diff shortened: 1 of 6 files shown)


docirl (approved):
Changes look good
Tests added to cover
Changeset added

mikicvi-SAP (approved):
Thanks @IainSAP 
- change small and LG ✅
- test covers the behaviour 
- changeset LGTM

cianmSAP (approved):
Updates look good
Changeset and tests added

longieirl (approved):
Changeset is good.
Tests are good.
Overall change is good.

PR #4888, opened 2026-06-24
fix(fiori mcp): Add odata service lookup

IainSAP:
## Summary

Internal ref: 38628

- **`findService` replaces `findServicePath`** in [sap-system.ts](packages/fiori-mcp-server/src/tools/services/sap-system.ts): catalog lookup now matches on `ODataServiceInfo.id` (case-insensitive) instead of `name`, and returns the full `ODataServiceInfo` object rather than just the path string. The old `ServiceSuggestion` / `FindServicePathResult` types are removed; suggestions are plain `ODataServiceInfo[]`.
- **Catalog path used verbatim**: `servicePath = result.service.path` ensures the exact URL returned by the catalog is used for the metadata request — fixing malformed requests that previously concatenated the service name onto the host.
- **`serviceName` input support in `download_odata_service_metadata`**: callers can now pass a service technical id instead of a full path. The tool resolves it via `findService` and falls back to `servicePath` when both are supplied. `serviceName` is never passed to `findSystem` (it is not a system identifier).
- **Richer error hints**: when a service id is not found, the error message includes the candidate list with `name`, `serviceVersion`, `path`, and OData version so the caller can identify and supply the correct path directly.

## Test plan

- [x] `download_odata_service_metadata` with `serviceName` resolves the catalog path and fetches metadata successfully
- [x] `download_odata_service_metadata` prefers `servicePath` when both `servicePath` and `serviceName` are supplied
- [x] Unrecognised `serviceName` returns suggestions containing name, version, path, and OData version
- [x] `findService` finds a service by exact case-insensitive id match and returns the full `ODataServiceInfo`
- [x] `findService` returns suggestions matching by name prefix, substring, and reverse-containment
- [x] `findService` falls back to V2 catalog when V4 yields no match
- [x] `findService` continues gracefully when one catalog lookup fails
- [x] `pnpm --filter @sap-ux/fiori-mcp-server test` passes (414 tests)


packages/fiori-mcp-server/src/tools/services/sap-system.ts
```diff
@@ -268,6 +289,62 @@ function checkMetadata(metadata: string): void {
...
+    ]);
+
+    const allServices: Array<ODataServiceInfo> = [];
+
+    for (const result of results) {
+        if (result.status === 'rejected') {
+            logger.debug(`Catalog lookup failed: ${result.reason}`);
+            continue;
+        }
+        const exact = result.value.find((s) => s.id.toLocaleLowerCase() === nameLower);
```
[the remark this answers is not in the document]
IainSAP: Will update the doc, implementation is as expected

packages/fiori-mcp-server/src/tools/download-odata-service-metadata.ts
```diff
@@ -37,7 +38,7 @@ export async function downloadODataServiceMetadata(
     }
 
     try {
-        const findResult = await findSystem(sapSystemQuery || servicePath);
+        const findResult = await findSystem(sapSystemQuery || servicePath || serviceName);
```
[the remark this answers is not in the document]
IainSAP: will fix

vinayhospete (dismissed):
code changes looks good to me

packages/fiori-mcp-server/skills/sap-fiori-create-cli/SKILL.md
```diff
@@ -4,7 +4,7 @@ description: Run, invoke, and test the @sap-ux/create CLI — generate, add, con
 argument-hint: command and subcommand (e.g., add mockserver-config, generate adaptation-project)
 metadata:
   author: sap-fiori-tools
-  version: "1.0.34"
+  version: "1.0.36"
```
heimwege: I fixed the issue causing version conflict with this module. I think if you revert this change you should be good

packages/fiori-mcp-server/src/types/input.ts
```diff
@@ -67,7 +67,18 @@ export const DownloadODataServiceMetadataInputSchema = zod.object({
...
+        .string()
+        .optional()
+        .describe(
+            'The path to the SAP service to fetch metadata for. Either servicePath or serviceName must be provided.'
+        ),
+    serviceName: zod
+        .string()
+        .optional()
+        .describe(
+            'The technical name of the OData service (e.g. ZMY_SALES_SRV). If provided instead of servicePath, a catalog lookup is performed to resolve the service path.'
```
heimwege: It might be better to use the `meta.examples` property instead of having the example in the `describe` (e.g. https://github.com/SAP/open-ux-tools/blob/518e0c87b6240d9cc7453e435eb1a878857f1da9/packages/fiori-mcp-server/src/tools/schemas/appgen-config-schema-props.ts#L72)

packages/fiori-mcp-server/src/tools/services/sap-system.ts
```diff
@@ -268,6 +289,62 @@ function checkMetadata(metadata: string): void {
     }
 }
 
+export type FindServiceResult =
```
heimwege, line 292: When both Promise.allSettled legs reject (V4 and V2 catalogs both unreachable), `findService` returns `{ found: false, suggestions: [] }`. The rejection reasons are only written to `logger.debug`. The caller in `downloadODataServiceMetadata` returns `"No service named 'X' found. No similar services found."` which looks like the service doesn't exist, when the system may simply be unreachable. There is no `status: 'Error'` vs `status: 'Unreachable'` distinction in the return type, so a caller or LLM cannot distinguish the two cases.
At minimum, the hint text should indicate that the catalog itself may have been unavailable, or the `FindServiceResult` should carry a third state.

packages/fiori-mcp-server/src/tools/services/sap-system.ts
```diff
@@ -268,6 +289,62 @@ function checkMetadata(metadata: string): void {
...
+            return { found: true, service: exact };
+        }
+        allServices.push(...result.value);
+    }
+
+    const suggestions = allServices
+        .filter((s) => {
+            // The ODataServiceInfo.name is a aggregate of the services's `${group.GroupId} > ${service.ServiceAlias || service.ServiceId}` - so looking for any match here is a best-effort approach.
+            const n = s.name.toLocaleLowerCase();
+            return n.startsWith(nameLower) || n.includes(nameLower) || nameLower.includes(n);
```
heimwege: Isn't that the same as 

```suggestion
            return n.includes(nameLower) || nameLower.includes(n);
```

because `n.includes` also covers `n.startsWith`?

packages/fiori-mcp-server/src/tools/services/sap-system.ts
```diff
@@ -268,6 +289,62 @@ function checkMetadata(metadata: string): void {
...
+ * @param system - The BackendSystem or Destination to connect to.
+ * @param serviceName - The technical id of the OData service to look up (e.g. `/DMO/SD_TRAVEL_MDSK`).
+ * @returns A promise resolving to a found result with the service info, or a not-found result with suggestions.
+ */
+export async function findService(
+    system: BackendSystem | Destination,
+    serviceName: string
+): Promise<FindServiceResult> {
+    const provider = isAppStudio()
+        ? createAbapServiceProviderForDestination(system as Destination)
```
heimwege, line 314: AGENTS.md: "_Prefer type guards over as SomeType casts. If a cast is unavoidable after a runtime check, add a comment explaining why._" The same pattern appears in `getServiceMetadata` (pre-existing), but the new `findService` adds two more unchecked cast sites. Since the type is a union and `isAppStudio()` is not a TypeScript type guard, an inline comment explaining the invariant (`// safe: callers ensure destination is an ABAP OData destination` and `// safe: isAppStudio() guarantees system matches the runtime environment.`) would satisfy the rule and help future maintainers.

packages/fiori-mcp-server/src/tools/download-odata-service-metadata.ts
```diff
@@ -22,13 +22,14 @@ const EMPTY_PARAMS: DownloadODataServiceMetadataOutput['parameters'] = {
 export async function downloadODataServiceMetadata(
```
heimwege, line 22: Why should we accept the SONAR issue for the cognitive complexity threshold being violated?

heimwege (commented):
see comments + changeset is missing

packages/fiori-mcp-server/src/tools/download-odata-service-metadata.ts
IainSAP: Because, IMO, the number alone is not an accurate representation of understandability/complexity as interpreted by a human. We often accept this (>15) if for a human it would be more complex to follow the code by artificially splitting functions just to satisfy a number of comparisons metric - this is a very basic way of interpreting complexity and has little to do with readability or understandability. If you feel its not readable or understandable then I can revisit?

packages/fiori-mcp-server/src/types/input.ts
IainSAP: I will remove this as its not really a good example - it can be any string

packages/fiori-mcp-server/src/tools/services/sap-system.ts
IainSAP: Yes thats what I thought...not sure why Claude code keeps doing this (the pattern is in other places in Fiori MCP also)

packages/fiori-mcp-server/src/tools/services/sap-system.ts
IainSAP: OK but we will need to get into complex root cause analysis of connectivity issues that apply on different platforms (BAS vs VSCode) in that case, as we do for `odata-service-inquirer`. I doubt we need to do that here ? The idea here is to find something that works for the current context. I will just return the error directly with another property - RCA of connectivity issues would be a new feature I feel.  We cannot tell if a catalog failure on v2 was the cause of service on v4 not being found, for example - we can report the catalog fails - I think thats the best option for now.

packages/fiori-mcp-server/src/tools/services/sap-system.ts
IainSAP: done

IainSAP:
Closing replaced with: https://github.com/SAP/open-ux-tools/pull/4907

PR #4905, opened 2026-06-29
Reverts #4858 — ADP preview `/i18n/...properties` pass-through

nikmace:
#4858 stopped the ADP preview from redirecting `/i18n/...properties` requests to local workspace files, so they fall through to the base app instead. Structurally that is the right behaviour, and it correctly fixed the `{@i18n>...}` annotation lookups it targeted.

  However, in adaptation projects built on top of Fiori Elements V2 ListReport/ObjectPage base apps, FEV2's `TemplateComponent` rebuilds the `i18n` model on every template view and replaces it with one seeded from the base app's bundle. Once requests pass through, the base app's i18n logic effectively shadows the ADP's local `i18n.properties` inside every template, so customer-defined keys are no longer reachable from inside the editor's bindings — even though the local file is served correctly.

The proper fix requires moving customer keys to the FEV2-safe `@i18n` model (backend merger + generator changes) and cannot be safely downported in this release window. Reverting restores the pre-#4858 behaviour, which has a smaller and better-understood failure mode for the affected customers.

A corrected fix will be reapplied once the backend and generator support is in place.

  ### What changes

  - Reverts `662e3c358e` — restores `AdpPreview.proxy()` and the related tests to the pre-#4858 state.
  - Adds a patch changeset for `@sap-ux/adp-tooling`.

  ### Verification

  - `pnpm --filter @sap-ux/adp-tooling test` — 793/793 pass.
  - Manual: FEV2-based adaptation projects again resolve their local `i18n.properties` keys inside ListReport/ObjectPage views.

### DIFF
--- a/.changeset/revert-adp-tooling-i18n-passthrough.md
+++ b/.changeset/revert-adp-tooling-i18n-passthrough.md
@@ -1,0 +1,9 @@
+---
+"@sap-ux/adp-tooling": patch
+---
+
+FIX: Revert the `/i18n/...properties` pass-through introduced in #4858.
+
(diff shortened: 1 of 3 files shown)


mmilko01 (approved):
Changeset OK
Changes look good
Tested manually

IvoSG (approved):
Changeset OK
Changes look good
Do not test manually

PR #4908, opened 2026-06-30
fix(eslint-plugin): Streamline plugin meta information

marcovieth:
Streamline meta information
- Remove stale hardcoded version: '0.0.1'
- Make sure, namespace is included
[- Keep named export for "meta" because it is used to check the version in Page Map]


AlinaGovoruhina (approved):
Missing namespace added to meta, plugin version info updated. Unit tests and changeset added.

igorsVoronins (approved):
Tested the PR using the Fiori V4 app - looks good to be merged✅

- Verified that the plugin rules are displayed correctly, along with the plugin version;
- Confirmed that no regressions were introduced.

PR #4914, opened 2026-07-01
chore: replace unmaintained prettify-xml with xml-formatter

mikicvi-SAP:
## Why

prettify-xml (v1.2.0) has not been published since June 2022 and ships no TypeScript types. xml-formatter (v3.7.0) is actively maintained and ships its own types.

#### @sap-ux/odata-service-writer - `annotations.ts` and `external-services.ts`

- Uses NodeNext module resolution (inherited from the root tsconfig). xml-formatter authors its types assuming "moduleResolution": "node" - under NodeNext the default export resolves to the merged namespace type, which TypeScript does not consider callable. The import therefore needs a one-time cast:
```js
  import _fmt from 'xml-formatter';
  const xmlFormatter = _fmt as unknown as (xml: string, options?: object) => string;
```

#### @sap-ux/generator-odata-downloader - `odata-download-generator.ts`

  Overrides to "module": "commonjs", "moduleResolution": "node" in its own tsconfig (it's a Yeoman generator bundled with esbuild). This matches the resolution context xml-formatter was authored for, so the default import works without a cast:

``` js
import xmlFormatter from 'xml-formatter';
```
---
## Important - reviewers attention

  Behaviour differences from prettify-xml

  xml-formatter is a real XML parser/serialiser; prettify-xml was regex-based. This produces unavoidable differences in generated metadata.xml files:

  - throwOnFailure: false is set so invalid/partial XML fragments (used in tests) are returned unchanged rather than throwing
  - Empty elements are collapsed to one line: <Tag></Tag> instead of <Tag>\n</Tag>
  - Self-closing tags no longer get a trailing space: <Tag/> instead of <Tag />
  - Attribute line-wrapping is not preserved: prettify-xml kept existing mid-tag line breaks in the source XML; xml-formatter always serialises each element to a single line. No option exists to
  change this.
  - Stray blank lines are not reproduced: prettify-xml occasionally inserted an extra blank line before certain tags (e.g. <edmx:DataServices>); xml-formatter does not.

  Content is identical; only whitespace changes. Users who regenerate an existing app will see a diff in their metadata.xml files on first run. Snapshots and expected-output fixture files in fiori-freestyle-writer, fiori-elements-writer, and fiori-app-sub-generator have been updated accordingly.


IainSAP:
Can we keep the formnatting the same, consumers would not expect diffs in service (metadata.xml) files etc ?

devinea:
> Can we keep the formnatting the same, consumers would not expect diffs in service (metadata.xml) files etc ?

@IainSAP While I agree, it seems like that may not be easy to achieve. For app generation it will not impact customers at all at they will have no diff. For odata download and service manager there can be a small diff on space(s) along with the actual changes they are syncing. So I think it is acceptable.
We need to replace old no longer maintained library.
BTW: MCP fetch metadata is already using the new library for xml formatting.

longieirl (approved):
Changesets are good.
Comments have been addressed.
Overall change is good.

cianmSAP (approved):
Updates look ok based on update
Integration tests in inner source repo will need updating also 
Changeset looks good

devinea:
Thanks. Approved.

devinea (approved):
Thanks @mikicvi-SAP . lgtm

kjose90 (approved):
updates are good ✅

PR #4915, opened 2026-07-02
Feat(eslint-plugin-fiori-tools): enable no-live-mode rule for OData V2 applications

AlinaGovoruhina:
## Description

This PR extends the `sap-no-live-mode` ESLint rule to detect and report `liveMode` usage in **OData V2 applications** (via flex `.change` files), in addition to the existing OData V4 `manifest.json` support.

Key changes:
- **`sap-no-live-mode` rule**: Now handles both `FioriJSONSourceCode` (V4 manifest) and `FioriChangeSourceCode` (V2 flex change files). For V2, it reports a warning when a property change file sets `liveMode` to `true` on a list report page, and offers an auto-fix to set `newValue` to `false`.
- **`fe-v2.ts` linker**: 
  - Renamed `getTablePropertyChangeConfig` → `getPropertyChangeConfig` and generalized it to support page-level properties (not just table properties).
  - Refactored `getPageTableChanges` → `getPageChanges` (now a named function) to support filtering page-level changes without requiring a section.
  - `createPageConfiguration` now accepts flex changes and extracts `liveMode` configuration from them.
  - `liveMode` added to `PageSetting` interface and `FlexChangeProperty` type.
  - Object page and list report page linking now pass page-level flex changes.
- **`fe-v4.ts` linker**: Moved `liveMode` into a proper `configuration` object on `FeV4ListReport`, introducing a `PageConfiguration` interface. Removed the incorrectly placed top-level `liveMode` property.
- **`diagnostics.ts`**: Updated `NoLiveMode` interface to include `property`, optional `manifest` (V4), and optional `changeFileUri` (V2).
- **`helpers.ts`**: Added `isV2Page()` utility function (complementing the existing `isV2Table()`). Also moved `isV2Table` out of `sap-copy-to-clipboard.ts` into `helpers.ts`.
- **Tests & Docs**: Updated test suite with V2 flex change scenarios (valid and invalid), and updated documentation with examples for incorrect/correct V2 property change files.

<details>

**Version:** `1.26.13`

- Correlation ID: ``
</details>


packages/eslint-plugin-fiori-tools/src/project-context/linker/fe-v2.ts
```diff
@@ -489,11 +509,13 @@ function linkObjectPagePage(
 
     const sections = collectSections('v2', entityType, mainService);
     const createMode = target.component?.settings?.createMode;
+    const componentName = 'sap.suite.ui.generic.template.ObjectPage';
+    const pageChanges = getPageChanges(context.app.changes, { type: 'object-page', componentName, entitySetName });
```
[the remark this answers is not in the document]
AlinaGovoruhina: Replaced with empty array, since currently no flex properties for object page exist, only table properties that are checked when getting sections.

packages/eslint-plugin-fiori-tools/test/rules/sap-no-live-mode.test.ts
```diff
@@ -18,6 +18,28 @@ const ruleTester = new RuleTester({
...
+    content: { ...v2FlexChangeLiveModeEnabledLR.content, property: 'liveMode', newValue: false }
+};
+
+const v2FlexChangeLiveModeEnabledOP = {
+    ...v2FlexChangeLiveModeEnabledLR,
+    selector: {
+        ...v2FlexChangeLiveModeEnabledLR.selector,
+        id: 'v2xmlstart::sap.suite.ui.generic.template.ObjectPage.view.Details::Z_SEPMRA_SO_SALESORDERANALYSIS--Products::Table'
+    }
+};
```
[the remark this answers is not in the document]
AlinaGovoruhina: Fixed.

lfindlaysap (dismissed):

packages/eslint-plugin-fiori-tools/src/utils/helpers.ts
```diff
@@ -850,3 +850,13 @@ export function checkAppTablesConfiguration<DiagnosticType>(
...
 }
+
+/**
+ * Checks if given page is ODataV2 type.
+ *
+ * @param page ODataV2 or ODataV4 page
+ * @returns true for ODataV2 page
+ */
+export function isV2Page(page: FeV4PageType | FeV2PageType): page is FeV2PageType {
+    return 'condensedTableLayout' in (page as FeV2PageType).configuration;
```
marcovieth: 'condensedTableLayout' could be confusing since it is also part of V4's TableSettings.
Wouldn't 'createMode' be a better discriminant, as hyperspace suggested?

packages/eslint-plugin-fiori-tools/src/rules/sap-no-live-mode.ts
```diff
@@ -70,7 +76,23 @@ const rule: FioriRuleDefinition = createFioriRule({
                     operation: 'delete'
                 })
             });
-        }
+        },
+    createChangeVisitorHandler(context) {
+        return function report(node: MemberNode): void {
+            const deepestPathResult = { validatedPath: ['content', 'newValue'], missingSegments: [] };
```
marcovieth: It seems that deepestPathResult is not needed here. Why not making it optional?

marcovieth (commented):
Code looks good to me. Just two questions below.

packages/eslint-plugin-fiori-tools/src/utils/helpers.ts
AlinaGovoruhina: Ok, to avoid confusion switched to `createMode` to check the page is ODataV2, as for ODataV4 `createMode` can be configured on the app level only.

packages/eslint-plugin-fiori-tools/src/rules/sap-no-live-mode.ts
AlinaGovoruhina: I updated to use a const as the `deepestPathResult` is the same for flex changes currently. The json path is still used in `createJsonFixer`, do you want to make it optional there?

igorsVoronins (dismissed):
Tested the rule with V2 Fiori app:
Correctly fires warning in the flex change file when user enable livemode to true;
Auto-fix is setting the value to false(instead of deleting in case of V4);
Checked that rule only applies to List Report page and Analytical page. Object pages being ignored;
PR is looking good to be merged ✅

packages/eslint-plugin-fiori-tools/src/rules/sap-no-live-mode.ts
marcovieth: I thought we could make it optional and only set when needed ("insert" operation, or when letting it auto-detect). But i think, using the constant is also ok.

packages/eslint-plugin-fiori-tools/src/utils/helpers.ts
```diff
@@ -850,3 +855,15 @@ export function checkAppTablesConfiguration<DiagnosticType>(
...
+/**
+ * Checks if given page is ODataV2 type.
+ * Create mode is a page configuration setting in ODataV2.
+ * But create mode can be set only on an application in ODataV4.
+ *
+ * @param page ODataV2 or ODataV4 page
+ * @returns true for ODataV2 page
+ */
+export function isV2Page(page: FeV4PageType | FeV2PageType): page is FeV2PageType {
+    return 'createMode' in (page as FeV2PageType).configuration;
```
marcovieth, line 868: Type cast not necessary (also in isV2Table above).
(But I would only fix it if there is something else to fix.)

marcovieth (dismissed):
Looks good to me.
Just a tiny comment which I would rate as optional to fix.

packages/eslint-plugin-fiori-tools/docs/rules/sap-no-live-mode.md
```diff
@@ -31,10 +31,24 @@ Enabling `liveMode` makes the content area load automatically during the initial
 ```
 `liveMode` is set to `true` in the settings of the list report page.
 
-### Correct
+#### Incorrect Property Change File for ODataV2 Applications
```
lfindlaysap:
```suggestion
#### Incorrect `.change` File for OData V2 Applications
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-no-live-mode.md
```diff
@@ -31,10 +31,24 @@ Enabling `liveMode` makes the content area load automatically during the initial
...
+
+```json
+"content": {
+    "property": "liveMode",
+    "newValue": true
+}
+```
+`liveMode` property `newValue` is set to `true`.
```
lfindlaysap:
```suggestion
The `newValue` property for `liveMode` is set to `true`.
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-no-live-mode.md
```diff
@@ -31,10 +31,24 @@ Enabling `liveMode` makes the content area load automatically during the initial
...
+
+### Correct `manifest.json` File for ODataV4 Applications
```
lfindlaysap:
```suggestion
### Correct `manifest.json` File for OData V4 Applications
```

packages/eslint-plugin-fiori-tools/docs/rules/sap-no-live-mode.md
```diff
@@ -31,10 +31,24 @@ Enabling `liveMode` makes the content area load automatically during the initial
...
 
 Set the `liveMode` property to `false` or remove it from the `manifest.json` file to restore the default settings.
 
+### Correct `.change` File for ODataV2 Applications
```
lfindlaysap:
```suggestion
### Correct `.change` File for OData V2 Applications
```

lfindlaysap (changes requested):

packages/eslint-plugin-fiori-tools/docs/rules/sap-no-live-mode.md
```diff
@@ -31,10 +31,24 @@ Enabling `liveMode` makes the content area load automatically during the initial
 ```
 `liveMode` is set to `true` in the settings of the list report page.
 
-### Correct
+#### Incorrect Property `.change` File for OData V2 Applications
```
lfindlaysap: @AlinaGovoruhina, why do we use Property `.change` File here but `.change` File in line 48? These should be the same.

If it's important to include property: "Incorrect `.change` File for a Property in OData V2 Applications", and use the same formulation for 48. Otherwise it should be the same as 48: "Incorrect `.change` File for OData V2 Applications"

lfindlaysap (changes requested):

packages/eslint-plugin-fiori-tools/docs/rules/sap-no-live-mode.md
AlinaGovoruhina: There wasn't any particular reason, so I removed "Property" from line 34. Now it should match.

lfindlaysap (approved):

marcovieth (approved):
Looks good to me.

PR #4918, opened 2026-07-03
fix(ui5-test-writer): Object pages navigation now expect any number of OPs

sap-sebelao:
- handle variable number of parent OPs rather than a single one

Internal JIRA
FXUBRQ34-12640


packages/ui5-test-writer/src/utils/objectPageUtils.ts
```diff
@@ -103,33 +104,46 @@ export function getObjectPages(applicationModel: ApplicationModel): PageWithMode
...
-            navigationParents.parentOPName = objectPage.name;
-
-            navigationParents.parentOPTableSection = routeToTargetOP.identifier;
+    const parentOPs: ObjectPageNavigationParent[] = [];
+    const visited = new Set<string>(); // guard against infinite loop in case of invalid manifest entries
+    let cursor = targetObjectPageKey;
+
+    while (!visited.has(cursor)) {
+        visited.add(cursor);
+        const childKey = cursor;
```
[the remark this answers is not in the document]
sap-sebelao: This suggestion breaks eslint. Avoiding both the lint rule and the extra variable would require introducing another loop.

packages/ui5-test-writer/src/utils/objectPageUtils.ts
```diff
@@ -103,33 +104,46 @@ export function getObjectPages(applicationModel: ApplicationModel): PageWithMode
...
-        if (routeToTargetOP) {
-            navigationParents.parentOPName = objectPage.name;
-
-            navigationParents.parentOPTableSection = routeToTargetOP.identifier;
+    const parentOPs: ObjectPageNavigationParent[] = [];
+    const visited = new Set<string>(); // guard against infinite loop in case of invalid manifest entries
+    let cursor = targetObjectPageKey;
+
+    while (!visited.has(cursor)) {
+        visited.add(cursor);
```
heimwege: the visited.add(cursor) at the top of the loop records the child key, but parentOPs.unshift(parent) happens before the new cursor value is checked. With a mutual-navigation cycle A↔B targeting A: after two iterations parentOPs = [{A}, {B}] - A is listed as its own ancestor.
Fix: check `visited.has(parent.name)` before the `unshift`, or seed `visited = new Set([targetObjectPageKey])` before the loop.

heimwege (commented):
see comment + `packages/ui5-test-writer/test/test-input/fin.test.v4.lr1` is committed as a git submodule (mode 160000) but there's no `.gitmodules` registration anywhere in the repo. It's also never referenced in any test. Remove it before merge - it will break `git clone --recurse-submodules` for everyone.

packages/ui5-test-writer/src/utils/objectPageUtils.ts
sap-sebelao: Fixed.

sap-sebelao:
> see comment + `packages/ui5-test-writer/test/test-input/fin.test.v4.lr1` is committed as a git submodule (mode 160000) but there's no `.gitmodules` registration anywhere in the repo. It's also never referenced in any test. Remove it before merge - it will break `git clone --recurse-submodules` for everyone.

Removed the test sample that was commited as submodule by accident.

Fixed my local git settings to (hopefully) avoid these in the future.

heimwege (dismissed):
- changeset ok
- review comments addressed
- coverage is perfect
- did NOT test manually

devinea (dismissed):
Code changes look good. 
Changeset ✅ 

merge conflicts

devinea (approved):
Re-approving after merge conflicts fixed.

heimwege (approved):
Re-approve

PR #4920, opened 2026-07-03
fix: conditional usage of inbound change types

mmilko01:
## Description

This fix introduces version-based conditional selection of inbound change types when generating FLP (Fiori Launchpad) configuration for ADP (Adaptation Project) projects.

**Problem:** The `appdescr_app_setInbounds` change type introduced in UI5 1.143 was being used unconditionally for all systems, including ABAP systems running UI5 1.142 or lower where only the legacy `appdescr_app_addNewInbound` + `appdescr_app_removeAllInboundsExceptOne` change types are supported.

**Solution:**
- ABAP systems running UI5 **1.142 or lower** → use legacy change types (`appdescr_app_addNewInbound` + `appdescr_app_removeAllInboundsExceptOne`)
- ABAP systems running UI5 **1.143 or higher** → use the new `appdescr_app_setInbounds` change type
- **CF projects always** use `appdescr_app_setInbounds` regardless of UI5 version

### Key Changes

- **`adp-tooling/writer/options.ts`**: Added `shouldUseLegacyInboundChangeTypes()` utility function that checks the system UI5 version against the `1.143.0` threshold. Extracted `appendLegacyInboundChanges()` and `appendSetInboundsChange()` helper functions. Updated `enhanceManifestChangeContentWithFlpConfig()` to accept a `useLegacyChangeTypes` flag.

- **`adp-tooling/writer/inbound-navigation.ts`**: Updated `generateInboundConfig()` to accept optional `systemUI5Version` and `isCfProject` parameters and delegate change type selection to `shouldUseLegacyInboundChangeTypes()`.

- **`adp-flp-config-sub-generator`**: Fetches the system UI5 version via `getSystemUI5Version()` during initialization and passes it along to `generateInboundConfig()`.

- **`create/cli/add/navigation-config.ts`**: Updated `getInbounds()` to return `{ data, systemUI5Version, isCfProject }` instead of just inbound data, and propagates version/CF context through to `generateInboundConfig()`.

- **Tests**: Updated across all affected packages to verify correct change type selection based on version and project type.

## Type of change

## How have you tested?
Unit tests added/updated across all affected packages covering:
- `shouldUseLegacyInboundChangeTypes()` for all version/CF combinations
- `generateInboundConfig()` with legacy vs. new change types
- Sub-generator and CLI command tests verifying correct parameters are passed to `generateInboundConfig()`

## Checklist:

<details>

**Version:** `1.26.14`

- Correlation ID: `dfce7370-76ea-11f1-9b04-490e900dea10`
</details>


packages/adp-tooling/src/writer/options.ts
```diff
@@ -312,29 +315,99 @@ function getInboundChangeContentWithNewInboundID(
...
+            inbound: {
+                [flpConfig.inboundId]: inboundChangeContent.inbounds[flpConfig.inboundId]
+            }
+        },
+        texts: {
+            'i18n': 'i18n/i18n.properties'
+        }
+    });
+    // Only add 'removeAllInboundsExceptOne' after the first inbound is added
+    if (isFirst) {
```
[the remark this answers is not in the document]
nikmace: This is intentional and preserves the pre-4633 behavior verbatim. appdescr_app_removeAllInboundsExceptOne is a base-app inbound cleanup, not a per-inbound operation:
- Iteration 0: emit addNewInbound(#_1) then removeAllInboundsExceptOne(#_1) — this wipes inbounds inherited from the base app, keeping only #_1.
- Iteration 1+: emit addNewInbound(#_N) — these stack on top of the surviving #_1 on the ABAP side.

Emitting removeAllInboundsExceptOne for every inbound would race with the ABAP merge and defeat the pattern. This is the same shape the code had before PR #4633 replaced it with setInbounds (see git show
8c4185adbd^:packages/adp-tooling/src/writer/options.ts).

Added a multi-inbound legacy test in [fcca09e](commit link) asserting the exact change order: [addNewInbound(#_1), removeAllInboundsExceptOne(#_1), addNewInbound(#_2)]. Expanded the inline comment so the intent is easier to catch on future review.

packages/adp-flp-config-sub-generator/src/app/index.ts
```diff
@@ -251,6 +259,9 @@ export default class AdpFlpConfigGenerator extends Generator {
             }
         ]);
         await this.prompt(prompts);
+        if (this.provider) {
+            this.systemUI5Version = await getSystemUI5Version(this.provider, this.toolsLogger);
```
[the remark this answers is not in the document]
nikmace: Good catch on the surface, but tracing the actual flow: `await this.prompt(prompts)` doesn't resolve until the credentials validator returns true (inquirer keeps re-prompting on non-true returns), which only happens on the auth-success path where `this.provider` is set. So if (this.provider) was effectively dead — the guard was defensive-only, and re-auth can't land here with provider unset (the flow either succeeds or the user cancels and `writing()` never runs).

That said, the call was unwrapped, so a transient failure of `getSystemUI5Version` after successful auth would crash the generator — which contradicts the pattern used in `_initializeAbapGenerator` (lines 542-546)
where the same call is wrapped in try/catch and errors are swallowed at debug.

Fixed in [fcca09e](commit link) by wrapping this call to match. The if (this.provider) guard is dropped since it was dead code.
> Note: the same unwrapped pattern exists in packages/create/src/cli/add/navigation-config.ts:161 (inside Promise.all). Not touched in this PR since it's outside the scope of the review comment, but worth a
follow-up if consistency is desired.

avasilev-sap (dismissed):
Looks good

packages/adp-tooling/src/writer/options.ts
```diff
@@ -312,29 +320,103 @@ function getInboundChangeContentWithNewInboundID(
...
+ * Determines whether the legacy inbound change types should be used based on the system UI5 version.
+ * Legacy change types ('appdescr_app_addNewInbound' + 'appdescr_app_removeAllInboundsExceptOne') are
+ * used for ABAP systems on UI5 1.142 or lower. CF projects always use the new type regardless of version.
+ *
+ * @param systemUI5Version - The UI5 version string reported by the system (e.g. '1.142.0').
+ * @param isCfProject - Whether this is a Cloud Foundry adaptation project.
+ * @returns True when the legacy change types must be used.
+ */
+export function shouldUseLegacyInboundChangeTypes(systemUI5Version: string | undefined, isCfProject: boolean): boolean {
+    if (isCfProject || !systemUI5Version) {
```
heimwege: Please double-check: `houldUseLegacyInboundChangeTypes` returns `false` when `systemUI5Version` is `undefined`, meaning a failed version fetch silently emits `appdescr_app_setInbounds` on an old ABAP system. The safe default for an ABAP system that can't report its version should be `true` (legacy), since systems too old to expose a version service are almost certainly also too old for `setInbounds`.

packages/adp-flp-config-sub-generator/src/app/index.ts
```diff
@@ -251,6 +259,11 @@ export default class AdpFlpConfigGenerator extends Generator {
             }
         ]);
         await this.prompt(prompts);
+        try {
+            this.systemUI5Version = await getSystemUI5Version(this.provider, this.toolsLogger);
```
heimwege, line 271: nit-pick: `getSystemUI5Version` including try-catch is placed in both `_initializeAbapGenerator()` (normal flow) and `_promptAuthentication()` (auth-required flow).
Might be worth extracting to a single `private async _fetchSystemUI5Version()` helper to have the same try catch behavior, log message, etc. that will now differ over time because one has been adjusted and the other one aws forgotten.

heimwege (commented):
see comments

packages/adp-tooling/src/writer/options.ts
nikmace: Makes total sense: [e062f05](https://github.com/SAP/open-ux-tools/pull/4920/commits/e062f05ec994ca4d4e01f5be0b105e33a8c9fdaa)

packages/adp-flp-config-sub-generator/src/app/index.ts
nikmace: Changed in: [e822f80](https://github.com/SAP/open-ux-tools/pull/4920/commits/e822f80f11b0319607fc912fe1c2ff602c0543eb)

heimwege (approved):
- changeset ok
- review comments addressed
- coverage ok
- did NOT test manually

nikmace:
Screenshot 2026-07-07 at 8 04 37

nikmace (approved):
Changes look good
Changeset is OK
Coverage is good
Did not test manually

avasilev-sap (approved):

PR #4922, opened 2026-07-06
fix(adp-tooling): Pass i18n.properties requests through to base app proxy

### LINKED ISSUE #4857: BUG - ADP preview redirects base app `i18n.properties` to local file, breaking `{@i18n>...}` annotation bindings
## Affected package

`@sap-ux/adp-tooling` (also surfaces via `@sap/ux-ui5-tooling` which bundles it).

### Description

When an adaptation project (ADP) is run in local dev preview (`npm start` → `fiori-tools-preview` middleware), requests for the base app's i18n bundles are 302-redirected to the ADP's own `webapp/i18n/` folder. The ADP's bundle contains only customer-prefixed keys (per the official ADP convention), so the runtime `i18n` and `@i18n` `ResourceModel`s end up loading a bundle that is missing every base-app key.

Annotation-side bindings such as `String="{@i18n>col.image}"` then fail to resolve. The UI renders raw key strings or empty cells.

The bug is **dev-only**. Once deployed, the runtime fetches the base bundle directly from the BSP (no preview middleware in the path), and the `appdescr_ui5_addNewModelEnhanceWith` change correctly merges the ADP's prefixed keys on top.

### Steps to reproduce

1. Create an FE V2 LROP base app with annotation file containing label bindings of the form `{@i18n>some.key}` and matching keys defined in `webapp/i18n/i18n.properties`.
2. Declare an `@i18n` `ResourceModel` in `sap.ui5.models` of the base manifest (required because annotation-side bindings target this model).
3. Deploy the base app to an ABAP system.
4. Generate an ADP project (`@sap-ux/generator-adp`) targeting the deployed base.
5. Run `npm start` in the ADP, open the FLP preview.
6. Inspect column headers / OP type names: they show raw keys such as `col.image`, `op.typeNamePlural`. Console logs `Assertion failed: could not find any translatable text for key 'col.image' in bundle file(s): '/sap/bc/.../i18n/i18n.properties'`.

### Expected behavior

The base app's i18n bundle is served from the BSP (or whatever resource server backs the base app), so all base keys resolve. The ADP's `appdescr_ui5_addNewModelEnhanceWith` change adds the customer-prefixed keys to the same model at runtime.

### Actual behavior

`fiori-tools-preview` 302-redirects `/sap/bc/<base-app>/i18n/i18n.properties` to `/i18n/i18n.properties` (the ADP's local file). The base app's bundle is never loaded; `{@i18n>...}` lookups fail; the SmartTable / Object Page render empty or raw-key labels.

### Screenshots

OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

## Root cause

`packages/adp-tooling/src/preview/adp-preview.ts`, method `proxy()`. The fall-through branch:

```ts
const files = await this.project.byGlob(req.path.replace('.js', '.*'));
if (files.length === 1) {
    res.redirect(302, req.path);
} else {
    next();
}
```

This redirects unconditionally for any path whose file exists in the ADP's webapp folder. That is correct for ADP overrides of fragments, controllers, and change files (full file substitution is the right semantic there) but wrong for `i18n.properties`, which is a partial bundle merged into the base at the model layer, not the file layer.

## Proposed fix

Add an early branch in `proxy()` that lets `i18n` `.properties` requests fall through to the next middleware without substitution:

```ts
} else if (req.path.startsWith('/i18n/') && req.path.endsWith('.properties')) {
    // i18n .properties files must always fall through to the base app's resource
    // server. ADP customer-prefixed keys are merged via appdescr_ui5_addNewModelEnhanceWith
    // at runtime, NOT by substituting the entire bundle.
    next();
}
```

The match is intentionally broad to cover:

- Default bundle: `/i18n/i18n.properties`
- Locale variants: `/i18n/i18n_de.properties`, `/i18n/i18n_en_US.properties`
- Per-page bundles: `/i18n/ListReport/<entity>/i18n.properties`
- Per-page locale variants: `/i18n/ListReport/<entity>/i18n_de.properties`

nikmace:
### Context

Re-submission of #4858, which was merged on Jun 24 and reverted on Jun 29 (#4905). The code itself is correct — the revert happened because the fix was incomplete on its own and regressed integration tests. This PR ships the same 3-file, 53-line diff; the integration tests will be adjusted before merge so they no longer depend on the redirect-as-shadowing behaviour this fix removes.

Fixes #4857.

### Problem

`AdpPreview.proxy()` was intercepting every `/i18n/...properties` request and, when a matching file existed in the ADP webapp, 302-redirecting to it. This substituted the ADP's **partial** bundle (customer keys only) in place of the base app's **complete** bundle, so `{@i18n>...}` annotation lookups (e.g. `UI.LineItem` column headers, ObjectPage type names) that resolve against the base bundle broke silently.

ADP customer keys are meant to reach the runtime via `appdescr_ui5_addNewModelEnhanceWith` — layered onto the model at manifest-merge time — not by substituting the entire bundle at the proxy.

### Fix

Add an early pass-through branch in `AdpPreview.proxy()` for any request matching `/i18n/...properties` (default bundle, locale variants, and per-page bundles like `/i18n/ListReport/<entity>/i18n.properties`). Requests fall through to the base app's resource server; the ADP's `enhanceWith` change still merges customer keys into the model at runtime.

### Files changed

- `packages/adp-tooling/src/preview/adp-preview.ts` — 10-line pass-through branch in `proxy()`
- `packages/adp-tooling/test/unit/preview/adp-preview.test.ts` — unit tests for pass-through behaviour
- `.changeset/adp-tooling-i18n-passthrough.md` — patch changeset

### Known scope

This fix on its own does not repair `{i18n>...}` inside FEV2 template views — FEV2's `fnEnhanceI18nModel()` rebuilds the `i18n` model on every template component, so customer keys layered via `enhanceWith` for `modelId: "i18n"` don't reach template views. The supported binding inside FEV2 templates is `{@i18n>...}`; making that work end-to-end for base apps that don't declare `@i18n` requires the backend `createIfMissing` merger feature + a generator update, tracked separately.

### Test plan

- [x] `pnpm --filter @sap-ux/adp-tooling test` passes
- [ ] Integration tests in `ui5-tooling` adjusted to page-specific i18n before merge
- [ ] Manual verification: `UI.LineItem` column-header labels resolve against base-app keys in an ADP preview
- [ ] Manual verification: ADP customer keys still reach the outer app component via `enhanceWith`


nikmace:
Screenshot 2026-07-06 at 11 42 06

avasilev-sap (approved):

IvoSG (approved):
Re-approval

PR #4929, opened 2026-07-08
fix(eslint-plugin-fiori-tools): check property change file has required properties defined

AlinaGovoruhina:
## Description

Improves the robustness of `.change` file parsing in the `eslint-plugin-fiori-tools` package by adding validation guards to prevent malformed or incomplete property change files from causing issues.

**Key changes:**

- **OData v2 version check** (`parser.ts`): Property change files are now only collected when the app's main service is OData v2 (`version === '2.0'`), avoiding unnecessary processing for non-v2 services.

- **Required field validation** (`utils.ts`): Added `isFlexChange()` guard that verifies `changeType`, `content`, and `selector` are all present before constructing a `FlexChange` object. Files missing any of these fields are silently skipped.

- **Improved `reparseChange` logic** (`parser.ts`): Refactored to use the new `getAppForPath()` utility for accurate app lookup. When a file is updated to an incomplete/invalid state, the existing index entry is now properly removed.

- **New `getAppForPath()` utility** (`utils.ts`): Resolves which app a given file path belongs to — returns the single app directly if only one exists, or matches by path for multi-app projects.

- **Test improvements** (`parser.test.ts`, `utils.test.ts`): Switched `beforeAll` to `beforeEach` for proper test isolation, added `fileCache.clear()` in setup, fixed assertions to use `reparsed.index`, and added new tests for applying changes, skipping empty `.change` files, removing stale entries, and `getAppForPath` behavior.

## Type of change

## How have you tested?
- Property change files missing `changeType`, `content`, or `selector` are skipped without errors.
- Property change files are only collected when the main service is OData v2.
- New unit tests cover empty file handling, stale entry removal, change updates, and `getAppForPath` lookup logic.
- All existing unit tests continue to pass.

## Checklist:

<details>

**Version:** `1.27.2`

- Correlation ID: `9564a690-7c2d-11f1-879a-4dfddeef1d2c`
</details>


packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
```diff
@@ -211,6 +213,7 @@ export class ApplicationParser {
                 const path = fileURLToPath(change.changeFileUri);
                 return existsSync(path);
             });
+            const existingChangeIndex = app.changes.findIndex((change) => change.changeFileUri === uri);
```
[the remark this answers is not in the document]
AlinaGovoruhina: Does not seem necessary to me to just skip existsSync.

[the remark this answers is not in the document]
AlinaGovoruhina: Added a check to get the app file belongs to, as this has come up before. Though multiple apps are only in a CAP project that doesn't work with .change files anyway.

packages/eslint-plugin-fiori-tools/test/project-context/parser/parser.test.ts
```diff
@@ -44,21 +45,56 @@ describe('Flex change', () => {
...
+        fileCache.set(nonExistentChangeFileUri, JSON.stringify(existingPropertyChange));
+        parsedProject.apps[''].changes = [existingPropertyChange];
+        const reparsed = parser.reparse(changeFileUri, parsedProject, fileCache); // reparse new change file
+        expect(reparsed.index.documents[changeFileUri]).toBeDefined();
+        expect(reparsed.index.apps[''].changes).toHaveLength(1); // non-existent-file change was deleted
+        expect(reparsed.index.apps[''].changes[0].changeFileUri).toStrictEqual(changeFileUri);
+    });
+
+    test('reparse: applies change in a .change file', () => {
+        const newChange = Object.assign(propertyChange) as FlexChange;
```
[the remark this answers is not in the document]
AlinaGovoruhina: Updated

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
```diff
@@ -204,38 +206,41 @@ export class ApplicationParser {
...
+        // Create and save the ast tree
+        const ast = parseJson(content, {
+            mode: 'json',
+            ranges: true,
+            tokens: true,
+            allowTrailingCommas: false
+        });
+        index.documents[uri] = ast;
+        // Create new change object
+        const jsonContent = JSON.parse(content);
```
marcovieth: I think we need a try..catch here (as we have in collectFlexChanges).
Edit a .change file to get malformed JSON  breaks eslint.

marcovieth (commented):
Just a partial review for now. Please check my comment.

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
AlinaGovoruhina: Added a try-catch block https://github.com/SAP/open-ux-tools/pull/4929/commits/2d84e5105546bd368a275b24b89bffb80f29ec56
When catching an error, it will remove previous flexChange information from the app changes.

packages/eslint-plugin-fiori-tools/src/project-context/utils.ts
```diff
@@ -15,6 +16,46 @@ export function getParsedServiceByName(parsedApp: ParsedApp, serviceName?: strin
...
+ * @param path - File path
+ * @returns Parsed app if found
+ */
+export function getAppForPath(apps: { [appName: string]: ParsedApp }, path: string): ParsedApp | undefined {
+    const appNames = Object.keys(apps);
+    if (appNames.length === 1) {
+        return apps[appNames[0]];
+    }
+    if (appNames.length > 1) {
+        const appName = appNames.find((appName) => path.includes(join(apps[appName].projectRootPath, appName)));
```
marcovieth: I think appName is a full URI and not a path segment so this won't work:
join(apps[appName].projectRootPath, appName)

Do you mean path.includes(fileURLToPath(appName)) ?

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
```diff
@@ -84,7 +84,9 @@ export class ApplicationParser {
                 const [parsedManifest, services] = this.parseManifest(webappPath, manifestUri, manifest);
                 const appRootUri = pathToFileURL(app.appRoot).toString();
                 const changes: FlexChange[] = [];
-                if (existsSync(join(webappPath, 'changes'))) {
+                const mainServiceName = getMainService(manifest) ?? '';
+                const mainService = services.find((service) => service.name === mainServiceName);
+                if (mainService?.version === '2.0' && existsSync(join(webappPath, 'changes'))) {
```
marcovieth, line 89: Can we also have change files for V4?

packages/eslint-plugin-fiori-tools/src/project-context/utils.ts
```diff
@@ -15,6 +16,46 @@ export function getParsedServiceByName(parsedApp: ParsedApp, serviceName?: strin
...
+ * Otherwise, looks for an app, which path is a substring in the file path.
+ *
+ * @param apps - Object containig parsed apps
+ * @param path - File path
+ * @returns Parsed app if found
+ */
+export function getAppForPath(apps: { [appName: string]: ParsedApp }, path: string): ParsedApp | undefined {
+    const appNames = Object.keys(apps);
+    if (appNames.length === 1) {
+        return apps[appNames[0]];
```
marcovieth: Do we also need path validation for a single app?

packages/eslint-plugin-fiori-tools/src/project-context/utils.ts
```diff
@@ -15,6 +16,46 @@ export function getParsedServiceByName(parsedApp: ParsedApp, serviceName?: strin
...
+ * @param changeObject - Object from parsed .change file
+ * @returns boolean
+ */
+export function isFlexChange(changeObject: Partial<FlexChange>): changeObject is FlexChange {
+    return (
+        'changeType' in changeObject &&
+        'content' in changeObject &&
+        typeof changeObject.content?.property === 'string' &&
+        'newValue' in changeObject.content &&
+        typeof changeObject.selector?.id === 'string'
```
marcovieth, line 56: Shall we also check for selector.type here since we need it later?
And maybe also typeof changeType === 'string'?

marcovieth (commented):
Please check my comments.

packages/eslint-plugin-fiori-tools/src/project-context/parser/parser.ts
AlinaGovoruhina: No, flex change is only relevant for ODataV2. Rules already check for V2 pages, but we can prevent checking for .change files in V4 applications here also.

packages/eslint-plugin-fiori-tools/src/project-context/utils.ts
AlinaGovoruhina: You are correct. I fixed this to use appUri instead of appId, also added a path check for single app as well.

packages/eslint-plugin-fiori-tools/src/project-context/utils.ts
AlinaGovoruhina: Added.

packages/eslint-plugin-fiori-tools/src/project-context/utils.ts
AlinaGovoruhina: Added selector type and changeType === "propertyChange" checks.

marcovieth (approved):
Looks good to me.

igorsVoronins (approved):
Tested the PR against V2 fiori app with App-Modeler extension with following scenarios:
- empty {}
- missing selector/content/changeType
No red toast in any case in App-Modeler. Valid .change entries continued to surface correctly in App Modeler. Looks good for merging ✅

PR #4933, opened 2026-07-10
feat: add skill for update/add vocabularies

gernotkuhr:
- add skill to add vocabulary and/or update existing vocabularies
- add docu how to add/update vocabularies
- add docu how to test local vocabulary changes in tools-suite


packages/odata-vocabularies/README.md
```diff
@@ -34,6 +34,41 @@ Library that contains the most recent copies of the OData vocabularies authored
...
+
+After completing, create a changeset with `pnpm cset`.
+
+> **Note:** `com.sap.cds.vocabularies.*` files (ObjectModel, AnalyticsDetails) are hand-crafted and cannot be added this way.
+
+## Local testing in tools-suite
+
+To test local changes to this package inside the XML annotation language server of `tools-suite`, use [yalc](https://github.com/wclr/yalc). `pnpm link` does not work there because the language server is an esbuild bundle and dependencies are resolved at build time.
+
+See [cross-repo-dev.md](https://github.wdf.sap.corp/ux-engineering/tools-suite/blob/master/docs/dev-guide/cross-repo-dev.md) *(SAP-internal)* in tools-suite for the full setup and iteration workflow.
```
heimwege: This URL should absolutely NOT be mentioned here! This is an internal URL that should never ever be posted to an external github.

.changeset/odata-vocabularies-readme-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-ux/odata-vocabularies": patch
+---
+
+FIX: Correct README — add missing SAP vocabularies (Auditing, CDS, Support) and fix wrong package import path
```
heimwege: Why is this a FIX? it adds a new skill, right? Shouldn't this be a `FEAT` (and `minor` bump) then?

.agents/skills/odata-vocabularies-sync/SKILL.md
```diff
@@ -0,0 +1,136 @@
+---
+name: odata-vocabularies-sync
+description: Sync odata-vocabularies package — update all vocabulary resource files to their latest published versions, and optionally add a new vocabulary given its JSON URL. Use when asked to update vocabularies, refresh vocabulary resources, or add a new vocabulary to @sap-ux/odata-vocabularies.
+argument-hint: (optional) URL of the new vocabulary in JSON format (ending in .json), e.g. https://sap.github.io/odata-vocabularies/vocabularies/SomeNew.json — XML URLs are not supported
+metadata:
+  author: sap-ux
```
heimwege: I would keep the author in sync with the customer facing skills (https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-mcp-server/skills)

```suggestion
  author: sap-fiori-tools
```

.agents/skills/odata-vocabularies-sync/SKILL.md
```diff
@@ -0,0 +1,136 @@
...
+'<namespace>': {
+    uri: '<the provided URL>'
+},
+```
+
+### 1c. Edit `src/resources/index.ts`
+
+Make all six additions. Insert each in alphabetical order within entries of the same vocabulary family. All six are mandatory — missing any one will cause a runtime error.
+
+1. **Import** (grouped with same-family imports):
```
heimwege: the actual index.ts does NOT group by family - OASIS and SAP imports are interleaved, apparently ordered by alias name (Aggregation, Auditing, Auth, Capabilities, CDS, CodeList...). An agent following the skill's instruction would insert in the wrong place. The instruction should say "in alphabetical order by import identifier (alias), in the single existing import block."

.agents/skills/odata-vocabularies-sync/SKILL.md
```diff
@@ -0,0 +1,136 @@
...
+   ```typescript
+   import <Alias> from './<namespace>.js';
+   ```
+   Note: the import identifier must be a valid JS identifier. Use the `$Alias` value from the JSON verbatim — check existing OASIS aliases for precedent (e.g. `Auth` for Authorization, `ODataJSON` for JSON).
```
heimwege: The skill says "Use the $Alias value from the JSON verbatim" but then notes "ODataJSON for JSON" as a parenthetical.
This is a real gotcha: the alias in the file (ODataJSON) differs from the JSON `$Alias` (JSON) because JSON is a reserved JS identifier. The skill should make this more prominent - something like: "If the alias is a reserved JS keyword or built-in (e.g. `JSON`, `Map`, `Error`), prefix it to make it a valid identifier (e.g. `ODataJSON`)."

.agents/skills/odata-vocabularies-sync/SKILL.md
```diff
@@ -0,0 +1,136 @@
...
+- `namespace` — the key of the top-level schema object (e.g. `Org.OData.NewThing.V1` or `com.sap.vocabularies.NewThing.v1`)
+- `alias` — the value of `$Alias` inside that schema object (e.g. `NewThing`)
+
+Determine the vocabulary family from the namespace:
+- Starts with `Org.OData.` → **OASIS vocabulary**
+- Starts with `com.sap.vocabularies.` → **SAP vocabulary**
+- Starts with `com.sap.cds.` → **CDS vocabulary — stop.** These files are hand-crafted and cannot be added via this skill. Explain this to the user.
+- Anything else → stop and ask the user to confirm the family before proceeding.
+
+### 1b. Edit `tools/update.ts`
```
heimwege, line 37: `com.sap.vocabularies.CDS.v1` has` update: false` in `update.ts`. The skill never explains this field. If a user adds a vocabulary they don't want auto-updated, they'd have no idea to add it. This could be added as a note in Step 1b.

.agents/skills/odata-vocabularies-sync/SKILL.md
```diff
@@ -0,0 +1,136 @@
...
+
+---
+
+## Step 3 — Run tests and update snapshots
+
+```bash
+pnpm --filter @sap-ux/odata-vocabularies test -- -u
+```
+
+Report any test failures that are not snapshot-related. Snapshot updates are expected and normal after vocabulary content changes.
```
heimwege: Step 3 says "Snapshot updates are expected and normal after vocabulary content changes." This is true for a sync/update run, but for a "no-arg update" where nothing upstream changed, there should be no snapshot changes. Framing them as always expected may hide real regressions. Better: "Snapshot updates are expected when vocabulary content has changed upstream; if running without a URL arg and snapshots change unexpectedly, inspect the diff before accepting."

.agents/skills/odata-vocabularies-sync/SKILL.md
```diff
@@ -0,0 +1,136 @@
...
+
+---
+
+## Step 4 — Report
+
+Summarise what was done:
+- If a new vocabulary was added: confirm the namespace, alias, family, and all files changed.
+- List vocabulary resource files regenerated by the update tool.
+- Confirm tests pass.
+- Remind the user to create a changeset with `pnpm cset` since `src/` files were modified.
```
heimwege: AGENTS.md covers changesets extensively. The reminder in Step 4 is redundant and adds noise. You can drop it entirely - AGENTS.md has it covered more thoroughly than the skill ever would.

packages/odata-vocabularies/README.md
```diff
@@ -34,6 +34,41 @@ Library that contains the most recent copies of the OData vocabularies authored
 * Support
 * UI
 
+## Maintaining Vocabularies
+
+Both operations are handled by the `odata-vocabularies-sync` skill (`.claude/skills/odata-vocabularies-sync/SKILL.md`).
+
+Use the slash command in Claude Code (CLI or VSCode extension):
```
heimwege: This skill can be used by agents in general, not only Claude Code, right?
Not sure if we need to mention how to use a skill at all 🤔 ...on the other hand the URL param needs to be mentioned somewhere.

packages/odata-vocabularies/README.md
```diff
@@ -34,6 +34,41 @@ Library that contains the most recent copies of the OData vocabularies authored
...
+The skill will:
+1. *(When adding)* Register the vocabulary in `tools/update.ts`, `src/resources/index.ts`, `src/loader.ts`, and `README.md`
+2. Download and regenerate all vocabulary resource files
+3. Run tests and update snapshots
+
+After completing, create a changeset with `pnpm cset`.
+
+> **Note:** `com.sap.cds.vocabularies.*` files (ObjectModel, AnalyticsDetails) are hand-crafted and cannot be added this way.
+
+## Local testing in tools-suite
```
heimwege, line 44: wrong place. This is an external repo

packages/odata-vocabularies/README.md
```diff
@@ -34,6 +34,41 @@ Library that contains the most recent copies of the OData vocabularies authored
...
 
+## Maintaining Vocabularies
+
+Both operations are handled by the `odata-vocabularies-sync` skill (`.claude/skills/odata-vocabularies-sync/SKILL.md`).
+
+Use the slash command in Claude Code (CLI or VSCode extension):
+- `/odata-vocabularies-sync` — update all vocabularies
+- `/odata-vocabularies-sync https://...` — add a new vocabulary (URL must end in `.json`, XML is not supported)
+
+The skill will:
```
heimwege: The README is documenting the skill's behavior and workflow instead of the skill documenting itself.

 - "The skill will: 1. Register the vocabulary... 2. Download... 3. Run tests" - this is exactly what the SKILL.md already describes step by step. It's duplicated, and the README copy will drift out of sync as the skill evolves.
  - After completing, create a changeset with pnpm cset - same as the changeset point already flagged: AGENTS.md covers it, and even if it didn't, it belongs in the skill not the README.

The README's "Maintaining Vocabularies" section should ideally just be a pointer - something like "use the /odata-vocabularies-sync skill" and maybe the two invocation forms (with/without URL). The step-by-step breakdown and the changeset reminder should be dropped from here entirely.

That's a cleaner separation: the README tells you that the skill exists and how to invoke it, the SKILL.md tells you what it does.

heimwege (commented):
see comments

packages/odata-vocabularies/README.md
gernotkuhr: parapraph removed

.agents/skills/odata-vocabularies-sync/SKILL.md
gernotkuhr: changed as proposed

.changeset/odata-vocabularies-readme-fix.md
gernotkuhr: fixed

.agents/skills/odata-vocabularies-sync/SKILL.md
gernotkuhr: fixed

.agents/skills/odata-vocabularies-sync/SKILL.md
gernotkuhr: improved

packages/odata-vocabularies/README.md
gernotkuhr: fixed the path, removed the Claude Code-specific slash command syntax, and kept the URL parameter documented.

packages/odata-vocabularies/README.md
gernotkuhr: correct. The note belongs in the skill (where it already lives) not in the public package README. Removed

.agents/skills/odata-vocabularies-sync/SKILL.md
gernotkuhr: fixed

packages/odata-vocabularies/README.md
gernotkuhr: thank you - improved

packages/odata-vocabularies/README.md
gernotkuhr: simplified README: just pointing to skill for vocabulary update

.agents/skills/odata-vocabularies-sync/SKILL.md
gernotkuhr: info added

.agents/skills/odata-vocabularies-sync/SKILL.md
gernotkuhr: removed

.agents/skills/odata-vocabularies-sync/SKILL.md
```diff
@@ -0,0 +1,137 @@
...
+
+### 1c. Edit `src/resources/index.ts`
+
+Make all six additions. All six are mandatory — missing any one will cause a runtime error.
+
+1. **Import** — insert alphabetically by alias in the single import block (OASIS and SAP imports are interleaved by alias, not grouped by family):
+   ```typescript
+   import <Alias> from './<namespace>.js';
+   ```
+   Use the `$Alias` value from the JSON as the import identifier. **Exception:** if the alias is a reserved JS keyword or built-in (e.g. `JSON`, `Map`, `Error`), prefix it to form a valid identifier (e.g. `JSON` → `ODataJSON`). Check existing imports for precedent.
```
heimwege: The skill tells agents to use `import <Alias> from './<namespace>.js'` where <Alias> is $Alias from the JSON. But `Org.OData.Authorization.V1` has `$Alias = "Authorization"` yet is imported as `Auth`.
`Authorization` is neither a keyword nor a built-in, so the skill gives no guidance for why it becomes `Auth`. Worth a note like "also check existing imports for precedent - aliases may be shortened for brevity."

heimwege (commented):
one additional comment left

.agents/skills/odata-vocabularies-sync/SKILL.md
gernotkuhr: good finding - thank you!

Info added

heimwege (approved):
- review comments addressed
- changeset ok
- did NOT test manually

vinayhospete (approved):

PR #4943, opened 2026-07-14
fix(ui5-test-writer): correct List Report and Object Page OPA test generation

dudajSAP:
## Summary

Fixes several List Report (LR) and Object Page (OP) OPA5 test-generation defects reported against `@sap-ux/ui5-test-writer`, so the generated tests match the actual `sap.fe.test` runtime API and pass reliably.

## Changes

**Filter fields**
- Standard filter fields now use the stable OData property object form `iCheckFilterField({ property })` instead of translatable display labels, so generated tests don't break under translation.
- Custom (manifest `template:`) filter fields, which cannot be matched by property, fall back to their resolved label string (`iCheckFilterField("<label>")`), cast to `FilterFieldIdentifier` to satisfy the mistyped `@sapui5/types` signature. Unresolved `{i18n>…}` labels are resolved via the property's OData `@Common.Label`.

**Actions (List Report and Object Page)**
- Toolbar/table/header/section actions now use the stable action identifier `iCheckAction({ service, action, unbound })` (and the matching `iPressAction`) instead of the translatable action label — e.g. `iCheckAction("Set To Booked", …)` → `iCheckAction({ service: "…", action: "setToBooked", unbound: false }, …)`.
- Applies to LR toolbar actions and to OP header, table-section, and form-section actions; the dynamic/static `enabled` handling is preserved.
- `buildActionButtonState` / `buildActionStateFromSpecModelKey` populate `action` (method name only), `service`, and `unbound`, which feed the identifier object.
- Note: this covers OData-bound actions. Custom (manifest-declared) actions with no OData service are out of scope here and tracked separately.

**Semantic-key "add to filter bar" test**
- Excludes properties annotated with `@UI.HiddenFilter` (never offered by the adaptation dialog) and properties already present in the filter bar.
- Corrects the `Given`/`When`/`Then` subjects (adaptation actions are `When`, assertions are `Then`).
- Adds a value placeholder to the `iChangeFilterField` example.

**Multi-tab List Reports**
- Targets the correct table per tab via `onTable("<key>")` (from the manifest `views.paths[].key`, skipping custom `template:` tabs; a single non-custom tab is treated as a single-table LR).
- Switches to each tab with `iGoToView({ key })` before checking its rows, then back to the first tab before pressing a row.

**Object Page navigation**
- Navigation from a multi-tab parent List Report now targets the parent's default (first non-custom) tab instead of an empty table identifier.

**Template whitespace**
- Removes stray blank lines / trailing whitespace in the generated LR journeys; keeps a single blank line between `opaTest` blocks.

## Downstream packages

The OPA output is snapshotted by app generators, so their snapshots were regenerated to match:
- `@sap-ux/fiori-elements-writer`
- `@sap-ux/fiori-app-sub-generator`

(Both receive an automatic patch bump via the internal-dependency mechanism; the snapshot updates are test-only.)

## Test plan

- [x] `pnpm --filter @sap-ux/ui5-test-writer build`
- [x] `pnpm --filter @sap-ux/ui5-test-writer test` — 425 tests, 24 snapshots pass; coverage ≥ baseline (96.6% stmts / 88.7% branch)
- [x] `pnpm --filter @sap-ux/ui5-test-writer lint` — 0 errors
- [x] `pnpm --filter @sap-ux/fiori-elements-writer test` and `pnpm --filter @sap-ux/fiori-app-sub-generator test` — pass with updated snapshots
- [x] Regenerated JS + TS output for sample apps (LR single-tab and multi-tab, OP); `tsc --noEmit` clean on the TS output


packages/ui5-test-writer/templates/v4/integration/ListReportJourney.js
```diff
@@ -34,75 +34,92 @@ sap.ui.define([
...
+            <%_ } else { -%>
             Then.onThe<%- startLR%>Generated.onTable().iCheckRows();
-<% if (navigatedOP) { %>
-            When.onThe<%- startLR%>Generated.onTable().iPressRow(0);
+            <%_ } -%>
+            <%_ if (navigatedOP) { -%>
+            <%_ if (tableIdentifiers && tableIdentifiers.length > 1) { _%>
+            When.onThe<%- startLR%>Generated.iGoToView({ key: "<%- tableIdentifiers[0] %>" });
+            <%_ } _%>
+            When.onThe<%- startLR%>Generated.onTable(<%- tableIdentifiers && tableIdentifiers.length > 0 ? '"' + tableIdentifiers[0] + '"' : '' %>).iPressRow(0);
```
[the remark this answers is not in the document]
dudajSAP: Addressed in 4d3a10d0f. Rather than only changing the template guard, I fixed the root cause: `getTableIdentifiers` now returns `[]` when `<= 1` non-custom tab survives filtering, so a single-path (or "one custom + one standard") manifest is treated like a single-table LR and uses the default `onTable()` form. With that, `tableIdentifiers` is only ever empty or `>= 2`, and I also simplified the guard to `> 0` for clarity.

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.ts
```diff
@@ -40,74 +39,91 @@ function journey() {
...
+        <%_ } else { -%>
         Then.onThe<%- startLR%>Generated.onTable("").iCheckRows();
-<% if (navigatedOP) { %>
-        When.onThe<%- startLR%>Generated.onTable("").iPressRow(0);
+        <%_ } -%>
+        <%_ if (navigatedOP) { -%>
+        <%_ if (tableIdentifiers && tableIdentifiers.length > 1) { _%>
+        When.onThe<%- startLR%>Generated.iGoToView({ key: "<%- tableIdentifiers[0] %>" });
+        <%_ } _%>
+        When.onThe<%- startLR%>Generated.onTable(<%- tableIdentifiers && tableIdentifiers.length > 0 ? '"' + tableIdentifiers[0] + '"' : '""' %>).iPressRow(0);
```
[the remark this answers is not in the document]
dudajSAP: Addressed in 4d3a10d0f — same fix as the JS template. The root cause is in `getTableIdentifiers` (now returns `[]` for `<= 1` surviving non-custom tab), so the single-surviving-tab case you describe is treated as single-table and never emits `onTable("<key>")`/`iPressRow` without a switch. Guard also simplified to `> 0`.

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.ts
```diff
@@ -19,8 +19,7 @@ import opaTest from "sap/ui/test/opaQunit";
 import type { Given, When, Then } from "./types/OpaJourneyTypes";
 <%_
 const usesFilterFieldIdentifier =
-    (!hideFilterBar && filterBarItems && filterBarItems.length > 0) ||
-    (semanticKey && semanticKey.missingFromFilterBar && semanticKey.missingFromFilterBar.length > 0);
+    !hideFilterBar && filterBarItems && filterBarItems.some(function(item) { return item.custom; });
```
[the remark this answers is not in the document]
dudajSAP: Fixed in 4d3a10d0f. Decoupled the import guard from `hideFilterBar`: `usesFilterFieldIdentifier = filterBarItems && filterBarItems.some(item => item.custom)`. This removes the fragile coupling you flagged.

packages/ui5-test-writer/src/utils/listReportUtils.ts
```diff
@@ -161,9 +165,41 @@ export function getListReportFeatures(
...
+        semanticKeyProperties?.length && filterBarProperties.length
+            ? semanticKeyProperties
+                  .filter((key) => !filterBarProperties.includes(key))
+                  // Exclude properties annotated with `@UI.HiddenFilter` — these are hidden
+                  // from the user's filter adaptation dialog, so trying to add them at runtime
+                  // via `iAddAdaptationFilterField({ property })` produces a "field not found"
+                  // failure.
+                  .filter(
+                      (key) => !convertedMetadata || !isHiddenFilter(convertedMetadata, listReportPage.entitySet, key)
+                  )
```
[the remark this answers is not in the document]
dudajSAP: Addressed in 4d3a10d0f. Extended the comment above the filter to note the graceful-degradation behaviour: when metadata is unavailable the `@UI.HiddenFilter` check is skipped and the property is kept rather than dropped.

packages/ui5-test-writer/src/utils/listReportUtils.ts
```diff
@@ -223,6 +260,180 @@ export function getFilterFieldNames(pageModel: TreeModel, log?: Logger): string[
...
+    if (!Array.isArray(paths) || paths.length === 0) {
+        return [];
+    }
+    const identifiers: string[] = [];
+    for (const path of paths) {
+        if (path && typeof path.key === 'string' && path.key.length > 0 && typeof path.template !== 'string') {
+            identifiers.push(path.key);
+        }
+    }
+    return identifiers;
```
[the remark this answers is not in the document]
dudajSAP: Fixed in 4d3a10d0f. `getTableIdentifiers` now returns `[]` when `identifiers.length <= 1`, so a single-path `views.paths` manifest is treated the same as a single-table LR (default `onTable()`/`onTable("")` form, no tab-switching). Implementation now matches the JSDoc contract. Added unit tests for the single-path and single-surviving-tab cases.

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.js
```diff
@@ -34,75 +34,92 @@ sap.ui.define([
...
         opaTest("Navigate to ObjectPage", function (Given, When, Then) {
             // Note: this test will fail if the ListReport page doesn't show any data
-            <% if (!hideFilterBar) { %>
+            <%_ if (!hideFilterBar) { -%>
             When.onThe<%- startLR%>Generated.onFilterBar().iExecuteSearch();
-            <%} %>
+            <%_ } -%>
+            <%_ if (tableIdentifiers && tableIdentifiers.length > 0) { -%>
+            <%_ tableIdentifiers.forEach(function(tabId) { _%>
+            <%_ if (tableIdentifiers.length > 0) { _%>
```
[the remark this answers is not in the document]
dudajSAP: Fixed in 4a8495090 — removed the always-true inner `if (tableIdentifiers.length > 0)` guard from the per-tab loop; `iGoToView` is now emitted unconditionally inside the loop.

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.ts
```diff
@@ -40,74 +39,91 @@ function journey() {
...
     opaTest("Navigate to ObjectPage", function (_Given: Given, When: When, Then: Then) {
         // Note: this test will fail if the ListReport page doesn't show any data
-        <% if (!hideFilterBar) { %>
+        <%_ if (!hideFilterBar) { -%>
         When.onThe<%- startLR%>Generated.onFilterBar().iExecuteSearch();
-        <%} %>
+        <%_ } -%>
+        <%_ if (tableIdentifiers && tableIdentifiers.length > 0) { -%>
+        <%_ tableIdentifiers.forEach(function(tabId) { _%>
+        <%_ if (tableIdentifiers.length > 0) { _%>
```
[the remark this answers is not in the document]
dudajSAP: Fixed in 4a8495090 — same dead-guard removal in the TS template.

packages/ui5-test-writer/src/utils/listReportUtils.ts
```diff
@@ -161,9 +165,42 @@ export function getListReportFeatures(
...
+                  .filter((key) => !filterBarProperties.includes(key))
+                  // Exclude properties annotated with `@UI.HiddenFilter` — these are hidden
+                  // from the user's filter adaptation dialog, so trying to add them at runtime
+                  // via `iAddAdaptationFilterField({ property })` produces a "field not found"
+                  // failure. When metadata is unavailable the check is skipped (graceful
+                  // degradation — the property is kept rather than dropped).
+                  .filter(
+                      (key) => !convertedMetadata || !isHiddenFilter(convertedMetadata, listReportPage.entitySet, key)
+                  )
             : undefined;
```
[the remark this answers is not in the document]
dudajSAP: Thanks — good catch on the intent, but this `&& filterBarProperties.length` guard is pre-existing on `main` (unchanged by this PR; my diff only swapped `filterBarItems.includes` → `filterBarProperties.includes` and added the HiddenFilter filter). Emitting the semantic-key adaptation block for an empty filter bar would change pre-existing behavior beyond this PR's scope, and the adaptation test builds on `onFilterBar()` interactions that assume filter fields exist. Keeping it out of scope here — happy to address separately if it's a genuine desired change.

packages/ui5-test-writer/src/utils/listReportUtils.ts
```diff
@@ -223,6 +261,186 @@ export function getFilterFieldNames(pageModel: TreeModel, log?: Logger): string[
...
+    convertedMetadata: ConvertedMetadata,
+    entitySetName: string | undefined,
+    propertyName: string
+): boolean {
+    if (!entitySetName) {
+        return false;
+    }
+    const entitySet = convertedMetadata.entitySets.find((es: EntitySet) => es.name === entitySetName);
+    const property = entitySet?.entityType?.entityProperties?.find((p) => p.name === propertyName);
+    return property?.annotations?.UI?.HiddenFilter !== undefined;
```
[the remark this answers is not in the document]
dudajSAP: Fixed in 4a8495090, though the suggested `=== true` wouldn't work: the converted `@UI.HiddenFilter` value from @sap-ux/annotation-converter is a **Boolean wrapper object** (it carries annotation metadata), so it is always truthy AND `=== true` against the primitive is always false. Verified empirically. The correct fix reads the underlying primitive: `property?.annotations?.UI?.HiddenFilter?.valueOf() === true`. Added tests for the Bool=false and unknown-property cases plus integration coverage that a hidden semantic key is dropped from missingFromFilterBar.

packages/ui5-test-writer/src/utils/listReportUtils.ts
```diff
@@ -223,6 +261,186 @@ export function getFilterFieldNames(pageModel: TreeModel, log?: Logger): string[
...
+              };
+          }
+        | undefined;
+    const paths = target?.options?.settings?.views?.paths;
+    if (!Array.isArray(paths) || paths.length === 0) {
+        return [];
+    }
+    const identifiers: string[] = [];
+    for (const path of paths) {
+        if (path && typeof path.key === 'string' && path.key.length > 0 && typeof path.template !== 'string') {
```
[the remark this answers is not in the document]
dudajSAP: Fixed in 4a8495090 — the guard now excludes only non-empty template strings: `!(typeof path.template === 'string' && path.template.length > 0)`, consistent with `getCustomFilterFieldProperties`. An empty-string `template: ""` is now treated as a standard tab. Added a unit test for that case.

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.js
```diff
@@ -34,75 +34,90 @@ sap.ui.define([
...
             <%_ semanticKey.missingFromFilterBar.forEach(function(property) { _%>
-            // Then.onThe<%- startLR%>Generated.onFilterBar().iChangeFilterField("<%- property %>");
+            // When.onThe<%- startLR%>Generated.onFilterBar().iChangeFilterField({ property: "<%- property %>" }, "<value to filter>", false);
             <%_ }); -%>
-            // Then.onThe<%- startLR%>Generated.onFilterBar().iExecuteSearch();
-            // Then.onThe<%- startLR%>Generated.onTable().iCheckRows();
-            // Then.onThe<%- startLR%>Generated.onTable().iSelectRows(0);
-            // Then.onThe<%- startLR%>Generated.onTable().iCheckAction("<Action Name>", { enabled: true });
+            // When.onThe<%- startLR%>Generated.onFilterBar().iExecuteSearch();
+            // Then.onThe<%- startLR%>Generated.onTable(<%- tableIdentifiers && tableIdentifiers.length > 0 ? '"' + tableIdentifiers[0] + '"' : '' %>).iCheckRows();
```
heimwege: nit-pick: The EJS expression `tableIdentifiers && tableIdentifiers.length > 0 ? '"' + tableIdentifiers[0] + '"' : ''` appears ~12 times in the JS template alone, and similarly in the TS template. This is error-prone and hard to maintain.

A `<%_  const tableId = tableIdentifiers && tableIdentifiers.length > 0 ? '"' + tableIdentifiers[0] + '"' : ''; _%>` variable at the top would reduce duplication significantly.

heimwege (commented):
just one nit-pick comment 😄

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.js
dudajSAP: @heimwege I was wondering about that during the implementation. But introducing a const will lead to change in JS template from `onTable()` to `onTable("")` for one table apps. There is no runtime change, but it just complicates the function call and readiness little bit. So the question is "What is worse?". But I am fine with chaning that.

packages/ui5-test-writer/src/utils/listReportUtils.ts
```diff
@@ -161,9 +165,42 @@ export function getListReportFeatures(
         }
     }
 
+    // Resolve unresolved i18n placeholders on custom filter field labels using the
```
sap-sebelao: A bit of a nit-pick:

I usually go by the rule of "If you need a multi line comment to explain your code, the code sucks and needs to be made more readable" - but AI tends to overexplain and leave lengthy comments even over obvious code.

Check the comments the AI left in the code (ther are multiple, I did not mark every single one) to see if they bring any value or if they are just overexplaining simple concepts.

packages/ui5-test-writer/src/utils/listReportUtils.ts
```diff
@@ -223,6 +261,194 @@ export function getFilterFieldNames(pageModel: TreeModel, log?: Logger): string[
...
+    const target = manifest['sap.ui5']?.routing?.targets?.[targetKey] as
+        | {
+              options?: {
+                  settings?: {
+                      controlConfiguration?: Record<string, unknown>;
+                  };
+              };
+          }
+        | undefined;
+    const cc = target?.options?.settings?.controlConfiguration;
```
sap-sebelao: Nit pick:

`cc => controlConfiguration`

I generally do not like naming to be too abbreviated, it sometimes requires you to scroll up to see the assignment line, to understand the abbreviation.

packages/ui5-test-writer/src/utils/listReportUtils.ts
```diff
@@ -223,6 +261,194 @@ export function getFilterFieldNames(pageModel: TreeModel, log?: Logger): string[
...
+    try {
+        const filterBarAggregations = getFilterFields(pageModel);
+        filterBarItems = getSelectionFieldItemsWithLabels(filterBarAggregations);
+    } catch (error) {
+        log?.debug(error);
+    }
+
+    if (!filterBarItems.length) {
+        log?.warn(
+            'Unable to extract filter fields from project model using specification. No filter field tests will be generated.'
```
sap-sebelao: This message is not translatable. Most logs (unless completely technical, like list of files), use text wired through the t() function to retrieve a translatable string.

packages/ui5-test-writer/src/utils/listReportUtils.ts
dudajSAP: Good point — swept the added comments and trimmed the over-explaining JSDoc/inline blocks down to intent-only across `types.ts`, `listReportUtils.ts` and `modelUtils.ts`. Fixed in 13fc14e4.

packages/ui5-test-writer/src/utils/listReportUtils.ts
dudajSAP: Agreed — renamed `cc` to `controlConfiguration`. Fixed in 13fc14e4.

packages/ui5-test-writer/src/utils/listReportUtils.ts
dudajSAP: Wired through `t()` as `warn.noFilterFieldTests`. Fixed in 13fc14e4.

packages/ui5-test-writer/src/translations/ui5-test-writer.i18n.json
```diff
@@ -10,7 +10,8 @@
     "warn": {
         "cannotUpdateJourneyRunner": "Could not update `JourneyRunner.js`. The existing file may be missing or unreadable. New pages were not registered.",
         "cannotUpdateOpaTestsQunit": "Could not update `opaTests.qunit.js`. The existing file may be missing or unreadable. New journey paths were not registered.",
-        "cannotUpdateOpaJourneyTypes": "Could not update `OpaJourneyTypes.d.ts`. The existing file may be missing or unreadable. New journey type definitions were not registered."
+        "cannotUpdateOpaJourneyTypes": "Could not update `OpaJourneyTypes.d.ts`. The existing file may be missing or unreadable. New journey type definitions were not registered.",
+        "noFilterFieldTests": "Unable to extract filter fields from project model using specification. No filter field tests will be generated."
```
lfindlaysap:
```suggestion
        "noFilterFieldTests": "Unable to extract filter fields from the project model using specification. No filter field tests are generated."
```

lfindlaysap (changes requested):

packages/ui5-test-writer/src/translations/ui5-test-writer.i18n.json
dudajSAP: Applied — reworded to match. Fixed in a47d2e68.

dudajSAP:
Reverted List Report `iCheckAction`/`iPressAction` back to the label form (`iCheckAction("<label>", { enabled })`) in `5634aaaaf`. **Object Page keeps the `{ service, action, unbound }` object form** — this asymmetry is intentional.

**Why:** the id-based matcher behind the object form (`service::action`-anchored regex) does not match List Report toolbar action button ids, whose ids carry an entity-type / `Collection` suffix — so the Travels-js integration gate failed on every LR toolbar action. The label form matches via `properties({ text })` on the button, bypassing the id regex, which is the behaviour that worked before this PR. Object Page's `onHeader()` / `onForm()` / `onTable(section)` matchers resolve the object form correctly, so OP stays on it.

**Caveat:** labels are translatable, so LR action assertions are translation-fragile again (the opposite of the property-stability direction we took for filter fields). This is a deliberate trade-off to make the gate pass.

The underlying matcher issue is tracked internally as **DINC0978831** (in progress). Once that is resolved we can switch LR `iCheckAction` back to the stable `{ service, action, unbound }` form to match Object Page.

lfindlaysap (dismissed):

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.ts
```diff
@@ -19,8 +19,7 @@ import opaTest from "sap/ui/test/opaQunit";
 import type { Given, When, Then } from "./types/OpaJourneyTypes";
 <%_
 const usesFilterFieldIdentifier =
-    (!hideFilterBar && filterBarItems && filterBarItems.length > 0) ||
-    (semanticKey && semanticKey.missingFromFilterBar && semanticKey.missingFromFilterBar.length > 0);
+    filterBarItems && filterBarItems.some(function(item) { return item.custom; });
```
docirl: @dudajSAP One more claude review comment, can you check based on the change on this line:

### Finding — FilterFieldIdentifier import emitted but unused when hideFilterBar + a custom filter field coincide

  File: `packages/ui5-test-writer/templates/v4/integration/ListReportJourney.ts:20-25`

  The PR narrowed the import guard to:
  ```const usesFilterFieldIdentifier =
      filterBarItems && filterBarItems.some(function(item) { return item.custom; });
```
  but the only use of FilterFieldIdentifier (the custom-field cast iCheckFilterField("…" as unknown as FilterFieldIdentifier)) lives inside the block guarded by `if (!hideFilterBar && filterBarItems.length > 0)`. The semantic-key block no longer casts (it now uses { property }).

  hideFilterBar is derived independently (from the manifest's settings.hideFilterBar) and does not force filterBarItems empty. So a List Report with hideFilterBar: true and a declared custom filter field emits
  `import type { FilterFieldIdentifier }` while every usage is suppressed → an unused type-only import in the generated .gen.ts, which noUnusedLocals / @typescript-eslint/no-unused-vars flag.

 #### Failure scenario
manifest with hideFilterBar: true + a controlConfiguration.@…SelectionFields.filterFields custom field → generated TS journey has a dangling unused import. Pre-PR the guard included !hideFilterBar, so this is a regression. Not covered by existing tests (the current custom-field test keeps the filter bar visible).

#### Suggested fix
align the import guard with the usage condition, e.g.
  ```const usesFilterFieldIdentifier =
      !hideFilterBar && filterBarItems && filterBarItems.some(function(item) { return item.custom; });
```

dudajSAP: Thanks for your finding. I introduced this bug when reviewing this PR after wrong suggestion from hyperspace-pr-bot. Sorry for that. Fixed here: c2f99fa

docirl (approved):
- Changes look good, tested with mock service and tests are passing
- Feedback comment addressed
- Tests added/updated
- Changeset ok

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.js
heimwege: also fine for me if it stays as it

packages/fiori-app-sub-generator/test/int/fiori-elements/expected-output/alp_v4/webapp/test/integration/MaterialDetailsObjectPageJourney.gen.js
```diff
@@ -56,8 +56,8 @@ sap.ui.define([
             Then.onTheMaterialDetailsObjectPageGenerated.onForm({ section: "MaterialDetailsFacet" }).iCheckField({ property: "FabricationCountry" });
             When.onTheMaterialDetailsObjectPageGenerated.iPressSectionIconTabFilterButton("MaterialRatingsFacet");
             Then.onTheMaterialDetailsObjectPageGenerated.iCheckSection({ section: "MaterialRatingsFacet" });
-            Then.onTheMaterialDetailsObjectPageGenerated.onTable({ property: "_MaterialRatings" }).iCheckAction("Material Ratings Bound Action", { enabled: true });
-            // When.onTheMaterialDetailsObjectPageGenerated.onTable({ property: "_MaterialRatings" }).iPressAction("Material Ratings Bound Action");
+            Then.onTheMaterialDetailsObjectPageGenerated.onTable({ property: "_MaterialRatings" }).iCheckAction({ service: "com.c_salesordermanage_sd_aggregate", action: "MaterialRatingsBoundAction", unbound: true }, { enabled: true });
```
heimwege, line 59: MaterialRatingsBoundAction -> unbound: true
that looks somewhat wrong. If it is a bound action as the name implies than I'd expect `unbound: false`. If it is an unbound action as the parameter value implies I'd expect the name `MaterialRatingsUnboundAction`

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.js
```diff
@@ -34,75 +36,90 @@ sap.ui.define([
...
-            Then.onThe<%- startLR%>Generated.onTable().iCheckDelete({ visible: true });
+            // When.onThe<%- startLR%>Generated.onTable(defaultTableId).iPressDelete();
+            Then.onThe<%- startLR%>Generated.onTable(defaultTableId).iCheckDelete({ visible: true });
             <%_ } _%>
             <%_ toolBarActions.forEach(function(item) { _%>
             <%_ if (item.visible) { _%>
-            // Then.onThe<%- startLR%>Generated.onTable().iPressAction("<%- item.label %>");
-            Then.onThe<%- startLR%>Generated.onTable().iCheckAction("<%- item.label %>", { enabled: <%- item.enabled === true %> });
+            // When.onThe<%- startLR%>Generated.onTable(defaultTableId).iPressAction("<%- item.label %>");
+            Then.onThe<%- startLR%>Generated.onTable(defaultTableId).iCheckAction("<%- item.label %>", { enabled: <%- item.enabled === true %> });
```
heimwege, line 88: LR toolbar actions still use the translatable label string (`item.label`) in the template even though `buildActionButtonState` now populates `item.service`, `item.action`, and `item.unbound` - the same stable identifier fields used for OP actions.

heimwege (commented):
see comments

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.js
dudajSAP: I introduced in the end the `defaultTableId` const - https://github.com/SAP/open-ux-tools/pull/4943/changes#diff-79b561c6c2d38b71730356ee091d10ca8f963271eb00a99f1c208e06937b3fa7R27

devinea (approved):
Changes lg. 
did not test locally.

packages/ui5-test-writer/templates/v4/integration/ListReportJourney.js
dudajSAP: As I mentioned here (https://github.com/SAP/open-ux-tools/pull/4943#issuecomment-5034545215) I reverted the LR toolbar actions back to using label. The original change to `item.service`, `item.action`, and `item.unbound` was causing test failures for travels app. The internal incident is still 'In Progress' and moreover we will need versioning for that.

heimwege: oh ok. sorry. I might have lost track on the comments

packages/fiori-app-sub-generator/test/int/fiori-elements/expected-output/alp_v4/webapp/test/integration/MaterialDetailsObjectPageJourney.gen.js
dudajSAP: From my POV it's a naming issue in the service, there is nothing I can do about it. The test pass just for `unbound: true`, it fails with `unbound: false`.

docirl: Ran this passed claude:

_________

Is the MaterialRatingsBoundAction / unbound: true a Valid Bug?

  No — the reviewer's concern is not a code bug. It is a test data naming issue.

  Here's why:

  - The action MaterialRatingsBoundAction is never defined in the OData metadata schema. It appears only as a reference in a UI.DataFieldForAction annotation.
  - In OData v4, an action that has no definition with IsBound="true" is by definition unbound.
  - The code in actionUtils.ts looks up the action in convertedMetadata.actions, finds nothing (because there's no definition), and correctly falls through to isEntityBound = false → unbound: true.
  - A dedicated test covers this exact case: "builds state for an unbound action (not found in metadata)" — and it passes.

  The action name is misleading, but the code is correct. The PR author's response ("it's a naming issue in the service, there is nothing I can do about it") is accurate. The unbound: true value is right, and it's
  an issue with whoever named the action in the test metadata.

______

Looks to be ok

docirl:
@heimwege I've added a reply to your review comment, looks to be ok.

@heimwege @devinea if comment is ok/addressed, could we try to get this merged today before branching? The reviewer groups are problematic again here, and had previous approvals. Thanks

packages/fiori-app-sub-generator/test/int/fiori-elements/expected-output/alp_v4/webapp/test/integration/MaterialDetailsObjectPageJourney.gen.js
heimwege: ok, thanks for double-checking. If it's just a misleading name it's ok for me. Sorry for my confusion

PR #4946, opened 2026-07-15
Add support to filter full service URL  backend systems from system selection                   

kjose90:
- Adds `backendSystemFilter?: BackendSystemFilter` to `SystemSelectionPromptOptions` in `odata-service-inquirer`                                  
- `repo-app-import-sub-generator` passes `{ connectionType: ['abap_catalog'] }` to exclude full service URL systems, and `{ odata_abap: true }` via destinationFilters for BAS
- `getFloorplanLabel` moved from `app-gen-info.ts` to `app-helpers.ts` in `fiori-generator-shared`

PR #4950, opened 2026-07-15
feat(fiori-generator-shared): export getFloorplanDescription and initI18n

devinea:
## Summary
- `getFloorplanDescription(templateType)` added to `app-gen-info.ts` — mirrors `getFloorplanLabel` but for `floorplans.description.*` keys
- Both `getFloorplanDescription` and `initI18n` exported from the public `index.ts`

## Background
The `floorplans` translations (labels **and** descriptions) were moved into `@sap-ux/fiori-generator-shared` in v1.1.0, but only `getFloorplanLabel` was added to the public API. Consumers that need description strings (e.g. `tools-suite` app generator) had no clean way to access them without importing from the internal `dist/i18n.js` subpath, which breaks in esbuild-bundled environments. Exporting `initI18n` allows callers to explicitly await initialization before using translation functions.

## Test plan
- [ ] Lint passes (`pnpm lint` — pre-existing warnings only, no new errors)
- [ ] Tests pass (`pnpm test` — 61 tests, 13 suites, all green)
- [ ] `getFloorplanDescription` returns correct description string once `initI18n()` is awaited

### DIFF
--- a/packages/fiori-generator-shared/src/app-gen-info.ts
+++ b/packages/fiori-generator-shared/src/app-gen-info.ts
@@ -94,3 +94,13 @@ export function generateAppGenInfo(destPath: string, appGenInfo: AppGenInfo, fs:
 export function getFloorplanLabel(templateType: string, odataVersion?: string): string {
     return t(`floorplans.label.${templateType}`, { defaultValue: templateType, odataVersion });
 }
+
+/**
+ * Returns the description for a given floorplan template type.
+ *
+ * @param templateType - the template type key (e.g. 'lrop', 'fpm')
+ * @returns the description string
+ */
+export function getFloorplanDescription(templateType: string): string {
+    return t(`floorplans.description.${templateType}`, { defaultValue: '' });
+}
--- a/packages/fiori-generator-shared/src/i18n.ts
+++ b/packages/fiori-generator-shared/src/i18n.ts
@@ -43,6 +43,14 @@ export function t(key: string, options?: TOptions): string {
     return (i18n.t as (key: string, opts?: TOptions) => string)(key, options);
 }
 
-initI18n().catch(() => {
+const _initPromise = initI18n().catch(() => {
     // Ignore any errors since the write will still work
 });
+
+/**
+ * Returns a promise that resolves when the module-level i18n initialization completes.
+ * Safe to call multiple times — always returns the same underlying promise.
+ */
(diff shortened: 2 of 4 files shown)


korotkovao (approved):
Straight forward change
Changeset

packages/fiori-generator-shared/src/i18n.ts
```diff
@@ -43,6 +43,14 @@ export function t(key: string, options?: TOptions): string {
...
-initI18n().catch(() => {
+const _initPromise = initI18n().catch(() => {
     // Ignore any errors since the write will still work
 });
+
+/**
+ * Returns a promise that resolves when the module-level i18n initialization completes.
+ * Safe to call multiple times — always returns the same underlying promise.
+ */
+export function waitForI18n(): Promise<void> {
```
IainSAP, line 54: Why is this needed? I dont see it used anywhere...usually consuming modules call initI18n() to ensure texts are loaded. We already have an established pattern to await text loading and this introduces a different approach. Since the texts have been moved then an additional call to init may be needed here: https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-app-sub-generator/src/fiori-app-generator/fioriAppGenerator.ts#L129

packages/fiori-generator-shared/src/app-gen-info.ts
```diff
@@ -94,3 +94,13 @@ export function generateAppGenInfo(destPath: string, appGenInfo: AppGenInfo, fs:
...
     return t(`floorplans.label.${templateType}`, { defaultValue: templateType, odataVersion });
 }
+
+/**
+ * Returns the description for a given floorplan template type.
+ *
+ * @param templateType - the template type key (e.g. 'lrop', 'fpm')
+ * @returns the description string
+ */
+export function getFloorplanDescription(templateType: string): string {
```
IainSAP, line 104: This function should be in the same place as `getFloorplanLabel` see:  https://github.com/SAP/open-ux-tools/pull/4946/changes#diff-41b83c017870301d41e0d6e0acb17b0d23d2658cc158c82e506cb8b253ed9a71R28 it was placed in the wrong file but that is being corrected. But again there are no references to this function in this PR 😕

IainSAP:
Will be fixed in https://github.com/SAP/open-ux-tools/pull/4946

PR #4977, opened 2026-07-23
feat: add PaletteBot - automated VSCode theme change monitor

LiamM2004:
## Summary
- Adds `design-tokens/vscode_themes.json` — the Figma variable collection for VSCode themes (2026 Light, 2026 Dark, Monokai, HC Black) committed as the source of truth
- Adds `scripts/check-vscode-themes.mjs` — fetches upstream VSCode theme files and TypeScript `registerColor()` sources from `microsoft/vscode`, diffs them against the stored JSON, and updates it in place. Changes are grouped by component and output as Markdown
- Adds `.github/workflows/vscode-theme-monitor.yml` — PaletteBot workflow that runs on the 1st of every month, commits any changes directly to `main`, and posts a diff summary as a comment on a persistent tracking issue

## How it works
PaletteBot runs monthly via GitHub Actions. If Microsoft changes any color tokens in the 2026 Light, 2026 Dark, Monokai, or HC Black themes, PaletteBot:
1. Updates `design-tokens/vscode_themes.json` automatically and commits it directly to `main` using a GitHub App token to bypass branch protection
2. Finds or creates the `PaletteBot — VSCode Theme Changes` tracking issue (labelled `palettebot`)
3. Posts a comment on the tracking issue with changes grouped by component (e.g. `button`, `editor`, `list`) showing token name, old hex, and new hex, along with a link to download the updated `vscode_themes.json`

If there are no changes, it posts a "no changes" comment to the tracking issue so there is always a monthly audit trail.

`@SAP/ux-design-reviewers` is mentioned in the issue body and every comment so the team is notified of both changes and monthly no-change confirmations.

## Color resolution
The script resolves color values in priority order:
1. **Theme JSON override** — explicit hex in the upstream theme file
2. **`registerColor()` default** — parsed from TypeScript source files in `microsoft/vscode` discovered dynamically via the GitHub code search API, resolved transitively (references, `transparent()`, `darken()`/`lighten()` using VSCode's exact HSL algorithm from `color.ts`)
3. **Transparent** — fallback when neither source has a value

Tokens with no `registerColor()` call anywhere are tagged `⚠️ no upstream registerColor` in the JSON and listed in every report so they can be investigated/monitored.


LiamM2004:
Example runs: https://github.com/SAP/open-ux-tools/issues/5161
First run expected to be large due to outdated vscode_themes.json - Not updated in this pr to preserve full history in the created issues.

devinea (commented):
Code review findings from automated analysis. See individual inline comments for details.

.github/workflows/vscode-theme-monitor.yml
```diff
@@ -0,0 +1,194 @@
...
+              id: commit
+              run: |
+                  git config user.name "github-actions[bot]"
+                  git config user.email "github-actions[bot]@users.noreply.github.com"
+                  git add design-tokens/vscode_themes.json
+                  git commit --no-verify -m "chore: PaletteBot — update vscode_themes.json with upstream theme changes"
+                  git push origin HEAD:main
+                  echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+
+            - name: Find or create tracking issue
```
devinea, line 75: **Bug: Post-comment runs even when `git push` fails**

The "Post comment on tracking issue" step is conditioned only on `has_changes == 'true'`, not on whether the commit/push succeeded. If `git push origin HEAD:main` fails (e.g. a concurrent push, branch-protection rule, or transient error), the step still runs but `COMMIT_SHA` is empty — the comment is posted with `(unknown)` as the commit reference, falsely implying the snapshot was successfully committed.

**Fix:** Add `&& steps.commit.outcome == 'success'` to the `if:` condition on the comment step.

.github/workflows/vscode-theme-monitor.yml
```diff
@@ -0,0 +1,194 @@
+name: PaletteBot
+
+on:
+    schedule:
+        # Runs on the 1st of every month at 08:00 UTC
+        - cron: '0 8 1 * *'
+    workflow_dispatch:
+
```
devinea, line 8: **Missing `concurrency` group — parallel runs will race**

There is no `concurrency:` key on this workflow. If the monthly cron fires while a `workflow_dispatch` is already running, both jobs will:
1. Check out the same commit
2. Detect the same changes
3. Both try to `git push origin HEAD:main` — the second push fails
4. Both attempt to post tracking-issue comments, creating duplicate entries

**Fix:**
```yaml
concurrency:
  group: palettebot
  cancel-in-progress: true
```

.github/workflows/vscode-theme-monitor.yml
```diff
@@ -0,0 +1,194 @@
...
+                  if [ -f "$DIFF_OUTPUT" ]; then
+                    echo "diff_file=$DIFF_OUTPUT" >> "$GITHUB_OUTPUT"
+                    echo "has_changes=true" >> "$GITHUB_OUTPUT"
+                  else
+                    echo "has_changes=false" >> "$GITHUB_OUTPUT"
+                  fi
+              env:
+                  DIFF_OUTPUT: ${{ runner.temp }}/theme-diff.json
+
+            - name: Commit updated vscode_themes.json
```
devinea, line 61: **`--no-verify` bypasses all pre-commit hooks on bot commits**

Every automated commit skips Husky/lint-staged hooks that guard every other commit in the repo. If `vscode_themes.json` is malformed, contains a pattern that triggers a secret-scanning hook, or violates a formatting rule, nothing catches it before it lands on `main`.

Consider removing `--no-verify` and ensuring the bot commit path satisfies hooks, or at minimum document why it is intentionally bypassed.

.github/workflows/vscode-theme-monitor.yml
```diff
@@ -0,0 +1,194 @@
...
+              if: steps.check.outputs.has_changes == 'true'
```
devinea, line 62: **Hardcoded `HEAD:main` — `workflow_dispatch` from any branch pushes to `main`**

If someone triggers this workflow from a feature branch (e.g. to test a change to the script), the workflow checks out that branch, runs the script, and then pushes that branch's state of `design-tokens/vscode_themes.json` directly to `main`, overwriting whatever was there.

**Fix:** Either restrict dispatch to `main` only, or push to the current branch:
```yaml
# Option A: restrict dispatch
on:
  workflow_dispatch:
    branches: [main]

# Option B: dynamic target
git push origin HEAD:${{ github.ref_name }}
```

.github/workflows/vscode-theme-monitor.yml
```diff
@@ -0,0 +1,194 @@
...
+                          '',
+                          'If a value looks wrong, the most likely cause is a cross-file color reference',
+                          '(e.g. `registerColor(\'foo\', someOtherToken, ...)`) where `someOtherToken` is defined',
+                          'in a file not in `TS_COLOR_SOURCES`. Add the missing file to fix the resolution.',
+                          '',
+                          '@SAP/ux-design-reviewers',
+                        ].join('\n'),
+                      });
+                      core.setOutput('issue_number', newIssue.number);
+
```
devinea, line 164: **`Number("")` = `NaN` when the tracking-issue step fails**

If the "Find or create tracking issue" step throws before `core.setOutput('issue_number', ...)` (e.g. a GitHub API outage, missing label, permissions error), the output is never written. The downstream steps still run because they only check `has_changes`. Then:

```js
const issueNumber = Number("${{ steps.tracking-issue.outputs.issue_number }}");
// → Number("") → NaN
octokit.issues.createComment({ issue_number: NaN, ... })
// → API error or silent no-op — theme diff is lost
```

**Fix:** Guard these steps with `if: steps.tracking-issue.outcome == 'success'` in addition to the `has_changes` check, or use `steps.tracking-issue.outputs.issue_number || 0` with an explicit failure check.

scripts/check-vscode-themes.mjs
```diff
@@ -0,0 +1,908 @@
...
+            const refHex = resolveRef(opaqueMatch[1].trim(), mode, depth + 1);
+            if (!refHex) return null;
+            return normalizeHex(refHex.slice(0, 7)); // strip alpha
+        }
+
+        // oneOf(a, b, ...) — use first resolvable
+        const oneOfMatch = val.match(/^oneOf\((.+)\)$/);
+        if (oneOfMatch) {
+            const args = oneOfMatch[1].split(',').map((s) => s.trim());
+            for (const arg of args) {
```
devinea: **Bug: `oneOf()` args split on every comma, breaking nested function calls**

`split(",")` tokenises the raw argument string, so any `oneOf` call whose arguments contain a comma (e.g. `transparent(someToken, 0.5)`) gets split incorrectly:

```js
// input: oneOf(transparent(contrastBorder, 0), editorWidgetBorder)
oneOfMatch[1].split(",")
// → ["transparent(contrastBorder", " 0)", " editorWidgetBorder"]
// none of the first two fragments resolve → wrong fallback used silently
```

VSCode source already uses `oneOf()` with const-ref args today, and adding a `transparent()` nested arg is a natural future change.

**Fix:** Use a paren-depth-aware splitter (the same logic already implemented elsewhere in this script for `BARE_START_RE`):
```js
function splitArgs(str) {
    const args = [];
    let depth = 0, start = 0;
    for (let i = 0; i < str.length; i++) {
        if (str[i] === "(") depth++;
        else if (str[i] === ")") depth--;
        else if (str[i] === "," && depth === 0) {
            args.push(str.slice(start, i).trim());
            start = i + 1;
        }
    }
    args.push(str.slice(start).trim());
    return args;
}
const args = splitArgs(oneOfMatch[1]);
```

scripts/check-vscode-themes.mjs
```diff
@@ -0,0 +1,908 @@
...
+ * expressions remain null (transparent).
+ */
+function parseTsColorDefaults(source) {
+    const results = {};
+    // Maps JS const name (e.g. MODERN_TAB_ACTIVE_BACKGROUND or listInactiveSelectionBackground) → token name
+    const constToToken = {};
+
+    // Match registerColor calls where the default is an object literal { dark: ..., light: ..., ... }
+    const OBJECT_RE =
+        /(?:(?:export\s+)?const\s+([A-Za-z_][A-Za-z0-9_]*)\s*=\s*)?registerColor\(\s*['"]([^'"]+)['"]\s*,\s*(\{[\s\S]*?\})\s*,/g;
```
devinea: **`OBJECT_RE` lazy `{[\\s\\S]*?}` stops at the first `}` — truncates nested-object defaults blocks**

The non-greedy match stops at the first closing brace it encounters. Any `registerColor` defaults block that contains a nested object or a closing brace in a sub-expression will be captured incompletely:

```js
// e.g. defaults with a type-cast or helper:
registerColor("foo", { dark: Color.fromHex("#fff").transparent(0.1), light: null, ... }, ...)
//                                                                ^--- OBJECT_RE stops here
```

The truncated capture means the per-key regex finds no mode values, the token is recorded as all-null, and it gets tagged `⚠️ no upstream registerColor` on every run.

**Fix:** Use a brace-depth-aware extraction instead of a regex, or use a greedy match paired with backtracking from the last `}` before the description argument.

scripts/check-vscode-themes.mjs
```diff
@@ -0,0 +1,908 @@
...
+const RAW_BASE = 'https://raw.githubusercontent.com/microsoft/vscode/main/';
+
+// The authoritative list of all registered VSCode color token CSS custom properties.
+const KNOWN_VARS_URL =
+    'https://raw.githubusercontent.com/microsoft/vscode/main/build/lib/stylelint/vscode-known-variables.json';
+
+// TypeScript source files in the VSCode repo that define colors via registerColor().
+// Each provides default values for tokens not overridden by theme JSON files.
+const TS_COLOR_SOURCES = [
+    // Platform colors
```
devinea: **`TS_COLOR_SOURCES` is a manually curated list — new VSCode files are silently missed**

When VSCode adds `registerColor()` calls in a new source file not on this list (a routine occurrence as features are added), all tokens defined there get resolved to transparent and tagged `⚠️ no upstream registerColor` on every subsequent run — with no error raised.

This will accumulate silent false-positive warnings until someone audits the list by hand.

**Consider:** dynamically discovering files via the GitHub tree API (filter for `.ts` files under `src/vs/` that contain `registerColor`) instead of a static allowlist. This would be resilient to VSCode restructures and new feature areas.

.github/workflows/vscode-theme-monitor.yml
LiamM2004: Addressed

.github/workflows/vscode-theme-monitor.yml
LiamM2004: Added concurrency group, but using cancel-in-progress: false to queue rather than cancel, if a run has already committed and pushed the JSON but not yet posted the comment, cancelling it would lose the report entirely.

.github/workflows/vscode-theme-monitor.yml
LiamM2004: No verify is needed to commit to this repo, I have copied the logic from commit workflows in the repo: lint.yml, renovate-pr-automation.yml, and pipeline.yml

.github/workflows/vscode-theme-monitor.yml
LiamM2004: Changed to bare git push, matching the pattern used in lint.yml, renovate-pr-automation.yml, and pipeline.yml.

.github/workflows/vscode-theme-monitor.yml
LiamM2004: Addressed

scripts/check-vscode-themes.mjs
LiamM2004: Addressed

scripts/check-vscode-themes.mjs
LiamM2004: Addressed

scripts/check-vscode-themes.mjs
LiamM2004: Addressed

devinea (dismissed):
Thanks for the updates.

mikicvi-SAP (approved):
Thanks @LiamM2004 
- thanks for incorporating feedback and fixing some small issues raised offline
- much more maintainable and easier to read, even though script is quite large.
- workflow looks very good 👍

devinea (approved):
Thanks @LiamM2004  new GHA workflow looks good.

PR #4992, opened 2026-07-29
fix: add input validation to mcp tools

mustafamengutay-sap:
## Summary

This PR improves input validation and error handling consistency in the `fiori-mcp-server` MCP tool handlers.

Previously, `list_functionality` silently accepted empty `appPath` values and swallowed internal errors as plain strings by bypassing the server's standard error response path. `execute_functionality` had a similar gap where omitting `parameters` caused a `TypeError` instead of a clear message.

## Changes

- `src/tools/list-functionalities.ts` - Added a guard that throws 'appPath parameter is required' and removed the `try-catch` so errors propagate to `server.ts`'s top-level error handler, which makes the error response consistent with all other tools.
- `src/types/input.ts` - Added at least one char constraint to `appPath` in `ListFunctionalitiesInputSchema` so the MCP JSON Schema rejects empty strings before the handler is called.
- `src/tools/execute-functionality.ts` — Added a guard that throws 'parameters is required', preventing a `TypeError` when parameters is `undefined`.
- `test/unit/tools/list-functionalities.test.ts` - Updated the existing error-path tests and added a new input validation describe block covering empty-string and whitespace-only `appPath` cases.
- `test/unit/tools/execute-functionality.test.ts` - Added a without-parameters test case asserting the new guard.
- `test/unit/tools/__snapshots__/index.test.ts.snap` - Updated the `list_functionality` input schema snapshot to include `"minLength": 1` reflecting the Zod schema change.

### DIFF
--- a/.changeset/fiori-mcp-server-input-validation.md
+++ b/.changeset/fiori-mcp-server-input-validation.md
@@ -1,0 +1,5 @@
+---
+"@sap-ux/fiori-mcp-server": patch
+---
+
+fix: add input validation and consistent error throwing to list_functionality and execute_functionality tools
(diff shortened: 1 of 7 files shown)


devinea (approved):
Changes look good and covered by tests

vinayhospete (approved):

PR #4994, opened 2026-07-29
feat(ui5-test-writer): Generate Object Page Title test

sap-sebelao:
Internal issue
36742
Internal JIRAs
FXUBRQ34-12490


packages/ui5-test-writer/src/utils/modelUtils.ts
```diff
@@ -84,7 +91,12 @@ export async function getAppFeatures(
         // readApp calls createApplicationAccess internally if given a path, but it uses the "live" version of project-access without fs enhancement
         const appAccess = await createApplicationAccess(basePath, { fs: fs });
         const specification = await appAccess.getSpecification<Specification>();
-        const appModel: ReadAppResult = await specification.readApp({ app: appAccess, fs: fs });
+        const readAppOptions: ReadAppParams & { includeAnnotationProperties: boolean } = {
+            app: appAccess,
+            fs: fs,
+            includeAnnotationProperties: true
+        };
```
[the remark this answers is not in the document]
sap-sebelao: This is a workaround because the proper alternative would be more complex so I left it for later.

The reason the type does not fit (yet) is because ui5-test-writer uses version 1.144.5 of ux-specification. The new `includeAnnotationProperties` flag was added with 1.144.8

If I bump the version to the `.8` patch, the imports break completely, because the ux-spec colleagues reworked the exports between versions in a way makes the current deep imports break.

So this is a temporary band aid fix to avoid mixing chores into a feature PR.

packages/ui5-test-writer/templates/v4/integration/ObjectPageJourney.js
```diff
@@ -61,6 +61,12 @@ sap.ui.define([
         });
 <% } -%>
 
+<% if (headerTitle) { -%>
+        opaTest("Check header title of the Object Page", function (Given, When, Then) {
+            Then.onThe<%- name%>Generated.onHeader().iCheckTitlePath("<%- headerTitle %>");
```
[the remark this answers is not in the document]
sap-sebelao: Fixed

packages/ui5-test-writer/templates/v4/integration/ObjectPageJourney.ts
```diff
@@ -80,6 +80,12 @@ function journey() {
     });
 <% } -%>
 
+<% if (headerTitle) { -%>
+    opaTest("Check header title of the Object Page", function (_Given: Given, _When: When, Then: Then) {
+        Then.onThe<%- name%>Generated.onHeader().iCheckTitlePath("<%- headerTitle %>");
```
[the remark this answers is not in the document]
sap-sebelao: Fixed

packages/ui5-test-writer/test/unit/fiori-elements.test.ts
```diff
@@ -744,6 +744,7 @@ export type Then = Opa5 & BaseArrangements & {
 
             const bookingObjPageJourneyContent =
                 fs.dump()['test/test-output/LROPv4/webapp/test/integration/BookingObjectPageJourney.gen.js'].contents;
+            expect(bookingObjPageJourneyContent).toContain('onHeader().iCheckTitlePath("BookingID")');
```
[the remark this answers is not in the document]
sap-sebelao: Fixed

packages/ui5-test-writer/src/utils/modelUtils.ts
```diff
@@ -84,7 +91,12 @@ export async function getAppFeatures(
         // readApp calls createApplicationAccess internally if given a path, but it uses the "live" version of project-access without fs enhancement
         const appAccess = await createApplicationAccess(basePath, { fs: fs });
         const specification = await appAccess.getSpecification<Specification>();
-        const appModel: ReadAppResult = await specification.readApp({ app: appAccess, fs: fs });
+        const readAppOptions: ReadAppParams & { includeAnnotationProperties: boolean } = {
```
heimwege, line 94: Same as here: https://github.com/SAP/open-ux-tools/pull/5012#discussion_r3710229258

packages/ui5-test-writer/templates/v4/1.150/integration/ObjectPageJourney.js
```diff
@@ -61,6 +61,12 @@ sap.ui.define([
         });
 <% } -%>
 
+<% if (headerTitle) { -%>
+        opaTest("Check header title of the Object Page", function (Given, When, Then) {
```
heimwege: Same here

```suggestion
        opaTest("Check header title of the Object Page", function (_Given, _When, Then) {
```

packages/ui5-test-writer/templates/v4/1.84/integration/ObjectPageJourney.js
```diff
@@ -61,6 +61,12 @@ sap.ui.define([
         });
 <% } -%>
 
+<% if (headerTitle) { -%>
+        opaTest("Check header title of the Object Page", function (Given, When, Then) {
```
heimwege:
```suggestion
        opaTest("Check header title of the Object Page", function (_Given, _When, Then) {
```

heimwege (commented):
see comments + 

In the generated journey, the new title check block is inserted after the navigate to page test but before header facets. This seems correct. However it's worth confirming: does `iCheckTitlePath` require the page to be in `display` mode (not `edit` mode)? If `edit` mode is possible as an initial state (I think it can be with some custom extension), the generated test could be fragile.

packages/ui5-test-writer/src/utils/modelUtils.ts
sap-sebelao: Done

packages/ui5-test-writer/templates/v4/1.150/integration/ObjectPageJourney.js
sap-sebelao: Done

packages/ui5-test-writer/templates/v4/1.84/integration/ObjectPageJourney.js
sap-sebelao: Done

sap-sebelao:
> see comments +
> 
> In the generated journey, the new title check block is inserted after the navigate to page test but before header facets. This seems correct. However it's worth confirming: does `iCheckTitlePath` require the page to be in `display` mode (not `edit` mode)? If `edit` mode is possible as an initial state (I think it can be with some custom extension), the generated test could be fragile.

Looked into the assertion implementation - `iCheckTitlePath` is looking for `ObjectPageDynamicHeaderTitle` which is always present, and then checks the actual binding path value, it should by all means be able to work even in edit mode, but I did not actually test it yet.

Custom implementation would be out of scope for the generator anyway, but edit mode can actually be active even without any custom implementation - in LROP with enabled draft, if the mockdata entry is marked as draft entry, it will auto open in edit mode. I ran into this issue earlier when implementing tests for Actions, and had to resort to generating one assertion commented out. (EDIT: the line(s) in question: https://github.com/SAP/open-ux-tools/blob/main/packages/ui5-test-writer/templates/v4/1.150/integration/ObjectPageJourney.js#L47)

One agreed upon limitation is that the generator does not read the mockdata, so it is up to the app developers to adjust which entry is being opened, if the default one (index 0) does not fit the generated test.

If the title test would be failing on edit mode (which I dont think based on the assertion implementation), we would anyway only be able to at best add a comment to instruct the developers to adjust the opened object page entry to a non-edit mode one - or we would have to generate this whole test case commented out, again for the app devs to decide if they can use it.

heimwege (approved):
- changeset ok
- review comments addressed
- coverage is excellent
- did NOT test manually

devinea (approved):
Did not test locally
changeset ✅

Issue #4997, opened 2026-07-29
BUG - validate-changesets misses cascade bump for packages that consume workspace dependencies via pnpm aliases.


avasilev-sap:
### Related Feature
Feature request: _issue number_

### Description
A clear description of what the bug is.

### Steps to Reproduce
1. Add any change to control-property editor or to preiview-middleware-client 
2. Create changeset: `pnpm cset`
3. Run `pnpm validate:changesets`
=> The cset is valid
Expects errors which explain that the preview-middleware client consumes those packages as alias dependencies and need to be included in the cset.

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #4998, opened 2026-07-29
fix: Enforce cascade changeset for package dependencies which use alias.

avasilev-sap:
#4997 

### Problem
  
When a workspace package is declared as a pnpm alias dependency — where the key in package.json differs from the real package name: `"@sap-ux/control-property-editor-sources": "workspace:@sap-ux/control-property-editor@*"` - the changesets **updateInternalDependencies** cascade does not fire, because it matches only by key. It never sees the real package name behind the alias.
  
### Impact
  
When `@sap-ux/control-property-editor` or `@sap-ux-private/preview-middleware-client` are bumped, @sap-ux/preview-middleware is not included in the version bump commit. Consumers (e.g. tools-suite) stay pinned to the old version of these packages until preview-middleware is manually re-published.
  
### Fix
  
   Extend validate-changesets.mjs to automatically detect alias dependencies by scanning all workspace package.json files and enforce a cascade changeset for the consumer when the aliased real package is being released.


heimwege (approved):
- changeset not needed
- SONAR happy
- did NOT test manually

devinea (approved):
Thanks for he enhancement @heimwege 
👍

PR #5000, opened 2026-07-30
feat(fiori-docs-embeddings): use skill reference files as embeddings source

devinea:
## Summary

- Delete `data_local/opa5_docu.md` and `data_local/fiori-tools-opa-guide.md` (manually maintained, now superseded by skill reference files)
- Add `src/scripts/copy-skill-refs.ts` — at build time, copies `SKILL.md` and all `references/*.md` from `fiori-mcp-server/skills/sap-fiori-opa5-test-development/` into `data_local/skills_copy/sap-fiori-opa5-test-development/`, converting `---` section dividers to `--------------------------------` (the embeddings chunk delimiter) so each section becomes a separately retrievable embedding chunk
- Update `build-embeddings.ts` to recurse `data_local/skills_copy/` subdirectories in addition to the top-level `data_local/` files
- Wire `copy-opa5-skill-refs` into the parallel `build` step in `package.json`
- Add `test/copy-skill-refs.test.ts` (100% coverage) and `test/build-embeddings-integration.test.ts` — reads the real `data_local/` files and verifies every file and skill subdir produces at least one non-empty document
- Update `search-docs-embeddings.test.ts` in `fiori-mcp-server` to reference content from the new skill files instead of the deleted ones

Only `sap-fiori-opa5-test-development` is included in the `SKILLS_TO_EMBED` allowlist — other skills are explicitly excluded until their reference files are ready to be embedded.

## Test plan

- [ ] `pnpm --filter @sap-ux/fiori-docs-embeddings test` — all tests pass
- [ ] `pnpm --filter @sap-ux/fiori-docs-embeddings build` — copies `sap-fiori-opa5-test-development` skill reference files into `data_local/skills_copy/` and rebuilds embeddings
- [ ] `build-embeddings-integration.test.ts` confirms every `data_local/` file and the `sap-fiori-opa5-test-development` skill subdir produce non-empty documents


packages/fiori-docs-embeddings/data_local/opa5_docu.md
```diff
@@ -532,51 +699,49 @@ When.onTheObjectPage.onTable({property: "items"}).iPressRow(0);
...
-};
+iClickMyBuildingBlockButton: function() {
+    return this.waitFor({
+        controlType: "sap.m.Button",
+        properties: { text: "My Building Block Button" },
+        actions: new Press(),
+        errorMessage: "Could not find building block button"
+    });
+}
 ```
```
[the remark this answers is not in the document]
d027132: We tested this coding in the OPA5 skill and it works. Therefore I do not want to change anything

d027132 (dismissed):

d027132 (dismissed):

.changeset/tender-games-exist.md
```diff
@@ -0,0 +1,6 @@
+---
+"@sap-ux/fiori-docs-embeddings": patch
+"@sap-ux/fiori-mcp-server": patch
+---
+
+FIX: Update opa5_docu.md based on testing
```
heimwege: I think what this PR should actually do is:

```suggestion
FIX: Align opa5_docu.md with sap-fiori-opa5-test-development content
```

And then I'm wondering how this can be done because there are currently still some open findings to be integrated into the skill 🤷🏻

packages/fiori-docs-embeddings/data_local/opa5_docu.md
```diff
@@ -6,32 +6,49 @@
...
 
 - Every `opaTest` MUST have at least one `Then` assertion — tests with 0 assertions fail
-- Every journey MUST end with `iTearDownMyApp()` in a final `opaTest`
+- Teardown must be called on `Given` (not chained on a page object), and always after at least one assertion:
+
+```javascript
+// ❌ Wrong — teardown chained on page object, no prior assertion
+opaTest("Clean up", function(Given, When, Then) {
+    Then.onTheList.iSeeThisPage()
+        .and.onTheList.iTeardownMyApp();
```
heimwege: `iTeardownMyApp()` is used throughout (lowercase d). Only one line mentions the actual difference in v2 vs v4

```
(V4: `Given.iTeardownMyApp()`, V2: `Given.iTeardownMyApp()`)
```

If this snipped is not returned by the search the MCP will lead to wrong code.

packages/fiori-docs-embeddings/data_local/opa5_docu.md
```diff
@@ -6,32 +6,49 @@
...
+    Given.iTeardownMyApp();
+});
+```
+
 - Use Given/When/Then semantics correctly: Given=preconditions, When=actions, Then=assertions
 - Method chain with `.and` for multiple actions on the same component
 
 ### Page Object Configuration
 
 - Register all page objects in `JourneyRunner.pages` before running journeys
```
heimwege: In contrast to the skill, the `JourneyRunner` is not explained at all in this documentation

packages/fiori-docs-embeddings/data_local/opa5_docu.md
```diff
@@ -433,62 +554,108 @@ When.onTheObjectPage.onDialog()
...
-    .iSelectRows({CategoryId: "CAT001"});
+// Filter bar field: pass { property: "<ODataPropertyName>" }
+When.onTheListReport.onFilterBar()
+    .iOpenValueHelp({ property: "CustomerID" });
+```
+
+**Searching within value help:**
+
+Search via the generic search field. The search field is only available when the value help entity set is annotated as searchable in `metadata.xml`. Depending on the backend:
+- **CAP**: `Search.SearchRestrictions` with `Searchable: true` (from `@Search.searchable: true` in CDS)
```
heimwege: this is wrong (fixed in the skill)

packages/fiori-docs-embeddings/data_local/opa5_docu.md
```diff
@@ -433,62 +554,108 @@ When.onTheObjectPage.onDialog()
...
-When.onTheObjectPage.onValueHelpDialog().iConfirm();
+Filter by a specific field. Two cases depending on the value help dialog layout. **Determine the case before writing the test** by grepping `metadata.xml` for `SearchRestrictions` on the value help target entity (found via the `NavigationProperty` path, e.g. `_Customer` → `Passenger` entity set):
 
-// Check field was updated
-Then.onTheObjectPage.onForm({section: "GeneralInfo"})
-    .iCheckField("Category", "Electronics");
+- `Searchable: true` present → **Case A**
+- Absent or `Searchable: false` → **Case B**
+
+> For RAP backends, Case A is the default — standard VH entities expose `SAP__capabilities.SearchRestrictions` with `Searchable: true`. When in doubt, check the metadata rather than assuming Case B.
```
heimwege: CAP is missing

heimwege (commented):
To me this document as is should not be merged. It must be aligned with the skill or it will produce different/wrong results.
See comments + e.g. QUnit autostart false is completely missing, Not everywhere RAP is mentioned, the CAP counterpart is provided, ...

Prompt:
```
check the document update at https://github.com/SAP/open-ux-tools/pull/5000 and compare to the information provided in the related skill file at https://github.com/SAP/open-ux-tools/blob/main/packages/fiori-mcp-server/skills/sap-fiori-opa5-test-development/SKILL.md The document in the PR should contain more or less the same information contined in the skill. Nothing should be missing and nothing should be contradicting.
```

I suggest to finalize the skill (because there are still some findings from Markus open) and then create the documentation based on the skill (maybe also automated generation like we do for the create readme). Else this will be out of sync sooner or later and that's what will lead to issues especially if the MCP and the skill are being used in one plugin (which is the way we recommend.

To be honest I would just delete this document because I don't see the benefit in having bot but I see the effort. But it's not up to me to decide 🤷🏻

PR #5007, opened 2026-07-31
fix: resolve application timeout

mustafamengutay-sap:
## Summary

This PR addresses the remaining acceptance criteria from the input validation work by adding a timeout mechanism to `resolveApplication()` in the fiori-mcp-server.

Previously, calling `list_functionality` or `execute_functionality` with an unreachable or slow-to-resolve path could cause the tool handler to hang indefinitely, as `resolveApplication()` had no upper bound on how long it would wait for `createApplicationAccess` or `findProjectRoot` to complete.

## Changes

- `src/utils/schema-utils.ts` — Wrapped `resolveApplication()` with a 7-second timeout using `Promise.race`. The internal logic was extracted into `resolveApplicationInternal()` and the timeout handle is cleared via `finally` to avoid open handles.
- `test/unit/tools/list-functionalities.test.ts` — Extended the input validation describe block with a non-existent path case (asserting static functionalities are still returned) and a timeout scenario using `jest.useFakeTimers`.
- `test/unit/tools/execute-functionality.test.ts` — Added a timeout scenario to the appPath validation describe block, verifying that a hanging `createApplicationAccess` is rejected after the timeout threshold.


815are (dismissed):
checked:
1. code change is easy to understand
2. changeset persists
3. tests are maintained

packages/fiori-mcp-server/src/utils/schema-utils.ts
```diff
@@ -6,13 +6,30 @@ import { join } from 'node:path';
 import * as zod from 'zod';
 import { logger } from './logger.js';
 
+const RESOLVE_APPLICATION_TIMEOUT_MS = 7000;
```
heimwege: `RESOLVE_APPLICATION_TIMEOUT_MS` is not exported, so tests hardcode `8000` as a magic number. If the constant changes, the tests silently test the wrong threshold. It would be better to export the constant and import in the tests.

heimwege (dismissed):
- changeset ok
- coverage perfect
- did NOT test manually

Only added one comment but it's also fine for me to merge as is.

packages/fiori-mcp-server/src/utils/schema-utils.ts
mustafamengutay-sap: Thanks for the catch, @heimwege!

Fixed here: 9912cdd1481f16671300d8d8025d219755d5aead

heimwege (approved):
re-approve

815are (approved):

PR #5012, opened 2026-07-31
feat(ui5-test-writer): Skip object page section marked with UI.Hidden

sap-sebelao:
Internal issue
36742
Internal JIRA
FXUBRQ34-12674


packages/ui5-test-writer/src/utils/modelUtils.ts
```diff
@@ -84,7 +89,12 @@ export async function getAppFeatures(
         // readApp calls createApplicationAccess internally if given a path, but it uses the "live" version of project-access without fs enhancement
         const appAccess = await createApplicationAccess(basePath, { fs: fs });
         const specification = await appAccess.getSpecification<Specification>();
-        const appModel: ReadAppResult = await specification.readApp({ app: appAccess, fs: fs });
+        const readAppOptions: ReadAppParams & { includeAnnotationProperties: boolean } = {
+            app: appAccess,
+            fs: fs,
+            includeAnnotationProperties: true
+        };
```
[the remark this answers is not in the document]
sap-sebelao: This is a workaround because the proper alternative would be more complex so I left it for later.

The reason the type does not fit (yet) is because ui5-test-writer uses version 1.144.5 of ux-specification. The new includeAnnotationProperties flag was added with 1.144.8

If I bump the version to the .8 patch, the imports break completely, because the ux-spec colleagues reworked the exports between versions in a way makes the current deep imports break.

So this is a temporary band aid fix to avoid mixing chores into a feature PR.

packages/ui5-test-writer/src/utils/objectPageUtils.ts
```diff
@@ -612,6 +621,17 @@ function isTableSection(section: BodySectionItem): boolean {
...
 
+/**
+ * Checks whether a section is hidden by a UI.Hidden annotation and should be skipped.
+ *
+ * @param section - section entry from ux specification
+ * @returns true if the section is marked hidden
+ */
+function isSectionHidden(section: SectionItem): boolean {
+    // hideByProperty holds a dynamic hide expression; treat any value as hidden.
+    return section.properties?.hidden?.value === true || !!section.properties?.hideByProperty?.value;
```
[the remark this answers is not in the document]
sap-sebelao: Fixed

packages/ui5-test-writer/src/utils/modelUtils.ts
```diff
@@ -84,7 +89,12 @@ export async function getAppFeatures(
         // readApp calls createApplicationAccess internally if given a path, but it uses the "live" version of project-access without fs enhancement
         const appAccess = await createApplicationAccess(basePath, { fs: fs });
         const specification = await appAccess.getSpecification<Specification>();
-        const appModel: ReadAppResult = await specification.readApp({ app: appAccess, fs: fs });
+        const readAppOptions: ReadAppParams & { includeAnnotationProperties: boolean } = {
```
heimwege, line 102: `ReadAppParams` (from `@sap/ux-specification`) does not declare `includeAnnotationProperties`. The intersection type is a compile-time workaround to pass an undocumented/unreleased option. Please add a comment explaining this is a forward-compatibility shim.

packages/ui5-test-writer/src/utils/objectPageUtils.ts
```diff
@@ -612,6 +621,17 @@ function isTableSection(section: BodySectionItem): boolean {
...
 
+/**
+ * Checks whether a section is hidden by a UI.Hidden annotation and should be skipped.
+ *
+ * @param section - section entry from ux specification
+ * @returns true if the section is marked hidden
+ */
+function isSectionHidden(section: SectionItem): boolean {
+    // hideByProperty holds a dynamic hide expression; any presence means the section may be hidden at runtime.
+    return section.properties?.hidden?.value === true || section.properties?.hideByProperty !== undefined;
```
heimwege: `hideByProperty.value` can be `string | boolean`. If it's false (a static false expression), the section is actually always visible but the current logic still skips it. The comment says "any presence means the section may be hidden at runtime" - but a `false` literal is not a dynamic expression; it's a static always-visible declaration.

Suggestion: Check `section.properties?.hideByProperty?.value !== false` or at minimum add a test case for `hideByProperty: { value: false }` to document the intentional conservative behavior.

packages/ui5-test-writer/test/unit/fiori-elements.test.ts
```diff
@@ -892,6 +892,25 @@ export type Then = Opa5 & BaseArrangements & {
             expect(bookingPressSupplement).toBeGreaterThan(bookingCheckSupplement);
             expect(targetSee).toBeGreaterThan(bookingPressSupplement);
         });
+
```
heimwege, line 906: The `SectionItem.properties` type gained `stashed` alongside `hidden` and `hideByProperty`. A stashed section is collapsed but still visible in the test journey. The existing `getSectionStashedFlag` function uses `properties.stashed.freeText` and is called after the `isSectionHidden` guard. This is correct - but there's no test verifying that a section with `properties: { stashed: ..., hidden: { value: true } }` is still skipped (i.e., `hidden` takes priority over `stashed`).

heimwege (commented):
see comments

packages/ui5-test-writer/src/utils/modelUtils.ts
sap-sebelao: Added a comment, including a TODO: line, to remove this when the dependency gets updated.

EDIT: The TODO is also reflected in an internal JIRA BLI: FXUBRQ34-12701

packages/ui5-test-writer/src/utils/objectPageUtils.ts
sap-sebelao: The test case was already there, just added an additional explicit check for `false` value.

packages/ui5-test-writer/test/unit/fiori-elements.test.ts
sap-sebelao: Done - updated a test fixture to include the stashed scenario.

heimwege (dismissed):
- changeset ok
- review comments addressed
- coverage is excellent
- open SONAR issue (todo) is ok
- did NOT test manually

devinea (dismissed):
Did not test locally
changeset ✅ 

As mentioned Workaround needs to be removed when specification is released next with exports fixed.

devinea (approved):
Re-approving after merge conflicts fixed.

PR #4956, opened 2026-07-17
chore(deps): bump patch-level dependencies

fionan-SAP:
## Description

Bumps patch-level dependencies across the monorepo with zero production code changes. Also aligns `typescript-eslint` to `8.64.0` in root and example packages (previously split between `8.57.2` and `8.61.1`); `@sap-ux/eslint-plugin-fiori-tools` remains at its prior version due to esbuild bundling issues (see Excluded). Adds cascade changesets for `@sap-ux/fiori-mcp-server`, `@sap-ux/eslint-plugin-fiori-tools`, and `sap-ux-sap-systems-ext`.

**Runtime dependencies bumped (affect published packages):**
| Dependency | Old | New | Packages affected | Notes |
|---|---|---|---|---|
| `semver` | 7.8.4 | 7.8.5 | 17 packages + 1 integration test | |
| `@sap-devx/yeoman-ui-types` | 1.25.0 | 1.25.1 | 22 packages (15 runtime deps, 7 devDeps) | |
| `@sap/bas-sdk` | 3.13.9 | 3.13.10 | btp-utils, environment-check, preview-middleware | |
| `qs` | 6.15.2 | 6.15.3 | axios-extension | |
| `@sap/approuter` | ^22.0.0 | 22.0.3 | backend-proxy-middleware-cf | Open range pinned to exact version |
| `applicationinsights` | 3.15.0 | 3.15.1 | telemetry | |
| `sanitize-html` | 2.17.5 | 2.17.6 | ui-components, ui-prompting-examples | |

**Dev/tool dependencies bumped (root, examples, devDeps only):**
| Dependency | Old | New |
|---|---|---|
| `@changesets/cli` | 2.31.0 | 2.31.1 |
| `@eslint/eslintrc` | 3.3.5 | 3.3.6 |
| `postcss` | 8.5.15 | 8.5.22 |
| `typescript-eslint` | 8.61.1 | 8.64.0 |
| `nock` | 14.0.15 | 14.0.16 |
| `fs-extra` | 11.3.5 | 11.3.6 |
| `@types/supertest` | 7.2.0 | 7.2.1 |
| `@types/http-proxy` | ^1.17.5 | 1.17.17 |
| `@ui5/cli` | 4.0.50 | 4.0.57 |
| `promptfoo` | 0.121.15 | 0.121.19 |
| `axios` (examples/simple-generator only) | 1.18.0 | 1.18.1 |
[...]

PR #5013, opened 2026-07-31
fix(telemetry): revert applicationinsights 3.15.1 to 3.15.0

fionan-SAP:
## Description

Reverts `applicationinsights` from `3.15.1` back to `3.15.0` in `@sap-ux/telemetry`.

`3.15.1` ([release notes](https://github.com/microsoft/ApplicationInsights-node.js/releases/tag/3.15.1)) updated `@azure/monitor-opentelemetry-exporter` to a version that ships a nested `@opentelemetry/api@0.10.2`. That old version depends on `@opentelemetry/context-base` — a package that no longer exists (it was absorbed into `@opentelemetry/api` itself in v0.12).

At runtime Node resolves this fine. When bundling with esbuild, esbuild follows every import statically and fails hard when it cannot find the missing package. This broke the `@sap/ux-ui5-tooling` build in downstream consumers.

The bump to `3.15.1` was introduced in #4956.

> **Note:** This is a temporary revert, not a permanent fix. The root cause is in `applicationinsights@3.15.1` itself — Microsoft needs to update `@azure/monitor-opentelemetry-exporter` to drop the stale `@opentelemetry/api@0.10.2` nested dep. When `applicationinsights` is bumped again in a future dependency sweep, the same issue will recur unless that upstream fix has landed first.

## Root cause chain

```
@sap-ux/telemetry
  → applicationinsights 3.15.1
    → @azure/monitor-opentelemetry-exporter (updated in 3.15.1)
      → (nested) @opentelemetry/api@0.10.2
        → @opentelemetry/context-base  ← does not exist
```

## Type of change

## Checklist


devinea:
Superseded by https://github.com/SAP/open-ux-tools/pull/5083

PR #5014, opened 2026-07-31
fix(fiori-mcp): validate metadata output directory

### LINKED ISSUE #4981: BUG - fiori-mcp `download_odata_service_metadata`: clarify `appPath` pre-condition to prevent LLMs from passing a non-existent folder
### Related Feature

Feature request: _N/A — this is a bug report on the existing `download_odata_service_metadata` tool exposed by `@sap-ux/fiori-mcp-server`._

### Description

The MCP tool `download_odata_service_metadata` (exposed by `@sap-ux/fiori-mcp-server`) requires its `appPath` parameter to point to an **already-existing** directory into which `metadata.xml` will be written. This pre-condition is not documented in the tool's input schema description, so LLMs consuming the MCP server routinely pass a folder that has not been created yet — for example, the target folder of a Fiori application they are *about* to scaffold. When that happens the metadata download fails and the whole "download metadata → generate app" chain breaks in a way that is hard for the LLM to recover from.

The problem is not the runtime behaviour per se (writing a file into a non-existent folder is a legitimate error) — the problem is that the tool's description gives the LLM no signal that this pre-condition exists.

**Current parameter description** (from the tool's JSON Schema):

> `appPath` (required, string): "Absolute path to the folder where metadata.xml will be saved. Typically the project target folder."

**Why this misleads the LLM:**

1. *"folder where metadata.xml will be saved"* describes the folder's purpose, not its lifecycle. Many filesystem primitives the LLM has been trained on (`mkdir -p && …`, various generator CLIs, some `fs.writeFile` variants) auto-create parent directories, so absent a rule to the contrary the LLM defaults to that convention.
2. *"Typically the project target folder"* reinforces the wrong intuition: for a **new** project, "the project target folder" is exactly the thing the LLM is about to create. So the LLM concludes it is safe to pass a path that does not exist yet.
3. The description contains no pre-condition line, no failure-mode hint, no post-condition line.
4. The parameter name `appPath` is semantically loaded ("the app's path"), which further hides the fact that this is just "an existing directory to drop a file into".

### Steps to Reproduce

1. From an MCP client (e.g. Claude Code, GitHub Copilot with `fiori-mcp` installed), start a conversation like:
   > "Create a Fiori elements List Report for the OData service at `https://…/sap/opu/odata/sap/<SERVICE>/` and put it under `C:\dev\my-new-fiori-app`."
2. The LLM chains the recommended workflow:
   a. Calls `download_odata_service_metadata` with `appPath: "C:/dev/my-new-fiori-app"` (a path that does **not** exist yet).
   b. Intends to follow up with `generate_fiori_app_odata`.
3. The metadata download step fails because the folder does not exist. Depending on the client, the LLM sees either a raw filesystem error or a truncated failure and cannot recover automatically.

### Expected results

Either:

- **(a)** The tool description makes the pre-condition explicit ("this folder MUST exist; this tool does not create directories; create it with `mkdir -p <appPath>` first"), so the LLM plans a preparatory `mkdir` step; **or**
- **(b)** The server-side handler creates the directory recursively (`fs.mkdirSync(appPath, { recursive: true })`) and the description states this behaviour, so the current LLM planning is correct-by-default; **or**
- **(c)** The server-side handler validates and returns a precise, actionable error message ("`appPath` does not exist. Create the directory before calling this tool: `mkdir -p <appPath>`. This tool does not create directories."), so the LLM can self-correct on the next turn.

**Recommended combination:** (a) + (c) — clarify the contract in the description **and** fail fast with a specific error string. Both are cheap and give defence in depth against LLMs that skim the schema.

**Proposed replacement description for `appPath`:**

> Absolute path to an **existing folder** where `metadata.xml` will be written. The folder MUST exist before this tool is invoked — this tool does **not** create directories. If the caller is about to scaffold a new Fiori project, create the target folder first (e.g. `mkdir -p <appPath>`) and only then call this tool. Typically this is the same folder that will later be passed as the project target to `generate_fiori_app_odata`.

**Proposed fail-fast guard at the top of the handler:**

```ts
if (!fs.existsSync(appPath)) {
  throw new Error(
    `appPath does not exist: ${appPath}. Create the directory before calling this tool ` +
    `(e.g. \`mkdir -p ${appPath}\`). This tool does not create directories.`
  );
}
```

### Actual results

The tool description does not state the pre-condition; the handler does not create the directory; and (based on the observed demo) the resulting error is not verbose enough for the LLM to recognise the root cause and self-correct.

### Screenshots

_N/A — text-based tool invocation. Trace/log excerpt can be added if useful._

OS:
- [ ] Mac OS
- [x] Windows
- [ ] Other

- Component: `@sap-ux/fiori-mcp-server`
- Tool: `download_odata_service_metadata`
- MCP client: GitHub Copilot (VS Code) — model: Claude 4.7 Opus (AI Core)
- fiori-mcp-server version: 1.9.1 (as currently on npm, `dist-tags.latest`)

### Problem
{describe the problem}

### Fix
{describe the fix}

### Why was it missed

### How can we avoid this
{if we don't want to see this type of issues anymore what we should do to prevent}

Pybsama:
## Description

Fixes #4981.

`download_odata_service_metadata` now rejects a missing `appPath` before system lookup, metadata download, or file writes. The exposed input schema now states that the output directory must already exist, that this tool does not create directories, and that `generate_fiori_app_odata` should be used when creating a new application. This also adds regression coverage and a patch changeset.

AI assistance: OpenAI Codex was used to assist with issue analysis, implementation, test authoring, and review. I reviewed the changes and test results; no third-party material was intentionally included.

## Type of change

## How have you tested?

- Focused tool and schema tests: 24 tests and 8 snapshots passed.
- `@sap-ux/fiori-mcp-server` test suite: 31 suites, 396 tests, and 35 snapshots passed.
- Repository test target: 87 of 87 projects succeeded.
- Repository lint target: 95 of 95 projects succeeded with no errors.
- Repository build target: 92 of 92 projects succeeded.
- Prettier, changeset validation, and `git diff --check` passed.

## Checklist:

- [x] Corresponding changes to the documentation has been done

### DIFF
--- a/.changeset/fiori-mcp-metadata-app-path.md
+++ b/.changeset/fiori-mcp-metadata-app-path.md
@@ -1,0 +1,5 @@
+---
+"@sap-ux/fiori-mcp-server": patch
+---
+
+FIX: Clarify and validate the metadata download output directory
--- a/packages/fiori-mcp-server/src/tools/download-odata-service-metadata.ts
+++ b/packages/fiori-mcp-server/src/tools/download-odata-service-metadata.ts
@@ -50,6 +50,13 @@ export async function downloadODataServiceMetadata(
     }
 
     try {
+        if (!fs.existsSync(params.appPath)) {
+            throw new Error(
+                `appPath does not exist: ${params.appPath}. Create the directory before calling this tool. ` +
+                    'This tool does not create directories.'
+            );
+        }
+
         const findResult = await findSystem(sapSystemQuery || servicePath);
         if (!findResult.system) {
             return {
--- a/packages/fiori-mcp-server/src/types/input.ts
+++ b/packages/fiori-mcp-server/src/types/input.ts
@@ -88,7 +88,12 @@ export const DownloadODataServiceMetadataInputSchema = zod.object({
         ), */
     appPath: zod
         .string()
-        .describe('Absolute path to the folder where metadata.xml will be saved. Typically the project target folder.')
+        .describe(
+            'Absolute path to an existing folder where `metadata.xml` will be written. ' +
+                'The folder must exist before this tool is called; this tool does not create directories. ' +
+                'Create the target folder first when scaffolding a new Fiori project. ' +
+                'Typically this is the same folder later passed as the project target to `generate_fiori_app_odata`.'
+        )
 });
 
 export const DocSearchInputSchema = zod.object({
--- a/packages/fiori-mcp-server/test/unit/tools/download-odata-service-metadata.test.ts
+++ b/packages/fiori-mcp-server/test/unit/tools/download-odata-service-metadata.test.ts
@@ -11,21 +11,26 @@ jest.unstable_mockModule('../../../src/tools/services/sap-system', () => ({
 }));
 
 const mockWriteFileSync = jest.fn<any>();
+const mockExistsSync = jest.fn<any>();
 const actualFs = await import('node:fs');
 jest.unstable_mockModule('node:fs', () => ({
     ...actualFs,
     default: {
         ...actualFs,
+        existsSync: mockExistsSync,
         writeFileSync: mockWriteFileSync
     },
+    existsSync: mockExistsSync,
     writeFileSync: mockWriteFileSync
 }));
 jest.unstable_mockModule('fs', () => ({
     ...actualFs,
     default: {
         ...actualFs,
+        existsSync: mockExistsSync,
         writeFileSync: mockWriteFileSync
     },
+    existsSync: mockExistsSync,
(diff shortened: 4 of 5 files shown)


Pybsama:
Thanks for the update. I compared this draft with #5103. The maintainer-side
PR now covers the same fail-fast runtime guard and LLM-visible schema contract,
and it additionally rejects an existing path that is not a directory. Its
cross-platform CI is also green.

To avoid duplicate review work, I am closing this draft in favor of #5103 and
will follow that PR for the resolution of #4981.

Issue #5017, opened 2026-08-03
BUG - Credentials from previously selected system are resubmitted when switching to a new system


avasilev-sap:
### Related Feature
Feature request: _issue number_

### Description
When a user selects a system that requires authentication, enters their credentials and logs in, then switches to a different
  system that also requires authentication, the previously entered password is automatically resubmitted against the new system.

  This happens because the password field retains its value when the system dropdown changes, and yeoman-ui re-triggers the validate
  callback on all visible fields on every model update. As a result, the old credentials are sent to the new system without the
  user's intent.

### Steps to Reproduce
1. Open the ADP generator configuration step
2. Select a system that requires authentication
3. Enter username and password and log in successfully
4. Change the system selection to a different system that also requires authentication
5. Observe: the old password is silently resubmitted to the new system

### Screenshots
If applicable, add screenshots to help explain the problem.

#### Version/Components/Environment	
Add any other context about the problem here
OS:                      
- [ ] Mac OS
- [ ] Windows
- [ ] Other

### Problem
{describe the problem} 

### Fix
{describe the fix} 

### Why was it missed
{Some explanation why this issue might have been missed during normal development/testing cycle}  

### How can we avoid this
{if we don’t want to see this type of issues anymore what we should do to prevent}

PR #5018, opened 2026-08-03
fix: Clear password field on system change to prevent stale credentials resubmission.

avasilev-sap:
#5017 

* Adds `applyDefaultWhenDirty: true` and `default: ''` to every password prompt across generator-adp, abap-deploy-config-inquirer, and inquirer-common.
* When the system dropdown changes, yeoman-ui re-applies the default value to dirty fields, clearing the password and preventing the old credentials from being silently resubmitted against the newly selected system.
* The applyDefaultWhenDirty is the only choice available here because the credentials are sent as part of the validate hook. 
* The doLogin hook for a password prompt, triggered only when the login button for a password prompt is pressed, is not working so far, this features is started 2 years ago but frozen. That's why we decided not to discard the credentials being send in the validate hook for now. Implementing the doLogin feature will require a major refactoring on all places we use the validate callback to log in to a system.

**Testing done**
Manually verified:
When we select system A, login to system A and then switch to system B which also requires login then the password is reset to '' and the system A credentials are not sent to system B. This is verified in the ADP application's page, ADP deploy page and the key-user changes page.


IvoSG (approved):
Changes look good
Changeset is OK
Coverage is good
Did not test manually

cianmSAP (approved):
Changes look good
Have not tested
Changeset looks appropriate

PR #5036, opened 2026-08-10
fix(pipeline): make vsix release resilient to npm publish/release split runs

devinea:
## Summary

- The `Detect extension publication` step was gated on `steps.changesetPublish.outputs.published == 'true'`, so if a previous run already published the extension to npm but crashed before creating the GitHub Release with the vsix, the next run would skip the vsix build entirely
- Adds a fallback path: if no extension was published in the current run, enumerate workspace extension packages, check npm for the latest version, and verify whether the corresponding GitHub Release already has a vsix asset — if not, proceed to build and attach it
- This mirrors how `changesets/action` is already resilient for npm (only publishes what's missing from the registry)

Fixes the missing vsix on https://github.com/SAP/open-ux-tools/releases/tag/sap-ux-sap-systems-ext%401.0.17 — the next merge to main will attach the vsix to that release automatically.

## Test plan

- [ ] Verify the next push to main attaches the vsix to the existing `sap-ux-sap-systems-ext@1.0.17` release
- [ ] Verify a normal release run (extension published fresh) still works via the primary path

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -253,11 +253,35 @@ jobs:
               env:
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
             - name: Detect extension publication
-              if: steps.changesetPublish.outputs.published == 'true'
               run: |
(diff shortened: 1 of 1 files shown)


cianmSAP (approved):
Thanks @devinea lgtm

Issue #5044, opened 2026-08-13
FEATURE - Ability to create entity specific controller extensions


IvoSG:
_As a developer,
I want to choose whether a controller extension in an Adaptation Project targets the base controller or the current page-specific controller,
so that I can extend the controller that best fits my use case._

### Description

Support the creation of instance-specific controller extensions in Adaptation Projects, similar to the functionality already available for base applications.

When creating a controller extension, the dialog should allow developers to choose between:

- **Base Page Controller** (default)
- **This Instance Specific Controller**

If the base controller is selected, the current behavior remains unchanged.

If the instance-specific controller is selected, the extension should be generated according to the RTA implementation
The feature should only be available for supported UI5 versions.

### Technical Design

Enhance the **Create Controller Extension** dialog in Adaptation Editor by adding a radio button group:

**Which controller would you like to extend?**
- Base Page Controller (default)
- This Instance Specific Controller

Use the existing RTA implementation to generate the corresponding controller extension.

### Acceptance Criteria

**Given** I am creating a controller extension in an Adaptation Project
**When** the dialog is opened
**Then** I can choose between extending the Base Page Controller and the Instance Specific Controller
**Given** I select Base Page Controller
**When** I save the controller extension
**Then** the current behavior is preserved
**Given** I select This Instance Specific Controller
**When** I save the controller extension
**Then** a page-specific controller extension is created according to the RTA implementation

### Notes

- Available only for supported UI5 versions.

### Tasks
- [ ] 
- [ ] Unit Tests added
- [ ] Test Cases defined

PR #5045, opened 2026-08-13
FEAT(adp): Ability to create entity specific controller extensions

IvoSG:
[#5044](https://github.com/SAP/open-ux-tools/issues/5044)

## Description

This PR introduces support for creating **entity-specific (instance-specific) controller extensions** in SAP Fiori Adaptation Project (ADP). Previously, controller extensions could only be created at the base page controller level. Now, when supported by the UI5 version (≥ 1.143), users can choose between extending the **Base Page Controller** or an **instance-specific controller** scoped to a particular view.

### Key changes:

**`packages/adp-tooling`**
- `routes-handler.ts`: Extended `GET /adp/api/code_ext` to accept an optional `viewId` query parameter. The response now distinguishes between `baseController*` and `instanceController*` fields, replacing the old single `controllerExists`/`controllerPath` pattern. Handles cases where both base and instance controllers can coexist.
- `types.ts`: Added optional `viewId` field to `CodeExtChange.content`.

**`packages/preview-middleware-client`**
- `api-handler.ts`: Updated `CodeExtResponse` interface and `getExistingController()` to support `viewId` and the new split response shape.
- `ControllerExtension.controller.ts`: Added UI5 version gate (≥ 1.143) for instance-specific controller support. Added radio button group for selecting controller type, separate VS Code open handlers (`onOpenBaseController`, `onOpenInstanceController`), and updated model to track `baseControllerExists`, `instanceControllerExists`, `isInstanceSpecific`, etc.
- `ControllerExtension.fragment.xml`: Added `RadioButtonGroup` with "Base Page Controller" / "This instance specific controller" options and updated the existing-controller form to show both controller paths separately with individual open links.
- `extend-controller.ts`: Added `instanceSpecific` flag to `DeferredExtendControllerData`.
- `utils.ts`: Added `getPendingCodeExtViewIds()` helper to inspect RTA command stack for pending `codeExt` changes per view, and improved `getControllerInfoForControl()` to walk up the view hierarchy when a nested view has no controller.
- `messagebundle.properties`: Added i18n keys for the new UI labels.
- Quick actions (`add-controller-to-page.ts`, `create-page-action.ts`, `create-table-action-config-change.ts`): Updated to use the new `baseController*`/`instanceController*` response fields.

**Tests**: All affected unit tests updated and extended to cover the new instance-specific controller logic, radio button enablement, view ID matching, and VS Code link handlers.

## Type of change
- [ ] Bug (non-breaking change which fixes an issue)

## How have you tested?
- Unit tests added and updated across `adp-tooling` and `preview-middleware-client` packages to cover:
  - Instance-specific vs. base controller detection on the server side
  - `getPendingCodeExtViewIds` utility
  - Radio button enablement based on existing extensions
  - Both VS Code link handlers
  - Version gating for the instance-specific option (UI5 ≥ 1.143)

## Checklist:
- [ ] The code conforms to the [general development principles](https://github.com/SAP/open-ux-tools/blob/main/docs/Guidelines.md)
- [ ] Supplied as many details as possible on this change
- [ ] The code is easy to read and maintainable by others
- [ ] Already existing and new unit tests pass locally

<details>

**Version:** `1.29.26`

- Correlation ID: `9bd3ed90-9719-11f1-90bb-f482b3b74505`
</details>


packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
```diff
@@ -216,47 +224,64 @@ export default class ControllerExtension extends BaseDialog<ControllerModel> {
...
+        if (baseExists && instanceExists) {
+            if (data.baseControllerExists || data.instanceControllerExists) {
+                this.updateModelForExistingController(data);
+            } else {
+                this.updateModelForExistingPendingChange();
+            }
+            return;
+        }
+
+        this.updateModelForNewController(viewId, data.isTsSupported, true, baseExists, instanceExists);
```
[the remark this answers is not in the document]
IvoSG: The radio button is disabled in the UI in this case

packages/preview-middleware-client/src/messagebundle.properties
```diff
@@ -42,6 +42,9 @@ ADP_ADD_FRAGMENT_WITH_TEMPLATE_NOTIFICATION = Note: The `{0}.fragment.xml` fragm
 ADP_CONTROLLER_EXTENSION_EXISTS = An existing controller extension has been found. Please use the previously created extension controller.
 ADP_CONTROLLER_PENDING_CHANGE_EXISTS = A pending change for controller extension has been found. Please use the previously created extension controller after saving the change or delete the pending change.
 ADP_CREATE_CONTROLLER_EXTENSION = Note: The `{0}` controller extension will be created once you save the change.
+ADP_CONTROLLER_TYPE_SELECTION_LABEL = Which controller would you like to extend?
+ADP_CONTROLLER_TYPE_BASE = Base Page Controller
+ADP_CONTROLLER_TYPE_INSTANCE = This instance specific controller
```
lfindlaysap: Can you provide a screenshot?

IvoSG: image

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -196,37 +201,67 @@ export default class RoutesHandler {
...
+                const fileName = isTsSupported ? baseFileName.replace('.js', '.ts') : baseFileName;
+                const controllerPath = getPath(sourcePath, fileName);
+                const controllerPathFromRoot = getPath(projectName, fileName);
+                const changeFilePath = getPath(projectName, file.getName(), '');
+
+                if (!fs.existsSync(controllerPath)) {
+                    missingControllerPath = controllerPath;
+                    missingChangeFilePath = changeFilePath;
+                    continue;
+                }
```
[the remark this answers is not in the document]
IvoSG: fixed in [2610a)](https://github.com/SAP/open-ux-tools/pull/5045/changes/7988ebba618cca1fbddf3204a845bca11dd2610a)

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
```diff
@@ -405,4 +450,33 @@ export default class ControllerExtension extends BaseDialog<ControllerModel> {
...
+    onOpenBaseController(): void {
+        window.open(`vscode://file${this.model.getProperty('/baseControllerPath')}`);
+    }
+
+    /**
+     * Opens the instance-specific controller extension file in VS Code.
+     */
+    onOpenInstanceController(): void {
+        window.open(`vscode://file${this.model.getProperty('/instanceControllerPath')}`);
+    }
```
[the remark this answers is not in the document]
IvoSG: The links can only render when /existingControllerFormVisibility is true, which is set in the same method that sets /isRunningInBAS, and all of it runs before dialog.open(). There's no window where a link shows with isRunningInBAS undefined.

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -173,66 +185,129 @@ export default class RoutesHandler {
...
+                instanceControllerPathFromRoot,
+                missingControllerPath,
+                missingChangeFilePath,
+                isTsSupported
+            } = lookup;
 
-            if (controllerExists && !fs.existsSync(controllerPath)) {
-                const errorMsg = `Please delete the change file at "${changeFilePath}" and retry creating the controller extension.`;
+            if (!baseControllerExists && !instanceControllerExists && missingControllerPath) {
+                const errorMsg = `Please delete the change file at "${missingChangeFilePath}" and retry creating the controller extension.`;
```
nikmace: Question

Could you check this, please: 

The NOT_FOUND guard only fires when *both* controllers are missing:
  ```ts
  if (!baseControllerExists && !instanceControllerExists && missingControllerPath) {
  ```
  If the base change file is stale but a valid instance change exists, \`instanceControllerExists=true\` bypasses the guard entirely. The stale base change is silently dropped with no feedback to the user.

The old code returned a hard 404 for any stale entry. The new silent-ignore is arguably correct (a stale base shouldn't block a valid instance), but the user has no way to know a dangling change file exists.
Suggestion: include a `staleChangeFilePath` field in the 200 response so the client can surface a warning. If silent-ignore is the intended design, a comment in the code and an assertion in the existing test ("ignores a stale base controller…") that `missingChangeFilePath` is *not* returned as an error would make the intent explicit.

The fix commit linked (2610a) doesn't appear to be in the current HEAD (12de852). The guard at line 206 still reads !baseControllerExists && !instanceControllerExists && missingControllerPath, so a stale base alongside a valid instance still silently drops the stale entry. Could you confirm whether the fix is included in this branch or is coming as a follow-up?

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
```diff
@@ -216,47 +224,64 @@ export default class ControllerExtension extends BaseDialog<ControllerModel> {
...
         this.model.setProperty('/inputFormVisibility', false);
         this.model.setProperty('/pendingChangeFormVisibility', false);
         this.model.setProperty('/existingControllerFormVisibility', true);
 
-        if (isRunningInBAS) {
-            this.dialog.getBeginButton().setVisible(false);
-        } else {
-            this.dialog.getBeginButton().setText('Open in VS Code').setEnabled(true);
-        }
+        this.dialog.getBeginButton().setVisible(false);
```
nikmace: `updateModelForExistingController` now always calls `getBeginButton().setVisible(false)` regardless of BAS vs VS Code. Before this PR, VS Code users got an "Open in VS Code" primary button; now the only affordance is the `<Link>` inside the form.

One thing to confirm: on UI5 < 1.143, updateModelForExistingController is reached at line 245 — was replacing the primary button with an in-form link intentional for that version too?

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
```diff
@@ -405,4 +450,33 @@ export default class ControllerExtension extends BaseDialog<ControllerModel> {
...
+     * Opens the instance-specific controller extension file in VS Code.
+     */
+    onOpenInstanceController(): void {
+        window.open(`vscode://file${this.model.getProperty('/instanceControllerPath')}`);
+    }
+
+    private async isInstanceSpecificSupported(): Promise<boolean> {
+        const ui5Version = await getUi5Version();
+        return !isLowerThanMinimalUi5Version(ui5Version, ControllerExtension.INSTANCE_SPECIFIC_MIN_UI5_VERSION);
+    }
```
nikmace: **Medium-Low**

`isControllerExtensionSupported()` and `isInstanceSpecificSupported()` each call `getUi5Version()`, which calls `VersionInfo.load()` — no module-level cache. Within a single `setup()` / `buildDialogData()` call both can be reached, so `VersionInfo.load()` runs twice per dialog open.

Suggestion: call `getUi5Version()` once at the top of `buildDialogData`, inline both version checks against the resolved result, and drop the two private helpers (or make them accept a pre-resolved `Ui5VersionInfo` parameter).

```ts
// in buildDialogData / setup
const ui5Version = await getUi5Version();
const showInstanceSpecificOption = !isLowerThanMinimalUi5Version(ui5Version, ControllerExtension.INSTANCE_SPECIFIC_MIN_UI5_VERSION);
// ... later, instead of this.isControllerExtensionSupported():
const isExtSupported = !isLowerThanMinimalUi5Version(ui5Version, ControllerExtension.CONTROLLER_EXT_MIN_UI5_VERSION);
```

nikmace (changes requested):

packages/adp-tooling/src/preview/routes-handler.ts
IvoSG: The fix is in HEAD and the guard is intentional, now documented with a comment and a test assertion

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
IvoSG: It was unintentional and has now been fixed in [76f374](https://github.com/SAP/open-ux-tools/pull/5045/changes/01ca9c6e3565d78bee64b5b7d4594f440a76f374)

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
IvoSG: Fixed in [76f374](https://github.com/SAP/open-ux-tools/pull/5045/changes/01ca9c6e3565d78bee64b5b7d4594f440a76f374)

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -173,66 +185,133 @@ export default class RoutesHandler {
...
                 this.logger.debug(errorMsg);
                 res.status(HttpStatusCodes.NOT_FOUND).send({ message: errorMsg });
                 return;
             }
 
             const isRunningInBAS = isAppStudio();
 
+            const toResponsePath = (controllerPath: string) =>
+                controllerPath && os.platform() === 'win32' ? `/${controllerPath}` : controllerPath;
```
heimwege: The `toResponsePath` helper calls `os.platform()` on every invocation - so twice per `/adp/api/code_ext` request - where the pre-PR code evaluated it once. A separate pre-existing instance at line 462 (`annotationPath: os.platform() === 'win32' ? ...`) also duplicates the same logic within the same class.

Hoist the check to avoid the redundant call, and consider making it a private method to cover line 462 too:

```ts
  const isWin32 = os.platform() === 'win32';
  const toResponsePath = (p: string) => p && isWin32 ? `/${p}` : p;
```

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
```diff
@@ -216,46 +226,68 @@ export default class ControllerExtension extends BaseDialog<ControllerModel> {
...
+        // Combine persisted (server) and pending (command stack) changes to determine whether a base
+        // page controller and/or an instance-specific controller for this view already exist.
+        const pendingViewIds = getPendingCodeExtViewIds(this.rta, controllerName);
+        const baseExists = data.baseControllerExists || pendingViewIds.some((id) => !id);
+        const instanceExists = data.instanceControllerExists || pendingViewIds.includes(viewId);
+
+        const showInstanceSpecificOption = this.isInstanceSpecificSupported();
+
+        if (!showInstanceSpecificOption) {
+            if (pendingViewIds.some((id) => !id)) {
```
heimwege: The old code used checkForExistingChange(rta, 'codeExt', 'selector.controllerName', controllerName) - any pending codeExt change for this controller showed the "pending change" form. The new pre-1.143 branch uses pendingViewIds.some((id) => !id), which only triggers when a pending base change is present. If a pending instance-specific codeExt change exists on a pre-1.143 system (manually crafted or migrated), it would silently fall through to the "create new controller" form, potentially allowing a second extension to be created.

On pre-1.143, the instance-specific concept doesn't exist, so any pending codeExt for this controller should block creation regardless of whether it carries a viewId. pendingViewIds.length > 0 restores the old "any pending codeExt" semantics:

```ts
  if (!showInstanceSpecificOption) {
      if (pendingViewIds.length > 0) {  // was: pendingViewIds.some((id) => !id)
          this.updateModelForExistingPendingChange();
      } else ...
```

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -173,66 +185,133 @@ export default class RoutesHandler {
...
+
+            const baseFileName = change.content.codeRef.replace('coding/', '');
+            const fileName = isTsSupported ? baseFileName.replace('.js', '.ts') : baseFileName;
+            const controllerPath = getPath(sourcePath, fileName);
+            const controllerPathFromRoot = getPath(projectName, fileName);
+            const changeFilePath = getPath(projectName, file.getName(), '');
+
+            if (!fs.existsSync(controllerPath)) {
+                lookup.missingControllerPath = controllerPath;
+                lookup.missingChangeFilePath = changeFilePath;
```
heimwege: When multiple stale change files are found (their referenced controller files are missing), lookup.missingControllerPath and lookup.missingChangeFilePath are overwritten on each iteration - only the last stale entry's path ends up in the 404 error message. If a user has two stale entries, they delete the named file, retry, and hit the 404 again for the second one. The fix is straightforward - collect all stale paths:

```ts
// change the interface:
  missingChangeFilePaths: string[];  // was: missingChangeFilePath: string

  // in the error message:
  const errorMsg = `Please delete the change file(s) at "${lookup.missingChangeFilePaths.join('", "')}" and retry.`;
```

heimwege (commented):
see comments

packages/adp-tooling/src/preview/routes-handler.ts
IvoSG: Fixed in [78f8](https://github.com/SAP/open-ux-tools/pull/5045/changes/51f7461f188505ae681d46bc702e4ab8121a78f8)

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
IvoSG: Fixed in [78f8](https://github.com/SAP/open-ux-tools/pull/5045/changes/51f7461f188505ae681d46bc702e4ab8121a78f8)

packages/adp-tooling/src/preview/routes-handler.ts
IvoSG: Fixed in [78f8](https://github.com/SAP/open-ux-tools/pull/5045/changes/51f7461f188505ae681d46bc702e4ab8121a78f8)

heimwege (dismissed):
- changeset ok
- review comments addressed
- coverage is great
- did NOT test manually

nikmace (dismissed):
Changes look good
Coverage is good
Changeset is OK
Did not test manually

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -217,22 +213,100 @@ export default class RoutesHandler {
...
         } catch (e) {
             this.handleErrorMessage(res, next, e);
         }
     };
 
+    /**
+     * Resolves base and instance-specific controller extensions from the workspace's codeExt change files.
+     *
+     * @param codeExtFiles codeExt change files read from the workspace
+     * @param controllerName Controller name to match against each change's selector
```
JannaLisa:
```suggestion
     * @param controllerName Controller name matched to each change's selector
```
hmmmm, I was a bit unsure about this 'match against'. Would my proposal also work, without changing the meaning?

packages/adp-tooling/src/preview/routes-handler.ts
```diff
@@ -217,22 +213,100 @@ export default class RoutesHandler {
...
+     * @param viewId Optional view ID used to detect an instance-specific extension for the current view
+     * @returns Existence flags and paths for base/instance controllers, plus any stale change reference
```
JannaLisa:
```suggestion
     * @returns Existence flags and paths for base or instance controllers, plus any stale change reference
```

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
```diff
@@ -401,8 +453,35 @@ export default class ControllerExtension extends BaseDialog<ControllerModel> {
...
 
-    private async isControllerExtensionSupported(): Promise<boolean> {
-        const ui5Version = await getUi5Version();
-        return !isLowerThanMinimalUi5Version(ui5Version, ControllerExtension.CONTROLLER_EXT_MIN_UI5_VERSION);
+    private isControllerExtensionSupported(): boolean {
+        return !isLowerThanMinimalUi5Version(this.ui5Version, ControllerExtension.CONTROLLER_EXT_MIN_UI5_VERSION);
+    }
+
+    /**
+     * Handles selection change on the controller type radio button group.
```
JannaLisa:
```suggestion
     * Handles the selection change on the controller type for the radio button group.
```
yikes. We usually try to avoid stringing more than 3 nouns in a row. Here, we have 5. Hope my proposal is fine.

packages/preview-middleware-client/src/adp/utils.ts
```diff
@@ -113,23 +113,102 @@ export function matchesChangeProperty(command: FlexCommand, propertyPath: string
...
+    getController?(): { getMetadata(): { getName(): string } } | undefined;
+    getParent?(): ManagedObject | null;
+}
+
+/**
+ * Resolves the controller name of a view, preferring the typed controller module name and falling
+ * back to the controller instance's metadata name. Returns an empty string when the view owns no
+ * controller.
+ *
+ * @param view The view to resolve the controller name for.
```
JannaLisa:
```suggestion
 * @param view The view for which to resolve the controller name.
```

JannaLisa (commented):
please get back to me in case my proposals don't make sense.

packages/adp-tooling/src/preview/routes-handler.ts
IvoSG: Changed in [239abcba](https://github.com/SAP/open-ux-tools/pull/5045/changes/f8b6bd4c3034b66d47e76ca7505a97fe239abcba)

packages/adp-tooling/src/preview/routes-handler.ts
IvoSG: Changed in [239abcba](https://github.com/SAP/open-ux-tools/pull/5045/changes/f8b6bd4c3034b66d47e76ca7505a97fe239abcba)

packages/preview-middleware-client/src/adp/controllers/ControllerExtension.controller.ts
IvoSG: Changed in [239abcba](https://github.com/SAP/open-ux-tools/pull/5045/changes/f8b6bd4c3034b66d47e76ca7505a97fe239abcba)

packages/preview-middleware-client/src/adp/utils.ts
IvoSG: Changed in [239abcba](https://github.com/SAP/open-ux-tools/pull/5045/changes/f8b6bd4c3034b66d47e76ca7505a97fe239abcba)

nikmace (approved):

JannaLisa (approved):
Thanks for the changes, looks good now for me!

heimwege (approved):
re-approve

PR #5050, opened 2026-08-14
fix(jest-file-matchers): fix updateSnapshot for nested folders

tom-sap:
[required for PR #38043 to fix int tests on windows]

## Description
When running `--updateSnapshot`, it deletes items from the expected/snapshot folder that no longer exist in the generated output (i.e., "stale" snapshot entries). 
- `unlinkSync` only works on files — if a diff entry is a folder it throws:
	EISDIR: illegal operation on a directory on Linux/macOS
	EPERM: operation not permitted on Windows
- `rmSync` handles both.


IainSAP (commented):
Thanks @tom-sap 

- Fix seems appropriate
- Is there a test to cover/update?

IainSAP (approved):
Thanks @tom-sap 

- Targeted update addresses file snapshot dir removal
-  Tests to cover

devinea (approved):
Fix looks valid and covered by tests. 
Changeset ✅

PR #5052, opened 2026-08-16
fix(fiori-mcp-server): guard against pre-existing app subfolder in generate tools

devinea:
## Summary

- `generate_fiori_app_odata` and `generate_fiori_app_cap` now check whether the target app subfolder already exists before invoking the headless generator, and return a clear error if it does — instead of letting `@sap/fiori:headless --force` silently delete and recreate it
- Updated `targetFolder` schema description to warn that `<targetFolder>/<project.name>` must not already exist
- Added a prominent `🚨 CRITICAL - TARGET FOLDER MUST NOT EXIST 🚨` block to both tool descriptions in `tools/index.ts`

## Root cause

`@sap/fiori:headless` calls `fs.rmSync(projectPath, { recursive: true, force: true })` when `--force` is passed and the app subfolder already exists. The MCP server always passed `--force`, so if an LLM pre-created the app folder, it was silently wiped and recreated — visible as a folder briefly vanishing in the filesystem.

## Test plan

- [ ] Existing unit tests pass (405 tests, 0 failures)
- [ ] New test `should return error when app subfolder already exists` verifies the early-exit behaviour for OData generator
- [ ] Manually verify: calling `generate_fiori_app_odata` with a `targetFolder/<name>` that already exists returns `status: "Error"` with a descriptive message and does not invoke `yo`


IainSAP (approved):
Thanks @devinea 

- Updates as expected
- Tests added to cover
- Manually checked

tom-sap (approved):

PR #5080, opened 2026-08-19
feat(project-access): add resolveApplicationTitle to resolve app title

mustafamengutay-sap:
Adds `resolveApplicationTitle` to `@sap-ux/project-access` to resolve i18n keys from `manifest.json`.


packages/project-access/src/project/app-title.ts
```diff
@@ -0,0 +1,51 @@
...
+        return rawTitle;
+    }
+
+    if (!manifestPath) {
+        return undefined;
+    }
+
+    try {
+        const i18nPaths = await getI18nPropertiesPaths(manifestPath, manifest, memFs);
+        const bundle = await getPropertiesI18nBundle(i18nPaths['sap.app'], memFs);
```
devinea, line 46: `.value.value` is the correct path (`I18nEntry.value` is a `TextNode = ValueNode<string>`, so the second `.value` gives the string), but the optional chain should extend all the way:

```ts
// current — throws if I18nEntry.value is somehow absent
return bundle[i18nKey]?.[0]?.value.value;

// should match preview-middleware config.ts:411
return bundle[i18nKey]?.[0]?.value?.value;
```

devinea (commented):
Overall clean and well-tested. A few issues inline — main blockers are the semver bump type and the missing `?.` in the optional chain.

.changeset/project-access-resolve-app-title.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-ux/project-access": patch
```
devinea, line 2: This is a new exported public function, so the bump type should be `minor` not `patch`. The `FEAT:` prefix in the summary already signals that — the bump type and prefix are inconsistent.

```diff
-"@sap-ux/project-access": patch
+"@sap-ux/project-access": minor
```

packages/project-access/src/project/app-title.ts
```diff
@@ -0,0 +1,51 @@
...
+        return bundle[i18nKey]?.[0]?.value.value;
```
devinea, line 47: `.value.value` is the correct path — `I18nEntry.value` is a `TextNode` (`ValueNode<string>`), so the second `.value` gives the actual string. But the optional chain should extend all the way:

```ts
// current — throws if I18nEntry.value is somehow absent (e.g. malformed parse result)
return bundle[i18nKey]?.[0]?.value.value;

// should be — matches preview-middleware config.ts:411
return bundle[i18nKey]?.[0]?.value?.value;
```

packages/project-access/src/project/app-title.ts
```diff
@@ -0,0 +1,51 @@
...
+    } catch {
```
devinea, line 48: Silent `catch {}` swallows all errors — corrupt manifest, wrong path derivation, and programmer errors all return `undefined` without any trace. `preview-middleware`'s equivalent `getI18nTextFromProperty` logs `logger.warn('Failed to load i18n properties bundle')` before returning. Consider at minimum a debug-level log here so failures are diagnosable in the field.

packages/project-access/test/project/app-title.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    getPropertiesI18nBundle: mockGetPropertiesI18nBundle
+}));
+
+const mockGetI18nPropertiesPaths = jest.fn<typeof i18nType.getI18nPropertiesPaths>();
+jest.unstable_mockModule('../../src/project/i18n/i18n.js', () => ({
+    getI18nPropertiesPaths: mockGetI18nPropertiesPaths
+}));
+
+const { resolveApplicationTitle } = await import('../../src/project/app-title.js');
+
```
devinea, line 25: Note for follow-up: `preview-middleware` has equivalent logic in `getI18nTextFromProperty` (`packages/preview-middleware/src/base/config.ts:395`) — same `extractDoubleCurlyBracketsKey` + `sap.app` bundle lookup pattern. Once this function is merged, `getI18nTextFromProperty` should be refactored to delegate to `resolveApplicationTitle` to avoid the two implementations diverging. Two behavioural differences to be aware of when doing that refactor: (1) `getI18nTextFromProperty` returns the raw key string when the key is not found in the bundle; `resolveApplicationTitle` returns `undefined`. (2) `getI18nTextFromProperty` logs a warning on error; this one is silent.

packages/project-access/src/project/app-title.ts
mustafamengutay-sap: Thanks @devinea

It catches and logs an error message: https://github.com/SAP/open-ux-tools/pull/5080/changes/7a44071072a969f401fc9e175da6e06f5fb14aa4

packages/project-access/src/project/app-title.ts
mustafamengutay-sap: Thanks @devinea

Fixed here: https://github.com/SAP/open-ux-tools/pull/5080/changes/7a44071072a969f401fc9e175da6e06f5fb14aa4

packages/project-access/src/project/app-title.ts
```diff
@@ -0,0 +1,55 @@
...
+ * @param params.manifest - pre-parsed manifest content; pass to avoid re-reading from disk
+ * @param memFs - optional mem-fs-editor instance
+ * @param logger - optional logger instance
+ * @returns the resolved title string, or undefined if it cannot be determined
+ */
+export async function resolveApplicationTitle(
+    params: { manifestPath?: string; manifest?: Manifest },
+    memFs?: Editor,
+    logger?: Logger
+): Promise<string | undefined> {
```
815are, line 24: maybe merge `memFs and logger to options object or existing params object? See other examples in code, where memFs and logger are in options parameter - https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/project/specification.ts#L27 https://github.com/SAP/open-ux-tools/blob/main/packages/project-access/src/command/npm-command.ts#L18

packages/project-access/src/project/app-title.ts
```diff
@@ -0,0 +1,55 @@
...
+        return undefined;
+    }
+
+    const i18nKey = extractDoubleCurlyBracketsKey(rawTitle);
+    if (!i18nKey) {
+        return rawTitle;
+    }
+
+    if (!manifestPath) {
+        return undefined;
```
815are: seems `manifestPath` property/param should be required and not optional? without passing `manifestPath` - logic in `getI18nPropertiesPaths` will not be able to resolve i18n paths

packages/project-access/src/project/app-title.ts
mustafamengutay-sap: Thanks @815are!

Applied the same parameter pattern: b2cf4f4aec0b1b0540de9a29bd86daa22d50f697

packages/project-access/src/project/app-title.ts
mustafamengutay-sap: Thanks @815are!

Refactored here: aacd4feac39c59ed3a843c262d800c0bde89f522

815are (approved):
checked:
1. code - code in new function is easy to understand. Review comments are addressed;
2. changelog exits
3. new unit tests are added

I did not test manually

devinea (approved):
new resolveApplicationTitle is clear and well tested. 
changesets ✅

PR #5098, opened 2026-08-24
fix: `archiveProject` method throws error `Cannot read properties of undefined (reading 'test')` because of incompatible versions of `ignore` package between `@sap-ux/environment-check` and  `glob-gitignore`(expected v5).

815are:
Issue after -> https://github.com/SAP/open-ux-tools/pull/4966
`ignore` depeendency was updated from v5 to v7.
In result [archiveProject ](https://github.com/SAP/open-ux-tools/blob/main/packages/environment-check/src/archive/index.ts#L83) method throws error `Cannot read properties of undefined (reading 'test')` because of incompatible versions of `ignore` package between `@sap-ux/environment-check` and  `glob-gitignore`(expected v5).
image
image

As I checked `glob-gitignore` expects that if we pass `ignore` 
 param as object, then it should be passed using `ignore` v5, because `glob-gitignore` has depeendnecy `ignore` as v5 -> https://github.com/kaelzhang/node-glob-gitignore/blob/master/package.json#L53
As solution we can avoid passing param `ignore` as object and pass raw value, because `glob-gitignore` applies same `ignore().add(ignores)` in its code -> https://github.com/kaelzhang/node-glob-gitignore/blob/master/src/util.js#L36

Also in documentation raw strings and arrays are acceptable -> https://www.npmjs.com/package/glob-gitignore?activeTab=readme#optionsignore


devinea (dismissed):
Thanks @815are 
Changes look good. 
changeset ✅

devinea (approved):
Re-approved

vadson71 (approved):
Bugfix changes look good
Addressed in tests
Approving

PR #5101, opened 2026-08-24
fix(telemetry): initialize reportingTelemetryClient lazily to avoid undefined on happy path

### LINKED ISSUE #5100: fix(telemetry): reportingTelemetryClient undefined when reportEnableTelemetryOnOff called on happy path
## Bug Description

`TELEMETRY_SETTINGS_INIT_FAILED` telemetry events are being received with the error:

```
Cannot read properties of undefined (reading 'trackEvent')
```

Stack:
```
reportEnableTelemetryOnOff at (sap-ux-service-modeler-extension-.../dist/extension-min.cjs:...)
asyncreadEnableTelemetry at (...)
asyncactivate at (...)
```

## Root Cause

In `packages/telemetry/src/base/utils/reporting.ts`, `reportingTelemetryClient` is a module-level variable that is **only ever assigned inside `reportRuntimeError`** (the error path):

```ts
let reportingTelemetryClient: appInsights.TelemetryClient; // never initialized on happy path

export const reportRuntimeError = (error: Error): void => {
    if (process.env.SAP_UX_FIORI_TOOLS_DISABLE_TELEMETRY?.trim() !== 'true') {
        reportingTelemetryClient = new appInsights.TelemetryClient(...); // only assigned here
        configAzureTelemetryClient(reportingTelemetryClient);
    }
    // ...
};

export const reportEnableTelemetryOnOff = (...): void => {
    // ...
    reportingTelemetryClient.trackEvent(...); // 💥 undefined when no error occurred
};
```

The normal activation flow is:

```
activate()
  → initTelemetrySettings()
    → readEnableTelemetry()
      → setEnableTelemetry(true/false)
        → reportEnableTelemetryOnOff()   ← crashes: reportingTelemetryClient is undefined
```

Since `initTelemetrySettings` succeeds (no error), `reportRuntimeError` is never called, `reportingTelemetryClient` stays `undefined`, and the crash fires — which is then ironically reported as a `TELEMETRY_SETTINGS_INIT_FAILED` event via a separate inline `TelemetryClient` constructed in `reportRuntimeError`.

## Fix

Replace the bare module-level variable with a lazy initializer so both functions share the same initialization path:

```ts
let reportingTelemetryClient: appInsights.TelemetryClient | undefined;

const getReportingTelemetryClient = (): appInsights.TelemetryClient => {
    if (!reportingTelemetryClient) {
        reportingTelemetryClient = new appInsights.TelemetryClient(TelemetrySettings.azureInstrumentationKey);
        configAzureTelemetryClient(reportingTelemetryClient);
    }
    return reportingTelemetryClient;
};
```

Both `reportRuntimeError` and `reportEnableTelemetryOnOff` then call `getReportingTelemetryClient().trackEvent(...)`.

## Additional Note

There is also a minor inconsistency in `reportRuntimeError`: the initialization guard uses `?.trim() !== 'true'` but the `trackEvent` guard uses `!== 'true'` (no trim). Worth aligning these.

## Affected Versions

Observed in `sap-ux-service-modeler-extension` 1.31.1 on Linux. all others also..

Check results with 
```KQL
customEvents
| where timestamp > ago(90d)
| where name has "TELEMETRY_SETTINGS_INIT_FAILED"
| summarize count() by bin(timestamp, 1d)
| order by timestamp asc
```

devinea:
## Summary

- `reportingTelemetryClient` was declared as a module-level variable but only ever assigned inside `reportRuntimeError` (the error path)
- `reportEnableTelemetryOnOff` also calls `.trackEvent()` on it, but on the normal activation flow (no error) `reportRuntimeError` is never called, leaving the variable `undefined` → crash
- Replaced with a lazy `getReportingTelemetryClient()` helper that initializes the client on first use; both functions now go through it

Fixes #5100

## Test plan

- [ ] Verify no `TELEMETRY_SETTINGS_INIT_FAILED` events fire on normal extension activation
- [ ] Verify `reportEnableTelemetryOnOff` correctly tracks the enable/disable event on first activation
- [ ] Verify `reportRuntimeError` still works correctly when called from the error path
- [ ] Run `pnpm test --filter @sap-ux/telemetry`


docirl (approved):
Changes look good
Not tested manually

IainSAP (approved):
Thanks @devinea 

- Refactored appropriately
- Not tested locally

PR #5102, opened 2026-08-25
feat: implement variables for spacing

mushegh-sap:
Creation of a standardized spacing system to align with product standards by using the structure of VS Code css variables:
0px: no variable needed, stays none
3px: var(--ui-spacing-xs);
5px: var(--ui-spacing-small);
10px: var(--ui-spacing-medium);
15px: var(--ui-spacing-large);
20px: var(--ui-spacing-xl);
50px: var(--ui-spacing-xxl);
=> Goal: Automatically update the spacing of the AIP with the help of AI based on the defined spacing standards
=> Note: As many existing spacing values ​​are odd, there could be a rounding mechanism like: 12px automatically rounded down to 10px, 14px rounded up to 15px etc.


815are (commented):
@mushegh-sap 

I think comment is valid -> https://github.com/SAP/open-ux-tools/pull/5102#discussion_r3850530886

mushegh-sap:
> @mushegh-sap
> 
> I think comment is valid -> [#5102 (comment)](https://github.com/SAP/open-ux-tools/pull/5102#discussion_r3850530886)

Andis, Thanks for checking. I reviewed the suggestion, but I believe the import should remain. _spacing.scss is shipped as an individual SCSS partial, so keeping @import './variables' makes it independently compilable instead of relying on the import order in ui-components.scss. This also follows the existing _shadows.scss pattern. Since _variables.scss only defines Sass variables and emits no CSS, it does not produce duplicate CSS output.  
If we want all partials to rely exclusively on the entry-point ordering, I think we should handle that consistently for _shadows.scss as well. What do you think?

815are:
> > @mushegh-sap
> > I think comment is valid -> [#5102 (comment)](https://github.com/SAP/open-ux-tools/pull/5102#discussion_r3850530886)
> 
> Andis, Thanks for checking. I reviewed the suggestion, but I believe the import should remain. _spacing.scss is shipped as an individual SCSS partial, so keeping @import './variables' makes it independently compilable instead of relying on the import order in ui-components.scss. This also follows the existing _shadows.scss pattern. Since _variables.scss only defines Sass variables and emits no CSS, it does not produce duplicate CSS output. If we want all partials to rely exclusively on the entry-point ordering, I think we should handle that consistently for _shadows.scss as well. What do you think?

I saw `_shadows.scss` - I think there also same issue with duplicate `variables`, but for example https://github.com/SAP/open-ux-tools/blob/main/packages/ui-components/src/styles/_mixins.scss does not import `variables` while they using them. Also additional indication is that all files are with `_` which seems indicates they are considered as private/internal, while `ui-components.scss` is considered as public.
But as I rechecked - at the end it should not give any side effect for now as `_variables.scss` only has sass variables and there no real styles which can produce duplication

815are (approved):
checked:
1. code changes are straightforward
2. changelog persists
3. storybook example is added

AlinaGovoruhina (approved):
Change for added spacing variables is clear, storybook examples and changesets added.

PR #5103, opened 2026-08-25
fix: appPath guard in download_odata_service_metadata returns

tom-sap:
# Description

- fix for #4981

## Summary

- Updated `appPath` description in the Zod schema to explicitly state the folder must exist before the tool is invoked, and that the tool does not create directories
- Added a fail-fast guard that returns a structured `{ status: 'Error' }` response when `appPath` does not exist, with an actionable message telling the LLM to create the directory first
- Fixed the `fs` mock in tests to include `existsSync`, and added a test covering the non-existent path case


packages/fiori-mcp-server/src/tools/download-odata-service-metadata.ts
```diff
@@ -38,6 +38,19 @@ export async function downloadODataServiceMetadata(
         };
     }
 
+    if (!fs.existsSync(params.appPath) || !fs.statSync(params.appPath).isDirectory()) {
+        return {
+            status: 'Error',
+            message:
+                `appPath does not exist or is not a directory: ${params.appPath}. Create the directory before calling this tool ` +
```
lfindlaysap, line 45:
```suggestion
                `appPath does not exist or is not a directory: ${params.appPath}. Create the directory before calling this tool, ` +
```

packages/fiori-mcp-server/src/tools/download-odata-service-metadata.ts
```diff
@@ -38,6 +38,19 @@ export async function downloadODataServiceMetadata(
...
+                `(e.g. on Unix: \`mkdir -p ${params.appPath}\`, on Windows: \`mkdir "${params.appPath}"\`). This tool does not create directories.`,
```
lfindlaysap, line 46:
```suggestion
                `for example on Unix: \`mkdir -p ${params.appPath}\` and on Windows: \`mkdir "${params.appPath}"\`). This tool does not create directories.`,
```

lfindlaysap (dismissed):

packages/fiori-mcp-server/src/tools/download-odata-service-metadata.ts
tom-sap: Thanks @lfindlaysap, but none of these are visible in the UI, or to the user. These are error messages for the LLM agent that was calling the mcp tool (so the fewer characters the better 😉). Do you mind if we keep them like they were?

devinea (approved):
@tom-sap changes look good. 
Changset ✅

PR #5105, opened 2026-08-25
chore: update @sap/ux-cds-compiler-facade to 1.25.0

Jimmy-Joseph19:
Update @sap/ux-cds-compiler-facade dependency from 1.23.0 to 1.25.0 in fiori-annotation-api and cds-odata-annotation-converter packages. Fix imports to use named exports instead of default export as the new version no longer provides a default export.


vadson71 (approved):
Dependency update looks good
Approving

devinea (approved):
package update looks good to me.

PR #5107, opened 2026-08-26
 feat(fiori-mcp-server): Add ADP controller extension and adaptation project generation tools

nikmace:
## Description
Issue: https://github.com/SAP/open-ux-tools/issues/5184

Adds opt-in SAP Fiori Adaptation Project (ADP) tooling to `@sap-ux/fiori-mcp-server` behind `SAP_FIORI_MCP_ADP_TOOLS=true`.

New MCP tools:
- generate_adaptation_project — scaffold a new ADP against a connected ABAP system; optionally imports key user changes from LREP
- adp_controller_extension — generate controller extensions and XML fragments from a prompt; validates paths, filters .change files, scans existing project context
- open_adaptation_editor — start the adaptation editor server for a project
- run_rta_workflow_step — dispatched step protocol over the adaptation editor (start → get_overlays → get_context → call_action → save → stop/restart)
- read_odata_metadata_adp — fetch and optionally persist OData metadata for an ADP from the connected ABAP system
- lookup_ui5_documentation — resolve aggregation, property, and event docs from a library's api.json, walking the inheritance chain and targeting the app's pinned UI5 version

Key changes:
- Adds MCP tools for ADP project generation, adaptation editor launch, controller extension authoring, RTA workflow automation, ADP OData metadata reading, and UI5 documentation lookup.
- Introduces a Playwright-backed RTA bridge with browser/page lifecycle handling, restart/stop support, frame targeting, and frontend-action dispatch.
- Adds progressive-disclosure controller extension support that returns ADP-specific generation rules/project context first, then safely writes AI-generated files with path traversal protection and `.change` file skipping.
- Adds UI5 `api.json` documentation lookup for aggregations, properties, and events, including inheritance resolution and caching.
- Consolidates shared ADP utilities into `@sap-ux/adp-tooling`, including default project name generation and DEFAULT key-user-change fetching.
- Updates generator-adp to consume the shared default project name utility.
- Adds quote-safe command execution via `runCmdArgs` for generator invocation.
- Adds package dependencies, TS references, changesets, Jest mocks, and broad unit test coverage.

## Type of change
- [ ] Bug (non-breaking change which fixes an issue)
- [x] Non-Breaking chores (Changes to tools, libraries, build process, documentation, etc)

## How have you tested?
Added and updated unit tests covering:
- ADP controller extension parsing, writing, scanning, lifecycle, and error handling
- Adaptation project generation, argv safety, optional key-user-change import, timeouts, and failures
- Adaptation editor launch behavior
- RTA dispatcher, command wrappers, and Playwright bridge lifecycle
- UI5 documentation lookup, cache, inheritance, and config resolution
- ADP metadata reading and shared utility behavior

## Checklist:
- [x] I have reviewed and addressed all Hyperspace bot findings (or explicitly explained dismissals)
- [x] I have done an Agentic review

<details>

**Version:** `1.31.37`

- LLM: `gpt-5.5`
- Correlation ID: `84d819b0-b0da-11f1-844b-801ac8eb824c`
</details>


packages/fiori-mcp-server/src/tools/get-adp-odata-metada.ts
```diff
@@ -0,0 +1,12 @@
...
+import { readAnnotationfromManifest } from './functionalities/manifest-changes/manifestContext.js';
+/**
+ * Reads the OData metadata for the specified ADP (Application Development Platform) project.
+ *
+ * @param params - Input parameters for reading OData metadata.
+ * @returns A promise resolving to an array of OData metadata.
+ */
+export async function readODataMetadataAdp(params: AdpMetadataInput): Promise<Array<any>> {
+    const metadataMap = await readAnnotationfromManifest(params.appPath, params.saveLocal);
+    return Array.from(metadataMap.values());
```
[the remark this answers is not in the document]
nikmace: Superseded — this file was replaced by `read-odata-metadata.ts` which builds a proper Map and accesses it via .get(name). The `Array.from(metadataMap.values())` call you flagged doesn't exist in the current implementation. No action needed.

packages/fiori-mcp-server/src/tools/get-adp-odata-metada.ts
```diff
@@ -0,0 +1,12 @@
+import type { AdpMetadataInput } from '../types/index.js';
+import { readAnnotationfromManifest } from './functionalities/manifest-changes/manifestContext.js';
+/**
+ * Reads the OData metadata for the specified ADP (Application Development Platform) project.
+ *
+ * @param params - Input parameters for reading OData metadata.
+ * @returns A promise resolving to an array of OData metadata.
+ */
+export async function readODataMetadataAdp(params: AdpMetadataInput): Promise<Array<any>> {
```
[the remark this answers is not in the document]
nikmace: Superseded — the file was renamed to read-odata-metadata.ts before this PR was opened. The typo doesn't exist in the current diff.

packages/fiori-mcp-server/src/tools/generate-adaptation-project.ts
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        await FSpromises.mkdir(finalTargetFolder, { recursive: true });
+
+        // Pass the JSON payload as a single argv element (not interpolated into a shell string) so
+        // quotes, spaces or apostrophes in values cannot corrupt it. A corrupted payload would make
+        // the generator silently fall back to interactive prompts and hang with no attached stdin.
+        const jsonString = JSON.stringify(jsonInput);
+        const { stdout, stderr } = await runCmdArgs('npx', ['-y', 'yo@4', '@sap-ux/adp', jsonString, '--force'], {
+            cwd: process.cwd(),
```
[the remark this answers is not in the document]
nikmace: Already fixed — the current implementation passes cwd: finalTargetFolder to runCmdArgs. The suggestion matches what's already in the code.

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/index.ts
```diff
@@ -0,0 +1,167 @@
...
+                return { saved: ok };
+            }
+            case 'stop': {
+                const { id } = getSession(input.sessionId);
+                sessions.delete(id);
+                if (sessions.size === 0) {
+                    await defaultTransport.stopBrowser();
+                }
+                return { stopped: true };
+            }
```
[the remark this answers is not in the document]
nikmace: Already handled — the current stop case calls await defaultTransport.disconnectSite(site) first (which closes the page and removes it from the registry), then checks isRegistryEmpty() before calling stopBrowser(). The pattern mirrors restart. No action needed.

packages/fiori-mcp-server/src/tools/adp-controller-extension/ai-response/parser.ts
```diff
@@ -0,0 +1,62 @@
...
+    let currentPath = '';
+    let inCodeBlock = false;
+    let currentCode = '';
+
+    for (const line of lines) {
+        const pathMatch = line.match(PATH_MARKER);
+        if (pathMatch) {
+            currentPath = pathMatch[1].trim();
+            continue;
+        }
```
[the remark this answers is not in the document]
nikmace: Changed in 3747bfd

packages/fiori-mcp-server/src/server.ts
```diff
@@ -273,6 +292,27 @@ Never skip steps or guess functionalityIds. Never use a functionalityId as a too
                     case 'generate_fiori_app_cap':
                         result = await generateFioriAppCap(args as GeneratorConfigCAP);
                         break;
+                    case 'generate_adaptation_project':
+                        result = await generateAdaptationProject(args as GenerateAdaptationProjectInput);
+                        break;
+                    case 'open_adaptation_editor':
```
devinea, line 306: test comment

devinea (commented):
test

packages/fiori-mcp-server/src/utils/utils.ts
```diff
@@ -21,3 +21,102 @@ export async function checkIfGeneratorInstalled(generatorVersion = '1.18.5'): Pr
...
+
+        child.stdout?.setEncoding('utf8');
+        child.stderr?.setEncoding('utf8');
+        child.stdout?.on('data', (data: string) => {
+            stdout += data;
+        });
+        child.stderr?.on('data', (data: string) => {
+            stderr += data;
+        });
+
```
devinea, line 111: test

devinea (commented):
test

packages/fiori-mcp-server/src/utils/utils.ts
```diff
@@ -21,3 +21,102 @@ export async function checkIfGeneratorInstalled(generatorVersion = '1.18.5'): Pr
...
+ * @throws {Error} If the process cannot be spawned, exits with a non-zero code, or exceeds `timeout`.
+ */
+export function runCmdArgs(cmd: string, args: string[], options: RunCmdArgsOptions = {}): Promise<RunCmdArgsResult> {
+    const { cwd, timeout } = options;
+    // On Windows, resolve .cmd shims (e.g. npx → npx.cmd) so we can spawn
+    // without shell: true — shell: true corrupts args that contain quotes.
+    const resolvedCmd = process.platform === 'win32' && !cmd.includes('.') ? `${cmd}.cmd` : cmd;
+    const spawnOptions: SpawnOptions = { cwd };
+
+    return new Promise<RunCmdArgsResult>((resolve, reject) => {
```
devinea, line 61: test

devinea (commented):
test

packages/fiori-mcp-server/src/tools/adp-controller-extension/ai-response/processor.ts
```diff
@@ -0,0 +1,79 @@
...
+    logger.info('Processing AI response to extract files');
+    const extractedFiles = extractFilesFromResponse(aiResponse);
+    logger.info(`Extracted ${extractedFiles.length} files from AI response`);
+
+    const changes: string[] = [];
+    const echo = echoableParams(params);
+
+    for (const file of extractedFiles) {
+        if (isChangeFile(file.path)) {
+            logger.debug(`Skipping change file: ${file.path}`);
```
devinea, line 30: test

devinea (commented):
test

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/browser/playwright-bridge.ts
```diff
@@ -0,0 +1,300 @@
...
+}
+
+/**
+ * Invokes a Joule frontend action on the page hosting `site`. Reuses an
+ * existing page if one is registered for the site; otherwise opens a new
+ * one. Throws on transport / lookup failures; returns the frontend action
+ * result envelope (including `isSuccess === false` payloads) for the caller to inspect.
+ *
+ * @param site URL to load (typically the adaptation editor URL).
+ * @param actionName Frontend action identifier, e.g. `com.sap.ui.flex.startRTA.v1`.
```
devinea, line 188: test

devinea (commented):
test

devinea (commented):
test

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/index.ts
```diff
@@ -0,0 +1,167 @@
...
+    try {
+        switch (input.step) {
+            case 'start': {
+                const site = requireString(input.payload, 'site');
+                const frameId = typeof input.payload?.frameId === 'string' ? input.payload.frameId : undefined;
+                const result = await startRta(defaultTransport, { site, frameId });
+                const sessionId = randomUUID();
+                sessions.set(sessionId, { site, frameId });
+                return { sessionId, ...result };
+            }
```
devinea, line 97: test

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/browser/playwright-bridge.ts
```diff
@@ -0,0 +1,300 @@
...
+ * 2. `PLAYWRIGHT_BROWSER_CHANNEL` — Playwright channel (`chrome`, `msedge`, etc.).
+ * 3. Default: `chrome` channel (system-installed Google Chrome).
+ * 4. Fallback: Playwright's bundled Chromium (requires `npx playwright install chromium`).
+ *
+ * @returns The connected Chromium browser instance.
+ */
+async function startBrowser(): Promise<Browser> {
+    if (browser?.isConnected()) {
+        return browser;
+    }
```
devinea, line 136: test

devinea (commented):
test

devinea (commented):
test

packages/fiori-mcp-server/src/tools/read-odata-metadata.ts
```diff
@@ -0,0 +1,108 @@
...
+ * @returns One entry per OData data source: id, service URL, formatted metadata XML, and bound model.
+ */
+export async function readODataMetadataAdp(params: ReadODataMetadataInput): Promise<ODataMetadataEntry[]> {
+    const { appPath, saveLocal = false } = params;
+    const abapProvider = await getProvider(appPath);
+    const variant = await getVariant(appPath);
+    const manifestService = await ManifestService.initMergedManifest(abapProvider, appPath, variant, logger);
+
+    const manifest = manifestService.getManifest();
+    const ui5Models = (manifest['sap.ui5']?.models ?? {}) as Record<string, Ui5Model>;
```
devinea, line 53: test

packages/fiori-mcp-server/src/utils/utils.ts
```diff
@@ -21,3 +21,113 @@ export async function checkIfGeneratorInstalled(generatorVersion = '1.18.5'): Pr
...
+                }, 5000);
+                // If the child exits before the force-kill timer fires, cancel it.
+                child.once('close', () => clearTimeout(forceKillTimer));
+                settle(() =>
+                    reject(
+                        new Error(`Command '${cmd} ${args.join(' ')}' timed out after ${timeout}ms and was terminated.`)
+                    )
+                );
+            }, timeout);
+        }
```
devinea, line 101: **[Security] Credential leak in error messages**

`args.join(' ')` embeds the full argument vector—including any JSON payload passed to the generator (which may contain a password or auth token)—directly into the timeout and exit-code error messages. These strings are returned to MCP clients and may land in logs.

Replace the detail string with a redacted form, e.g. only include `cmd` and the argument count:
```ts
new Error(`Command '${cmd}' timed out after ${timeout}ms (${args.length} args)`)
```

packages/fiori-mcp-server/src/utils/utils.ts
```diff
@@ -21,3 +21,113 @@ export async function checkIfGeneratorInstalled(generatorVersion = '1.18.5'): Pr
...
+ * @throws {Error} If the process cannot be spawned, exits with a non-zero code, or exceeds `timeout`.
+ */
+export function runCmdArgs(cmd: string, args: string[], options: RunCmdArgsOptions = {}): Promise<RunCmdArgsResult> {
+    const { cwd, timeout } = options;
+    // On Windows, resolve .cmd shims (e.g. npx → npx.cmd) so we can spawn
+    // without shell: true — shell: true corrupts args that contain quotes.
+    const resolvedCmd = process.platform === 'win32' && !cmd.includes('.') ? `${cmd}.cmd` : cmd;
+    const spawnOptions: SpawnOptions = { cwd };
+
+    return new Promise<RunCmdArgsResult>((resolve, reject) => {
```
devinea, line 61: **[Bug] Windows: .cmd shims still fail with ENOENT without `shell: true`**

Appending `.cmd` to the command name is not enough—Windows cannot execute `.cmd` files directly via `spawn` unless `shell: true` is set (they are not PE binaries; they require `cmd.exe` to interpret them). The current code will still throw `ENOENT` for `npx.cmd` and every other `.cmd` shim on Windows.

Set `shell: true` on Windows while keeping it `false` on POSIX:
```ts
const spawnOptions: SpawnOptions = { cwd, shell: process.platform === 'win32' };
```
You can then drop the `.cmd` suffix logic—`shell: true` on Windows resolves shims automatically via `PATH`.

packages/fiori-mcp-server/src/server.ts
```diff
@@ -273,6 +295,24 @@ Never skip steps or guess functionalityIds. Never use a functionalityId as a too
                     case 'generate_fiori_app_cap':
                         result = await generateFioriAppCap(args as GeneratorConfigCAP);
                         break;
+                    case 'generate_adaptation_project':
```
devinea, line 303: **[Security] `SAP_FIORI_MCP_ADP_TOOLS=false` guard is bypassable**

The env-var check only controls whether ADP tools appear in the `list_tools` response. The `call_tool` switch unconditionally dispatches to `generateAdaptationProject`, `openAdaptationEditor`, `adpControllerExtension`, `runRtaWorkflowStep`, and `readODataMetadataAdp`. A caller that already knows the tool names can invoke them directly, bypassing the flag entirely.

Add an explicit guard in each ADP case (or before the switch):
```ts
if (process.env.SAP_FIORI_MCP_ADP_TOOLS === 'false') {
    throw new McpError(ErrorCode.MethodNotFound, `Tool ${name} is disabled`);
}
```

packages/fiori-mcp-server/src/tools/read-odata-metadata.ts
devinea, line 53: **[Bug] Structured output schema mismatch — `structuredContent` is always skipped**

The tool's `outputSchema` declares `{ entries: ODataMetadataEntry[] }`, but `readODataMetadataAdp` returns a plain `ODataMetadataEntry[]` (an array, not an object). The server's `convertResultToCallToolResult` logic that populates `structuredContent` will never match, so MCP clients that rely on structured output always receive only the text fallback.

Either wrap the return value:
```ts
return { entries: await readODataMetadataAdp(params) };
```
or change the `outputSchema` to `z.array(oDataMetadataEntrySchema)` and adjust any downstream consumers.

packages/fiori-mcp-server/src/tools/adp-controller-extension/ai-response/processor.ts
```diff
@@ -0,0 +1,65 @@
...
+        try {
+            const relativePath = writeExtractedFile(appPath, file);
+            changes.push(`Created ${relativePath}`);
+        } catch (error) {
+            if (error instanceof PathTraversalError) {
+                logger.error(error.message);
+                return { status: 'Error', message: error.message, appPath, changes: [] };
+            }
+            const message = `Failed to write file ${file.path}: ${
+                error instanceof Error ? error.message : String(error)
```
devinea, line 43: **[Bug] PathTraversalError returns `changes: []` after earlier files were already written**

When a `PathTraversalError` is thrown mid-loop, the handler returns `changes: []` even though the loop may have successfully written several files to disk in earlier iterations. The caller receives an empty list and may believe it is safe to retry the whole batch—leading to duplicate writes for the already-persisted files.

Return the actual partial writes so callers can see what was already committed:
```ts
return { status: 'Error', message: error.message, appPath, changes };
```

devinea (commented):
Security and correctness review findings — 8 issues identified.

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/browser/playwright-bridge.ts
```diff
@@ -0,0 +1,310 @@
...
+}
+
+/**
+ * Invokes a Joule frontend action on the page hosting `site`. Reuses an
+ * existing page if one is registered for the site; otherwise opens a new
+ * one. Throws on transport / lookup failures; returns the frontend action
+ * result envelope (including `isSuccess === false` payloads) for the caller to inspect.
+ *
+ * @param site URL to load (typically the adaptation editor URL).
+ * @param actionName Frontend action identifier, e.g. `com.sap.ui.flex.startRTA.v1`.
```
devinea, line 188: **[Resource leak] Browser context/page not closed when `page.goto()` or `createPageRPC()` throws**

If `page.goto()` times out or `createPageRPC()` throws, the newly-created `context` and `page` are abandoned—neither is closed, and the page is never added to `connectionRegistry`, so `stopBrowser()` cannot clean it up either.

Wrap the setup in a try/catch (or try/finally) and close the context on failure:
```ts
const context = await activeBrowser.newContext({ viewport: null });
const page: Page = await context.newPage();
try {
    // ... goto, createPageRPC ...
    connectionRegistry.set(site, rpc);
} catch (err) {
    await context.close().catch(() => undefined);
    throw err;
}
```

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/index.ts
```diff
@@ -0,0 +1,175 @@
...
+                // This prevents a new project opened on the same port from reusing the
+                // previous project's stale Playwright page.
+                await defaultTransport.disconnectSite(site);
+                const result = await startRta(defaultTransport, { site, frameId });
+                return { site, frameId, ...result };
+            }
+            case 'get_overlays': {
+                const site = requireSite(input.site);
+                const { overlays, actionsCatalog } = await getOverlays(defaultTransport, {
+                    site,
```
devinea, line 100: **[API contract] Top-level `site` is marked required for all steps but only used after `start`**

The Zod input schema marks top-level `site` as required, yet the `start` step reads `site` from `payload.site` (not from `input.site`). LLM callers following the schema will pass `site` at the top level for `start`, get an error, then discover the correct location only through trial and error.

Mark top-level `site` as optional (it is only needed for steps after `start`) and add a clear discriminated-union or `superRefine` that requires `payload.site` when `step === 'start'` and requires `input.site` otherwise.

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/browser/playwright-bridge.ts
```diff
@@ -0,0 +1,310 @@
...
+ * 2. `PLAYWRIGHT_BROWSER_CHANNEL` — Playwright channel (`chrome`, `msedge`, etc.).
+ * 3. Default: `chrome` channel (system-installed Google Chrome).
+ * 4. Fallback: Playwright's bundled Chromium (requires `npx playwright install chromium`).
+ *
+ * @returns The connected Chromium browser instance.
+ */
+async function startBrowser(): Promise<Browser> {
+    if (browser?.isConnected()) {
+        return browser;
+    }
```
devinea, line 136: **[Fragility] `isMissingBrowserError` uses hardcoded English substring matching**

The function matches against five hardcoded lowercase English phrases. Playwright error messages are version-specific and may change between releases; they also vary by locale when Playwright surfaces OS error strings. A mismatch means the fallback Chromium launch is silently skipped and the original error propagates without the `INSTALL_HINT`.

Consider testing a structured property instead of free-text: Playwright's `BrowserType.launch` throws errors with a `name` of `'Error'` but often sets a machine-readable `code` or wraps an OS `errno`. Alternatively, catch *all* launch errors and always attempt the bundled fallback, only re-throwing if the fallback also fails—this makes the function unnecessary.

packages/fiori-mcp-server/src/utils/utils.ts
nikmace: Committed as 02870df

packages/fiori-mcp-server/src/utils/utils.ts
nikmace: The `.cmd` suffix approach is intentional here. Setting `shell: true` on Windows causes Node to pass the full command line through `cmd.exe`, which re-parses all arguments — any JSON payload in args containing &, |, ^, or quotes becomes a shell injection vector. The whole point of `runCmdArgs` (vs the existing runCmd) is to keep args verbatim and never expose them to shell parsing.

Node can spawn `.cmd` shims directly with `shell: false` when the extension is explicit — Windows resolves the interpreter via PATHEXT/CreateProcess without re-parsing argv.

(The automated security scanner flagged shell: true as a command injection risk for exactly this reason when we tried the change.)

Happy to switch to `shell: true` with per-arg `cmd.exe` quoting if you'd prefer, but the current approach is the safer default. Let us know.

packages/fiori-mcp-server/src/server.ts
nikmace: The risk is low in practice, but here is the change [67a18e8](https://github.com/SAP/open-ux-tools/pull/5107/commits/67a18e8cf59d10c0c8775096e7c95a7a87b6ffe5)

nikmace: Used !== 'true' rather than === 'false' so the guard fires when the env var is unset as well — with === 'false' a caller with no env var set at all would still reach the ADP tools.

packages/fiori-mcp-server/src/tools/read-odata-metadata.ts
nikmace: [2fb826a](https://github.com/SAP/open-ux-tools/pull/5107/commits/2fb826aab72ffb0f80d583e9189cc4d40b5d6846)

packages/fiori-mcp-server/src/tools/adp-controller-extension/ai-response/processor.ts
nikmace: Changed in [aec272b](https://github.com/SAP/open-ux-tools/pull/5107/commits/aec272b3178f40c54bbede7cf940eefd8bac1f82)

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/browser/playwright-bridge.ts
nikmace: Changed in [28d6f0b](https://github.com/SAP/open-ux-tools/pull/5107/commits/28d6f0b67b8d0f509286db755b5b553cbbede2f8)

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/index.ts
nikmace: Changed in: [350ad11](https://github.com/SAP/open-ux-tools/pull/5107/commits/350ad110f5eb4d1596b6f93b5f47805591a841ea)

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/browser/playwright-bridge.ts
nikmace: Changed in: [45a4d86](https://github.com/SAP/open-ux-tools/pull/5107/commits/45a4d86390b761243502cd7897be05da824aa0aa)

IainSAP (commented):
Code review findings — 4 confirmed bugs, all correctness issues.

packages/fiori-mcp-server/src/utils/utils.ts
```diff
@@ -21,3 +21,117 @@ export async function checkIfGeneratorInstalled(generatorVersion = '1.18.5'): Pr
...
+ * Unlike {@link runCmd} (which wraps `exec` and interpolates the command into a shell), this
+ * passes each argument as a distinct argv element. That means arbitrary argument values —
+ * such as a JSON payload containing quotes, spaces or apostrophes — are never re-parsed by a
+ * shell and cannot corrupt the command. Output is streamed, so there is no `maxBuffer` limit,
+ * and an optional `timeout` guards against a child that hangs (e.g. a generator that silently
+ * dropped into an interactive prompt with no attached stdin).
+ *
+ * On Windows, `.cmd`/`.bat` shims (like `npx`) are not directly executable, so the command is
+ * run through the shell there; each argument is still passed as a separate argv element.
+ *
```
IainSAP, line 51: **Bug: `.cmd` shims spawned without `shell: true` — ENOENT on every Windows invocation**

On Windows, `runCmdArgs` appends `.cmd` to produce e.g. `yo.cmd`, then calls `spawn("yo.cmd", args, { cwd })` with no `shell` option. `CreateProcess` cannot execute `.cmd` files directly; the child process emits `error` with ENOENT, causing `generate_adaptation_project` to fail on every Windows call.

Fix: add `shell: process.platform === "win32"` (or unconditionally `shell: true`) to `spawnOptions` when a `.cmd` suffix is appended.

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/index.ts
```diff
@@ -0,0 +1,180 @@
...
+    if (!site.startsWith('http://') && !site.startsWith('https://')) {
+        throw new Error(`site must be an http:// or https:// URL. Received: "${site}"`);
+    }
+    return site;
+}
+
+/**
+ * Runs one step of the RTA workflow. Dispatches by `input.step` and forwards
+ * to the corresponding `rta` command wrapper. `site` and `frameId` are passed
+ * directly on every call — the tool holds no server-side session state.
```
IainSAP, line 78: **Bug: `start` step reads `site` from `payload.site` but the Zod schema places it at the top level**

The `start` case calls `requireString(input.payload, "site")`, but `payload.site` is always `undefined` — the schema declares `site` as a required top-level field and all other steps read `input.site`. Every call to the `start` step throws `"payload.site is required"`, making the RTA workflow impossible to start.

Fix: change to `requireString(input, "site")` (matching the other step cases).

packages/fiori-mcp-server/src/tools/open-adaptation-editor.ts
```diff
@@ -0,0 +1,228 @@
...
+export async function openAdaptationEditor(params: OpenAdaptationEditorInput): Promise<OpenAdaptationEditorOutput> {
+    const { appPath } = params;
+
+    try {
+        const isWindows = process.platform === 'win32';
+        const { command, args } = resolveFioriBin(appPath, isWindows);
+
+        logger.info(`Spawning editor process: ${command} ${args.join(' ')} in ${appPath}`);
+
+        const childProcess: ChildProcess = spawn(command, args, {
```
IainSAP: **Bug: `.bin/fiori.cmd` and `npm.cmd` spawned without `shell: true` on Windows**

`resolveFioriBin` returns `<appPath>/node_modules/.bin/fiori.cmd` (or `npm.cmd` as fallback) on Windows. `spawn(command, args, { shell: false })` uses `CreateProcess` which cannot run `.cmd` files without `cmd.exe`. The child never starts, no URL is emitted on stdout, and `waitForEditorUrl` times out after 30 s — the adaptation editor is unreachable on Windows.

Fix: pass `shell: process.platform === "win32"` in the spawn options when a `.cmd` binary is used.

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/browser/playwright-bridge.ts
```diff
@@ -0,0 +1,285 @@
...
+        });
+
+        page.on('close', () => {
+            connectionRegistry.delete(site);
+        });
+
+        try {
+            await page.goto(site, { waitUntil: 'networkidle', timeout: 60000 });
+            rpc = await createPageRPC(page);
+            connectionRegistry.set(site, rpc);
```
IainSAP, line 217: **Bug: `BrowserContext` created per-site is never closed, leaking contexts for the process lifetime**

Each `callFrontendAction` for a new site calls `activeBrowser.newContext()`, but `rpc.close()` (called by `disconnectSite`) only calls `page.close()` — the `BrowserContext` is never `context.close()`d. After N stop+start cycles, N contexts accumulate inside the browser process, each retaining cookies, storage, and network state. Over a long-running MCP server session the browser's memory grows without bound.

Fix: store the `BrowserContext` alongside the page in the connection registry and call `context.close()` in `rpc.close()`.

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/index.ts
nikmace: Thanks for the report, Iain! This was already caught and fixed in commit 350ad110f5 ("fix(fiori-mcp-server): make start step read site from input.site, consistent with all other steps") earlier on this branch. The current code reads site via requireSite(input.site) before the switch statement, so all steps including start use the top-level field consistently. No further action needed.

packages/fiori-mcp-server/src/utils/utils.ts
nikmace: Fixed in [cf0fe36](https://github.com/SAP/open-ux-tools/pull/5107/commits/cf0fe36a7c9376280341998035a34392e2d3c09a). We switched to cross-spawn instead of enabling shell: true. The reason we want to avoid shell: true is that it routes arguments through cmd.exe, which re-parses them — meaning any user-supplied value containing metacharacters like &, |, ^, <, > could inject additional commands. cross-spawn handles .cmd shim execution on Windows without a shell, so arguments are never re-parsed. It's already used in the monorepo (app-config-writer).

packages/fiori-mcp-server/src/tools/open-adaptation-editor.ts
nikmace: Fixed in the same [cf0fe36](https://github.com/SAP/open-ux-tools/pull/5107/commits/cf0fe36a7c9376280341998035a34392e2d3c09a), same reasoning. resolveFioriBin no longer produces .cmd-suffixed paths — cross-spawn is used at the spawn call site and handles the platform difference transparently while keeping shell: false semantics.

packages/fiori-mcp-server/src/utils/utils.ts
nikmace: Fixed in [cf0fe36](https://github.com/SAP/open-ux-tools/pull/5107/commits/cf0fe36a7c9376280341998035a34392e2d3c09a). We switched to cross-spawn instead of enabling shell: true. The reason we want to avoid shell: true is that it routes arguments through cmd.exe, which re-parses them — meaning any user-supplied value containing metacharacters like &, |, ^, <, > could inject additional commands. cross-spawn handles .cmd shim execution on Windows without a shell, so arguments are never re-parsed. It's already used in the monorepo (app-config-writer).

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/browser/playwright-bridge.ts
nikmace: Fixed. `createPageRPC` now accepts the `BrowserContext` as a second parameter and the close() closure calls both `page.close()` and `context.close()` — matching the error-path behaviour already present in the goto failure handler. No behaviour change across a running session: the context is created once per site on first `callFrontendAction` and stays open for all subsequent calls to that site; it only closes when `disconnectSite` (the stop step) or `stopBrowser` is explicitly called. [11a3da6](https://github.com/SAP/open-ux-tools/pull/5107/commits/11a3da66d55e73655aec8b2c7780f6ca31070b7f)

packages/fiori-mcp-server/package.json
```diff
@@ -83,11 +83,13 @@
         "@sap-ux/vocabularies-types": "0.16.0",
         "@sap/ux-specification": "1.144.5",
         "@types/json-schema": "7.0.15",
+        "@types/cross-spawn": "6.0.6",
         "@types/mem-fs": "1.1.2",
         "@types/mem-fs-editor": "7.0.1",
         "@sap-ux/adp-tooling": "workspace:*",
         "prettify-xml": "1.2.0",
         "cross-env": "10.1.0",
+        "cross-spawn": "7.0.6",
```
devinea: https://www.npmjs.com/package/cross-spawn
hasn't been released in 2 years 🤔 . Probably ok,  but have you considered alternatives that are more regaularly maintained:
https://www.npmjs.com/package/tinyexec
https://www.npmjs.com/package/nano-spawn

devinea (commented):
Thanks for applyng updates. 1 small query.

packages/fiori-mcp-server/package.json
nikmace: Changed to nano-spawn: [f1167e0](https://github.com/SAP/open-ux-tools/pull/5107/commits/f1167e0ceb22d4a389a14f0c24568f44e1858815)

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/index.ts
```diff
@@ -0,0 +1,179 @@
...
+    try {
+        url = new URL(site);
+    } catch {
+        throw new Error(`site must be a valid URL. Received: "${site}"`);
+    }
+    if (url.protocol !== 'http:' && url.protocol !== 'https:') {
+        throw new Error(`site must be an http:// or https:// URL. Received: "${site}"`);
+    }
+    if (url.hostname !== 'localhost') {
+        throw new Error(`site must point to localhost. Received: "${url.hostname}"`);
```
IvoSG: `requireSite` rejects any host that is not exactly `localhost`. Is `127.0.0.1` (which the editor CLI may emit) intentionally rejected? 
Worth confirming the editor always emits `localhost` (see the assumption stated at line 50).

IvoSG (commented):
Overall, things look good, except for one minor comment.

packages/fiori-mcp-server/src/tools/run-rta-workflow-step/index.ts
nikmace: Changed in: [71f46e3](https://github.com/SAP/open-ux-tools/pull/5107/commits/71f46e3da83162ea85e18d0480da55e20f68950c)

IvoSG (dismissed):
Review comments addressed
Test coverage is good
Changeset ok
Manually tested

PR #5142, opened 2026-09-01
Chore - update .github/CODEOWNERS based on feedback

devinea:
Removing @SAP/ux-tools-architects where it is 2nd reviewer as it causes noise on the PRs.


heimwege (approved):
removing @SAP/ux-tools-architects is fine for me as they were supposed to be optional but that concept sadly is not covered by CODEOWNERS

PR #5153, opened 2026-09-03
fix: prevent YUI webview crash on service selection step

### LINKED ISSUE #5152: VS Code Yeoman-UI webview crashes on ABAP service selection (both on-prem and BTP), CLI works fine with identical inputs
**Environment:** VS Code 1.136.0, SAP Fiori Tools Extension Pack, sapos.yeoman-ui 1.26.1, @sap/generator-fiori 1.31.1, Node 24.20.0, Windows

**Steps to reproduce:**

Fiori: Open Application Generator → List Report Page → Connect to a System → ABAP On Premise (also reproduces with BTP ABAP Environment/Cloud Foundry discovery)
Enter system URL, client, credentials → Login
Crash immediately on service selection step, before any service is chosen

**Error:**
Could not update method 'validate' in 'systemSelection:serviceSelection' / 'abapOnPrem:serviceSelection' question in generator '@sap/fiori:app' - evaluateMethod() Cannot read properties of null (reading 'value')
at odata-service-inquirer/index.js:27:48249

**Contrast test:** Running npx yo @sap/fiori from the terminal with identical URL/client/credentials succeeds — catalog fetches 248 service groups, service list is fully browsable. Same generator version, only the invocation path differs (CLI vs VS Code Yeoman-UI webview).

longieirl:
## Summary

Fixes #5152 — VS Code Yeoman-UI webview crashes immediately on service selection step (on-prem and BTP), while CLI works fine.

**Root cause:**
- `isCli()` in `fiori-generator-shared` checks `process.stdin.isTTY` to detect CLI context. On Windows with Node 24.x, VS Code's extension host has `process.stdin.isTTY === true`, so `isCli()` incorrectly returns `true` inside VS Code.
- This causes `useAutoComplete: true` for the `serviceSelection` prompt (should be `false` in YUI).
- With `useAutoComplete: true`, prompt type becomes `autocomplete`. YUI calls `validate(null)` on first render before any selection, hitting `(null as ListChoiceOptions).value` → `TypeError: Cannot read properties of null (reading 'value')`.

**Two fixes:**
1. **Defensive fix** (`odata-service-inquirer`): guard `service != null` before accessing `.value` in `validate` — prevents crash regardless of environment detection.
2. **Root cause fix** (`fiori-generator-shared`): `isCli()` now checks `VSCODE_PID` env var and returns `false` early, preventing the `process.stdin.isTTY` false positive.

**Reference:** `VSCODE_PID` is set by VS Code's electron main process and inherited by all child processes including the extension host:
https://github.com/microsoft/vscode/blob/main/src/vs/code/electron-main/main.ts#L477

## Test plan

- [ ] New test: `validate(null)` with `useAutoComplete: true` resolves without throwing
- [ ] New test: `isCli()` returns `false` when `VSCODE_PID` is set, even with `process.stdin.isTTY === true`
- [ ] Both package test suites pass
- [ ] Manual verification: open App Generator in VS Code on Windows, select ABAP On Premise, log in — service selection step should no longer crash

### DIFF
--- a/.changeset/fix-icli-vscode-pid.md
+++ b/.changeset/fix-icli-vscode-pid.md
@@ -1,0 +1,5 @@
+---
+"@sap-ux/fiori-generator-shared": patch
+---
+
+FIX: Prevent isCli() returning true inside VS Code extension host on Windows/Node 24.x by checking VSCODE_PID environment variable
--- a/.changeset/fix-service-selection-null-crash.md
+++ b/.changeset/fix-service-selection-null-crash.md
@@ -1,0 +1,5 @@
+---
+"@sap-ux/odata-service-inquirer": patch
+---
+
+FIX: Guard against null service in validate when useAutoComplete is true; prevents crash when YUI calls validate before any service is selected
--- a/packages/fiori-generator-shared/src/environment.ts
+++ b/packages/fiori-generator-shared/src/environment.ts
@@ -7,6 +7,11 @@ import { hostEnvironment, type HostEnvironmentId } from './types/index.js';
  * @returns true if it is a cli environment, false otherwise
  */
 export function isCli(): boolean {
+    // VSCODE_PID is set by VS Code in its extension host; process.stdin.isTTY can be true
+    // on Windows with Node 24.x even inside VS Code, so exclude this case from CLI detection
+    if (process.env.VSCODE_PID) {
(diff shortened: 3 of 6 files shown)


longieirl:
Not applicable, unable to reproduce issue.

PR #5156, opened 2026-09-04
feat(fiori-mcp-server): productize lookup_ui5_documentation as MCP tool

mmilko01:
## Summary

- Promotes the standalone `scripts/lookup-aggregation.mjs` into a proper `lookup_ui5_documentation` MCP tool available in the Fiori Tools MCP server
- The tool looks up UI5 control aggregation/property/event documentation from a library's designtime `api.json`, resolves the UI5 host from `ui5.yaml` (falling back to `https://ui5.sap.com`), and walks the inheritance chain for inherited members
- Removes the now-superseded `lookup-aggregation.mjs` script
- Updates `SKILL.md` with `lookup_ui5_documentation` usage rules for the ADP controller extension flow

## Changes

- `src/tools/lookup-ui5-documentation/` — full implementation (api-json fetcher/cache, aggregation/property/event lookups, inheritance chain walker)
- `src/tools/index.ts`, `src/server.ts`, `src/types/` — tool wiring
- `package.json` — adds `@sap-ux/ui5-config` dependency for `ui5.yaml` resolution
- `test/unit/tools/lookup-ui5-documentation/` — unit tests
- `skills/adp-controller-extension-flow/SKILL.md` — usage rules added

## Test plan

- [ ] Unit tests pass: `pnpm --filter @sap-ux/fiori-mcp-server test`
- [ ] MCP tool is listed when the server starts
- [ ] `lookup_ui5_documentation` correctly resolves aggregation/property/event from designtime api.json
- [ ] Inheritance chain walking returns members from parent controls
- [ ] ADP e2e harness (`ProductDeveloper` scenario) passes with the tool available — verified 2026-09-04, 100% pass (14/14 requirements)


packages/fiori-mcp-server/src/tools/lookup-ui5-documentation/lookups/aggregation.ts
```diff
@@ -0,0 +1,54 @@
...
+    params: LookupUi5DocumentationInput,
+    source: LookupSource
+): AggregationLookupResult {
+    const { library, control, member } = params;
+    if (!member) {
+        throw new Error('The "member" parameter is required when lookupType is "aggregation".');
+    }
+
+    const found = findMemberInChain(chain, selectAggregations, member);
+    if (!found) {
```
nikmace: `lookupAggregation`, `lookupEvent`, and `lookupProperty` all seem to use the same logic except for the UI5 metadata object.

Recommendation: extract the duplicated guard + search logic into a shared helper (e.g. resolveMember) that handles the member check, findMemberInChain call, and the "not found" error — but keep the three handlers as separate functions that call it. That gets you the DRY win without sacrificing type narrowing or testability. The result shaping stays per-type because TypeScript can't meaningfully narrow it otherwise.

packages/fiori-mcp-server/src/tools/lookup-ui5-documentation/lookups/aggregation.ts
```diff
@@ -0,0 +1,54 @@
+// Aggregation lookup: extracts a single control aggregation's documentation, walking the control's
+// inheritance chain so inherited aggregations (often declared in an ancestor library) resolve too.
+
```
nikmace: Maybe these top-level comments in a lot of the files are not really needed. What do you think?

packages/fiori-mcp-server/src/types/input.ts
```diff
@@ -206,3 +206,44 @@ export const AdpMetadataInputSchema = zod.object({
...
+            'Which kind of UI5 control member to look up: "aggregation" (type, cardinality, visibility, ' +
+                'since, description), "property" (type, defaultValue, group, bindable, visibility, since, ' +
+                'description), or "event" (parameters, visibility, since, description). Inherited members are ' +
+                "resolved by walking the control's inheritance chain across libraries."
+        ),
+    library: zod.string().describe('The UI5 library the control belongs to (e.g. "sap.ui.comp").'),
+    control: zod.string().describe('Fully-qualified control name (e.g. "sap.ui.comp.smarttable.SmartTable").'),
+    member: zod
+        .string()
+        .optional()
```
nikmace: `member` is optional in the Zod schema but required at runtime — schema mismatch.

packages/fiori-mcp-server/src/tools/lookup-ui5-documentation/api-json.ts
```diff
@@ -0,0 +1,381 @@
...
+ * @returns Candidate dotted library names, most-specific first.
+ */
+function candidateLibraries(fqName: string): string[] {
+    const prefixSegments = fqName.split('.').slice(0, -1);
+    const candidates: string[] = [];
+    for (let len = prefixSegments.length; len >= 2; len--) {
+        candidates.push(prefixSegments.slice(0, len).join('.'));
+    }
+    return candidates.slice(0, 4);
+}
```
nikmace, line 67: Optional / Needs to be checked: 
`candidateLibraries` probing is `O(n)` network calls with no short-circuit on a definitive 404 diff-only

`candidateLibraries("sap.ui.comp.smarttable.SmartTable")` yields `["sap.ui.comp.smarttable", "sap.ui.comp", "sap.ui"]`. For a valid class, it will always fail on the first candidate (sub-package) before succeeding on the second. Every miss fires a full resolveApiJson waterfall (up to 4 HTTP attempts). For a control deep in a hierarchy (e.g. `sap.m.InputBase` → `sap.ui.core.Control`), the chain walk could issue a dozen network calls. The disk cache (24 h) prevents repeat fetches across requests, but on a cold start this is significant latency. No concern for correctness; worth flagging for performance in production use.

packages/fiori-mcp-server/src/tools/lookup-ui5-documentation/api-json.ts
```diff
@@ -0,0 +1,381 @@
+// lookupType-agnostic core for the `lookup_ui5_documentation` tool.
+//
+// Responsible for everything that is independent of *which* piece of documentation is requested:
+// discovering and reading ui5.yaml (to resolve the configured UI5 base URL and version), fetching and
+// caching a library's designtime api.json, and locating a control symbol within it. The per-lookupType
+// extractors (see ./lookups/*) operate on the Ui5Symbol this module returns.
+//
+// Raw fetch + ~/.cache TTL are deliberate — public anonymous GETs with no suitable common-lib equivalent.
+
```
nikmace: `api-json.ts` responsibilities — the file mixes 4+ concerns (yaml discovery, cache I/O, HTTP fetch, chain walking) under a name that describes only one of them. The file owns ui5.yaml discovery and parsing (findUi5Yaml, extractUi5Config, resolveUi5Config), manifest.json version fallback (resolveMinUi5VersionFromManifest), disk cache read/write (readCache, writeCache, cachePathFor), in-memory class→library memo (classLibraryCache, resolveLibraryForClass), and inheritance chain walking (resolveControlChain).  

Suggest splitting into at least: ui5-config-resolver.ts (yaml discovery + version resolution), api-json-cache.ts (disk cache + HTTP fetch + fallback chain), and keeping api-json.ts for findControl, resolveLibraryForClass, resolveControlChain. This would also make the caching policy independently testable.

packages/fiori-mcp-server/src/tools/lookup-ui5-documentation/api-json.ts
```diff
@@ -0,0 +1,381 @@
...
+ */
+export function findControl(apiJson: ApiJson, controlName: string): Ui5Symbol | null {
+    if (!Array.isArray(apiJson.symbols)) {
+        return null;
+    }
+    return apiJson.symbols.find((s) => s.name === controlName) ?? null;
+}
+
+/** In-memory memo for class→api.json resolution, keyed by base|version|fqName. Process-lifetime. */
+const classLibraryCache = new Map<string, ResolveApiJsonResult | null>();
```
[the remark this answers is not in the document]
nikmace: This needs to be checked also 🟡

packages/fiori-mcp-server/src/tools/lookup-ui5-documentation/api-json.ts
```diff
@@ -0,0 +1,381 @@
...
+import type { ApiJson, FetchApiJsonResult, ResolveApiJsonResult, Ui5Symbol } from './types.js';
+
+/** Resolved (url, version) pair read from a ui5.yaml `ui5:` mapping. */
+interface Ui5YamlConfig {
+    url: string | null;
+    version: string | null;
+}
+
+const FALLBACK_BASE = 'https://ui5.sap.com';
+const CACHE_DIR = join(homedir(), '.cache', 'fiori-mcp-ui5-doc');
```
nikmace: **Cache has no eviction**

Files in `~/.cache/fiori-mcp-ui5-doc/` accumulate indefinitely. Expired files are overwritten on re-fetch of the same base × version × library key, but files for keys you no longer use (old versions, switched projects) are never deleted.

The in-memory `classLibraryCache` has the same issue: a null result (e.g. CDN unreachable at lookup time) is cached for the entire process lifetime with no TTL or eviction. A subsequent call for the same class will silently return a shorter inheritance chain even if the CDN becomes reachable again.

Suggestions: add a max-entries bound or TTL to `classLibraryCache`, and either document that `~/.cache/fiori-mcp-ui5-doc/` is safe to delete manually, or add a periodic cleanup of files older than N days.

nikmace (changes requested):

packages/fiori-mcp-server/src/tools/lookup-ui5-documentation/api-json.ts
mmilko01: I think we should leave this as-is. We can't tell which library owns a class from its name, so we try prefixes in order — but it's capped at 4 tries and only happens on a cold start, since the 24h disk cache handles every call after the first. Not worth adding complexity for a one-time cost.

packages/fiori-mcp-server/src/types/input.ts
```diff
@@ -206,3 +206,44 @@ export const AdpMetadataInputSchema = zod.object({
...
+});
+
+export const ODataServiceInputSchema = zod.object({
+    appPath: zod
+        .string()
+        .describe(
+            'Absolute path to the adaptation project root directory (where webapp/manifest.appdescr_variant resides).'
+        ),
+    filter: zod.string().optional().describe('Filter string to match service names')
+});
```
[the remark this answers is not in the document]
mmilko01: Fixed with b798a077a

packages/fiori-mcp-server/src/tools/lookup-ui5-documentation/api-json.ts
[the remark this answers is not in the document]
mmilko01: Fixed with 3e0dcd872

packages/fiori-mcp-server/src/tools/lookup-ui5-documentation/lookups/aggregation.ts
mmilko01: Fixed with b4c817997

packages/fiori-mcp-server/src/tools/lookup-ui5-documentation/lookups/aggregation.ts
mmilko01: Fixed with ac8b76fa8

packages/fiori-mcp-server/src/types/input.ts
mmilko01: Fixed with 8bd45fd86

packages/fiori-mcp-server/src/tools/lookup-ui5-documentation/api-json.ts
mmilko01: Fixed with 7267102ab

packages/fiori-mcp-server/src/tools/lookup-ui5-documentation/api-json.ts
mmilko01: Fixed with 3e0dcd872

nikmace (approved):
Changes look good
Comments addressed

PR #5164, opened 2026-09-10
fix(adp-tooling): route UI5 version requests through proxy-aware HTTP layer

### LINKED ISSUE #5163: BUG - Adaptation Project generator cannot validate UI5 versions behind a corporate proxy
### Description
When VS Code runs behind a corporate HTTP proxy, the **SAPUI5 Adaptation Project** generator cannot validate any UI5 version. Every version (system version and latest) is rejected with:

> The SAPUI5 version you have selected is not compatible with the SAPUI5 Adaptation Editor. Please select a different version.

Root cause: the generator's UI5-CDN calls bypass the proxy-aware HTTP layer used elsewhere in the tooling. `validateUI5VersionExists` used a raw `axios.get`, and the version-list calls used the native Node `fetch`:
- Raw axios relies on axios's built-in proxy handling, which builds a `CONNECT` request (`Host` without port) that strict corporate proxies reject. The resulting error is mapped to the "not compatible" message.
- Node's native `fetch` ignores `HTTP(S)_PROXY` entirely, so in a proxy-only network those calls cannot reach the CDN.

### Steps to Reproduce
1. Run VS Code on a machine whose only internet access is a strict corporate HTTP proxy (`HTTPS_PROXY`/`HTTP_PROXY` set).
2. Start the SAPUI5 Adaptation Project generator; select a system and application.
3. Reach the UI5 version step.
4. Every version is rejected with "not compatible with the SAPUI5 Adaptation Editor".

### Expected results
The version validates through the proxy and the wizard proceeds (as it does without a proxy, and as the tooling's other requests already do).

### Actual results
All versions rejected with the "not compatible" error; the wizard cannot continue.

- Components: `@sap-ux/adp-tooling` (`@sap-ux/generator-adp`), `@sap-ux/axios-extension`
- Reported on Node 24.x, VS Code (desktop), plain unauthenticated corporate HTTP proxy, with `GLOBAL_AGENT_*` / `HTTP(S)_PROXY` / `NO_PROXY` env vars set.
OS:
- [x] Windows
- [ ] Mac OS
- [x] Other (any desktop behind a strict HTTP proxy)

### Problem
ADP UI5-version requests to the SAP CDN bypassed the proxy-aware HTTP layer: `validator.ts` used raw `axios.get` (axios's built-in proxy CONNECT is rejected by strict proxies) and `fetch.ts` used native `fetch` (which ignores proxy env vars). The proxy rejection surfaced to the user as "version not compatible".

### Fix
Route the ADP UI5-version calls through the existing proxy handling. Extracted the proxy logic from `axios-extension`'s `factory.ts` into a reusable `getProxyAgentConfig(url)` and used it in `validator.ts` and `fetch.ts` (the latter switched from native `fetch` to axios). Honours `NO_PROXY` and is skipped in SAP Business Application Studio.

### Why was it missed
The proxy handling lived only inside the `axios-extension` service-provider factory; these UI5-CDN calls were written as standalone `axios`/`fetch` calls and were only exercised in environments with direct internet or a lenient proxy. The proxy rejection is indistinguishable, to the validator, from the CDN legitimately reporting an unknown version.

### How can we avoid this
Prefer the shared proxy-aware HTTP helper for all outbound requests; avoid native `fetch` for requests that must honour proxy configuration; add proxy-path test coverage.

mmilko01:
## Description
Behind a corporate HTTP proxy, the SAPUI5 Adaptation Project generator rejects every UI5 version with *"The SAPUI5 version you have selected is not compatible with the SAPUI5 Adaptation Editor."*

The generator's UI5-CDN calls bypassed the proxy-aware HTTP layer: `validateUI5VersionExists` used a raw `axios.get` (axios's built-in proxy handling builds a `CONNECT` that strict proxies reject), and the version-list calls used Node's native `fetch` (which ignores `HTTP(S)_PROXY`). This routes them through the same proxy handling `@sap-ux/axios-extension` already uses everywhere else.

Changes:
- `@sap-ux/axios-extension`: extract the proxy logic from `factory.ts` into a new `proxy.ts` exporting **`getProxyAgentConfig(url, agentOptions?)`** (`getProxyForUrl` + `PatchedHttpsProxyAgent`/`HttpProxyAgent` + `proxy: false`, honouring `NO_PROXY`, skipped in SAP Business Application Studio). The factory interceptor now consumes it — no behaviour change. `PatchedHttpsProxyAgent` moved with it and is still re-exported.
- `@sap-ux/adp-tooling`: `validator.ts` and `fetch.ts` now use `getProxyAgentConfig`; the two `fetch()` calls in `fetch.ts` are converted to proxy-aware `axios.get`.

Fixes #5163

## Type of change

## How have you tested?
- Updated unit tests: `adp-tooling` `fetch.test.ts` (mocks axios) and `validator.test.ts`; added `getProxyAgentConfig` branch tests in `axios-extension` `factory.test.ts` (proxy set / `NO_PROXY` bypass / App Studio). All `@sap-ux/adp-tooling` unit tests pass (810).
- Reproduced the failure/fix mechanism in plain Node against a proxy that emulates a strict gateway: the raw-axios (built-in proxy) request is rejected, while the explicit-agent request (this fix) is accepted — matching the approach the customer independently validated against their real proxy.

## Checklist:
- [ ] I have reviewed and addressed all Hyperspace bot findings (or explicitly explained dismissals)
- [x] I have done an Agentic review

### DIFF
--- a/packages/adp-tooling/src/ui5/fetch.ts
+++ b/packages/adp-tooling/src/ui5/fetch.ts
@@ -1,4 +1,6 @@
 import type { ToolsLogger } from '@sap-ux/logger';
+import axios from 'axios';
+import { getProxyAgentConfig } from '@sap-ux/axios-extension';
 
 import type { UI5Version } from '../types.js';
 import { buildFallbackMap } from './format.js';
@@ -13,11 +15,8 @@ import { UI5_VERSIONS_CDN_URL, UI5_VERSIONS_NEO_CDN_URL, LATEST_VERSION } from '
  */
 export async function fetchPublicVersions(logger?: ToolsLogger): Promise<UI5Version> {
     try {
-        const response = await fetch(UI5_VERSIONS_CDN_URL);
-        if (!response.ok) {
-            throw new Error(`Failed to fetch public UI5 versions. Status: ${response.status}`);
-        }
-        return (await response.json()) as UI5Version;
+        const { data } = await axios.get<UI5Version>(UI5_VERSIONS_CDN_URL, getProxyAgentConfig(UI5_VERSIONS_CDN_URL));
+        return data;
     } catch (e) {
         logger?.warn(
             '[ui5-info] Falling back to built-in UI5 version list: ' + (e instanceof Error ? e.message : String(e))
@@ -33,8 +32,7 @@ export async function fetchPublicVersions(logger?: ToolsLogger): Promise<UI5Vers
  * @returns {Promise<string[]>} A promise that resolves to an array of formatted internal version strings.
  */
 export async function fetchInternalVersions(latestVersion: string): Promise<string[]> {
-    const response = await fetch(UI5_VERSIONS_NEO_CDN_URL);
-    const data = await response.json();
+    const { data } = await axios.get(UI5_VERSIONS_NEO_CDN_URL, getProxyAgentConfig(UI5_VERSIONS_NEO_CDN_URL));
 
     return data?.routes?.map((route: { target: { version: string } }) => {
         return route.target.version === latestVersion
--- a/packages/adp-tooling/src/ui5/fetch.ts
+++ b/packages/adp-tooling/src/ui5/fetch.ts
@@ -32,11 +32,17 @@ export async function fetchPublicVersions(logger?: ToolsLogger): Promise<UI5Vers
  * @returns {Promise<string[]>} A promise that resolves to an array of formatted internal version strings.
  */
 export async function fetchInternalVersions(latestVersion: string): Promise<string[]> {
-    const { data } = await axios.get(UI5_VERSIONS_NEO_CDN_URL, getProxyAgentConfig(UI5_VERSIONS_NEO_CDN_URL));
+    try {
+        const { data } = await axios.get(UI5_VERSIONS_NEO_CDN_URL, getProxyAgentConfig(UI5_VERSIONS_NEO_CDN_URL));
 
-    return data?.routes?.map((route: { target: { version: string } }) => {
-        return route.target.version === latestVersion
-            ? `${route.target.version} ${LATEST_VERSION}`
-            : route.target.version;
-    });
+        return (
+            data?.routes?.map((route: { target: { version: string } }) => {
+                return route.target.version === latestVersion
+                    ? `${route.target.version} ${LATEST_VERSION}`
+                    : route.target.version;
+            }) ?? []
+        );
+    } catch {
+        return [];
+    }
 }
--- a/packages/adp-tooling/src/ui5/fetch.ts
+++ b/packages/adp-tooling/src/ui5/fetch.ts
@@ -1,6 +1,6 @@
 import type { ToolsLogger } from '@sap-ux/logger';
 import axios from 'axios';
-import { getProxyAgentConfig } from '@sap-ux/axios-extension';
+import { getProxyAgentConfig, isAxiosError } from '@sap-ux/axios-extension';
 
 import type { UI5Version } from '../types.js';
 import { buildFallbackMap } from './format.js';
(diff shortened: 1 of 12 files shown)


packages/adp-tooling/src/ui5/fetch.ts
```diff
@@ -33,8 +32,7 @@ export async function fetchPublicVersions(logger?: ToolsLogger): Promise<UI5Vers
  * @returns {Promise<string[]>} A promise that resolves to an array of formatted internal version strings.
  */
 export async function fetchInternalVersions(latestVersion: string): Promise<string[]> {
-    const response = await fetch(UI5_VERSIONS_NEO_CDN_URL);
-    const data = await response.json();
+    const { data } = await axios.get(UI5_VERSIONS_NEO_CDN_URL, getProxyAgentConfig(UI5_VERSIONS_NEO_CDN_URL));
```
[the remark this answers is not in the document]
mmilko01: Good catch — addressed in 3c214c14b8d3b7981a61b081f140c0b5231bf406: `fetchInternalVersions` now wraps the request in try/catch and returns an empty list on failure, consistent with `fetchPublicVersions`. Added a unit test for the failure path.

nikmace (dismissed):
Changes look good
Coverage is good
Changeset is OK
Did not test manually

cianmSAP (dismissed):
Update looks good, factory sitll behaves as before
Good testing
Changeset looks appropriate

packages/adp-tooling/src/ui5/fetch.ts
```diff
@@ -33,12 +32,21 @@ export async function fetchPublicVersions(logger?: ToolsLogger): Promise<UI5Vers
...
  */
 export async function fetchInternalVersions(latestVersion: string): Promise<string[]> {
-    const response = await fetch(UI5_VERSIONS_NEO_CDN_URL);
-    const data = await response.json();
+    let data;
+    try {
+        ({ data } = await axios.get(UI5_VERSIONS_NEO_CDN_URL, getProxyAgentConfig(UI5_VERSIONS_NEO_CDN_URL)));
+    } catch (e) {
+        if (isAxiosError(e) && e.response) {
+            return [];
```
heimwege, line 44: The old code would have propagated any error. The new silent-empty behaviour may confuse users who get no internal versions and no explanation. Compare with `fetchPublicVersions` which at least calls `logger?.warn(...)` on failure. Consider adding a logger parameter (matching `fetchPublicVersions`) and logging the swallowed error, or at minimum accepting this is intentional and adding a comment.

packages/axios-extension/src/proxy.ts
```diff
@@ -0,0 +1,58 @@
...
+import { type AgentOptions } from 'node:https';
+import { type HttpsProxyAgentOptions, HttpsProxyAgent } from 'https-proxy-agent';
+import { getProxyForUrl } from 'proxy-from-env';
+import { isAppStudio } from '@sap-ux/btp-utils';
+
+/**
+ * PatchedHttpsProxyAgent is a custom implementation of HttpsProxyAgent that allows to pass additional options, currently not supported by the original implementation when calling tls.connect
+ */
+export class PatchedHttpsProxyAgent<Uri extends string> extends HttpsProxyAgent<Uri> {
+    private readonly extraOptions: any;
```
heimwege: General: Why all these `any` types? Since the code is being relocated and re-exported as a new public API, it's a good opportunity to type it properly

heimwege (commented):
see comments

packages/adp-tooling/src/ui5/fetch.ts
mmilko01: My thinking was to actually keep the old behavior. Previously, the implementation used `fetch` with doesn't throw on responses like 400, 404, 500, etc. It only throws on DNS/CORS issues. I agree that logging is a good idea so I added it in commit - 33c665cb7149faa82a60e1c2e5ad22ff398bb432

packages/axios-extension/src/proxy.ts
mmilko01: Good point! I added types where `any` was used. It is not completely ideal, because type `AgentConnectOpts` is now described a bit ugly, but otherwise I will need to introduce new dependency to `axios-extension` - `agent-base`, which I prefer to avoid. Fixed in: 7e33ec0b37cdd03bdf5922a9f6226542c35dc4a2 and 5e5cf1d60248bd0a275055754f29073221149719

heimwege (approved):
- review comments addressed
- coverage is good
- did not test locally

nikmace (approved):
Re-approve

cianmSAP (approved):
Re-approving

PR #5176, opened 2026-09-11
Update navigation handling for multiple properties targeting the same entity set

marcovieth:
Summary from AI

In FE V4 apps, the "Add Subpage" quick action was incorrectly disabled when multiple navigation properties all target the same entity set and at least one of them already has a subpage configured
Root cause: prepareNavigationData called isPageExists(targetEntitySet) which only matched by entity set name — so a page existing for _Subtype1 → Child01 would block _ZZSubtype13 → Child01 as well
Fix: replace the entity-set check with a route-pattern check — a nav property is only skipped if the manifest already contains a route whose pattern includes /{navProperty}(
Changes

packages/preview-middleware-client/src/adp/quick-actions/fe-v4/add-new-subpage.ts — prepareNavigationData now checks manifest routes instead of entity set names; nav properties that already have a route are filtered out, all others are offered
packages/preview-middleware-client/test/unit/adp/quick-actions/fe-v4.test.ts — updated isNewPageUnavailable object-page test case to include the _BookSupplement route (a target page must have a route — the test was incomplete); added a dedicated regression test for the multi-nav-prop-to-same-entity-set scenario
Test plan

 pnpm --filter @sap-ux-private/preview-middleware-client test — all 63 tests pass
 Manually verified with hcm.fio.myfamilymemberss1 on QKZ: "Add Subpage" is now enabled and offers _ZZSubtype13 as an option


Jimmy-Joseph19 (approved):
Changes looks good.
Changeset found.
Test coverage ok.
Tested with int test https://github.com/SAP/open-ux-tools/actions/runs/34827724531

vadson71 (approved):
Code changes look good
Thoroughly covered by tests
Tested locally

PR #5181, opened 2026-09-14
fix: use password type for smartlinks credential prompt

mustafamengutay-sap:
## Problem

When running `@sap-ux/create add smartlinks-config` and entering credentials, the password field gave no visual feedback. Users couldn't tell how many characters they had typed.

## Proposed Solution

Changed `type: 'invisible'` to `type: 'password'` in `src/prompt/smartlinks-config.ts`.

The `password` type in the `prompts` library masks each character with `*`, confirming keystrokes without exposing the actual value.


heimwege (approved):
lgtm

vinayhospete (approved):

PR #5197, opened 2026-09-17
feat(ui5-test-writer): Reset mockdata before each Journey

sap-sebelao:
Internal JIRA
FXUBRQ34-12762


dudajSAP (approved):
Looks good to me.
Not tested manually.

I was confused, why `iResetMockData` and `iResetTestData` should be part of OpaJourneyTypes.d.ts, but it's done due to .and. chaining -> so it's fine for this PR.

docirl (approved):

PR #5200, opened 2026-09-18
feat(project-access): support fallbackLocale in i18n path resolution

mustafamengutay-sap:
UI5 allows configuring `fallbackLocale` in `sap.app.i18n`. Previously, Fiori Tools ignored this field and caused false "key not available" warnings when a key only existed in the fallback locale file.

This PR resolves and merges fallback locale entries into the primary i18n bundle, so keys defined only in the fallback file are visible to consumers. Primary file entries always win on key collision.


packages/project-access/src/project/i18n/i18n.ts
```diff
@@ -92,32 +170,25 @@ function getI18nAppPath(manifest: Manifest): string {
...
-                continue;
-            }
-            if (i18nModel.settings.bundleUrl) {
-                result[modelKey] = { path: join(i18nModel.settings.bundleUrl) };
-                continue;
             }
+            continue;
         }
         if (i18nModel.uri) {
             result[modelKey] = { path: join(i18nModel.uri) };
```
devinea, line 194: **Regression: models with `settings` but neither `bundleName` nor `bundleUrl` (e.g. only `enhanceWith`) silently lose their `uri` fallback.**

The unconditional `continue` here was added to replace the two earlier `continue` statements inside the `bundleName`/`bundleUrl` branches. But those inner `continue`s only fired when a path was actually found; this one fires for _every_ model that has a `settings` object — including those where `extractBundlePath` returns `undefined` and the model's `uri` property should still be consulted.

Before this PR, a model like:
```json
{
  "type": "sap.ui.model.resource.ResourceModel",
  "settings": { "enhanceWith": [...] },
  "uri": "i18n/i18n.properties"
}
```
would fall through to the `if (i18nModel.uri)` check and resolve correctly. Now it is silently dropped.

Fix: only `continue` when `path` is truthy (i.e. move the `continue` inside `if (path) { ... }`), so the `uri` fallback below is still reachable when `extractBundlePath` returns `undefined`.

packages/project-access/src/project/i18n/i18n.ts
```diff
@@ -39,10 +52,46 @@ export async function getI18nPropertiesPaths(
...
+ * @param locale - fallback locale string (e.g. "en", "de")
+ * @returns - path to the fallback locale file
+ */
+function computeFallbackLocalePath(basePath: string, locale: string): string {
+    const ext = '.properties';
+    const stem = basePath.endsWith(ext) ? basePath.slice(0, -ext.length) : basePath;
+    return `${stem}_${locale}${ext}`;
+}
+
+/**
```
devinea, line 80: **Security: `fallbackLocale` is not validated before being embedded in a file path, enabling path traversal.**

`computeFallbackLocalePath` concatenates the caller-supplied `locale` string directly into the path without any sanitization. A manifest containing:
```json
{ "sap.app": { "i18n": { "fallbackLocale": "../../../../etc/passwd", "bundleUrl": "i18n/i18n" } } }
```
produces `i18n/i18n_../../../../etc/passwd.properties`, which after `path.join(manifestFolder, ...)` normalises to an absolute path outside the project root. `getPropertiesI18nBundle` will then attempt to read that path; if it exists, its contents are silently merged into the app's i18n bundle.

Recommended fix: validate that `fallbackLocale` only contains safe locale characters before use, e.g.:
```typescript
if (!/^[A-Za-z0-9_-]+$/.test(locale)) {
    return basePath; // or throw/log
}
```

packages/project-access/src/project/i18n/read.ts
```diff
@@ -45,6 +43,20 @@ export async function getI18nBundles(
...
     }
 
+    if (i18nPropertiesPaths['sap.app.fallbackLocale']) {
+        try {
+            const fallbackBundle = await getPropertiesI18nBundle(i18nPropertiesPaths['sap.app.fallbackLocale'], fs);
+            // Fallback entries supplement the primary bundle; primary entries take precedence on key collision
+            result['sap.app'] = { ...fallbackBundle, ...result['sap.app'] };
+            // Fallback recovered the data — the primary-file error is no longer relevant
+            delete result.errors?.['sap.app'];
+        } catch (error) {
```
devinea: **Bug: non-ENOENT primary-file error is silently cleared when the fallback loads successfully.**

`delete result.errors?.['sap.app']` runs unconditionally whenever the fallback read succeeds. If the primary file failed with a non-ENOENT error (e.g. `EACCES`, `EPERM`), that error is stored in `result.errors` and then immediately discarded here. The caller receives a bundle populated from the fallback only, with no error entry, giving the false impression that the primary file is accessible and healthy.

The `delete` should be guarded so it only fires when the primary error _was_ `ENOENT` — i.e. the file was genuinely missing and the fallback is a complete substitute — not when there is an underlying permission or I/O problem:

```typescript
const primaryError = result.errors?.['sap.app'] as NodeJS.ErrnoException | undefined;
if (primaryError?.code === 'ENOENT') {
    delete result.errors!['sap.app'];
}
```

The same issue exists in the models fallback block below.

packages/project-access/src/project/i18n/read.ts
```diff
@@ -54,6 +66,20 @@ export async function getI18nBundles(
...
+        const fallbackLocalePath = i18nPropertiesPaths.models[key].fallbackLocalePath;
+        if (fallbackLocalePath) {
+            try {
+                const fallbackBundle = await getPropertiesI18nBundle(fallbackLocalePath, fs);
+                result.models[key] = { ...fallbackBundle, ...result.models[key] };
+                // Fallback recovered the data — the primary-file error is no longer relevant
+                delete result.errors?.[`models.${key}`];
+            } catch (error) {
+                if ((error as NodeJS.ErrnoException).code !== 'ENOENT') {
+                    throw error;
```
devinea: **Bug: if the fallback read also throws a non-ENOENT error, the original primary-file error is permanently lost.**

The fallback block runs regardless of _why_ the primary read failed. If the primary threw `EPERM` (stored in `result.errors`) and then the fallback throws `ENETDOWN`, this `throw error` propagates `ENETDOWN` as an unhandled rejection. The caller sees `ENETDOWN` with no trace of the original permission issue, and the partial `result` (with any models that had already been read) is also discarded.

Two fixes are needed:
1. Only attempt the fallback when the primary error was `ENOENT` — other error codes indicate the file exists but is unreadable, and a fallback cannot substitute for that.
2. If the fallback itself fails with a non-ENOENT error, store it in `result.errors` (like primary errors are) rather than throwing, to preserve the function's existing soft-error contract.

The same issue exists in the `sap.app` fallback block above.

devinea (commented):
Follow-up review: all 4 previously raised issues are now fixed. The following issues were found during re-review.

- Finding in `read.ts` is a regression introduced by the fix commit.
- Findings in `i18n.ts` and `access.test.ts` are pre-existing issues (present before the fix commit) surfaced during re-review.

packages/project-access/src/project/i18n/read.ts
```diff
@@ -11,12 +12,49 @@ import type { Editor } from 'mem-fs-editor';
...
+): Promise<I18nBundle> {
+    const primaryError = result.errors?.[primaryKey] as NodeJS.ErrnoException | undefined;
+    if (primaryError && primaryError.code !== 'ENOENT') {
+        return current;
+    }
+    try {
+        const fallbackBundle = await getPropertiesI18nBundle(fallbackPath, fs);
+        if (primaryError?.code === 'ENOENT') {
+            delete result.errors![primaryKey];
+        }
```
devinea, line 50: **Bug (introduced in this fix commit): `result.errors` left as `{}` (truthy) after the only error key is deleted.**

After a successful fallback read, `delete result.errors![primaryKey]` removes the last key but leaves `result.errors` as an empty object `{}` rather than `undefined`. Any consumer that guards with `if (result.errors)` will enter the error-handling branch and iterate over an empty object, producing spurious logging or false "errors present" signals.

Fix:
```ts
delete result.errors![primaryKey];
if (Object.keys(result.errors!).length === 0) {
    result.errors = undefined;
}
```

packages/project-access/src/project/i18n/i18n.ts
```diff
@@ -39,10 +52,46 @@ export async function getI18nPropertiesPaths(
...
+ * Get the fallback locale i18n path from sap.app.i18n part of the manifest.
+ *
+ * @param manifest - parsed content of manifest.json
+ * @param appPath - resolved base i18n path for sap.app
+ * @returns - fallback locale path, or undefined if not configured
+ */
+function getI18nAppFallbackLocalePath(manifest: Manifest, appPath: string): string | undefined {
```
devinea, line 87: **Bug (pre-existing): `typeof null === 'object'` causes a `TypeError` crash when `i18n` is `null`.**

`typeof manifest?.['sap.app']?.i18n === 'object'` passes when `i18n` is `null` because `typeof null === 'object'`. Accessing `.fallbackLocale` on `null` then throws `TypeError: Cannot read properties of null`.

Fix: add an explicit null guard:
```ts
const i18n = manifest?.['sap.app']?.i18n;
if (i18n !== null && typeof i18n === 'object') { ... }
```

The same pattern exists in `getI18nAppPath` (pre-existing), but this new function replicates it.

packages/project-access/src/project/i18n/i18n.ts
```diff
@@ -65,9 +114,11 @@ function getI18nAppPath(manifest: Manifest): string {
     if (typeof manifest?.['sap.app']?.i18n === 'object') {
         // bundleName wins over `bundleUrl`
         if ('bundleName' in manifest['sap.app'].i18n) {
-            // module name is in dot notation
-            const withoutAppId = manifest['sap.app'].i18n.bundleName.replace(manifest['sap.app'].id, '');
-            const i18nPath = `${join(...withoutAppId.split('.'))}.properties`;
+            // module name is in dot notation; strip appId as a prefix
+            const bundleName = manifest['sap.app'].i18n.bundleName;
```
devinea: **Bug (pre-existing): malformed path `'.properties'` when `bundleName === appId` exactly.**

When `bundleName` equals `appId` exactly (e.g. both `'sample.app'`), `suffix` is `''`, `suffix.split('.')` is `['']`, `path.join('')` is `'.'`, and the final path becomes `'.properties'`. `computeFallbackLocalePath` then produces `'._en.properties'`. This was a pre-existing bug in `getI18nAppPath` that `extractBundlePath` replicates.

packages/project-access/src/project/i18n/i18n.ts
```diff
@@ -79,6 +130,33 @@ function getI18nAppPath(manifest: Manifest): string {
...
+ * Resolves the i18n file path from a resource model's settings object.
+ *
+ * @param appId - application id from sap.app namespace
+ * @param settings - resource model settings object
+ * @param settings.bundleName - bundle name in dot notation (e.g. "sample.app.i18n")
+ * @param settings.bundleUrl - bundle URL path (e.g. "i18n/i18n.properties")
+ * @returns - resolved relative path, or undefined if it cannot be determined
+ */
+function extractBundlePath(appId: string, settings: { bundleName?: string; bundleUrl?: string }): string | undefined {
+    if (settings.bundleName) {
```
devinea, line 148: **Bug (pre-existing): `extractBundlePath` silently drops the model when `appId` is an empty string.**

When `bundleName` is set but `appId` is an empty string, `if (!appId) { return undefined; }` causes `extractBundlePath` to return `undefined` early. The model is then silently omitted if it also has no `uri`. The comment "fall back to full bundleName otherwise" only applies in the `startsWith` branch — the `!appId` guard does not fall back.

Consider using `bundleName` directly as the path when `appId` is absent, rather than returning `undefined`.

packages/project-access/test/project/access.test.ts
```diff
@@ -15,6 +15,7 @@ const mockCreateAnnotationI18nEntries = jest.fn<typeof i18nWriteType.createAnnot
 const mockCreateUI5I18nEntries = jest.fn<typeof i18nWriteType.createUI5I18nEntries>();
 const mockCreateManifestI18nEntries = jest.fn<typeof i18nWriteType.createManifestI18nEntries>();
 const mockCreateCapI18nEntries = jest.fn<typeof i18nWriteType.createCapI18nEntries>();
+const mockCreateI18nEntriesAtPath = jest.fn<typeof i18nWriteType.createI18nEntriesAtPath>();
```
devinea, line 18: **Test coverage (pre-existing): `createI18nEntriesAtPath` is mocked but never called through the `ApplicationAccess` interface.**

`mockCreateI18nEntriesAtPath` is registered here but no test case invokes `applicationAccess.createI18nEntriesAtPath(...)` through the `ApplicationAccessImp` public interface. The `write.ts` unit tests cover the function in isolation, but the forwarding of `this.project.root` and `this.options?.fs` through `ApplicationAccessImp` has no coverage — a regression in that wiring would go undetected.

PR #5203, opened 2026-09-21
fix(ui5-test-writer): Tests for actions with parameters & deselect rows after test

sap-sebelao:
Internal JIRA
FXUBRQ34-12787
FXUBRQ34-12789


sap-sebelao:
> I wasn’t able to post inline comments before the prompt limit. The main issue I found is that the new “deselect” calls use `iSelectRows(0)`, which selects the row rather than clearing selection; this likely leaves the table state unchanged and can still affect following actions. Consider using the FE test API’s row state parameter (for example `iSelectRows(0, { selected: false })`, if supported by the target UI5 version/API) or the appropriate deselection helper in all added deselect locations.

Looking into the behavior of `iSelectRows()` is something I've done at the very beginning of this implementation, as the name implies the function only selects. This is not the case, the action simply presses the row selector, regardless of it's state, meaning that it will toggle the selection state of the row. The naming is just unfortunate.

Furthermore the suggested replacement `iSelectRows(0, { selected: false })` is nonsense, the function does not accept any second argument.

No action needed here, it works as expected.

dudajSAP:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

dudajSAP (approved):

docirl (approved):

PR #5205, opened 2026-09-21
fix: Error `Cannot read properties of undefined (reading 'trackEvent')` is thrown during `reportRuntimeError` call

815are:
Following error is thrown during `reportRuntimeError`:
```
TypeError: Cannot read properties of undefined (reading 'trackEvent')
```

while debugging issue, root issue is that instrumentation key is not accepted by `appInsights.TelemetryClient`
image

As we noticed reason is that we need call wrapper function [ClientFactory.buildConnectionString](https://github.com/SAP/open-ux-tools/blob/main/packages/telemetry/src/base/client/index.ts#L47) before initializing [appInsights.TelemetryClient](https://github.com/SAP/open-ux-tools/blob/main/packages/telemetry/src/base/utils/reporting.ts#L51)


devinea (approved):
Thanks @815are 
code changes look correct for missing call to buildConnectionString when creating new telemetry client.

PetrStribrny (approved):

